PR #53895, opened 2024-12-19
registerProductBlockType: Update signature and usages

tjcafferkey:

We've been experiencing reports that the function signature for `registerProductBlock` (or rather `registerBlockSingleProductTemplate` which was its former name) is confusing because it requires `blockName` and `blockMetadata`.

This PR aims to refactor the function so that the function signature is more aligned with Gutenbergs native `registerBlockType` with the addition of our custom properties:

* `isVariationBlock`
* `variationName`
* `isAvailableOnPostEditor`

For example what was once:

```
registerBlockSingleProductTemplate( {
	blockName: metadata.name,
	// @ts-expect-error: `metadata` currently does not have a type definition in WordPress core
	blockMetadata: metadata,
	blockSettings: {
		icon,
		edit,
		save,
	},
	isAvailableOnPostEditor: false,
} );
```

is now:

```
registerProductBlockType( metadata, {
	isAvailableOnPostEditor: false,
} );
```

1. Verify blocks using `registerProductBlockType` registers correctly in both single product templates and post/page editor. **Please check various blocks here including ones we've deprecated such as Products (Beta)**
2. Check transitions between templates using both the command UI (CMD+K) and manual navigation maintain correct block behavior
3. Verify existing block implementations continue to work without changes
4. Ensure E2E tests pass

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce-blocks/assets/js/blocks/product-query/variations/related-products.tsx
```diff
@@ -117,31 +117,30 @@ export const INNER_BLOCKS_TEMPLATE: InnerBlockTemplate[] = [
 	],
 ];
 
-registerProductBlockType( {
-	blockName: QUERY_LOOP_ID,
-	blockMetadata: {},
-	blockSettings: {
-		description: __( 'Display related products.', 'woocommerce' ),
-		name: 'Related Products Controls',
```
tjcafferkey, line 125: I don't even know why this is here, as far as I know `name` is the block nomenclature not the title. But what makes this move confusing is that `title` is also defined right below it.

plugins/woocommerce-blocks/assets/js/atomic/utils/register-product-block-type.ts
```diff
@@ -14,25 +14,30 @@ import { subscribe, select } from '@wordpress/data';
...
+ * @property {boolean}                     [isVariationBlock]      - Whether this block is a variation
+ * @property {string}                      [variationName]         - The name of the variation if applicable
+ * @property {boolean}                     isAvailableOnPostEditor - Whether the block should be available in post editor
  */
-type BlockConfig = {
-	blockName: string;
-	blockMetadata?: string | Partial< BlockConfiguration >;
-	blockSettings: Partial< BlockConfiguration >;
+type ProductBlockRegistrationConfig = {
+	settings?: Partial< BlockConfiguration >;
```
kmanijak: `ProductBlockRegistrationConfig` is used to type `settings`:

```
export const registerProductBlockType = (
	blockNameOrMetadata: string | Partial< BlockConfiguration >,
	settings?: ProductBlockRegistrationConfig
): void => {
```

so if I get it right, it should not contain `settings`

```suggestion
```

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/related-products/index.tsx
```diff
@@ -11,14 +11,14 @@ import edit from './edit';
...
+	...metadata,
+	icon: { src: icon },
+	edit,
+	save,
+	isAvailableOnPostEditor: false,
+};
+
+registerProductBlockType( blockConfig, {
 	isAvailableOnPostEditor: false,
 } );
```
kmanijak, line 24: `isAvailableOnPostEditor` is `false` by default. I think we can simplify these declarations even more and remove the settings wherever `isAvailableOnPostEditor: false` is passed.

```suggestion
registerProductBlockType( blockConfig );
```

kmanijak (approved):
I didn't spot any issues! Tested the block and variations in scope of
- Single Product template - they are registered globally
- posts - they are registered with the ancestors
- patterns - they are registered with the ancestors
- classic theme - widgets - nothing crashes 😅 

Great job! 🚀 Left some minor comments but pre-approving.

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/related-products/index.tsx
tjcafferkey: We _could_ do but I intentionally left that in so it's more declarative, but good spot!

plugins/woocommerce-blocks/assets/js/atomic/utils/register-product-block-type.ts
tjcafferkey: Ah yes good spot. We still need settings but not as its own key. I've updated here https://github.com/woocommerce/woocommerce/pull/53895/commits/83e6febb983700f5967570eb3dda238e7352f3fd

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/related-products/index.tsx
tjcafferkey: PS. Im inclined to leave it like it is now, so its limitations are clear.

kmanijak: Yeah I saw you even add it where is wasn't there initially. I don't have a strong opinion. Actually I get it and agree. But then, maybe it should not be optional. Because it seems it's not 😄 It's a non-blocking detail though.

tjcafferkey: Yeah good point about it being optional. This setting was historical optional so I kept it that way to simply the refactor for now.

plugins/woocommerce-blocks/assets/js/atomic/utils/register-product-block-type.ts
```diff
@@ -14,34 +14,53 @@ import { subscribe, select } from '@wordpress/data';
...
+ * Internal block config type used by the BlockRegistrationManager
+ *
+ * @typedef {Object} ProductBlockConfig
+ * @property {string}                      blockName            - The name of the block
+ * @property {Partial<BlockConfiguration>} settings             - Block settings configuration
+ * @property {ProductBlockSettings}        productBlockSettings - Product block settings
+ */
+type ProductBlockConfig = ProductBlockSettings & {
+	blockName: string;
+	settings: Partial< BlockConfiguration >;
```
gigitux, line 40: I am not convinced that this should be partial given that `ProductBlockSettings` has already optional keys

plugins/woocommerce-blocks/assets/js/atomic/utils/register-product-block-type.ts
```diff
@@ -14,34 +14,53 @@ import { subscribe, select } from '@wordpress/data';
...
+/**
+ * Configuration object for registering a product block type.
+ *
+ * @typedef {Object} ProductBlockRegistrationConfig
+ * @property {Partial<BlockConfiguration>} settings                - Block settings configuration
+ * @property {boolean}                     [isVariationBlock]      - Whether this block is a variation
+ * @property {string}                      [variationName]         - The name of the variation if applicable
+ * @property {boolean}                     isAvailableOnPostEditor - Whether the block should be available in post editor
+ */
+type ProductBlockRegistrationConfig = Partial< BlockConfiguration > &
```
gigitux, line 52: Same here

plugins/woocommerce-blocks/assets/js/atomic/utils/register-product-block-type.ts
gigitux: If we want to keep the `isAvailableOnPostEditor` optional, let's update the `ProductBlockSettings`:

```ts
type ProductBlockSettings = {
	isVariationBlock?: boolean;
	variationName?: string | undefined;
	isAvailableOnPostEditor?: boolean;
};
```

gigitux (approved):
Nice refactor! I like the idea of streamlining this function to the `registerBlockType` function!

I left a few NIT comments, mostly to improve the clarity around TS types. Feel free to skip them!

plugins/woocommerce-blocks/assets/js/atomic/utils/register-product-block-type.ts
tjcafferkey: >I am not convinced that this should be partial given that ProductBlockSettings has already optional keys

@gigitux can you explain please? `BlockConfiguration` is Gutenbergs block config.

gigitux: Ok, I missed that. All good :D Sorry for the confusion!

PR #49311, opened 2024-07-09
Ensure local state is not overwritten by server when using `extensionCartUpdate`

opr:
### Changes proposed in this Pull Request:

~~Add new property to the `wc/store/cart` data store. The new property is in the `metaData` object and is called `isCustomerDataDirty`. This will track whether an address change is pending (i.e. not pushed to the server yet). When a request to update customer data completes, `isCustomerDataDirty` is set to false.~~

- When the shipping or billing addresses are updated, add a flag to `localStorage` to mark whether the addresses are dirty (i.e. not synced with server yet).
- When the address is pushed, set this flag to false.
- **Push entire address object when updating customer, not just dirty keys.** - This is required because if you change the address in tab A, then only new changes in tab B are synced, problematic when country is changed and things like State and Postcode are synced and "invalid".
- Add some e2e tests for `extensionCartUpdate` - may not completely cover its functionality but covers this use-case.
- Add new arg to `extensionCartUpdate` called `overwriteDirtyCustomerData` - this can be used by extensions if they don't care about overwriting pending data. This defaults to `false`.
- Update documentation to reflect new args

Closes #49203

### How to test the changes in this Pull Request:

Using the [WooCommerce Testing Instructions Guide](https://github.com/woocommerce/woocommerce/wiki/Writing-high-quality-testing-instructions), include your detailed testing instructions:

#### Developer only testing

1. Install the [Code Snippets](https://wordpress.org/plugins/code-snippets/) plugin. Add a new snippet containing the following:

```php
use Automattic\WooCommerce\StoreApi\Schemas\ExtendSchema;
use Automattic\WooCommerce\StoreApi\StoreApi;

add_action(
	'woocommerce_init',
	function () {
		$extend = StoreApi::container()->get( ExtendSchema::class );
		if (
			is_callable(
				array(
					$extend,
					'register_update_callback',
				)
			)
		) {
			$extend->register_update_callback(
				array(
					'namespace' => 'woocommerce-blocks-test-extension-cart-update',
					'callback'  => function ( $data ) {
						if ( ! empty( $data['test-name-change'] ) ) {
							WC()->cart->get_customer()->set_shipping_first_name( 'Mr. Test' );
							WC()->cart->get_customer()->save();
						}
					},
				)
			);
		}
	}
);
```
2. In an incognito session, add an item to your cart and go to the Checkout block. **Be sure not to fill in any data unless the test instructions say so**.
[...]

PR #53896, opened 2024-12-19
Update `applyExtensionCartUpdate` to always return response

opr:

In https://github.com/woocommerce/woocommerce/pull/49311/files#diff-104ed0ccbbe914c26add1fcc3f567985816b8ab4653855eea51dc4247811322dL103 we removed the `return response;` line - this wouldn't have been a problem if the args had `overwriteDirtyCustomerData` set to `true` by default, but in the case that customer data wasn't dirty, response was not returned.

This PR ensures response is always returned.

1. Add the following code to be executed server-side:

```php
use Automattic\WooCommerce\StoreApi\Schemas\ExtendSchema;
use Automattic\WooCommerce\StoreApi\StoreApi;

add_action(
	'woocommerce_init',
	function () {
		$extend = StoreApi::container()->get( ExtendSchema::class );
		if (
			is_callable(
				array(
					$extend,
					'register_update_callback',
				)
			)
		) {
			$extend->register_update_callback(
				array(
					'namespace' => 'woocommerce-blocks-test-extension-cart-update',
					'callback'  => function ( $data ) {
						if ( ! empty( $data['test-name-change'] ) ) {
							WC()->cart->get_customer()->set_shipping_first_name( 'Mr. Test' );
							WC()->cart->get_customer()->save();
						}
					},
				)
			);
		}
	}
);
```

2. Go to the Checkout block and open the JS console. Execute the following code:

```js
wc.blocksCheckout.extensionCartUpdate( {
        namespace: 'woocommerce-blocks-test-extension-cart-update',
		overwriteDirtyCustomerData: true,
        cartPropsToReceive: [ 'extensions' ],
    } ).then( ( response ) => {
        console.log( { response } );
    } ).catch( ( error ) => {
        console.error( { error } );
    } );
```

3. Ensure the response is logged correctly, and that you don't see `undefined` as the response, the response should contain the cart data.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce-blocks/assets/js/data/cart/thunks.ts
+++ b/plugins/woocommerce-blocks/assets/js/data/cart/thunks.ts
@@ -125,6 +125,7 @@ export const applyExtensionCartUpdate =
 				return response;
 			}
 			dispatch.receiveCart( response );
+			return response;
 		} catch ( error ) {
 			dispatch.receiveError( isApiErrorResponse( error ) ? error : null );
 			return Promise.reject( error );
--- a/plugins/woocommerce-blocks/tests/e2e/tests/checkout/checkout-block-extensibility.shopper.block_theme.spec.ts
+++ b/plugins/woocommerce-blocks/tests/e2e/tests/checkout/checkout-block-extensibility.shopper.block_theme.spec.ts
@@ -41,6 +41,47 @@ test.describe( 'Shopper → Extensibility', () => {
 		await frontendUtils.goToCheckout();
 	} );
 	test.describe( 'extensionCartUpdate', () => {
+		test( 'Response is not undefined in any code path', async ( {
+			checkoutPageObject,
+		} ) => {
+			// With no additional args.
+			let response = await checkoutPageObject.page.evaluate(
+				"wc.blocksCheckout.extensionCartUpdate( { namespace: 'woocommerce-blocks-test-extension-cart-update' } ).then( ( response ) => response );"
+			);
+			let resolvedResponse = await Promise.resolve( response );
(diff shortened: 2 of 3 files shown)


nielslange (approved):
Thanks for working on this, @opr. I've successfully tested your change. Great that you added tests here. 👏

While reviewing this PR, I noticed that [our docs for `applyExtensionCartUpdate`](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce-blocks/docs/third-party-developers/extensibility/data-store/cart.md?plain=1#L365-L385) do not show the return value. In fact, the entire docs file only lists the parameters of the functions. Did we simply miss that when creating these docs? Do you think we should document the return values? What do you think?

opr:
> While reviewing this PR, I noticed that [our docs for applyExtensionCartUpdate](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce-blocks/docs/third-party-developers/extensibility/data-store/cart.md?plain=1&rgh-link-date=2024-12-23T10%3A58%3A38Z#L365-L385) do not show the return value. In fact, the entire docs file only lists the parameters of the functions. Did we simply miss that when creating these docs? Do you think we should document the return values? What do you think?

We don't really intend for the value to be used, so I'm hesitant about including it in the docs, I think it's there if folks really need it, but the preferred method is reading the data store for changes as per this dev advisory: https://developer.woocommerce.com/2024/12/19/developer-advisory-return-value-of-response-in-extensioncartupdate/

Issue #53765, opened 2024-12-16
Improve error handling for `createBlocksFromInnerBlocksTemplate`


gigitux:
### Issue Description

In the past weeks, @kmanijak and I have worked to prevent crashes when `createBlocksFromInnerBlocksTemplate` throws an error.

- https://github.com/woocommerce/woocommerce/issues/53640
- https://github.com/woocommerce/woocommerce/issues/53078

During internal discussions, we discovered that some WordPress sites might unregister `core/post-title` or other blocks needed by the Product Collection, leading to crashes in the Gutenberg editor. For this reason, it is necessary a further improvement.

One solution that comes to mind is to ensure that all blocks are registered via `getBlockType` and to omit the ones that aren't before invoking `createBlocksFromInnerBlocksTemplate`.

In pseudocode:

```
const templates = [...]
const templateWithRegisteredBlocks = templates.filter((block) => getBlockType(block) !== undefined)
createBlocksFromInnerBlocksTemplate(templateWithRegisteredBlocks) 
```

kmanijak:
There are two actions taken:

### 👌 Gutenberg

First, more future-proof action is in Gutenberg: https://github.com/WordPress/gutenberg/pull/68043. I received initial feedback and addressed it. Hopefully it gets merged and that will actually solve our problem. But to solve it we need to wait for WordPress L-1.

### ❌ WooCommerce

Second, temporary action in WooCommerce: https://github.com/woocommerce/woocommerce/pull/54065. This is my attempt with the local util that verifies the template first as per this issue. However, that DOESN'T solve the actual problem.

Only after implementing this I realised the editor crash comes from Gutenberg internals, more specifically from component used in the inserter: [InserterListItem](https://github.com/WordPress/gutenberg/blob/1ba8152f09f6ffc30699145705ada748cf5e917e/packages/block-editor/src/components/inserter-list-item/index.js#L47). That said, it means the possibility to crash Editor WAS ALWAYS there. But to make it happen two conditions needs to be met:
1. Block must rely on unavailable inner block
2. Block must be exposed in the inserter

#### Example 

Let's take `core/post-title` as an example of unavailable block. We know some plugins unregister it.

**Why editor doesn't crash considering Query Loop relies on it?**

[...]

Issue #53691, opened 2024-12-13
Opt-In Block Registration for Widget Area


gigitux:
This issue is based on the discussion in https://github.com/woocommerce/woocommerce/discussions/53659. Changing to an opt-in approach ensures that only blocks explicitly designated to work in the Widget Area are included.

**Next Steps:**

1. Triage blocks to identify those that work well and are appropriate for the widget area.
2. Update the registration logic to opt-in:
https://github.com/woocommerce/woocommerce/blob/48eb0030dce131947367d632b8dfecedb2d71def/plugins/woocommerce/src/Blocks/BlockTypesController.php#L473-L489
3. Add E2E tests to ensure blocks render correctly in the widget area.

PR #53898, opened 2024-12-19
[Experimental] Add mechanism to upgrade/downgrade between Add to Cart with Options block versions

### LINKED ISSUE #53208: [Enhancement]: Add mechanism to Upgrade/Downgrade the Blockified Add to Cart Experience
### Description
Add a way to allow users to upgrade or downgrade to the blockified `Add to Cart Form` experience. We can follow the same pattern we use to upgrade/downgrade from Products (Beta) to Product Collection block.

**Upgrade to**
Screenshot 2024-11-07 at 5 25 08 PM

**Revert to**
Screenshot 2024-11-07 at 5 27 30 PM

### Details

- **Functionality**
  - Provide a UI control for users to switch between the legacy and blockified Add to Cart Forms.
  - Handle any necessary cleanup or initialization when switching forms to prevent conflicts.

- **Implementation Considerations**
  - Ensure that the mechanism works across different product types and maintains compatibility with third-party extensions.

```[tasklist]
### Tasks
- [ ] Add the UI elements for upgrading/downgrading the Add to Cart Form.
- [ ] Test the mechanism thoroughly across various scenarios and product types.
```

Aljullu:
Closes #53208.

2. Go to `WooCommerce Admin Test Helper` -> `Features` and enable `blockified-add-to-cart` and `add-to-cart-with-options-stepper-layout`.
2. Create a post or page and add the Single Product block.
3. Select the existing _Add to Cart with Options_ block.
4. Verify in the sidebar there is a notice with a button to update. Click on it and verify it switches the block to the blockified version.
5. Do the same to switch back.
6. Try changing the display style to "Stepper" mode and verify it's preserved.

### DIFF
--- a/plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/edit.tsx
+++ b/plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/edit.tsx
@@ -5,11 +5,11 @@ import { useEffect } from '@wordpress/element';
 import {
 	BlockControls,
 	InnerBlocks,
+	InspectorControls,
 	useBlockProps,
 } from '@wordpress/block-editor';
 import { BlockEditProps } from '@wordpress/blocks';
 import { __ } from '@wordpress/i18n';
-import type { InnerBlockTemplate } from '@wordpress/blocks';
 
 /**
  * Internal dependencies
@@ -17,6 +17,8 @@ import type { InnerBlockTemplate } from '@wordpress/blocks';
 import { useIsDescendentOfSingleProductBlock } from '../../atomic/blocks/product-elements/shared/use-is-descendent-of-single-product-block';
 import { AddToCartOptionsSettings } from './settings';
 import ToolbarProductTypeGroup from './components/toolbar-type-product-selector-group';
+import { DowngradeNotice } from './components/downgrade-notice';
+import getInnerBlocksTemplate from './utils/get-inner-blocks-template';
 import useProductTypeSelector from './hooks/use-product-type-selector';
 
 export interface Attributes {
@@ -31,26 +33,6 @@ export type FeaturesProps = {
 };
 
 export type UpdateFeaturesType = ( key: FeaturesKeys, value: boolean ) => void;
-
-const INNER_BLOCKS_TEMPLATE: InnerBlockTemplate[] = [
-	[
-		'core/heading',
-		{
-			level: 2,
-			content: __( 'Add to Cart', 'woocommerce' ),
-		},
-	],
-	[ 'woocommerce/add-to-cart-with-options-variation-selector' ],
-	[ 'woocommerce/product-stock-indicator' ],
-	[ 'woocommerce/add-to-cart-with-options-quantity-selector' ],
-	[
-		'woocommerce/product-button',
-		{
-			textAlign: 'center',
-			fontSize: 'small',
-		},
-	],
-];
 
 const AddToCartOptionsEdit = ( props: BlockEditProps< Attributes > ) => {
 	const { setAttributes } = props;
@@ -80,12 +62,16 @@ const AddToCartOptionsEdit = ( props: BlockEditProps< Attributes > ) => {
 		unregisterListener,
 	] );
 
+	const innerBlocksTemplate = getInnerBlocksTemplate();
+
 	return (
 		<>
+			<InspectorControls>
+				<DowngradeNotice />
+			</InspectorControls>
 			<BlockControls>
 				<ToolbarProductTypeGroup />
 			</BlockControls>
-
 			<AddToCartOptionsSettings
 				features={ {
 					isBlockifiedAddToCart: true,
@@ -93,7 +79,7 @@ const AddToCartOptionsEdit = ( props: BlockEditProps< Attributes > ) => {
 			/>
 
 			<div { ...blockProps }>
-				<InnerBlocks template={ INNER_BLOCKS_TEMPLATE } />
+				<InnerBlocks template={ innerBlocksTemplate } />
 			</div>
 		</>
 	);
--- a/plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/edit.tsx
+++ b/plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/edit.tsx
@@ -9,7 +9,6 @@ import {
 	useBlockProps,
 } from '@wordpress/block-editor';
 import { BlockEditProps } from '@wordpress/blocks';
-import { __ } from '@wordpress/i18n';
 
 /**
  * Internal dependencies
--- a/plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/edit.tsx
+++ b/plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/edit.tsx
@@ -67,7 +67,7 @@ const AddToCartOptionsEdit = ( props: BlockEditProps< Attributes > ) => {
 	return (
 		<>
 			<InspectorControls>
-				<DowngradeNotice />
+				<DowngradeNotice blockClientId={ props?.clientId } />
 			</InspectorControls>
 			<BlockControls>
 				<ToolbarProductTypeGroup />
--- a/plugins/woocommerce-blocks/assets/js/blocks/product-elements/add-to-cart-form/components/upgrade-notice/index.tsx
+++ b/plugins/woocommerce-blocks/assets/js/blocks/product-elements/add-to-cart-form/components/upgrade-notice/index.tsx
@@ -1,0 +1,88 @@
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { Notice, Button } from '@wordpress/components';
+import { createInterpolateElement } from '@wordpress/element';
+import { recordEvent } from '@woocommerce/tracks';
+import { dispatch, select } from '@wordpress/data';
+import { findBlock } from '@woocommerce/utils';
+import {
+	createBlock,
+	// @ts-expect-error Type definitions for this function are missing in Gutenberg
+	createBlocksFromInnerBlocksTemplate,
+} from '@wordpress/blocks';
+
+/**
+ * Internal dependencies
+ */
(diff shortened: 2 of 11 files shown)


plugins/woocommerce-blocks/assets/js/blocks/product-elements/add-to-cart-form/components/upgrade-notice/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+		<Notice isDismissible={ false }>
+			<>{ notice }</>
+			<br />
+			<br />
+			<Button variant="link" onClick={ handleClick }>
+				{ buttonLabel }
+			</Button>
+		</Notice>
+	);
+};
```
kmanijak, line 101: Can't help but notice the similarity between:
- plugins/woocommerce-blocks/assets/js/blocks/product-elements/add-to-cart-form/components/upgrade-notice/index.tsx
- plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/components/downgrade-notice/index.tsx
- https://github.com/woocommerce/woocommerce/blob/c7648abe5b5e1387c9e8ead887f1913b75e5443a/plugins/woocommerce-blocks/assets/js/blocks/product-query/inspector-controls/upgrade-notice.tsx#L9
- https://github.com/woocommerce/woocommerce/blob/c7648abe5b5e1387c9e8ead887f1913b75e5443a/plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/summary/upgrade.tsx#L38
- https://github.com/woocommerce/woocommerce/blob/c7648abe5b5e1387c9e8ead887f1913b75e5443a/plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/upgrade-notice.tsx#L33

For some of them I am of course guilty of repeating the same code 😅 Not the scope of this PR but we should consider extracting the notice as a separate block. Not expecting it in this PR but just mentioning for transparency.

plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/edit.tsx
```diff
@@ -79,20 +61,24 @@ const AddToCartOptionsEdit = ( props: BlockEditProps< Attributes > ) => {
 		unregisterListener,
 	] );
 
+	const innerBlocksTemplate = getInnerBlocksTemplate();
+
 	return (
 		<>
+			<InspectorControls>
+				<DowngradeNotice blockClientId={ props?.clientId } />
+			</InspectorControls>
```
kmanijak, line 70: In the current shape, if you add Add to Cart with Options (Experimental) block, you'd see the notice even though you didn't upgrade. In scope of Product Collection I used `convertedFromProducts` attribute to mark blocks upgraded from Products (Beta) and showed the notice only for them.

https://github.com/woocommerce/woocommerce/blob/c7648abe5b5e1387c9e8ead887f1913b75e5443a/plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/index.tsx#L253

add to cart

That's probably increases the scope of the whole feature but also we added couple of criteria to stop showing "Switch back" notice:

> // Upgrade Notice should be displayed only if:
> // - block is converted from Products
> // - user haven't acknowledged seeing the notice
> // - less than X hours since the notice was first displayed
> // - notice was displayed less than X times

https://github.com/woocommerce/woocommerce/blob/c7648abe5b5e1387c9e8ead887f1913b75e5443a/plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/index.tsx#L244C6-L262C3.

But considering the block is experimental I'll let you decide if that's blocking or not.

kmanijak (commented):
PR is testing well! 🚀 

Based on experience in Product Collection block I shared a UX concern. But considering the block is experimental, if you don't plan to work on it at the moment, I'm happy to approve but would love to hear your opinion first!

plugins/woocommerce-blocks/assets/js/blocks/product-elements/add-to-cart-form/components/upgrade-notice/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+		'Upgrade to the blockified Add to Cart with Options block',
+		'woocommerce'
+	);
+
+	const handleClick = () => {
+		upgradeToBlockifiedAddtoCartWithOptions( blockClientId );
+		recordEvent( 'blocks_add_to_cart_with_options_migration', {
+			transform_to: 'blockified',
+		} );
+	};
```
kmanijak: After recent troubles we had with `createBlocksFromInnerBlocksTemplate` in various areas (https://github.com/woocommerce/woocommerce/issues/53765), I wanted to check how it behaves in Widgets and Patterns.

It seems the upgrade doesn't happen in widgets in classic theme and I think it's ok, because Single Product block doesn't work there anyway.

However, the event is recorded anyway. Considering the `upgradeToBlockifiedAddtoCartWithOptions` may not pass due to:

```
    if ( ! foundBlock ) {
		return;
	}
```

I think the `upgradeToBlockifiedAddtoCartWithOptions` should return boolean and the track should be recorded or not depends on the outcome:

```suggestion
	const handleClick = () => {
		const upgraded = upgradeToBlockifiedAddtoCartWithOptions( blockClientId );
		if ( upgraded ) {
			recordEvent( 'blocks_add_to_cart_with_options_migration', {
			    transform_to: 'blockified',
		    } );
		}
	};
```

It's minor and rather unlikely scenario but feels cleaner. The same would apply to downgrade.

octaedro (approved):
Good job @Aljullu! This is testing well on my end and the code LGTM 🚀 

For a follow-up PR (or a cooldown task), it might be a good idea to consolidate all the `upgrade` and `downgrade` notices into a single component for consistent use throughout the `WooCommerce-Blocks` project.

**UPDATE**
I just saw Karol's comments and reproduced what he mentioned. I think that we should address his comments before merging this PR.

plugins/woocommerce-blocks/assets/js/blocks/product-elements/add-to-cart-form/components/upgrade-notice/index.tsx
Aljullu: Oh, good catch, I wasn't aware we had that many! Fully agree that we should unify them. :100:  I opened an issue here: https://github.com/woocommerce/woocommerce/issues/54076.

plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/edit.tsx
Aljullu: Good points! I'm a bit on the fence on this. Personally, I would prefer to leave it as "always visible" for now, at least until we have a release plan and we have investigated extension support.

If we release this block in an experimental state, I think it will be better to have the button always visible to switch to the stable block. What do you think? We currently do something similar with the "Revert to Classic Template" button in templates, it's always visible no matter if the user actively switched to the blockified template or not.

plugins/woocommerce-blocks/assets/js/blocks/product-elements/add-to-cart-form/components/upgrade-notice/index.tsx
Aljullu: Good catch. We should probably also hide the block in the widgets screen. :grimacing: I will take a look at it.

For now, I implemented your suggestion in 50ce7c1c8834df7c32af375464da9f2812bfe5e2.

Aljullu:
Thanks for the reviews, @kmanijak @octaedro. I fixed/answered the comments, would you mind taking another look? :bow:

plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/edit.tsx
kmanijak: Yeah, I don't mind and it sounds ok! But wanted to mention this aspect anyway knowing we faced that earlier 🙌

plugins/woocommerce-blocks/assets/js/blocks/product-elements/add-to-cart-form/components/upgrade-notice/index.tsx
kmanijak: > Good catch. We should probably also hide the block in the widgets screen. 😬 I will take a look at it.

FWIW it can be hidden in here (wouldn't be surprised if you knew this already 😅 ):

https://github.com/woocommerce/woocommerce/blob/62b81a321a219f8e241e7f12ca9394fbd3aa9e97/plugins/woocommerce/src/Blocks/BlockTypesController.php#L521C6-L536C4

kmanijak (approved):
All changes look good to me! It tests well and as mentioned in comments, for now the notice will be displayed in every block and indefinitely no matter if it was transformed from previous block or not 👍 

Great job!

PS. Thanks for creating the issue to unify notices! 🚀

plugins/woocommerce-blocks/assets/js/blocks/product-elements/add-to-cart-form/components/upgrade-notice/index.tsx
Aljullu: Thanks! I opened a PR here: https://github.com/woocommerce/woocommerce/pull/54087. Apparently switching from `registerBlockType()` to `registerProductBlockType()` was enough. Not sure if we should add the block to the PHP list as well, but considering we have https://github.com/woocommerce/woocommerce/issues/53691, I don't think it's needed. :thinking:

kmanijak: > considering we have https://github.com/woocommerce/woocommerce/issues/53691, I don't think it's needed. 🤔

Actually, I even picked it up today and PR is nearly ready (https://github.com/woocommerce/woocommerce/pull/54086) so definitely not worth the effort!

Aljullu: Awesome! :heart_eyes: Feel free to ping me if you need a review!

Issue #53860, opened 2024-12-19
[Settings] Load and run external scripts


psealock:
The Settings Proof of Concept accomplished this in 3 parts

1. Make sure scripts we'd see on current Settings register and enqueue. [Source](https://github.com/woocommerce/woocommerce/pull/48340/files#diff-93e330e41dcee0a7211c92b86b52ba9c4c1f2c0424457958abdfc8b2fb45e659)
2. Grab registered scripts and pass the `src` along with `settingsData` [Source](https://github.com/woocommerce/woocommerce/pull/48340/files#diff-2eda4b4f15229298aab5b05def76b8abadde4d0d2931fd9929de97ac5e2b53f3)
3. Apply the scripts on tab/section change
https://github.com/woocommerce/woocommerce/blob/09cdb35a1ebbc422f87c5dea1795b0b7824d254e/plugins/woocommerce-admin/client/settings/index.js#L32

PR #53909, opened 2024-12-20
[Settings] Register and render internal slotFill

psealock:

Register and render slotFills from our own codebase. This task is normally [handled in wc-admin](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/admin/client/embedded-body-layout/render-embedded-layout.tsx#L112) but we avoid loading the wc-admin app.

To use `useActiveRoute` it must be called from a component within `RouterProvider`, so that needed to be exported from the package and used at the root level. This allows re-rendering of slotFills when the tab or section change.

SlotFills generated by extensions won't be rendered until https://github.com/woocommerce/woocommerce/issues/53860 is handled.

3. WooCommerce > Settings > Site visibility. The launch your store site visibility settings should appear but they will be non functional.
4. WooCommerce > Settings > Tax. See the error warning if applicable.
image

5. WooCommerce > Settings > Payments. On a fresh site, this slotFill won't render anything but confirm no error messages appear.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Register and render internal slotFills in the new Settings.

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/admin/client/settings/index.js
```diff
@@ -1,16 +1,46 @@
...
 
+const Settings = () => {
+	const { activePage, activeSection } = useActiveRoute();
+
+	// Render the settings slots every time the page or section changes.
+	useEffect( () => {
+		possiblyRenderSettingsSlots();
+	}, [ activePage, activeSection ] );
+
+	// Register the settings slots only once.
```
rjchow: we shouldn't rely on useEffect with empty dependency arrays to [run only exactly once on mount](https://medium.com/@naveen-rk/why-useeffect-is-running-twice-on-the-initial-render-react-strictmode-c99e987f9956)

In the gutenberg repo it looks like they just call the registerPlugin in the module scope so it's loaded when the JS loads. If that's not suitable here I guess we could write some custom hook that tracks execution with a useRef or similar? I've seen this pattern of useEffect with empty dep array being used frequently in this repo and it'll be a bit of a cleanup issue when we move to react 19

rjchow (changes requested):
Left a comment on useEffect usage but otherwise lgtm!

plugins/woocommerce/client/admin/client/settings/index.js
ilyasfoo: Good catch. Alternatively, does de-registering fills on unmount make sense?

Edit: I think another way is to simply declare a local variable in module scope like `hasRegisteredFills` and use that to ensure it's only called once per page load

rjchow: That’ll work too!

psealock: It turns out these `registerPlugin` can be called at any time after Gutenberg is available. In https://github.com/woocommerce/woocommerce/pull/53909/commits/2c537135456b516fb18ae56c908cbdbce5243997 I removed those function calls from `useEffect` inside the component and just call them in the file.

rjchow (approved):
lgtm, thank you!

PR #53910, opened 2024-12-20
Add coming soon image gallery pattern

### LINKED ISSUE #53302: Coming soon template 5
### Issue Description

Build coming soon template variant. The scope of this task is specifically for building a simple template, functionality for adding product images can be done as a follow-up.

P2: pfwcF5-13s-p2

Figma: 0BWMerqRR1f468EwC7g5tg-fi-3652_27731

Pre-requisite: #52393 and #52397

Image

### Acceptance Criteria

1. Build template using blocks
1. Ensure images and copy are editable

adrianduffell:

Closes https://github.com/woocommerce/woocommerce/issues/53302.

This adds an image gallery pattern to the coming soon template, allowing users to showcase a preview of what they are selling with a variety of small images.

It is based off the T24 homepage pattern "[Portfolio home template with post featured images](https://github.com/WordPress/wordpress-develop/blob/b8ecfbf9542198b1d88230221faeca6b71fa0ac7/src/wp-content/themes/twentytwentyfour/patterns/template-home-portfolio.php#L4)", which displays blog posts featured images. I have replaced those with image blocks to allow the user to easily change the sample images with their own.

A follow-up PR will explore using existing images in the product catalogue to make it even simpler to use.

Screen Shot 2024-12-21 at 12 11 51 am

1. Create a fresh site with  coming-soon-newsletter-template feature flag enabled.
1. Enable coming soon mode in site visibility settings.
1. Go to Appearance > Editor > Templates > Coming soon
1. Choose Coming Soon Image Gallery design on the right sidebar (Template > Design > Coming Soon Image Gallery)
1. Confirm the layout matches the design in Figma.
1. Confirm that you can customize the styles, replace images, change the content, etc.
1. Select Coming Soon Image Gallery  Image template again and save the template.
1. Go to the site front end incognito tab to make sure the template is loaded correctly. The page should be responsive.
1. Go to Appearance > Editor > Templates and confirm the coming soon template preview shows the selected template.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce-blocks/assets/js/blocks/coming-soon/entire-site.scss
```diff
@@ -202,3 +202,28 @@ body:has(.woocommerce-coming-soon-banner) {
...
+			line-height: normal;
+			text-align: center;
+			text-decoration: none;
+			padding: 12px 16px;
+		}
+	}
+
+	.wp-block-image {
+		img {
+			border-radius: 14px;
```
ilyasfoo: Nit: I think this is not the most optimal since we're not using site editor's built-in border-radius settings. However, it's still possible to override it in via the `Border & Shadow` under the block styles tab on individual images.

Still, I wonder if it's better to instead use the site editor's styles' border radius on individual images to make the template more "Gutenberg native"?

plugins/woocommerce/assets/images/pattern-placeholders/gallery-12.jpg
ilyasfoo, line 1: 2px shaved from this file due to white lines appearing when converting from Figma which contains transparent pixels

plugins/woocommerce/assets/images/pattern-placeholders/gallery-2.jpg
ilyasfoo, line 1: 2px shaved from this file due to white lines appearing when converting from Figma which contains transparent pixels

ilyasfoo (commented):
@adrianduffell  Looking good! A few minor comments.

Also, I've pushed f5048e7 which does the following:

1. Remove while lines, which is observable when changing background color ([example](https://github.com/user-attachments/assets/bddd0590-4eda-4512-9250-60fc59061fad)) 
1. Optimized JPG sizes using https://tinyjpg.com/.

Edit: I also noticed that changelog is missing

plugins/woocommerce-blocks/assets/js/blocks/coming-soon/entire-site.scss
adrianduffell: @ilyasfoo Good idea! I fixed this in 26d9b59. The sample images have a border embedded in the image file too. It might be worth a follow-up PR to remove this as well so users can adjust the radius on these?

rjchow (commented):
Looks amazing! I have this feedback:

image
I think the images should be centered when in column layout

image

The font color doesn't follow the settings - and I was wondering if it should also be possible to change the background color?

plugins/woocommerce-blocks/assets/js/blocks/coming-soon/entire-site.scss
ilyasfoo: > The sample images have a border embedded in the image file too. It might be worth a follow-up PR to remove this as well so users can adjust the radius on these?

@adrianduffell Right, sounds good for a follow-up! Might need to convert the images to png to make it transparent 🤔

ilyasfoo:
> The font color doesn't follow the settings - and I was wondering if it should also be possible to change the background color?

@rjchow Right, I think this is a problem with all coming soon patterns, maybe because of a CSS rule overriding them. I think it's a good follow-up to fix all of them

ilyasfoo:
I'll make the changes

ilyasfoo:
> I think the images should be centered when in column layout

@rjchow I've updated the alignment to center in 1853be6, along with some minor cleanliness improvements.

image


I've also added ID to the pattern in bb43742 for tracks.

Please have another look!

plugins/woocommerce-blocks/assets/js/blocks/coming-soon/entire-site.scss
```diff
@@ -202,3 +202,22 @@ body:has(.woocommerce-coming-soon-banner) {
 		aspect-ratio: 1 / 1;
 	}
 }
+
+// Product image gallery template styles
+.woocommerce-coming-soon-image-gallery {
+	.wp-block-loginout {
+		display: flex;
+
+		a {
```
chihsuan, line 211: We should remove these custom CSS styles and use the block settings to control the styles instead so that users can adjust the styles in block editor.

https://github.com/user-attachments/assets/c7176ad6-1c5b-4965-8d57-f3591ecb749f

chihsuan (commented):
Nice work! Just noticed that:

1. Can't change the loginout button font size due to the CSS override (Left a comment in code).

2. The heading padding/margin doesn't match the design. It should be 100px but it's 85px.

Screenshot 2024-12-25 at 13 08 54

Expected:
Screenshot 2024-12-25 at 13 08 59

3. The spacing between images doesn't match the design. It seems that different groups of images need to have different spacing between them.

Screenshot 2024-12-25 at 13 08 03

Expected:
Screenshot 2024-12-25 at 13 07 52

ilyasfoo:
Great feedback, @chihsuan!

> Can't change the loginout button font size due to the CSS override (Left a comment in code).

Good catch! However, this seems to be an issue that impacts [at least 2 patterns](https://github.com/woocommerce/woocommerce/blob/2d639722a5a995d087d55eb1b5f5530e04fa7eb6/plugins/woocommerce-blocks/assets/js/blocks/coming-soon/entire-site.scss#L74-L88), I'm thinking it's better to be addressed in a follow-up, where we could audit all patterns' login button together. Wdyt?

> The heading padding/margin doesn't match the design. It should be 100px but it's 85px.

Good catch! Fixed it in 61d06c5.

> The spacing between images doesn't match the design. It seems that different groups of images need to have different spacing between them.

Thanks, I tried this but it seems it's not ideal UI for mobile where the gap doesn't make sense. I'm not able to find a workaround for mobile-specific UI, I'm thinking it's better to keep the gap small so it can work in both for now.

image

chihsuan (approved):
Thanks, LGTM! 👍 

> I'm thinking it's better to be addressed in a follow-up, where we could audit all patterns' login button together. Wdyt?

Sounds good!

>  I'm not able to find a workaround for mobile-specific UI, I'm thinking it's better to keep the gap small so it can work in both for now.

Yeah, I think we need Core to support this or need css override which is not ideal. I'm okay with this for now.

plugins/woocommerce-blocks/assets/js/blocks/coming-soon/entire-site.scss
ilyasfoo: On a second thought, I looked at the images when it's converted to png, the size increase is quite substantial. Given that this may only benefit small percentage of users who'd likely to reduce or remove the border radius, I think the better solution might be to provide image without any border radius, so it will not have that odd white borders when users remove border radius.

rjchow: (the source images on figma have rectangular corners)

ilyasfoo: Created a follow-up issue: https://github.com/woocommerce/woocommerce/issues/54036

ilyasfoo:
I've created the following follow-up issues:

- https://github.com/woocommerce/woocommerce/issues/54036
- https://github.com/woocommerce/woocommerce/issues/54037
- https://github.com/woocommerce/woocommerce/issues/54039

FYI @adrianduffell @rjchow  @chihsuan

rjchow (approved):
thanks for working on this Ilyas! lgtm now

PR #53913, opened 2024-12-20
[Settings] Handle outputed script tags

### LINKED ISSUE #52975: [Settings] Custom views - Handle outputed script tags
Sometimes extensions will place a `<script />` tag in the custom output of a settings field, such as in https://github.com/woocommerce/woocommerce/issues/52317. This script tag may include javascript or an external script. Make sure this is passed on in SettingsData to the client. Note that these scripts should be their own Setting element with type `script` (or something).

The proof of concept used `WP_HTML_Tag_Processor` for this.

https://github.com/woocommerce/woocommerce/blob/09cdb35a1ebbc422f87c5dea1795b0b7824d254e/plugins/woocommerce/includes/admin/settings/class-wc-settings-page.php#L182-L193

chihsuan:

Closes https://github.com/woocommerce/woocommerce/issues/52975

This PR extracts `<script>` tag contents from custom settings fields using a DOM parser and injects them into the DOM.

Additionally, this PR fixes the type definitions:

1. Update `type` of `SettingsField` to be explicitly defined constants instead of `string`.
2. Add `CustomSettingsField` and fix the `GroupSettingsField` type.
3. Fix SettingsSection.settings type. (`SettingField` is undefined.)

I use the same way as POC https://github.com/woocommerce/woocommerce/pull/48340 to inject the script into the DOM. Since it's usually dangerous to do this, I think we should look at it more deeply and ensure that injecting the script into the DOM is safe. Also, we need to clean up the script when the component unmounts. However, if scripts create global variables or event listeners, it might cause unexpected behavior and cause memory leaks. But this is beyond the scope of this PR. 

3. Use the [Settings Tester](https://github.com/woocommerce/woocommerce-settings-migration-tester) to test Modern screens. Clone it locally and `npm install && npm run plugin-zip` / [settings-tester.zip](https://github.com/user-attachments/files/18208344/settings-tester.zip)
3. Upload and activate the Tester
4. Go to `/wp-admin/admin.php?page=wc-settings&tab=settings-tester&section=javascript`
5. Toggle `Echoing HTML` checkbox and confirm the yellow background is shown when the checkbox is checked.

https://github.com/user-attachments/assets/b9d31e78-c1b4-447b-8077-138cf531c205

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

This is behind the settings feature flag.

</details>


packages/js/settings-editor/src/legacy/content.tsx
```diff
@@ -27,20 +27,20 @@ export const LegacyContent = ( {
 	return (
 		<form>
 			<div className="woocommerce-settings-content">
-				{ section.settings.map( ( data ) => {
-					const key =
-						data.id +
-						'-' +
-						( data.title ?? '' ).replace( /\s+/g, '-' ) +
-						'-' +
-						data.type;
```
chihsuan, line 36: `CustomSettingsField` doesn't have `id` and `title` properties so I use type + `key` as the unique identifier.

packages/js/settings-editor/typings/global.d.ts
```diff
@@ -38,13 +48,14 @@ declare global {
...
 		| BaseSettingsField
+		| CustomSettingsField
 		| GroupSettingsField
 		| CheckboxGroupSettingsField
 		| CheckboxSettingsField;
 
 	interface SettingsSection {
 		label: string;
-		settings: SettingField[];
+		settings: SettingsField[];
```
chihsuan, line 82: `SettingField` doesn't exist.

packages/js/settings-editor/typings/global.d.ts
```diff
@@ -1,7 +1,7 @@
 declare global {
 	interface BaseSettingsField {
 		title?: string;
-		type: string;
+		type: 'text' | 'password' | 'datetime' | 'datetime-local' | 'date' | 'month' | 'time' | 'week' | 'number' | 'email' | 'url' | 'tel'  | 'select' | 'radio' | 'textarea' | 'sectionend' | 'slotfill_placeholder';
```
rjchow: ❤️

packages/js/settings-editor/src/components/custom-view/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+	} );
+
+	return {
+		cleanHTML: sanitize( doc.documentElement.outerHTML ),
+		scripts,
+	};
+};
+
+const SettingsScript = ( { content }: { content: string } ) => {
+	useEffect( () => {
```
rjchow, line 44: I believe this will execute twice if strictmode is turned on

packages/js/settings-editor/src/components/custom-view/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+		return content;
+	} );
+
+	return {
+		cleanHTML: sanitize( doc.documentElement.outerHTML ),
+		scripts,
+	};
+};
+
+const SettingsScript = ( { content }: { content: string } ) => {
```
rjchow, line 43: Not sure if it's worth putting in right now, but I think it might be helpful to consider adding some ways to manipulate the script tag from outside this component, for example to add ways to defer the rendering or to remove the script node (e.g we have a long settings screen and want to lazy load or something. or perhaps this script has a dependency on something else being rendered to the dom first, so we want to be able to append the script in a controlled way)

rjchow, line 43: I'm not sure we need the .range or useLayoutEffect for our purposes but leaving this here for reference since I came across it while doing research for this PR review https://macarthur.me/posts/script-tags-in-react/

rjchow (commented):
Leaving some comments first, having a closer look tomorrow

packages/js/settings-editor/src/components/custom-view/index.tsx
chihsuan: I think most of the scripts need to be run after the elements are mounted. 

To ensure this is fully compatible with classic settings, they should be run in the order they appear in the document since these scripts are inline scripts in the HTML. However, I think that might be tricky and need more research.

chihsuan: Good point! I think we can explore if plugins do anything like this. From what I can see, the scripts are used to manipulate the DOM in its own custom output. (cc @psealock if you know more about this)

packages/js/settings-editor/src/components/custom-view/index.tsx
chihsuan: Yes 😅.  I think we need a follow-up issue to improve the script injection code. I want to keep this PR focused on the original issue.

rjchow (approved):
lgtm, thanks!

psealock (approved):
Nice work @chihsuan, this is working as expected and code looks good. I created an issue to track some important issues brought up by @rjchow 

https://github.com/woocommerce/woocommerce/issues/54180

PR #53941, opened 2024-12-20
Introducing payment method constants

wjrosa:

This PR introduces constants for all payment method identifiers. The idea is to make it easier to reference them, avoid typos, make the code stricter, make it easier to find status usages, centralize them, improve documentation, and many other reasons that should help developers create code related to them.

- Checkout and build this branch on your test environment (`dev/introducing-payment-method-constants`)
    - Store setup
    - Product creation
    - Common checkout
    - Shortcode checkout
    - My account page navigation
    - Order listing/details

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 
Not required since it is just a constant introduction.

</details>


plugins/woocommerce/includes/admin/class-wc-admin-setup-wizard.php
```diff
@@ -1528,7 +1529,7 @@ public function get_wizard_in_cart_payment_gateways() {
 		$can_paypal  = $this->is_paypal_supported_currency( $currency );
 
 		if ( ! current_user_can( 'install_plugins' ) ) {
-			return $can_paypal ? array( 'paypal' => $gateways['paypal'] ) : array();
+			return $can_paypal ? array( PaymentMethods::PAYPAL => $gateways['paypal'] ) : array();
```
barryhughes:
```suggestion
			return $can_paypal ? array( PaymentMethods::PAYPAL => $gateways[PaymentMethods::PAYPAL] ) : array();
```

I might be mistaken, but shouldn't we use the constant twice in this case?

plugins/woocommerce/tests/legacy/unit-tests/payment-gateways/payment-gateways.php
```diff
@@ -42,7 +44,7 @@ public function test_wc_set_current_gateway_from_session() {
 
 		$gateways = WC()->payment_gateways()->get_available_payment_gateways();
 		$gateways['bacs']->chosen = false;
-		WC()->session->set( 'chosen_payment_method', 'bacs' );
+		WC()->session->set( 'chosen_payment_method', PaymentMethods::BACS );
```
barryhughes: Just curious, why don't we use the new constant on the previous line? That is, like this:

```php
$gateways[PaymentMethods::BACS]->chosen = false;
```

(I'm only flagging this one, but I think there are a few areas like this.)

barryhughes (commented):
Thanks!

Left a couple of minor notes. Also wanted to double check if anything flagged using the following one-liner needs to be updated:

```sh
grep -Ern "'(bacs|cheque|cod|paypal|woocommerce_payments)'" ./ | grep .php
```

For instance, in [class-wc-settings-payment-gateways.php](https://github.com/woocommerce/woocommerce/blob/b4020a8432f669e0c67b32c3beedcb3238654d0d/plugins/woocommerce/includes/admin/settings/class-wc-settings-payment-gateways.php) should we update the following, and use the new constants for the array keys?

```php
$wcpay_setup        = isset( $payment_gateways['woocommerce_payments'] ) && ! $payment_gateways['woocommerce_payments']->needs_setup();
```

wjrosa:
> Thanks!
> 
> Left a couple of minor notes. Also wanted to double check if anything flagged using the following one-liner needs to be updated:
> 
> ```shell
> grep -Ern "'(bacs|cheque|cod|paypal|woocommerce_payments)'" ./ | grep .php
> ```
> 
> For instance, in [class-wc-settings-payment-gateways.php](https://github.com/woocommerce/woocommerce/blob/b4020a8432f669e0c67b32c3beedcb3238654d0d/plugins/woocommerce/includes/admin/settings/class-wc-settings-payment-gateways.php) should we update the following, and use the new constants for the array keys?
> 
> ```
> $wcpay_setup        = isset( $payment_gateways['woocommerce_payments'] ) && ! $payment_gateways['woocommerce_payments']->needs_setup();
> ```

Thanks for the review, @barryhughes ! I replaced all occurrences returned by your command and accepted your suggestions as well 👍

plugins/woocommerce/tests/php/src/Blocks/BlockTypes/OrderConfirmation/Totals.php
```diff
@@ -136,7 +136,7 @@ public function test_order_notes_cleaned() {
 				),
 				'create_account'   => true,
 				'customer_note'    => '<a href="http://attackerpage.com/csrf.html">This text should not save inside an anchor.</a><script>alert("alert")</script>',
-				'payment_method'   => 'bacs',
+				'payment_method'   => PaymentMethods::BACS,
```
barryhughes: Many of our checks are failing, and I think one of the causes may be that in some cases we haven't imported the enum (ie, we need a `use` statement at the top of this file) ... same applies to a few other files we touched.

Example of a failing test:

```
2) Automattic\WooCommerce\Tests\Blocks\BlockTypes\OrderConfirmation\Totals::test_order_notes_cleaned
Error: Class 'Automattic\WooCommerce\Tests\Blocks\BlockTypes\OrderConfirmation\PaymentMethods' not found
```

barryhughes (changes requested):
Thanks for this!

1. Some test failures we need to address (looks like we haven't imported the enum in a few cases).
2. Also, and perhaps in some ways it's a little late to be asking this (though, better late than never ... and I'd anyway be interested in fellow reviewer @kalessil's thoughts) but in this particular case I actually wonder if introducing a new class, as we're doing here, is preferable to adding constants to the gateway classes themselves. 

To clarify that second point, I guess my feeling is that something like `WC_Gateway_Cheque::ID` (or `::SLUG`) might feel a little more natural than `PaymentMethods::CHEQUE` (why *wouldn't* the class know its own slug/ID/moniker)?. This would also remove the need to introduce yet another class.

Again, I wanted to pause and ask the question, since we'll be working with this new enum class for a long time, but I'm also not opposed to it in its current form.

wjrosa:
> Thanks for this!
> 
> 1. Some test failures we need to address (looks like we haven't imported the enum in a few cases).
> 2. Also, and perhaps in some ways it's a little late to be asking this (though, better late than never ... and I'd anyway be interested in fellow reviewer @kalessil's thoughts) but in this particular case I actually wonder if introducing a new class, as we're doing here, is preferable to adding constants to the gateway classes themselves.
> 
> To clarify that second point, I guess my feeling is that something like `WC_Gateway_Cheque::ID` (or `::SLUG`) might feel a little more natural than `PaymentMethods::CHEQUE` (why _wouldn't_ the class know its own slug/ID/moniker)?. This would also remove the need to introduce yet another class.
> 
> Again, I wanted to pause and ask the question, since we'll be working with this new enum class for a long time, but I'm also not opposed to it in its current form.

I actually think it makes total sense! And it matches what was done with WooPayments and the Stripe extension already. I have updated the main implementation in https://github.com/woocommerce/woocommerce/pull/53941/commits/69e77454c4e54c976a397556e6e33ee93a018a91. I had to remove `woocommerce_payments` value as we don't have the gateway class in this repository.

plugins/woocommerce-beta-tester/api/tools/class-fake-wcpayments.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
barryhughes: I think if we nix this line then we can remove this file from the changeset.

```suggestion
```

plugins/woocommerce-beta-tester/changelog/53941-dev-introducing-payment-method-constants
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: dev
+
+Introduces and implements payment method identifier constants in the WooCommerce plugin codebase.
```
barryhughes: Following my earlier comment, if we are not making substantive changes to the beta tester then we don't need a separate changelog entry for it.

plugins/woocommerce/includes/abstracts/abstract-wc-payment-gateway.php
```diff
@@ -206,7 +206,7 @@ public function get_method_description() {
 	 */
 	public function admin_options() {
 		$is_reactify_enabled      = Features::is_enabled( 'reactify-classic-payments-settings' );
-		$offline_payment_gateways = array( 'bacs', 'cheque', 'cod' );
+		$offline_payment_gateways = array( WC_Gateway_BACS::ID, \WC_Gateway_Cheque::ID, \WC_Gateway_COD::ID );
```
barryhughes: This isn't wrong, of course, and it doesn't break any linting rules, but I'd personally think this is cleaner:

```suggestion
		$offline_payment_gateways = array( WC_Gateway_BACS::ID, WC_Gateway_Cheque::ID, WC_Gateway_COD::ID );
```

Just because we are already within the root namespace, so using `\` to qualify the classes doesn't add much (and it's a little odd to do it for just two of the referenced classes).

plugins/woocommerce/includes/admin/class-wc-admin-setup-wizard.php
```diff
@@ -1401,13 +1401,13 @@ protected function get_wizard_available_in_cart_payment_gateways() {
...
-				'class'       => 'checked stripe-logo',
-				'repo-slug'   => 'woocommerce-gateway-stripe',
-				'settings'    => array(
+			'stripe'               => array(
+				'name'             => __( 'WooCommerce Stripe Gateway', 'woocommerce' ),
+				'image'            => WC()->plugin_url() . '/assets/images/stripe.png',
+				'description'      => $stripe_description,
+				'class'            => 'checked stripe-logo',
+				'repo-slug'        => 'woocommerce-gateway-stripe',
+				'settings'         => array(
```
barryhughes: Here but also in one or more of the following blocks, can we verify the spacing before the `=>` is correct? PHPCS is complaining and I think the problem is alignment should be "per level". That is, within the inner array we don't need alignment with the arrows from the outer array.

```suggestion
				'name'        => __( 'WooCommerce Stripe Gateway', 'woocommerce' ),
				'image'       => WC()->plugin_url() . '/assets/images/stripe.png',
				'description' => $stripe_description,
				'class'       => 'checked stripe-logo',
				'repo-slug'   => 'woocommerce-gateway-stripe',
				'settings'    => array(
```

Etc, repeating as needed.

plugins/woocommerce/includes/admin/class-wc-admin-setup-wizard.php
```diff
@@ -1589,19 +1589,19 @@ public function get_wizard_in_cart_payment_gateways() {
 	public function get_wizard_manual_payment_gateways() {
 		_deprecated_function( __CLASS__ . '::' . __FUNCTION__, '4.6.0', 'Onboarding is maintained in WooCommerce Admin.' );
 		$gateways = array(
-			'cheque' => array(
+			\WC_Gateway_Cheque::ID => array(
```
barryhughes: Similar to a previous comment, throughout the changeset (there's another example on line 1604 where we reference `\WC_Gateway_COD`) we should be consistent in terms of where we do and don't qualify our class names.

```suggestion
			WC_Gateway_Cheque::ID => array(
```

plugins/woocommerce/tests/legacy/unit-tests/rest-api/Helpers/OrderHelper.php
```diff
@@ -332,7 +333,7 @@ public static function create_complex_data_store_order( $data_store = null ) {
 		$order->set_billing_phone( '555-32123' );
 
 		$payment_gateways = WC()->payment_gateways->payment_gateways();
-		$order->set_payment_method( $payment_gateways['bacs'] );
+		$order->set_payment_method( $payment_gateways[ WC_Gateway_BACS::class ] );
```
barryhughes:
```suggestion
		$order->set_payment_method( $payment_gateways[ WC_Gateway_BACS::ID ] );
```

I think we intend to use the ID and not the class name here, right? This should fix a set of failing tests.

barryhughes (changes requested):
Thanks for implementing that change (adding an ID constant to the existing gateway classes). Looks good, though left some minor notes and suggestions:

- Need to tweak formatting in a few places to satisfy our linting rules.
- Not essential, but we could be more consistent in terms of how we reference classes (unnecessary qualifiers in lots of places).
- One or more checks are failing, but I think I spotted one of the problems.

kalessil:
@wjrosa @barryhughes, sorry for the late response - adding the PR to review queue on Monday.

plugins/woocommerce/includes/abstracts/abstract-wc-payment-gateway.php
wjrosa: Oh right. I removed all of them and imported the class when necessary on my latest commits.

wjrosa:
> Thanks for implementing that change (adding an ID constant to the existing gateway classes). Looks good, though left some minor notes and suggestions:
> 
> * Need to tweak formatting in a few places to satisfy our linting rules.
> * Not essential, but we could be more consistent in terms of how we reference classes (unnecessary qualifiers in lots of places).
> * One or more checks are failing, but I think I spotted one of the problems.

Thanks, Barry! I have fixed the lint issues, removed the changelog entry, the \ before the class name calls and saw tests are passing. When you have some time, I would appreciate additional 👀

barryhughes (commented):
Great! Nothing else to add on my end, so hitting approve (although linting still seems to be failing in relation to arrow alignment, so we should find a way to address that if possible).

barryhughes (approved):

Issue #52646, opened 2024-11-07
notice filling logs in upcoming wordpress 6.7/6.7.1 when using non-english default still happens with wc 9.4.2[also 9.5. beta] (and wrong translations also with wc 9.3.3)


Danielspain22:
### Prerequisites

- [x] I have carried out troubleshooting steps and I believe I have found a bug.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

When testing 6.7 and 6.7.1 and woocommerce 9.4.1(and now 9.4.2 also)  using main spanish language(or try any non-english default, but don't forget to update your translations), the following notice start filling the logs(also with wc 9.3.3):

[03-Nov-2024 02:36:53 UTC] PHP Notice:  Function _load_textdomain_just_in_time was called <strong>incorrectly</strong>. Translation loading for the <code>woocommerce</code> domain was triggered too early. This is usually an indicator for some code in the plugin or theme running too early. Translations should be loaded at the <code>init</code> action or later. Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 6.7.0.) in /home/local/mysite/wp-includes/functions.php on line 6087

The translations are ok if you previously update them , but the php notice persists.
the wc translations work with wc 9.41/9.4.2 but NOT anymore with  wc 9.3.3!!!! and wp 6.7!

### Expected behavior

no notice should appear in logs, this would be very conflictive(i know it's just a notice, but will fill the logs fast)

### Actual behavior

the logs being filled fast with the notice pasted before

### Steps to reproduce

install wp 6.7/6.7.1and woo 9.4.1/9.4.2(also happens with wc 9.3.3), enable or use the main spanish language for your site(or any non-english default),update translations, watch the logs

### WordPress Environment ###

WordPress address (URL): https://local.ar
Site address (URL): https://local.ar
WC Version: 9.4.0-rc.3 (also tested up to wc 9.4.1)
Legacy REST API Package Version: El plugin Legacy REST API no está instalado en este sitio.
Action Scheduler Version: ✔ 3.8.2
Log Directory Writable: ✔
WP Version: 6.7-RC3
WP Multisite: –
WP Memory Limit: 256 MB
WP Debug Mode: ✔
WP Cron: ✔
Language: es_ES
External object cache: –

### Server Environment ###

Server Info: LiteSpeed
PHP Version: 8.1.30
[...]

Issue #51843, opened 2024-10-02
Error: _load_textdomain_just_in_time Called Incorrectly for WooCommerce in WP 6.7.0-beta


kmjamsheed0:
### Prerequisites

- [X] I have carried out troubleshooting steps and I believe I have found a bug.
- [X] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

I'm currently running WordPress 6.7.0-beta, and I'm encountering the following error when visiting a product page or interacting with WooCommerce menus:

Function _load_textdomain_just_in_time was called incorrectly. Translation loading for the woocommerce domain was triggered too early. This is usually an indicator for some code in the plugin or theme running too early. Please see [Debugging in WordPress](https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/) for more information. (This message was added in version 6.7.0.)

The error seems to indicate that the translation loading for WooCommerce is happening too early in the execution order, likely due to a plugin or theme code being triggered prematurely.

### Expected behavior

No errors related to translation loading should be triggered during normal WooCommerce usage.

### Actual behavior

 The error message about _load_textdomain_just_in_time is displayed, indicating that translation loading was triggered too early.

### Steps to reproduce

1. Install WordPress 6.7.0-beta.
2. Install and activate WooCommerce.
3. Visit a product page or interact with WooCommerce menus.
4. The error message appears on the screen or in the debug log.

### WordPress Environment

WordPress Version: 6.7-beta1-59156
WooCommerce Version: 9.3.3
Theme: Storefront

### Isolating the problem

- [X] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [X] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/products/storefront/).
- [X] I can reproduce this bug consistently using the steps above.

webdados:
I can confirm this happens with WooCommerce 9.4.0-beta.2 and WordPress 6.7-beta1-59158

The notice has since been updated to 

`_load_textdomain_just_in_time` means that WooCommerce hasn't yet called `load_plugin_textdomain` (which is happening correctly on `init`) but a i18n function with the `woocommerce` textdomain was already called and WordPress is trying to load the domain "just in time".

[...]

PR #53955, opened 2024-12-20
Translation changes to improve compatibility with WP 6.7 upwards.

barryhughes:
This change attempts to remedy a problem with text domain handling under WP 6.7, while retaining compatibility with text domain handling under WP 6.6 (remembering that, at present, we [don't support versions older than 6.6](https://github.com/woocommerce/woocommerce/blob/9.5.1/plugins/woocommerce/woocommerce.php#L11)).

Updates #52646 → *does not close it, as the original basis of the issue is a problem with translation-related 'doing it wrong' notices, which this PR does not address directly.*

Let's establish some baselines around translation handling as it worked prior to WP 6.7. 

1. Setup WordPress 6.6 (the version is important to let us see translations in effect before the recent changes).
2. Install and activate [WooCommerce 9.4.3](https://downloads.wordpress.org/plugin/woocommerce.9.4.3.zip) (the version is important because the way we do [text domain unloading](https://github.com/woocommerce/woocommerce/blob/9.3.4/plugins/woocommerce/includes/class-woocommerce.php#L876) changed slightly after this version).
3. I propose enabling the French `fr_FR` translation, which is very complete. If you prefer to use something else that's fine, but you may need to adapt some of the following steps accordingly. I also ask that, initially, you ensure your user profile is using the default site language (in this case, that's French).
4. At this point, confirm that all (or most) strings are translated to French.
5. Now let's install a custom translation:

[woocommerce-fr_FR.zip](https://github.com/user-attachments/files/18216082/woocommerce-fr_FR.zip)

6. The above asset is a zip (GitHub doesn't allow `*.mo` files to be uploaded directly) ... unzip it, though, and you can extract the `*.mo` file, which you should copy to:

```
wp-content/languages/woocommerce/woocommerce-fr_FR.mo
```

7. What you need to know about this custom translation is that it translates only one string. Products is normally translated to *Produits,* but this file changes it to *Les Produits*  (this probably isn't good French, but the point is to be able to see that the custom translation is taking effect).
8. With this in position, if you refresh the admin environment you should see the **Produits** admin menu entry is now **Les Produits**. In addition, the remainder of the UI should still be translated, as per the canonical French translation.

Please now update to WP 6.7.1 and switch to this branch.

1. You should be able to verify that:
    1. Translations still work.
    2. *Custom* translations (like our `*.mo` file containing *Les Produits)* still work as before.
2. Let's now simulate one of the premature translation doing-it-wrong notices described in the original issue, by introducing some custom code like the following (which could temporarily live in `wp-content/mu-plugins/test-52646.php` or similar):

```php
<?php

// "Spike" the translation loading process.
add_action( 'plugins_loaded', function () {
    __( 'Early translation trigger', 'woocommerce' );
} );
```

3. With this in place (and activated, if necessary), refresh the admin UI. By monitoring your error logs, or by using a tool like [Query Monitor](https://wordpress.org/plugins/query-monitor/), you should see that a *doing-it-wrong* notice concerning the `woocommerce` text domain has been emitted. However, despite the notice, both the translation and custom translation should still function as before/should still take effect (you can compare this to how things behave with WooCommerce 9.5.1 to see the problematic handling some users are currently facing).

Last but not least, let's confirm that other facets of translation handling are not impacted. In particular, let's take a look at emails. Still using WP 6.7.1 and this branch, run through the following steps.

1. You will need some method of monitoring emails, like [Email Log](https://wordpress.org/plugins/email-log/).
2. With that in place, edit an order and **send order details to customer** *('Envoyer les détails de la commande au client')*.
3. The dispatched email should be in French (actually, this is a spot where gaps in the translation may show up ... so all we really want to know is that at least some of the email is being translated).
4. Now edit your user profile, change your preferred language to English (or to something other than French).
5. Repeat the test and trigger a fresh email. The email content (or a portion of it) should still be in French.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/includes/class-woocommerce.php
+++ b/plugins/woocommerce/includes/class-woocommerce.php
@@ -871,16 +871,21 @@ final class WooCommerce {
 	 *      - WP_LANG_DIR/plugins/woocommerce-LOCALE.mo
 	 */
 	public function load_plugin_textdomain() {
-		$locale = determine_locale();
-
 		/**
 		 * Filter to adjust the WooCommerce locale to use for translations.
 		 */
-		$locale = apply_filters( 'plugin_locale', $locale, 'woocommerce' ); // phpcs:ignore WooCommerce.Commenting.CommentHooks.MissingSinceComment
+		$locale                  = apply_filters( 'plugin_locale', determine_locale(), 'woocommerce' ); // phpcs:ignore WooCommerce.Commenting.CommentHooks.MissingSinceComment
+		$custom_translation_path = WP_LANG_DIR . '/woocommerce/woocommerce-' . $locale . '.mo';
+		$plugin_translation_path = WP_LANG_DIR . '/plugins/woocommerce-' . $locale . '.mo';
 
-		unload_textdomain( 'woocommerce', true );
-		load_textdomain( 'woocommerce', WP_LANG_DIR . '/woocommerce/woocommerce-' . $locale . '.mo' );
-		load_plugin_textdomain( 'woocommerce', false, plugin_basename( dirname( WC_PLUGIN_FILE ) ) . '/i18n/languages' );
+		// If a custom translation exists (by default it will not, as it is not a standard WordPress convention)
+		// we unload the existing translation, then essentially layer the custom translation on top of the canonical
+		// translation. Otherwise, we simply step back and let WP manage things.
+		if ( is_readable( $custom_translation_path ) ) {
+			unload_textdomain( 'woocommerce' );
+			load_textdomain( 'woocommerce', $custom_translation_path );
+			load_textdomain( 'woocommerce', $plugin_translation_path );
+		}
 	}
 
 	/**
--- a/plugins/woocommerce/changelog/fix-52646-translation-loading
+++ b/plugins/woocommerce/changelog/fix-52646-translation-loading
@@ -1,0 +1,4 @@
+Significance: patch
+Type: tweak
+
+Adjust translation handling; improve support for custom translations under WP 6.7 upwards.


barryhughes:
I'm applying the `impact: high` label because translations impact a lot of users, and because translation issues in general became a hot topic following the release of WP 6.7 ... so we may want to take some extra care.

jorgeatorres (commented):
@barryhughes: Thanks for working on this and apologies for the delay. This looks good to me and tests well. Hopefully nothing too serious, but I've left some comments inline.

jorgeatorres:
(GH seems to have lost my feedback 😓. Please bear with me while I re-post).

plugins/woocommerce/includes/class-woocommerce.php
```diff
@@ -871,16 +871,21 @@ public function init() {
...
-		load_plugin_textdomain( 'woocommerce', false, plugin_basename( dirname( WC_PLUGIN_FILE ) ) . '/i18n/languages' );
+		$locale                  = apply_filters( 'plugin_locale', determine_locale(), 'woocommerce' ); // phpcs:ignore WooCommerce.Commenting.CommentHooks.MissingSinceComment
+		$custom_translation_path = WP_LANG_DIR . '/woocommerce/woocommerce-' . $locale . '.mo';
+		$plugin_translation_path = WP_LANG_DIR . '/plugins/woocommerce-' . $locale . '.mo';
+
+		// If a custom translation exists (by default it will not, as it is not a standard WordPress convention)
+		// we unload the existing translation, then essentially layer the custom translation on top of the canonical
+		// translation. Otherwise, we simply step back and let WP manage things.
+		if ( is_readable( $custom_translation_path ) ) {
+			unload_textdomain( 'woocommerce' );
```
jorgeatorres, line 885: I really like that we're now enabling the custom code only when a custom translation is already in place, but I wonder if we should remove the `$reloadable = true` arg in the `unload_textdomain()`. Depending on what other things happen during script execution, wouldn't this be useful to have? Have you seen that break something else?

plugins/woocommerce/includes/class-woocommerce.php
```diff
@@ -871,16 +871,21 @@ public function init() {
...
-		$locale = determine_locale();
-
 		/**
 		 * Filter to adjust the WooCommerce locale to use for translations.
 		 */
-		$locale = apply_filters( 'plugin_locale', $locale, 'woocommerce' ); // phpcs:ignore WooCommerce.Commenting.CommentHooks.MissingSinceComment
-
-		unload_textdomain( 'woocommerce', true );
-		load_textdomain( 'woocommerce', WP_LANG_DIR . '/woocommerce/woocommerce-' . $locale . '.mo' );
-		load_plugin_textdomain( 'woocommerce', false, plugin_basename( dirname( WC_PLUGIN_FILE ) ) . '/i18n/languages' );
```
jorgeatorres, line 883: We have a README file in `i18n/languages` that indicates that custom translations should instead be stored in `WP_LANG_DIR` but thanks to `load_plugin_textdomain()` they still worked. Those wouldn't work anymore after this change. Is this truly something we want do to now?

Unless I'm missing something, `load_plugin_textdomain()` will _still_ attempt to load the translation from `WP_LANG_DIR` _first_ so maybe it could actually simplify things, though I might be missing something.

I'm not against removing support for the old behavior, just mentioning it because I'm not sure if we've truly walked the deprecation path.

plugins/woocommerce/includes/class-woocommerce.php
barryhughes: Right! One might think so. In practice, at least in my testing, this would mean that the "custom" translation (the one that translates *products* as *les produits)* will no longer take effect.

barryhughes: To simulate, follow the testing instructions then modify this code as you suggest. On refreshing the dash, you should see that *products* is translated as *produits* (ie, the custom translation file is skipped).

plugins/woocommerce/includes/class-woocommerce.php
barryhughes: > Is this truly something we want do to now?

It felt like a good opportunity to me, but you're right to challenge it, as we don't want to cause unnecessary disruption. 

Yet, someone who actively utilizes this is in a difficult spot when the time comes to update WooCommerce, and is going against the advice you mentioned ... which has been in place for a considerable length of time (almost a decade, in fact [*](https://github.com/woocommerce/woocommerce/tree/3.0.0/i18n/languages)).

> Unless I'm missing something, load_plugin_textdomain() will still attempt to load the translation from WP_LANG_DIR first so maybe it could actually simplify things, though I might be missing something.

So I think you mean replacing this:

```php
load_textdomain( 'woocommerce', $custom_translation_path );
load_textdomain( 'woocommerce', $plugin_translation_path );
```

With:

```php
load_plugin_textdomain( 'woocommerce', false, $path_to_woo_i18n_dir );
load_textdomain( 'woocommerce', $custom_translation_path );
```

Or possibly the above, but with the lines reversed. I can't actually get either formulation to work as one might expect (at least, not under WP 6.7 ... but, perhaps this is now [just how things are](https://github.com/woocommerce/woocommerce/issues/51843#issuecomment-2416087379).

barryhughes: ☝🏼 That was a long and meandering response but, in summary, if you feel that this is not the right time to drop support for language files living in `i18n/languages`, I don't have any particular opposition to that. 

Also quite happy if we want to retain our use of `load_plugin_textdomain`, but I can't immediately see how that will work (could certainly be missing something, and doubtless something obvious, on my end).

plugins/woocommerce/includes/class-woocommerce.php
jorgeatorres: I didn't see that happening when I tried and I just tried again, but I might be missing something. At least on my instance (WP 6.7.1) the file wasn't skipped.
I don't think this is particularly serious but would be good if we could both reproduce the same results.

jorgeatorres: Scratch that. So... things are more complicated as usual and I'm the one who's wrong here. If I don't force an "early" translation, things do work (despite the `$reload = true`), but if I do, then it works as you described. Weird, but 🤷.

I'm not particularly happy of an early translation making things work entirely different but I guess that's beyond what we could do.

plugins/woocommerce/includes/class-woocommerce.php
jorgeatorres: > It felt like a good opportunity to me, but you're right to challenge it, as we don't want to cause unnecessary disruption.

I'm with you here and not against the removal of this functionality. Was just thinking that despite the warning it's been working, so we should at least add a note or something to the release notes to indicate that, well, it will no longer work.

> So I think you mean replacing this: [...]

Yeah, though now I can't reproduce. I tried so many things while reviewing last time that I probably left some file somewhere. In any case, as mentioned above, I'm good with removing the functionality as long as we at least have something bout it in the release notes.

jorgeatorres (approved):
I'm approving as this tests as described. I also tried to test how this worked in relation with other translation plugins (LocoTranslate in particular) and things were ok: default translation + custom translation + LocoTranslate translation are all correctly loaded.

I wish we didn't have to do all this custom code but we could even partially blame WP for it. In any case, this is definitely an improvement. Thank you @barryhughes!

Only thing I would note is that with this PR, translations in `<plugin dir>/i18n/languages` will stop working. Despite this being discouraged since (at least) 2014 (see https://github.com/woocommerce/woocommerce/commit/5a53a238c5b9f5fbbbf28979404eff0e91c03d88), we should probably feature something about it in the release notes for 9.8.0.

plugins/woocommerce/includes/class-woocommerce.php
barryhughes: Yeah, that's now beyond our control (or, is beyond our control as of WP 6.7). At least as I currently understand things.

barryhughes:
Thank you for the review!

> We should probably feature something about it in the release notes for 9.8.0.

I agree. I've applied `release: highlight` and will follow-up in the relevant places for that.

plugins/woocommerce/includes/class-woocommerce.php
naman03malhotra: Trying to wrap my head around this PR, I am wondering what will happen to the sites that might have some custom translation sitting in `/i18n/languages` folder after this change?

From the discussion above, it seems like it won't be supported anymore? Unless I am missing something, should we preserve it for a few releases and plan a deprecation path informing the users a few versions in advance?

Sorry if I missed something obvious here.

naman03malhotra: Trying to wrap my head around this PR, I am wondering what will happen to the sites that might have some custom translation sitting in `/i18n/languages` folder after this change?

From the discussion above, it seems like it won't be supported anymore? Unless I am missing something, should we preserve it for a few releases and plan a deprecation path informing the users a few versions in advance?

Sorry if I missed something obvious here.

plugins/woocommerce/includes/class-woocommerce.php
```diff
@@ -871,16 +871,21 @@ public function init() {
...
+		$locale                  = apply_filters( 'plugin_locale', determine_locale(), 'woocommerce' ); // phpcs:ignore WooCommerce.Commenting.CommentHooks.MissingSinceComment
+		$custom_translation_path = WP_LANG_DIR . '/woocommerce/woocommerce-' . $locale . '.mo';
+		$plugin_translation_path = WP_LANG_DIR . '/plugins/woocommerce-' . $locale . '.mo';
+
```
naman03malhotra, line 880: based on my previous comment, we may also add a support for `$path_to_woo_i18n_dir` if we still want to keep it.

plugins/woocommerce/includes/class-woocommerce.php
naman03malhotra: It's me again! So I did miss reading about this [commit](https://github.com/woocommerce/woocommerce/commit/5a53a238c5b9f5fbbbf28979404eff0e91c03d88), it has been a while (since 2014) since we have discouraged it. That said, there could be sites out there with translations in i18 directory, and as someone bitten in the past of depreciating something obvious, it bit us back. Should we consider giving a heads up to the community about this removal, a version or two earlier?

All that said, I am down with the direct removal as well, if we all agree.

barryhughes: > From the discussion above, it seems like it won't be supported anymore?

This is true. Note that translations placed in that location would anyway be overwritten upon update, so someone who both maintains custom translations in this location *and* actively updates their plugin has work to do in any case. Additionally:

> Unless I am missing something, should we preserve it for a few releases and plan a deprecation path.

Well, as touched on above, we've been recommending against placing translations in this location for over a decade. From version 2.3 (released back in 2015): ["WARNING! DO NOT PUT CUSTOM TRANSLATIONS HERE! WooCommerce will delete all custom translations placed in this directory."](https://github.com/woocommerce/woocommerce/tree/2.3.0/i18n/languages)

Not opposed to supplementing that with a further and renewed period of deprecation, though, if we think that's best and safest.

barryhughes: > All that said, I am down with the direct removal as well, if we all agree.

It seems reasonable to me, and all the more so if we put out a dev advisory (and/or some other comms) ahead-of-time.

barryhughes: ...If we merged soonish, the change would make its way into 9.8, scheduled for Mar 31, which gives us time to raise awareness. I guess, also thinking of previous deprecation experiences, targeted in-app notifications (ie, flag that a change is coming if we detect translations from this directory are in place) may be the gold standard, because not everyone follows the dev blog. I'm admittedly a little unsure if that is necessary, in this particular scenario.

barryhughes:
✍🏼 Rebased/updated the branch, as it was somewhat out-of-sync.

naman03malhotra (approved):
Thank you for working on it, its much better now!

Though I am not able to reproduce the doing it wrong notice, sending your the details over DM.

barryhughes:
Possibly a conflict? I logged in to the test site, deactivated everything except WooCommerce and the mail logger (plus Query Monitor), and the notice surfaced.

naman03malhotra (approved):
image
Everything works, thanks for sorting it out!

barryhughes:
@naman03malhotra @jorgeatorres just want to verify we're all aligned, to eliminate the potential for any misunderstandings ... it seems that once checks are passing, we can hit merge (and it will then target 9.8). Does that sound good to you both?

Zero problem on my end if we'd rather hold and test/explore/talk out the solution some more.

jorgeatorres:
That sounds good to me @barryhughes. Thanks!

naman03malhotra:
Sounds like a plan! 🚢

barryhughes:
Thanks all!

gigitux:
On WordPress 6.8 beta 1, this warning is triggered:


https://github.com/user-attachments/assets/4a1b1ae7-f526-4e58-8891-47a3ab680635

@barryhughes can you take a look at it?

barryhughes:
Thanks for flagging, @gigitux!

It looks to me like this problem originates with [this change](https://github.com/woocommerce/woocommerce/pull/55754) (heads-up @triple0t and @costasovo) and specifically from [this line](https://github.com/woocommerce/woocommerce/blob/c938f12469ee29c0e6b92f752f34600c87ea7648/plugins/woocommerce/src/Internal/EmailEditor/Package.php#L38), which results in the loading of translatable strings (in this case, the strings that are a part of the feature definition) very early in the request. This is currently happening during `plugins_loaded`, but we should probably be waiting until `before_woocommerce_init` if possible.

It's not really in the scope of this PR to handle this situation, so I'd propose one or both of the following as ways to proceed:

1. Possibly get input from the team who authored the EmailEditor package. Can interactions with `FeaturesUtil` be deferred until later in the request? If that's feasible, then we can avoid calling WP's translation functions pre-init, which is what leads to the warning you observed.
2. Some months ago, there were some thoughts about allowing early interaction with `FeaturesUtil` in a way that would avoid this problem (one idea was to use 'lazy strings' in our feature definitions). If we think there may still be value in this, then we'd probably need @woocommerce/proton to drive it.

gigitux:
Thanks for checking this, @barryhughes!

> Possibly get input from the team who authored the EmailEditor package. Can interactions with FeaturesUtil be deferred until later in the request? If that's feasible, then we can avoid calling WP's translation functions pre-init, which is what leads to the warning you observed.

They worked on PR to fix this warning https://github.com/woocommerce/woocommerce/issues/56113. Also, they opened [a CFE for WooCommerce 9.8](https://github.com/woocommerce/woocommerce/issues/56135). I would say all good! 

Thanks again!

PR #53999, opened 2024-12-23
Add document about removing checkout fields and add FAQ doc

opr:

Add "Removing Checkout fields" document and an FAQ document.

The FAQ Document is intended to be a work in progress that we will continue to iterate on.

1. Read the documents and ensure they don't have grammar or formatting errors.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


docs/cart-and-checkout-blocks/faq.md
```diff
@@ -0,0 +1,72 @@
...
+        add_action(
+          'woocommerce_cart_calculate_fees',
+          function( $cart ) {
+            $cart->add_fee( 'Custom Fee', 10 );  
+        } );
+      }
+    ]
+  );
+} );
+```
```
senadir: This example is incomplete as the fee won't persist on checkout submit flow. Maybe you can change this to something else (apply a coupon programmatically, select a payment method...) anything that actually persists in cart or session.

See https://nadir.blog/2024/07/02/reacting-to-selected-payment-method-in-woocommerce-checkout-block/

docs/cart-and-checkout-blocks/faq.md
```diff
@@ -0,0 +1,72 @@
...
+extensionCartUpdate( {
+	namespace: 'extension-unique-namespace',
+} ).then( () => {
+	// Cart has been updated with the fee.
+  // Don't need to do anything else here as it's automatically updated.
+} ).catch( ( error ) => {
+	// Handle error.
+	processErrorResponse(error);
+} );
+```
```
senadir: Code example works fine without adding `then` and catching directly, but maybe we keep it so people know?

docs/cart-and-checkout-blocks/faq.md
```diff
@@ -0,0 +1,72 @@
...
+
+### How to force-refresh the cart from the server
+
+This can be achieved using [`extensionCartUpdate`](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce-blocks/docs/third-party-developers/extensibility/rest-api/extend-rest-api-update-cart.md) which is the preferred way, but it is also possible by executing the `receiveCart` action on the `wc/store/cart` data store with a cart object, like so:
+
+```js
+const { dispatch } = window.wp.data;
+
+dispatch( 'wc/store/cart' ).receiveCart( cartObject )
```
senadir, line 116: The challenge here is that you need a valid `cartObject`?

Another way to force refresh without a valid cart element is to use:

```js
wp.data.dispatch('wc/store/cart').invalidateResolutionForStore()
```

Keep note that will cause the address block to flash, something we should probably fix on our side.

docs/cart-and-checkout-blocks/removing-checkout-fields.md
```diff
@@ -0,0 +1,197 @@
+---
+post_title: Cart and Checkout - Removing checkout fields
+menu_title: Extensibility
+tags: reference
+---
+<!-- markdownlint-disable MD041 -->
+
+If you're trying to extend the new Checkout block, you might notice the previous `checkout_fields` isn't working. This is intentional, we're taking back control of how to render and handle fields, yet, one of the most common requests we get is how to disable Checkout fields for certain order types. For a long time, our answer was "you can't". You actually can, but we don't encourage it.
```
senadir: Not sure if we want to include this verbatim? I'm more comfortable speaking like this in my blog but dunno from Woo? Let's see keep it and see what dev advocacy say later.

docs/cart-and-checkout-blocks/removing-checkout-fields.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Why we don't encourage removing fields
+
+The simple reason is that Checkout is complex, and many plugins need different data, payment methods will need your address info to run fraud detection, or validate your card, tax (including core) will need your full address to calculate taxes correctly, this is something we still think about in Checkout and are looking for a universal solution, until then, you can go with code, and here's how:
+
+## Disabling fields for a single country
+
+Checkout fields still country locale, which means you can modify fields for a country via PHP and this would work fine. The following billing form has 10 fields, 7 of them are required:
```
senadir:
```suggestion
Checkout fields still respects country locale, which means you can modify fields for a country via PHP and this would work fine. The following billing form has 10 fields, 7 of them are required:
```

docs/cart-and-checkout-blocks/removing-checkout-fields.md
```diff
@@ -0,0 +1,197 @@
...
+```php
+add_filter('woocommerce_get_country_locale', function( $locale ) {
+
+	$locale['DZ']['postcode']['required'] = false;
+	$locale['DZ']['postcode']['hidden'] = true;
+
+	$locale['DZ']['city']['required'] = false;
+	$locale['DZ']['city']['hidden'] = true;
+
+		return $locale;
```
senadir:
```suggestion
	 return $locale;
```

docs/cart-and-checkout-blocks/removing-checkout-fields.md
```diff
@@ -0,0 +1,197 @@
...
+		'required' => false,
+		'hidden'   => true,
+	];
+
+	$locale['DZ']['state'] = [
+		'required' => false,
+		'hidden'   => true,
+	];
+
+		return $locale;
```
senadir:
```suggestion
	return $locale;
```

docs/cart-and-checkout-blocks/removing-checkout-fields.md
```diff
@@ -0,0 +1,197 @@
...
+
+	$locale['DZ']['city'] = [
+		'required' => false,
+		'hidden'   => true,
+	];
+
+	$locale['DZ']['state'] = [
+		'required' => false,
+		'hidden'   => true,
+	];
```
senadir, line 71:
```suggestion
	$locale['DZ']['company'] = [
		'required' => false,
		'hidden'   => true,
	];
	
	$locale['DZ']['state'] = [
		'required' => false,
		'hidden'   => true,
	];
	
	$locale['DZ']['phone'] = [
		'required' => false,
		'hidden'   => true,
	];
```

Since published, we can now remove phone and company via PHP, I added them here.

docs/cart-and-checkout-blocks/removing-checkout-fields.md
```diff
@@ -0,0 +1,197 @@
...
+```
+
+This is the result:
+
+![Image](https://github.com/user-attachments/assets/19c82877-3405-4762-82ce-e952746abe66)
+
+You notice 2 things:
+
+- Address Line 2 is not visible, this is because it will always be skipped if Address Line 1 is skipped.
+- Phone and Company can be controlled the same way as others.
```
senadir:
```suggestion
You notice that Address Line 2 is not visible, this is because it will always be skipped if Address Line 1 is skipped.
```

docs/cart-and-checkout-blocks/removing-checkout-fields.md
```diff
@@ -0,0 +1,197 @@
...
+
+If you want to change the value depending on locale (overriding the merchant value for that country), you can use the examples above.
+
+Till now, all changes are applied to Algeria only, customers switching would see the correct fields for each country:
+
+![Image](https://github.com/user-attachments/assets/3b8cb49a-1c95-4fab-8aaa-26b14ce22aad)
+
+## Applying changes to all countries you sell to
+
+As with the above, this is not something I recommend unless you sell to a very controlled set of countries and tested this with each country. Some payment gateways would only enable fraud detection on production mode and not test mode, and your fields that passed test mode fails in production mode, but here it is regardless:
```
senadir:
```suggestion
As with the above, this is not something we recommend unless you sell to a very controlled set of countries and tested this with each country. Some payment gateways would only enable fraud detection on production mode and not test mode, and your fields that passed test mode fails in production mode, but here it is regardless:
```

senadir (approved):
Looking good, left some suggetions.

docs/cart-and-checkout-blocks/faq.md
opr: Why won't it persist on checkout? It's using the `woocommerce_cart_calculate_fees` action

senadir: Because it's called inside `woocommerce_store_api_register_update_callback` and that's only called once (per your action) and not on further request.


https://github.com/user-attachments/assets/75a03ce1-1d8d-46a9-9d28-2d263b36fd06



Fees are never persisted to anything, they're runtime values, so for this to work on submit you need to also run the action again.

docs/cart-and-checkout-blocks/faq.md
opr: Hm, I'll remove it, the link to `extensionCartUpdate` contains this full example

docs/cart-and-checkout-blocks/faq.md
opr: Ahh, I see what you mean, alright thanks will update

opr: The point about fees is that it is an FAQ, specifically for fees, so that's why we want to keep it.

opr: Oh, this doesn't need to be an extensionCartUpdate example, you're right.

docs/cart-and-checkout-blocks/removing-checkout-fields.md
opr: I reworded it.

opr:
@senadir I made some changes would you mind re-checking them?

senadir (approved):
Looking good! Thank you Thomas

PR #54001, opened 2024-12-23
Introducing product visibility constants

wjrosa:

This PR introduces constants for all product visibility identifiers. The idea is to make it easier to reference them, avoid typos, make the code stricter, make it easier to find status usages, centralize them, improve documentation, and many other reasons that should help developers create code related to them.

- Checkout and build this branch on your test environment (`dev/introducing-visibility-constants`)
    - Store setup
    - Product creation
    - Common checkout
    - Shortcode checkout
    - My account page navigation
    - Order listing/details

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Introduces and implements product visibility constants in the WooCommerce plugin codebase.

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/changelog/54001-dev-introducing-visibility-constants
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: tweak
+Comment: Just introduces new product visibility constants. A changelog is not required.
+
```
wjrosa, line 3: I think the changelog generator might be broken 🤔

barryhughes: What did we do for related changes (where we also introduced similar enum-style classes)? It feels like we should probably have at least one changelog entry to cover them, assuming they all ship in the same release.

barryhughes: ...In any case... why do you think it might be broken? It looks valid, and passes the validation check 🤔

plugins/woocommerce/src/Enums/ProductVisibility.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace Automattic\WooCommerce\Enums;
+
+/**
+ * Enum class for all the product visibility.
+ */
+class ProductVisibility {
```
barryhughes: Elsewhere in the codebase, but also I think in the docs, we tend to speak of *catalog* visibility. To that end, what do you think of renaming this class as `CatalogVisibility`? Feels like that might align a little more tightly.

plugins/woocommerce/src/Enums/ProductVisibility.php
```diff
@@ -0,0 +1,32 @@
...
+	 */
+	const CATALOG = 'catalog';
+
+	/**
+	 * Product visible in search.
+	 */
+	const SEARCH = 'search';
+
+	/**
+	 * Product invisible.
```
barryhughes: Especially if we decide *not* to rename the class, it may be worth flagging what hidden means. Specifically, we should probably highlight that the product is still accessible if navigated to directly.

barryhughes (commented):
Nice! Left a couple of notes. I'm also going to select a reviewer from Proton to replace me here (just because it feels like for any further changes such as these, they ought to see the PRs).

plugins/woocommerce/src/Enums/ProductVisibility.php
coreymckrill: +1, I think matching the name of the class up with the name of the product property will make it easier to reason about the code.

plugins/woocommerce/src/Enums/ProductVisibility.php
coreymckrill: It might be good to use the labels from `wc_get_product_visibility_options` in the doc blocks here, as they succinctly describe what each of the values actually does. But I also agree that adding a bit more to the `hidden` option to clarify that you can still access the product directly would also be good.

plugins/woocommerce/changelog/54001-dev-introducing-visibility-constants
wjrosa: > What did we do for related changes (where we also introduced similar enum-style classes)? It feels like we should probably have at least one changelog entry to cover them, assuming they all ship in the same release.

For the previous PRs, I have included the changelog, but someone said it was not needed in this case (not sure if it was you or Vladimir).

> ...In any case... why do you think it might be broken? It looks valid, and passes the validation check 🤔

I thought that by setting a message for the "changelog not needed" portion of the PR description the actions would skip the changelog creation, but it created a changelog with that message instead. I tried to roll it back today by adding the changelog entry again, but now it was ignored. Not sure how to proceed

wjrosa: Now it worked (https://github.com/woocommerce/woocommerce/pull/54001/commits/6e5a53a27d455c1969fde505496457d22a3e1a57)

plugins/woocommerce/src/Enums/ProductVisibility.php
wjrosa: Oh right, renamed it in https://github.com/woocommerce/woocommerce/pull/54001/commits/5c1204ea361cfb5aeb9957c472900706eabda3c5

plugins/woocommerce/src/Enums/ProductVisibility.php
wjrosa: Makes sense! Improved the docs on https://github.com/woocommerce/woocommerce/pull/54001/commits/41ace7941403a285e9edc0baf71d5636bc0ea734

plugins/woocommerce/changelog/54001-dev-introducing-visibility-constants
barryhughes: > I thought that by setting a message for the "changelog not needed" portion of the PR description the actions would skip the changelog creation

One or more changelog _files_ are needed for each PR (usually just one), but a changelog _entry_ is not always needed.

So, in the second case where we don't actually want to create a changelog entry for the final release, the changelog file is structured like this (similar to the original version):

```
Significance: patch
Type: tweak
Comment: <EXPLAIN WHY A CHANGELOG IS NOT NEEDED>
```

As I understand it, this gives the reviewer a chance to think about this and indicate if they think that, actually, a changelog *is* needed.

The normal form, as in commit https://github.com/woocommerce/woocommerce/commit/6e5a53a27d455c1969fde505496457d22a3e1a57, doesn't contain a comment field so that will lead to an actual changelog entry being created when these files are parsed closer to the actual release.

(I might be misunderstanding and/or you may already know this ... so just ignore if it's not helpful!)

> Someone said it was not needed in this case (not sure if it was you or Vladimir).

I also don't remember. Not a big problem and, to clarify my view on this, it just felt like a situation where we only need one changelog per release, since all these enum PRs are doing pretty similar stuff from a developer's perspective.

plugins/woocommerce/src/Enums/CatalogVisibility.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace Automattic\WooCommerce\Enums;
+
+/**
+ * Enum class for all the catalog visibility values.
+ */
+class CatalogVisibility {
```
kalessil: Shouldn't it be final?

kalessil: Also, I belive we need to update https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Enums/README.md to mention this enum.

kalessil (approved):
LGTM!

Leaving a couple of comments to address before merge.

plugins/woocommerce/src/Enums/CatalogVisibility.php
wjrosa: Right! Updated it on https://github.com/woocommerce/woocommerce/pull/54001/commits/168958b2e244b0141bbe56163f999e2af633d842

wjrosa: Done in https://github.com/woocommerce/woocommerce/pull/54001/commits/5d78609cc6fa817dabfdc8f4136394b81c6e31ac

PR #54002, opened 2024-12-23
[Settings]: update @wordpress/edit-site dependency to 6.14.0

### LINKED ISSUE #53818: [Settings] Update Router usage to Calypso site admin package
https://github.com/Automattic/wp-calypso/pull/99754 Provides `automattic/site-admin` package that contains a router for use in new site admins.

Update `RouterProvider` to the one from the Calypso package. This will remove Settings dependency on Gutenberg 19.8.

retrofox:

Currently, using WooCommerce Settings with the Gutenberg development branch is not possible.
The `@wordpress/edit-site` API has changed considerably, introducing breaking ones.

This PR updates the package's dependency and relevant changes to make it work:

### `routes`

This PR introduces the settings routes used to decorate the different parts of the layout in the same way the core does.
It uses the `registerRoute` private action, and also the `getRoutes` private selector.

### `Sidebar`

* It uses the `SidebarNavigationItem` core component to render the sidebar
* It removes the custom `SettingsItem` component e76317fbb94b578ccbd176d9b811a8109d0c73f4

### Styles

It uses core mixings to reset part of the layout

Closes https://github.com/woocommerce/woocommerce/issues/53818

1. Update the dependencies and build the app
2. Confirm it's possible to change the Settings section by clicking on the sidebar

https://github.com/user-attachments/assets/75a2c49f-b07b-4b1e-8659-156270829e71

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


packages/js/settings-editor/src/components/routes/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+					settingsPage={ settingsPage }
+					activeSection="default"
+				/>
+			),
+			mobile: <div>Mobile</div>,
+		},
+	};
+} );
+
+export function useRegisterSiteEditorRoutes() {
```
chihsuan: Is it necessary to register the routes to the edit-site store? I feel it is a bit confusing to have the `useRegisterSiteEditorRoutes` function in the setting editor and also have another `route.tsx` file in the setting editor. 🤔 

If this is needed, would it be better to move it to a separate folder since this is not a React component?

packages/js/settings-editor/src/components/routes/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Internal dependencies
+ */
+import { Sidebar } from '../sidebar';
+import { LegacyContent } from '../../legacy';
+import { getSettingsSectionPath } from './utils';
+
+const settingsData: SettingsData = window.wcSettings?.admin?.settingsData || [];
+
+const routes = Object.values( settingsData ).map( ( settingsPage ) => {
```
chihsuan: Do we need to include modern page routes (e.g. register the routes via JS filter) here?

chihsuan (commented):
Thanks for working on this! @retrofox 👍


It looks like CI is failing because the lockfile is out of date. Could you fix that? 

Here's the error message:


```bash
 ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with packages/js/settings-editor/package.json
```

I also left a few comments in the code first.

packages/js/settings-editor/src/components/routes/index.tsx
retrofox: Probably yes. Here is where I need help to be aware of the implementation details. Looking 👀

packages/js/settings-editor/src/components/routes/index.tsx
retrofox: > Is it necessary to register the routes to the edit-site store?

This PR aims to bring the implementation closer to the core. I researched how the Core handles the layout decoration. This component is a copy, meaning that we could probably reuse the core one.

Technically, it's a component (React component) :-)
Core also puts the file which defines it into a `components/` folder.
But if we consider only visual components, can we move this component into a `hooks/` folder?

chihsuan: > This PR aims to bring the implementation closer to the core. I researched how the Core handles the layout decoration. This component is a copy, meaning that we could probably reuse the core one.

Thanks for the context! However, IMO it's unclear why we need to register routes to editSite store. 🤔  I think it makes sense to reuse UI components, but we should avoid coupling the setting editor with the edit Site store. (I might missed something, if so, please let me know!)

I was also wondering why we need to register routes here but also have original route.tsx. Is it possible to merge them?

> But if we consider only visual components, can we move this component into a hooks/ folder?

Yeah, that's what I was thinking! Thanks. 🙂

packages/js/settings-editor/src/components/routes/index.tsx
retrofox: I've reorganized the files

packages/js/settings-editor/src/components/routes/index.tsx
retrofox: > Thanks for the context! However, IMO it's unclear why we need to register routes to editSite store.

The short answer is that that's what Core does. If the question is why we use the edit-site store instead of a different one, I think the routes API covers the requirements for addressing the layout decoration.

The contra question is: why will we do something different when it's already addressed in core?

> I was also wondering why we need to register routes here but also have original route.tsx. Is it possible to merge them?

If we agree that the core should address it, I think the file should be removed, and everything should be handled through the core API.

retrofox: We should avoid duplication as much as possible.  
For example, we previously had a `<SidebarItem />` component that relied on a custom `useLink()` hook.
However, the core already provides an internal [useLink()](https://github.com/WordPress/gutenberg/blob/trunk/packages/router/src/link.tsx#L12) hook, which made our custom implementation redundant.  

To address this, I removed the custom component in this commit e76317fbb94b578ccbd176d9b811a8109d0c73f4.  

This kind of duplication increases maintenance overhead and introduces inconsistencies, which we should strive to minimize.

chihsuan: > The short answer is that that's what Core does. 

I agree that we should be closer to Gutenberg and reuse as much as possible.

> If the question is why we use the edit-site store instead of a different one, I think the routes API covers the requirements for addressing the layout decoration.

If I’m not mistaken, the Routing API is provided by `@wordpress/routes,` while the edit-site package is private and intended solely for purposes related to block editing. The edit-site store provides APIs for registering and retrieving routes. But it feels strange to use the edit-site store to register our routes. Also, shouldn't we consider whether this API is intended to be opened up for use by other packages later? 

It looks like we need to provide the routes to the `RouterProvider`. Can't we just pass the routes to the `RouterProvider`?

```
<RouterProvider routes={ routes } pathArg="page">
```

chihsuan: > To address this, I removed the custom component in this commit https://github.com/woocommerce/woocommerce/commit/e76317fbb94b578ccbd176d9b811a8109d0c73f4.

> This kind of duplication increases maintenance overhead and introduces inconsistencies, which we should strive to minimize.

Yeah, I saw the commit. That's a great improvement!

psealock: I believe the edit site package is will never be exposed publicly. I understand the desire to replicate whats in core, but the long term goal will be to transition to apis as they become public, so it makes sense to avoid using the edit site store if we can.

> It looks like we need to provide the routes to the RouterProvider. Can't we just pass the routes to the RouterProvider?

Would this work? It seems logical to me. Plus, this avoids two round trip requests to the server on page load if I'm not mistaken?

retrofox: > I understand the desire to replicate whats in core, but the long term goal will be to transition to apis as they become public, so it makes sense to avoid using the edit site store if we can.

We are already importing packages from the `/build-module` folder. That's why I started to import other ones. What are the criteria for importing non-public libs from there?

I prefer to pick private resources instead of creating custom ones because:
* It forces us to get more aligned with the core
* It avoids re-writing code, not re-inventing the wheel.
* In theory, it should be more productive

Once private resources become public, changing how they are imported will be a matter of adjusting. In the worst-case scenario, we'll create custom implementations in the Analytics repository.

I prefer this instead of creating something new, TBH.

chihsuan: I think we can import private apis from the build-module folder as long as they will be public in the future and are intended to be used by other plugins. 

I totally agree with the points you made, but I just think not all code in Core is designed to be used by other plugins. There are some codes that are only intended to be used by the editor, and IMO, `useRegisterSiteEditorRoutes` is one of them. More importantly, it seems like we don't need to use it to use `RouterProvider`. It looks like simply pass the routes to the `RouterProvider` and it will work.

retrofox: `useRegisterSiteEditorRoutes` is a custom hook. I've renamed it with something more familiar: `useRehgisterSettingsRoutes`.

>  It looks like simply pass the routes to the RouterProvider and it will work.

We can pass the routes to the `RouterProvider` through its `routes` property, among other props.
The question is, why not follow a similar implementation to Core's? For instance, using the store to register the routes.

I don't know what Core's plan is about publicizing private APIs.

AFAICS, we are importing the following `@wordpress/edit-site` components/libs from build-module/:

* `unlock()`
* `<SiteHub />`
* `<SidebarContent />`
* `<SidebarNavigationScreen />`
* `<SidebarNavigationItem />` (imported here in this PR)

And also using the following private API.

* `<RouterProvider />` from '@wordpress/router'
* `getRoutes()` selector from `@wordpress/edit-site` (this PR)
* `<NavigableRegion />` from `@wordpress/editor`
* `useLocation()` from '@wordpress/router'
* `useHistory` from '@wordpress/router'
* `registerRoute()` action from `@wordpress/edit-site` (this PR)

Should we ask for feedback from devs more familiar with the implementation?

This huge PR contains the most relevant changes https://github.com/WordPress/gutenberg/pull/67199, part of https://github.com/WordPress/gutenberg/issues/60584

chihsuan: > We can pass the routes to the RouterProvider through its routes property, among other props.
The question is, why not follow a similar implementation to Core's? For instance, using the store to register the routes.

If we want to follow Core's implementation, IMO we should create a wc-settings store instead of using edit-site store. I don't see the reason why we need to register the routes in the store.

> Should we ask for feedback from devs more familiar with the implementation?

Sounds good! The `@wordpress/edit-site` is a special case. It's not intended to be used in other plugins and doesn't exports these private APIs in root index.js. We basically hack it by importing from build-module folder. So I think it's very different from `@wordpress/router`.

retrofox: > > We can pass the routes to the RouterProvider through its routes property, among other props.
> > The question is, why not follow a similar implementation to Core's? For instance, using the store to register the routes.
> 
> If we want to follow Core's implementation, IMO we should create a wc-settings store instead of using edit-site store. I don't see the reason why we need to register the routes in the store.

Yeah, I don’t know. I tend to think the opposite: why create a new store that does the same, which is, in short, dealing with the app routing? I'm okay creating a new store, tho. Let's ask to dev folks.

PR #54017, opened 2024-12-24
dev: added env vars to allow disabling and or hiding typescript warnings

rjchow:

After merging https://github.com/woocommerce/woocommerce/pull/53531 the long list of typescript warnings made it impossible to use `pnpm watch:build` because it takes over a few minutes to finish printing the warnings each time you save.

Adding env vars to allow
- disabling typescript checking for webpack (default: true - typescript disabled)
- hiding typescript warnings for webpack (default: false - not hidden)

This default combination means that anyone who's not actively working on fixing the typescript errors is not impeded by the warnings.

There exists the combination of enabling TS checking but also hiding the warnings, this is to facilitate working on fixing the errors. 

When the specific error code in `webpack.config.js` is commented out, it is converted to an error. This error however becomes hard to find in the backscroll of warnings, so it's been made possible to hide the warnings and focus only on the focused errors.

There's however an issue I couldn't resolve - it doesn't seem like ignoreWarnings works when ForkTsCheckerWebpackPlugin is running in async mode. It runs in async mode when `webpack --watch` is used. So, the workaround is to only use `npx webpack` to print the errors.

So for example, if I'd like to work on 'TS2578' // Unused '@ts-expect-error' directive.

1. I'd go to [webpack.config.js](https://github.com/woocommerce/woocommerce/blob/1de7ef6be3e826cfc995594ab0bab4ef9777efc3/plugins/woocommerce/client/admin/webpack.config.js#L195)
1. Uncomment the line with TS2578
1. go to `./plugins/woocommerce/client/admin` and run `DISABLE_TYPESCRIPT_CHECKING=true HIDE_TYPESCRIPT_WARNINGS=true npx webpack`

1. Build should be unchanged and still work
3. `pnpm watch:build` in ./plugins/woocommerce/client/admin should not flood with typescript warnings
4. Above steps to work on specific TS errors should work

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/dev-typescript-env-var-disable
+++ b/plugins/woocommerce/changelog/dev-typescript-env-var-disable
@@ -1,0 +1,4 @@
+Significance: patch
+Type: dev
+
+Added env vars to allow for disabling typescipt checking and hiding warnings.
--- a/plugins/woocommerce/client/admin/package.json
+++ b/plugins/woocommerce/client/admin/package.json
@@ -302,6 +302,16 @@
 		},
 		"watch:build:project:bundle": {
 			"command": "webpack --watch",
+			"env": {
+				"DISABLE_TYPESCRIPT_CHECKING": {
+					"external": true,
+					"default": "true"
+				},
+				"HIDE_TYPESCRIPT_WARNINGS": {
+					"external": true,
+					"default": "false"
+				}
+			},
 			"service": true
 		},
 		"dependencyOutputs": {
--- a/plugins/woocommerce/client/admin/webpack.config.js
+++ b/plugins/woocommerce/client/admin/webpack.config.js
@@ -89,6 +89,10 @@ require( 'fs-extra' ).ensureSymlinkSync(
 
 const webpackConfig = {
 	mode: NODE_ENV,
+	ignoreWarnings:
+		process.env.HIDE_TYPESCRIPT_WARNINGS === 'true'
+			? [ { message: /TS\d{4,6}:\ / } ]
+			: [],
 	entry: getEntryPoints(),
 	output: {
 		filename: ( data ) => {
@@ -179,7 +183,9 @@ const webpackConfig = {
 	plugins: [
 		...styleConfig.plugins,
 		// Runs TypeScript type checker on a separate process.
-		! process.env.STORYBOOK && new ForkTsCheckerWebpackPlugin(),
+		! process.env.STORYBOOK &&
+			! ( process.env.DISABLE_TYPESCRIPT_CHECKING === 'true' ) &&
+			new ForkTsCheckerWebpackPlugin(),
 		! process.env.STORYBOOK &&
 			new TypeScriptWarnOnlyWebpackPlugin( [
 				// these are the errors that have been converted into warnings during the react-18 upgrade


ilyasfoo:
This is working as advertised.

On attempting to run TS check however, running `DISABLE_TYPESCRIPT_CHECKING=true;HIDE_TYPESCRIPT_WARNINGS=true npx webpack` gives me:

```
ERROR in main
Module not found: Error: Can't resolve './src' in '[redacted]/woocommerce/plugins/woocommerce'
resolve './src' in '[redacted]/woocommerce/plugins/woocommerce'
  using description file: [redacted]/woocommerce/plugins/woocommerce/package.json (relative path: .)
    Field 'browser' doesn't contain a valid alias configuration
    using description file: [redacted]/woocommerce/plugins/woocommerce/package.json (relative path: ./src)
      no extension
        Field 'browser' doesn't contain a valid alias configuration
        [redacted]/woocommerce/plugins/woocommerce/src is not a file
      .js
        Field 'browser' doesn't contain a valid alias configuration
        [redacted]/woocommerce/plugins/woocommerce/src.js doesn't exist
...
```

I tried running in both repo root and in `plugins/woocommerce`.

Running `pnpx webpack` bugs me with installing `webpack-cli`, which I already did (I'm sure something is missing in my setup).

Would it also be possible to run typescript checks on changed files? I think this would be the ideal intermediate solution where devs could see only relevant errors. If nothing is readily available, perhaps a custom Webpack plugin?

rjchow:
Hey @ilyasfoo thanks for the review! In my late night christmas eve haste I pasted the wrong command - to set multiple env vars for a single command you separate them with a ` ` (space), not a semicolon!

so the correct command for TS checking is: `DISABLE_TYPESCRIPT_CHECKING=false HIDE_TYPESCRIPT_WARNINGS=true npx webpack` and this is to be run in `./plugins/woocommerce/client/admin` . Sorry bout the confusion!

> Would it also be possible to run typescript checks on changed files? I think this would be the ideal intermediate solution where devs could see only relevant errors. If nothing is readily available, perhaps a custom Webpack plugin?

When I was working on this I decided it's not possible because accurate type checking only really works as a whole codebase thing - otherwise you would have to make most-likely invalid assumptions of the interfacing types (between the changed files and the existing files). 

It did however also occur to me that you could do a full type check, save it to file, and only output new errors. I'm not really well placed to develop this piece of work though - and it would still have the issue of relying on the wrong types (lets say there were some old errors on a function a(), then a new function b() relies on its' type - you'd then be developing new types based on the wrong function a() type).

ilyasfoo (approved):
> so the correct command for TS checking is: DISABLE_TYPESCRIPT_CHECKING=false HIDE_TYPESCRIPT_WARNINGS=true npx webpack and this is to be run in ./plugins/woocommerce/client/admin . Sorry bout the confusion!

Np! Thanks, this command works for me.

> When I was working on this I decided it's not possible because accurate type checking only really works as a whole codebase thing - otherwise you would have to make most-likely invalid assumptions of the interfacing types (between the changed files and the existing files).
> 
> It did however also occur to me that you could do a full type check, save it to file, and only output new errors. I'm not really well placed to develop this piece of work though - and it would still have the issue of relying on the wrong types (lets say there were some old errors on a function a(), then a new function b() relies on its' type - you'd then be developing new types based on the wrong function a() type).

Ah yes, that makes sense to me. Perhaps it's not realistic, at least not in the scope of this PR.

LGTM! 🚢

PR #54040, opened 2024-12-26
Track 'Launch Your Store' settings

menakas:

After the introduction of the 'Launch Your Store' feature, adding tracking for these settings.

1. Send a snapshot by calling `send_tracking_data()` 
2. Check if the values for the three options related to Lauch Your Store are received in Nosara.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Adding Launch Your Store settings to the Tracker
</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/includes/class-wc-tracker.php
+++ b/plugins/woocommerce/includes/class-wc-tracker.php
@@ -998,6 +998,9 @@ class WC_Tracker {
 			'hpos_transactions_level'               => get_option( 'woocommerce_db_transactions_isolation_level_for_custom_orders_table_data_sync' ),
 			'show_marketplace_suggestions'          => get_option( 'woocommerce_show_marketplace_suggestions' ),
 			'admin_install_timestamp'               => get_option( 'woocommerce_admin_install_timestamp' ),
+			'coming_soon'                           => get_option( 'woocommerce_coming_soon' ),
+			'store_pages_only'                      => get_option( 'woocommerce_store_pages_only' ),
+			'private_link'                          => get_option( 'woocommerce_private_link' ),
 		);
 	}
 
--- a/plugins/woocommerce/changelog/54040-woodata-tracker_lys
+++ b/plugins/woocommerce/changelog/54040-woodata-tracker_lys
@@ -1,0 +1,4 @@
+Significance: minor
+Type: dev
+
+Adding Launch Your Store settings to the Tracker


plugins/woocommerce/includes/class-wc-tracker.php
```diff
@@ -998,6 +998,9 @@ private static function get_all_woocommerce_options_values() {
 			'hpos_transactions_level'               => get_option( 'woocommerce_db_transactions_isolation_level_for_custom_orders_table_data_sync' ),
 			'show_marketplace_suggestions'          => get_option( 'woocommerce_show_marketplace_suggestions' ),
 			'admin_install_timestamp'               => get_option( 'woocommerce_admin_install_timestamp' ),
+			'coming_soon'                           => get_option( 'woocommerce_coming_soon' ),
```
EveChen, line 1001: I've followed this [Slack thread](https://a8c.slack.com/archives/C06D0919X0U/p1735045974240699) but I'm not so sure if "coming_soon" indicates it's a staging WooCommerce site? 🤔 
image

EveChen: Also, based on the screenshot above, there is a option "Live", which I believe is the "Launch Your Store" feature?

EveChen (commented):
LGTM though I only know little thing about PHP 😅. Still waiting for all checks to pass successfully. While waiting, I'm also curious about the definition of those loggings. e.g.
1. `coming_soon`: staging site or test site?
2. When a merchant hits "Live" button, that's "Launch Your Store"?

plugins/woocommerce/includes/class-wc-tracker.php
menakas: @EveChen , as far as I understand, the `Coming soon` option is used on the live site, but the site is still hidden while the owner is trying something out. 
`Coming soon` and `Live` modes are mutually exclusive (radio button) and sets the `woocommerce_coming_soon` options to `yes` or `no` respectively.

EveChen: Thanks for the explanation @menakas! One follow up question, do you know if we regard merchants who chose `Coming soon` option as "staging sites"? The reason why I'm asking is that Payments Pod is working on monetized GMV and we want to filter out "staging sites". If the above is true, then we can use this logging to filter out staging sites 😃.

EveChen (dismissed):
LGTM! Also, all the checks are passed successfully. Thank you for writing this PR, Menaka! 🚀

plugins/woocommerce/includes/class-wc-tracker.php
menakas: @EveChen yes, this PR is to address the same problem : to exclude test orders made in the `Coming Soon` mode. But they're not staging sites because this option can be toggled - y'day `Coming soon`, today `Live` and tomorrow `Coming soon` again. So, we need some order-specific data. 

IMHO, we shouldn't regard these sites as staging sites. But, staging sites would also have the same options and that's going to complicate things, I guess.

EveChen: Gotcha! Just like you and Alex mentioned, looks like this feature is not that reliable. But still good to know we have this toggle for merchants to use (though I guess some of merchants may also confused about this, just like me 😅 )

menakas:
Closing this because this dropped off my radar and it is 1.5 yrs + since I opened this. May not be relevant now.

Issue #52316, opened 2024-10-24
[Settings] Introduce DataForms to Settings


psealock:
Depends on https://github.com/woocommerce/woocommerce/issues/52315

Map the configured values by `type` to existing Gutenberg components using DataForms. Determine base types not supported by DataForms and make a note on this issue or create new issue(s) to either handle these cases upstream or in WooCommerce.

louwie17:
@psealock not sure if you already decided on what to power the form with and if you are planning on using DataForms.
But incase you are using DataForms, it already supports some base types that are converted to react components, like `text`, `integer`, and `datetime`: https://github.com/WordPress/gutenberg/blob/0ceac8f5e8b74cbeb7ef36c44009c67ec5fabbf7/packages/dataviews/src/field-types/index.tsx#L15

psealock:
After some consideration, I don't think `<DataForm />` is helpful. The settings configuration has a whole array of types to consider, amongst them is a custom type that handles raw HTML. The DataForm doesn't easily lend itself well to mixing elements. Some of the custom code will assume a regular HTML form, so handling the submit must take that into account. DataForm requires an internal state to track values.

psealock:
From @gigitux (paraphrased)

>DataView allows you to implement dedicated fields, if you need to render data differently. For instance, this is the field used to render [the feature image of a post](https://github.com/WordPress/gutenberg/pull/65196). You could create a dedicated field for each type, developers that want to update the Settings page will be able to re-use the existent field that we have created.

Implementing a dedicated field can fill in the gaps where DataForms are missing.

Further:

>I would like to highlight that dataviews will have an extensibility API ([this is a first PR regarding extensibility](https://github.com/WordPress/gutenberg/pull/67175)), enabling developers to register new fields and extend form fields. I think that it would be beneficial for Woo Settings too.

Either of these two avenues will be helpful to solve the issue of custom HTML, which will be generated by PHP and [passed to the client](https://github.com/woocommerce/woocommerce/issues/52317).


I will re-open this issue so we can investigate.

psealock:
> I will re-open this issue so we can investigate.

We will use DataForms for Modern Screens. Legacy screens will implement its own form using Gutenberg input components.

The main reason for this is plugins inject custom HTML from tables, inputs, all the way to entire apps using slotFill. We can't know what folks will do, so its not a good idea to wrap those unknowns inside of a DataForm.

[...]

PR #54041, opened 2024-12-26
[Settings] Use DataForms to render settings fields

chihsuan:

Related to https://github.com/woocommerce/woocommerce/issues/52316

This PR shows how we can use DataForms to render settings fields instead of custom code. 

To use DataForm, we need to ensure each setting field has an `id` property, so I updated the setting data transformer to add a random id if it's missing.

https://github.com/user-attachments/assets/c466cbc0-0b44-47af-a3e2-a244303137b7

---

DataForm is a component that allows us to render form fields with a consistent UI. It supports a variety of field types, including text, number, checkbox, and select. Additionally, it's possible to render custom fields by providing a custom component. 

For more details, see:

- [DataForm README](https://github.com/WordPress/gutenberg/blob/227fcd87f1eccf4b853850c8f439fd95ba1e5f5f/packages/dataviews/README.md#dataform)
- [DataForm storybook](https://wordpress.github.io/gutenberg/?path=/docs/dataviews-dataform--docs&args=type:regular)

`@wordpress/dataviews@4.10.0` has a bug in package.json that causes the tests to fail. It has been fixed in https://github.com/WordPress/gutenberg/pull/67962 but not yet released so I've created a patch to fix it for now. Once it's released, we can remove the patch.

3. Use the [Settings Tester](https://github.com/woocommerce/woocommerce-settings-migration-tester) to test Modern screens. Clone it locally and `npm install && npm run plugin-zip` / [settings-tester.zip](https://github.com/user-attachments/files/18208344/settings-tester.zip)
3. Upload and activate the Tester
4. Go to `WooCommerce > Settings` and poke around and see checkboxes, selects, and inputs rendered with values.
5. Go to `WooCommerce > Settings > Emails` and see a custom HTML table rendered.
6. Next go to Site visibility to confirm our slotFill root is available by querying the document in the console.

```js
document.getElementById('wc_settings_site_visibility_slotfill');
```

7. Go to `/wp-admin/admin.php?page=wc-settings&tab=settings-tester&section=javascript`
8. Toggle `Echoing HTML` checkbox and confirm the yellow background is shown when the checkbox is checked.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 
Use DataForms to render settings fields 

</details>


packages/js/settings-editor/src/utils/dataforms-transformers.tsx
```diff
@@ -0,0 +1,227 @@
...
+				id: setting.id,
+				type: 'text',
+				label: setting.desc + ' (TO BE IMPLEMENTED)',
+				elements: Object.entries( setting.options || {} ).map(
+					( [ label, value ] ) => ( {
+						label,
+						value,
+					} )
+				),
+				Edit: 'select',
```
psealock: Maybe its a problem upstream but I'm seeing this error:
```
react-dom.js?ver=18:73 Warning: The `value` prop supplied to <select> must be a scalar value if `multiple` is false.
```

psealock: Also an upstream problem, but this select does not apply the id to the element, instead using something like `inspector-select-control-1`. We could use `SelectControl` in a similar way that you've used `InputControl` in this PR. 

It will be important for input elements to have the `id` applied because the legacy forms follow this pattern. Some extensions assume this is the case for things like form manipulation or even creating their own submit button.

Alternatively we could make this contribution upstream.

packages/js/settings-editor/src/utils/dataforms-transformers.tsx
```diff
@@ -0,0 +1,227 @@
...
+ * @param settings Array of settings to process
+ * @return         Form layout configuration
+ */
+export const generateForm = ( settings: SettingsField[] ) => ( {
+	type: 'regular' as const,
+	labelPosition: 'top' as const,
+	fields: settings
+		.map( transformToFormField )
+		.filter( ( field ) => field !== false ),
+} );
```
psealock: We'll need to add some js unit tests for these functions 😄

psealock (commented):
This is really awesome @chihsuan !

This is working well, I just noted issues with ids and select.

>DataForm is a component that allows us to render form fields with a consistent UI. It supports a variety of field types, including text, number, checkbox, and selects

The consistency in UI is great, but given that we will be passing in `InputControl`, possibly `SelectControl` and a few others do you think the added complexity is worthwhile? One thought I have is this depends on what changes we can get merged upstream. For instance, if can get `InputControl` rendered by dataviews instead of what it does now, this will support several more text types (email, password, etc) out of the box and we can remove the `getInputEdit` you've added here. This example seems like a no-brainer, but I was curious if you have any further thoughts on this?

package.json
```diff
@@ -77,7 +77,8 @@
 			"react-resize-aware": "3.1.1"
 		},
 		"patchedDependencies": {
-			"@wordpress/edit-site@5.15.0": "bin/patches/@wordpress__edit-site@5.15.0.patch"
+			"@wordpress/edit-site@5.15.0": "bin/patches/@wordpress__edit-site@5.15.0.patch",
+			"@wordpress/dataviews@4.10.0": "patches/@wordpress__dataviews@4.10.0.patch"
```
kalessil: It looks like the new patch location needs to be updated. We keep patches under `bin/patches/` directory.

kalessil (changes requested):

package.json
chihsuan: Ah, that's right. Thanks for catching this! I just noticed that `@wordpress/dataviews@4.11.1` has been released, which includes the fix for the issue so I'll update the dependency and remove the patch.

chihsuan:
> The consistency in UI is great, but given that we will be passing in InputControl, possibly SelectControl and a few others do you think the added complexity is worthwhile?
>This example seems like a no-brainer, but I was curious if you have any further thoughts on this?


@psealock Yeah, as discussed in the meetup, I think we could contribute upstream so that we don't need to implement our custom InputControl, SelectControl, etc (though we're pausing on contributing upstream now 😅 ).

packages/js/settings-editor/src/utils/dataforms-transformers.tsx
chihsuan: Good catch! I'll add a SelectControl.

chihsuan: Added in 7bd2444ef4968e6db4458ac587d0acfbb3df6f1f.

package.json
chihsuan: Updated in fc4e1498112a92a76e51d3faf65e301ed2f011de.

chihsuan:
@psealock  I've addressed the issues you mentioned. I also added a `useSettingsForm` hook to make it easier to understand. Could you please review it again? 🙏  Thanks!

package.json
kalessil: Super, thanks!

kalessil (approved):
LGTM (I reviewed only patch/deps bump part)!

louwie17:
Just a drive by comment here :)
Nice work @chihsuan, and happy to see you got this also working with DataForm's. I see in some previous comments you mentioned around making changes upstream. Given the pause, it may be good in the mean time to surface some of your feedback either by creating an issue in Gutenberg, sharing some feedback in [this Slack channel](https://wordpress.slack.com/archives/C082S8Q9PL3), or on P2 ( internally ). 

Btw I saw the big switch statement, we currently have something similar within the DataForm package: https://github.com/WordPress/gutenberg/blob/0ceac8f5e8b74cbeb7ef36c44009c67ec5fabbf7/packages/dataviews/src/field-types/index.tsx#L15. The plan is to elaborate on this, for example your `getInputEdit` solution may be a great addition to this. We also want to create some functions around this to be able to register field types, so you don't have to explicitly these definitions through the `fields` list.

Anyhow glad to see you make it work and the more use cases the better :)

psealock (approved):
This is looking great, code looks good and tests well 🚢

moon0326 (approved):
LGTM and tested well 👍 🚀

psealock:
One thing I just noticed is when a text field has no `desc` then the label is just the id. For `text` types we should also respect `placeholder` values and figure out what to do with `desc_tip`. We could address this in a follow up or just create an issue for the backlog.

image

chihsuan:
> One thing I just noticed is when a text field has no `desc` then the label is just the id. For `text` types we should also respect `placeholder` values and figure out what to do with `desc_tip`. We could address this in a follow up or just create an issue for the backlog.

Thanks, I'll create an issue for this.

PR #54064, opened 2024-12-27
Fix logic for skipping found rows query in OrdersTableQuery.

### LINKED ISSUE #53065: OrdersTableQuery.php is still running count queries even when `no_found_rows == true`
### Prerequisites

### Describe the bug

There is an error in the logic on this line:

https://github.com/woocommerce/woocommerce/blob/4b0913a066cd89f2c22919d49bb563f2bed2784b/plugins/woocommerce/src/Internal/DataStores/Orders/OrdersTableQuery.php#L1352

To avoid running the found rows query when `no_found_rows` is truthy, It should be:
```
if ( ( $this->arg_isset( 'no_found_rows' ) && $this->args['no_found_rows'] ) || empty( $this->orders ) ) {
```

This is causing calls to `wc_get_orders()` to run extra queries even when the count isn't needed.

### Expected behavior

When  `no_found_rows` is truthy, the count query should be avoided.

### Actual behavior

When  `no_found_rows` is truthy, the first half of the statement `( $this->arg_isset( 'no_found_rows' ) && $this->args['no_found_rows'] )` evaluates to false, causing the fall through to continue running the count query.

### Steps to reproduce

1. Create a script to run wc_get_orders() or run a WC_Order_Query() without a `paginate` query var set to have `no_found_rows` get set to `true` by default.
2. Use some sort of query monitor to evaluate the queries run.

### WordPress Environment

N/A

### Isolating the problem

prettyboymp:

Closes #53065.

The check to skip the found rows query was checking against `! $this->args['no_found_rows']`, however, we want to skip when `no_found_rows` is truthy.  This removes the negation.

1. The simplest way to test is with a test site running WooCommerce Subscriptions setup with the ability to monitor DB queries.
2. Visit the admin order listing page. 
3. Check the queries run for the page.  Check that queries run to get related subscriptions for orders do not include a separate count query after.

Alternatively, through a script or active shell with query monitoring available:
1. Call `wc_get_orders()`
2. Verify that no additional count query is run when retrieving the results.

### DIFF
--- a/plugins/woocommerce/changelog/fix-53065-skip-count-for-no-found-rows
+++ b/plugins/woocommerce/changelog/fix-53065-skip-count-for-no-found-rows
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix logic for skipping found rows query in OrdersTableQuery
--- a/plugins/woocommerce/src/Internal/DataStores/Orders/OrdersTableQuery.php
+++ b/plugins/woocommerce/src/Internal/DataStores/Orders/OrdersTableQuery.php
@@ -1351,7 +1351,7 @@ class OrdersTableQuery {
 		$this->orders = array_map( 'absint', $wpdb->get_col( $this->sql ) ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
 
 		// Set max_num_pages and found_orders if necessary.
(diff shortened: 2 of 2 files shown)


joshuatf (approved):
Good catch!  Testing well and LGTM 🚢

PR #54074, opened 2024-12-30
Add to Cart with Options: use correct esc function in Stepper layout

Aljullu:

This PR replaces an instance of `esc_html()` with `esc_attr()`, given that `aria-label` is an attribute, it seems like the correct approach.

2. Go to `WooCommerce Admin Test Helper` -> `Features` and enable `add-to-cart-with-options-stepper-layout`.
3. Create a post or page and add the Single Product block. Select any of the products.
4. Select the _Add to Cart with Options_ block and change the layout style to _Stepper_.
5. In the frontend, use the browser devtools (<kbd>F12</kbd> -> _Inspector_) to verify the Quantity Selector buttons have the correct `aria-label` attribute.

### DIFF
--- a/plugins/woocommerce/changelog/fix-add-to-cart-with-options-esc-function
+++ b/plugins/woocommerce/changelog/fix-add-to-cart-with-options-esc-function
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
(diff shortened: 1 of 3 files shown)


gigitux (approved):
LGTM 🚢

PR #54087, opened 2024-12-30
[Experimental] Hide Add to Cart with Options (Experimental) block from widgets screen

Aljullu:

This PR updates the function used to register the _Add to Cart with Options_ block from `registerBlockType()` to `registerProductBlockType()`, which takes care of not displaying the _Add to Cart with Options_ block in unwanted contexts like the widgets screen.

I also took the opportunity to fix a TypeScript error affecting the old _Add to Cart Form_ block.

2. Go to `WooCommerce Admin Test Helper` -> `Features` and enable `experimental-blocks` and `blockified-add-to-cart`.
3. Create a post or page and verify you can't add the _Add to Cart with Options (Experimental)_ block directly into the post/page.
4. Now, add the Single Product block and select a product.
5. Verify you can add the _Add to Cart with Options (Experimental)_ block inside the Single Product block.
6. Go to Appearance > Editor > Templates > Single Product.
7. Verify you can add the _Add to Cart with Options (Experimental)_ block anywhere in the template.
8. Switch to a classic theme and go to Appearance > Widgets.
9. Verify you can't add the _Add to Cart with Options (Experimental)_ block in the widget areas.

### DIFF
--- a/plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/index.tsx
+++ b/plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/index.tsx
@@ -18,6 +18,7 @@ import metadata from './block.json';
 import AddToCartOptionsEdit from './edit';
 import save from './save';
 import './style.scss';
+import type { Attributes } from './types';
 
 // Pick the value of the "blockify add to cart flag"
 const isBlockifiedAddToCart = getSettingWithCoercion(
@@ -25,11 +26,6 @@ const isBlockifiedAddToCart = getSettingWithCoercion(
 	false,
 	isBoolean
 );
-
-export interface Attributes {
-	className?: string;
-	isDescendentOfSingleProductBlock: boolean;
-}
 
 export const shouldBlockifiedAddToCartWithOptionsBeRegistered =
 	isExperimentalBlocksEnabled() && isBlockifiedAddToCart;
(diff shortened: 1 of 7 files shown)


plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/index.tsx
```diff
@@ -25,6 +26,11 @@ const isBlockifiedAddToCart = getSettingWithCoercion(
 	isBoolean
 );
 
+export interface Attributes {
+	className?: string;
+	isDescendentOfSingleProductBlock: boolean;
+}
+
```
kmanijak: Nitpick: what about moving this to `plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/types.ts`? The file is already there so seems like this interface would belong there 🙌

kmanijak (approved):
Works as expected! And thanks for adding better typing! <3

I left minor comment but approving anyway!

plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/index.tsx
Aljullu: Oh right, good catch! Done in f128d8af177ecc75d857eb168f10608540e07e13.

PR #54117, opened 2025-01-02
Fix registerProductBlockType() types

Aljullu:

This PR adds some TypeScript fixes on top of the changes made in https://github.com/woocommerce/woocommerce/pull/54087/.

This PR only includes changes to types, so there are no user-facing changes.

1. Open `plugins/woocommerce-blocks/assets/js/atomic/utils/register-product-block-type.ts` and `plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/index.tsx`.
2. Verify there are no TypeScript errors introduced by this PR.

### DIFF
--- a/plugins/woocommerce-blocks/assets/js/atomic/utils/register-product-block-type.ts
+++ b/plugins/woocommerce-blocks/assets/js/atomic/utils/register-product-block-type.ts
@@ -66,7 +66,7 @@ export class BlockRegistrationManager {
 		string,
 		ProductBlockSettings & {
 			blockName: string;
-			settings: Record< string, unknown >;
+			settings: Partial< BlockConfiguration< BlockAttributes > >;
 		}
 	> = new Map();
 	/** Current template ID being edited */
@@ -149,9 +149,7 @@ export class BlockRegistrationManager {
 				subscribe( () => {
 					const previousTemplateId = this.currentTemplateId;
 					this.currentTemplateId = this.parseTemplateId(
-						editSiteStore.getEditedPostId<
-							string | number | undefined
-						>()
+						editSiteStore.getEditedPostId()
 					);
 
 					if ( previousTemplateId !== this.currentTemplateId ) {
@@ -319,7 +317,7 @@ export class BlockRegistrationManager {
 		config: ProductBlockConfig< T >
 	): void {
 		const key = config.variationName || config.blockName;
-		this.blocks.set( key, config );
+		this.blocks.set( key, config as ProductBlockConfig< BlockAttributes > );
 		this.registerBlock( config );
 	}
 }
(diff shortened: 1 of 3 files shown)


plugins/woocommerce-blocks/assets/js/atomic/utils/register-product-block-type.ts
```diff
@@ -60,7 +62,13 @@ export class BlockRegistrationManager {
 	/** Singleton instance of the manager */
 	private static instance: BlockRegistrationManager;
 	/** Map storing block configurations keyed by block name or variation name */
-	private blocks: Map< string, ProductBlockConfig > = new Map();
+	private blocks: Map<
+		string,
+		ProductBlockSettings & {
+			blockName: string;
+			settings: Record< string, unknown >;
+		}
```
Aljullu: I didn't know how to keep this specific to `ProductBlockConfig` as it now depends on a generic that we don't know when building `BlockRegistrationManager`. A bit unfortunate, but given that all methods used to interact with this property use specific types, I think it shouldn't be a big deal. However, I'm open to suggestions on how to improve this. :slightly_smiling_face:

tjcafferkey:
I'm getting the TS error `Untyped function calls may not accept type arguments.ts` for the following bit of code

```
this.currentTemplateId = this.parseTemplateId(
						editSiteStore.getEditedPostId<
							string | number | undefined
						>()
					);
```

I think we can update it to be

```
this.currentTemplateId = this.parseTemplateId(
						editSiteStore.getEditedPostId()
					);
```

tjcafferkey:
Thanks for working on this @Aljullu! I was thinking something along the lines of this fix https://github.com/woocommerce/woocommerce/pull/54120

Just curious how you think it compares to your changes?

Aljullu:
> I'm getting the TS error `Untyped function calls may not accept type arguments.ts` for the following bit of code

Yeah, sorry. I thought that was an unrelated TS error so I ignored it. :sweat_smile:  But your suggestion makes sense!

> Thanks for working on this @Aljullu! I was thinking something along the lines of this fix https://github.com/woocommerce/woocommerce/pull/54120
> 
> Just curious how you think it compares to your changes?

I'm not sure, to be honest. By doing `Partial< BlockConfiguration >` we lose the attributes types, no? Wouldn't that make types a bit less specific? :thinking: 

Having said that, I don't have a strong opinion, and I'm fine going with your solution. It looks simpler and, to be honest, I'm not 100% confident about mine.

tjcafferkey:
>I'm not sure, to be honest. By doing Partial< BlockConfiguration > we lose the attributes types, no? Wouldn't that make types a bit less specific? 🤔

It makes `attributes` _optional_, but if `attributes` is present then its appropriately typed as far as I understand it.

tjcafferkey (approved):
Thank you for this @Aljullu ! And thanks for our DM convo around it too.

Aljullu (commented):
Thanks for the additional commits, the review and all the research associated, @tjcafferkey!

Everything looks good to me too and confirming there are no more TS errors in those files. :100:

PR #54099, opened 2024-12-31
Fix Stock Indicator not loading when "experimental-blocks" is disabled

Aljullu:
### Changes proposed in this Pull Request:

Fixes https://github.com/woocommerce/woocommerce/issues/54068.

This PR makes it so the product type store is always registered. Until now, we were only registering it when the `experimental-blocks` feature flag was enabled, but the store is also used by the _Stock Indicator_ block, which is not an experimental one.

This PR also adds some minor TypeScript fixes (it might be easier to review the two commits separately).

### How to test the changes in this Pull Request:

2. Go to Pages > Add New Page
3. Add a `Single Product` block and select the product you just created.
4. Add a `Stock indicator` block.
5. Go to Tools > WCA Test Helper > Features
6. Disable the feature: `experimental-blocks`
6. Go back to the page you created and edit it.
7. Verify the `Stock indicator` block is still visible without errors.

PR #54118, opened 2025-01-02
[Experimental] Move Product Type template state store to its own folder and rename some constants

Aljullu:

As @kmanijak mentioned in https://github.com/woocommerce/woocommerce/pull/54113#pullrequestreview-2527387951, the store used by the Product Type template state was using names too generic. In https://github.com/woocommerce/woocommerce/pull/54099 we might remove the `registerStore()` function. In this one I renamed the `store` export and moved the store from `plugins/woocommerce-blocks/assets/js/shared/store/` to `plugins/woocommerce-blocks/assets/js/shared/stores/product-type-template-state/`.

This PR doesn't introduce any user-facing changes, so testing means making sure there are no regressions. Because of that, the testing instructions are the same as in #54113.

2. Go to `WooCommerce Admin Test Helper` -> `Features` and enable `add-to-cart-with-options-stepper-layout` and `blockified-add-to-cart`.
3. Smoke test the `Add to Cart with Options (Experimental)` block.
4. Create a new product.
5. Go to Pages > Add New Page.
6. Add a `Single product` block and select the product you just created.
7. Remove the current _Add to Cart with Options_ block.
8. Everything should work well
9. Now go to the Single Product template and add an `Add to Cart with Options (Experimental)` block.
10. Verify that after adding the block the `Type switcher` selector is now visible in the Inspector > Template.
11. Confirm the selected product type is consistent.
12. Confirm you do not see the console errors shown in the above screenshot when you're in the Gutenberg editor on the Single Product template, or a post/page.
13. Confirm all E2E tests pass.

https://github.com/user-attachments/assets/0c56abdb-fc03-416c-b893-b4cca89049ca


plugins/woocommerce-blocks/assets/js/shared/stores/product-type-template-state/index.ts
```diff
@@ -130,12 +130,12 @@ const reducer = ( state: StoreState = DEFAULT_STATE, action: Actions ) => {
 	}
 };
 
-export const store = createReduxStore( STORE_NAME, {
```
retrofox: All stores registered by core use the `store` constant name. For instance [here](https://github.com/WordPress/gutenberg/blob/a2daa2fc0874e6af82d0607c65b324d873c9e4ff/packages/preferences/src/store/index.js#L21).

We can use anything we choose, but it was intentional to follow the core coding convention. Just FYI.

Indeed you saw how it imports the stores, for instance:

```es6
import { store as noticesStore } from '@wordpress/notices';
```

...which is another convention :-)

Aljullu: Aaah, good point. I don't have a strong opinion. I think in WooCommerce we are a bit more liberal with that. In fact, I saw several different patterns in our codebase. :sweat: I'm fine keeping this aligned with core, so I will undo these changes.

Aljullu: Done in 3fe15e8a1b82de67e470235be04e0805265660e2. Thanks for the feedback, @retrofox!

retrofox (approved):
LGTM 🚀 thanks for enhancing the code 💯

PR #54121, opened 2025-01-02
add log context normalization (#53595)

### LINKED ISSUE #53595: [Enhancement]: Serialize exceptions in log context

Similarly to [monolog/monolog](https://github.com/Seldaek/monolog/blob/main/src/Monolog/Formatter/NormalizerFormatter.php#L175), WooCommerce could serialize all content passed as context to logger. This would enhance the developer experience when passing `Stringable` object or exceptions.

Current behavior:

```php
<?php

try {
  throw new \Exception('Something bad happened...');
} catch (Throwable $e) {
  wc_get_logger->debug('oh, no exception', ['exception_in_context' => $e]);
}
// DEBUG 'oh, no exception' CONTEXT: {"exception_in_context": {}}
```

This forces to manually serialize exceptions and any objects, which could be easily formatted as strings. FWIW, `JsonSerializable` implementors are good, as context is running through `json_encode` function, nevertheless at little expense this could be enhanced further.

If you generally agree upon the above, I'd be happy to provide a PR.

It _might_ be possible to filter this with `woocommerce_format_log_entry`, but including this in various plugins would add unnecessary overhead on customer sites contrary to centralized solution, because using the filter would require to build whole log message from scratch.

### Additional context

_No response_

bart-jaskulski:

Closes #53595 .

1. Add log statement somewhere in code with `wc_get_logger->debug('Message', ['error' => new \Exception()]);`
2. Go to WooCommerce -> Status -> Logs
3. Log message will contain rich context about exception

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

Improve context serialization in log statements.

</details>

<details>

#### Comment 

</details>


bart-jaskulski:
With this PR, more complex data structures passed as log context are normalized before saving to file. Previously it only serialized to json string, but now the normalization step is added.

Example change is for the exceptions. Previously, when exception was added in context log produced following line:

```php
try {
} catch (\Exception $e) {
  wc_get_logger()->debug('Oh, no...', ['error' => $e]);
}
// DEBUG 'Oh, no...' CONTEXT: {"error": {}}
```

The suggested change enhances the format, making it more usable. The normalization is borrowed from monolog/monolog

```php
try {
} catch (\Exception $e) {
  wc_get_logger()->debug('Oh, no...', ['error' => $e]);
}
// DEBUG 'Oh, no...' CONTEXT: {"error": {"class": "\Exception", "message": "something broke", "code": 0, "file": "wp-content/plugins/my-plugin/my-file.php:12", "trace": []}}
```

bart-jaskulski:
Added `[WIP]` in title as this needs test cases in code (atm struggling with test suite setup on local machine)

albarin:
Hey @bart-jaskulski! Thanks a lot for your contribution!
Is the PR ready to be reviewed? We are still seeing some commented-out code.

bart-jaskulski:
Hi @albarin
Yes, this PR is ready. About commented test cases: I'd be willing to accept some suggestions, how the test could be improved to actually validate correct exception normalization, while maintaining current test's style.

I think, those three cases could be moved to own test method, separate from the provided, but I am not sure if this approach would be better than tweaking the current test method for less verbose handling of context assertions. In short: treat the commented out code as a question about the best standard in the project to handle that cases.

albarin:
@bart-jaskulski thanks for your reply.

Thinking about it, what about adding:
```php
```

albarin:
@bart-jaskulski thanks for your reply!

Thinking about it, what about adding something like this:
```php
class ExceptionWithTrace extends \Exception implements \JsonSerializable {
	private static $mockFile = '/mock/path/to/test.php';
	private static $mockLine = 100;

	public function jsonSerialize(): array {
		return array(
			'class'   => 'Exception',
			'message' => $this->getMessage(),
			'code'    => $this->getCode(),
			'file'    => self::$mockFile . ':' . self::$mockLine,
			'trace'   => array(
				self::$mockFile . ':' . self::$mockLine,
			),
		);
	}
}
```

and then using it in the test:
```php
yield 'exception' => array(
	array( 'ctx' => new ExceptionWithTrace( 'foo', 123 ) ),
	$context_delineator . wp_json_encode(
		array(
			'ctx' => array(
				'class'   => 'Exception',
				'message' => 'foo',
				'code'    => 123,
				'file'    => '/mock/path/to/test.php:100',
				'trace'   => array(
					'/mock/path/to/test.php:100',
				),
			),
		),
	JSON_UNESCAPED_SLASHES
	),
);
```

bart-jaskulski:
Actually, taking a deep breath and looking at my PR again, I think formatting should be encapsulated to another class. There we could test against data structures, not the string output, what should be easier to mock in specific cases.

Initially I was thinking it would be more cumbersone with adding new dependency in class graph to `LogHandlerFileV2`, but this shouldn't be a case with calling `wc_get_container` in `LogHandlerFileV2::__constructor`

I'll rewrite my PR accordingly.

albarin:
Closing this as there's been no activity since January.

PR #54129, opened 2025-01-02
Fix TS errors in `plugins/woocommerce/client/admin/client/products/`

### LINKED ISSUE #54123: Fix TS errors `plugins/woocommerce/client/admin/client/products/`
Part of https://github.com/woocommerce/woocommerce/issues/53939

kmanijak:

Fix or workaround TS errors in `plugins/woocommerce/client/admin/client/products/`

Closes https://github.com/woocommerce/woocommerce/issues/54123

To be tested by developers only!

1. Change in [webpack config](https://github.com/woocommerce/woocommerce/blob/d856cb9f8485a5483f4ac12222c58be40efd224a/plugins/woocommerce/client/admin/package.json#L301C7-L314C5):

```
"DISABLE_TYPESCRIPT_CHECKING": {
					"external": false,
					"default": "false"
				},
				"HIDE_TYPESCRIPT_WARNINGS": {
					"external": false,
					"default": "false"
				}
```

2. Run `pnpm build > ./ts-output.txt`
3. Check the newly created file `ts-output.txt` and search for `./client/products/`
4. Expected: There should be no occurrences 

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/admin/client/products/fills/more-menu-items/classic-editor-menu-item.tsx
```diff
@@ -30,8 +30,12 @@ export const ClassicEditorMenuItem = ( {
 			select( OPTIONS_STORE_NAME );
 
 		const allowTrackingOption =
+			// Todo: awaiting a more global fix, demo:
+			// https://github.com/woocommerce/woocommerce/pull/54146
 			getOption( ALLOW_TRACKING_OPTION_NAME ) || 'no';
```
Aljullu: Do you have some more context how this relates to #54146? I'm not sure to understand it. :sweat_smile:

plugins/woocommerce/client/admin/client/products/product-page.tsx
```diff
@@ -71,9 +71,11 @@ export default function ProductPage() {
...
 					<>
 						<WooProductMoreMenuItem>
-							{ ( { onClose }: { onClose: () => void } ) => (
-								<MoreMenuFill onClose={ onClose } />
-							) }
+							{
+								( ( { onClose }: { onClose: () => void } ) => (
+									<MoreMenuFill onClose={ onClose } />
+								) ) as unknown as React.ReactNode
+							}
```
Aljullu: I know it wasn't introduced in this PR, but I'm a bit confused by this pattern. :thinking: Why are we defining a function that takes the `onClose` prop simply to pass it down to a component? And the usage of  `as unknown` seems not ideal IMO.

I wonder if this could be refactored to simply:

```suggestion
							<MoreMenuFill />
```

And then make the `onClose` prop in `MoreMenuFill` optional. :thinking:

plugins/woocommerce/client/admin/client/products/product-variation-page.tsx
```diff
@@ -41,15 +41,19 @@ export default function ProductPage() {
...
+									<DeleteVariationMenuItem
+										onClose={ onClose }
+									/>
+									<MoreMenuFill
+										productType="product_variation"
+										onClose={ onClose }
+									/>
+								</>
+							) ) as unknown as React.ReactNode
+						}
```
Aljullu: Same. :slightly_smiling_face:

Aljullu (commented):
Thanks for working on this, @kmanijak! Overall looking good, I just left a couple of questions for things I'm not sure to understand.

plugins/woocommerce/client/admin/client/products/fills/more-menu-items/classic-editor-menu-item.tsx
kmanijak: Ah yeah, it's about changing the way we register stores from using string to store object. Then, the types are properly propagated. @rjchow is working on that and we agreed to have it fixed globally rather then each area separately.

More context: p1735902356485879-slack-C02FL3X7KR6

plugins/woocommerce/client/admin/client/products/product-page.tsx
kmanijak: This is a Slot Fill pattern used in multiple places in Woo. I'm not convinced to that either.

> I wonder if this could be refactored to simply:

The internal logic behaves differently when you provide React component vs function returning component. So it should not be the way to go.

> And the usage of as unknown seems not ideal IMO.

But I agree with that 😅 let me try to adjust the internal types to accommodate the case with function returning a component as such.

plugins/woocommerce/client/admin/client/products/fills/more-menu-items/classic-editor-menu-item.tsx
Aljullu: Oooh, ok. I wasn't aware of that combo, thanks!

Btw, sharing with you this thread from another PR where we are disabling these errors: https://github.com/woocommerce/woocommerce/pull/54243#discussion_r1905538395, in case we want to do the same in this PR.

kmanijak: Thank you for pointing it out! I updated those comments!

Change: https://github.com/woocommerce/woocommerce/pull/54129/commits/be080bca7251ec2e22f52785620bcbc3507c2d4f

plugins/woocommerce/client/admin/client/products/product-page.tsx
kmanijak: I spent some time investigating it but with no luck. I even consulted this with @gigitux and we cannot figure it out. I added `ts-expect-error` for now in those two cases and if we're ok with that, I'll create a dedicated issue for these two. I know it's not ideal.

Change: https://github.com/woocommerce/woocommerce/pull/54129/commits/70f6ba9293ba4ac8e4a65521e2263e852cc9ad8c

louwie17: > Why are we defining a function that takes the onClose prop simply to pass it down to a component? And the usage of as unknown seems not ideal IMO.

Just for some extra context around this. The `WooProductMoreMenuItem` is a wrapper around the Slot/Fill components available within GB. A `Slot` component can take `fillProps`, which can be anything, that is passed down into each `Fill` component. In order to pass it down and adhering to type checking the additional function wrapper is necessary. See the `fillProps` example in the docs here: https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/slot-fill/README.md#props

Technically we don't even know if `onClose` exists or not, if we wanted to we could enforce the `WooProductMoreMenuItem.Slot` type to take `onClose` as its `fillProps` and in turn also update the `WooProductMoreMenuItem` fill to guarantee that `onClose` exists ( from a type perspective ). But for now I pushed these changes: https://github.com/woocommerce/woocommerce/pull/54312/commits/d32cc4f22dac5020abd92765a0573253d35917ca which makes use of the GB `FillProps` type, which can be anything: https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/slot-fill/types.ts#L17

kmanijak: @louwie17, thank you ❤️ This one and the other similar is fixed in https://github.com/woocommerce/woocommerce/pull/54129/commits/bcd77ab3f81f79d319206f930231bfa7525288f1

plugins/woocommerce/client/admin/client/products/product-variation-page.tsx
kmanijak: Fixed in https://github.com/woocommerce/woocommerce/commit/bcd77ab3f81f79d319206f930231bfa7525288f1

plugins/woocommerce/client/admin/client/products/product-page.tsx
Aljullu: Thanks for the explanation, @louwie17! It makes sense.

Aljullu (approved):
Thanks for the updates and the additional context. LGTM!

PR #54133, opened 2025-01-02
Customize Your Store: fix TS errors

### LINKED ISSUE #54125: Fix TS errors `plugins/woocommerce/client/admin/client/customize-store`
Part of https://github.com/woocommerce/woocommerce/issues/53939

gigitux:

Closes https://github.com/woocommerce/woocommerce/issues/54125 .

1. Visit `wp-admin/admin.php?page=wc-admin&path=%2Fcustomize-store`
2. Click on "Start designing"
3. Do smoke test CYS

#### Only dev 

1. Go to `plugins/woocommerce/client/admin`
2. Comment [these lines](https://github.com/woocommerce/woocommerce/blob/d856cb9f8485a5483f4ac12222c58be40efd224a/plugins/woocommerce/client/admin/webpack.config.js#L189-L215)
3. Run `npx tsc | grep "customize-store"`
4. Ensure that no TS error appear.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Customize Your Store: fix TS errors.

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/admin/client/customize-store/assembler-hub/auto-block-preview.tsx
```diff
@@ -165,8 +181,7 @@ function ScaledBlockPreview( {
 		: 0;
 
 	// Initialize on render instead of module top level, to avoid circular dependency issues.
-	MemoizedBlockList = MemoizedBlockList || pure( BlockList );
```
gigitux, line 168: `pure` is a deprecated method

plugins/woocommerce/client/admin/client/customize-store/transitional/index.tsx
```diff
@@ -74,9 +74,6 @@ export const Transitional = ( {
 				</Modal>
 			) }
 			<SiteHub
-				variants={ {
```
gigitux, line 77: This prop isn't used

plugins/woocommerce/client/admin/client/customize-store/intro/index.tsx
```diff
@@ -377,9 +377,6 @@ export const Intro: CustomizeStoreComponent = ( { sendEvent, context } ) => {
 			) }
 			<div className="woocommerce-customize-store-header">
 				<SiteHub
-					variants={ {
-						view: { x: 0 },
```
gigitux, line 381: This prop isn't used

plugins/woocommerce/client/admin/client/customize-store/assembler-hub/sidebar/sidebar-navigation-screen.tsx
```diff
@@ -117,7 +117,7 @@ export const SidebarNavigationScreen = ( {
 						}
 						color={ '#e0e0e0' /* $gray-200 */ }
 						level={ 1 }
-						size={ 20 }
```
gigitux, line 120: `size` isn't a valid prop

plugins/woocommerce/client/admin/client/customize-store/assembler-hub/block-editor-container.tsx
```diff
@@ -186,7 +186,7 @@ export const BlockEditorContainer = () => {
 		<BlockEditor
 			renderedBlocks={ blocks }
 			isScrollable={ isScrollable }
-			onChange={ onChange }
+			onChange={ ( blockInstance ) => onChange( blockInstance, {} ) }
```
louwie17: This may actually break things right, I went down the rabbit hole a bit and it looks like the `onChange` would get called with the `blockInstance` and `selection` data: https://github.com/WordPress/gutenberg/blob/609a3c8d38a0608b41b1827ec4a8c1f812487ef2/packages/block-editor/src/components/provider/use-block-sync.js#L273-L283

The above change would also remove the the `selection` data. But it also means the types on our end are wrong for the `BlockEditor` onChange method.

plugins/woocommerce/client/admin/client/customize-store/assembler-hub/block-editor.tsx
```diff
@@ -21,12 +21,13 @@ export const BlockEditor = memo(
 		onChange,
 	}: {
 		renderedBlocks: BlockInstance[];
-		settings: Record< string, unknown > & {
+		settings: EditorSettings & {
 			styles: string[];
+			[ key: string ]: unknown;
```
louwie17, line 26: Hmm looks like this was needed? If so, its all good!
It just always feels weird to me adding a generic item like this in addition to distinct typed values.

plugins/woocommerce/client/admin/client/customize-store/assembler-hub/toolbar/shuffle.tsx
```diff
@@ -79,7 +80,9 @@ export default function Shuffle( { clientId }: { clientId: string } ) {
 		patternName: string;
 	} = useSelect(
 		( select ) => {
-			const { getBlockAttributes } = select( blockEditorStore );
+			const { getBlockAttributes } = select( blockEditorStore ) as {
+				getBlockAttributes: ( clientId: string ) => BlockAttributes;
+			};
```
louwie17: Should we move some of these `blockEditorStore` to a central place, like a types file? and use them here. Given we manually set these in a couple places, like here: https://github.com/woocommerce/woocommerce/blob/843187ceff692bb78d59e3110a436f7213805f4a/plugins/woocommerce/client/admin/client/customize-store/assembler-hub/auto-block-preview.tsx#L131-L133

louwie17 (changes requested):
This tested well, and changes look good for the most part. I did leave a couple comments, one important one around the `onChange` method. I am not sure if the change affected anything, but would break things.

plugins/woocommerce/client/admin/client/customize-store/assembler-hub/toolbar/shuffle.tsx
gigitux: As we decided, I added a comment (4fdd78f).

plugins/woocommerce/client/admin/client/customize-store/assembler-hub/block-editor-container.tsx
gigitux: > The above change would also remove the the selection data. But it also means the types on our end are wrong for the BlockEditor onChange method.

I updated the type definition with [4048b51](https://github.com/woocommerce/woocommerce/pull/54133/commits/4048b51a3cf2d44234cc9f6d6e8142b44d7406ae)

plugins/woocommerce/client/admin/client/customize-store/assembler-hub/hooks/use-editor-blocks.ts
```diff
@@ -7,21 +7,22 @@
 import { useEntityBlockEditor } from '@wordpress/core-data';
 import { BlockInstance } from '@wordpress/blocks';
 
-export type ChangeHandler = (
+type InputHandler = ( blocks: BlockInstance[] ) => void;
+type ChangeHandler = (
 	blocks: BlockInstance[],
-	options: Record< string, unknown >
+	options?: Record< string, unknown >
```
louwie17: I believe `options` is actually required here, so this would be the wrong type - https://github.com/WordPress/gutenberg/blob/609d573625be035953f3782efdc0589f617b4b78/packages/core-data/src/hooks/use-entity-block-editor.js#L95

I was actually thinking that we should update the `onChange` type in the `BlockEditor` component: https://github.com/woocommerce/woocommerce/blob/5345873399b7c7439a45fc6236238a336772f5e8/plugins/woocommerce/client/admin/client/customize-store/assembler-hub/block-editor.tsx#L30
And in turn also in the `BlockPreview` `onChange` prop: https://github.com/woocommerce/woocommerce/blob/816aee20f4e6eea96ff60724c043b8efa7c3c1bf/plugins/woocommerce/client/admin/client/customize-store/assembler-hub/block-preview.tsx#L51

I noticed though that unfortunately the type for the `BlockEditorProvider` is out of date: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/wordpress__block-editor/components/provider.d.ts#L30-L37
As you can see in the actual code it does call the `onChange` or `onInput` with a second parameter that contains the selection information: https://github.com/WordPress/gutenberg/blob/609a3c8d38a0608b41b1827ec4a8c1f812487ef2/packages/block-editor/src/components/provider/use-block-sync.js#L276
In our case I did prefer we add a `@ts-expect-error` comment with our use of `BlockEditorProvider`, instead of adopting to the wrong type.

louwie17 (changes requested):
Thanks for the updates @gigitux, I just left a follow up comment around the `useEditorBlocks` and `onChange` method updates. Let me know what you think.

plugins/woocommerce/client/admin/client/customize-store/assembler-hub/hooks/use-editor-blocks.ts
gigitux: Mmm, good point 🤔 Thanks for going in-depth! Fixed with ca7324dba057e76c01ce201170022832e65a7734

plugins/woocommerce/client/admin/client/customize-store/assembler-hub/block-editor-container.tsx
```diff
@@ -186,6 +186,7 @@ export const BlockEditorContainer = () => {
 		<BlockEditor
 			renderedBlocks={ blocks }
 			isScrollable={ isScrollable }
+			// @ts-expect-error The type is not up to date.
```
louwie17: Sorry to be nitpicky here, but  the `BlockEditor` component is actually an internal component we can update ourselves.
I was thinking of adding the above comment here: https://github.com/woocommerce/woocommerce/blob/816aee20f4e6eea96ff60724c043b8efa7c3c1bf/plugins/woocommerce/client/admin/client/customize-store/assembler-hub/block-preview.tsx#L49
As the `BlockEditorProvider` is imported from the GB package.

louwie17 (changes requested):
Sorry to be nitpicky, and I am sure you are ready to call it quits on these type changes. But left one more suggestion, I am also happy to push the change up though :)

plugins/woocommerce/client/admin/client/customize-store/assembler-hub/block-editor-container.tsx
gigitux: No worries! I appreciated it! I'm dealing with multiple TS errors in different branches, and it creates some confusion for me :D 

Fixed with 95ef4cdca179a935189e292cf46d315bb477955e

louwie17 (approved):
This looks good to me, thanks for addressing the several suggestions. It also remained to test well 👍

PR #54146, opened 2025-01-03
fix/demo migrate onboarding data

rjchow:
- **dev: patch @wordpress/data to add types for resolveSelect**
- **dev: patch @wordpress/data to add types for store metadata actions**
- **fix: demo onboarding store migration from registerStore() to register()**

Due to incomplete types in \@wordpress/data, this PR adds a patch to the type definitions for the package. This is an interim solution while an upstream fix to the types is properly prepared. The type definitions in this PR mostly result from conversion of the relevant parts of \@wordpress/data to Typescript, so this should be sound.

The other changes are a migration for the onboarding store, meant to serve both as a demo of the migration and also as a proving ground for the types since it is non-trivial.

There are also a bunch of TS exclusions for the product editor as I've deemed it out of scope to do the migration in this PR. Some of these exclusions result from incorrect types, and will require fixes in \@woocommerce/product-editor package.

1. Review the patch for \@wordpress/data for any feedback on the types
1. Review the \@woocommerce/data changes for the onboarding store migration
2. Honestly the scope of changes span large areas of WC and it is not really possible for any one person to test all the covered area. However the changes are quite homogeneous, involving a change of ONBOARDING_STORE_NAME string const to the storeDescriptor object. Testing a few of the features should provide confidence that it works as intended.
3. There are some actual code changes to some of the `invalidateResolution` calls, as they require a parameter for the args. (I'm not 100% certain it is required, but the JSDocs don't indicate that they are optional and the code doesn't seem to indicate either way). It would be good to test these areas out to ensure that they have not been broken.
  - payment recommendations (./plugins/woocommerce/client/admin/client/payments/payment-recommendations.tsx, ./plugins/woocommerce/client/admin/client/payments/use-payments-banner.ts)
  - deprecated tasks (?) (./plugins/woocommerce/client/admin/client/task-lists/fills/deprecated-tasks.tsx)
  - marketing task (./plugins/woocommerce/client/admin/client/task-lists/fills/Marketing/index.tsx)
  - task list (./plugins/woocommerce/client/admin/client/task-lists/task-lists.tsx)

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

Patch \@wordpress/data types to add types for metadata selectors, actions, and resolveSelect. Migrate onboarding store and usages to the new register(storeDescriptor) API

</details>

<details>

#### Comment 

</details>


docs/extension-development/handling-merchant-onboarding.md
```diff
@@ -171,7 +171,7 @@ Next, we create a [functional component](https://reactjs.org/docs/components-and
 
 ```js
 const Task = ( { onComplete, task } ) =&gt; {
-	const { actionTask } = useDispatch( ONBOARDING_STORE_NAME );
+	const { actionTask } = useDispatch( onboardingStore );
```
chihsuan, line 177: Could you please update the `#### Handle imports` section to include the import for `onboardingStore`? It appears we may have overlooked it.

bin/patches/@wordpress__data@10.0.2.patch
```diff
@@ -0,0 +1,428 @@
...
+- * ```js
+- * import { resolveSelect } from '@wordpress/data';
+- * import { store as myCustomStore } from 'my-custom-store';
+- *
+- * resolveSelect( myCustomStore ).getPrice( 'hammer' ).then(console.log)
+- * ```
+- *
+- * @return {Object} Object containing the store's promise-wrapped selectors.
+- */
+-export const resolveSelect: any;
```
chihsuan, line 36: Nice Improvement for removing `any` type here and properly defining it! 👍

bin/patches/@wordpress__data@10.0.2.patch
```diff
@@ -0,0 +1,428 @@
...
+diff --git a/build-types/redux-store/metadata/selectors.d.ts b/build-types/redux-store/metadata/selectors.d.ts
+index 3a44c74e8bb021444a3072331b546d1b16f4d58e..41bb4f3a712c9c20dafb03ed3750960f3cd1f1e5 100644
+--- a/build-types/redux-store/metadata/selectors.d.ts
++++ b/build-types/redux-store/metadata/selectors.d.ts
+@@ -14,7 +14,7 @@
+  *
+  * @return {StateValue|undefined} isResolving value.
+  */
+-export function getResolutionState(state: State, selectorName: string, args: unknown[] | null): StateValue | undefined;
++export declare function getResolutionState<Selectors extends Record<string, (...args: any[]) => any>>(state: Record<keyof Selectors, State>, selectorName: keyof Selectors, args: SelectorParameters<Selectors[typeof selectorName]> | undefined): StateValue | undefined;
```
chihsuan: This changed `null` to `undefined` for `args`. Was the change from null to undefined intentional?

The code seems to accept both `null` and `undefined` for `args`, but JSDoc only mentions `null`. I think we should include `null` for the type. 🤔

chihsuan: Other resolution selector functions (getResolutionState, getIsResolving, hasStartedResolution...) also changed from `null` to using `undefined`.

bin/patches/@wordpress__data@10.0.2.patch
```diff
@@ -0,0 +1,428 @@
...
++ * Given a store descriptor, returns an object containing the store's selectors pre-bound to state
++ * so that you only need to supply additional arguments, and modified so that they return promises
++ * that resolve to their eventual values, after any resolvers have ran.
++ *
++ * @param {StoreDescriptor|string} storeNameOrDescriptor The store descriptor. The legacy calling
++ *                                                       convention of passing the store name is
++ *                                                       also supported.
++ *
++ * @example
++ * ```js
```
chihsuan: nit:

```suggestion
+ * ```ts
```

chihsuan (approved):
Nice work!  @rjchow Tested the changes and did not find any issues. 👍 

I've left a few comments on the PR, but none of them are blocking.


>  There are some actual code changes to some of the `invalidateResolution` calls, as they require a parameter for the args. (I'm not 100% certain it is required, but the JSDocs don't indicate that they are optional and the code doesn't seem to indicate either way). It would be good to test these areas out to ensure that they have not been broken.
 
I guess you meant to change `hasFinishedResolution` instead of `invalidateResolution`? I only see `hasFinishedResolution` used in the codebase you mentioned.

Looking at the code, the [`selectorArgsToStateKey`](https://github.com/WordPress/gutenberg/blob/3416bf4b0db6679b86e8e4226cbdb0d3387b25d7/packages/data/src/redux-store/metadata/utils.ts#L41-L51) that used in the resolution functions converts `undefined` and `null` to empty array so I believe the changes are fine.

bin/patches/@wordpress__data@10.0.2.patch
rjchow: No it wasn't intentional, overlooked it. Thanks for picking up on it!

packages/js/data/src/index.ts
```diff
@@ -6,7 +6,6 @@ import '@wordpress/core-data';
 // Export store names
 export { SETTINGS_STORE_NAME } from './settings';
 export { PLUGINS_STORE_NAME } from './plugins';
-export { ONBOARDING_STORE_NAME } from './onboarding';
```
louwie17: Hmmm should we remove this export in one go? It could break plugins using it, but I am not entirely sure how we can deprecate a `const` in a good manner.

Having said that, I don't think it hurts still exporting this as using `select( store_string_name )` is still expected to work. I think just recommending to use `onboardingStore` in our docs would suffice.

packages/js/data/src/products/index.ts
```diff
@@ -26,3 +28,10 @@ registerStore< State >( STORE_NAME, {
 } );
 
 export const PRODUCTS_STORE_NAME = STORE_NAME;
+
+// This is necessary for the correct typing of resolveSelect until the migration to register(storeDescriptor) is complete.
```
louwie17, line 32: Could you add this comment in the product variations and tags stores as well?

packages/js/data/src/index.ts
rjchow: That's a great point I didn't consider 3rd party plugins using it! I'll put it back

rjchow:
Addressed Louren's PR feedback and also realised that I didn't fix the TS errors for ./data/onboarding, which I have done now

plugins/woocommerce/client/admin/client/core-profiler/index.tsx
```diff
@@ -615,15 +617,15 @@ const preFetchJetpackAuthUrl = assign( {
...
-	resolveSelect( ONBOARDING_STORE_NAME ).getFreeExtensions()
+	resolveSelect( onboardingStore ).getFreeExtensions()
 );
 
 const getPlugins = fromPromise( async () => {
-	dispatch( ONBOARDING_STORE_NAME ).invalidateResolution(
+	dispatch( onboardingStore ).invalidateResolutionForStoreSelector(
 		'getFreeExtensions'
 	);
 	const extensionsBundles: ExtensionList[] = await resolveSelect(
```
louwie17:
```suggestion
	const extensionsBundles = await resolveSelect(
```
Should we remove these types now? given they come correctly from `getFreeExtensions()` function?
Btw in this instance given they should be identical I don't mind leaving it in either.

louwie17 (approved):
Thanks for addressing my other suggestions, I just have one more slight suggestion, although to my that is low priority. For the rest things look good from my perspective.
I also just did a little bit of testing, not extensively.

plugins/woocommerce/client/admin/client/core-profiler/index.tsx
rjchow: lgtm, thanks ☺️

chihsuan (approved):
I fixed the conflicts and tested onboarding area again. It's working as expected.

PR #54156, opened 2025-01-03
Fix stock indicator visibility for Variable products

### LINKED ISSUE #53720: Stock Indicator for Variations visibility is not working well
### Prerequisites

### Describe the bug

The `Stock Indicator` block should be visible only when a variable product has no variations with a price. In such cases, the stock indicator message should read:
`This product is currently out of stock and unavailable.`
However, we are not showing that message.

Additionally, the `Stock Indicator` block should not be visible for variable products in any other circumstances. Currently, it is visible when it shouldn't be.

### Expected behavior

The Stock Indicator block should be visible when a variable product doesn't have variations with price.

The Stock Indicator block should be hidden for variable products in any other circumstances.

### Steps to reproduce

1. Create a new variable product; create a few variations, but don't add a price to them.
2.  Go to Pages > Add New Page.
3. Add a Single Product block and select the product you just created (not the variations, but the parent).
4. Add a Stock indicator inside the Single Product block.
5. The stock indicator should say:
```
This product is currently out of stock and unavailable.
```
 Instead, it will say:
```
Out of stock
```
6. Add a price for every variation and go back to the page you were creating.
7. No stock indicator message should be shown for variable products.

### WordPress Environment

It's happening on `trunk`.

### Isolating the problem

octaedro:

This PR fixes an error with the `Stock Indicator` message for variable products (in the `Single Product` block) when they don't have variations with a price. In such case, the stock indicator message will read:
`This product is currently out of stock and unavailable.`

| Before | After |
|--------|--------|
| Image | Image |
| Image | Image | 

Additionally, the `Stock Indicator` block should not be visible for variable products in any other circumstances. This PR also fixes that error.

Closes #53720.

1. Create a new variable product; create a few variations, but don't add a price to them.
2.  Go to Pages > Add New Page.
3. Add a `Single Product` block and select the product you just created (not the variations, but the parent).
4. Add a `Stock indicator` inside the `Single Product` block.
5. The stock indicator should say:
```
This product is currently out of stock and unavailable.
```
6. Save the page and view it.
7. Verify the text looks correct in the front end as well.
8. Add a price for every variation and go back to the page you created.
9. No stock indicator message should be shown for variable products.

Screenshot 2025-01-03 at 1 17 52 PM

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/src/Blocks/BlockTypes/ProductStockIndicator.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/ProductStockIndicator.php
@@ -2,6 +2,7 @@
 namespace Automattic\WooCommerce\Blocks\BlockTypes;
 
 use Automattic\WooCommerce\Blocks\Utils\StyleAttributesUtils;
+use Automattic\WooCommerce\Blocks\Utils\ProductAvailabilityUtils;
 use Automattic\WooCommerce\Enums\ProductType;
 
 /**
@@ -45,7 +46,7 @@ class ProductStockIndicator extends AbstractBlock {
 	 * @return array
 	 */
 	protected function get_product_types_without_stock_indicator() {
-		return array( ProductType::EXTERNAL, ProductType::GROUPED, ProductType::VARIABLE );
+		return array( ProductType::EXTERNAL, ProductType::GROUPED );
 	}
 
 	/**
@@ -82,15 +83,13 @@ class ProductStockIndicator extends AbstractBlock {
 			return '';
 		}
 
-		$availability = $product->get_availability();
+		$availability = ProductAvailabilityUtils::get_product_availability( $product );
 
 		if ( empty( $availability['availability'] ) ) {
 			return '';
 		}
 
-		$low_stock_amount   = $product->get_low_stock_amount();
 		$total_stock        = $product->get_stock_quantity();
-		$is_low_stock       = $low_stock_amount && $total_stock <= $low_stock_amount;
 		$classes_and_styles = StyleAttributesUtils::get_classes_and_styles_by_attributes( $attributes );
 
 		$classnames  = isset( $classes_and_styles['classes'] ) ? ' ' . $classes_and_styles['classes'] . ' ' : '';
--- a/plugins/woocommerce/src/Blocks/Utils/ProductAvailabilityUtils.php
+++ b/plugins/woocommerce/src/Blocks/Utils/ProductAvailabilityUtils.php
@@ -1,0 +1,48 @@
+<?php
+declare(strict_types=1);
+namespace Automattic\WooCommerce\Blocks\Utils;
+
+use Automattic\WooCommerce\Blocks\Templates\ProductStockIndicator;
+use Automattic\WooCommerce\Enums\ProductType;
+/**
+ * Utility functions for product availability.
+ */
+class ProductAvailabilityUtils {
+
+	/**
+	 * Get product availability information.
+	 *
+	 * @param \WC_Product $product Product object.
+	 * @return string[] The product availability class and text.
+	 */
+	public static function get_product_availability( $product ) {
+		$product_availability = array(
+			'availability' => '',
+			'class'        => '',
+		);
+
+		if ( ! $product ) {
+			return $product_availability;
+		}
+
+		// If the product is a variable product, check if it has any available variations.
+		// We will show a custom availability message if it does.
+		if ( $product->get_type() === ProductType::VARIABLE ) {
+			$available_variations = $product->get_available_variations();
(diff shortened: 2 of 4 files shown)


plugins/woocommerce/src/Blocks/BlockTypes/ProductStockIndicator.php
```diff
@@ -82,15 +83,13 @@ protected function render( $attributes, $content, $block ) {
...
-		$availability = $product->get_availability();
+		$availability = ProductAvailabilityUtils::get_product_availability( $product );
 
 		if ( empty( $availability['availability'] ) ) {
 			return '';
 		}
 
-		$low_stock_amount   = $product->get_low_stock_amount();
 		$total_stock        = $product->get_stock_quantity();
-		$is_low_stock       = $low_stock_amount && $total_stock <= $low_stock_amount;
```
octaedro, line 93: We don't use these two variables (`$low_stock_amount` and `$is_low_stock`) anymore.

plugins/woocommerce/src/Blocks/BlockTypes/ProductStockIndicator.php
```diff
@@ -45,7 +46,7 @@ protected function get_block_type_uses_context() {
 	 * @return array
 	 */
 	protected function get_product_types_without_stock_indicator() {
-		return array( ProductType::EXTERNAL, ProductType::GROUPED, ProductType::VARIABLE );
+		return array( ProductType::EXTERNAL, ProductType::GROUPED );
```
octaedro, line 49: Variable products only display a stock indicator under the circumstances outlined in this PR. For this reason, we need to remove it from this list.

plugins/woocommerce/src/Blocks/Utils/ProductAvailabilityUtils.php
```diff
@@ -0,0 +1,48 @@
...
+			return $product_availability;
+		}
+
+		// If the product is a variable product, check if it has any available variations.
+		// We will show a custom availability message if it does.
+		if ( $product->get_type() === ProductType::VARIABLE ) {
+			$available_variations = $product->get_available_variations();
+			if ( empty( $available_variations ) && false !== $available_variations ) {
+				$product_availability['availability'] = __( 'This product is currently out of stock and unavailable.', 'woocommerce' );
+				$product_availability['class']        = 'out-of-stock';
```
octaedro, line 34: We set the product availability class to `out-of-stock` for consistency with the variable product page. We assume it's out of stock [here](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/templates/single-product/add-to-cart/variable.php#L32).

Aljullu (commented):
Overall looks good, but I'm a bit confused now. If I have the _Stock Indicator_ and _Add to Cart with Options_ (the existing one) on the same _Single Product_ block, I see the message twice in the frontend:

imatge

Can you reproduce too? I think it's a bit unfortunate that WooCommerce is using the Variable product Add to Cart template to render this message instead of the Stock Indicator space.

I'm not sure what's a good solution here, to be honest. :confused:  Wondering if there is a way to make it so the message is only displayed conditionally, to make sure we don't display it twice.

octaedro:
Hi @Aljullu, thank you for reviewing this PR!

> If I have the _Stock Indicator_ and _Add to Cart with Options_ (the existing one) on the same _Single Product_ block, I see the message twice in the frontend:

I can reproduce this behavior, but I’m not sure if it’s a bug. The `Add to Cart with Options` block mimics the add to cart form used on the product page. Since we display the stock indicator on the product page, it makes sense for it to also appear in the `Add to Cart with Options` block ([add-to-cart-form](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartForm.php)).

> Wondering if there is a way to make it so the message is only displayed conditionally, to make sure we don't display it twice.

Considering that both the old and new versions of the `Add to Cart with Options` block automatically include the `Stock Indicator`, I believe that when a user chooses to add an additional `Stock Indicator` to the `Single Product` block, it indicates they intentionally want two stock indicators on the page.

One of the benefits of blocking the `Add to Cart with Options` is that it allows the user to remove the `Stock Indicator` from the add to cart and place it elsewhere on the page if desired.

Please let me know if I understood your comments correctly.

Aljullu (approved):
Yeah, sorry I was wrong, I thought it was a special case for this "special" availability message, but it happens in all of them, so I agree it's the expected behavior.

imatge

Disregard my previous comment and sorry for the confusion! :sweat_smile: 

LGTM :ship:

PR #54166, opened 2025-01-03
Add order by option to Product Collection block when in default query mode

Aljullu:

Part of https://github.com/woocommerce/woocommerce/discussions/52840.

This PR tries a new approach to handle global settings directly from the _Product Collection_ block, until now those settings were only accessible from the _Customizer_.

[Enregistrament de pantalla del 2025-01-09 18-30-34.webm](https://github.com/user-attachments/assets/f7f563ba-906b-4447-beb9-8d1e99425a78)

Specifically, this PR adds the _Default sorting_ option for queries to the _Product Collection_ block UI. In the future, we could expand it to also include _Products per page_ and _Columns_ options (https://github.com/woocommerce/woocommerce/issues/54240).

This approach is inspired by a similar implementation in the _Checkout_ block, which migrated from attributes to global settings (see https://github.com/woocommerce/woocommerce/pull/52784).

1. Go to the Customizer (`/wp-admin/customize.php`, note: if you are using a block theme, it's expected this option won't be visible in the UI) > WooCommerce > Product Catalog > Default product sorting and one of the options (ie: _Sort by most recent_). Click on _Publish_ to save.
3. Go to the _Shop_ page (`/shop`) in the frontend and verify blocks are sorted accordingly.
4. With a block theme, go to Appearance > Editor > Templates > Product Catalog (`/wp-admin/site-editor.php?postId=woocommerce%2Fwoocommerce%2F%2Farchive-product&postType=wp_template&canvas=edit`) and select the _Product Catalog_ block.
5. Verify _Query type_ is set to _Default_ and _Order by_ to _Newest to oldest_.
6. Change the _Order by_ value to something different (ie: _Price, low to high_).
7. Click on _Save_ and verify you can save the _Default product sorting_ separately from the template.
8. Continue saving and go to the _Shop_ frontend again.
9. Verify products are now sorted by price (descending).

### DIFF
--- a/plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/index.tsx
+++ b/plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/index.tsx
@@ -176,9 +176,7 @@ const ProductCollectionInspectorControls = (
 				{ showCustomOrderControl && (
 					<CustomQueryOrderByControl { ...queryControlProps } />
 				) }
-				{ showDefaultOrderControl && (
-					<CustomQueryOrderByControl { ...queryControlProps } />
-				) }
+				{ showDefaultOrderControl && <DefaultQueryOrderByControl /> }
 				{ showOffsetControl && (
 					<OffsetControl { ...queryControlProps } />
 				) }
--- a/plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/index.tsx
+++ b/plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/index.tsx
@@ -39,7 +39,8 @@ import {
 	InheritQueryControl,
 	FilterableControl,
 } from './use-page-context-control';
-import OrderByControl from './order-by-control';
+import DefaultQueryOrderByControl from './order-by-control/default-query-order-by-control';
+import CustomQueryOrderByControl from './order-by-control/custom-query-order-by-control';
 import OnSaleControl from './on-sale-control';
 import StockStatusControl from './stock-status-control';
 import KeywordControl from './keyword-control';
@@ -91,7 +92,7 @@ const ProductCollectionInspectorControls = (
 		isArchiveTemplate && shouldShowFilter( CoreFilterNames.INHERIT );
 	const showFilterableControl =
 		! isArchiveTemplate && shouldShowFilter( CoreFilterNames.FILTERABLE );
-	const showOrderControl =
+	const showCustomOrderControl =
 		showQueryControls && shouldShowFilter( CoreFilterNames.ORDER );
 	const showOffsetControl =
 		showQueryControls && shouldShowFilter( CoreFilterNames.OFFSET );
@@ -171,8 +172,10 @@ const ProductCollectionInspectorControls = (
 					<ProductsPerPageControl { ...queryControlProps } />
 				) }
 				<ColumnsControl { ...displayControlProps } />
-				{ showOrderControl && (
-					<OrderByControl { ...queryControlProps } />
+				{ showCustomOrderControl ? (
+					<CustomQueryOrderByControl { ...queryControlProps } />
+				) : (
+					<DefaultQueryOrderByControl />
 				) }
 				{ showOffsetControl && (
 					<OffsetControl { ...queryControlProps } />
--- a/plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/index.tsx
+++ b/plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/index.tsx
@@ -164,6 +164,14 @@ const ProductCollectionInspectorControls = (
 				{ showInheritQueryControl && (
 					<InheritQueryControl { ...queryControlProps } />
 				) }
+				{ showCustomOrderControl && (
+					<CustomQueryOrderByControl { ...queryControlProps } />
+				) }
+				{ showDefaultOrderControl && (
+					<DefaultQueryOrderByControl
+						trackInteraction={ trackInteraction }
+					/>
+				) }
 				{ showFilterableControl && (
 					<FilterableControl { ...queryControlProps } />
 				) }
@@ -173,14 +181,6 @@ const ProductCollectionInspectorControls = (
 					<ProductsPerPageControl { ...queryControlProps } />
 				) }
 				<ColumnsControl { ...displayControlProps } />
-				{ showCustomOrderControl && (
-					<CustomQueryOrderByControl { ...queryControlProps } />
-				) }
-				{ showDefaultOrderControl && (
-					<DefaultQueryOrderByControl
-						trackInteraction={ trackInteraction }
-					/>
-				) }
 				{ showOffsetControl && (
 					<OffsetControl { ...queryControlProps } />
 				) }
--- a/plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/index.tsx
+++ b/plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/index.tsx
@@ -87,21 +87,21 @@ const ProductCollectionInspectorControls = (
 		tracksLocation === 'product-catalog' ||
 		tracksLocation === 'product-archive';
 
-	const showQueryControls = inherit === false;
+	const showCustomQueryControls = inherit === false;
 	const showInheritQueryControl =
 		isArchiveTemplate && shouldShowFilter( CoreFilterNames.INHERIT );
 	const showFilterableControl =
 		! isArchiveTemplate && shouldShowFilter( CoreFilterNames.FILTERABLE );
 	const showCustomOrderControl =
-		showQueryControls && shouldShowFilter( CoreFilterNames.ORDER );
-	const showDefaultOrderControl = inherit === true;
+		showCustomQueryControls && shouldShowFilter( CoreFilterNames.ORDER );
+	const showDefaultOrderControl = ! showCustomQueryControls;
 	const showOffsetControl =
-		showQueryControls && shouldShowFilter( CoreFilterNames.OFFSET );
+		showCustomQueryControls && shouldShowFilter( CoreFilterNames.OFFSET );
 	const showMaxPagesToShowControl =
-		showQueryControls &&
+		showCustomQueryControls &&
 		shouldShowFilter( CoreFilterNames.MAX_PAGES_TO_SHOW );
 	const showProductsPerPageControl =
-		showQueryControls &&
+		showCustomQueryControls &&
 		shouldShowFilter( CoreFilterNames.PRODUCTS_PER_PAGE );
 	const showOnSaleControl = shouldShowFilter( CoreFilterNames.ON_SALE );
 	const showStockStatusControl = shouldShowFilter(
@@ -185,7 +185,7 @@ const ProductCollectionInspectorControls = (
 				) }
 			</ToolsPanel>
 
-			{ showQueryControls ? (
+			{ showCustomQueryControls ? (
 				<ToolsPanel
 					label={ __( 'Filters', 'woocommerce' ) }
 					resetAll={ ( resetAllFilters: ( () => void )[] ) => {
--- a/plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/index.tsx
+++ b/plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/index.tsx
@@ -176,7 +176,11 @@ const ProductCollectionInspectorControls = (
 				{ showCustomOrderControl && (
 					<CustomQueryOrderByControl { ...queryControlProps } />
 				) }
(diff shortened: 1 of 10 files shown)


Aljullu:
@imanish003 @kmanijak I'm pinging you two directly for review as I think you are the ones who worked the longest on the _Product Collection_ block and I would really appreciate your feedback. :pray:

imanish003:
@Aljullu Thanks for the ping, I will try to review it today.

plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/order-by-control/default-query-order-by-control.tsx
```diff
@@ -0,0 +1,73 @@
...
+		} );
+	};
+
+	return (
+		<OrderByControl
+			selectedValue={ value }
+			orderOptions={ orderOptions }
+			onChange={ onChange }
+			help={ __(
+				'This setting is sincyed across all Product Collection blocks using the default query type.',
```
imanish003:
```suggestion
				'This setting is synced across all Product Collection blocks using the default query type.',
```

imanish003 (dismissed):
@Aljullu I have some UX concerns about the current implementation:

1. The UI layout could be improved for clarity:
   - The "Default" query type setting and "Order By" option are currently separated, making their relationship unclear
   - We may consider grouping these related settings together to make their connection more obvious

image

2. Since this affects all Product Collection blocks using the "Default" query type:
   - We should make this behavior more prominent to users
   - Maybe we should consider adding a warning or info icon to highlight this global impact?
   - This would help prevent unintended changes to multiple blocks

IMO We should involve @j111q to provide design/UX guidance on these points. What do you think?

**Side Note**: I haven't reviewed code yet. I will review it once we agree on the UX approach. I hope that's okay.

plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/index.tsx
```diff
@@ -91,8 +92,9 @@ const ProductCollectionInspectorControls = (
 		isArchiveTemplate && shouldShowFilter( CoreFilterNames.INHERIT );
 	const showFilterableControl =
 		! isArchiveTemplate && shouldShowFilter( CoreFilterNames.FILTERABLE );
-	const showOrderControl =
+	const showCustomOrderControl =
 		showQueryControls && shouldShowFilter( CoreFilterNames.ORDER );
+	const showDefaultOrderControl = inherit === true;
```
kmanijak: Nitpick: What about renaming `showQueryControls` to `showCustomQueryControls`. And in that case I'd make the next condition as such:

```suggestion
	const showCustomOrderControl =
		showCustomQueryControls && shouldShowFilter( CoreFilterNames.ORDER );
	const showDefaultOrderControl = ! showCustomQueryControls;
```

This is to show explicitly they're mutually exclusive so we either show one or another.

plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/order-by-control/default-query-order-by-control.tsx
```diff
@@ -0,0 +1,73 @@
...
+	const [ value, setValue ] = useState(
+		settings.woocommerce_default_catalog_orderby || 'menu_order'
+	);
+
+	const onChange = ( newValue: string ) => {
+		setValue( newValue );
+		dispatch( coreStore ).editEntityRecord( 'root', 'site', undefined, {
+			[ `woocommerce_default_catalog_orderby` ]: newValue,
+		} );
+	};
```
kmanijak: I think we could track this as well to see the usage among other filters:

```suggestion
	const onChange = ( newValue: string ) => {
		setValue( newValue );
		dispatch( coreStore ).editEntityRecord( 'root', 'site', undefined, {
			[ `woocommerce_default_catalog_orderby` ]: newValue,
		} );
		trackInteraction( CoreFilterNames.DEFAULT_ORDER );
	};
```

You'd have to define new `CoreFilterNames`. WDYT?

plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/order-by-control/custom-query-order-by-control.tsx
```diff
@@ -90,33 +85,24 @@ const OrderByControl = ( props: QueryControlProps ) => {
...
-		<ToolsPanelItem
-			label={ __( 'Order by', 'woocommerce' ) }
+		<OrderByControl
+			selectedValue={ orderValue }
 			hasValue={ () =>
 				order !== DEFAULT_QUERY.order ||
 				orderBy !== DEFAULT_QUERY.orderBy
 			}
-			isShownByDefault
+			orderOptions={ orderOptions }
```
kmanijak, line 94: I think we should add a help text in this case as well to somehow explain to users this one is going to affect only this specific block. If help text in default order is `This setting is sincyed across all Product Collection blocks using the default query type.`, maybe something like:

```suggestion
			orderOptions={ orderOptions }
			help={ __( 'Set the products order in this collection.' ) }
```

kmanijak (commented):
I love that improvement @Aljullu! <3 Thanks for taking care of it! I added couple of minor comments.

regarding @imanish003's comment I'm also inclined to move the Order By settings up (both of them) below Query Type to highlight the relation between those two settings.

plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/order-by-control/default-query-order-by-control.tsx
Aljullu: Good catch, fixed!

plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/index.tsx
Aljullu: Good point, done in a11d09e7655179c1021917bb22d619f7262f38bc.

plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/order-by-control/default-query-order-by-control.tsx
Aljullu: Sounds good! Done in b06a85ec260739cc335888a18e006d3c66dada1d.

plugins/woocommerce-blocks/assets/js/blocks/product-collection/edit/inspector-controls/order-by-control/custom-query-order-by-control.tsx
Aljullu: Done in 7005065a99426172992efe83b6a19c8c5d0bbf07.

Aljullu:
Thanks for the reviews, folks! I applied your suggestions.

@j111q could we get your feedback on this one?

_tl;dr_ Product Collection can load products from either a Custom Query (customizable in the editor) or the Default Query (inherited from WordPress context). The Custom Query allows merchants to customize settings such as product order, the number of products displayed, additional filters, and more. In contrast, the Default Query, typically used in templates, offers more limited customization options.

This PR introduces the ability to modify the default product order when using the Default Query.

This is achieved by altering the default WordPress query, meaning the changes will apply to all Product Collection blocks set to "Default Query". However, we wonder whether this behavior might confuse users.

In the PR description, I posted a video showcasing how it currently looks in this PR. Depending on the selected query type, we display a different description under the _Order by_ controls:

* Default Query: _This setting is synced across all Product Collection blocks using the default query type_
* Custom Query: _Set the products order in this collection_

However, we wonder if this might not be enough and whether there is a better UX/UI way to implement it. Additionally, in https://github.com/woocommerce/woocommerce/issues/54240, we would also like to allow customizing the _Products per page_ and _Columns_ options for the Default Query (aka applying the changes to all Product Collection blocks using the Default Query mode).

For additional context, these settings were in the _Customizer_ in classic themes and we are now exploring the best location for them in block themes.

j111q:
Hmmmm interesting problem, thanks for flagging this 🙏 

A bit of a weird thought: I don't suppose we could use `Global Styles > Blocks > Product Collection Block` for this 😅 It's not really a _style_, more of a Setting, but it would make the "global impact" of this setting more obvious. But, it would also be much be hidden. So that's probably not a good idea...

A few more realistic suggestions:

Change `This setting is synced across all Product Collection blocks using the default query type` to 
`All Product Collection blocks using the Default Query will sync to this sort order.`

Change the copy of `Order By` to `Default Sorting` to match the dropdown copy in the front end. 

Would we be able to detect how many PC blocks with default query are in use by a store? 

I assume no, but if we could, then we could show something like, showing a modal or inline message that says:
`This sort order will be automatically synced to 4 product collection blocks in your store.`

Aljullu:
Thanks for taking a look, @j111q!

> A bit of a weird thought: I don't suppose we could use `Global Styles > Blocks > Product Collection Block` for this 😅 It's not really a _style_, more of a Setting, but it would make the "global impact" of this setting more obvious. But, it would also be much be hidden. So that's probably not a good idea...

I took a brief look and I don't think it's possible. My understanding is that it's intentional that plugins don't have a lot of control on the Global Styles panel. Gutenberg displays the styles blocks support, but it's not expected for plugins hook in there.

> Change `This setting is synced across all Product Collection blocks using the default query type` to 
`All Product Collection blocks using the Default Query will sync to this sort order.`

This one was easy, implemented in 761fc25065b92a910de15561a4b212dfa82bd864. :slightly_smiling_face: 

> Change the copy of `Order By` to `Default Sorting` to match the dropdown copy in the front end. 

It makes sense, but just to clarify: should we do this change in both cases (when Query Type is _Default_ and _Custom_) or only when the Query Type is _Default_?

> Would we be able to detect how many PC blocks with default query are in use by a store?

I don't think it's possible. The block can be in posts, pages, templates, widgets, added by plugins using PHP hooks, etc. so I don't think we have a reliable way to know how many instances of the block exist. That would also require us to define what counts as different _instances_. Ie: a template, widget or template part which is rendered in different frontend pages in the frontend, would count as one instance or as many instances as pages it's used in? In the end, I think this might be even more confusing than not saying the number of blocks it impacts. :sweat_smile: What do you think?

j111q:
> It makes sense, but just to clarify: should we do this change in both cases (when Query Type is Default and Custom) or only when the Query Type is Default?

Both please if we can! 🙏 

> In the end, I think this might be even more confusing than not saying the number of blocks it impacts. 😅 What do you think?

Yeah okay never mind that, hahaha thank you for your detailed reply 😆

kmanijak:
> > It makes sense, but just to clarify: should we do this change in both cases (when Query Type is Default and Custom) or only when the Query Type is Default?
> Both please if we can! 🙏

To be fair, if the `query type` is `custom`, then order is fixed and buyers cannot change it. So the order set by merchant is not only a default sorting. Calling it "_Default_ sorting" in this case may be misleading.

EDIT: On the other hand I understand the value of unified labels. "Order by" changing to "Default sorting" and the other way around is confusing.

And to make it even harder, Query Loop uses "Order by" label which is the same as the current Product Collection label 🙈

j111q:
@kmanijak Oh hmmmm 😅 That is tough. 
Let's use `Default sort by` for `Default Query` and `Order by` for `Custom Query`. 
I think that helps to bridge the gap between the two and align with `Query Loop` too.

Aljullu:
> Let's use `Default sort by` for `Default Query` and `Order by` for `Custom Query`. 
> I think that helps to bridge the gap between the two and align with `Query Loop` too.

Sounds good, done in e9e595af6c81b68836a404664ebe3462f09ad875.

@imanish003 @kmanijak, this should be ready for another review, now. :slightly_smiling_face:

plugins/woocommerce-blocks/tests/e2e/tests/product-collection/product-collection.block_theme.spec.ts
```diff
@@ -807,12 +807,61 @@ test.describe( 'Product Collection', () => {
...
+
 	test.describe( 'Editor: In taxonomies templates', () => {
 		test( 'Products by specific category template displays products from this category', async ( {
 			admin,
 			page,
 			editor,
 		} ) => {
+			await wpCLI(
+				'option update woocommerce_default_catalog_orderby price'
+			);
```
Aljullu, line 863: To be honest, I'm not 100% sure why this is needed. But the default sorting was working differently when the tests were run locally and when they were run in the GitHub jobs. I think because `menu_order` might return inconsistent results if products have never been reordered.

kmanijak (approved):
Works as expected! Thank you for addressing all of the comments and patience with figuring out copy! 🙌

Aljullu:
@imanish003 do you want to take another look? Otherwise, you will need to remove your review because I think it blocks merging. :thinking:

imanish003:
@Aljullu I have dismissed my review since @kmanijak has reviewed the PR. I don't believe my review is necessary, but let me know if you think it is. Great work ❤️

PR #54168, opened 2025-01-03
Show a notice when a site is disconnected from WooCommerce.com

thilinah:

This PR add a notice under `WooCommerce->Extensions->My Subscriptions` menu when a store is disconnected from WooCommerce.com. This is to inform all the site admins that the site is disconnected from a specific WooCommerce.com user account.

Closes # .

1. Create a WooCommerce build based on this PR. (`npm run build:zip`) and replace WooCommerce in your test site with the new build.
2. Connect your site to WooCommerce.com and disconnect it.
3. You should see a notice similar to this one under `WooCommerce->Extensions->My Subscriptions`
Screenshot 2025-01-03 at 18 15 39
4. Notice should be dismissible.
5. If the site is connected and disconnected the notice should be visible again with the correct email address it got disconnected from.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

Show a notice when a site is disconnected from WooCommerce.com

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/includes/admin/helper/class-wc-helper.php
```diff
@@ -2346,6 +2346,11 @@ public static function disconnect() {
 			)
 		);
 
+		$data = WC_Helper_Options::get( 'auth_user_data' );
+		WC_Helper_Options::update( 'last_disconnected_user_data', $data );
+		// Ignore all previously dismissed disconnect notices.
+		delete_metadata('user', 0, 'woo_disconnect_notice_dismiss', '', true);
```
anaspk: Can we use the `PluginsHelper::DISMISS_DISCONNECT_NOTICE` constant here instead of hard coded string `'woo_disconnect_notice_dismiss'` ?

anaspk (approved):
Thank you @thilinah , LGTM and works as described 👍

plugins/woocommerce/client/admin/client/marketplace/components/my-subscriptions/my-subscriptions.tsx
```diff
@@ -64,20 +66,47 @@ export default function MySubscriptions(): JSX.Element {
...
+			const data = {
+				notice_id: 'woo-disconnect-notice',
+				dismiss_notice_nonce: wccomSettings?.dismissNoticeNonce || '',
+			};
+			apiFetch( {
+				path: `/wc-admin/notice/dismiss`,
+				method: 'POST',
+				data,
+			} );
+			localStorage.setItem(
```
hesahesa, line 79: why do we need to manipulate this localstorage? Is this to force the `<Notice` to always show the `Notice` according to the component logic so that we can show/hide notice based on backend option value?

hesahesa (approved):
LGTM

plugins/woocommerce/client/admin/client/marketplace/components/my-subscriptions/my-subscriptions.tsx
```diff
@@ -22,6 +22,8 @@ import { RefreshButton } from './table/actions/refresh-button';
 import Notices from './notices';
 import InstallModal from './table/actions/install-modal';
 import { connectUrl } from '../../utils/functions';
+import Notice from '../notice/notice';
+import apiFetch from '@wordpress/api-fetch';
```
hesahesa: we probably should move this to External dependencies above (line 7)

hesahesa:
also @thilinah i found some weird behaviour when I dismiss the notice and then go to another tab, then go back to the my subscription tab, the notice is still shown

when I refresh the page, the notice disappeared tho (so this probably a small issue only)


https://github.com/user-attachments/assets/677ae969-90df-4dca-b10a-8566d6b08ec0

plugins/woocommerce/client/admin/client/marketplace/components/my-subscriptions/my-subscriptions.tsx
thilinah: Our Notice component adds an entry to the local storage when the notice is closed. For most of the cases even if the notice dismiss state is managed in backend this is fine. But what if:
1. Site is disconnected.
2. Notice is added as dismissed to the localStorage.
3. The site is connected again and disconnected.

Since we first relay on localStorage, the dismiss notice will not be displayed the second time. 

I didn't want to change the default behaviour as the Noticed is used in many places. But here just removing the entry from localStorage so we only depend on the dismiss status stored in DB.

plugins/woocommerce/client/admin/client/marketplace/components/my-subscriptions/my-subscriptions.tsx
thilinah: Thanks @hesahesa, fixed with https://github.com/woocommerce/woocommerce/pull/54168/commits/36b704b9505fbed76952dc4d5cc1c37274ccfa6e

plugins/woocommerce/includes/admin/helper/class-wc-helper.php
thilinah: Thanks @anaspk. Fixed with https://github.com/woocommerce/woocommerce/pull/54168/commits/f8c19a4848a2e8683ad13966773ed5277dac3e9e

PR #52797, opened 2024-11-14
Add defensive checks to `wc_get_order_status_name()`

barryhughes:
As documented in the [linked issue](https://github.com/woocommerce/woocommerce/issues/51690), we've seen that fatal errors can occur if `wc_get_order_status_name()` is called with something other than a string (such as `null`).

This change tries to mitigate that, and also seeks to inform developers/site operators of any problems by issuing a *doing it wrong* notice if either of the following are true:

- The provided status is unknown/unregistered.
- The provided status is not a string.

Notes:

- PHP can coerce some types (like integers) into strings without any problem, but other types like arrays will lead to notices, warnings or errors being issued (the severity level depends on the PHP runtime being used). My stance is that is fine (we shouldn't seek to prevent those from taking place, we just want to avoid fatal errors).
- There are a few cases where a fatal error still happens, such as if an object is supplied instead of a string. I think it would also be possible to catch this and handle it gracefully ... but this is such an odd case that I don't think we *should.* We also don't know of any real-world cases where this is happening.

Closes #51690.

### How to test the changes in this Pull Request:

Real-world testing can be done in the following ways:

1. Install and activate Query Monitor, or ensure you have some other means of detecting errors and problems (including *doing it wrong* notices).
2. Ensure you have some order data.
3. Visit **WooCommerce ‣ Orders**. There should be no errors that map back to the `wc_get_order_status_name()` function (this function will be called multiple times when this screen is rendered, so this test is just to give us additional confidence that nothing has broken as a result of this change).

You can also do some console-based testing via `wp shell`. Examples:

```sh
>>> wc_get_order_status_name( 'pending' )
=> "Pending payment"
```

```sh
>>> wc_get_order_status_name( 'unknown' )
PHP Notice:  Function wc_get_order_status_name was called <strong>incorrectly</strong>. An invalid order status slug was supplied. 
=> "unknown"
```

```sh
>>> wc_get_order_status_name( 5 )
PHP Notice:  Function wc_get_order_status_name was called <strong>incorrectly</strong>. An invalid order status slug was supplied. 
=> "5"
```

```sh
>>> wc_get_order_status_name( null )
PHP Notice:  Function wc_get_order_status_name was called <strong>incorrectly</strong>. An invalid order status slug was supplied.
=> ""
```

```sh
>>> wc_get_order_status_name( [] )
PHP Warning:  Array to string conversion
PHP Notice:  Function wc_get_order_status_name was called <strong>incorrectly</strong>. An invalid order status slug was supplied. 
[...]

PR #54172, opened 2025-01-03
Relax check in `wc_get_order_status_name()`

jorgeatorres:

In #52797 (targeting 9.6) we introduced some defensive checks to prevent fatal errors which could result from calling `wc_get_order_status_name()` with an invalid parameters (such as a NULL value). In those cases, the function would trigger a doing-it-wrong notice.

It'll also do so when passed an order status that doesn't exist, which can have an effect on other plugins using order-derived types (such as Subscriptions) and using their own order statuses (which aren't included in the default order statuses as they aren't truly **order** statuses). See https://github.com/Automattic/woocommerce-payments/issues/10053

This PR reintroduces prior behavior where passing a non-existent status wouldn't trigger a doing-it-wrong notice. We keep the checks to ensure the parameter is a string to prevent any fatal errors.

Related: p1735820321777489-slack-C0E1AV8T0

(Slightly adapted from #52797)

   > Real-world testing can be done in the following ways:
   > 
   > 1. Install and activate Query Monitor, or ensure you have some other means of detecting errors and problems (including *doing it wrong* notices).
   > 2. Ensure you have some order data.
   > 3. Visit **WooCommerce ‣ Orders**. There should be no errors that map back to the `wc_get_order_status_name()` function (this function will be called    > multiple times when this screen is rendered, so this test is just to give us additional confidence that nothing has broken as a result of this change).
   > 
   > You can also do some console-based testing via `wp shell`. Examples:
   > 
   > ```sh
   > >>> wc_get_order_status_name( 'pending' )
   > => "Pending payment"
   > ```
   > 
   > ```sh
   > >>> wc_get_order_status_name( 'unknown' )
   > => "unknown"
   > ```
   > 
   > ```sh
   > >>> wc_get_order_status_name( 5 )
   > PHP Notice:  Function wc_get_order_status_name was called <strong>incorrectly</strong>. An invalid order status slug was supplied. 
   > => "5"
   > ```
   > 
   > ```sh
   > >>> wc_get_order_status_name( null )
   > PHP Notice:  Function wc_get_order_status_name was called <strong>incorrectly</strong>. An invalid order status slug was supplied.
   > => ""
   > ```
   > 
   > ```sh
   > >>> wc_get_order_status_name( [] )
   > PHP Warning:  Array to string conversion
   > PHP Notice:  Function wc_get_order_status_name was called <strong>incorrectly</strong>. An invalid order status slug was supplied. 
   > => "Array"
   > ```

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/fix-51690-redux
+++ b/plugins/woocommerce/changelog/fix-51690-redux
@@ -1,0 +1,5 @@
+Significance: patch
+Type: tweak
+Comment: This change slightly relaxes a check that was introduced in a prior change (and for which a changelog entry already exists).
+
+
--- a/plugins/woocommerce/includes/wc-order-functions.php
+++ b/plugins/woocommerce/includes/wc-order-functions.php
@@ -164,7 +164,7 @@ function wc_get_is_pending_statuses() {
  */
 function wc_get_order_status_name( $status ) {
 	// "Special statuses": these are in common usage across WooCommerce, but are not normally returned by
-	// wc_get_order_statuses(). We add them only to avoid unnecessary doing-it-wrong errors.
+	// wc_get_order_statuses().
 	$special_statuses = array(
 		'wc-' . OrderStatus::AUTO_DRAFT => OrderStatus::AUTO_DRAFT,
 		'wc-' . OrderStatus::TRASH      => OrderStatus::TRASH,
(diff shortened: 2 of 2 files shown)


jorgeatorres:
@mattallan: Tagging you as reviewer to confirm this helps with the doing-it-wrong notices coming from Subscriptions.

coreymckrill (approved):
👍 Tried the commands in wp shell and works as expected, with `unknown` _not_ triggering the notice. Although, the notice was also not triggered for `wc_get_order_status_name( null )`, it just returned an empty string. This could, perhaps, be because my test environment is running PHP 8.1, which I think is a bit more liberal with recasting things on the fly.

Approving, but not merging so that @mattallan can confirm first that this fixes the Subscriptions issue.

mattallan (approved):
Thanks @jorgeatorres ! I've tested these changes against Woo Subscriptions and can confirm the `_doing_it_wrong()` is no longer triggered on our Change Payment page.

Thanks for the quick PR!

Issue #36689, opened 2023-01-30
FlexSlider clean-up


apenchev:
### Describe the solution you'd like

**jquery.flexslider.js** contains legacy code for unsupported browsers that is no longer needed (checking for `-webkit-transform` and `-moz-transform` in addition to the `transform` CSS property or checking for `msGesture` support, for instance). From my testing on our website, the code can be safely removed.

PR #54176, opened 2025-01-04
FlexSlider variable slider.transitions renamed to slider.transforms

apenchev:

FlexSlider now always uses CSS3 transitions, but can be configured to either use CSS 3 transforms or margin-[left|right] for positioning. We should rename the obsolete `slider.transitions` variable to `slider.transforms` so that reading the FlexSlider code makes sense. We keep the old `slider.transitions` variable name just in case that some third-party plugin checks for its value.

Updates #36689.

1. Confirm that the gallery of a single product page with multiple images successfully changes between images when using a slide animation.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 
FlexSlider variable slider.transitions renamed to slider.transforms
</details>

<details>

#### Comment 

</details>


barryhughes:
I didn't test, though it looks reasonable to me :-) ... Requesting a different reviewer, though, as this touches the classic storefront (if I've requested the wrong team/person, apologies + please reassign!).

tjcafferkey (approved):
I've tested this with multiple themes, and also checked there was no regressions with `useCss: false;` and everything appears to be working as expected.

Thanks for this contribution @apenchev!

My only feedback would be some more thorough testing instructions to take the guessing out of anyone reviewing who may not be familiar. E.g.

* Do we need to test against classic _and_ block themes (even naming some popular ones e.g. Storefront)?
* Regression test against `useCSS: false;` setting
* Are there any scenarios we can ignore?

Similarly if you have any feedback for us on how easy/difficult to found contributing we'd love to hear them!

Again, much appreciated. Approving 😄 Thanks!

Issue #54235, opened 2025-01-07
Product Variations List API enhacements


albarin:
We need to implement the same filters (except `include/exclude_type`) in the product variations list API endpoint.
For context: pdfdoF-68d-p2#comment-7252

`include_status`: Array of variation statuses to include as a comma-separated list, taking precedence over status
`exclude_status`: Array of variation statuses to exclude as a comma-separated list, taking precedence over include_status
`virtual`: Boolean – whether to include or exclude virtual variations.
`downloadable`: Boolean – whether to include or exclude downloadable variations.

We need to update the docs API also.

PR #54246, opened 2025-01-07
Add new `include_status` and `exclude_status` filter on the products variations endpoint

albarin:

This PR introduces the new `include_status` and `exclude_status` parameters to the list product variations endpoint.
It allows filtering by multiple statuses simultaneously: `any`, `draft`, `pending`, `private`, and `publish`. 

With this implementation, `include_status` takes precedence over `status`, if both of them are present.
Also, `exclude_status` takes precedence over `status` and `include_status` if any of them is present in the same request.

Part of https://github.com/woocommerce/woocommerce/issues/54235

**include_status test**
1. Make sure you have some variable products with variations and different statuses.
2. Go to `wp-admin/admin.php?page=wc-settings&tab=advanced&section=keys` and add a new key with `Read/Write` permissions.
3. Make the following request using the consumer key and secret generated in the previous step, adding the `include_status` param:
```bash
curl --insecure -u [KEY]:[SECRET] --request GET --url 'https://[STORE_URL]/wp-json/wc/v3/products/[VARIABLE_PRODUCT_ID]/variations?include_status=publish' --header 'Content-Type: application/json'
```
4. Ensure the result returns only products with the status `publish`.
5. Repeat the request with multiple statuses (comma-separated) and check the response makes sense.
6. Try with a non-existing status, it should return a 400 response.
7. Try with `any`, it should return all products independently of their status.

**exclude_status test**
1. Make the following request adding the `exclude_status` param:
```bash
curl --insecure -u [KEY]:[SECRET] --request GET --url 'https://[STORE_URL]/wp-json/wc/v3/products/[VARIABLE_PRODUCT_ID]/variations?exclude_status=publish' --header 'Content-Type: application/json'
```
1. Ensure the result returns only products with a status different than `publish`.
2. Repeat the request with multiple statuses (comma-separated) and check the response makes sense.
3. Try with a non-existing status, it should return a 400 response.
4. Try any other combinations you can think of (with both params and also `status`), and check the responses are consistent.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Add new `include_status` and `exclude_status` filter on the products variations endpoint.
</details>

<details>

#### Comment 

</details>


coreymckrill:
🤔 Is there actually a way to set the publish status of individual variations? I'm not seeing one in the web UI. If we're actually only setting the statuses of variable products, and not their child variations, it seems like we could just do something like

```
GET wc/v3/products?type=variable&include_status=draft,pending
```

If the variations will always have the same status as their parent, I'm not sure I see the value in adding these parameters to the variations endpoint. But I could be missing something.

albarin:
> 🤔 Is there actually a way to set the publish status of individual variations?

From the admin UI you can only change the `Enabled` flag for individual variations, which will set their status as `publish` or `private`. 
Screenshot 2025-01-10 at 09 33 38

Also, they can be set to other statuses programmatically, so I think there's value in having the filter, wdyt? @coreymckrill

coreymckrill:
Ah, I didn't realize the `Enable` control used the status. TIL! In that case, this does seem to have a clear use case.

coreymckrill (approved):
👍 Tested well locally

* Tested `include_status` with a single status, multiple statuses, an invalid status, `any`
* Tested `exclude_status` with a single status, multiple statuses, an invalid status
* Tested that `exclude_status` takes precedence over `include_status` and `status`

Very thorough set of unit tests, thank you!

joshheald:
@albarin Thanks for the fast work on this and the other filters.

I tested them today, using Jurassic Ninja pointed at trunk via WC Beta Tester. That resulted in the build named `9.7.0-dev-12709385361-gb898415`

Downloadable and Virtual variation filters worked as expected, but I couldn't get the status filters to work. I'm assuming that the build I've used doesn't actually include the changes... but is there a way to check from the build name?

Here's an example request:
```
rest-api/?json=true&path=/wc/v3/products/99/variations%26_method%3Dget&query=%7B%22per_page%22%3A%2225%22%2C%22downloadable%22%3A%22false%22%2C%22page%22%3A%221%22%2C%22exclude_status%22%3A%22publish%22%7D
```

And part of the response – you can see the published variation is included:

```
{
  "data": [
    {
      "id": 146,
      "type": "variation",
      "date_created": "2025-01-13T19:35:08",
      "date_created_gmt": "2025-01-13T19:35:08",
      "date_modified": "2025-01-13T20:11:32",
      "date_modified_gmt": "2025-01-13T20:11:32",
      "description": "",
      "permalink": "https://grebe-of-kites.jurassic.ninja/product/aerodynamic-paper-bag/?attribute_pa_size=xxl&attribute_dolor-praesentium-harum=beatae&attribute_pa_color=nihil-ut",
      "sku": "aerodynamic-paper-bag-53105264",
      "global_unique_id": "",
      "price": "783.27",
      "regular_price": "783.27",
      "sale_price": "318.51",
      "date_on_sale_from": null,
      "date_on_sale_from_gmt": null,
      "date_on_sale_to": "1973-04-08T23:59:59",
      "date_on_sale_to_gmt": "1973-04-08T23:59:59",
      "on_sale": false,
      "status": "publish",
      "purchasable": true,
      "virtual": true,
      "downloadable": false,
      "downloads": [],
      "download_limit": -1,
      "download_expiry": -1,
      "tax_status": "taxable",
      "tax_class": "",
      "manage_stock": false,
      "stock_quantity": null,
      "stock_status": "instock",
      "backorders": "no",
      "backorders_allowed": false,
      "backordered": false,
      "low_stock_amount": null,
      "weight": "90",
      "dimensions": {
        "length": "107",
        "width": "193",
        "height": "80"
      },
      "shipping_class": "",
      "shipping_class_id": 0,
      "image": {
        "id": 14,
        "date_created": "2025-01-13T19:34:48",
```

albarin:
@joshheald I guess the only way to check would be to go to `wp-admin/plugin-editor.php` on your site and check the controller. I'm not sure what's going on there, I tried with a new JN site and `trunk` and could make it work, I'll share the site with you.

joshheald:
@albarin I've tested on the side you shared and include/exclude work correctly there. Thanks!

I checked my test site in the plugin editor, and confirmed that these changes aren't on the branch it's built from... which is weird, because I built it from trunk yesterday, but never mind. It still explains why it didn't work!

Thanks again for the quick work on this.

PR #54262, opened 2025-01-08
Harden type in BlockRegistrationManager

Aljullu:

Follow-up of https://github.com/woocommerce/woocommerce/pull/54117. After merging that PR I realized we could directly use `ProductBlockConfig` in one of the type definitions to harden and simplify it.

This PR only includes changes to types, so there are no user-facing changes.

1. Open `plugins/woocommerce-blocks/assets/js/atomic/utils/register-product-block-type.ts`.
2. Verify there are no TypeScript errors introduced by this PR.


tjcafferkey (approved):
Good update! Thanks @Aljullu

PR #54202, opened 2025-01-06
Fix - Skip variations without prices during bulk price adjustments to prevent 500 error

imanish003:
### Changes proposed in this Pull Request:

Closes #54140

This PR fixes an issue with bulk price adjustments for variations where the operation would fail silently with an infinite loading spinner if any variation was missing a price. The fix skips variations without prices set, allowing the bulk update to proceed for variations that do have prices.

### How to test the changes in this Pull Request:

1. Create a variable product with multiple variations
2. Leave some variations without prices set
3. Click on "Bulk actions" dropdown and select "Increase regular prices (fixed amount or percentage)"
4. Enter a value (e.g. "10" or "10%") and click OK
5. Verify that:
	- The operation completes successfully
	- Only variations with existing prices are updated
	- Variations without prices are skipped
	- No infinite loading spinner appears
6. Perform the same steps i.e. step 1-5 for following options:
	- "Decrease regular prices (fixed amount or percentage)"
	- "Increase sale prices (fixed amount or percentage)"
	- "Decrease sale prices (fixed amount or percentage)"

Image

### Changelog entry

-   [x] Automatically create a changelog entry from the details below.

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Details</summary>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

-   [x] Fix - Fixes an existing bug
-   [ ] Add - Adds functionality
-   [ ] Update - Update existing functionality
-   [ ] Dev - Development related task
-   [ ] Tweak - A minor adjustment to the codebase
-   [ ] Performance - Address performance issues
-   [ ] Enhancement - Improvement to existing functionality

#### Message 
Fix - Skip variations without prices during bulk price adjustments to prevent 500 error
</details>

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

PR #54266, opened 2025-01-08
Enhancement: Improve UX by showing price validation warnings before bulk updates for variations

imanish003:

Context: https://github.com/woocommerce/woocommerce/pull/54202#pullrequestreview-2532742828

This PR improves the user experience during bulk price adjustments for variations by adding a clear warning about sale price validation rules. When users attempt to modify prices in bulk, they'll now see an upfront warning that explains how sale prices are validated against regular prices.

This helps prevent user confusion by explaining the validation rules before prices are modified, rather than after invalid prices are automatically removed.

We need to show this message only if there is the possibility that the sale price could become more than the regular price. This happens in two cases:
- Increase sale price
- Decrease regular price

Closes # .

1. Create a variable product with multiple variations
2. Set regular price and sale price for a variation. For example, set regular price to $50 and sale price to $45.
3. Click on "Bulk actions" dropdown and select "Increase sale prices (fixed amount or percentage)"
4. Verify that: In alert, you can see a warning message with text "Warning: Sale prices will be removed if they are not lower than regular prices."
image

5. Enter a value (e.g. "10") and click OK
6. Verify that:
	- Sale price is removed for variations where because of bulk operation sale prices becomes higher than regular prices
7. Perform the same steps i.e. step 1-5 for "Decrease regular prices (fixed amount or percentage)"

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 
Enhancement: Improve UX by showing price validation warnings before bulk updates for variations
</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/54266-enhance-bulk-price-validation-warning
+++ b/plugins/woocommerce/changelog/54266-enhance-bulk-price-validation-warning
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+Enhancement: Improve UX by showing price validation warnings before bulk updates for variations
--- a/plugins/woocommerce/client/legacy/js/admin/meta-boxes-product-variation.js
+++ b/plugins/woocommerce/client/legacy/js/admin/meta-boxes-product-variation.js
@@ -1228,9 +1228,12 @@ jQuery( function ( $ ) {
 				case 'variable_regular_price_decrease':
 				case 'variable_sale_price_increase':
 				case 'variable_sale_price_decrease':
-					value = window.prompt(
-						woocommerce_admin_meta_boxes_variations.i18n_enter_a_value_fixed_or_percent
-					);
+					const message =
+						woocommerce_admin_meta_boxes_variations.i18n_enter_a_value_fixed_or_percent +
+						'\n\n' +
+						woocommerce_admin_meta_boxes_variations.i18n_sale_price_warning;
+
+					value = window.prompt( message );
 
 					if ( value != null ) {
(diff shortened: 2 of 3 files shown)


octaedro (approved):
Nice job @imanish003! The code looks good to me and is testing well on my end.

While the code meets the requirements outlined in the testing instructions, I’d also like to hear @j111q’s input on this. The warning message definitely enhances the user experience by informing them that the sale price will be deleted under those circumstances. However, after giving it more thought, I think it might be worth reviewing the action the user is facing. For example, we could consider adding a JS validation to the prompt instead of relying solely on a warning.

I’ll pre-approve this PR, but I’d recommend waiting for @j111q’s feedback before merging.

imanish003:
@octaedro I agree we should wait for @j111q's input. However, IMO we should merge this PR, since it simply adds a warning message for existing behavior. We can update the message copy if needed based on feedback. Any additional UX improvements could be explored in a future iteration if deemed valuable.

> For example, we could consider adding a JS validation to the prompt instead of relying solely on a warning.

FWIW While this is an interesting suggestion, implementing JS validation would require an additional API call to fetch all variations and will require more effort. Happy to implement such solution if we think it's worth the effort.

octaedro:
Hi @imanish003,

> IMO we should merge this PR, since it simply adds a warning message for existing behavior. We can update the message copy if needed based on feedback. Any additional UX improvements could be explored in a future iteration if deemed valuable.

Sounds good to me! Let's merge this PR. Thank you for working on this!

PR #54288, opened 2025-01-08
Disable the Store Notice when switching to a block theme

### LINKED ISSUE #54237: Disable Store Notice when switching from a classic theme to a block theme
Based on the discussion in https://github.com/woocommerce/woocommerce/discussions/52840, we would like to disable the Store Notice when switching from a classic theme to a block theme. This needs to be done in the following way:

1. It shouldn't impact live stores using the Store Notice, even if they are in a block theme.
2. The Store Notice should be recovered when switching back to a classic theme.

~To accomplish that, I thought the implementation could work as follows: when switching from a classic theme to a block theme, we save an option in the database to disable the Store Notice in the frontend. Then, in the code responsible for displaying the notice, we check if the store is using a block them and check whether the option is set, in that case, the Store Notice won't be displayed. This way, we guarantee backwards compatibility and only disable the Store Notice in stores switching themes once the feature is implemented.~

Aljullu:

Closes https://github.com/woocommerce/woocommerce/issues/54237.

This PR makes it so:

* when switching from a classic theme to a block theme:
  * if the Store Notice is inactive, we do nothing
  * if the Store Notice is active, we deactivate it and save and option to flag that we will need to reactivate it when switching back to a classic theme (`woocommerce_enable_store_notice_in_classic_theme`)
* when switching from a block theme to a classic theme:
  * if `woocommerce_enable_store_notice_in_classic_theme` is not set, we do nothing
  * if `woocommerce_enable_store_notice_in_classic_theme` is set to `yes`, we re-enable the Store Notice

In other words, we disable the Store Notice when switching to a block theme and reactivate it when switching to a classic theme (assuming it was active).

1. Switch your theme to a classic theme (ie: Storefront) and go to Appearance > Customizer > WooCommerce > Store Notice and enable it. Click on _Publish_ to save.
2. Verify the store notice is visible on the frontend.
3. Switch to a block theme (ie: Twenty Twenty Five).
4. Verify the store notice is no longer visible on the frontend.
5. Switch back to a classic theme.
6. Verify the store notice is visible again on the frontend.
7. Go to Appearance > Customizer > WooCommerce > Store Notice and disable it. Click on _Publish_ to save.
8. Switch to a block theme.
9. Switch back to a classic theme.
10. Verify the store notice isn't visible on the frontend (it wasn't reactivated by mistake).

### DIFF
--- a/plugins/woocommerce/includes/wc-template-functions.php
+++ b/plugins/woocommerce/includes/wc-template-functions.php
@@ -4205,6 +4205,60 @@ function wc_set_hooked_blocks_version() {
 }
 
 /**
+ * Attach functions that listen to theme switches.
+ *
+ * @since 9.7.0
+ *
+ * @param string    $old_name Old theme name.
+ * @param \WP_Theme $old_theme Instance of the old theme.
+ * @return void
+ */
+function wc_after_switch_theme( $old_name, $old_theme ) {
+	wc_set_hooked_blocks_version_on_theme_switch( $old_name, $old_theme );
+	wc_update_store_notice_visible_on_theme_switch( $old_name, $old_theme );
+}
+
+/**
+ * Update the Store Notice visibility when switching themes:
+ * - When switching from a classic theme to a block theme, disable it.
+ * - When switching from a block theme to a classic theme, re-enable only if it was enabled when switching from a classic theme to a block theme.
+ *
+ * @since 9.7.0
+ *
+ * @param string    $old_name Old theme name.
+ * @param \WP_Theme $old_theme Instance of the old theme.
+ * @return void
+ */
+function wc_update_store_notice_visible_on_theme_switch( $old_name, $old_theme ) {
+	$enable_store_notice_in_classic_theme_option = 'woocommerce_enable_store_notice_in_classic_theme';
+	$is_store_notice_active_option               = 'woocommerce_demo_store';
+
+	if ( ! $old_theme->is_block_theme() && wc_current_theme_is_fse_theme() ) {
+		/*
+		 * When switching from a classic theme to a block theme, check if the store notice is active,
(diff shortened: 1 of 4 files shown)


plugins/woocommerce/includes/wc-template-functions.php
```diff
@@ -4204,6 +4204,61 @@ function wc_set_hooked_blocks_version() {
...
+	if ( ! $old_theme->is_block_theme() && wc_current_theme_is_fse_theme() ) {
+		/*
+		 * When switching from a classic theme to a block theme, check if the store notice is active,
+		 * if it is, disable it but set an option to re-enable it when switching back to a classic theme.
+		 */
+		$store_notice_is_active = get_option( $is_store_notice_active_option, 'no' );
+
+		if ( 'yes' === $store_notice_is_active ) {
+			update_option( $is_store_notice_active_option, 'no' );
+			add_option( $enable_store_notice_in_classic_theme_option, 'yes' );
```
tjcafferkey: I know we are using `yes` and `no` here but in WC there are `wc_bool_to_string` and `wc_string_to_bool` functions [here](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/wc-formatting-functions.php#L18-L42). Shall we use these instead?

tjcafferkey (approved):
I've left a non-blocking comment but pre-approving as everything looks good and works as expect! Thanks adding unit tests too!

plugins/woocommerce/includes/wc-template-functions.php
Aljullu: Good point, better to be cautious just in case. :slightly_smiling_face: I fixed it here b6c795c38595d1aa8dc5ff30501b55047c48abe1. I also made it so we use the existing `is_store_notice_showing()` function instead of getting the option directly.

PR #54312, opened 2025-01-09
Product Editor: fix remaining TS errors

gigitux:

Closes # .

1. Run `pnpm run --filter="@woocommerce/plugin-woocommerce" build`
2. Ensure that the build works correctly.


packages/js/product-editor/src/components/attribute-combobox-field/index.tsx
```diff
@@ -219,6 +219,7 @@ const AttributesComboboxControl: React.FC<
 			<BaseControl label={ label } help={ help } id={ labelFor }>
 				<ComboboxControl
 					className="woocommerce-attributes-combobox"
+					// @ts-expect-error allowReset is a valid prop of ComboboxControl.
 					allowReset={ false }
```
louwie17, line 224: This `ts-expect-error` is not necessary, this is actually an issue with how ComboboxControlProps is defined above: https://github.com/woocommerce/woocommerce/blob/0462094b58e3c994ab55ef5a5f7a5444a4439ff7/packages/js/product-editor/src/components/attribute-combobox-field/index.tsx#L32
If you change:
```
interface ComboboxControlProps
	extends Omit< typeof CoreComboboxControl, 'label' | 'help' > {
	__experimentalRenderItem?: ( args: {
		item: ComboboxControlOption;
	} ) => string | JSX.Element;
	className?: string;
}
```
to
```
interface ComboboxControlProps
	extends Omit< React.ComponentProps< typeof CoreComboboxControl >, 'label' | 'help' > {
	__experimentalRenderItem?: ( args: {
		item: ComboboxControlOption;
	} ) => string | JSX.Element;
	className?: string;
}
```

The `allowReset` is correctly set. I am not entirely sure how the others are working semi correctly 🤷‍♂️

gigitux: Thanks! Fixed with 0dc6c1c840f6d3e02cb111543008e9904e99029d

packages/js/product-editor/src/components/add-new-shipping-class-modal/add-new-shipping-class-modal.tsx
```diff
@@ -129,11 +125,13 @@ function ShippingClassForm( { onAdd, onCancel }: ShippingClassFormProps ) {
 			/>
 
 			<InputControl
-				{ ...getInputProps( 'slug' ) }
+				{ ...getInputProps< string >( 'slug' ) }
```
louwie17: Hmm this shouldn't be necessary with the TS updates in the Form component on `trunk`, but they don't seem to be reflected in this branch yet

packages/js/product-editor/src/components/custom-fields/create-modal/create-modal.tsx
```diff
@@ -236,10 +236,12 @@ export function CreateModal( {
...
-									onChange={ changeHandler(
-										customField,
-										'key'
-									) }
+									onChange={ ( value ) =>
+										changeHandler(
+											customField,
+											'key'
+										)( value )
+									}
```
louwie17: I don't think this change is necessary anymore after your `| undefined` addition in the `changeHandler` function.

packages/js/product-editor/src/components/product-section-layout/product-field-section.tsx
```diff
@@ -15,6 +15,7 @@ type ProductFieldSectionProps = {
 	title: string;
 	description: string | JSX.Element;
 	className?: string;
+	children: JSX.Element | JSX.Element[];
```
louwie17: While I don't overly mind this, should we use `React.ReactElement` instead?
Or instead of using `React.FC`, just setting the props using: `React.PropsWithChildren< ProductFieldSectionProps >`

packages/js/product-editor/src/components/product-section-layout/product-field-section.tsx
```diff
@@ -29,11 +30,13 @@ export const ProductFieldSection: React.FC< ProductFieldSectionProps > = ( {
...
-			</CardBody>
-		</Card>
+		<>
+			<Card>
+				<CardBody>
+					{ children }
+					<WooProductFieldItem.Slot name="" section={ id } />
+				</CardBody>
+			</Card>
+		</>
```
louwie17: With the update in `ProductSectionLayout` the use of `Fragment` isn't necessary anymore, we can remove that.

packages/js/product-editor/src/components/product-section-layout/product-section-layout.tsx
```diff
@@ -1,14 +1,20 @@
...
+} from '@wordpress/element';
 import { FormSection } from '@woocommerce/components';
 
-type ProductSectionLayoutProps = {
+interface ProductSectionLayoutProps {
 	title: string;
 	description: string | JSX.Element;
 	className?: string;
-};
+	children: Element[] | Element;
```
louwie17: Similar comment as above: https://github.com/woocommerce/woocommerce/pull/54312/files#r1910079027
Using `React.PropsWithChildren< ProductSectionLayoutProps >`?

Having said that, these components can also be removed as they are currently not used. But that can be a follow up.

packages/js/product-editor/src/components/radio-field/types.ts
```diff
@@ -3,8 +3,18 @@
...
+export type RadioFieldProps = Omit<
+	typeof RadioControl,
+	'label' | 'value' | 'options'
+> & {
 	title: string;
 	description?: string;
 	disabled?: boolean;
+	className?: string;
+	onChange( value: string ): void;
+	selected: string | undefined;
```
louwie17: Any particular reason you went with `| undefined` versus the optional `:?`?

packages/js/product-editor/src/components/require-password/require-password.tsx
```diff
@@ -67,7 +67,11 @@ export function RequirePassword( {
 					<InputControl
 						id={ postPasswordId }
 						value={ postPassword }
-						onChange={ onInputChange }
+						onChange={ ( value ) => {
+							if ( value ) {
+								onInputChange( value );
+							}
+						} }
```
louwie17:
```suggestion
						onChange={ ( value ) => {
							onInputChange( value ?? '' );
						} }
```
This breaks the `password` field, and actually cause the last character to always be present.
The above would be better.

packages/js/product-editor/src/components/schedule-publish-modal/schedule-publish-modal.tsx
```diff
@@ -67,7 +67,11 @@ export function SchedulePublishModal( {
 
 				<DateTimePicker
 					currentDate={ date }
-					onChange={ handleDateTimePickerChange }
+					onChange={ ( val ) => {
+						if ( val ) {
+							handleDateTimePickerChange( val );
+						}
+					} }
```
louwie17:
```suggestion
					onChange={ ( val ) => {
						handleDateTimePickerChange( val ?? '' );
					} }
```
I didn't explicitly test this one, but I believe the if statement would cause the same issue as the password field.

louwie17: Having said that, maybe we should update the `handleDateTimePickerChange` function instead.

packages/js/product-editor/src/components/tabs/tabs.tsx
```diff
@@ -102,10 +97,10 @@ export function Tabs( { selected, onChange }: TabsProps ) {
 		}
 	}
 
-	function renderFills( fills: readonly ( readonly ReactElement[] )[] ) {
+	function renderFills( fills: ReactNode ) {
 		return (
 			<TabFills
-				fills={ fills }
+				fills={ fills as readonly ( readonly ReactElement[] )[] }
```
louwie17, line 103: Hmm given this will likely be removed or altered, I am bound to let this slide.
But ideally we should update `TabFills` to also receive `ReactNode` and make use of `Children.toArray`.
Although the `sortFillsByOrder` is a pretty well duplicate of the one available in the components package: https://github.com/woocommerce/woocommerce/blob/d32cc4f22dac5020abd92765a0573253d35917ca/packages/js/components/src/utils.tsx#L93
So we can either mimic that or use that one.

Anyway, no hard feelings if you decide to ignore :)

packages/js/product-editor/src/components/text-control/text-control.tsx
```diff
@@ -46,7 +46,11 @@ export const TextControl = forwardRef( function ForwardedTextControl(
 			}
 			required={ required }
 			help={ error || help }
-			onChange={ onChange }
+			onChange={ ( value: string | undefined ) => {
+				if ( value && onChange ) {
+					onChange( value );
+				}
+			} }
```
louwie17:
```suggestion
			onChange={ ( value: string | undefined ) => {
				if ( onChange ) {
					onChange( value ?? '' );
				}
			} }
```
Same as password field. Also its weird to me that the `onChange` is optional here. Should we just make it required? or will that cause a entire other set of TS errors?

packages/js/product-editor/src/components/variations-table/variations-filter/variations-filter.tsx
```diff
@@ -181,6 +181,9 @@ export function VariationsFilter( {
 
 	const handleInputControlChange = useDebounce(
 		function handleInputControlChange( value: string | undefined ) {
+			if ( value === undefined ) {
+				return;
+			}
```
louwie17: I think we should use `value ?? ''` where `value` is used.

packages/js/product-editor/src/hooks/use-number-input-props.ts
```diff
@@ -64,7 +64,10 @@ export const useNumberInputProps = ( {
...
 			}
 		},
-		onChange( newValue: string ) {
+		onChange( newValue ) {
+			if ( ! newValue ) {
+				return;
+			}
 			let sanitizeValue = parseNumber(
 				newValue.replace( NOT_NUMBERS_OR_SEPARATORS_OR_MINUS_REGEX, '' )
 			);
```
louwie17:
```suggestion
		onChange( newValue ) {
			let sanitizeValue = parseNumber(
				( newValue ?? '').replace( NOT_NUMBERS_OR_SEPARATORS_OR_MINUS_REGEX, '' )
			);
```

louwie17 (changes requested):
@gigitux I left a bunch of comments ( sorry I did it as individual comments instead of them being part of a specific review 🤦‍♂️ ).
There are a couple breaking changes, so it be great to address them :)

packages/js/product-editor/src/components/add-new-shipping-class-modal/add-new-shipping-class-modal.tsx
gigitux: This branch is updated with trunk. I think that this happens because this logic is wrapped in another component. I believe that we should update the definition of `ShippingClassFormProps`, but I'm not sure that it makes sense at the current state. What do you think?

louwie17: Oooh wait, I see now, the `getInputProps` changes that I made are done when using the `Form` component and not the `useFormContext` hook. Which unfortunately points to another type.
See: https://github.com/woocommerce/woocommerce/blob/67cde87fd1cd67a08c11c16bbe7f0f552f1f59ed/packages/js/components/src/form/form-context.ts#L41 versus https://github.com/woocommerce/woocommerce/blob/0462094b58e3c994ab55ef5a5f7a5444a4439ff7/packages/js/components/src/form/form.tsx#L141

We should make use of the latter. If you want I can update this?

packages/js/product-editor/src/components/custom-fields/create-modal/create-modal.tsx
gigitux: Fixed with 06fb5308c899110d213a379c7f3034037e08c970

packages/js/product-editor/src/components/product-section-layout/product-field-section.tsx
gigitux: Fixed with 0822966434e46bd39dbbcddba65d7b2149c5092a

packages/js/product-editor/src/components/product-section-layout/product-field-section.tsx
gigitux: Fixed with 0822966434e46bd39dbbcddba65d7b2149c5092a

packages/js/product-editor/src/components/product-section-layout/product-section-layout.tsx
gigitux: Fixed with 0822966434e46bd39dbbcddba65d7b2149c5092a

packages/js/product-editor/src/components/radio-field/types.ts
gigitux: Good point. I fixed it with 074c3aa8a9fa9c7e24cdbcf33865803c912ce4c8

packages/js/product-editor/src/components/require-password/require-password.tsx
gigitux: Fixed with e1f833ffa51d6d0e7100adcce0347930fee5be5f

packages/js/product-editor/src/components/schedule-publish-modal/schedule-publish-modal.tsx
gigitux: Fixed with e1f833ffa51d6d0e7100adcce0347930fee5be5f

packages/js/product-editor/src/components/text-control/text-control.tsx
gigitux: Fixed with e1f833ffa51d6d0e7100adcce0347930fee5be5f

packages/js/product-editor/src/components/variations-table/variations-filter/variations-filter.tsx
gigitux: Fixed with e1f833ffa51d6d0e7100adcce0347930fee5be5f

packages/js/product-editor/src/hooks/use-number-input-props.ts
gigitux: Fixed with e1f833ffa51d6d0e7100adcce0347930fee5be5f

gigitux:
Thanks for your review! I addressed all your feedback!

packages/js/product-editor/src/components/tabs/tabs.tsx
gigitux: I'm fine with both solutions, at the same fine, I'm more inclined to keep the code as it is given that we already made multiple changes across the codebase (especially given that we will remove or change it)

packages/js/product-editor/src/components/add-new-shipping-class-modal/add-new-shipping-class-modal.tsx
```diff
@@ -131,9 +127,11 @@ function ShippingClassForm( { onAdd, onCancel }: ShippingClassFormProps ) {
...
 				{ ...getInputProps( 'slug' ) }
 				label={ __( 'Slug', 'woocommerce' ) }
-				onChange={ ( value: string ) => {
-					setPrevNameValue( '' ); // clean the previous name value.
-					getInputProps( 'slug' ).onChange( value );
+				onChange={ ( value ) => {
+					if ( value ) {
+						setPrevNameValue( '' ); // clean the previous name value.
+						getInputProps( 'slug' ).onChange( value );
+					}
```
louwie17:
```suggestion
				onChange={ ( value ) => {
					setPrevNameValue( '' ); // clean the previous name value.
					getInputProps( 'slug' ).onChange( value ?? '' );
				}
```

gigitux: Fixed with 0d1572fe56cf7f7d68cdf76d784aa094048b0834

louwie17 (approved):
Nice work!!! Looks good 👍

PR #54313, opened 2025-01-09
Improve the copy of the snack bar when one plugin is installed/activated

### LINKED ISSUE #53596: [Enhancement]: @woocommerce/data Improve plugins installed/activated snackbar notices copy

When using `installAndActivatePlugins` with only one plugin, the notice copy should use singular, not plural.

Instead of `Plugins were successfully installed and activated.` we should see `A plugin was successfully installed and activated.`
Even better would be use the plugin title (or slug if it is too complicated) in the copy - something like `WooPayments was successfully installed and activated.`

_No response_

### Additional context

_No response_

elazzabi:

Closes https://github.com/woocommerce/woocommerce/issues/53596.

Improve the notice copy when one plugin is installed, instead of multiple, by providing a singular and plural version.

**Pre-requisites for local testing:**

- Have woocommerce installed locally and synced within WooPayments. See Pe4Hk0-3f-p2 for details.
- Enable the feature flag to see reactified settings page. pe4Hk0-1ly-p2.
- Check out this branch.
- Run `pnpm i && pnpm run build`

How to test:
1. Navigate to WooCommerce > Settings > Payments
2. Find a plugin that is not yet installed. If it is already, deactivate & delete it.
3. Click on `Install`
4. Wait for it to finish and double-check the copy is `A plugin was successfully installed and activated` instead of `Plugins were successfully installed and activated.`

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

Improve the copy of the snack bar when one plugin is installed/activated

</details>

<details>

#### Comment 

</details>


elazzabi:
@vladolaru FYI, I looked at the available data and sadly, we only have access to the plugin slug (not the name). I think having the name would be better in the copy and the slug looks weird, so I didn't include it.

However, if the plan is to go with the slug, I can adjust the change.

dmallory42 (approved):
LGTM, clever solution! I didn't know about using `_n` to have a singular and plural variation. 

I agree we shouldn't show the slug in the message, if the title isn't possible then this message is fine IMO!

mordeth (approved):
Thank you for working on this, Ahmed!

The changes look great, and the manual tests have passed successfully. I also agree that the slug shouldn’t be included if the title is unavailable.

:shipit:

vladolaru:
Thank you for working on this @elazzabi! My instinct says we should bite the bullet and truly deliver a top-notch UX by making the plugin name available and included in the message. It adds a big boost in confidence about what is happening - this is the primary reason it is worth the trouble.

It is a bummer it is not available in the response, but we can make it available in a backward-compatible manner (avoid existing key collisions). Here is a suggestion patch I crafted as a starting point:

```
Index: plugins/woocommerce/src/Admin/API/Plugins.php
IDEA additional info:
<+>UTF-8
diff --git a/plugins/woocommerce/src/Admin/API/Plugins.php b/plugins/woocommerce/src/Admin/API/Plugins.php
--- a/plugins/woocommerce/src/Admin/API/Plugins.php	(revision 25c5d43becbe7784fae8b6df9f6b043728d2e5fe)
+++ b/plugins/woocommerce/src/Admin/API/Plugins.php	(date 1736428892508)
@@ -355,10 +355,29 @@
 
 		$data = PluginsHelper::activate_plugins( $plugins );
 
-		return( array(
+		// Gather some plugin details for each activated plugin.
+		$plugin_details = array();
+		if ( is_array( $data['activated'] ) ) {
+			foreach ( $data['activated'] as $plugin_slug ) {
+				$plugin_data = PluginsHelper::get_plugin_data( $plugin_slug );
+				if ( empty( $plugin_data ) ) {
+					continue;
+				}
+
+				$plugin_details[ $plugin_slug ] = array(
+					'name'        => $plugin_data['Name'],
+					'description' => $plugin_data['Description'],
+					'uri'         => $plugin_data['PluginURI'],
+					'version'     => $plugin_data['Version'],
+				);
+			}
+		}
+
+		return ( array(
 			'data'    => array(
-				'activated' => $data['activated'],
-				'active'    => $data['active'],
+				'activated'      => $data['activated'],
+				'active'         => $data['active'],
+				'plugin_details' => $plugin_details,
 			),
 			'errors'  => $data['errors'],
 			'success' => count( $data['errors']->errors ) === 0,

```

I included the details that would satisfy most needs (including ours). The same could be done for the [install_plugins](https://github.com/woocommerce/woocommerce/blob/25c5d43becbe7784fae8b6df9f6b043728d2e5fe/plugins/woocommerce/src/Admin/API/Plugins.php#L242-L242), although there it might be tricker because of edge-cases around the code being loaded or not just yet.

WDYT?

elazzabi:
@vladolaru sounds good! I'll try your patch and update the changes accordingly 👍

elazzabi:
https://github.com/user-attachments/assets/d19c96e9-363d-4a77-93a4-f421127a4e07

Thanks for the diff @vladolaru. I made some changes to include the name (and version) of the installed plugin. I also added the code to the `install_plugins` method, which will depend on the availability of the code installed, but I think we'll default to empty data, which is fine.

vladolaru:
Awesome, @elazzabi ! I like it a lot!

PR #54317, opened 2025-01-09
Decode error messages caused by the Add to Cart button block

Aljullu:

This PR makes it so error messages caused by the Add to Cart button block and returned by the server are decoded before being rendered. Fixing things like `&quot;` into `"`.

1. Load the `/shop` page of your store (assuming it uses the _Product Collection_ block).
2. Without closing that tab, go to the admin and modify a product which was in stock to make it so it's out of stock.
3. In the `/shop` tab, without reloading, try to add to the cart the product you just marked as out of stock.
4. Verify the error is rendered with the correct characters.

Before | After
--- | ---
imatge | imatge


imanish003 (approved):
@Aljullu I've reviewed the PR and the changes look good 🚀

PR #54331, opened 2025-01-09
Prepare WC Blocks data stores for TypeScript fixes

opr:

This PR:

- Removes the `@wordpress/data` type augmentation we added in the WC Blocks data stores:
	- `wc/store/cart`
	- `wc/store/checkout`
	- `wc/store/payment`
	- `wc/store/store-notices`
	- `wc/store/validation`

- Exports the store (`StoreDescriptor`) for each of these into `@woocommerce/block-data`.
- Removes the defined types of thunks & actions from the Cart, Checkout, and Payment data stores (these are no longer needed)
- Removes `mapped-types.ts` these are no longer needed since we get the types from `@wordpress/data` now.
- Created a new type in each store that uses thunks file called `{Store}ThunkArgs` which describes the select, dispatch, and registry args passed to each thunk.

The reason this PR is required is because of type updates made in `@wordpress/data` our types are now out of date (historically we were using types from DefinitelyTyped, but Gutenberg is now supplying types with their packages) and so we are getting TS errors when we try to use `useSelect` and `useDispatch`.

This PR does not fix any of those issues, but it does lay the groundwork for doing so. As a follow up, I suggest we go by data store key and update each use in the codebase. There should be at least 5 follow up PRs. It is possible that the effort will be split down even further.

This PR contains lint errors which have been ignored by 

1. Smoke test the shopper journey, adding items to cart, changing quantities, adding coupons, changing address, checking out etc.

### Dev testing

1. Check each of the changed files in your editor and ensure no TS errors relating to the changes appear.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce-blocks/assets/js/data/cart/index.ts
```diff
@@ -45,7 +43,8 @@ const store = createReduxStore( STORE_KEY, {
 			...( persistenceLayer.get() || {} ),
 		},
 	},
-} );
+};
+export const store = createReduxStore( STORE_KEY, config );
```
ralucaStan, line 47: should this be a default export? I'm thinking in the future this might be the only thing we export from this store. Applies to the other changes files from this PR.

plugins/woocommerce-blocks/assets/js/data/cart/index.ts
```diff
@@ -17,21 +17,19 @@ import * as selectors from './selectors';
...
-import { ResolveSelectFromMap } from '../mapped-types';
 import {
 	hasCartSession,
 	persistenceLayer,
 	isAddingToCart,
 } from './persistence-layer';
 import { defaultCartState } from './default-state';
 
-const store = createReduxStore( STORE_KEY, {
+export const config = {
```
ralucaStan, line 32: how is `config` used outside of this file?

plugins/woocommerce-blocks/assets/js/data/index.ts
```diff
@@ -6,13 +6,16 @@ import '@wordpress/notices';
 /**
  * Internal dependencies
  */
-export { CART_STORE_KEY } from './cart';
-export { CHECKOUT_STORE_KEY } from './checkout';
+export { CART_STORE_KEY, store as cartStore } from './cart';
```
ralucaStan, line 11: if we remove CART_STORE_KEY in the future we just import the default `CartStore` from `./cart`.

ralucaStan (approved):
I followed the testing steps and left some small comments. The PR looks good to me. Thank you for taking the first steps to fix the TS issues for the data store.

plugins/woocommerce-blocks/assets/js/data/cart/index.ts
opr: They are used in tests

plugins/woocommerce-blocks/assets/js/data/cart/index.ts
opr: ~Yeah, I think we could do that, let's see how it looks~

Actually, since we export `wcBlocksData` onto the window we still have to export the key, so we would be exporting the store description and the key. I don't think we should export a default.

plugins/woocommerce-blocks/assets/js/data/index.ts
opr: > if we remove CART_STORE_KEY in the future

we export `wcBlocksData` onto the window which contains this, it is exposed publicly, we would have to deprecate it (possible, but not in scope for this PR).

opr:
@samueljseay I requested a review from you as we looked at this problem this morning and I think it would be great to get your eyes on it, I think specifically these thunk arg types as I'm a tad unsure about them: https://github.com/woocommerce/woocommerce/pull/54331/files#diff-0203990004d095220d702753b81b8611b2908efff8507c4a869950720db5a945R36-R40

plugins/woocommerce-blocks/assets/js/data/index.ts
samueljseay: fwiw I much prefer named exports anyway :D they're easier for IDE auto import resolution to pick up.

samueljseay:
Just letting you know I won't have time to review this today, but will try for Monday. Just takes a little bit of time looking close at the types that you wanted me to look at. It's top of my todo list for next week.

plugins/woocommerce-blocks/assets/js/data/checkout/actions.ts
```diff
@@ -7,7 +7,6 @@ import { AdditionalValues } from '@woocommerce/settings';
  * Internal dependencies
  */
 import { ACTION_TYPES as types } from './action-types';
-import { ReturnOrGeneratorYieldUnion } from '../mapped-types';
 
 // `Thunks are functions that can be dispatched, similar to actions creators
```
nielslange:
```suggestion
// Thunks are functions that can be dispatched, similar to actions creators
```

Unrelated to your PR, but this backtick here looks accidental. There's another one in https://github.com/woocommerce/woocommerce/blob/785f1a482bd2ce437ed771cff24275c6613b9ace/plugins/woocommerce-blocks/assets/js/data/payment/actions.ts#L17.

nielslange (commented):
Thanks for working on this, @opr. I've successfully tested this PR based on your instructions. I did not spot TS errors as a result of your changes, and the smoke test went fine. That said, as @samueljseay is more skilled with TS, I'd leave the final approval of this PR with him.

Below, you can find two minor nitpicks, I noticed. Not related to your PR, though.

samueljseay (approved):
Tested and looks good but furthermore I dug into the thunk types and i think you've used the right ones here. It looks great. I really appreciate you paying down the technical debt on this. Great work 🎉

PR #54340, opened 2025-01-09
Fix cart and checkout page detection logic for edge cases

### LINKED ISSUE #53592: Cart shortcode disappears if its classic shortcode block, and no Cart page is set
Cart shortcode would vanish if it's not the default cart, you have no default cart, and are using classic shortcode block.

### Steps to reproduce:

1. Remove the cart page from WooCommerce -> Advanced settings.
2. Change the cart block to Classic Cart block.
3. On Cart page, hit apply coupon
4. Notice that the cart shortcode disappears.

### Expected behavior
Cart should persist.

### Current behavior
Cart is broken

### LINKED ISSUE #53628: Cart and Checkout content information on status page doesn't check for Classic Block type
### Prerequisites

### Describe the bug

On WooCommerce > Status, on Cart and Checkout entries in the WooCommerce pages section, the info is based solely on the presence of Classic Block when evaluating if shortcode is present. Classic Block can be Cart or Checkout.

### Expected behavior

It should state "Page does not contain..." if the wrong Classic Block is being used, eg Cart Classic Block on Checkout page.

### Actual behavior

It states the shortcode version is present, when the wrong classic block type is on the page.

### Steps to reproduce

1. Clear the contents of the cart page
2. Add the Classic Checkout to the page
3. Go to WooCommerce > Status, and check that in the WooCommerce pages section, the Cart page states the Cart block is being used:

### WordPress Environment

n/a

### Isolating the problem

mikejolley:

This PR fixes 2 similar issues relating to cart and checkout page detection, namely:

1. The cart page not being correctly detected resulting in page content disappearing after AJAX requests.
2. The system status report not correctly reporting mistakes if the wrong shortcode/block is used on a cart or checkout page.

Screenshot 2025-01-09 at 15 27 05

Fixes #53592
Fixes #53628

1. Remove the cart page from WooCommerce -> Advanced settings.
2. Change the cart block to Classic Cart block.
3. On Cart page, hit apply coupon
4. The page content should refresh but should not disappear.

To test the status page,

1. To ease testing, open up a tab with WooCommerce > Settings > Advanced, WooCommerce > Status, and the cart page open in the editor
2. Ensure cart and checkout pages are set under **advanced**. You should see the status report is happy with this.
3. Edit the cart page and use a cart block. Check the status report states `Contains the woocommerce/cart block`
4. Edit the cart page and convert the cart block into a classic shortcode block (via the transform menu). Check the status report states `Contains the woocommerce/classic-shortcode block`
5. Edit the cart page and go into the HTML editor. Change the block markup and set the "shortcode" attribute on the classic shortcode block manually, editing it to read "checkout" instead of "cart". Save. Check the status report indicates an error on this page due to missing shortcode/block.
6.  Edit the cart page and delete all blocks. Insert a "shortcode" block and manually input `[woocommerce_cart]`. Check that the status page indicates you're using the shortcode.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/fix-53592-cart-shortcode-detection
+++ b/plugins/woocommerce/changelog/fix-53592-cart-shortcode-detection
@@ -1,0 +1,4 @@
+Significance: minor
+Type: update
+
+Fix cart and checkout block/shortcode detection in status report and frontend.
--- a/plugins/woocommerce/src/Blocks/Utils/CartCheckoutUtils.php
+++ b/plugins/woocommerce/src/Blocks/Utils/CartCheckoutUtils.php
@@ -5,6 +5,79 @@ namespace Automattic\WooCommerce\Blocks\Utils;
  * Class containing utility methods for dealing with the Cart and Checkout blocks.
  */
 class CartCheckoutUtils {
+	/**
+	 * Returns true if:
+	 * - The cart page is being viewed.
+	 * - The page contains a cart block, cart shortcode or classic shortcode block with the cart attribute.
+	 *
+	 * @return bool
+	 */
+	public static function is_cart_page() {
+		global $post;
+
+		$page_id      = wc_get_page_id( 'cart' );
+		$is_cart_page = $page_id && is_page( $page_id );
+
+		if ( $is_cart_page ) {
+			return true;
+		}
+
+		// Check page contents for block/shortcode.
+		return is_a( $post, 'WP_Post' ) && ( wc_post_content_has_shortcode( 'woocommerce_cart' ) || self::has_block_variation( 'woocommerce/classic-shortcode', 'shortcode', 'cart' ) );
+	}
+
+	/**
+	 * Returns true if:
+	 * - The checkout page is being viewed.
+	 * - The page contains a checkout block, checkout shortcode or classic shortcode block with the checkout attribute.
+	 *
+	 * @return bool
+	 */
+	public static function is_checkout_page() {
+		global $post;
+
+		$page_id          = wc_get_page_id( 'checkout' );
+		$is_checkout_page = $page_id && is_page( $page_id );
+
+		if ( $is_checkout_page ) {
+			return true;
+		}
+
+		// Check page contents for block/shortcode.
+		return is_a( $post, 'WP_Post' ) && ( wc_post_content_has_shortcode( 'woocommerce_checkout' ) || self::has_block_variation( 'woocommerce/classic-shortcode', 'shortcode', 'checkout' ) );
+	}
+
+	/**
+	 * Check if the current post has a block with a specific attribute value.
+	 *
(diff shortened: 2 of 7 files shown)


senadir (approved):
Feels like this might start surfacing some weird things, I know people were using `is_checkout` and `is_cart` and possibly not expecting it to work with blocks? I'm not saying we change anything, but if a bug arises, this is a good start to point people to.

We might want to tweak the changelog to indicate that this includes `is_cart` and `is_checkout` functions.

plugins/woocommerce/src/Blocks/Utils/CartCheckoutUtils.php
```diff
@@ -5,6 +5,77 @@
...
+		$page_id          = wc_get_page_id( 'checkout' );
+		$is_checkout_page = $page_id && is_page( $page_id );
+
+		if ( $is_checkout_page ) {
+			return true;
+		}
+
+		// Check page contents for block/shortcode.
+		return is_a( $post, 'WP_Post' ) && ( wc_post_content_has_shortcode( 'woocommerce_checkout' ) || self::has_block_variation( 'woocommerce/classic-shortcode', 'shortcode', 'checkout', $post->post_content ) );
+	}
```
senadir, line 48: Where is the check for the checkout/cart blocks?

plugins/woocommerce/changelog/fix-53592-cart-shortcode-detection
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: update
+
+Fix cart and checkout block/shortcode detection in status report and frontend.
```
senadir, line 4: We might want to change this
```suggestion
Tweak is_cart and is_checkout functions to always account for block/shortcode versions of cart/checkout.
```

mikejolley:
@senadir R/e your feedback,

> I know people were using is_checkout and is_cart and possibly not expecting it to work with blocks?

The behaviour hasn't changed. These functions only return true for the assigned cart/checkout pages and those using shortcode. The "block" check is for the shortcode served via block.

> Where is the check for the checkout/cart blocks?

There isn't one. And changing this would make your first point true.

I think you may have confused the classic shortcode block with the actual cart/checkout blocks.

senadir:
Good point Mike, I assumed `CartCheckoutUtils::is_cart_page` would also check for the block.

mikejolley:
@senadir do you think it should or do we leave it as before? I think we'd ideally want to map only to the assigned pages; the shortcode handling is there from legacy times to avoid script breakage. Blocks don't break in the same way.

senadir:
Agree, let's not add the block, `is_cart_page` should only the actual page. Let's keep it as is

PR #54342, opened 2025-01-09
[e2e tests] Clean up devices in Playwright configs

adimoldovan:

The devices definitions in Playwright configs are not needed, we are not using multiple devices. 
Cleaned up the config files.

The change should not affect any e2e test, they should all pass as before.


lanej0 (approved):
Looks good.  I think that was leftover from the original setup, and you're right -- was never used.

👍🏼

PR #54344, opened 2025-01-09
Remove legacy Puppeteer e2e tests

### LINKED ISSUE #40119: Remove WooCommerce Core's Legacy Puppeteer Tests
These tests are no longer used, and as such, are not needed any longer. We should remove them and rename the `e2e-pw` directory to just `e2e`. There are some potential places we need to update this directory though as they may be depending on that path. One example being the QIT.

adimoldovan:

Closes https://github.com/woocommerce/woocommerce/issues/40119

Removed the old and unused e2e tests (e2e folder) and all related dependencies and scripts.
The js packages which were in use by the tests are not in scope of this PR, we'll follow-up on that.

CI is green


lanej0 (approved):
Ran the test suite on trunk and on this PR.  No difference in test results, so looks good.

👍🏼

PR #51185, opened 2024-09-06
Change event to document.body for dynamically added elements

Lenny4:
Change event to document.body to allow third party plugin to add, remove, update elements. Because this approach works for dynamically added elements.

### Submission Review Guidelines:

-   I have followed the [WooCommerce Contributing Guidelines](https://github.com/woocommerce/woocommerce/blob/trunk/.github/CONTRIBUTING.md) and the [WordPress Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/).
-   I have checked to ensure there aren't other open [Pull Requests](https://github.com/woocommerce/woocommerce/pulls) for the same update/change.
-   I have reviewed my code for [security best practices](https://developer.wordpress.org/apis/security/).
-   Following the above guidelines will result in quick merges and clear and detailed feedback when appropriate.

### Changes proposed in this Pull Request:

These changes allow a third party plugin to modify the content of the order details in javascript without having to "recall" all the javascript event linked to these elements.

Even in the future if WooCommerce starts to modify the content of `#woocommerce-order-data` this will be needed.

### How to test the changes in this Pull Request:

1. Before checking out this branch, verify the current behavior in trunk.
1. Install this snippet in your test site: https://gist.github.com/Lenny4/448cc3894b940d3877f02754a9893cdf
2. In WP Admin, navigate to the screen to add a new order. You'll see two buttons at the top of the screen added by the snippet.
3. Click the pencil icon to edit the Billing Address for the order. It should expand to a set of fields. Then click the first "Run Test" button at the top of the screen. This will regenerate the Billing Address section, so the fields are collapsed again. Now if you click the pencil icon again, nothing will happen, because the set of fields is no longer initialized.
4. Try the same with the second "Run Test" button, and you will get the same result.
5. Now check out this branch. You may need to rebuild the admin scripts with `pnpm --filter='@woocommerce/plugin-woocommerce' build`
6. Do the same procedure with the pencil icon and the two "Run Test" buttons. The first button should still result in an unresponsive pencil icon. But the second button should result in being able to re-expand the set of billing address fields.

### Changelog entry

-   [ ] Automatically create a changelog entry from the details below.

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Details</summary>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

-   [ ] Fix - Fixes an existing bug
-   [ ] Add - Adds functionality
[...]

PR #54357, opened 2025-01-10
Reduce scope of globalized order metabox functions

coreymckrill:

In #51185 we exported the order metabox scripts outside of their encapsulation in order to let 3rd party scripts re-initialize UI elements if they were modified. However, this also opened up many other methods to the global scope that don't need to be global. This amends that change so that only the init methods are available globally.

Follow the testing instructions in #51185 and make sure they still work, except use [this modified version](https://gist.github.com/coreymckrill/eae1fcacc3c2c9291f3d22ad875afb48) of the test script.


jorgeatorres (approved):

coreymckrill:
Closing in favor of #54390

PR #54376, opened 2025-01-10
Update Store Notices data store use to use `StoreDescriptor` instead of key

opr:

Updates uses of `wc/store/store-notices` to use the StoreDescriptor instead of store name, follow up to (and based on) https://github.com/woocommerce/woocommerce/pull/54331

1. Install Stripe
2. Go to the Cart block, open your JS console and type `wp.data.dispatch( wp.notices.store ).createInfoNotice( 'This is a notice', { context: 'wc/cart' } )` Expect to see a notice in the cart.
3. Go to the Checkout block, repeat step 2 (**use `wc/cart`** not checkout)
4. Reload and type `wp.data.dispatch( wp.notices.store ).createInfoNotice( 'This is a notice', { context: 'wc/checkout' } )` and `wp.data.dispatch( wp.notices.store ).createInfoNotice( 'This is a notice', { context: 'wc/checkout/shipping-methods' } )`
5. Ensure you see a notice at the top of the cart and one in the shipping methods.
6. Type `wp.data.select( wc.wcBlocksData.storeNoticesStore ).getRegisteredContainers()` and then try adding a notice to each of these registered containers. Note, there is a bug with the `woocommerce/checkout-totals-block` container, so don't try that one (PR: will fix this)
7. Try using one of Stripe's cards that fails (https://docs.stripe.com/testing#declined-payments) - ensure a notice appears (this will appear at the top level).

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce-blocks/packages/components/store-notices-container/index.tsx
+++ b/plugins/woocommerce-blocks/packages/components/store-notices-container/index.tsx
@@ -2,10 +2,7 @@
  * External dependencies
  */
 import { useSelect, useDispatch } from '@wordpress/data';
-import {
-	PAYMENT_STORE_KEY,
-	STORE_NOTICES_STORE_KEY,
-} from '@woocommerce/block-data';
+import { PAYMENT_STORE_KEY, storeNoticesStore } from '@woocommerce/block-data';
 import { getNoticeContexts } from '@woocommerce/base-utils';
 import type { Notice } from '@wordpress/notices';
 import { useMemo, useEffect } from '@wordpress/element';
@@ -31,16 +28,14 @@ const StoreNoticesContainer = ( {
 	context = '',
 	additionalNotices = [],
 }: StoreNoticesContainerProps ): JSX.Element | null => {
-	const { registerContainer, unregisterContainer } = useDispatch(
-		STORE_NOTICES_STORE_KEY
-	);
+	const { registerContainer, unregisterContainer } =
+		useDispatch( storeNoticesStore );
 	const { suppressNotices, registeredContainers } = useSelect(
 		( select ) => ( {
 			suppressNotices:
 				select( PAYMENT_STORE_KEY ).isExpressPaymentMethodActive(),
-			registeredContainers: select(
-				STORE_NOTICES_STORE_KEY
-			).getRegisteredContainers(),
+			registeredContainers:
+				select( storeNoticesStore ).getRegisteredContainers(),
 		} )
 	);
 	const contexts = useMemo< string[] >(
--- a/plugins/woocommerce-blocks/docs/third-party-developers/extensibility/data-store/store-notices.md
+++ b/plugins/woocommerce-blocks/docs/third-party-developers/extensibility/data-store/store-notices.md
@@ -77,7 +77,9 @@ This action will register a new container.
 #### _Example_ <!-- omit in toc -->
 
 ```javascript
-dispatch( registerContainer( 'someContainerContext' ) );
+import { storeNoticesStore } from '@woocommerce/block-data';
+
+dispatch( storeNoticesStore ).registerContainer( 'someContainerContext' );
 ```
 
 ### unregisterContainer( containerContext )
@@ -97,7 +99,9 @@ This action will unregister an existing container.
 #### _Example_ <!-- omit in toc -->
 
 ```js
-dispatch( unregisterContainer( 'someContainerContext' ) );
+import { storeNoticesStore } from '@woocommerce/block-data';
+
+dispatch( storeNoticesStore ).unregisterContainer( 'someContainerContext' );
 ```
 
 ## Selectors
--- a/plugins/woocommerce-blocks/docs/third-party-developers/extensibility/data-store/store-notices.md
+++ b/plugins/woocommerce-blocks/docs/third-party-developers/extensibility/data-store/store-notices.md
@@ -17,10 +17,16 @@ The Store Notices Store allows to register and unregister containers for notices
 
 ## Usage
 
-To utilize this store you will import the `STORE_NOTICES_STORE_KEY` in any module referencing it. Assuming `@woocommerce/block-data` is registered as an external pointing to `wc.wcBlocksData` you can import the key via:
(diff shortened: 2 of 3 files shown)


plugins/woocommerce-blocks/assets/js/data/checkout/reducers.ts
```diff
@@ -1,12 +1,18 @@
+/**
+ * External dependencies
+ */
+import type { Reducer } from 'redux';
```
alexflorisca: Should we be importing things straight from redux? I thought the point of wp.data is an abstraction on top, so if they ever want to swap it out for something other than redux, code that depends on it shouldn't break

plugins/woocommerce-blocks/docs/third-party-developers/extensibility/data-store/README.md
```diff
@@ -55,15 +55,17 @@ const { SCHEMA_STORE_KEY } = window.wc.wcBlocksData;
 The **[Store Notices Store (wc/store/store-notices)](store-notices.md)** is dedicated to handling various store notices and alerts. To access this store using a constant, you can use:
 
 ```ts
-const { STORE_NOTICES_STORE_KEY } = window.wc.wcBlocksData;
+const { storeNoticesStore } = window.wc.wcBlocksData;
```
alexflorisca: Should we rename this to `noticesStore` instead? Bit of a mouthful otherwise

plugins/woocommerce-blocks/docs/third-party-developers/extensibility/data-store/README.md
```diff
@@ -55,15 +55,17 @@ const { SCHEMA_STORE_KEY } = window.wc.wcBlocksData;
...
 ```
 
+This `StoreDescriptor` can be passed in the `mapSelect` function of `useSelect` and in `useDispatch`.
```
alexflorisca: if we're explaining this here, should we do this for every other store as well?

plugins/woocommerce-blocks/packages/components/store-notices-container/index.tsx
```diff
@@ -2,10 +2,7 @@
  * External dependencies
  */
 import { useSelect, useDispatch } from '@wordpress/data';
-import {
-	PAYMENT_STORE_KEY,
-	STORE_NOTICES_STORE_KEY,
-} from '@woocommerce/block-data';
+import { PAYMENT_STORE_KEY, storeNoticesStore } from '@woocommerce/block-data';
```
alexflorisca, line 5: There's a select in this file that still uses `PAYMENT_STORE_KEY` instead of `storeNoticesStore` descriptor

plugins/woocommerce/changelog/fix-data-store-types-2
alexflorisca, line 1: I think this file is left over from previous PR - not sure why it's being added here again?

alexflorisca (commented):
Nice job, love that you've removed almost 200 lines of code, most of which were types and what not! Left a few inline comments, but mostly looks good 👍

plugins/woocommerce-blocks/docs/third-party-developers/extensibility/data-store/README.md
alexflorisca: Ah I realise that you've splitting this into multiple PRs, so ignore the above

plugins/woocommerce-blocks/docs/third-party-developers/extensibility/data-store/README.md
opr: While it is a little longer, it correlates directly with the name of the store: `wc/store/store-notices` I think I'd prefer to keep this direct reference.

plugins/woocommerce/changelog/fix-data-store-types-2
opr: Rebased and it is gone ✨

plugins/woocommerce-blocks/docs/third-party-developers/extensibility/data-store/README.md
alexflorisca: What about renaming the store to `notices` to better fit with the other store names?

opr: It's possible, but I don't think it's a necessary change. The purpose of this store is to work with `StoreNoticesContainer`. In my opinion the name is useful and specific.

opr: Also, it's unlikely that anyone is using it, but this data store is technically public API so renaming it would break that.

alexflorisca: Ok I'm convinced

opr: Thanks mate, if you're happy with the PR please could you approve it 🙇🏼

plugins/woocommerce-blocks/assets/js/data/checkout/reducers.ts
opr: This was removed from the diff after the rebase, but `@wordpress/data` doesn't seem to export any alternative type, for now I think the type def import from redux is fine, as we're not actually using any code.

alexflorisca:
I feel like there were a lot more than 3 files changed before you force pushed, can you just double check that all the work is here?

plugins/woocommerce-blocks/docs/third-party-developers/extensibility/data-store/README.md
alexflorisca: Yeah no worries, just see my [other comment](https://github.com/woocommerce/woocommerce/pull/54376#issuecomment-2598252693). In future, would be helpful if you can re-request the review so it pops up on my notifications when you're ready 👍

opr:
> I feel like there were a lot more than 3 files changed before you force pushed, can you just double check that all the work is here?

Yes this was because it was based on https://github.com/woocommerce/woocommerce/pull/54331 which is now merged

alexflorisca (approved):
This looks good now 👍

PR #54380, opened 2025-01-10
Prevent a "Maximum update depth exceeded" error when adding notices to the `woocommerce/checkout-totals-block` context

opr:

When adding to the `woocommerce/checkout-totals-block` store notices context, an infinite loop occurs due to a logic bug. Previously, the code would try to remove the only notice we add to this context. It did this by removing the notice with a specific ID. If a notice was added with a different ID, then the `removeNotice` call wouldn't have any effect, but would still trigger the hook to re-run. 

This PR adds a check for the presence of the notice by ID before trying to remove it.

In this PR I also updated the version of `@wordpress/notices` so we can benefit from the types.

1. Add an item to your cart and go to the Checkout block.
2. Open the JS console and type: `wp.data.dispatch( wp.notices.store ).createInfoNotice( 'This is a notice', { context: 'woocommerce/checkout-totals-block' } )`
3. Check the notice appears in the sidebar and you don't get any React/JS errors. 

**Regression testing (Steps from #49718):**

1. Setup some shipping methods and some local pickup locations under WooCommerce > Settings > Shipping. Additionally, create a zone for the US with no shipping methods. We'll use this later.
2. Go to checkout.
3. Using a non-US address,Toggle between local pickup and shipping. Ensure when you toggle, the first shipping rate or pickup location is automatically selected and totals are updated as expected. Also ensure that the notice "Totals will be recalculated when a valid shipping method is selected." does not become visible.
4. Switch to shipping and enter a US address (ensure you select a state also in the form as there is an unrelated issue with the select boxes preventing default state from being selected).
5. Check that the "Totals will be recalculated when a valid shipping method is selected." method appears and there are no shipping methods available.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


wavvves (commented):
This seems ok Thomas, but trying to replicate on trunk doesn't happen anymore, ~and there are some conflicting change although Github doesn't complain 🤔~
~I tried merging, but it seems that lock files need rebuilding and I think you should be the one making sure there's nothing missing in the end. Re-ping me for review please~

Just noticed this isn't being merged to trunk. Heads up there's also a pnpm bump coming from there.

So since this works out here, I'll approve, and then resolving the trunk conflicts on [fix/store-notices-data-store](https://github.com/woocommerce/woocommerce/tree/fix/store-notices-data-store) seems the best option 👍🏼

wavvves (approved):
Approved, see comment above ⬆️

opr:
Thanks Paulo! I will merge this after #54376

Issue #54381, opened 2025-01-10
Available block context does not flow to blocks inside the Product Template block


helgatheviking:
### Prerequisites

- [x] I have carried out troubleshooting steps and I believe I have found a bug.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

Any block context made available on the Product Template block (or ideally the Product Collection block) cannot be inherited by the nested blocks such as Post Title and Product Image because of how the Product Template renders each product without passing the currently available context to the `new WP_Block` instance [here](https://github.com/woocommerce/woocommerce/blob/3dd035917b982093ac15f3472d7bc7df0e56b94a/plugins/woocommerce/src/Blocks/BlockTypes/ProductTemplate.php#L96-L104). Note, Woo is doing this slightly differently than core does the Post Template block, but with the same result. 

I [reported this to WP core](https://github.com/WordPress/gutenberg/issues/68608) as well.

### Expected behavior

The custom context key is defined and so then we can display the value next to the title "we love tacos"
![Image](https://github.com/user-attachments/assets/4d68da94-17c3-4cd0-a38f-10cdaf596609)

### Actual behavior

The custom context key is not defined so the undefined value is displayed "no tacos"

![Image](https://github.com/user-attachments/assets/797f3582-8dba-424d-ac77-0bfc0a9f7655)

### Steps to reproduce

1. Activate snippet
```

add_filter( 'render_block_context', function( $context, $parsed_block ) {
    if ( 'woocommerce/product-template' === $parsed_block['blockName'] ) {
        $context['helga/taco'] = 'we love tacos';
    }

    return $context;
}, 10, 2 );
 
add_filter( 'block_type_metadata', function( $metadata ) {
    if ( 'core/post-title' === $metadata['name'] ) {
        $metadata['usesContext'] ??= [];
        $metadata['usesContext'][] = 'helga/taco';
    }
    return $metadata;
} );
 
add_action( 'render_block_core/post-title', function( $block_content, $block, $instance ) {
    
	if ( ! isset( $instance->context ) ) {
        return $block_content;
    }

    $taco = $instance->context['helga/taco'] ??= 'no tacos';

    return $taco . ' ' . $block_content;

}, 10, 3 );
```

2. insert any product collection and view on the front end

### WordPress Environment

Woo 9.7.0-dev

### Isolating the problem

- [x] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
[...]

PR #54382, opened 2025-01-10
Relay the block context from Product Template to the inner blocks

helgatheviking:

Closes ##54381 .

Follow steps outlined in #54381.
with patch applied you should see the successful result of "We love tacos" appearing before every title in the product collection.

1. Activate snippet
```

add_filter( 'render_block_context', function( $context, $parsed_block ) {
    if ( 'woocommerce/product-template' === $parsed_block['blockName'] ) {
        $context['helga/taco'] = 'we love tacos';
    }

    return $context;
}, 10, 2 );
 
add_filter( 'block_type_metadata', function( $metadata ) {
    if ( 'core/post-title' === $metadata['name'] ) {
        $metadata['usesContext'] ??= [];
        $metadata['usesContext'][] = 'helga/taco';
    }
    return $metadata;
} );
 
add_action( 'render_block_core/post-title', function( $block_content, $block, $instance ) {
    
	if ( ! isset( $instance->context ) ) {
        return $block_content;
    }

    $taco = $instance->context['helga/taco'] ??= 'no tacos';

    return $taco . ' ' . $block_content;

}, 10, 3 );
```

2. insert any product collection and view on the front end.
3. Ensure that `we love tacos` string is visible.

| Before | After |
|--------|--------|
| image | image | 

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

Relay the block context from Product Template to the inner blocks

#### Comment 

</details>


nielslange:
I've removed @Konamiman and myself as reviewers since this functionality is maintained by @woocommerce/woo-fse. Therefore, @roykho is the appropriate reviewer.

helgatheviking:
I have been thinking the _complete_ solution might be `WP_Block_List` but this does kinda work... or at least it relays anything from the _product template_ block down. Context on the collection block would not flow automatically however.

Reminder that I opened a similar [issue upstream](https://github.com/WordPress/gutenberg/issues/68608)

plugins/woocommerce/src/Blocks/BlockTypes/ProductTemplate.php
```diff
@@ -91,15 +91,21 @@ protected function render( $attributes, $content, $block ) {
 			// This ensures that for the inner instances of the Post Template block, we do not render any block supports.
 			$block_instance['blockName'] = 'core/null';
 
+			// Relay the block context to the inner blocks.
+			$block_context = array_merge(
```
imanish003: Nitpick: I think it would be better to rename `$block_context` to `$available_context` as 2nd argument of `WP_Block` constructor is called `available_context`. What do you think?
```suggestion
			$available_context = array_merge(
```

imanish003 (dismissed):
@helgatheviking Thank you for the PR. I've tested the changes and they work as expected. While there is one technical consideration to note, I believe we can proceed with merging this PR.

**Technical Consideration**

The second argument of the `WP_Block` constructor is `available_context`, not `context`. Ideally, we should pass all available context from parent blocks, rather than just the Product Template block's context. For instance, if the parent "Product Collection" block has additional context, inner blocks currently won't be able to access it since it's not included in the `available_context` as 2nd argument of `WP_Block` constructor. However, given the current architecture, IMO this is an acceptable limitation. I can't think of a better solution for this. 🤔

For example, ideally we would prefer passing `$block->available_context` as 2nd argument of `WP_Block` constructor i.e.

```php
$block_content = (
	new WP_Block(
		$block_instance,
		$block->available_context
	)
)->render( array( 'dynamic' => false ) );
```

But as it's protected property, we can't access it directly. We may consider using Reflection API, but IMO it would be hacky solution. Probably, going with proposed solution in this PR is the best solution for now. Also, you mentioned `WP_Block_List` but I am not sure how that would help us with this limitation. 🤔

helgatheviking:
>The second argument of the WP_Block constructor is available_context, not context. Ideally, we should pass all available context from parent blocks, rather than just the Product Template block's context. 

I agree and that's what I was hoping to be able to do, but as you noted, that's a private property and cannot be accessed. This does mean that the context has to be on the Product Template block itself in order to get inherited by child blocks, but context on say the Product Collection will not automatically flow through. 

It's still a limitation, but an improvement from before.

I'm not really certain how  `WP_Block_List` would be used either, I just noticed that's what core WP does when trying to render inner blocks.

plugins/woocommerce/src/Blocks/BlockTypes/ProductTemplate.php
```diff
@@ -91,15 +91,21 @@ protected function render( $attributes, $content, $block ) {
...
 			// Render the inner blocks of the Post Template block with `dynamic` set to `false` to prevent calling
 			// `render_callback` and ensure that no wrapper markup is included.
 			$block_content = (
 				new WP_Block(
 					$block_instance,
-					array(
-						'postType' => get_post_type(),
-						'postId'   => $product_id,
-					)
+					$block_context
```
imanish003: Can you please also update the name here? 

```suggestion
					$available_context
```

imanish003:
> It's still a limitation, but an improvement from before.

+1. This PR looks good to merge after addressing the minor variable renaming suggestion in my comment above.

gigitux (approved):
@helgatheviking Thank you for your contribution 🙇. I updated the PR with the improved name based on @imanish00’s suggestion. Additionally, I revised the PR description to include testing instructions and added the changelog using the `pnpm --filter='@woocommerce/block-library' changelog add` command.

<hr />


I followed the testing instructions and I'm able to see the `we love tacos` in the Product Collection block:


image

:shipit:

PR #54384, opened 2025-01-10
Enable DB auto-update by default

### LINKED ISSUE #54250: [Spike] Enabling of DB auto-updates
Though we've supported DB auto-updates through the `woocommerce_enable_auto_update_db` filter, it's currently set to `false` by default. We aim to enable it by default, but have to make sure that there are no unwanted side effects as a result (for example, for CLI users, etc.) or that it is easy to disable again.

Ideally, this would be the new default behavior:

> - Set database updates to happen automatically after the plugin is updated - this could be as simple as changing the default woocommerce_enable_auto_update_db value to true
> - Ensure the option to make it manual still remains
> - Potentially add a PHP constant for this setting (this was suggested as preferred over a filter, though I propose both).
> - Ensure user is redirected to where they were if they click the update database button.
>
> Scenarios:
> - Auto updates on, database set to automatic: WooCommerce updates automatically and database updates are scheduled behind the scenes.
> - Auto updates on, database updates set to manual: Same as now, WooCommerce updates automatically but user has to confirm via the notice to update the database.
> - Auto updates off, database updates automatic: User has to manually update WooCommerce, the database is also updated as part of that process automatically.
> - Auto updates off, database updates manual: User manually updates WooCommerce, and manually triggers database updates if required.

jorgeatorres:

This PR implements database auto-updating by default. This implies that upgrade routines included in any version of WC will be scheduled automatically without admin intervention, something that was only possible by overriding the default behavior via the `woocommerce_enable_auto_update_db` filter. See #54268 for some discussion around this.

Summary of the changes introduced in this PR:

- Filter `woocommerce_enable_auto_update_db` can still be used to prevent auto-db updates by ensuring its result is `false`.
- When auto updates are enabled (the new default), the filter `woocommerce_db_update_schedule_delay` can be used to add a delay (in seconds) to the time the updates are scheduled to run. This delay is, by default, `0`, which means updates will run as soon as possible (likely during the next page load).
   Users/hosts can use this filter to make sure updates are scheduled in the future.
- Logging has been added to various parts of the update process, both for manual or auto updates. Log file `wc-updater` will be written to with entries indicating the time the updates are scheduled to run as well as details on the "version" jump:

   ```
   Automatic database update triggered.
   Scheduling database updates (from 9.2.0 to 9.7.0)...
       Updates will begin running in approximately 9 minutes.
       Updates from version 9.3.0 scheduled.
       Updates from version 9.4.0 scheduled.
       Updates from version 9.5.0 scheduled.
   Database updates scheduled.
   ```

Closes #54250.

#### Pre-requisites
1. Drop the following PHP snippet as a .php file inside `wp-content/mu-plugins`. This will disable Action Scheduler from automatically running updates, allowing us to check that they are indeed being scheduled:
   ```php
   <?php
   add_filter( 'action_scheduler_allow_async_request_runner', '__return_false' );
   add_action( 'init', fn () => remove_action( ActionScheduler_QueueRunner::WP_CRON_HOOK, [ ActionScheduler_QueueRunner::instance(), 'run' ] ) );
   ```

#### Test the new default behavior
1. Run the following WP-CLI commands to "simulate" WC being updated from 9.2.0 to the current version:
   ```
   wp transient delete wc_installing
   wp option set woocommerce_db_version 9.2.0
   wp option set woocommerce_version 9.2.0
   ```
1. Visit any page on the admin.
1. Run the following command to check that the updates with group `woocommerce-db-updates` were scheduled without user intervention:
   ```
   wp action-scheduler action list --group=woocommerce-db-updates --status=pending --per_page=0
   ```
   Alternatively, check this in WC > Status > Scheduled Actions > Pending instead.
1. Confirm also that these actions were scheduled for as soon as possible (i.e. current time or slightly i the past).
1. Go to WC > Status > Logs and select the `wc-updater` log file.
1. Confirm that the log file includes the following entries:
   ```
   Automatic database update triggered.
   Scheduling database updates (from 9.2.0 to 9.7.0)...
      Updates will begin running in approximately 9 minutes.
      Updates from version 9.3.0 scheduled.
      Updates from version 9.4.0 scheduled.
      Updates from version 9.5.0 scheduled.
   Database updates scheduled.
   ```
1. Run all the pending migration routines at once by running the following command:
   ```
   wp action-scheduler action run $(wp action-scheduler action list --group=woocommerce-db-updates --status=pending --format=ids --per_page=0)
   ```
1. Confirm that there are no more pending actions in the group `woocommerce-db-updates` by running `wp action-scheduler action list --group=woocommerce-db-updates --status=pending` again.
1. Enable the following snippet on your site, and repeat the steps above. Confirm that this time (in step 4) the updates are scheduled to run in "20 minutes from now".
   ```php
   <?php
   add_filter( 'woocommerce_db_update_schedule_delay', fn() => 20 * 60 );
   ```

#### Use the filter to bring back the old behavior (requiring confirmation)
1. Add the following snippet to the mu-plugin file you previously created or to a different one:
   ```php
   <?php
   add_filter( 'woocommerce_enable_auto_update_db', '__return_false' );
   ```
1. Run the following WP-CLI commands to "simulate" WC being updated from 9.2.0 to the current version:
   ```
   wp transient delete wc_installing
   wp option set woocommerce_db_version 9.2.0
   wp option set woocommerce_version 9.2.0
   ```
1. Visit any page on the admin.
1. Confirm that no actions in group `woocommerce-db-updates` were automatically scheduled by running the following command:
   ```
   wp action-scheduler action list --group=woocommerce-db-updates --status=pending
   ```
1. Confirm that your admin pages show a notice on top asking to run the updates:
   Screenshot 2025-01-15 at 17 34 16
1. Click the **Update WooCommerce Database** button on this notice.
1. Run the following command to confirm that this has scheduled the upgrade routines. Check that their scheduled time is your current time (with at most a few second difference).
   ```
   wp action-scheduler action list --group=woocommerce-db-updates --status=pending --per_page=0
   ```
1. Go to WC > Status > Logs and select the `wc-updater` log file.
1. Confirm that the log file includes the following entries:
   ```
   Manual database update triggered.
   Scheduling database updates (from 9.2.0 to 9.7.0)...
      Updates from version 9.3.0 scheduled.
      Updates from version 9.4.0 scheduled.
      Updates from version 9.5.0 scheduled.
   Database updates scheduled.
   ```
1. Run all the pending migration routines at once by running the following command:
   ```
   wp action-scheduler action run $(wp action-scheduler action list --group=woocommerce-db-updates --status=pending --format=ids --per_page=0)
   ```

#### After testing
1. Don't forget to remove any snippets you added, be kind to others and enjoy life.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/includes/class-wc-install.php
+++ b/plugins/woocommerce/includes/class-wc-install.php
@@ -666,8 +666,7 @@ class WC_Install {
 
 	/**
 	 * Is DB auto-update enabled? This controls whether database updates are applied without prompting the admin.
-	 * This is the default behavior since version 9.7.0 and can be overridden via filter
-	 * 'woocommerce_enable_auto_update_db' or by setting the constant WC_DISABLE_DB_AUTO_UPDATE to `true`.
+	 * This is the default behavior since 9.8.0 and can be overridden via filter 'woocommerce_enable_auto_update_db'.
 	 *
 	 * @since 9.7.0
 	 *
@@ -679,9 +678,7 @@ class WC_Install {
 		 *
 		 * @since 3.2.0
 		 */
-		return
-			! Constants::is_true( 'WC_DISABLE_DB_AUTO_UPDATE' )
-			&& (bool) apply_filters( 'woocommerce_enable_auto_update_db', true );
+		return (bool) apply_filters( 'woocommerce_enable_auto_update_db', true );
 	}
 
 	/**
--- a/plugins/woocommerce/includes/class-wc-install.php
+++ b/plugins/woocommerce/includes/class-wc-install.php
@@ -666,9 +666,9 @@ class WC_Install {
 
 	/**
 	 * Is DB auto-update enabled? This controls whether database updates are applied without prompting the admin.
-	 * This is the default behavior since 9.8.0 and can be overridden via filter 'woocommerce_enable_auto_update_db'.
+	 * This is the default behavior since 9.9.0 and can be overridden via filter 'woocommerce_enable_auto_update_db'.
 	 *
-	 * @since 9.8.0
+	 * @since 9.9.0
 	 *
 	 * @return bool TRUE if database auto-updates are enabled. FALSE otherwise.
 	 */
@@ -764,7 +764,7 @@ class WC_Install {
 			 * Filters the maximum delay in seconds to apply to the scheduling of database updates when automatic
 			 * updates are enabled.
 			 *
-			 * @since 9.8.0
+			 * @since 9.9.0
 			 *
 			 * @param int $delay The maximum delay in seconds. Default is 1800 (30 minutes).
 			 */
--- a/plugins/woocommerce/includes/class-wc-install.php
+++ b/plugins/woocommerce/includes/class-wc-install.php
@@ -680,7 +680,7 @@ class WC_Install {
 		 */
 		return (bool) apply_filters(
 			'woocommerce_enable_auto_update_db',
-			! defined( 'WC_DISABLE_DB_AUTO_UPDATE' ) || false === (bool) WC_DISABLE_DB_AUTO_UPDATE
+			! Constants::is_true( 'WC_DISABLE_DB_AUTO_UPDATE' )
 		);
 	}
 
--- a/plugins/woocommerce/includes/class-wc-install.php
+++ b/plugins/woocommerce/includes/class-wc-install.php
@@ -751,15 +751,37 @@ class WC_Install {
 	private static function update( bool $schedule_with_delay = false ) {
 		$current_db_version = get_option( 'woocommerce_db_version' );
 		$current_wc_version = WC()->version;
-
-		// We add some randomness here prevent overloading servers after an upate has been installed.
-		$scheduled_time = time() + ( $schedule_with_delay ? wp_rand( 1, 15 * MINUTE_IN_SECONDS ) : 0 );
-		$loop           = 0;
+		$scheduled_time     = time();
 
 		wc_get_logger()->info(
 			sprintf( 'Scheduling database updates (from %s to %s)...', $current_db_version, $current_wc_version ),
 			array( 'source' => 'wc-updater' )
 		);
+
+		// We add some randomness here to prevent overloading servers after an upate has been installed.
+		if ( $schedule_with_delay ) {
+			/**
+			 * Filters the maximum delay in seconds to apply to the scheduling of database updates when automatic
+			 * updates are enabled.
+			 *
+			 * @since 9.8.0
+			 *
+			 * @param int $delay The maximum delay in seconds. Default is 1800 (30 minutes).
+			 */
+			$scheduled_time_delay = absint( apply_filters( 'woocommerce_db_update_schedule_delay', 30 * MINUTE_IN_SECONDS ) );
+			$scheduled_time_delay = wp_rand( 0, $scheduled_time_delay );
+
+			if ( $scheduled_time_delay > 0 ) {
+				wc_get_logger()->info(
+					sprintf( '  Updates will begin running in approximately %s.', human_time_diff( 0, $scheduled_time_delay ) ),
+					array( 'source' => 'wc-updater' )
+				);
+
+				$scheduled_time += $scheduled_time_delay;
+			}
+		}
+
+		$loop = 0;
 		foreach ( self::get_db_update_callbacks() as $version => $update_callbacks ) {
 			if ( version_compare( $current_db_version, $version, '<' ) ) {
 				foreach ( $update_callbacks as $update_callback ) {
--- a/plugins/woocommerce/includes/class-wc-install.php
+++ b/plugins/woocommerce/includes/class-wc-install.php
@@ -706,7 +706,7 @@ class WC_Install {
 			 */
 			if ( self::is_db_auto_update_enabled() ) {
 				wc_get_logger()->info( 'Automatic database update triggered.', array( 'source' => 'wc-updater' ) );
-				self::update();
+				self::update( true );
 			} else {
 				WC_Admin_Notices::add_notice( 'update', true );
 			}
@@ -745,11 +745,16 @@ class WC_Install {
 
 	/**
 	 * Push all needed DB updates to the queue for processing.
+	 *
+	 * @param bool $schedule_with_delay Whether updates are scheduled to run with a random delay (of up to 15 mins).
 	 */
-	private static function update() {
+	private static function update( bool $schedule_with_delay = false ) {
 		$current_db_version = get_option( 'woocommerce_db_version' );
 		$current_wc_version = WC()->version;
-		$loop               = 0;
+
+		// We add some randomness here prevent overloading servers after an upate has been installed.
+		$scheduled_time = time() + ( $schedule_with_delay ? wp_rand( 1, 15 * MINUTE_IN_SECONDS ) : 0 );
+		$loop           = 0;
 
 		wc_get_logger()->info(
 			sprintf( 'Scheduling database updates (from %s to %s)...', $current_db_version, $current_wc_version ),
@@ -759,7 +764,7 @@ class WC_Install {
 			if ( version_compare( $current_db_version, $version, '<' ) ) {
 				foreach ( $update_callbacks as $update_callback ) {
 					WC()->queue()->schedule_single(
-						time() + $loop,
+						$scheduled_time + $loop,
 						'woocommerce_run_update_callback',
 						array(
 							'update_callback' => $update_callback,
@@ -780,7 +785,7 @@ class WC_Install {
 		if ( version_compare( $current_db_version, $current_wc_version, '<' ) &&
 			! WC()->queue()->get_next( 'woocommerce_update_db_to_current_version' ) ) {
 			WC()->queue()->schedule_single(
-				time() + $loop,
+				$scheduled_time + $loop,
 				'woocommerce_update_db_to_current_version',
 				array(
 					'version' => $current_wc_version,
--- a/plugins/woocommerce/includes/class-wc-install.php
+++ b/plugins/woocommerce/includes/class-wc-install.php
@@ -663,6 +663,27 @@ class WC_Install {
 	}
 
 	/**
+	 * Is DB auto-update enabled? This controls whether database updates are applied without prompting the admin.
+	 * This is the default behavior since version 9.7.0 and can be overridden via filter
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/includes/class-wc-install.php
```diff
@@ -451,6 +452,7 @@ protected static function run_update_callback_end( $callback, $result ) {
 	public static function install_actions() {
 		if ( ! empty( $_GET['do_update_woocommerce'] ) ) { // WPCS: input var ok.
 			check_admin_referer( 'wc_db_update', 'wc_db_update_nonce' );
+			wc_get_logger()->info( 'Manual database update triggered.', array( 'source' => 'wc-updater' ) );
```
naman03malhotra, line 459: Thanks for adding this.

plugins/woocommerce/includes/class-wc-install.php
```diff
@@ -739,22 +769,28 @@ private static function update() {
 					);
 					++$loop;
 				}
+
+				wc_get_logger()->info(
+					sprintf( '  Updates from version %s scheduled.', $version ),
```
naman03malhotra, line 795:
```suggestion
					sprintf( 'Updates from version %s scheduled.', $version ),
```
Are these extra spaces in the front intentional?

plugins/woocommerce/includes/class-wc-install.php
```diff
@@ -739,22 +769,28 @@ private static function update() {
...
+				$scheduled_time + $loop,
 				'woocommerce_update_db_to_current_version',
 				array(
 					'version' => $current_wc_version,
 				),
 				'woocommerce-db-updates'
 			);
 		}
+
+		wc_get_logger()->info( 'Database updates scheduled.', array( 'source' => 'wc-updater' ) );
```
naman03malhotra, line 814: Should we also consider logging the `$scheduled_time`?

plugins/woocommerce/includes/class-wc-install.php
```diff
@@ -721,16 +741,26 @@ public static function get_db_update_callbacks() {
...
+	 * @param bool $schedule_with_delay Whether updates are scheduled to run with a random delay (of up to 15 mins).
 	 */
-	private static function update() {
+	private static function update( bool $schedule_with_delay = false ) {
 		$current_db_version = get_option( 'woocommerce_db_version' );
-		$loop               = 0;
+		$current_wc_version = WC()->version;
 
+		// We add some randomness here prevent overloading servers after an upate has been installed.
+		$scheduled_time = time() + ( $schedule_with_delay ? wp_rand( 1, 15 * MINUTE_IN_SECONDS ) : 0 );
```
naman03malhotra:
```suggestion
		$scheduled_time = time() + ( $schedule_with_delay ? wp_rand( 1, 30 * MINUTE_IN_SECONDS ) : 0 );
```
Perhaps I am being over cautious, since this was manual before the gap was like to be spead out in days if not weeks as users come across this update. Should we consider increasing this time bracket to 30 or 60 minutes (or more) to spread out the load a bit more, I am worried that 15 minutes window could be a tight space for the server load to spike, wdyt?

naman03malhotra (commented):
The new auto update logic works well 👏 , left some small comments regarding server load to consider, nothing major.

✅  Confirm also that these actions were scheduled for a time in the future (this should be a random time between "now" and 15 minutes from "now")

✅  Confirm that the log file entires

✅  Repeat the above steps and check that the scheduled time is different (again, between "now" and "15 minutes from now") - time is always random.

✅  Confirm that your admin pages show a notice on top asking to run the updates - previous behavior works well.



Screenshot 2025-02-26 at 7 24 57 PM
Screenshot 2025-02-26 at 7 25 04 PM
Screenshot 2025-02-26 at 7 31 52 PM

Screenshot 2025-02-26 at 7 43 13 PM

plugins/woocommerce/includes/class-wc-install.php
jorgeatorres: Yes, so that subtasks look "indented" in the logs:

Screenshot 2025-02-26 at 19 45 49

A similar strategy is used by WP's auto plugin updates, but happy to adjust if you think it's confusing.

plugins/woocommerce/includes/class-wc-install.php
jorgeatorres: Added in 39d24b8e0cdcd282ccb8aa582baea240c02bca63 as part of other changes. A human-readable version of the delay is logged. That's not `$scheduled_time` per se, but the log entry already has the current time, so I felt "updates will run in 10 minutes" was enough to convey the situation.

plugins/woocommerce/includes/class-wc-install.php
jorgeatorres: Agreed. I changed it to 30 mins and also made the [max delay filterable](https://github.com/woocommerce/woocommerce/pull/54384/commits/39d24b8e0cdcd282ccb8aa582baea240c02bca63#diff-b082a396ea78d8aaa870e896edc388647bd548df1c4eddc9b4953df241faf99bR759-R767), so that it's configurable, probably be hosts. Do you think this is reasonable?

Some other stuff I was thinking was whether it made sense for this to be in seconds (or entire minutes, though seconds increases the likelihood of updates not occurring at the same time) and also whether the minimum should be something like 5 minutes and not 0. Any thoughts?

jorgeatorres:
@naman03malhotra: Thank you for the thorough review. I've made some adjustments. Feel free to take another look when you have the chance.

plugins/woocommerce/includes/class-wc-install.php
naman03malhotra: >it made sense for this to be in seconds 

I agree; if we keep the `$scheduled_time` in seconds, we are providing a wider range of timestamps for the update to occur. 

> the minimum should be something like 5 minutes and not 0

I am not sure if that would make much difference as we'll have a wide range (with seconds) of randomized timestamps for which the update will be scheduled, but just to be safe, we can keep an initial delay of 2-5 minutes.

plugins/woocommerce/includes/class-wc-install.php
naman03malhotra: Makes sense. Good to have this. :unagi: xD

naman03malhotra (commented):
Thanks for making the changes, looks great to me.

naman03malhotra (approved):
Only a small discussion is left about keeping the initial delay, we can check the same with the atomic folks. I am okay with either way.

plugins/woocommerce/includes/class-wc-install.php
```diff
@@ -725,16 +745,48 @@ public static function get_db_update_callbacks() {
 
 	/**
 	 * Push all needed DB updates to the queue for processing.
+	 *
+	 * @param bool $schedule_with_delay Whether updates are scheduled to run with a random delay (of up to 15 mins).
 	 */
-	private static function update() {
+	private static function update( bool $schedule_with_delay = false ) {
```
jmdodd: Should the default behavior here be to $schedule_with_delay if is_db_auto_update_enabled returns true to avoid many auto updates at the same time, or alternately to allow a filter to set whether the auto update should have jitter added?

jorgeatorres: Hey @jmdodd! Appreciate you taking a look.

Just to confirm, what you're saying is that instead of the `$schedule_with_delay` arg here, when auto-db updates are enabled we apply a filter to determine whether the auto update should have delay or not. I like that idea as it reduces some of the changes outside this function. Should the default be to have jitter or not? 🤔 

Since we have you here already, beyond [this filter](https://github.com/woocommerce/woocommerce/blob/c81d1acb75d84f6abfc0a1e8a5da895a4d2c75a6/plugins/woocommerce/includes/class-wc-install.php#L771) to set the maximum value for the delay, do you think we'll need a "minimum" delay too? So that the random time chosen for the updates is always *at least* a given number of seconds from the time the plugin was updated? If so, does 5 minutes sound reasonable or should we also make this filterable?
 
Thanks! 🙇

jmdodd: I think if someone (like a host) has auto-db updates enabled, then they may absolutely want to be able to ensure that the auto-db updates are also not all happening at the same time. In short -- I do not want auto-db updates to exclude the possibility of also needing/specifying jitter.

I think we do not need to specify a minimum delay of 5 minutes as opposed to 0, but having a filterable value will always give more control to the folks running the plugin.

jorgeatorres:
Based on @jmdodd's comment above, I decided to rework the code a bit. Auto-db updates are now all scheduled to run "as soon as possible" but the filter `woocommerce_db_update_schedule_delay` can be used to add a delay.

In the previous implementation the actual delay was randomly chosen (using `woocommerce_db_update_schedule_delay` as the possible maximum), but I've removed that code. I don't think WC should decide how the delay is applied. With the filter, admin/hosts can add a delay in the way that suits them best, be it fixed or random.

@naman03malhotra: Because of the above I'm requesting yet another review from you (sorry and thank you!)and have updated the testing instructions based on the most recent changes.

@jmdodd: If you have the chance to review or comment on this approach, that would be awesome.

jmdodd:
> In the previous implementation the actual delay was randomly chosen (using `woocommerce_db_update_schedule_delay` as the possible maximum), but I've removed that code. I don't think WC should decide how the delay is applied. With the filter, admin/hosts can add a delay in the way that suits them best, be it fixed or random.

I like this approach. Filters allow the platform to specify a preference at a reasonably low priority while still allowing hosts, managing agencies, and users to override the filter.

naman03malhotra (approved):
Thanks Jorege for making this change.

✅  With the latest change, I can confirm that the actions were scheduled immediately

Screenshot 2025-03-04 at 6 21 17 PM
✅  The delay values are filterable and adds a delay to the action schedule.
Screenshot 2025-03-04 at 6 21 06 PM


**Comms prep:**

> I decided to rework the code a bit. Auto-db updates are now all scheduled to run "as soon as possible" but the filter woocommerce_db_update_schedule_delay can be used to add a delay.

Whenever we roll out this update, nearly every hosting provider and agency will need to configure a custom delay value to prevent system spikes. Although this may seem obvious, it’s important to emphasize that we should proactively raise awareness about this change several releases in advance.

Additionally, we must prepare internal communications for VIP, WP.com, WP Cloud, and other relevant teams.

PR #54386, opened 2025-01-10
Cart/Checkout Block: Allow HTML on shipping rates

### LINKED ISSUE #54385: Cart and Checkout Block does not render HTML in product names on shipping rates
### Prerequisites

### Describe the bug

Currently we support HTML on products names, it's rendered correctly on the cart and checkout blocks items listing.

Some plugins, like WooCommerce Subscriptions, adds the product name on the shipping rates to specify what's related to each shipping rate, and the HTML on the product names are not loaded as expected.

This happens because the `Label` component [here](https://github.com/woocommerce/woocommerce/blob/3dd035917b982093ac15f3472d7bc7df0e56b94a/plugins/woocommerce-blocks/packages/components/label/index.tsx#L48) does not load HTML.

Image

Image

### Expected behavior

Load the HTML in product names on the shipping rates. 

### Actual behavior

The HTML source code is showed in the product name.

### Steps to reproduce

1. Install WooCommerce Subscriptions.
2. Create a subscription with HTML in the name.
3. Add this subscription to the cart.
4. Add a normal product to the cart.
5. Access the cart or checkout page.
6. The HTML in the product name for the shipping rate is not loaded.

### WordPress Environment

This is happening on `trunk`.

### Isolating the problem

alefesouza:

This PR allows HTML to be rendered on the shipping rates section of cart and checkout block to handle products with HTML in the name.

| Before | After |
|-|-|
| Screenshot 2025-01-10 at 5 14 04 PM | Screenshot 2025-01-10 at 5 11 58 PM |
| Screenshot 2025-01-10 at 6 01 38 PM | Screenshot 2025-01-10 at 5 38 18 PM |

Closes #54385 .

1. Install WooCommerce Subscriptions.
2. Create a subscription with HTML in the name.
3. Add this subscription to the cart.
4. Add a normal product to the cart.
5. Access the blocks cart and checkout page.
6. The HTML in the product name for the shipping rate is not loaded.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

This PR does not need a changelog because it's a simple change that does not affect so many users.

</details>


plugins/woocommerce-blocks/packages/components/label/index.tsx
```diff
@@ -45,7 +47,17 @@ const Label = ( {
 	if ( hasLabel && hasScreenReaderLabel && label !== screenReaderLabel ) {
 		return (
 			<Wrapper { ...wrapperProps }>
-				<span aria-hidden="true">{ label }</span>
+				{ allowHTML ? (
+					<span
+						aria-hidden="true"
+						dangerouslySetInnerHTML={ {
```
mikejolley: There is a `RawHTML` component which may be more appropriate here.

plugins/woocommerce-blocks/packages/components/label/index.tsx
```diff
@@ -20,6 +21,7 @@ export interface LabelProps extends HTMLProps< HTMLElement > {
  */
 const Label = ( {
 	label,
+	allowHTML,
```
mikejolley: Should we default to `false` here?

plugins/woocommerce-blocks/packages/components/label/index.tsx
```diff
@@ -45,7 +47,17 @@ const Label = ( {
 	if ( hasLabel && hasScreenReaderLabel && label !== screenReaderLabel ) {
 		return (
 			<Wrapper { ...wrapperProps }>
-				<span aria-hidden="true">{ label }</span>
+				{ allowHTML ? (
+					<span
+						aria-hidden="true"
+						dangerouslySetInnerHTML={ {
+							__html: label,
```
mikejolley: How about doing some form of sanitization using `sanitizeHTML`? This uses DomPurify library.

mikejolley (changes requested):
Hi @alefesouza 

I think allowing HTML is fine but I think we need some protections and limitations in place. See my comments.

Thanks

alefesouza:
@mikejolley thanks! Updated on https://github.com/woocommerce/woocommerce/pull/54386/commits/cba713897a0bde5ac7be7964bf20f80cba4b6d3a, I was using the same approach used by the cart product item name [here](https://github.com/woocommerce/woocommerce/pull/54386/commits/cba713897a0bde5ac7be7964bf20f80cba4b6d3a), maybe it would be good to use your suggestions there too.

plugins/woocommerce-blocks/packages/components/label/index.tsx
```diff
@@ -45,7 +48,25 @@ const Label = ( {
 	if ( hasLabel && hasScreenReaderLabel && label !== screenReaderLabel ) {
 		return (
 			<Wrapper { ...wrapperProps }>
-				<span aria-hidden="true">{ label }</span>
+				{ allowHTML ? (
+					<RawHTML>
+						{ sanitizeHTML( label, {
+							tags: [
```
mikejolley, line 54: Not blocking, but I am fairly sure if you leave this out, it uses our defaults.

```
const ALLOWED_TAGS = [ 'a', 'b', 'em', 'i', 'strong', 'p', 'br' ];
const ALLOWED_ATTR = [ 'target', 'href', 'rel', 'name', 'download' ];
```

I see no reason why those defaults couldn't be expanded to include span and style.

mikejolley (approved):
Thanks for making those changes. I made one comment but its not blocking. Approving.

plugins/woocommerce-blocks/packages/components/label/index.tsx
alefesouza: I think maybe we didn't want to include the `style` in the past so this can break existing systems where we are not expecting customized styles.

PR #54390, opened 2025-01-10
Revert #51185

coreymckrill:

There are some concerns about de-encapsulating the objects in the meta-boxes-order.js script and making them globally available. In the PR that made this change, the goal was to be able to re-initialize elements in the UI that had been replaced and had their event listeners deleted. There are other ways to accomplish this without exposing the script objects, so we should go back to the drawing board and try something different.

Check the original changeset from https://github.com/woocommerce/woocommerce/pull/51185 and make sure it all has been completely removed. Since it was only additive, and no other PRs depend on it, removing the added code should not cause any regressions.

### DIFF
--- a/plugins/woocommerce/changelog/revert-51185
+++ b/plugins/woocommerce/changelog/revert-51185
@@ -1,0 +1,5 @@
+Significance: patch
+Type: fix
+Comment: This reverts functionality slated for release in 9.6, so it actually removes a changelog entry, and doesn't requite its own
+
+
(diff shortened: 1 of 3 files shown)


jorgeatorres (approved):

naman03malhotra:
This was reverted as part of https://github.com/woocommerce/woocommerce/pull/54405 due to conflicts from trunk removing `cherry pick failed` label.

PR #54399, opened 2025-01-13
Add Cost of Goods Sold fields in product editors in admin

Konamiman:

This pull request adds the ability to edit the Cost of Goods Sold values for products in the product editor of the classic admin UI. It also adds a (yet incomplete) "Cost" column to the products list table in the admin area.

#### Simple and variable product editors

The "General" tab of the product editor gets a new cost column that gets tooltip'ed differently depending on whether the product is variable or not:

image

image

This field will show a value of zero for products that don't have an explicit value set. There's nothing in the database for these, and `get_cogs_value` for the corresponding product object will return `null`; but as per https://github.com/woocommerce/woocommerce/pull/54458 this is equivalent to zero, and an explicit zero is clearer than an empty value.

#### Variation editor

Variations get their own cost field. For variations that don't have an explicit cost value set, a placeholder indicating the default value configured for the parent is displayed:

image

image

#### Quick and bulk edit

The cost can be edited using the standard quick and bulk editors too. For variable products the value that gets modified is the default value that gets applied to variations:

image

image

image

#### Bulk actions for variations

One new bulk action is added for variations to remove the custom cost value of all variations (thus effectively reverting to the default value). Confirmation is requested before proceeding:

image

image

💡 An additional added improvement is that now the action selector dropdown will go back to the default "Bulk actions" text after the action has run (or the prompt/confirmation has been cancelled). This is useful when having to retry an action after having supplied invalid data.

#### Cost column

A new "Cost" column is added to the products list table:

image

Two new filters are added to customize how this column is rendered, similar to the existing filter for prices: `woocommerce_get_cogs_html` and `woocommerce_empty_cogs_html`

For now this column is not sortable, as this would require the addition of a new column in the `wc_product_meta_lookup` database table. This will be implemented in a separate pull request.

Also, for variable products this column only shows the default cost value for variations. Additional information (like the range of costs defined for existing variations, similarly to how the price column shows a range of variation prices) might be added in an additional pull request too.

1. Enable the Cost of Goods Sold feature if you haven't done that yet (_WooCommerce - Settings - Advanced - Features_, check the box for "Cost of Goods Sold" and save).

2. Create a few products, simple and variable with some variations.

3. Go to the products table ("Products" in the left menu). Open the regular editor for some of the products, simple and variable. The cost value that you see in the "General" tab should be zero. Verify that you can modify the value and save it, and when you reload the editor the value is still there.

4. Try the quick editor ("Quick edit" link in the product rows in the products table). Verify that the cost shown in the "Cost" field is initially 0, and that you can change it and save, and the value gets persisted too.

5. Try the bulk editor (mark the checkbox for a few products in the products table, select "Edit" in the "Bulk actions" dropdown, then click "Apply"). Verify that the cost is modified for the selected products only if the cost field is set to "Change to:", and that an empty value sets a cost of zero.

6. Open the editor for a variation ("Variations" tab when editing a variable product). Verify that the "Cost" field shows a placeholder with the cost value set for the parent product in the "General" tab (or zero, if none has been set). Verify that you can set a value and save it, and that you can then delete the value and save it too (note that an empty value is not equivalent to zero in this case).

7. Try the "Remove custom costs" bulk action for variations (edit a variable object, go to the "Variations" tab, click the "Bulk actions" dropdown). Try accepting and cancelling the confirmation prompt (it should remove the custom cost of all the variations in the first case, and no costs should be changed in the second case).

8. Verify that the "Cost" column in the products table shows the proper values for each product (zero is indicated with a hyphen).

9. Try the cost column HTML filters with the following snippet:

```php
add_filter('woocommerce_empty_cogs_html', function($html, $product)  {
    return "<p>No cost for " . $product->get_id() . "</p>";
}, 2, 10);

add_filter('woocommerce_get_cogs_html', function($html, $value, $product)  {
    return
        0.0 === $value ?
        $html :
        '<p>Cost for ' . $product->get_id() . ': ' . $value . '</p>';
}, 3, 10);
```

10. Finally, disable the Cost of Goods Sold feature and verify that none of the added UI elements that you just tested is visible anymore, but you can still edit the rest of the product fields (using regular editor, quick edit, bulk edit, variation editor and variation bulk action).

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


findingsimple:
@woocommerce/ventures @budzanowski can we also get someone from Venures to help with the review here pretty please?

dwainm (commented):
Most of instructions work, but I wanted to note some deviations while testing:

1. The 9.8.0-dev admin.css version didn't load for me, but the CSS works when I added it manually.

Screenshot 2025-02-06 at 11 36 17

2. The quick edit for a product that had no COGS set didn't show zero:
> Try the quick editor ("Quick edit" link in the product rows in the products table). Verify that the cost shown in the "Cost" field is initially 0, and that you can change it and save, and the value gets persisted too.

Screenshot 2025-02-06 at 11 47 41

dwainm (approved):
No issues obvious while looking at the code and from my previous comment, everything works.

plugins/woocommerce/includes/abstracts/abstract-wc-product.php
```diff
@@ -1947,6 +1947,37 @@ public function get_price_html( $deprecated = '' ) {
...
+			/**
+			 * Filter to customize how an empty Cost of Goods Sold value for a product gets rendered to HTML.
+			 *
+			 * @param string $html The rendered HTML.
+			 * @param WC_Product $product The product for which the cost is rendered.
+			 *
+			 * @since 9.7.0
+			 */
+			apply_filters( 'woocommerce_empty_cogs_html', '', $this ) :
+			wc_price( $value ) . $this->get_price_suffix();
```
coreymckrill: Nitpick: this is kind of hard to read. It took me a minute to figure out why this was indented and that the `apply_filters` and `wc_price` lines are an either/or. I think this would be easier to read and grok if you used a standard if/else here instead of a ternary.

plugins/woocommerce/includes/abstracts/abstract-wc-product.php
```diff
@@ -1947,6 +1947,37 @@ public function get_price_html( $deprecated = '' ) {
...
+
+		/**
+		 * Filter to customize how the Cost of Goods Sold value for a product gets rendered to HTML.
+		 *
+		 * @param string $html The rendered HTML.
+		 * @param float $value The cost value that is being rendered.
+		 * @param WC_Product $product The product for which the cost is rendered.
+		 *
+		 * @since 9.7.0
```
coreymckrill: Noting that these will presumably need to get updated to 9.8.0.

```suggestion
		 * @since 9.8.0
```

plugins/woocommerce/includes/admin/class-wc-admin-assets.php
```diff
@@ -319,6 +319,7 @@ public function admin_scripts() {
 					'i18n_edited_variations'              => esc_js( __( 'Save changes before changing page?', 'woocommerce' ) ),
 					'i18n_variation_count_single'         => esc_js( __( '1 variation', 'woocommerce' ) ),
 					'i18n_variation_count_plural'         => esc_js( __( '%qty% variations', 'woocommerce' ) ),
+					'i18n_variation_cost_remove_warning'  => esc_js( __( 'The custom cost of goods sold will revert back to their default value for all the variations - would you like to continue?', 'woocommerce' ) ),
```
coreymckrill:
```suggestion
					'i18n_variation_cost_remove_warning'  => esc_js( __( 'The custom cost of goods sold values will revert back to their defaults for all the variations. Would you like to continue?', 'woocommerce' ) ),
```

Konamiman:
Hi @dwainm 

> The 9.8.0-dev admin.css version didn't load for me
> The quick edit for a product that had no COGS set didn't show zero:

I guess you are testing locally after checking out the corresponding branch. Try running `pnpm run build` first, that should fix both issues.

plugins/woocommerce/includes/abstracts/abstract-wc-product.php
Konamiman: Fair enough. Changed.

plugins/woocommerce/changelog/pr-54399
```diff
@@ -0,0 +1,4 @@
+Significance: minor
```
jorgeatorres, line 1: Not sure if this is truly 'minor' but ok.

plugins/woocommerce/client/legacy/js/admin/meta-boxes-product-variation.js
```diff
@@ -1284,7 +1285,16 @@ jQuery( function ( $ ) {
 					if ( value != null ) {
 						data.value = value;
 					} else {
-						return;
+						cancel = true;
+					}
+					break;
+				case 'variable_unset_cogs_value':
+					if(
```
jorgeatorres:
```suggestion
					if (
```

plugins/woocommerce/client/legacy/js/admin/meta-boxes-product-variation.js
```diff
@@ -1284,7 +1285,16 @@ jQuery( function ( $ ) {
...
+						! window.confirm(
```
jorgeatorres:
```suggestion
						! window.confirm (
```

plugins/woocommerce/client/legacy/js/admin/meta-boxes-product-variation.js
```diff
@@ -1317,41 +1327,48 @@ jQuery( function ( $ ) {
...
 					}
 					break;
 			}
 
-			if ( 'delete_all' === do_variation_action && data.allowed ) {
-				$( '#variable_product_options' )
-					.find( '.variation-needs-update' )
-					.removeClass( 'variation-needs-update' );
-				$( '.generate_variations' ).text( 'Generate variations' );
+			if( cancel ) {
```
jorgeatorres:
```suggestion
			if ( cancel ) {
```

plugins/woocommerce/client/legacy/js/admin/meta-boxes-product-variation.js
```diff
@@ -1317,41 +1327,48 @@ jQuery( function ( $ ) {
...
+				document.getElementById( 'field_to_edit' ).selectedIndex = 0;
```
jorgeatorres: While there's technically nothing wrong with this, all the code in this file is using jQuery to fetch the `#field_to_edit` field. For consistency, maybe we could do this using jQuery too?

Maybe something like `$( '#field_to_edit' ).val( 'bulk_actions' );` 

Obviously not a blocker, but this part does stand out given all the code around it. Same below.

plugins/woocommerce/client/legacy/js/admin/meta-boxes-product-variation.js
```diff
@@ -1317,41 +1327,48 @@ jQuery( function ( $ ) {
...
 
-			wc_meta_boxes_product_variations_ajax.block();
+				wc_meta_boxes_product_variations_ajax.block();
 
-			$.ajax( {
-				url: woocommerce_admin_meta_boxes_variations.ajax_url,
-				data: {
-					action: 'woocommerce_bulk_edit_variations',
-					security:
+				$.ajax({
```
jorgeatorres:
```suggestion
				$.ajax( {
```

plugins/woocommerce/client/legacy/js/admin/meta-boxes-product-variation.js
```diff
@@ -1317,41 +1327,48 @@ jQuery( function ( $ ) {
...
+						data: data,
+					},
+					type: 'POST',
+					success: function () {
+						wc_meta_boxes_product_variations_pagenav.go_to_page(
+							1,
+							changes
+						);
+					},
+					complete: function() {
```
jorgeatorres:
```suggestion
					complete: function () {
```

plugins/woocommerce/includes/class-wc-ajax.php
```diff
@@ -3657,6 +3675,35 @@ private static function check_locked_orders( $response, $data ) {
 		return wc_get_container()->get( Automattic\WooCommerce\Internal\Admin\Orders\EditLock::class )->check_locked_orders_ajax( $response, $data );
 	}
 
+	// phpcs:disable Generic.CodeAnalysis.UnusedFunctionParameter.FoundAfterLastUsed
+	/**
+	 * Render a variation editor.
+	 *
+	 * NOTE! Do NOT remove the apparently unused function arguments.
+	 * These are actually used inside the included html-variation-admin template.
+	 *
```
jorgeatorres, line 3685: Can we add a `@since` tag?

plugins/woocommerce/includes/class-wc-ajax.php
```diff
@@ -3657,6 +3675,35 @@ private static function check_locked_orders( $response, $data ) {
...
+	 * @param WC_Product $product_object PArent product of the variation being edited.
+	 * @param WC_Product $variation_object Variation being edited.
+	 * @param int        $loop Index of the variation being rendered.
+	 * @param float|null $base_cost Default cost for variations, null if the Cost of Goods Sold feature is disabled.
+	 */
+	private static function render_variation_html( WC_Product $product_object, WC_Product $variation_object, $loop, ?float $base_cost ) {
+		$variation_id   = $variation_object->save();
```
jorgeatorres: I think it's great we're extracting mostly identical code to this function but why would we need to save the variation object to render some details about it?

While this is true for the AJAX callback for `add_variation` above ([ref](https://github.com/woocommerce/woocommerce/blob/5715bbd2f253eb3a43f3e60c0873ee9c09a70d0c/plugins/woocommerce/includes/class-wc-ajax.php#L837)), it's not true for `load_variations()` (see https://github.com/woocommerce/woocommerce/blob/5715bbd2f253eb3a43f3e60c0873ee9c09a70d0c/plugins/woocommerce/includes/class-wc-ajax.php#L2416).

Even if it ends up being innocuous, I'd prefer if we don't unnecessarily call `save()`.

jorgeatorres (changes requested):
Hi @Konamiman!

Thanks for your work on this. I followed the instructions and this tests well. I left some feedback but nothing too serious. I also have a few questions/comments:

- > There's nothing in the database for these, and get_cogs_value for the corresponding product object will return null; but as per https://github.com/woocommerce/woocommerce/pull/54458 this is equivalent to zero, and an explicit zero is clearer than an empty value.

   I read the linked PR but it's still unclear to me why 0 is clearer than an empty value. I mean, if (for some reason) something costs me 0 to make why is that the same as "I haven't even looked at this"? I'm missing a bunch of context around COGS so apologies if this has been discussed before.
- > Additional information (like the range of costs defined for existing variations, similarly to how the price column shows a range of variation prices) might be added in an additional pull request too.

   I guess it's fine if this comes in a later PR but the wording here makes me think it might not come at all. If that's the case, I believe we need to find a way to convey that the cost for variations might not be accurate, as otherwise this would just be confusing: you go and set prices for variations and then the table shows something else.
- I tried to test the new COGS functions in a WP-CLI shell but I was greeted by this error:
   ```
   wp> $p = wc_get_product ( ... );
   wp> $p->get_cogs_value();
   Notice: Function WC_Product::get_cogs_value was called <strong>incorrectly</strong>. The Cost of Goods sold feature is disabled, thus the method called will do nothing and will return dummy data.
   ```

   The same code as a snippet tied to a hook worked correctly, but I believe once the feature is enabled it should also work on the shell. Why is that not the case?
- COGS values don't seem to play nicely with different decimal separators. Changing the decimal separator in WC > Settings > General from `.` to `,` correctly changes things for the display and saving of prices, but not for COGS. Entering `2,5` as a COGS value will result in `2` being saved, which is incorrect.
   Similarly, COGS values are always rounded (`2.0` becomes `2`) as opposed to normal product prices.

Konamiman:
> I read the linked PR but it's still unclear to me why 0 is clearer than an empty value. I mean, if (for some reason) something costs me 0 to make why is that the same as "I haven't even looked at this"?

For all practical purposes (including cost calculations for orders) WooCommerce will consider that the cost of any product that hasn't got a cost value explicitly set is equal to zero, so I thought that it's important to make that zero explicitly stated. The fact that zero values aren't actually stored is an implementation detail (which could cease to be true in the future).

Note that for _variations_ zero and empty have indeed different meanings, and that's why in that case empty values are preserved in the UI.

>>Additional information (like the range of costs defined for existing variations, similarly to how the price column shows a range of variation prices) might be added in an additional pull request too.

> I guess it's fine if this comes in a later PR but the wording here makes me think it might not come at all. 

This is indeed planned (although possibly not for the MVP). Displaying the default variation cost seemed like the most reasonable approach until a proper range display is implemented, but I'm open to discussing alternatives.

> Notice: Function WC_Product::get_cogs_value was called <strong>incorrectly</strong>. The Cost of Goods sold feature is disabled, thus the method called will do nothing and will return dummy data.

This is a bug in the features controller that will get fixed in https://github.com/woocommerce/woocommerce/pull/55303.

> Changing the decimal separator in WC > Settings > General from . to , correctly changes things for the display and saving of prices, but not for COGS. Entering 2,5 as a COGS value will result in 2 being saved, which is incorrect.

Good catch. Fixed in [`b0c4f79` (#54399)](https://github.com/woocommerce/woocommerce/pull/54399/commits/b0c4f796294cfe173f8507d009560d1c6b4db390).

plugins/woocommerce/changelog/pr-54399
Konamiman: Yes it's not really a minor change but the only other alternatives are "Backwards-compatible bug fixes" and "Broke backwards compatibility in some way", which don't seem appropriate either.

plugins/woocommerce/client/legacy/js/admin/meta-boxes-product-variation.js
Konamiman: These and the other linting issues mentioned are fixed in [`432499d` (#54399)](https://github.com/woocommerce/woocommerce/pull/54399/commits/432499d06c88b66b909bfa8b2ac9c511e42b2011).

plugins/woocommerce/client/legacy/js/admin/meta-boxes-product-variation.js
Konamiman: I always try to [KISS](https://en.wikipedia.org/wiki/KISS_principle) and so I use built-in language/framework funcionality instead of external libraries whenever I can. But it's true that this seems inconsistent. I have chaned it to use JQuery in [`432499d` (#54399)](https://github.com/woocommerce/woocommerce/pull/54399/commits/432499d06c88b66b909bfa8b2ac9c511e42b2011).

plugins/woocommerce/includes/class-wc-ajax.php
Konamiman: Aren't those only for public methods?

plugins/woocommerce/includes/class-wc-ajax.php
Konamiman: Good catch. Modified in [`51aee76` (#54399)](https://github.com/woocommerce/woocommerce/pull/54399/commits/51aee76b53e7a9fbcd066eb69cfcc07a948cfb6d) so that the saving happens externally to the method and only when needed.

jorgeatorres (approved):
LGTM. Thanks for making the adjustments @Konamiman!

I'm not merging yet in case @woocommerce/ventures wants to review, per [@jconroy's comment](https://github.com/woocommerce/woocommerce/pull/54399#issuecomment-2638382403) above.

message-dimke (approved):
Looks perfect! Thank you, @Konamiman!

I went through all the 10 steps described and clicked here and there on my own - not a single glitch. Awesome!

PR #54418, opened 2025-01-13
Reverting ProductStatus constants in rest API due class not being loaded

wjrosa:

Closes p1736777313459879-slack-C011ENB20Q1

This PR reverts the usage of the new ProductStatus constant class (introduced in https://github.com/woocommerce/woocommerce/pull/53938) from the product variations API controller class due it not being available upon installation.

Code review should be enough.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 
This is a simple fix reverting a constant usage.

</details>

### DIFF
--- a/plugins/woocommerce/includes/rest-api/Controllers/Version3/class-wc-rest-product-variations-controller.php
+++ b/plugins/woocommerce/includes/rest-api/Controllers/Version3/class-wc-rest-product-variations-controller.php
@@ -615,7 +615,7 @@ class WC_REST_Product_Variations_Controller extends WC_REST_Product_Variations_V
 					'description' => __( 'Variation status.', 'woocommerce' ),
 					'type'        => 'string',
 					// Not using ProductStatus constants here due the class not being loaded upon installation.
-					// https://github.com/woocommerce/woocommerce/issues/37464
+					// See: https://github.com/woocommerce/woocommerce/issues/37464.
 					'default'     => 'publish',
 					'enum'        => array_keys( get_post_statuses() ),
 					'context'     => array( 'view', 'edit' ),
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/includes/rest-api/Controllers/Version3/class-wc-rest-product-variations-controller.php
```diff
@@ -614,7 +614,7 @@ public function get_item_schema() {
 				'status'                => array(
 					'description' => __( 'Variation status.', 'woocommerce' ),
 					'type'        => 'string',
-					'default'     => ProductStatus::PUBLISH,
+					'default'     => 'publish', // Not using ProductStatus constants here due the class not being loaded upon installation.
```
barryhughes:
```suggestion
					// Not using ProductStatus constants here due the class not being loaded upon installation.
					// https://github.com/woocommerce/woocommerce/issues/37464
					'default'     => 'publish',
```

A bit of a nit, and this is entirely optional, but linking to the specific issue in question might be useful if at some future point we want to address this and need additional context.

plugins/woocommerce/includes/rest-api/Controllers/Version3/class-wc-rest-product-variations-controller.php
```diff
@@ -614,7 +614,9 @@ public function get_item_schema() {
 				'status'                => array(
 					'description' => __( 'Variation status.', 'woocommerce' ),
 					'type'        => 'string',
-					'default'     => ProductStatus::PUBLISH,
+					// Not using ProductStatus constants here due the class not being loaded upon installation.
+					// https://github.com/woocommerce/woocommerce/issues/37464
```
barryhughes: Sorry, I suggested this change but it seems PHPCS doesn't like it.

```suggestion
					// See: https://github.com/woocommerce/woocommerce/issues/37464.
```

(I'm not using `@see` or `@link` because those are really only valid in an actual `/**` docblock, though that's another possibility.)

barryhughes (approved):
Looks good to me! One further change needed to satisfy the linter, otherwise I think we're good.

PR #53845, opened 2024-12-18
Replace the Cost of Goods Sold "overrides parent" flag for variations with an "is additive" flag

Konamiman:
### Changes proposed in this Pull Request:

This is a modification to the handling of Cost of Goods Sold values for variations that was introduced in https://github.com/woocommerce/woocommerce/pull/51675.

Currently, the cost value for variations is additive (the effective value is the sum of the parent product value and the value defined for the variation), but there's a per variation "cost overrides parent" flag that makes the defined value to become the final value for the variation.

This pull request changes the name and reverses the meaning of the flag:

* Its name is now "is additive"
* When false (default) the defined value for the variation is its final cost value; when true, the final cost value is the sum of the parent value and the defined value for the variation.

This translates in changes in the codebase as follows:

* The methods to handle the flag in `WC_Product_Variation` are now named `get_cogs_value_is_additive` and `set_cogs_value_is_additive`.
* The name of the meta key where the flag is stored is now `_cogs_value_is_additive`.
* The name of the filters that fire when the flag is loaded or saved are now `woocommerce_load_product_cogs_is_additive_flag` and `woocommerce_save_product_cogs_is_additive_flag`.
* The name of the field in the JSON data for REST API requests is now `defined_value_is_additive`.
* And of course, the logic in `WC_Product_Variation::get_cogs_effective_value_core` is the reverse as with the old "overrides" flag.

The motivation for this change is some research that seems to suggest that variations having their own absolute cost values will be the most common use case.

### How to test the changes in this Pull Request:

Same testing instructions of https://github.com/woocommerce/woocommerce/pull/51675 with the corresponding changes:

* For step 4, the expected values change:

```php
$v=wc_get_product($variation_id);

$v->set_cogs_value(20.5);

echo $v->get_cogs_value(); // 20.5
echo $v->get_cogs_effective_value(); // 20.5
echo $v->get_cogs_total_value(); // 20.5
```

* Same for steps 5 and 6, also note the changes in the names of methods and keys:

```php
echo $v->get_cogs_value_is_additive(); // false
$v->set_cogs_value_is_additive(true);
echo $v->get_cogs_value_is_additive(); // true

echo $v->get_cogs_value(); // 20.5
echo $v->get_cogs_effective_value(); // 32.84
echo $v->get_cogs_total_value(); // 32.84

$v->save();

echo get_post_meta($variation_id, '_cogs_value_is_additive', true); // 'yes'

$v->set_cogs_value_is_additive(false);
$v->save();

[...]

PR #51675, opened 2024-09-25
Add the Cost of Goods Sold related code and REST APIs for the product class

Konamiman:
### Changes proposed in this Pull Request:

This pull request is the beginning of the implementation of a Cost of Goods Sold feature in WooCommerce, it introduces the code and REST API changes to support programmatically setting and getting one single cost value for any product.

#### Changes in code API

* A new feature is incorporated (handled by `FeaturesController`): `cost_of_goods_sold`
* The following methods are added to the `WC_Abstract_Legacy_Product` class:
  * `set_cogs_value( float $value ): void`
  * `get_cogs_value(): float`
  * `get_cogs_effective_value(): float`
  * `get_cogs_total_value(): float`
* The following methods are added (additionally to the ones above) to the `WC_Product_Variation` class:
  * `get_cogs_value_overrides_parent(): bool`
  * `set_cogs_value_overrides_parent( bool $value )`
* The `WC_Product_Data_Store_CPT` and `WC_Product_Variation_Data_Store_CPT` get the modifications needed to persist the new values (see below).

Currently `get_cogs_value`, `get_cogs_effective_value` and `get_cogs_total_value` return all the same value, with one exception: in products that are variations, `get_cogs_effective_value` and `get_cogs_total_value` return the stored value for the variation added to the stored value for the parent product, unless `set_cogs_value_overrides_parent(true)` was executed.

#### Storage

The value set by `set_cogs_value` is stored as product meta using the `_cogs_total_value` key, but only if the feature is enabled and the value is not zero (a missing meta value for a product will be treated as equivalent to a value of zero).

The value set by `set_cogs_value_overrides_parent` is stored as product meta using the `_cogs_value_overrides_parent` key, but only if the feature is enabled and the value is `true` (a missing meta value for a product will be treated as equivalent to a value of `false`).

#### Changes in REST API

The following is added to the REST API schema for products and variations (`/wp-json/wc/v3/products/<product id>` and `/wp-json/wc/v3/products/<product id>/variations` endpoints, for both `GET` and `POST` requests):

```json
{
    "cost_of_goods_sold": {
        "values": [
            {
                "defined_value": <float>
                "effective_value": <float>
            }
        ],
        "total_value": <float>,
        "defined_value_overrides_parent": <bool>
    }
}
```

`defined_value_overrides_parent` is present only for variations. `effective_value` and `total_value` are read-only fields.

[...]

PR #54458, opened 2025-01-14
Make the defined Cost of Goods Sold value nullable

Konamiman:

When the programmatic API for the Cost of Goods Sold feature was introduced for products in https://github.com/woocommerce/woocommerce/pull/51675, variations had an "additive" flag set by default that meant "the total cost of the variation is the variation's effective cost added to the parent's effective cost". Cost was zero by default too, so that meant that the default total cost for variations was "parent cost plus zero", effectively inheriting the parent product cost. This was on purpose, and  that's a sensible approach for handling the variation cost defaults.

However, later on https://github.com/woocommerce/woocommerce/pull/53845 was merged, which made the "additive" flag reset by default. This poses a problem that was overlooked back then: now the default actual cost value for variations is zero. For a variation to inherit the parent product cost, now the "additive" flag would have to be set to true and thus persisted to the database; this is not ideal for an intended default behavior.

To fix this, this pull request introduces the following changes:

* The defined cost value (the one set via `set_cogs_value`) is now nullable, and is null by default. This is true for all product types.
  * This implies that the data store class will persist the value only when it's not null (previously it was "when it's not zero").
  * By default (as implemented in `WC_Product`) a defined value of null translates to an effective value of zero; this is overridden in variations as detailed below.
  * Trying to set a defined cost value of zero will actually set null (again, this is the default, and it's overridable).
    * This is implemented this way because clients of the API will naturally and rightfully do `set_cogs_value(0)` when willing to remove the cost from a product, and we don't want these useless default values stored in the database.
* For variations with the "additive" flag set, nothing changes: the total cost value is the sum of the parent's and the variation's effective values. Null will be interpreted as zero in this case.
* For variations with the "additive" flag not set (which is still the default):
  * A defined value of null means that the variation's total cost is equal to the parent's effective cost.
  * Any other value (including zero) will be taken as the variation's final, total cost value.

So in sum: for products in general a cost of zero is still the default (although represented as null now), and for variations the default is now truly "inherit the parent value".

Additionally, the calculation of the total value for variations was mistakenly in `get_cogs_effective_value`; it has been moved to `get_cogs_total_value`, where it truly belongs (the effective value must be the same as the defined value until something like costs that are a percent of the price is implemented).

You can test in the command line using `wp shell`:

1. Test that a new simple product will have a defined value of zero by default, and that attempting to set a value of zero will actually set null:

```php
$p = new WC_Product_Simple();
echo $p->get_cogs_value(); //NULL
$p->set_cogs_value(12.34);
echo $p->get_cogs_value(); //12.34
$p->set_cogs_value(0);
echo $p->get_cogs_value(); //NULL
```

2. Test that the cost value is persisted only when it's not null/zero. Continuing the previous `wp shell` session:

```php
$p->set_cogs_value(12.34);
$p->save();
echo get_post_meta($p->get_id(), '_cogs_total_value', true); //12.34

$p->set_cogs_value(null);
$p->save();

global $wpdb; 
echo $wpdb->get_var("SELECT COUNT(1) FROM {$wpdb->postmeta} WHERE meta_key='_cogs_total_value' AND post_id=" . $p->get_id()); // 0
```

3. Now let's assume that you have created a product with id `$product_id` which has a variation with id `$variation_id`. Set a cost value for the parent product, then verify the defaults and the data storage for the variation:

```php
$p=wc_get_product($product_id);
$p->set_cogs_value(10);
$p->save();

$v=wc_get_product($variation_id);
echo $v->get_cogs_value(); //NULL
echo $v->get_cogs_value_is_additive(); //false

$v->set_cogs_value(0);
echo $v->get_cogs_value(); //0 - unlike other product types where it would be NULL
$v->set_cogs_value_is_additive(true);
$v->save();
echo get_post_meta($v->get_id(), '_cogs_total_value', true); //"0"
echo get_post_meta($v->get_id(), '_cogs_value_is_additive', true); //"yes"

$v->set_cogs_value(null);
$v->set_cogs_value_is_additive(false);
$v->save();

echo $wpdb->get_var("SELECT COUNT(1) FROM {$wpdb->postmeta} WHERE (meta_key='_cogs_total_value' OR meta_key='_cogs_value_is_additive') AND post_id=" . $v->get_id()); //0
``` 

4. Finally, it's time to test the calculation of the total value for the variation:

```php
echo $v->get_cogs_total_value(); //10 - not additive, defined value of null --> inherited from parent

$v->set_cogs_value(0);
echo $v->get_cogs_total_value(); //0 - not additive, overrides parent value
$v->set_cogs_value(12.34);
echo $v->get_cogs_total_value(); //12.34 - same

$v->set_cogs_value_is_additive(true);
echo $v->get_cogs_total_value(); //22.34 - parent + variation defined values

$v->set_cogs_value(null);
echo $v->get_cogs_total_value(); //10 - in additive mode, a null defined value is interpreted as zero
```

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/includes/abstracts/abstract-wc-product.php
```diff
@@ -2213,30 +2213,62 @@ protected function get_availability_class() {
...
+	 */
+	protected function adjust_cogs_value_before_set( ?float $value ): ?float {
+		return 0.0 === $value ? null : $value;
+	}
+
 	/**
 	 * Get the defined value of the Cost of Goods Sold for this product.
+	 * See set_cogs_value.
 	 *
 	 * WARNING! If the Cost of Goods Sold feature is disabled this method will always return zero.
```
coreymckrill:
```suggestion
	 * WARNING! If the Cost of Goods Sold feature is disabled this method will always return null.
```

plugins/woocommerce/includes/class-wc-product-variation.php
```diff
@@ -628,19 +627,40 @@ public function set_cogs_value_is_additive( bool $value ): void {
...
+	}
 
-		return parent::get_cogs_value();
+	/**
+	 * Get the Cost of Goods Sold effective value of the parent product.
+	 *
+	 * @return float Cost of Goods Sold effective value of the parent product.
+	 */
+	public function get_parent_cogs_effective_value(): float {
+		return (float) get_post_meta( $this->get_parent_id(), '_cogs_total_value', true );
```
coreymckrill: Given that in `WC_Product`, cogs total value is filterable via `woocommerce_get_product_cogs_total_value`, it seems like maybe instead of getting this value straight from the post meta table it should be doing something like this?

```
$parent_product = wc_get_product( $this->get_parent_id() );
return $parent_product ? $parent_product->get_cogs_total_value() : 0; 
```

Wdyt?

coreymckrill (commented):
👍 Worked well testing locally. I tweaked the testing instructions a tiny bit because there was one spot where a `$p` should have been a `$v` in the shell commands.

One comment/question to discuss, but otherwise this looks good.

plugins/woocommerce/includes/class-wc-product-variation.php
Konamiman: It's a bit confusing because of the different values involved, but note that what we are obtaining here is the _effective_ value, not the _total_ value which is what gets filtered. So the correct snippet would be:

```php
$parent_product = wc_get_product( $this->get_parent_id() );
return $parent_product ? $parent_product->get_cogs_effective_value() : 0; 
```

As of now the effective value is the same as the defined value, except that null is interpreted as zero (with the exception of the values for the variations themselves) and is not affected by the `woocommerce_get_product_cogs_total_value` filter. Reading directly from the postmeta table is just a shortcut for what the data store does, without having to read the entire product object from the database.

That said: there _is_ actually a hook in the data store to alter the retrieved cost value: `woocommerce_load_product_cogs_value`, this is the one that should be used here. I'll do the change (and document a bit what's going on).

Konamiman: Implemented in https://github.com/woocommerce/woocommerce/commit/961438cca19939c8bdb9e30d4e5fc199c732a32f. In the end, and to avoid duplicating code, I went with the suggested snippet but with the addition of some caching (since we are already caching all the parent data anyway).

coreymckrill (approved):

PR #54466, opened 2025-01-14
[e2e tests] Fix site reset incorrectly reported as successful in global setup

### LINKED ISSUE #53107: [e2e tests] Site reset incorrectly reported as successful
The site reset was implemented by #51720 as a plugin that exposes an API endpoint [called in the global setup](https://github.com/woocommerce/woocommerce/blob/afd3469e79f837e9f3d14f8eadd6adbe6273b581/plugins/woocommerce/tests/e2e-pw/global-setup.js#L39) of the e2e tests.

The code considers a successful call also a successful site reset which is incorrect. A 403, 404, even 500 response will be reported as successful reset.

```
Resetting site...
Reset successful: 403
```

```
Resetting site...
Reset successful: 404
```

Fix the implementation by checking the response http code and correctly logging the reset outcome.

adimoldovan:

Closes https://github.com/woocommerce/woocommerce/issues/53107

Fixes the incorrect logging of site reset as successful when the response is a not ok code.

Run any e2e test against a site that has the cleanup plugin missing or deactivated. The reset should fail in the global setup, with a 404 response. The reset failure is logged as error.

### DIFF
--- a/plugins/woocommerce/tests/e2e-pw/global-setup.js
+++ b/plugins/woocommerce/tests/e2e-pw/global-setup.js
@@ -36,7 +36,12 @@ module.exports = async ( config ) => {
 						},
 					}
 				);
-				console.log( 'Reset successful:', response.status() );
+
+				if ( response.ok() ) {
+					console.log( 'Reset successful:', response.status() );
+				} else {
+					console.error( 'ERROR! Reset failed:', response.status() );
+				}
 			} catch ( error ) {
 				console.error(
 					'Reset failed:',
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/tests/e2e-pw/global-setup.js
```diff
@@ -36,7 +36,12 @@ module.exports = async ( config ) => {
 						},
 					}
 				);
-				console.log( 'Reset successful:', response.status() );
+
+				if ( response.ok() ) {
+					console.log( 'Reset successful:', response.status() );
+				} else {
+					console.error( 'ERROR! Reset failed:', response.status() );
+				}
```
lanej0, line 44: Does it make sense to abort the test run if the reset fails?  The chances that the test run will succeed or yield accurate results is pretty low.  

Your PR is to just report the correct status -- which this does, so 👍🏼  -- but something to consider for the future.

lanej0 (approved):

PR #54468, opened 2025-01-14
Update uses of `VALIDATION_STORE_KEY` and use the `StoreDescriptor` instead

### LINKED ISSUE #54428: Update uses of `VALIDATION_STORE_KEY` and use the `StoreDescriptor` instead.
After some `@wordpress/data` types were updated our codebase started to show errors with `useSelect` and `useDispatch`. The cure for this is to use the types that are now provided with `@wordpress/data` and to update uses of the string `wc/store/validation` (often imported as `VALIDATION_STORE_KEY` and use the `StoreDescriptor` (i.e. the result of calling `createReduxStore`) instead.

Please see [this example PR](https://github.com/woocommerce/woocommerce/pull/54376) with some changes for the `wc/store/store-notices`  store.

Where we would previously use:

```js
import { VALIDATION_STORE_KEY } from '@woocommerce/block-data';

const {... } = useDispatch(
		VALIDATION_STORE_KEY
	);

const { ... } = useSelect ( ( select ) => select( VALIDATION_STORE_KEY ).someSelector() );
```

we would now do it like so:

```js
import { validationStore } from '@woocommerce/block-data';

const {... } = useDispatch(
		validationStore
	);

const { ... } = useSelect ( ( select ) => select( validationStore ).someSelector() );
```

opr:

This PR updates instances where we used the key to access the validation data store and instead uses the `StoreDescription`. It also adds dependency arrays to `useSelect` calls as is the best practice.

I left instances of `VALIDATION_STORE_KEY` in [`FormSubmit`](https://github.com/woocommerce/woocommerce/blob/aba5aa590e70c684d3bd246931986be7272bc688/plugins/woocommerce-blocks/assets/js/base/context/providers/add-to-cart-form/form/submit/index.js#L23) and [`AddToCartFormStateContextProvider`](https://github.com/woocommerce/woocommerce/blob/ec9b7852f9ecfe98a0bce387eac7fb0c64eee61e/plugins/woocommerce-blocks/assets/js/base/context/providers/add-to-cart-form/form-state/index.js#L18) since these are JS files anyway and wouldn't benefit from TS, but also because I couldn't test the changes. Maybe someone from @woocommerce/woo-fse  could help me out on how to test those?

The testing steps here are quite extensive, so I recommend two people review this. @samueljseay since you are chosen as the main reviewer, I suggest you delegate half the testing steps to someone else.

Closes #54428

#### `checkout-shipping-method-block` 1

1. Go to WC -> Settings -> Shipping -> Local Pickup and enable it. Add some local pickup options.
2. Ensure you have some regular shipping methods set up too, remember which country has available options.
3. Remove the "Rest of the world" options, ensure no options are enabled for this zone.
4. Add an item to your cart and go to the Checkout block.
5. In the shipping method selection, choose "Shipping". Change address to a location with no shipping options.
6. Try to check out. Ensure you see an error: `Sorry, this order requires a shipping option.` and that you cannot check out.
7. Change address to one with valid shipping options and check out successfully.

https://www.loom.com/share/7c9571e30cd24cb9a049024ebce9317e

#### `checkout-shipping-method-block` 2

*This test is required because this block uses the validation data store to determine whether to show "calculated with an address" or a value in the "Shipping" selector. If the address has validation errors, then it'll show the message, if not it'll show the lowest price.*

1. Go to WC -> Settings -> Shipping -> Local Pickup and enable it. Add some local pickup options.
3. Ensure you have some regular shipping methods set up too, remember which country has available options.
4. Edit your Checkout block, select the "Shipping / Local pickup" selectors and enable the "Show costs" option

Screenshot 2025-01-14 at 14 36 16

9. Log out and add an item to the cart and go to the Checkout block.
10. Ensure that the Shipping selector says "calculated with an address" and that the local pickup one says "FREE"
11. Enter an address for a country with shipping methods (that you set up in step 2).
12. Ensure the "Shipping" selector shows "FROM xxx" where xxx is the lowest price

#### `validateCountry` and `validateState`

1. Go to WC -> Settings -> General. Set `Default customer location` to "No location by default".
2. Log out and add an item to your cart, then go to the Checkout block.
3. Try to check out without selecting a country
4. Ensure you see an error on the country input.
5. Choose USA and ensure the error disappears
6. Try to check out without entering a state.
7. Ensure you see an error on the state input.
8. Select a state, enter a full address and ensure you can check out.

#### `ValidationInputError`

1. Go to the Checkout block, execute the following code in the console:

```js
const container = document.createElement('div');
document.body.appendChild(container);
const ValidationInputError = window.wc.blocksCheckout.ValidationInputError;
const root = ReactDOM.createRoot(container);
root.render(React.createElement(ValidationInputError, { propertyName: 'test-error' } ) ) ;
wp.data.dispatch( wc.wcBlocksData.validationStore ).setValidationErrors( {'test-error': { message: 'Test error', hidden: false } } )
```

2. Scroll to the bottom of the page and ensure you see a red error message.
3. Execute `wp.data.dispatch( wc.wcBlocksData.validationStore ).clearValidationErrors()`
4. Ensure it disappears.

#### `ShippingCalculatorAddress`

1. Go to WC -> Settings -> Shipping Settings -> check Enable the shipping calculator on the Cart Page.
2. Log out and go to the Cart block.
3. Open the shipping calculator and enter each text box except `Country`
4. When you leave the field, ensure you see an error like so:

image

5. Enter text into each box, ensuring the error disappears on each one, but leave `Country` without a value. Press `Check delivery options`.
6. Ensure there is an error on `Country`.
7. Select a country and ensure the error disappears. 
8. Press `Check delivery options`. Ensure the form is hidden without errors.

#### `TotalsCoupon`

1. Go to the Cart block, enter an invalid coupon and press submit.
2. Ensure you see an error message.
3. Leave the input and ensure the message disappears.
4. Repeat on the Checkout block.

#### `ValidatedTextInput`

1. Add an item to your cart and go to the Checkout block.
2. In the shipping address, enter text in the `Address` field then leave the field.
3. Ensure no error appears. Go back and delete the text then leave the field.
4. Ensure an error appears:

image

5. Repeat for first name, last name, postcode, and state (choose UK to enter a state as text).
6. Repeat on billing address.

#### `Select`

1. Install Additional Checkout Fields Tester
[additional-checkout-fields-tester-main (6).zip](https://github.com/user-attachments/files/18413698/additional-checkout-fields-tester-main.6.zip)
2. Log out, add an item to your cart and go to the Checkout page.
3. Fill in all details except the `What is your main role at your company?` select.
4. Try to check out, ensure that field shows an error.
5. Select an option and ensure the error disappears and you can check out.

#### `useValidation`

1. Download and install WooCommerce EU VAT Number.
2. Go to WC -> Settings and check ` Enable tax rates and calculations`
3. Add an item to your cart and go to the Checkout block.
4. Change your country to France (or any other EU country) and scroll to the bottom of the page, the EU VAT Input should be there.
5. Enter invalid text into the box and ensure a validation error shows.
6. Enter a valid VAT Number for the country (e.g. `FR09445330103` for France) Ensure the error disappears.

#### `CheckoutEventsContext`

1. Go to `assets/js/extensions/payment-methods/bacs/index.js` change the contents to:

```js
/**
 * External dependencies
 */
import { registerPaymentMethod } from '@woocommerce/blocks-registry';
import { __ } from '@wordpress/i18n';
import { getPaymentMethodData } from '@woocommerce/settings';
import { decodeEntities } from '@wordpress/html-entities';
import { sanitizeHTML } from '@woocommerce/utils';
import { RawHTML, useEffect } from '@wordpress/element';

/**
 * Internal dependencies
 */
import { PAYMENT_METHOD_NAME } from './constants';

const settings = getPaymentMethodData( 'bacs', {} );
const defaultLabel = __( 'Direct bank transfer', 'woocommerce' );
const label = decodeEntities( settings?.title || '' ) || defaultLabel;

/**
 * Content component
 */
const Content = ( { eventRegistration } ) => {
	useEffect( () => {
		const unsubscribe = eventRegistration.onCheckoutValidation( () => {
			return {
				type: 'failure',
				errorMessage: 'error!',
				validationErrors: {
					shipping_first_name: {
						message: 'Shipping first name should become invalid',
						hidden: false,
					},
				},
			};
		} );
		return unsubscribe;
	}, [] );
	return <RawHTML>{ sanitizeHTML( settings.description || '' ) }</RawHTML>;
};

/**
 * Label component
 *
 * @param {*} props Props from payment API.
 */
const Label = ( props ) => {
	const { PaymentMethodLabel } = props.components;
	return <PaymentMethodLabel text={ label } />;
};

/**
 * Bank transfer (BACS) payment method config object.
 */
const bankTransferPaymentMethod = {
	name: PAYMENT_METHOD_NAME,
	label: <Label />,
	content: <Content />,
	edit: <Content />,
	canMakePayment: () => true,
	ariaLabel: label,
	supports: {
		features: settings?.supports ?? [],
	},
};

registerPaymentMethod( bankTransferPaymentMethod );
```

2. Go to WC -> Settings -> Payment and enable BACS.
3. Add an item to the cart and go to the Checkout block.
4. Open the shipping address.
5. Try to place the order with BACS as your payment method.
6. Ensure you see a validation error on the First Name field.
7. In the code above, replace `type: 'failure'` with `type: 'error'` and replace `onCheckoutValidation` with `onPaymentProcessing`.
8. Repeat steps 4, 5, and 6.

#### Checkout terms block

1. Go to the Checkout block in the editor
2. Click on the `You must accept our Terms and Conditions and PrivacyPolicy to continue with your purchase.` block and in the inspector enable `Require checkbox`.
3. Add an item to your cart and go to the Checkout block.
4. Try to check out without ticking the terms and conditions box.
5. Ensure the text goes red and you can't check out.
6. Check the box and ensure you can check out successfully.

#### Force editing shipping/billing cards

1. Place an order as a logged in customer.
2. As an admin, edit the customer (WP -> Users) and delete their "City" in _billing_ address.
3. Log back in as the user and go to the Checkout block. Ensure the Billing Address is expanded and is not displaying as a card.
4. As an admin, edit the customer (WP -> Users) and delete their "City" in _shipping_ address.
5. Log back in as the user and go to the Checkout block. Ensure the Shipping Address is expanded and is not displaying as a card.

#### Delayed account creation form

1. Go to WC -> Settings -> Accounts & Privacy.
2. Check `Enable guest checkout (recommended)` and `Allow customers to create an account -> After checkout`
3. Uncheck `Send password setup link (recommended)`
4. As a logged-out user, add an item to your cart and go to the Checkout block.
5. Place the order.
6. On the order confirmation screen click into the password box then out of it without entering anything. Ensure you see an error.
7. Enter a strong password and ensure the error goes away and you can create the account.

#### `validateDirtyProps`

1. Add an item to your cart and go to the Checkout block.
2. Open the network tab of dev tools.
3. In the shipping address, fill in a full address and watch a batch request be made to update the address.
4. Edit the Address field to be empty.
5. Ensure no batch request is made.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/validate-country.ts
+++ b/plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/validate-country.ts
@@ -4,7 +4,7 @@
 import { __ } from '@wordpress/i18n';
 import { type ShippingAddress } from '@woocommerce/settings';
 import { select, dispatch } from '@wordpress/data';
-import { VALIDATION_STORE_KEY } from '@woocommerce/block-data';
+import { validationStore } from '@woocommerce/block-data';
 
 // If it's the shipping address form and the user starts entering address
 // values without having set the country first, show an error.
@@ -14,18 +14,18 @@ const validateCountry = (
 ): void => {
 	const validationErrorId = `${ addressType }_country`;
 	const hasValidationError =
-		select( VALIDATION_STORE_KEY ).getValidationError( validationErrorId );
+		select( validationStore ).getValidationError( validationErrorId );
 
 	if (
 		! values.country &&
 		( values.city || values.state || values.postcode )
 	) {
 		if ( hasValidationError ) {
-			dispatch( VALIDATION_STORE_KEY ).showValidationError(
+			dispatch( validationStore ).showValidationError(
 				validationErrorId
 			);
 		} else {
-			dispatch( VALIDATION_STORE_KEY ).setValidationErrors( {
+			dispatch( validationStore ).setValidationErrors( {
 				[ validationErrorId ]: {
 					message: __( 'Please select your country', 'woocommerce' ),
 					hidden: false,
@@ -35,9 +35,7 @@ const validateCountry = (
 	}
 
 	if ( hasValidationError && values.country ) {
-		dispatch( VALIDATION_STORE_KEY ).clearValidationError(
-			validationErrorId
-		);
+		dispatch( validationStore ).clearValidationError( validationErrorId );
 	}
 };
 
--- a/plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/validate-state.ts
+++ b/plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/validate-state.ts
@@ -3,7 +3,7 @@
  */
 import { dispatch, select } from '@wordpress/data';
 import { KeyedFormField, ShippingAddress } from '@woocommerce/settings';
-import { VALIDATION_STORE_KEY } from '@woocommerce/block-data';
+import { validationStore } from '@woocommerce/block-data';
 import { __, sprintf } from '@wordpress/i18n';
 import isShallowEqual from '@wordpress/is-shallow-equal';
 
@@ -32,7 +32,7 @@ export const validateState = (
 ) => {
 	const validationErrorId = `${ addressType }_state`;
 	const hasValidationError =
-		select( VALIDATION_STORE_KEY ).getValidationError( validationErrorId );
+		select( validationStore ).getValidationError( validationErrorId );
 	const isRequired = stateField.required;
 
 	const lastAddress =
@@ -46,12 +46,12 @@ export const validateState = (
 	if ( hasValidationError ) {
 		if ( ! isRequired || values.state ) {
 			// Validation error has been set, but it's no longer required, or the state was provided, clear the error.
-			dispatch( VALIDATION_STORE_KEY ).clearValidationError(
+			dispatch( validationStore ).clearValidationError(
 				validationErrorId
 			);
 		} else if ( ! addressChanged ) {
 			// Validation error has been set, there has not been an address change so show the error.
-			dispatch( VALIDATION_STORE_KEY ).showValidationError(
+			dispatch( validationStore ).showValidationError(
 				validationErrorId
 			);
 		}
@@ -62,7 +62,7 @@ export const validateState = (
 		values.country
 	) {
 		// No validation has been set yet, if it's required, there is a country set and no state, set the error.
-		dispatch( VALIDATION_STORE_KEY ).setValidationErrors( {
+		dispatch( validationStore ).setValidationErrors( {
 			[ validationErrorId ]: {
 				message: sprintf(
 					/* translators: %s will be the state field label in lowercase e.g. "state" */
--- a/plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/address.tsx
+++ b/plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/address.tsx
@@ -38,8 +38,7 @@ const ShippingCalculatorAddress = ( {
 				isCustomerDataUpdating:
 					select( CART_STORE_KEY ).isCustomerDataUpdating(),
 			};
-		},
-		[]
+		}
 	);
 
 	const validateSubmit = () => {
--- a/plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/address.tsx
+++ b/plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/address.tsx
@@ -6,7 +6,7 @@ import Button from '@woocommerce/base-components/button';
 import { useState } from '@wordpress/element';
 import isShallowEqual from '@wordpress/is-shallow-equal';
 import type { ShippingAddress, FormFields } from '@woocommerce/settings';
-import { VALIDATION_STORE_KEY, CART_STORE_KEY } from '@woocommerce/block-data';
+import { validationStore, CART_STORE_KEY } from '@woocommerce/block-data';
 import { useDispatch, useSelect } from '@wordpress/data';
 import { useFocusReturn } from '@woocommerce/base-utils';
 /**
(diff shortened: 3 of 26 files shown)


samueljseay (approved):
1. Code changes look good
2. One thing that slightly bugs me is that this increases the bundle size by 700B, which I kinda expected just due to the import changing from the string. I don't think we should spend the time trying to fix this, but we'll likely see a similar increase for each store we make this change on
3. One thing that would have made these changes much less risky to make is excluding changing the deps to `useSelect`, I updated the description of the PR to mention this (and it would be good to have in changelog as well). It's definitely best to have the deps but they're probably the thing introducing the most actual risk to runtime changes here. Just mentioning that :) 
4. I went through all the test steps but just had 2 small issues (and I suspect 2 is just some issue with the test code)

1. I couldn't test eu vat plugin. it says my server needs SOAP support. Do you know how to fix this?
2. The second case of testing bacs with modified code `onPaymentProcessing` would not trigger correctly for me like the first case did.

Everything else tested well.

opr:
@samueljseay 

> One thing that would have made these changes much less risky to make is excluding changing the deps to useSelect, I updated the description of the PR to mention this (and it would be good to have in changelog as well). It's definitely best to have the deps but they're probably the thing introducing the most actual risk to runtime changes here. Just mentioning that :)

I think you're right, is it worth going back. andremoving this change and shelfing it for later, or change it on a "if you're in the file and notice it" basis?

opr:
> I couldn't test eu vat plugin. it says my server needs SOAP support.

Not sure how I did it, it's a PHP extension I think so you'll need to install it and enable it, alternatively you could use live branches to test this on Pressable?

opr:
I removed the deps I added in [c8eafe6](https://github.com/woocommerce/woocommerce/pull/54468/commits/c8eafe63d1f61e98bf1066ffa1f71654848bf671) so if we want to re-add them in a separate effort where we fully explore the runtime implications it'll be easy to revert that commit.

opr:
> The second case of testing bacs with modified code onPaymentProcessing would not trigger correctly for me like the first case did.

I just tested that one out again and it worked OK for me, I don't think there's any reason for it not to work on yours. I am happy to merge without you being able to replicate it, I think we'd see failures during Stripe checkout if it wasn't working properly anyway.

opr:
@samueljseay just requested one more approval after agreeing that removing the changes to `useSelect` deps is OK - so it's purely a store key to store descriptor PR now.

samueljseay (approved):
Looking great, now that dep changes are removed this could barely be considered runtime change and is safe to ship imo. 🚢

Issue #52289, opened 2024-10-24
Modern-looking transactional emails


NeosinneR:
Update transactional email styles and the layout for a modern, refreshed look. 

![Image](https://github.com/user-attachments/assets/67f5b424-252d-42ed-8588-f7c63e0295e4)

PR #54491, opened 2025-01-14
Modernized email layout styles

### LINKED ISSUE #53738: Modern-looking email layout
Modernize email heading. This includes:
- Logo
- Main heading
- Description
- Footer
- Overall layout

Figma: FAA2sMXXfT8Qs40AC3V1gL-fi-569_14241

Image

NeosinneR:

Part of #52289. Depends on #54440 (must be merged first, before merging this PR). 

This PR modernizes the layout of core emails. Closes #53738.

| Screenshot 2025-01-14 at 18 27 33 | Screenshot 2025-01-14 at 18 22 08 |
|-------------------------|-------------------------|

1. Enable the `Email improvements` experimental feature in **WooCommerce > Settings > Advanced > Features**.
2. Go to **WooCommerce > Settings > Email**.
3. Check any email in the email preview that they are rendered properly and no information is missing (compared to the current state of `trunk`).


plugins/woocommerce/templates/emails/email-styles.php
```diff
@@ -127,8 +127,13 @@
 	background-color: inherit;
 }
 
-<?php if ( $header_alignment ) : ?>
+<?php if ( $email_improvements_enabled ) : ?>
```
alex-mpoet, line 130: The check for `$header_alignment` non-emptiness was removed. In theory, this variable can be empty if the respective option and transient aren't present, so the `text-align` property [below](https://github.com/woocommerce/woocommerce/pull/54491/commits/c476dc92cae09820fa5b0e1b1365b1f38c617b1b#diff-0f841c5365b79f78f7ad1a0e3714b26b02cbc5106fab762346319932438c6160L132) may end up being invalid. I think we should add a fallback value.

plugins/woocommerce/templates/emails/admin-cancelled-order.php
```diff
@@ -51,7 +51,11 @@
  * Show user-defined additional content - this is set in each email's settings.
  */
 if ( $additional_content ) {
-	echo wp_kses_post( wpautop( wptexturize( $additional_content ) ) );
+	?>
+	<div class="email-additional-content">
```
alex-mpoet: This doesn't seem to be under the feature flag, but the HTML structure is changed which we planned to avoid as much as possible. Can this wrapping be conditional? This applies to all similar cases.

plugins/woocommerce/templates/emails/admin-cancelled-order.php
```diff
@@ -24,8 +24,10 @@
 */
 do_action( 'woocommerce_email_header', $email_heading, $email ); ?>
 
+<div class="email-introduction">
```
alex-mpoet: Is it totally safe to change HTML structure of legacy emails and not putting this under the feature flag? What if, say, customers have parsers set up which react on specific email content, they might be broken after that.

plugins/woocommerce/templates/emails/email-styles.php
```diff
@@ -414,7 +415,8 @@
 	max-width: 100%;
 }
 
-.email-order-detail-heading span {
+h2.email-order-detail-heading span {
+	color: <?php echo esc_attr( $footer_text ); ?>;
```
alex-mpoet, line 425: This isn't conditional, is this intentional? It seems to be applied to the old templates, but why then a new color property is added?

alex-mpoet (commented):
@NeosinneR I have a few notes. 

Also, just to make sure, do the new layout and order details need to be tested in Litmus?

plugins/woocommerce/templates/emails/email-styles.php
NeosinneR: The `email-order-detail-heading` class is added conditionally, so this CSS is only applied when the feature flag is enabled. The new color is to differentiate subtitle in order summary:

Screenshot 2025-01-16 at 09 27 48

plugins/woocommerce/templates/emails/email-styles.php
NeosinneR: Good point. Added fallback when feature flag is on (when it's off, there is a check for non-emptiness). Fixed in 726b9c1.

plugins/woocommerce/templates/emails/admin-cancelled-order.php
NeosinneR: Good thinking. I specifically didn't add a feature flag here, because the CSS is flagged, but you're right that this would still change the HTML structure. Addressed in e8e3bff.

plugins/woocommerce/templates/emails/admin-cancelled-order.php
NeosinneR: The same as above, addressed in e8e3bff.

NeosinneR:
@alex-mpoet it's rebased.

> do the new layout and order details need to be tested in Litmus?

Yes. I haven't yet done it, and since we keep it behind a feature flag for one more release, it's not super urgent to do now. I've created #54572 so we don't forget about that.

alex-mpoet:
The "Customer note" email template produces the following error when previewed:

> Deprecated: preg_split(): Passing null to parameter #2 ($subject) of type string is deprecated in /www/wp-includes/formatting.php on line 3069

NeosinneR:
@alex-mpoet where do you see this error? I can't replicate it locally.

alex-mpoet:
@NeosinneR I can reproduce it both locally and on jurassic.ninja. `WP_DEBUG` needs to be enabled.

alex-mpoet (approved):
Thank you for the fixes! Looks good!

PR #54493, opened 2025-01-14
Update $slug.php.mustache

georgybu:
fix import Setup file

Was fixed import issues that happens in Linux.
Tested with `Ubuntu 22.04` and `Debian 12`

image


nerrad:
There have been several changes to the create-woo-extension since this PR was created and it's likely the issue this PR fixes no longer exists. I'm going to close as a result but if this is still occurring definitely report! Worth noting that with the advent of AI, scaffolding tools like this one are becoming less and less relevant.

Issue #54161, opened 2025-01-03
[Checkout block] In the editor the shipping option are displaying wrong data


ralucaStan:
### Prerequisites

- [ ] I have carried out troubleshooting steps and I believe I have found a bug.
- [ ] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

When I have at least one shipping method defined, the editor for the Checkout blocks shows confusing options for the Shipping options inner block. This was reported in a bigger ticket  #48939 for the checkout block that is not on the default checkout page, but the behaviour is similar for the default checkout page as well.

### Expected behavior

The options shown in the editor should reflect the settings. People trying out the checkout block would like to see the end result, and with the current experience we might be impacting adoption

We need a team discussion on the expected result because the **shipping options** might be region-based, so the merchant viewing the page shouldn't in theory see all of the shipping options. 

The **local pickup options** are straightforward to show, as those aren't region based. 

@pierorocca @woocommerce/rubik 

### Actual behavior

###1.  When no shipping options/local pickup are entered the editor shows correctly this: 

<img width="614" alt="Image" src="https://github.com/user-attachments/assets/38d05d9f-d567-420b-961f-8e651eb71a05" />

### 2. When I add one shipping option and no local pickup. 

<img width="1407" alt="Image" src="https://github.com/user-attachments/assets/48070184-cd49-409a-adeb-89e3108ec242" />
 
I see
<img width="684" alt="Image" src="https://github.com/user-attachments/assets/2b4e71af-f7ba-4548-906d-f87f4f3d6308" />
- flat rate: which I did not add
- free shipping: this might be a dummy one, that I didn't add
- 2 local pickup options (local pickup showing has a different ticket [#42098](https://github.com/woocommerce/woocommerce/issues/42098))

### 3. When I add a local pickup option and no shipping options I see
<img width="797" alt="Image" src="https://github.com/user-attachments/assets/e53cb5a3-4611-47d9-914a-b3191ac16444" />

- flat rate: which I did not add
- free shipping which I did not add
- no local pickup option

### Steps to reproduce

1. Set up a new store
2. See the editor for the checkout block, notice the result from the section 1. from above
3. Add 1 shipping option and see the editor for the checkout block, notice the result from the section 2. from above
4. Remove the option and add a local pickup option
5. notice the result from the section 3. from above

### WordPress Environment

n/a

### Isolating the problem

[...]

PR #54533, opened 2025-01-15
Add real local pickup preview data to Checkout Block Editor

### LINKED ISSUE #42098: Local pickup functionality is visible in the page editor, even when no pickup locations have been added or activated
## Describe the bug

The Local Pickup option remains visible in the page editor's Checkout block, even when no active pickup locations are set up. This issue arises when local pickup is enabled, but either no locations are added or the added locations are not activated.

## To reproduce

1. Go to `/wp-admin/admin.php?page=wc-settings&tab=shipping&section=pickup_location`.
2. Enable local pickup.
3. Add one pickup location, but do not enable it.
4. Create a test page and add the Checkout block.
5. Select the shipping method `Local Pickup`.
6. See that the Local Pickup functionality is visible.
7. Go back to `/wp-admin/admin.php?page=wc-settings&tab=shipping&section=pickup_location`.
8. Delete the pickup location added in step 3.
9. Go back to the test page with the Checkout block.
10. Select the shipping method `Local Pickup`.
11. See that the Local Pickup functionality is still visible.

## Current behavior

Currently, the Local Pickup functionality is visible in the page editor, even when no pickup locations have been added or when none of the added pickup locations have been activated.

## Expected behavior

I expect the Local Pickup functionality only to be visible when local pickups are enabled and when at least one active pickup location exists.

### No activated pickup locations

Screenshot 2023-11-15 at 13 49 37

### No pickup locations available

Screenshot 2023-11-15 at 13 53 21

mikejolley:

Updates the local pickup section in the editor to show real store data instead of sample data. This also means that if local pickup is not configured (no locations) the preview is helpful.

Closes #42098

1. Open up 2 tabs for easy testing. WC > Settings > Shipping > Local Pickup, and the editor for the Checkout Block
2. Disable Local Pickup. Confirm there is no local pickup in the editor preview.
3. Enable Local Pickup without any pickup locations. Confirm there is no local pickup in the editor preview.
4. Create some pickup locations. Confirm they are shown in the editor preview.
5. Toggle the "enabled" checkbox for all locations. Confirm there is no local pickup in the editor preview.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce-blocks/assets/js/previews/shipping-rates.ts
+++ b/plugins/woocommerce-blocks/assets/js/previews/shipping-rates.ts
@@ -29,8 +29,8 @@ const localPickupRates = localPickupLocations
 				delivery_time: '',
 				price: '0',
 				taxes: '0',
-				rate_id: `pickup_location:${ index }`,
-				instance_id: index,
+				rate_id: `pickup_location:${ index + 1 }`,
+				instance_id: index + 1,
 				meta_data: [
 					{
 						key: 'pickup_location',
--- a/plugins/woocommerce-blocks/assets/js/previews/shipping-rates.ts
+++ b/plugins/woocommerce-blocks/assets/js/previews/shipping-rates.ts
@@ -20,34 +20,36 @@ const localPickupLocations = getSetting<
 	}[]
 >( 'localPickupLocations', {} );
 
-const localPickupRates = Object.values( localPickupLocations ).map(
-	( location, index: number ) => ( {
-		...API_SITE_CURRENCY,
-		name: __( 'Local pickup #1', 'woocommerce' ),
-		description: '',
-		delivery_time: '',
-		price: '0',
-		taxes: '0',
-		rate_id: `pickup_location:${ index }`,
-		instance_id: index,
-		meta_data: [
-			{
-				key: 'pickup_location',
-				value: location.name,
-			},
-			{
-				key: 'pickup_address',
-				value: location.formatted_address,
-			},
-			{
-				key: 'pickup_details',
-				value: location.details,
-			},
-		],
-		method_id: 'pickup_location',
-		selected: false,
-	} )
-);
+const localPickupRates = localPickupLocations
+	? Object.values( localPickupLocations ).map(
+			( location, index: number ) => ( {
+				...API_SITE_CURRENCY,
+				name: __( 'Local pickup #1', 'woocommerce' ),
+				description: '',
+				delivery_time: '',
+				price: '0',
+				taxes: '0',
+				rate_id: `pickup_location:${ index }`,
+				instance_id: index,
+				meta_data: [
+					{
+						key: 'pickup_location',
+						value: location.name,
+					},
+					{
+						key: 'pickup_address',
+						value: location.formatted_address,
+					},
+					{
+						key: 'pickup_details',
+						value: location.details,
+					},
+				],
+				method_id: 'pickup_location',
+				selected: false,
+			} )
+	  )
+	: [];
 
 export const previewShippingRates: CartResponseShippingRate[] = [
 	{
--- a/plugins/woocommerce-blocks/assets/js/previews/shipping-rates.ts
+++ b/plugins/woocommerce-blocks/assets/js/previews/shipping-rates.ts
@@ -20,6 +20,8 @@ const localPickupLocations = getSetting<
 	}[]
 >( 'localPickupLocations', {} );
 
+const localPickupCost = getSetting< string >( 'localPickupCost', '' );
+
 const localPickupRates = localPickupLocations
 	? Object.values( localPickupLocations ).map(
 			( location, index: number ) => ( {
@@ -27,7 +29,7 @@ const localPickupRates = localPickupLocations
 				name: __( 'Local pickup #1', 'woocommerce' ),
 				description: '',
 				delivery_time: '',
-				price: '0',
+				price: displayForMinorUnit( localPickupCost, 0 ) || '0',
 				taxes: '0',
 				rate_id: `pickup_location:${ index + 1 }`,
 				instance_id: index + 1,
--- a/plugins/woocommerce-blocks/assets/js/previews/shipping-rates.ts
+++ b/plugins/woocommerce-blocks/assets/js/previews/shipping-rates.ts
@@ -3,11 +3,46 @@
  */
 import { __, _x } from '@wordpress/i18n';
 import type { CartResponseShippingRate } from '@woocommerce/types';
+import { getSetting } from '@woocommerce/settings';
 
 /**
  * Internal dependencies
  */
 import { API_SITE_CURRENCY, displayForMinorUnit } from './utils';
+
+// Get local pickup locations from the settings and format into some preview shipping rates for the response.
+const localPickupLocations = getSetting<
+	{
+		enabled: boolean;
+		name: string;
+		formatted_address: string;
+	}[]
+>( 'localPickupLocations', {} );
+
+const localPickupRates = Object.values( localPickupLocations ).map(
+	( location, index: number ) => ( {
+		...API_SITE_CURRENCY,
+		name: __( 'Local pickup #1', 'woocommerce' ),
+		description: '',
+		delivery_time: '',
+		price: '0',
+		taxes: '0',
+		rate_id: `pickup_location:${ index }`,
+		instance_id: index,
+		meta_data: [
+			{
+				key: 'pickup_location',
+				value: location.name,
+			},
+			{
+				key: 'pickup_address',
+				value: location.formatted_address,
+			},
+		],
+		method_id: 'pickup_location',
+		selected: false,
+	} )
+);
 
 export const previewShippingRates: CartResponseShippingRate[] = [
 	{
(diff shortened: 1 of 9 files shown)


nielslange (commented):
Thanks for working on this, @mikejolley. I briefly looked into this PR, based on your comment in https://github.com/woocommerce/woocommerce/issues/42098#issuecomment-2590326890, and noticed that the price is not displayed in the editor:

Screenshot 2025-01-16 at 14 05 32

Apart from that, I also noticed a discrepancy in how the pickup details are displayed:

- **In the editor:** The entire pickup details are shown.
- **On the frontend:** The details are truncated to two lines, with a "Read more" link visible if the content is too long.

mikejolley:
@nielslange I've added price. There are two caveats to this you'll probably notice which I cannot find a good fix for right now. I don't think it blocks adding real rates, but could be improved.

1. When you toggle to local pickup, the "order summary" area will not update pricing based on those rates. It's all hardcoded.
2. As you noted, the Read More functionality doesn't work in the editor. I believe its a bug in that component; it doesn't seem to be able to read the height in the editor. The implementation is correct.

alexflorisca:
I notice Niels is reviewing this, do you need a second pair of eyes on it or are you happy with Niels review?

nielslange:
> @nielslange I've added price. There are two caveats to this you'll probably notice which I cannot find a good fix for right now. I don't think it blocks adding real rates, but could be improved.
> 
> 1. When you toggle to local pickup, the "order summary" area will not update pricing based on those rates. It's all hardcoded.
> 2. As you noted, the Read More functionality doesn't work in the editor. I believe its a bug in that component; it doesn't seem to be able to read the height in the editor. The implementation is correct.

Thanks for clarifying this, @mikejolley. 

> 1. When you toggle to local pickup, the "order summary" area will not update pricing based on those rates. It's all hardcoded.

While it would be great to display the correct prices in the editor, I assume it would require a big refactor. @pierorocca what are your thoughts on this? The problem is, that when displaying the local pickup options with the pickup prices defined by the merchants, these pickup prices will not appear in the order summary in the editor. 

> 2. As you noted, the Read More functionality doesn't work in the editor. I believe its a bug in that component; it doesn't seem to be able to read the height in the editor. The implementation is correct.

I created the initial Read More component for the local pickup section. Due to the dummy data in the editor, I haven't checked the functionality against the editor. There might be either a bug in the way I developed the component or that it's related to the GB DOM within the editor. Personally, I think it's fine to not have the Read More component in the editor for now. We could look into this in the future, but I would not delay this PR due to that.

nielslange:
> I notice Niels is reviewing this, do you need a second pair of eyes on it or are you happy with Niels review?

@alexflorisca It would be great if you also review the PR. Four eyes always spot more than two eyes. 😀

plugins/woocommerce-blocks/assets/js/previews/shipping-rates.ts
```diff
@@ -3,12 +3,56 @@
...
 import type { CartResponseShippingRate } from '@woocommerce/types';
+import { getSetting } from '@woocommerce/settings';
 
 /**
  * Internal dependencies
  */
 import { API_SITE_CURRENCY, displayForMinorUnit } from './utils';
 
+// Get local pickup locations from the settings and format into some preview shipping rates for the response.
+const localPickupLocations = getSetting<
```
alexflorisca: nitpick: can we extract the type into a separate variable? This makes it quite hard to read, where the type is longer than the actual code.

alexflorisca (commented):
Code looks ok and tests well. Left a nitpick comment. 

One thought I had was about these two points

> - When you toggle to local pickup, the "order summary" area will not update pricing based on those rates. It's all hardcoded.
> - As you noted, the Read More functionality doesn't work in the editor. I believe its a bug in that component; it doesn't seem to be able to read the height in the editor. The implementation is correct.

While they are minor inconveniences, I think it's worth trying to address them. We want to get the editor experience to match the front-end as close as possible, and leaving small discrepancies like this might add up with other PRs and other work. Saying that, I imagine it would be a significant amount of work to get the totals to update in the editor though, so we have to balance that obviously. What do you think Mike?

mikejolley:
> Saying that, I imagine it would be a significant amount of work to get the totals to update in the editor though, so we have to balance that obviously. What do you think Mike?

Yes, this would increase the scope by a fair amount. I should also point out, the shipping section is currently also showing sample data so this would need a similar update in order to show dynamic values instead of hardcoded data.

So we'd need to:

1. Update shipping in the same way
2. Remove (or calculate) `totals` in the sample data
3. Update test coverage where it looks for sample values

I'm not 100% sure if the client would work in the same way in the editor in order to track selections/do calculations, so we would also need to look into how this functions or add a suitable editor only replacement cc @senadir 

TBH I don't think we should be doing calculations editor side. I think its good to represent real data where possible (so rates/pickup methods with price display), but I think generating correct taxes and totals on the fly is too much. Instead those sections could show some kinda of placeholder or loading state to avoid showing "wrong" totals or confusing data.

senadir:
Updating the summary seems like a nice to have at best :) If we can make sure strings get updated (collect vs ship and whatnot) that's nice to have, totals shouldn't matter, everyone know this is a fake state given the fake product.

alexflorisca:
Ok I kind of agree, I think the general principle still stands (get editor experience as close to front-end as possible) but I think it's too much to calculate totals in this case. Nadir is right, as long as we update the language that would work.

The readmore functionality would be good to fix so that the block doesn't take up too much vertical space in the case of long store descriptions.

mikejolley:
@alexflorisca As part of this? It feels like a follow-up issue seeing as that component is separate to the pickup locations.

pierorocca:
AFK today but wanted to bring @yjailin into the conversation around the principle of getting the editor as close as possible to the front-end experience.  I agree with that in principle and the balancing of scope vs. what can practically be accomplished. Whatever isn't real, it's hard to tell today what's real vs. dummy data. That's where I think design can provide guidance.

plugins/woocommerce-blocks/assets/js/previews/utils.ts
```diff
@@ -19,16 +19,18 @@ export const API_SITE_CURRENCY = {
 /**
  * Preview data is defined with 2dp. This converts to selected currency settings.
  */
-export const displayForMinorUnit = ( value: string ): string => {
+export const displayForMinorUnit = ( value: string, precision = 2 ): string => {
```
nielslange, line 22: Nitpic: Given that `value` has an explicit type, shouldn't we also provide an explicit type for `precision`?

```suggestion
export const displayForMinorUnit = ( value: string, precision: number = 2 ): string => {
```

I'm thinking about the case in which we call `displayForMinorUnit()` and the value of `precision` is not a number. By adding an explicit type, we can catch such an error during development.

nielslange (approved):
I've successfully tested the PR, and it works as expected. I'm pre-approving it, but left a nitpic question regarding explicit typing. 

Before merging this PR, let's wait for @yjailin's reply to [this comment](https://github.com/woocommerce/woocommerce/pull/54533#issuecomment-2602884189).

alexflorisca:
> @alexflorisca As part of this?

Happy for this to be a follow up issue

plugins/woocommerce-blocks/assets/js/previews/utils.ts
mikejolley: This does not appear to be needed. The type is implied by the default we set (2). Checked to confirm:

Screenshot 2025-01-22 at 14 25 38

pierorocca:
This doesn't seem to work for me with 1) Local Pickup disabled 2) Previously created pickup locations that are enabled. I have to also disable the pickup locations for it to display correctly.

image

image

When enabled it is working as expected:

image

P.S. We're ok to move ahead without design as I already connected with Yoann and the preference was strongly to use real data. That works well here.

mikejolley:
@pierorocca thanks for looking. Fixed in https://github.com/woocommerce/woocommerce/pull/54533/commits/1dda0f142d4b90bb8c74d6fb4bb7ad8897139b8d

> P.S. We're ok to move ahead without design as I already connected with Yoann and the preference was strongly to use real data. That works well here.

Thanks, in that case I'll merge this once @nielslange or @alexflorisca give final approval.

plugins/woocommerce-blocks/assets/js/previews/utils.ts
nielslange: Thanks for clarifying this, @mikejolley. I was thinking about consistency and demonstrating best practices to the community, but you're right that the type is already implied by the default value. I'll mark this suggestion as resolved.

nielslange (approved):
> Thanks, in that case I'll merge this once @nielslange or @alexflorisca give final approval.

All good from my side, @mikejolley. The PR works as expected. Let's ⛴️ this improvement.

While reviewing, I noticed that the Cart block editor is still displaying both the shipping and local pickup dummy data. To address this, I created #54825 for tracking the issue. Fixing it should be handled in a separate PR.

ralucaStan:
Great improvements 👏 
I've created [55068](https://github.com/woocommerce/woocommerce/issues/55068) to also bring merchant-defined local pickup options in the Cart block.
And https://github.com/woocommerce/woocommerce/issues/55071 to update the label we show in the order summary.
Both bugs will need to be fixed in time for WooCommerce 9.8, when the change from this PR is going to be released.

We'll discuss updating the shipping options display in https://github.com/woocommerce/woocommerce/issues/54161

PR #54542, opened 2025-01-15
Sanitize localized shipping method cost

waclawjacek:
An alternative to #54181. It is marked as a draft but is ready for testing. There are no unit tests as I don't know if we will be pursuing this route.

---

Only allow shipping method cost expressed using numbers and the thousand and decimal separators set in WC settings.

If a value is outside of that character set, display a notice.

Reformat these numbers to is_numeric() strings for DB persistence and then reformat them to localized strings for display.


plugins/woocommerce/src/Utilities/NumberUtil.php
```diff
@@ -49,4 +49,36 @@ public static function array_sum( array $arr ): float {
...
+			'/^[0-9\%s\%s]*$/',
+			wc_get_price_thousand_separator(),
+			wc_get_price_decimal_separator()
+		);
+
+		if ( 1 !== preg_match( $allowed_characters_regex, $value ) ) {
+			throw new \InvalidArgumentException( __( 'Please enter a valid number', 'woocommerce' ) );
+		}
+
+		$value = str_replace( wc_get_price_thousand_separator(), '', $value );
```
rjchow, line 75: There's actually a `wc_format_decimal()` function [here](https://github.com/woocommerce/woocommerce/blob/027aff026aa11a0e65c6733556cd9fdb857b9e70/plugins/woocommerce/includes/wc-formatting-functions.php#L288) - I think this function might be better located there.

Issue #47317, opened 2024-05-09
Different handling of optional states between Shortcode and Block Checkouts


senadir:
Countries who declare their state hidden and optional, but still have states, will show up in Checkout shortcode but are not visible in Checkout block. A prime example is Germany.

https://github.com/woocommerce/woocommerce//blob/a66075ee115c3d699c37240d7a597fa64ed175c5/plugins/woocommerce/includes/class-wc-countries.php#L1048-L1056

### Steps to reproduce

1. Sell to all countries.
2. Navigate to Checkout block, select Germany, see that state field is not visible.
3. Navigate to Checkout shortcode, select Germany, see that state field is visible and optional.

ralucaStan:
@senadir it's not clear to me what experience this ticket is targeting. How is it impacted by  https://github.com/woocommerce/woocommerce/pull/47319?

Could you rewrite the ticket using the issue template for bugs?

senadir:
I don't know which outcome is best, so I'm not sure what should be expected behavior, just that we have 2 behaviors here.

When talking to Mike, him and I agreed that hiding the states regardless should be the preferred outcome, but we won't touch shortcode Checkout to fix this.

senadir:
Added a won't fix label.

pmcpinto:
To recap, we currently don't show the state field for countries where the state is optional and don't offer an easy way for users to display this field, correct?

senadir:
To be exact, if states is marked as "hidden", we don't show it.
In shortcode Checkout, if state is marked as "hidden" but still had a predefined list of states, it will be shown regardless.

There isn't any cases of this so far, only Germany, and I already fixed it.

A way for a merchant to show it is to mark the state as visible,

```php
add_filter('woocommerce_get_country_locale', function( $locale ) {
$locale['DE']['state']['hidden'] = false;
}, 10, 1 );
```

It seems this might only happen by mistake that you end up adding states to a country but not enable the state button.

pmcpinto:
Thanks for the clarification. Can we close this one, or is there anything missing?

senadir:
I think we can close it for now, I documented it as a difference between Block and Shortcode

PR #54543, opened 2025-01-15
Add postcode and state configuration for Botswana

daledupreez:

This PR adds baseline address configuration for Botswana, which doesn't have postal codes and doesn't require a district (the approximate equivalent of a state). More specifically, this change specifies the following:
 - For the `postcode` field:
    * the field is not required
    * the field should be hidden
 - For the `state` field:
    * the field is not required
    * the field should be hidden (though this doesn't seem to apply in all contexts)
    * the field should use the `District` label

I picked this up while helping someone with a WooCommerce store based in Botswana.

#### Test setup

1. You need a functioning WooCommerce store with at least one shippable product defined.
2. To fully test the shipping UI, you also need to define a shipping zone and shipping method for shipping to Botswana. You can add that as follows:
  a. Navigate to WooCommerce > Settings
  b. Click on the Shipping tab
  c. Click on "Add Zone"
  d. Enter "Botswana" for the zone name
  e. Type "Botswana" into the "Zone region" field, and ensure that "Botswana" is checked in the UI
  f. Click on "Add shipping method" to add a shipping method
  g. Click on "Local pickup", and then click on "Create and save"

#### Testing

1. As a logged-out user of the site, navigate to the site's shop page.
2. Click on the test product on that page.
3. Add the product to your cart.
4. Navigate to your cart page (normally `/cart/`), which you should be able to do by clicking on the cart in your menu, or by following the "View cart" link that may be displayed after adding the product.
5. Click on the "Enter address to check delivery options" link.
6. Check that there are 4 fields displayed as part of the address. (For me, those are "Country/Region", "City", "State/County", and "Postal code", but may vary slightly depending on language.)
7. From the Country/Region dropdown, select `Botswana`.
8. Verify that only the "Country/Region" and "City" fields remain visible.
9. Enter 'Gaborone' in the City field.
10. Click on "Check delivery options".
11. Verify that your local pickup option is displayed. It should also be selected by default. (If it's not selected, select it.)
12. Click on "Proceed to Checkout".
13. Check that the "Billing Details" section includes fields for Town/City, State/Province (or similar), and Postcode/ ZIP (though exact labels may vary depending on your language and/or locale).
14. Select `Botswana` from the "Country/Region" dropdown.
15. Verify that the Postcode/ZIP field is no longer visible.
16. If your checkout page is built using blocks, verify that the `state` field is not displayed. If your checkout page is using a shortcode, verify that the State/Province field now has the "District" label, and is marked as optional.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Add default address configuration for Botswana

Postcode and state are not required, and both fields are hidden when possible. State is also called "District" for cases where it is still displayed.

</details>

<details>

#### Comment 

</details>


albarin:
Hi @daledupreez thanks for this contribution and the very detailed testing steps 👌 

I'm having some issues after step `14`, I select `Botswana`, but then in step `15` it says `Verify that the State/Province field now has the "District" label, and is marked as optional.`
I'm not seeing the `District` field, but isn't that expected since it is not required and hidden in your changes?

Screenshot 2025-01-16 at 11 29 34

Thank you!

daledupreez:
Thanks for taking a look and testing this, @albarin!

I took another look and I realised that my checkout page was using a shortcode rather than blocks, so that seems to be a bug/gap for that case. Given that, I think it's worth keeping both `hidden: true` and the alternate translation so the correct term is available in that context.

Here is a screenshot of my test environment with `hidden: true` using the checkout shortcode.
Screenshot 2025-01-16 at 21 10 13

albarin:
Thanks, @daledupreez, I did a bit more research and this is a known issue https://github.com/woocommerce/woocommerce/issues/47317

albarin (approved):

PR #54548, opened 2025-01-15
Add/54496 add new rest api to import single step

### LINKED ISSUE #54496: Blueprint - Add a new REST endpoint to import a single step
To provide more detailed import progress and avoid uploading a file to the server, create a new REST endpoint for importing a single step. Update the frontend to parse the schema and send requests to this endpoint to import.

**Benefits**:

- Users don’t need to upload schema files to the server, eliminating the need for saving or cleaning up those files.
- We can provide more detailed progress updates for each step, rather than waiting for all steps to be completed.

Note: We will not implement asynchronous import for now. If we decide to implement it in the future, we can still use the existing endpoints.

Note: To implement a "resume" feature, we can save the MD5 hash and the JSON path position, allowing the import to resume later if the user provides the same schema file.

## Acceptance Criteria

- Add POST `/wp-json/wc-admin/blueprint/import-step` endpoint
- The endpoint should accept a step definition as its payload
- The endpoint should return [JsonResultFormatter](https://github.com/woocommerce/woocommerce/blob/trunk/packages/php/blueprint/src/ResultFormatters/JsonResultFormatter.php#L32)

moon0326:

Closes #54496 

This PR changes the import process to parse the schema file on the frontend.

1. Added a new REST API `/wp-json/wc-admin/blueprint/import-step` to import a single step.(https://github.com/woocommerce/woocommerce/pull/54548/commits/8c67255a5dd0094f528f0cd2b2af25f2ae342ca5)
2. Frontend parses JSON directly in the user's browser without uploading the file to the server (https://github.com/woocommerce/woocommerce/pull/54548/commits/09b4d3baf3a7eddb895450960cb41a7f480ea90d)

Note: I think it would be nice to display real-time progress to users instead of just showing a spinning GIF. I'll bring this up to Veronica.

1. Create a new JN with this branch. Make sure to enable `blueprint` feature flag.
2. Go to `WooCommerce -> Settings -> Advanced -> Blueprint`
3. Click on `Export` button and save the file.
4. Open browser console.
5. Drag and drop the exported file to the dropzone.
6. Click on `Import` button.
7. When prompted with "Your configuration will be overridden", click on `Import`
8. You should see import logs and requests to `/wp-json/wc-admin/blueprint/import-step`

Screenshot 2025-01-15 at 1 09 38 PM

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

This feature is behind a feature flag and not visible to the public by default.
</details>

### DIFF
--- a/plugins/woocommerce/client/admin/client/blueprint/components/BlueprintUploadDropzone.tsx
+++ b/plugins/woocommerce/client/admin/client/blueprint/components/BlueprintUploadDropzone.tsx
@@ -54,49 +54,85 @@ type BlueprintImportResponse = {
 	};
 };
 
-const uploadBlueprint = async ( file: File ) => {
-	const formData = new FormData();
-	formData.append( 'file', file );
-
-	if ( window?.wcSettings?.admin?.blueprint_upload_nonce ) {
-		formData.append(
-			'blueprint_upload_nonce',
-			window.wcSettings.admin.blueprint_upload_nonce
-		);
-	} else {
-		throw new Error( 'Blueprint upload nonce not found' ); // TODO: write a more user friendly error
-	}
-
-	const response = await apiFetch< BlueprintQueueResponse >( {
-		path: 'wc-admin/blueprint/queue',
-		method: 'POST',
-		body: formData,
-	} );
-
-	if ( response.error_type ) {
-		throw new Error( response.errors?.[ 0 ] ?? 'Unknown error' );
-	}
-
-	return response;
+type BlueprintImportStepResponse = {
+	success: boolean;
+	messages: {
+		step: string;
+		type: string;
+		message: string;
+	}[];
 };
 
-const importBlueprint = async ( process_nonce: string, reference: string ) => {
-	const { processed, message } = await apiFetch< BlueprintImportResponse >( {
-		path: '/wc-admin/blueprint/process',
-		method: 'POST',
-		data: {
-			reference,
-			process_nonce,
-		},
-	} );
+const importBlueprint = async ( file: File ) => {
+	const errors = [];
+	try {
+		// Create a FileReader instance
+		const reader = new FileReader();
 
-	if ( ! processed ) {
-		throw new Error( message );
+		// Create a promise to handle async file reading
+		const fileContent: string = await new Promise( ( resolve, reject ) => {
+			reader.onload = () => resolve( reader.result as string );
+			reader.onerror = () => reject( reader.error );
+			reader.readAsText( file );
+		} );
+
+		// Parse the file content as JSON
+		const steps = JSON.parse( fileContent ).steps;
+
+		// Ensure the parsed data is an array
+		if ( ! Array.isArray( steps ) ) {
+			throw new Error( 'Invalid JSON format: Expected an array.' );
+		}
+
+		// Loop through each step and send it to the endpoint
+		for ( const step of steps ) {
+			const logLabel = `${ step.step }: Importing `;
+			console.log( logLabel );
+			console.time( step.step );
+
+			const response = await apiFetch< BlueprintImportStepResponse >( {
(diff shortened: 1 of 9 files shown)


psealock (approved):
As far as I can tell this is all looking good. I get the same error you have in the screenshot and details in the console. I assume this is supposed to happen? Approving in either case.

moon0326:
> As far as I can tell this is all looking good. I get the same error you have in the screenshot and details in the console. I assume this is supposed to happen? Approving in either case.

Thank you!

Yup! The error is expected without the changes from https://github.com/woocommerce/woocommerce/pull/54301 PR

rjchow (changes requested):
Thanks for working on this! Left some suggestions on the XState bit

moon0326:
> Thanks for working on this! Left some suggestions on the XState bit

Thanks! For some reason, I don't see any other comments 🤔

rjchow:
> > Thanks for working on this! Left some suggestions on the XState bit
> 
> Thanks! For some reason, I don't see any other comments 🤔

Seems to have gone missing here too, that's really frustrating it was a long comment with details and examples

plugins/woocommerce/client/admin/client/blueprint/components/BlueprintUploadDropzone.tsx
```diff
@@ -54,49 +54,85 @@ type BlueprintImportResponse = {
...
 
-	if ( window?.wcSettings?.admin?.blueprint_upload_nonce ) {
-		formData.append(
-			'blueprint_upload_nonce',
-			window.wcSettings.admin.blueprint_upload_nonce
-		);
-	} else {
-		throw new Error( 'Blueprint upload nonce not found' ); // TODO: write a more user friendly error
-	}
+const importBlueprint = async ( file: File ) => {
```
rjchow, line 66: If we'd like to gain the benefits of using XState, we should avoid long sequences of code with flow control in it. Actors and Actions, should do only 1 thing and leave the flow control to the XState definition. 

It'll definitely be more verbose and more work than plain code, as the intent of XState is to fully codify the possible states and be explicit about how to get from one state to another. If we mix XState with long blocks of internal flow control, then we have the worst of both worlds - the verbosity and boilerplate around I/O between actors and the context, and we also don't get any clarity of statees, transitions, and explicit error handling.

For example, the below changes could be split out into:
- ReadBlueprintFile - perhaps a [callback actor](https://stately.ai/docs/callback-actors) that sends out events corresponding to the FileReader API
  - onload -> sendBack( { type: BLUEPRINT_READ_SUCCESS, output: reader.result } )
  - onerror -> sendBack( { type: BLUEPRINT_READ_ERROR, error: reader.error } )
  
- The steps flow control can be modelled after the one in [PluginInstallerMachine](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/admin/client/core-profiler/services/installAndActivatePlugins.ts#L103), perhaps looping around an internal state that that invokes a [fromPromise Actor](https://stately.ai/docs/actors#frompromise) that wraps the API call

moon0326: @rjchow Thanks! I think that's a bit out of scope for this PR though. With the new REST API approach, we need to refactor the XState code in this file (including what you've mentioned already). Could we handle that in a separate PR?

I've created a new issue here: https://github.com/woocommerce/woocommerce/issues/55155

rjchow: In view that the timeline for Blueprints has been extended, let's work on it in #55155. I'll add some details and ACs there to provide assistance

rjchow (approved):

PR #54552, opened 2025-01-16
Add client side notices support to the woocommerce/store-notices block

### LINKED ISSUE #54207: Migrate StoreNotices block to Interactivity API
One blocking issue right now for migrating to Interactivity API is the requirement to remove React rendering from the ProductButton block. The reason this block uses React is to add an ad-hoc store notice to the StoreNotices block.

StoreNotices is in need of a refresh to support both front-end and server rendered notices.

To do that we will:

1. Make the Store Notices block interactive (with the Interactivity API) this will include rendering error template's wrapped by interactivity directives
2. Provide a public `woo/notices` store with some like `actions.addNotice` that other blocks can use to add notices client side

We will need to test some common themes for visual regressions because `1` will require some small HTML structure change. 

As part of this we should also release a dev advisory to mention the HTML structure changes. And we should make sure to circulate information about the new usage available to add store notices via the Interactivity API (that info could potentially be in a dev advisory as well and perhaps some docs added about it.)

samueljseay:

The idea of this PR is to keep existing server rendered notices intact while also allowing for client-side notices to be added. It's worth noting that this isn't a generic solution to adding client-side notices right now. It solves a specific problem which is to replace a client-side generated notice that was being added within `woocommerce/product-button`. That notice was injected in the page via React and this new implementation adds the notice to `woocommerce/store-notices` via an additional HTML region for interactivity API. 

In doing this we remove `woocommerce/product-button`'s  dependency on React as part of our goal to migrate interactivity powered blocks to script modules / the latest version of interactivity API.

We use block hooks to ensure the StoreNotices is present, **but** that won't stop a user deleting the block. The block is also not rendered in the editor so it will be difficult for the user to break.

In future we need to consider what a fully interactivity API powered notices solution should look like and we will investigate this, but **imo** we should treat this PR as a solution that is in place to ensure the absolute minimum disruption to the existing functionality **of the single error notice added by product-button.**

Closes https://github.com/woocommerce/woocommerce/issues/54207

1. Add `woocommerce/product-collection` block to a post or page.
2. Make sure you have some products in stock in WooCommerce 
3. Choose a product and limit it to 1 per cart. You can do that in product settings:
4. Visit the post or page you created and add one of that limited product to your cart
5. Now try to add another

**Expected:**

* The error notice should scroll into view (screenshot below)
Screenshot 2025-01-20 at 2 21 23 PM

6. Now scroll down the page again to the product with limit 1 per cart
7. Click add to cart again

**Expected:**

* The error notice should scroll into view again

8. Now dismiss the error notice
9. Now try add to cart again

**Expected:**

* The error notice should scroll into view again

10. Now check the error message is still displayed full width on product catalog template. Go to your shop page.
11. Try add the limited product to cart again. 

**Expected:**

* The error message should scroll into view, it should also be full width as we are not hooking it when it's already present:

Screenshot 2025-01-20 at 2 21 05 PM

12. To simulate the notice block not being present, comment out the block hooks filter code in Controller.php of this PR
13. Now go to the page or post you made containing the product collection block. repeat the steps to add the limited 1 per cart item
14. The error should not display because the store notices was not hooked, but also there should be no fatal JS error. Open the dev console to confirm. You should still see a `console.error` for the notice, which is how it behaved before.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Use Interactivity API to render the error store notices triggered by woocommerce/product-button

</details>

<details>

#### Comment 

</details>


samueljseay:
@gigitux Hey I've dragged you in to review this from the side of store notices and see what you think. This won't be immediately merged to trunk as it's part of a feature branch, but I am still trying to keep this to the standard we'd expect when merging to trunk. I also do expect that we'd bring you and others in for a review of the final PR for the feature branch too.

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/button/frontend.tsx
```diff
@@ -87,6 +68,7 @@ const getContext = ( ns?: string ) => getContextFn< Context >( ns );
 
 const { state } = store< Store >( 'woocommerce/product-button', {
 	state: {
+		noticeId: '',
```
luisherranz: You don't need to add this if you are already adding it on the server. 

If you're doing it because of TypeScript, you can do what [I did in this PR](https://github.com/woocommerce/woocommerce/blob/13bfcfea3458735b6fda6dd2c8b4bd60be90c328/plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/button/frontend.tsx#L53) to update the types of the `store` with the ones in `@wordpress/interactivity` because the ones in `@woocommerce/interactivity` are outdated.

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
```diff
@@ -0,0 +1,133 @@
...
+			const noticeId = generateNoticeId();
+			state.notices = [
+				...state.notices,
+				{
+					...notice,
+					id: noticeId,
+				},
+			];
+
+			return noticeId;
```
luisherranz: You don't need to use immutability with the Interactivity API. Not only is it unnecessary, but it also performs worse.	

```suggestion
			notice.id = generateNoticeId();
			state.notices.push( notice );
            return notice.id;
```

plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/Controller.php
```diff
@@ -77,9 +77,48 @@ protected function initialize() {
...
+	 * @param string $position The position of the block.
+	 * @param string $anchor_block The anchor block.
+	 * @return array The array of hooked blocks.
+	 */
+	public function block_hook_fallback_store_notice( $hooked_blocks, $position, $anchor_block ) {
+		if ( 'core/post-content' === $anchor_block && 'before' === $position ) {
+			$hooked_blocks[] = 'woocommerce/store-notices';
+		}
+
+		return $hooked_blocks;
```
gigitux: This approach works well when the `core/post-content` block is implicit, as the `Store Notice` block remains hidden in the editor.  

However, when the block is explicit, the `Store Notice` block becomes visible in the editor.  

For example, after installing WooCommerce with this PR, the TT5 blog template displays the block as follows:

image


Not very expert with Block Hooks API - do you think that we should improve the logic to ensure that the block is hooked only when the “Product Collection” block is present in the page?

plugins/woocommerce/src/Blocks/BlockTypes/StoreNotices.php
```diff
@@ -48,23 +48,62 @@ protected function render( $attributes, $content, $block ) {
...
+			'woocommerce/store-notices',
+			array(
+				'notices' => array(),
+			)
+		);
+
+		ob_start();
+		// Adding the is-layout-constrained class to the wrapper is a hack, to ensure the notice looks the same as it used to when prepended to post content.
+		?>
+		<div data-wc-interactive="<?php echo esc_attr( wp_json_encode( $namespace ) ); ?>" class="woocommerce-notices-wrapper">
```
gigitux: In our codebase we need to escape iAPI data with: 
```php
wp_json_encode( $interactive, JSON_NUMERIC_CHECK | JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_AMP ) 
```

For instance:

https://github.com/woocommerce/woocommerce/blob/4b987cfe4e5bc81c0a61e7126bcc4df5391745f5/plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php#L177-L180

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/button/frontend.tsx
```diff
@@ -146,23 +128,27 @@ const { state } = store< Store >( 'woocommerce/product-button', {
...
+						noticesStore.actions.removeNotice( state.noticeId );
+					}
+
+					const noticeId = noticesStore.actions.addNotice( {
+						notice: decodeEntities( message ),
+						type: 'error',
+						dismissible: true,
+					} );
+
+					state.noticeId = noticeId;
```
gigitux: I'm not sure that this check works well. I commented these two lines:

https://github.com/woocommerce/woocommerce/blob/4b987cfe4e5bc81c0a61e7126bcc4df5391745f5/plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/Controller.php#L81-L82


After adding the product, this is the error that I saw in the console:

image

gigitux (changes requested):
Thanks for working on this! 🙇 

I left a few comments, but I think that we need to revisit the Block Hook API approach, given that creates a weird scenario on not Woo-related templates.

plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/Controller.php
samueljseay: Ah right. That was my intent to only do it when product collection is on the page but naive to run this code here. Yes i will try and see if thats possible. Thanks for spotting that.

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/button/frontend.tsx
samueljseay: Whoops 🤦‍♂️ i had already made this check more defensive but looks like i didnt push the change. Thanks for spotting that

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/button/frontend.tsx
```diff
@@ -146,23 +128,27 @@ const { state } = store< Store >( 'woocommerce/product-button', {
...
+				// and we should not add a notice.
+				if ( noticesStore ) {
+					// The old implementation always overwrites the last notice, so
+					// we remove the last notice before adding a new one.
+					if ( state.noticeId !== '' ) {
+						noticesStore.actions.removeNotice( state.noticeId );
+					}
+
+					const noticeId = noticesStore.actions.addNotice( {
+						notice: decodeEntities( message ),
```
luisherranz: Is `decodeEntities` really necessary considering that we will end up injecting the message into the HTML using `ref.innerHTML`? It seems like a React leftover.

Take into account that we also need to get rid of `@wordpress/html-entities`.

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
```diff
@@ -0,0 +1,133 @@
+/**
+ * External dependencies
+ */
+import { getContext, getElement, store } from '@woocommerce/interactivity';
+
+type Notice = {
+	notice: string;
```
luisherranz, line 17: This is referred as `$message` in the server API. Perhaps we should use `message` instead of `notice` here as well?

https://woocommerce.github.io/code-reference/files/woocommerce-includes-wc-notice-functions.html#function_wc_add_notice

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
```diff
@@ -0,0 +1,133 @@
...
+				},
+			];
+
+			return noticeId;
+		},
+
+		removeNotice: ( noticeId: string ) => {
+			state.notices = state.notices.filter(
+				( notice ) => notice.id !== noticeId
+			);
```
luisherranz: Same here about immutability.


```suggestion
			const index = state.notices.findIndex(
				( { id } ) => id === noticeId
			);
			if ( index !== -1 ) state.notices.splice( index, 1 );
```

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
```diff
@@ -0,0 +1,133 @@
...
+	callbacks: {
+		dismissNotice: () => {
+			const context = getContext< { notice: NoticeWithId } >();
+
+			state.notices = state.notices.filter(
+				( notice ) => notice.id !== context.notice.id
+			);
+		},
+
+		getNoticeClass: () => {
```
luisherranz: This should be derived state instead of a callback.

The same applies for the other getters and the `isNoticeDismissible` one.

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
```diff
@@ -0,0 +1,133 @@
...
+				{
+					...notice,
+					id: noticeId,
+				},
+			];
+
+			return noticeId;
+		},
+
+		removeNotice: ( noticeId: string ) => {
```
luisherranz: Should we rename this to `clearNotice` to match the server API?

https://woocommerce.github.io/code-reference/files/woocommerce-includes-wc-notice-functions.html#function_wc_clear_notices

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
```diff
@@ -0,0 +1,133 @@
...
+};
+
+const generateNoticeId = () => {
+	// semi-random with low collision probability.
+	return `${ Date.now() }-${ Math.random()
+		.toString( 36 )
+		.substring( 2, 15 ) }`;
+};
+
+const { state } = store< StoreNoticesStore >( 'woocommerce/store-notices', {
```
luisherranz: It'd be good to think about what parts of this store must be public (`actions.addNotice` and `actions.removeNotice`?) and move everything else that is internal to a private store.

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
```diff
@@ -0,0 +1,133 @@
...
+	state: {
+		notices: [],
```
luisherranz: You don't need to add this if you are already creating the array on the server.

Again, if it's because of the types, we can import them from `@wordpress/interactivity`.

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/button/frontend.tsx
```diff
@@ -146,23 +128,27 @@ const { state } = store< Store >( 'woocommerce/product-button', {
...
-				if ( domNode ) {
-					injectNotice( domNode, decodeEntities( message ) );
+				// If the user deleted the hooked store notice block, the store won't be present
+				// and we should not add a notice.
+				if ( noticesStore ) {
+					// The old implementation always overwrites the last notice, so
+					// we remove the last notice before adding a new one.
+					if ( state.noticeId !== '' ) {
+						noticesStore.actions.removeNotice( state.noticeId );
+					}
```
luisherranz: So, the assumption here is that we never want the errors produced by this block to be concatenated, right? They should always override the previous one.

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
```diff
@@ -0,0 +1,133 @@
...
+			);
+		},
+	},
+	callbacks: {
+		dismissNotice: () => {
+			const context = getContext< { notice: NoticeWithId } >();
+
+			state.notices = state.notices.filter(
+				( notice ) => notice.id !== context.notice.id
+			);
```
luisherranz: Same here about immutability. 

By the way, why not reuse the `removeNotice` action here?


```suggestion
      actions.removeNotice( context.notice.id );
```

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
```diff
@@ -0,0 +1,133 @@
...
+		getNoticeClass: () => {
+			const context = getContext< { notice: NoticeWithId } >();
+
+			const noticeTypeClass = {
+				error: 'is-error',
+				success: 'is-success',
+				notice: 'is-info',
+			}[ context.notice.type ];
+
+			return `wc-block-components-notice-banner ${ noticeTypeClass }`;
```
luisherranz: Binding the whole class makes it less extensible as it blocks any other class modification. Why not use three different `data-wp-class` directives instead? Just to avoid some repetition?

It'd be more solid to do it this way:

```html
<div
  data-wp-class--is-error="state.isNoticeError"
  data-wp-class--is-success ="state.isNoticeSuccess"
  data-wp-class--is-info ="state.isNoticeInfo"
  ...
</div>
```

```js
   state: {
     get isNoticeError() {
       const { notice } = getContext();
       return notice.type === 'error';
     },
     get isNoticeSuccess() {
       const { notice } = getContext();
       return notice.type === 'success';
     },
     get isNoticeInfo() {
       const { notice } = getContext();
       return notice.type === 'notice';
     },
   }
```

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
```diff
@@ -0,0 +1,133 @@
...
+				return 'alert';
+			}
+
+			return 'status';
+		},
+
+		isNoticeDismissible: () => {
+			const context = getContext< { notice: NoticeWithId } >();
+			return context.notice.dismissible;
+		},
```
luisherranz: Oh, you don't need this. You can use `"context.notice.dismissible"` in the HTML.

plugins/woocommerce/src/Blocks/BlockTypes/StoreNotices.php
```diff
@@ -48,23 +48,62 @@ protected function render( $attributes, $content, $block ) {
...
+			>
+				<div data-wc-init="callbacks.scrollIntoView" data-wc-bind--class="callbacks.getNoticeClass" data-wc-bind--role="callbacks.getNoticeRole" >
+					<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="24" height="24" aria-hidden="true" focusable="false">
+						<path data-wc-bind--d="callbacks.getNoticeIconPath"></path>
+					</svg>
+					<div class="wc-block-components-notice-banner__content">
+						<span data-wc-init="callbacks.renderNoticeContent"></span>
+					</div>
+					<button
+						data-wc-bind--hidden="!callbacks.isNoticeDismissible"
```
luisherranz: You can use the value from context directly.

```suggestion
						data-wc-bind--hidden="!context.notice.dismissible"
```

plugins/woocommerce/src/Blocks/BlockTypes/StoreNotices.php
```diff
@@ -48,23 +48,62 @@ protected function render( $attributes, $content, $block ) {
...
+						class="wc-block-components-button wp-element-button wc-block-components-notice-banner__dismiss contained"
+						aria-label="<?php esc_attr_e( 'Dismiss this notice', 'woocommerce' ); ?>"
+						data-wc-on--click="callbacks.dismissNotice"
```
luisherranz: It's not mandatory, but to keep consistency with the way we usually model actions vs callbacks, `dismissNotice` should be an action instead of a callback.

luisherranz (changes requested):
I've added several small improvement suggestions, but overall it's looking good 🙂

I don't know anything about block hooks, so I'll leave that up to you.

I also saw the error when the Store Notice block is not present. We should think what happens when that's the case, like showing a proper error message in the console so that the person developing the site/theme knows that the Store Notice block is missing.

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/button/frontend.tsx
samueljseay: @luisherranz yeah I did it to make TS happy, sure I'll use your approach here.

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/button/frontend.tsx
samueljseay: @luisherranz That's right, and I only do that to exactly match how it behaves right now. That's a concern of this block if they want to change the behaviour later they can with the APIs available.

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
samueljseay: Ok thanks, I didn't know that!  🗒️

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
samueljseay: Good point 😅

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
samueljseay: @luisherranz Ah you mean just move these into state with `get`? sure yep. I always find it very odd to be accessing context within derived state, the 2 concepts overlap in a way that is confusing from the outside (like why don't we just have derived context too? I think we have discussed this before 😅 )

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
samueljseay: Good point! Thanks! The only issue is that derived state does not work for me inside the template.

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/button/frontend.tsx
samueljseay: Resolved in https://github.com/woocommerce/woocommerce/pull/54552/commits/4b987cfe4e5bc81c0a61e7126bcc4df5391745f5

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
samueljseay: The arg is `message` but the object on the template had property `notice`, the code in that function:

```
if ( ! empty( $message ) ) {
		$notices[ $notice_type ][] = array(
			'notice' => $message,
			'data'   => $data,
		);
	}
```

😅 that said, I'm not too fussed I didn't really like `notice.notice` so maybe it's nicer to go with message here, which as you say matches the public method signature better.

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
samueljseay: Yeah solid point. So we'd call the private store from within the actions of the public one. Would you just name it like `woocommerce/store-notices/private` or something?

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
samueljseay: I've resolved this, but still using spread to assign id to notice variable mainly for type reasons, I don't want `Notice` to have `id?` because I don't want users to think they have an option to pass that.

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
samueljseay: Good point!

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
samueljseay: @luisherranz I had already tried derived state, that doesn't seem to work inside the each template when I'm referencing `data-wc-each--notice="state.notices"` I assumed this was either: a bug in our version of iAPI, or this case of referencing derived state inside a template had not been considered.

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
samueljseay: Fixed this and also introduced a basic style to ensure it's actually hidden which I missed.

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
luisherranz: That's fine 🙂

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
luisherranz: Up to you 🙂

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
luisherranz: Yeah, something like that. 

I would also like to consider whether it makes sense to have a public store for store notices or if those actions should be part of the general public WooCommerce store. In principle, it seems to me that it makes sense for it to have its own public store, but let's keep an eye on it in case we see that it makes more sense to integrate it.

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/button/frontend.tsx
luisherranz: Ok, thanks.

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
luisherranz: It's working fine for me. Could you describe what is failing for you?

I just pushed a change: https://github.com/woocommerce/woocommerce/pull/54552/commits/c461b01ab0f90f2df68a0c5f24f0216f8080731c

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
luisherranz: I just pushed this change, and it's working fine for me.

Let's discuss here: 
- https://github.com/woocommerce/woocommerce/pull/54552#discussion_r1924977004

plugins/woocommerce/src/Blocks/BlockTypes/StoreNotices.php
luisherranz: I did this in https://github.com/woocommerce/woocommerce/pull/54552/commits/c461b01ab0f90f2df68a0c5f24f0216f8080731c.

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
luisherranz: By the way, we're planning to introduce the `data-wp-class` directive without a suffix (default) so that it can return an object where the keys are the class names and the values are booleans indicating whether each class should be included.

In this case, the code would be simplified to something like:

```html
data-wp-class="state.noticeClass"
```

```js
   state: {
     get noticeClass() {
       const { notice } = getContext();
       return { [ `is-${notice.type}` ]: true };
   }
```

More info [here](https://github.com/WordPress/gutenberg/pull/65803#issuecomment-2468475673) and any feedback is welcomed 🙂

plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/Controller.php
```diff
@@ -112,13 +136,40 @@ public function block_hook_fallback_store_notice( $hooked_blocks, $position, $an
...
+		}
+
+		
 		// Do not default to wide, as it will break the layout for this fallback block.
 		$parsed_hooked_block['attrs']['align'] = '';
 
 		return $parsed_hooked_block;
 	}
 
+	public function block_hooks_support_post_content() {
```
tjcafferkey: We can probably simplify this code, looking [here](https://github.com/WordPress/gutenberg/pull/67272/files#diff-76236aad75203b8bb9c5c1bb8c6c6bdd9751d9226f624c4e89327af5c11cb0b4R155) and [here](https://github.com/WordPress/wordpress-develop/pull/7889/files#diff-4112a9a0bb18713bc2fc7868ceffe4403d765722f605e42b313ef2cdc218177fR195) we could in theory do something like:

```php
if ( has_filter('the_content', 'gutenberg_apply_block_hooks_to_post_content') || has_filter('the_content', 'apply_block_hooks_to_content') ) {
	return true;
}

return false;
```

tjcafferkey: I think it might be better being moved into [wc-conditional-functions.php](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/wc-conditional-functions.php) too

plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/Controller.php
```diff
@@ -78,23 +78,47 @@ protected function initialize() {
...
+	 * @param string $position The position of the block.
+	 * @param string $anchor_block The anchor block.
+	 * @return array The array of hooked blocks.
+	 */
+	public function block_hook_fallback_no_post_content_support( $hooked_blocks, $position, $anchor_block ) {
+		// In WordPress versions prior to 6.8, contents of post_content was not available on the context,
+		// so it's hard to detect if the product collection block is being rendered. As a workaround,
+		// we don't hook the block in admin in this case, so that the only negative side effect
+		// on the frontend is that an empty store notices block will be rendered.
+		if ( ! is_admin() && 'core/post-content' === $anchor_block && 'before' === $position ) {
```
tjcafferkey: Probably better using `! block_hooks_support_post_content &&  ! is_admin()` to determine whether the filter is even needed to be added rather than adding it unconditionally and firing this check in the callback

plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/Controller.php
```diff
@@ -78,23 +78,47 @@ protected function initialize() {
 		add_filter( 'render_block_data', array( $this, 'disable_enhanced_pagination' ), 10, 1 );
 
 		// Hook the store notices block to ensure woocommerce/product-button can add error notices client-side.
-		add_filter( 'hooked_block_types', array( $this, 'block_hook_fallback_store_notice' ), 1, 4 );
+		$filter = $this->block_hooks_support_post_content() ? 'block_hook_fallback_post_content_support' : 'block_hook_fallback_no_post_content_support';
+		add_filter( 'hooked_block_types', array( $this, $filter ), 1, 4 );
```
tjcafferkey: You could avoid adding a filter altogether unnecessarily with this approach.

```php
if ( $this->block_hooks_support_post_content() ) {
	add_filter( 'hooked_block_types', 'block_hook_fallback_post_content_support', 1, 4 );
} else if ( ! is_admin() ) {
	add_filter( 'hooked_block_types', 'block_hook_fallback_no_post_content_support', 1, 4 );
}
```

plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/Controller.php
tjcafferkey: as per [this comment](https://github.com/woocommerce/woocommerce/pull/54552/commits/096f6e1c691d8e5aa24458fdefd19872f1bf6fed#r1925325914)

plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/Controller.php
samueljseay: Oh nice, great suggestions thank you

plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/Controller.php
samueljseay: The approach you suggested is working well thanks for the suggested improvements. Resolved in [e5f1c07](https://github.com/woocommerce/woocommerce/pull/54552/commits/e5f1c0759616a27575532665ca438c7928e28e70)

plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/Controller.php
samueljseay: I think this is resolved now, but will be keen for some re-testing

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/button/frontend.tsx
samueljseay: Resolved in https://github.com/woocommerce/woocommerce/pull/54552/commits/4d9ea04107cb5357a35df61d61782f5da980cebc

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/button/frontend.tsx
samueljseay: Oh yes you're absolutely right, I should have spotted that. Actually the console error should probably decode entities to make it more readable but I don't think I'll do that here, as you say we also need to get rid of that dependency.

Resolved in https://github.com/woocommerce/woocommerce/pull/54552/commits/a867a76c3fcb0d987b386923f6833fe9b7ee01b1

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
samueljseay: I thought about this, I think it's fine to leave as is. I don't feel strongly either way.

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
samueljseay: Thanks, I'm not sure why this wasn't working for me 😅

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
samueljseay: Just my opinion, but I don't think we need to match the server API here. In a lot of other places we have some kind of client side notices we use the terminology `removeNotice`. At least to me that also feels a bit more intutitive than clearNotice. I don't feel super strongly about it though. I will leave this as is, but if there is some strong feelings about it I can revisit.

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
samueljseay: @luisherranz yeah makes sense. How do you feel about a follow up for this as we get closer to merging the feature branch to trunk? 

I feel that it is simpler to reason about from 3PD perspective if we keep stores granular per "business area" (ie cart, notices, etc), but it could be good to discuss this and hear your thoughts about that.

samueljseay:
@gigitux I've added some additional testing steps in the description to account for the situation you ran into with store notices being hooked when it shouldn't be. I have done some initial testing and it seems to be working as I'd expect now. 

@tjcafferkey Thanks for the review of the hooks code, would you mind also being a reviewer for this implementation. Would love one more set of eyes on it.

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
luisherranz: Cool, no problem.

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
luisherranz: I don't have strong feelings, it was simply a question. So I agree with leaving it like that 🙂

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
luisherranz: Sure. 

I don't have strong opinions about how the store should be structured; we just need to find something that makes sense.

gigitux:
Thanks, @samueljseay! I tested with Gutenberg trunk installed.

### Post
<video src="https://github.com/user-attachments/assets/83bc527a-6e5f-44bb-ba10-0cb0b8b44f2c"> </video>
The notice isn't visible on the post, likely because the `content` block is missing.

### Page
<video src="https://github.com/user-attachments/assets/0999cb5f-08e1-4fe9-80d4-bf9e781124a2"> </video>
The notice isn't visible on the page either, though the `content` block is present.

### Archive Template

#### With Store Notice block
<video src="https://github.com/user-attachments/assets/ef8746b0-0d62-49a0-bd7c-60b1cb277f8d"> </video>
The notice is visible on the frontend when the block is present in the template.

#### Without Store Notice
<video src="https://github.com/user-attachments/assets/c4c8ef04-1c24-4569-a032-428f84db8415"> </video>
The notice isn't visible on the frontend when the block is absent.

The behavior seems correct for the Archive Template, but I’m unsure about the Post and Page scenarios. Is there something wrong with my env? 🤔

samueljseay:
@gigitux Thanks for reviewing this, I'm sorry it's not working as expected. 

> The notice isn't visible on the post, likely because the content block is missing.

Content block is on posts, but on post it's implicit (not visible in editor), you'll see it rendered in the frontend and the hook fires for it. I also see that the hook fires for it and for `first_child` but for some reason `first_child` seems not to be working on posts and I missed that. If I change it to `before` it works on posts. I'll try tweak this some more.

> The notice isn't visible on the page either, though the content block is present.

I cannot reproduce this but I'm using Gutenberg 20.0 not trunk. I wonder if there is a bug on trunk, I'm definitely seeing block hooks be inconsistent just by the behaviours we've seen here. Could be something I'm not understanding about the API. I will test with trunk as well.

samueljseay:
@gigitux 🤞🏻 that it's working as expected now. I'm sorry it didn't at all do that before. I've removed Gutenberg testing steps because we no longer use conditional code based on available filters.

Would you mind testing again?

I've tested post and page with inserted product collection block. But I'm still not quite clear what you were testing with the archive template, It looks like from your screen recording that the block was being auto inserted on editor side there or something and you deleted it at the beginning?

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
```diff
@@ -0,0 +1,142 @@
...
+
+				if ( index !== -1 ) {
+					state.notices.splice( index, 1 );
+				}
+			},
+
+			dismissNotice: () => {
+				const context = getContext();
+				actions.removeNotice( context.notice.id );
+			},
```
luisherranz: Perhaps we could merge these two:

```suggestion
			removeNotice: ( noticeId?: string ) => {
        noticeId = noticeId || getContext().notice.id;
				const index = state.notices.findIndex(
					( { id } ) => id === noticeId
				);

				if ( index !== -1 ) {
					state.notices.splice( index, 1 );
				}
			},
```

luisherranz (approved):
Looks good to me 🙂 

I leave the block hooks implementation up to you, and we should note somewhere that the store should be divided between private and public so we don't forget to do it later.

gigitux:
@samueljseay, thanks for working on this! Below my findings: 

## Without Gutenberg

### Post - TT5 👍

image

### Page - TT5 👍

image

### Product Catalog - TT5 👍 

image


### Other Templates - TT5 👍 

The store notice block isn't injected in not Woo related templates.

## With Gutenberg

### Post - TT5 👍

image

### Page - TT5 ❌

image

The Store Notice is injected on a new page

### Product Catalog - TT5 👍

image

### Other Templates - TT5 👍 

The store notice block isn't injected in not Woo related templates.



I don't think that this is an issue of this PR and I have no idea how we could fix this, but, on TT4 theme, the store notice width is too wide:

image

plugins/woocommerce-blocks/assets/js/blocks/store-notices/frontend.ts
samueljseay: Yep good point. Will do this.

samueljseay:
Thanks @gigitux for the thorough testing, it really helped me out here. As we discussed we'll go ahead and merge this to the feature branch. I've created some issues:

https://github.com/woocommerce/woocommerce/issues/54907

and 

https://github.com/woocommerce/woocommerce/issues/54912

which we will address before merging this feature to trunk.

Issue #54512, opened 2025-01-15
Product Gallery: Remove thumbnail settings from parent block


tjcafferkey:
Remove the thumbnail settings from the parent block and remove the functionality that syncs them to the Thumbnails block

EDIT: Remove the Thumbnails limit completely

kmanijak:
Closed via https://github.com/woocommerce/woocommerce/pull/54799

PR #54570, opened 2025-01-16
Product Gallery: Simplify Pager to just display "current image/total images"

### LINKED ISSUE #54510: Product Gallery: Remove Pager settings from parent block
Remove the pager settings from the parent block and remove the functionality that syncs them to the pager block

### LINKED ISSUE #54516: Product Gallery: Replace Pager with current/total label
Replace the Pager with current/total label

kmanijak:

This PR removes the Pager variants (dots/numbers) and displays "current image/total images" instead. It also adds multiple global styles to the block.

Closes https://github.com/woocommerce/woocommerce/issues/54510
Closes https://github.com/woocommerce/woocommerce/issues/54516

3. Make sure the block displays "3/7" in editor
4. Play around with global styles
5. Save and go to frontend
6. **Expected:** Make sure it displays 1/N (depending on how many images your products has)
7. Change the image through arrows and clicking on thumbnail
8. **Expected:** Make sure Pager gets updated

> [!NOTE]  
> You may encounter the case in which Pager says for example "3/5" suggesting there's more images but right arrow gets disabled. This is temporarily expected. Current logic limits the number of images shown by default to number of thumbnails while there's more images to show in the full page image. It'll be fixed in https://github.com/woocommerce/woocommerce/issues/54512.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
```diff
@@ -45,51 +45,41 @@ const getImageIndex = ( context: ProductGalleryContext, imageId: string ) => {
...
-	context.disableLeft = nextImageIndex === 0;
-	context.disableRight = nextImageIndex === imagesIds.length - 1;
+	context.disableLeft = nextImageIndex === 1;
+	context.disableRight = nextImageIndex === imagesIds.length;
 };
 
 const selectImage = (
 	context: ProductGalleryContext,
-	type: 'prev' | 'next' | 'current'
+	type: 'prev' | 'next' | 'current' | 'first'
```
tjcafferkey: Im curious why we have `first` and not `last`? Also if we're able to use `Math.min` to calculate `last` why can't we do similar to programmatically calculate `first`?

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
```diff
@@ -45,51 +45,41 @@ const getImageIndex = ( context: ProductGalleryContext, imageId: string ) => {
...
+	let newImageIndex = 1;
 
-	// explicit "current"
-	let nextPotentialIndex = selectedImageIdIndex;
+	if ( type === 'first' ) {
+		newImageIndex = 1;
+	}
+
+	if ( type === 'current' ) {
+		newImageIndex = getImageIndex( context, imageId ) + 1;
```
tjcafferkey: It looks like we use both zero-based (e.g. in `disableArrows`) and one-based (here) indexing. Caught me out when reading the code. Ideally we'd be consistent.

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/inner-blocks/product-gallery-pager/edit.tsx
```diff
@@ -1,109 +1,12 @@
...
-
-interface EditProps {
-	attributes: BlockAttributes;
-	setAttributes: ( newAttributes: BlockAttributes ) => void;
-	context: ProductGalleryPagerContext;
-}
-
-export const Edit = ( props: EditProps ): JSX.Element => {
-	const { context } = props;
+export const Edit = (): JSX.Element => {
```
tjcafferkey: This type isn't necessary

plugins/woocommerce/src/Blocks/BlockTypes/ProductGallery.php
```diff
@@ -186,8 +186,8 @@ protected function render( $attributes, $content, $block ) {
 				'data-wc-context',
 				wp_json_encode(
 					array(
-						'selectedImage'                   => $product_gallery_first_image_id,
 						'firstMainImageId'                => $product_gallery_first_image_id,
+						'selectedImageIndex'              => 1,
```
tjcafferkey: I guess this is why we mix zero/one based indexing. Any danger is defaulting to `0`?

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryPager.php
```diff
@@ -49,123 +49,29 @@ protected function get_block_type_uses_context() {
...
-		if ( 'off' === $pager_display_mode ) {
-			return null;
-		}
-
-		$number_of_thumbnails = $block->context['thumbnailsNumberOfThumbnails'] ?? 0;
-		$wrapper_attributes   = get_block_wrapper_attributes();
-		$post_id              = $block->context['postId'] ?? '';
-		$product              = wc_get_product( $post_id );
+		$post_id = $block->context['postId'] ?? '';
+		$product = wc_get_product( $post_id );
```
tjcafferkey: if `$post_id` is `''` is there any point in executing `wc_get_product`? We could just immediately return `''` I guess?

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryPager.php
```diff
@@ -49,123 +49,29 @@ protected function get_block_type_uses_context() {
 	 * @return string Rendered block type output.
```
tjcafferkey, line 49: We dont return a `string` if `$product` is falsey. Perhaps we should return an empty string

tjcafferkey:
I think we've left some references we can now remove here for `pagerDisplayMode`:

* https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce-blocks/assets/js/blocks/product-gallery/edit.tsx#L149
* https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce-blocks/assets/js/blocks/product-gallery/edit.tsx#L149

There could be more that I just havent found

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
kmanijak: > Im curious why we have first and not last?

There's no use case for `last`. We move to `first` when closing the dialog. This is because the image set is bigger in dialog than regular gallery which we'll change when revisiting thumbnails. But for now I suggest falling back to first image. Once we display all images in regular gallery we can easily stay on the same image without additional logic.

> Also if we're able to use Math.min to calculate last why can't we do similar to programmatically calculate first?

I'm using Math.min and Math.max for next/previous buttons so we don't move outside of image ranges. I'm not sure how that applies to `first` when we now explicitly what first is? Or you mean something else?

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
kmanijak: Yeah, so the reason for using one-based is I'm using `selectedImageIndex` as the actual value that gets rendered on the frontend via `data-wc-text` directive [here](https://github.com/woocommerce/woocommerce/pull/54570/files#diff-646bdfbca017740548f7a8a7218bd10914aaac440748cc747d9052e3d803b7a9R69):

```
<span class="wc-block-product-gallery-pager__current-index" data-wc-text="context.selectedImageIndex"></span>
```

Perhaps changing the variable's name to `selectedImageNumber` and corresponding variables (e.g. `newImageIndex` -> `newImageNumber`) would be good move? Plus comments.

Unfortunately, `data-wc-text` directive doesn't allow modifying the output which would be ideal.

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
kmanijak: I removed the `first` but introduced special `closeDialog` case. It's temporary until we have different number of images in "both" galleries. I wanted it to be more explicit. Also, renamed the variable to show better it's one-based.

https://github.com/woocommerce/woocommerce/pull/54570/commits/54a2ae6c20ddc32a687cf7a1124679ec02bfc019

kmanijak:
Thanks for review @tjcafferkey! I addressed all of your comments and answered to some of them! 🙌

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
tjcafferkey: Thanks for the explanation! I wouldn't consider this a blocker but just something I noticed that could potentially improved at the time

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
```diff
@@ -264,7 +269,7 @@ const productGallery = {
 			);
 
 			const selectFirstImage = () => {
-				context.selectedImage = context.firstMainImageId;
+				selectImage( context, 'first' );
```
tjcafferkey, line 272: `first` needs updating here I think

tjcafferkey (changes requested):
Just one comment, then we're good to merge!

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
kmanijak: Added a `first` as accepted option, forgot it's still in use in here. Also, as we discussed elsewhere I'll double check why changing images with variation change doesn't work.

https://github.com/woocommerce/woocommerce/pull/54570/commits/41f56ee7ca9e57c5f11fd04aea5b4f477f6f7e8c

kmanijak: Ok, I verified and variable product images change as on `trunk` which means:
- change the image when it's in available images (it won't be changed if it's only available on dialog but we'll fix it)
- change the image to first upon clearing.

So I guess we're good to go with this PR.

https://github.com/user-attachments/assets/56e5033b-5644-4354-9f6a-2a6b7884f229

tjcafferkey (approved):

PR #54575, opened 2025-01-16
[e2e tests] Move products test files into new folder structure and merge tests

adimoldovan:

Part of pdV5qK-Ie-p2

Move all product related tests into the new `product` folder. This is a first move to try out a new project structure focused on features and moving away from the merchant/shopper split. More context: pcr5Cj-70A-p2

Also merged 2 tests from the old `shopper/product-simple.spec.js`:
1. The test for the product review was moved in the `product-reviews.spec.js` and was refactored:
- use the customer account to add a review, instead of the admin account
- updated locators
- added more checks
2. The test for the product description was removed, because this is already checked in the `product-create-simple.spec.js`. I updated one of the tests in this one to use HTML in the product description though.

Run the tests in the `tests/product` folder. They should all pass.


lanej0 (approved):
Ran the tests, went to lunch came back and everything had passed.  So all good!

PR #54576, opened 2025-01-16
Add to Cart (Product button) block: Prepare for migration to `@wordpress/interactivity`

### LINKED ISSUE #54502: Stop reading from the wp.data store and move logic to the iAPI store.
The goal here is to remove all `@wordpress/data` usage within iAPI block frontends and replace those with an iAPI store (or stores).

luisherranz:

Fixes #54502.

The main change is to remove the dependency of the block on `@wordpress/data` to prepare it for the migration to the WordPress Core Interactivity API. 

Besides that, this pull request also implements:

1. It moves the logic of adding items to the cart to a global WooCommerce store so that it can be shared by other blocks like the block.
2. It implements optimistic updates so that the user doesn't have to wait for the database to update to see on the screen that the item has been added to the cart (this is something we should test more thoroughly). I've kept the animation since I think it's a good way to indicate to the user that something has happened and it will also be useful when we start to fetch the cart on page load.

**Note**: The synchronization with the Minicart block has been lost, but that is something that @samueljseay and I want to address in a different pull request.

It also adds annotations of things we have to do when the migration progresses and we start importing from `@wordpress/interactivity` directly.

**Note**: We probably shouldn't merge this into trunk as it contains a lot of todo items that can't be finished until we progress with the migration and start importing `@wordpress/interactivity` directly. So maybe we should merge it into a branch where we work on the migration and once all the changes are ready, they can be merged all at once. It's also dependent on @samueljseay work on the Store Notices block migration to the Interactivity API, as I simply removed the old React-based notices functionality.

I've made a video to explain in more detail the changes that have been made so far and what is missing.

<div>
    <a href="https://www.loom.com/share/b7ab41f169754751b91c7b378f789b2b">
      
    </a>
  </div>

https://www.loom.com/share/b7ab41f169754751b91c7b378f789b2b

### Remaining tasks 

_(I'll be adding more as I progress)_

- [x] Move the `wp_initial_state` call of the `woocommerce` store to its own file so that it can be included in multiple blocks and it's called only once.
- [x] Fetch the cart on page load.

1. Add a Product Collection block that contains the Product Button block.
2. Go to the frontend and add products to the cart
3. Verify that cart updates happen instantly (optimistic updates)
4. Check that the cart state persists correctly
5. Disable network connection or make the backend fail and attempt to add items to cart
6. Verify that optimistic updates are properly rolled back

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>
To be done...
<details>

#### Comment 

</details>


luisherranz:
@samueljseay @Aljullu, even though this is not finished, I would love to have your initial feedback on these changes.

Aljullu (commented):
I agree that moving Cart status to a shared store will be beneficial, and it will allow different blocks to be in sync.

> so that it can be shared by other blocks like the block

I guess this sentence is incomplete? :smile: 

---

It might not be in the scope of this PR, but when an update fails, the optimistic update doesn't seem to be reset correctly on my end. (In `trunk` there is a visual issue that makes the text not to be correctly aligned :sweat_smile: ):

`trunk` | This branch
--- | ---
trunk | branch

To reproduce:

1. Load a page with the Product Collection block.
2. Add a product to your cart.
3. In a separate tab, edit that product and make it so it's sold individually.
4. In the initial tab, try adding the product again to your cart.

Can you reproduce as well?

---

Besides that, code changes look good to me and it's testing well on my end. I will leave the final review to @samueljseay, though, as he has a better understanding of the project.

plugins/woocommerce-blocks/assets/js/base/stores/add-to-cart.ts
```diff
@@ -0,0 +1,70 @@
+/**
```
samueljseay, line 1: Nitpick, but I wonder if we should namespace the stores in the folder structure like `stores/cart/add-to-cart.ts`. It might be easier to organize and work with as we add more stores.

samueljseay:
@luisherranz since this is not finished I won't give review of nit pick items or issues with TS types or lint. 

On a high level these changes look great to me. Obviously it would be good to investigate whether the issue raised by @Aljullu is related to this, but otherwise looking good.

luisherranz:
Thanks, folks!

I have made these three changes:

- Move the wp_initial_state call of the woocommerce store to its own file so that it can be included in multiple blocks and it's called only once.
- Fetch the cart on page load.
- Fix optimistic updates on store API failure.

I made a video to explain these changes and also explain how to simulate a cached page.

<div>
    <a href="https://www.loom.com/share/56a52c1ecc8646878dbd5a113defea75">
      
    </a>
  </div>

https://www.loom.com/share/56a52c1ecc8646878dbd5a113defea75

For me, the functionality I wanted to include in this PR is already completed.

I am opening comments on all the questions I left in the code, so we can discuss them before merging.

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/button/frontend.tsx
```diff
@@ -38,205 +39,136 @@ interface Store {
...
+				const { actions } = ( yield import(
+					'../../../../base/stores/add-to-cart'
+				) ) as WooStore;
 
-			try {
-				yield dispatch( storeKey ).addItemToCart(
+				actions.addToCart(
 					productId,
-					quantityToAdd
+					// Question: is quantityToAdd available in the cart or we need to pass it down?
```
luisherranz: Question: is `quantityToAdd` available in the cart, or do we need to pass it down?

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/button/frontend.tsx
```diff
@@ -38,205 +39,136 @@ interface Store {
...
-				const storeNoticeBlock =
-					document.querySelector( storeNoticeClass );
-
-				if ( ! storeNoticeBlock ) {
-					document
-						.querySelector( '.entry-content' )
-						?.prepend( createNoticeContainer() );
+			},
+			*refreshCart() {
+				// Question: use `splitTask` here? Does it make sense?
```
luisherranz: Question: use `splitTask` here? Does it make sense?

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/button/frontend.tsx
```diff
@@ -38,205 +39,136 @@ interface Store {
...
+				// Question: does it make sense to use a dynamic import instead of a static one?
```
luisherranz: Question: does it make sense to use a dynamic import instead of a static one?

plugins/woocommerce-blocks/assets/js/base/stores/add-to-cart.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+let pendingRefresh = false;
+
+// Todo: Remove the type cast once we import from `@wordpress/interactivity`.
+export const { state, actions } = ( store as typeof StoreType )< Store >(
+	'woocommerce',
+	{
+		actions: {
+			// Question: Should we name this `addCartItem` to match the Store API naming?
```
luisherranz: Question: Should we name this `addCartItem` to match the Store API naming?

It would be good to take a look at the Store API to see what other actions we need to create and if it makes sense for the names of those actions to correspond more closely.

plugins/woocommerce-blocks/assets/js/base/stores/add-to-cart.ts
```diff
@@ -0,0 +1,102 @@
...
+			// Question: Should we use regular arguments or wrap them in an object?
```
luisherranz: Question: Should we use regular arguments or wrap them in an object?

It would be good to maintain consistency among all actions and use either arguments or objects in all of them.

`addToCart( id, quantity )` vs `addToCart( { id, quantity } )`

To do this, it would be good to take a look at the Store API to see what other actions we need to add and see if in some of them it makes more sense to use objects.

plugins/woocommerce-blocks/assets/js/base/stores/add-to-cart.ts
```diff
@@ -0,0 +1,102 @@
...
+							headers: {
+								Nonce: state.nonce,
+								'Content-Type': 'application/json',
+							},
+							body: JSON.stringify( item ),
+						}
+					);
+
+					if ( ! res.status.toString().startsWith( '2' ) )
+						// Question: what error string should we use? Is there a standard in Woo?
```
luisherranz: Question: what error string should we use? Is there a standard in Woo?

plugins/woocommerce-blocks/assets/js/base/stores/add-to-cart.ts
```diff
@@ -0,0 +1,102 @@
...
+				} catch ( error ) {
+					// Todo: Handle error using the new Store Notices block.
+					// const { actions } = store('woocommerce/store-notices');
+					// actions.addNotice(...);
+
+					// Reverts the optimistic update.
+					item.quantity = previousQuantity;
+				}
+			},
+			// Question: Is `refreshCart` the best name for this action?
```
luisherranz: Question: Is `refreshCart` the best name for this action?

Specifically for this action that does not exist directly in the Store API (it is simply the endpoint `/cart`), it would be good to think of a name that represents what it does and the use that will be given to it.

plugins/woocommerce-blocks/assets/js/base/stores/add-to-cart.ts
```diff
@@ -0,0 +1,102 @@
...
+
+				try {
+					const res: Response = yield fetch(
+						`${ state.restUrl }wc/store/v1/cart`,
+						{ headers: { 'Content-Type': 'application/json' } }
+					);
+					state.cart = yield res.json();
+
+					if ( ! res.status.toString().startsWith( '2' ) )
+						// Question: what error string should we use? Is there a standard in Woo?
```
luisherranz: Question: what error string should we use? Is there a standard in Woo?

plugins/woocommerce-blocks/assets/js/base/stores/add-to-cart.ts
```diff
@@ -0,0 +1,102 @@
...
+						throw new Error( 'Failed to refresh the cart.' );
+				} catch ( error ) {
+					// Question: what should we do if this fails? Should we retry? Inform the user?
```
luisherranz: Question: what should we do if this fails? Should we retry? Inform the user?

What is happening today when that occurs in other places, such as the Minicart block?

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
```diff
@@ -82,9 +83,14 @@ protected function render( $attributes, $content, $block ) {
 		$post_id = isset( $block->context['postId'] ) ? $block->context['postId'] : '';
 		$product = wc_get_product( $post_id );
 
+		// Initialize the "Add To Cart" store part.
+		// Question: Is this ok for 3PD or should we use a global function like `woocommerce_interactivity_use_add_to_cart_store()`.
```
luisherranz, line 97: Question: Is this ok for 3PD or should we use a global function like `woocommerce_interactivity_use_add_to_cart_store()`?

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
```diff
@@ -132,13 +137,14 @@ protected function render( $attributes, $content, $block ) {
 			* @param number $default_quantity The default quantity.
 			* @param number $product_id The product id.
 			*/
+			// Question: Is this info in the Store API cart?
```
luisherranz: Question: Is this info in the Store API cart?

plugins/woocommerce-blocks/assets/js/base/stores/add-to-cart.ts
samueljseay: There is an error string returned by the API that describes the possible error or validation issue, seems to me we should just use that.

or do you mean a string for fallback when there is no message, ie a 500 or something? I don't know of a single standard string for that kind of thing.

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/button/frontend.tsx
luisherranz: It's not the same. It looks like this is a [legacy filter](https://github.com/woocommerce/woocommerce/blob/579ecb010843352573c1810168b16e1da03018d3/docs/cart-and-checkout-blocks/hooks/migrated-hooks.md).

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/button/frontend.tsx
luisherranz: Not really, as yield will also split the task.

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
luisherranz: It's not the same. It looks like this is a [legacy filter](https://github.com/woocommerce/woocommerce/blob/579ecb010843352573c1810168b16e1da03018d3/docs/cart-and-checkout-blocks/hooks/migrated-hooks.md).

plugins/woocommerce-blocks/assets/js/base/stores/add-to-cart.ts
luisherranz: Cool, I have refactored it to use that.

plugins/woocommerce-blocks/assets/js/base/stores/add-to-cart.ts
luisherranz: It now uses the errors returned by the Store API.

luisherranz:
I have refactored the code a bit to address some of the questions and to use the `items` endpoint instead of the general `cart` endpoint.

I still wonder how we should structure the global store, if it still should be something closer to the structure of the Store API.

plugins/woocommerce-blocks/assets/js/base/stores/add-to-cart.ts
luisherranz: For now, I have introduced a retry with a timeout that grows exponentially.

plugins/woocommerce-blocks/assets/js/base/stores/add-to-cart.ts
luisherranz: For now, I have left it as `refreshCartItems`. Although, I am not sure if it is better to only have the `refreshCart` action (refresh of the complete cart).

plugins/woocommerce-blocks/assets/js/base/stores/add-to-cart.ts
luisherranz: As there are parts of the Store API that have a few arguments, including optional arguments ([for example](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/StoreApi/docs/checkout.md#process-order-and-payment)), I believe that using objects as the standard is going to lead to a better DX.

plugins/woocommerce-blocks/assets/js/base/stores/add-to-cart.ts
luisherranz: For now, I've renamed it to `addCartItem` to try to make it more similar to the Store API, but we should keep an eye on this as we progress.

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/button/frontend.tsx
luisherranz: For now, I'm going to leave it as a dynamic import, but I still want to think about whether it makes sense to avoid preloading this module.

plugins/woocommerce-blocks/assets/js/base/stores/cart-items.ts
```diff
@@ -0,0 +1,136 @@
...
+
+let pendingRefresh = false;
+let refreshTimeout = 3000;
+
+// Todo: Remove the type cast once we import from `@wordpress/interactivity`.
+// Question: disable "used before defined" lint rule?
+export const { state, actions } = ( store as typeof StoreType )< Store >(
+	'woocommerce',
+	{} // Todo: Fix this type problem, we should not require an empty object here.
+);
```
luisherranz: I am testing to infer the types of the client's store part, but I am not very clear on what is best here, especially because it is not very clean due to the rule that does not allow referencing variables that have not yet been defined.

luisherranz: I have reverted this for now and went back to manually typing the whole store.

samueljseay (approved):
Happy to merge this as is. As we discussed we can continue to refine in the feature branch

luisherranz:
I have added more questions and I'm still not convinced about how to structure certain things, especially in the aspect of error handling, but I believe we can merge this pull request as it is to continue with the migration and finish those designs later.

luisherranz:
By the way, the CI is complaining that I haven't added a changelog. If you think it's necessary, we can add it later.

luisherranz:
Other e2e tests are failing as well, but I'm going to merge this now and we can look at them in the parent PR.

PR #54597, opened 2025-01-17
New endpoint in orders/{id}/actions to send email

### LINKED ISSUE #53810: Enable sending order-related customer emails via REST API request
In #52050 we added a new REST endpoint to "Send order details" to the customer, which matches one of the actions you can take on an order in WP Admin. However, we have since identified the need to be able to send other order-related emails to the customer via REST as well. This is because for some REST clients, the typical flow is to create and order and process and payment _before_ adding customer details like an email address to the order. The order may progress to the `completed` status before an email address is added, in which case, none of the usual transactional emails to the customer will be sent.

So the proposal here is:

* Create a new endpoint, or set of endpoints, off of `wc/v3/orders/{order ID}/` for sending specific emails related to the order to the customer.
* The endpoint(s) should have an optional parameter to specify the email address to send to. This would get added to the order as the customer's billing address during the fulfillment of the request. This way the address can be added and the email sent all in one request.

coreymckrill:

This revamps the existing OrdersActionsRestController and adds some additional endpoints:
* A GET endpoint at `orders/{id}/actions/email_templates` to get a list of email templates that can be sent for this particular order.
* A POST endpoint at `orders/{id}/actions/send_email` where you can trigger the sending of an email template for a particular order.

This also adds some other new things:
* A helper method for adding or updating the billing email of the order.
* Filter and action hooks so that extensions that register additional email templates can include them in the list of available templates for an order.

Initially this new functionality was going to be added as an entirely separate endpoint, but after reviewing the existing functionality in the `send_order_details` order action, we decided that this would fit better, and allow us to have less duplicated code, if we just expanded the existing endpoint.

Fixes #53810

### Post-merge

- [x] Developer advisory about change to the `send_order_details` endpoint, and about embedded data in `orders` endpoint
- [x] Update REST API docs

1. Install a plugin like [WP Mail Logging](https://wordpress.org/plugins/wp-mail-logging/) on your test site so you can observe outgoing email.
1. Via the web UI, create an order with at least two products in it (or one product with at least 2 for quantity). For now you can leave the status at Pending, and don't add a billing email.
2. Now fire up your REST API client of choice and make a request for the order you just created:
    `GET wc/v3/orders/{order ID}`
3. In the response, look for the `_links` property, and find the new `email_templates` item within it.
4. Make a request to the URL provided in the `email_templates` link (`GET wc/v3/orders/{order ID}/actions/email_templates`). You should get back an empty array, because the order doesn't have a billing email set yet, so there are no email templates available to send.
5. Try to send an email for the order using `POST wc/v3/orders/{order ID}/actions/send_email`. You'll need to include a template ID in the payload:
    `{"template_id": "customer_invoice"}`
    You should get a message back saying "Order does not have an email address.".
8. Ok, now make the same request to `send_email`, but include params for both `template_id` and `email`:
    `{"template_id": "customer_invoice", "email": "someone@example.org"}`
    This should update the order to have a billing address. Check the mail logger and see that an Order Details email was sent.
9. The order has a billing email now, so make another request to get the order data, but with one extra parameter:
    `GET wc/v3/orders/{order ID}?_embed=email_templates`
10. In the response, you should see a new `_embedded` property. Within that should be an `email_templates` item with an array of available template IDs. This order should currently only have one available template, `customer_invoice`.
11. If you make another request to the `email_templates` link, you should get back a similar array as the one in the `_embedded` property, but each item should also include a title and description in addition to the ID.
12. Update the order status to Processing, and then request the order data again. Now you should have two template IDs available, `customer_invoice` and `customer_processing_order`.
13. Try to send the `customer_processing_order` template, but with a different email address:
    `POST wc/v3/orders/{order ID}/actions/send_email`
    `{"template_id": "customer_processing_order", "email": "someone.else@example.org"}`
14. You should get an error, because the order already has a billing email set.
15. Try again, but include the force param:
    `{"template_id": "customer_processing_order", "email": "someone.else@example.org", "force_email_update": "true"}`
16. This time the email should send successfully, and the order should now have the new value for the billing email.
17. Try various other order statuses and see which email templates are available. Try making a partial refund on the order, and see what other template then becomes available for it.


plugins/woocommerce/src/Internal/Orders/OrderActionsRestController.php
```diff
@@ -130,20 +524,95 @@ public function send_order_details( WP_REST_Request $request ) {
...
+
+		if ( $existing_email === $email ) {
+			return '';
+		}
+
+		if ( $existing_email && true !== $force ) {
+			return new WP_Error(
+				'woocommerce_rest_order_billing_email_exists',
+				__( 'Order already has a billing email.', 'woocommerce' ),
+				array( 'status' => 500 )
```
albarin: Would it make more sense to use a 400 status code here instead of 500, since this is a validation case (existing email) rather than an unexpected server error? 🤔

plugins/woocommerce/src/Internal/Orders/OrderActionsRestController.php
```diff
@@ -130,20 +524,95 @@ public function send_order_details( WP_REST_Request $request ) {
...
+	 * @param string   $email The email address to maybe add to the order.
+	 * @param bool     $force Optional. True to update the order even if it already has a billing email. Default false.
+	 *
+	 * @return string|WP_Error A message upon success, otherwise an error.
+	 */
+	private function maybe_update_billing_email( WC_Order $order, string $email, ?bool $force = false ) {
+		$existing_email = $order->get_billing_email( 'edit' );
+
+		if ( $existing_email === $email ) {
+			return '';
```
albarin, line 595: Returning an empty string here makes the `$messages` array to have two elements (empty string + the actual message) so when imploding we end up with an extra space at the beginning of the message: 👇 
https://github.com/woocommerce/woocommerce/blob/cceffbee664162384398dd262dda53abc3a8a4c8/plugins/woocommerce/src/Internal/Orders/OrderActionsRestController.php#L542-L544
Screenshot 2025-01-27 at 14 45 09

albarin (commented):
Nice work @coreymckrill 👏 
I've left a couple of review comments and also this small thing I found about the testing steps:

On step `6. Try sending an email for the order using POST wc/v3/orders/{order ID}/actions/send_email. You should get a message back saying "Order does not have an email address."`, I'm getting:
```json
{
  "code": "rest_missing_callback_param",
  "message": "Missing parameter(s): template_id",
  "data": {
    "status": 400,
    "params": [
      "template_id"
    ]
  }
}
```

I'm only getting the `"Order does not have an email address."` when I send the `template_id` in the request body, I think that's correct right? It's the test step that needs to be updated.

plugins/woocommerce/src/Internal/Orders/OrderActionsRestController.php
coreymckrill: Yeah, that makes more sense, good call. Updated in 17e8d4c2fb

plugins/woocommerce/src/Internal/Orders/OrderActionsRestController.php
coreymckrill: Oops, I added `array_filter` to `send_email` to address this issue, but forgot to add it to `send_order_details`. Good catch. Fixed in 0c4fad6634

albarin (approved):
Thanks for the fixes, is looking good! ✅

staskus:
👍  Thank you again! This adds comprehensive emailing capabilities that I'm excited to see!

✅  I tested all the scenarios and they work as described!

I also tried to model these new APIs within our current IPP/POS flows. This is how I imagine they could be integrated in the beginning:

#### 🟢 Succesful payment

Merchant selects "Email Receipt", and enters an email address. We send:

PATH: `wc/v3/orders/{order ID}/actions/send_email`
BODY: `{"template_id": "customer_completed_order", "email": "someone@example.org"}`

Optionally, we would immediately pass `"force_email_update": "true"` with a note that this action overrides the billing email address.

It sends a “Thanks for shopping with us” email. The same one is sent if customer's email is attached to the order before the payment. 

#### 🔴 Failed payment

Same flow. Merchant selects "Email Receipt", and enters an email address. We send an email with a different template:

PATH: `wc/v3/orders/{order ID}/actions/send_email`
BODY: `{"template_id": "customer_failed_order", "email": "someone@example.org"}`

It sends "Sorry, your order was unsuccessful” email 

### Question

These flows would depend on hardcoded `template_id` to avoid making another API call to:
- `GET wc/v3/orders/{order ID}/actions/email_templates` or
- `GET wc/v3/orders/{order ID}?_embed=email_templates`

As a designer of this API, do you think it's reasonable to hardcode template_ids? We would like to avoid making another API call, especially if it doesn't add additional functionality.

I do imagine additional flows, that could utilize `actions/email_templates`. For example, having a broader email functionality within order details would allow merchants to pick from a list of available templates to send to the customer.

coreymckrill:
@albarin sorry, forgot to mention that I also updated the testing instructions from your feedback. Does it look better now?

coreymckrill:
@staskus 

> These flows would depend on hardcoded template_id to avoid making another API call...

I was assuming that there would be a time somewhere in the mobile app flow where you would GET the order data and could store the available template IDs for future use. 🤔 But maybe that wouldn't always be the case? One easy thing we could do here is add an enum to the schema for the `template_id` property so it would show what all the possible IDs are, even if they aren't all available for the current order. Then you could do an OPTIONS request to get that list and cache it for later use... would that be helpful?

> do you think it's reasonable to hardcode template_ids?

I don't think hardcoding the template ID values is necessarily problematic, you're probably already doing that with other things like order statuses? But, because this new endpoint is extensible, and other extensions such as WooPayments could add their own email templates to the list, you'd probably still want to have a flow somewhere that retrieves the list just to see what's available.

coreymckrill:
Ok, I added the enum to the schema...

staskus:
@coreymckrill 

> I was assuming that there would be a time somewhere in the mobile app flow where you would GET the order data and could store the available template IDs for future use. One easy thing we could do here is add an enum to the schema for the template_id property so it would show what all the possible IDs are, even if they aren't all available for the current order

Yes, it's more likely we prefer to cache available email templates rather than rely on additional calls during the payment process. Unless we update our flows to let the merchant pick a desired template, we will still need to decide within the code which templates we want to send at which moment in time. Therefore, having hardcoded template_ids within the code will be hard to avoid. The ability to fetch all available or all template IDs allows to add of an extra layer of stability to the solution.

> But, because this new endpoint is extensible, and other extensions such as WooPayments could add their own email templates to the list, you'd probably still want to have a flow somewhere that retrieves the list just to see what's available.

Yes. I think what this endpoint opens up - the ability to have a more comprehensive emailing functionality from within the app. That's where the functionality to fetch templates with their titles and allow merchants to pick the desired one could be a nice addition.

PR #54600, opened 2025-01-17
Fix text color inheritance in Coming Soon templates

### LINKED ISSUE #54037: Coming soon block text color is not applied
### Issue description

Follow-up to https://github.com/woocommerce/woocommerce/pull/53910#issuecomment-2559565087

It seems texts in coming soon block does not respect the text color settings. This applies to multiple patterns, including the default entire site pattern.

Image

#### To reproduce:

1. Ensure your site is set to coming soon entire site mode
1. Go to `Appearance > Editor > Templates`
2. Open coming soon template
3. Select the coming soon block and set the text color to some other color
4. Observe the color of heading remains unchanged
5. Create a new paragraph and type something
6. Observe it also does not follow the text color from coming soon block

### Acceptance criteria

1. I think the best solution is to ensure text blocks adhere to the following CSS specificity, in order from highest to lowest priority:
    - Styles from its own block
    - Styles from coming soon block
    - Styles from theme / WordPress
2. Ensure this fix is applied to all coming soon patterns

chihsuan:

Closes https://github.com/woocommerce/woocommerce/issues/54037.

This PR ensures that text color settings in "Coming Soon" templates are properly applied and respect the intended CSS specificity hierarchy:

1. Styles explicitly set on the text block.
2. Styles inherited from the "Coming Soon" block.
3. Styles provided by the theme or WordPress defaults.

There are two reason why the custom color is not applied to the heading:

1. We have CSS set heading color to `var(--wp--preset--color--contrast);`
2. The Cover block default set the text color to black or white when the text color is not set.

To fix this

1. I removed the heading CSS and use the block settings styles instead.
2. I set the default Cover block text color to `inherit`.

1. Create a fresh site with `coming-soon-newsletter-template` feature flag enabled.
1. Set your site to "Coming Soon"
2. Navigate to `Appearance > Editor > Templates` and open the "Coming Soon" template.
3. Test the following scenarios:
   - Change the text color of the entire "Coming Soon" block and verify that:
     - All headings and paragraphs within the block should be updated to match the new text color.
   - Apply a custom color to individual text blocks (e.g., a heading or paragraph) and confirm that:
     - The custom color overrides the "Coming Soon" block color.
4. Create new text elements (e.g., paragraphs or headings) and verify they inherit the "Coming Soon" block's text color by default.
5. Verify the fix works in all "Coming Soon" templates (Sidebar > Templates > Design).

Note that some templates have blocks that explicitly set the text color to match the design so you will need to clear the text color setting in the block settings to verify it can be set by the "Coming Soon" block color. See recording:

https://github.com/user-attachments/assets/099a69b2-4075-4282-b216-e3f96213ac75

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


moon0326 (approved):
LGTM and tested well 👍 🚀

adrianduffell (changes requested):
This is testing well, thanks @chihsuan !

I'm just wondering about the Login button. I think it could be better to prevent it from inheriting colors on the default (purple) design since that has a different style to the rest of the design (black button / white text).

Currently, if I change the coming soon block text color from black to a dark blue, the button is no longer readable. I expected it to stay as white text. I think it makes sense for the button to inherit the colors for other designs though 🤔 

image

chihsuan:
> I'm just wondering about the Login button. I think it could be better to prevent it from inheriting colors on the default (purple) design since that has a different style to the rest of the design (black button / white text).
> 
> Currently, if I change the coming soon block text color from black to a dark blue, the button is no longer readable. I expected it to stay as white text. I think it makes sense for the button to inherit the colors for other designs though 🤔

Thanks for the feedback! @adrianduffell 

That's a good point. However, I couldn't find a way to do this. We've already set the custom color for the button, but the parent/layout color can still override it. The Editor dynamically adjusts the CSS. I think this would need a feature request to Gutenberg to support this.

https://github.com/woocommerce/woocommerce/pull/54600/files#diff-f8c9f2c4bab26df8d603ab7c79c01aa2f0a462e1cbc3bfc517212d5e358d4055R38

Screenshot 2025-01-22 at 15 28 53

I can also reproduce this with the TT5 theme, so it's not specific to our templates.

https://github.com/user-attachments/assets/564461e8-0007-41cd-bf47-37487d8b3ab5

adrianduffell (approved):
Thanks for looking into the button. Since it looks to be an issue outside our scope I'm approving!  🚀 

> That's a good point. However, I couldn't find a way to do this. We've already set the custom color for the button, but the parent/layout color can still override it. The Editor dynamically adjusts the CSS. I think this would need a feature request to Gutenberg to support this.

Gotcha, yeah sounds like it is worth creating an issue for on the Gutenberg side 🤔

chihsuan:
> Gotcha, yeah sounds like it is worth creating an issue for on the Gutenberg side 🤔

Thanks, I'll look into it more and create an issue if needed.

PR #54628, opened 2025-01-17
[e2e tests] Migrate global setup and teardown to project dependencies

### LINKED ISSUE #53187: [e2e tests] Migrate global setup and teardown to project dependencies
There are two ways to configure global setup and teardown with Playwright: using a global setup file and setting it in the config under [globalSetup](https://playwright.dev/docs/test-global-setup-teardown#option-2-configure-globalsetup-and-globalteardown) or using [project dependencies](https://playwright.dev/docs/test-global-setup-teardown#option-1-project-dependencies).
The recommended way is project dependencies, because it will produce traces and artifacts, while using `globalSetup` and `globalTeardown` will not produce traces or artifacts, and options like headless or specified in the config file are not applied.

We have a heavy setup and we will benefit from traces and artifacts. This, we should migrate our `globalSetup` and `globalTeardown` to project dependencies.

See more: https://playwright.dev/docs/test-global-setup-teardown

adimoldovan:

Closes https://github.com/woocommerce/woocommerce/issues/53187

Note: there are many changed files because most test files were updated, but most changes are small, like updated imports or a constant rename.

Replaced the global-setup and global-teardown with multiple setup files, used as project dependencies. This is the recommended way by Playwright.
It has some advantages over the classical approach:
- will produce traces or artifacts by default
- any test options (like headless mode) will be used by default
- more flexible - we can use project dependencies instead of conditionals (e.g. for the future the reset function could be move in its own setup file and use in the config of external environments only)
- better reporting and easier debugging - each setup/teardown function is reported just like a test is and we can easily see how long it takes and what exactly fails.

Other changes:
- the global authentication is done via an API request instead of the UI. This is faster an more reliable
- replaced the storage state environment variables with constants defined and exported by the config file
- moved the setup of the coming soon mode in a global setup function instead of setting it in each shopper test
- removed the `site.js` util which remained unused

All e2e tests pass. Check with multiple environments.
Nothing should change in the way the e2e tests work and are executed, the same setup should happen, only running from different .js files instead of a single one.
Each setup function should be logged as a test.


lanej0 (approved):
This looks good.  I ran the tests a bunch of times both locally and against an external environment.  There was a slight uptick in failures locally from trunk (46 vs. 43) but probably just due to general flakiness.

👍🏼

PR #54669, opened 2025-01-20
Fix fatal error on `WC_Customer_Data_Store_Session`

### LINKED ISSUE #50844: [Fatal] error: Uncaught Error: Call to a member function set() on null
### Prerequisites

### Describe the bug

The following error was observed post deployment in our logs:

```
PHP Fatal error:  Uncaught Error: Call to a member function set() on null in /wordpress/plugins/woocommerce/9.2.1/includes/data-stores/class-wc-customer-data-store-session.php:114
Stack trace:
#0 /wordpress/plugins/woocommerce/9.2.1/includes/data-stores/class-wc-customer-data-store-session.php(63): WC_Customer_Data_Store_Session->save_to_session(Object(WC_Customer))
#1 /wordpress/plugins/woocommerce/9.2.1/includes/class-wc-data-store.php(186): WC_Customer_Data_Store_Session->create(Object(WC_Customer))
#2 /wordpress/plugins/woocommerce/9.2.1/includes/abstracts/abstract-wc-data.php(241): WC_Data_Store->create(Object(WC_Customer))
#3 /wordpress/core/6.6.1/wp-includes/class-wp-hook.php(324): WC_Data->save('')
#4 /wordpress/core/6.6.1/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters(NULL, Array)
#5 /wordpress/core/6.6.1/wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#6 /wordpress/core/6.6.1/wp-includes/load.php(1280): do_action('shutdown')
#7 [internal function]: shutdown_action_hook()
#8 {main}
  thrown in /wordpress/plugins/woocommerce/9.2.1/includes/data-stores/class-wc-customer-data-store-session.php on line 114
```

It could be possibly due to the session object is not being initialized yet, or somehow session reaches an invalid state.

### Expected behavior

Fatal error shouldn't occur while calling `set` function on the session object.

### Actual behavior

Fatal error occurs.

### Steps to reproduce

There aren't any clear steps to replicate at the moment - To be updated after investigation.

### WordPress Environment

NA

### Isolating the problem

albarin:

This PR checks the `WC()->session` is not null before setting any data to avoid fatal errors.

Closes https://github.com/woocommerce/woocommerce/issues/50844

While we couldn't reproduce the fatal error organically, this is a small script to reproduce it and check it's fixed.

1. On `trunk`, add this script to `woocommerce.php`:
```php
add_action(
	'admin_init',
	function () {
		$customer = new WC_Customer();
		$customer->set_email( 'customer@example.com' );

		$session = new WC_Customer_Data_Store_Session();
		$session->save_to_session( $customer );
	}
);
```
2. Go to `/wp-admin` on your store and see that you get the fatal error: `Fatal error: Uncaught Error: Call to a member function set() on null in /plugins/woocommerce/plugins/woocommerce/includes/data-stores/class-wc-customer-data-store-session.php:116`
3. Checkout to this branch PR.
4. Reload the page and see you no longer get the fatal error.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Avoid fatal error on `WC_Customer_Data_Store_Session`.
</details>

<details>

#### Comment 

</details>


plugins/woocommerce/includes/data-stores/class-wc-customer-data-store-session.php
```diff
@@ -111,7 +111,10 @@ public function save_to_session( $customer ) {
 				$data[ $session_key ] = (string) $session_value;
 			}
 		}
-		WC()->session->set( 'customer', $data );
+
+		if ( WC()->session ) {
+			WC()->session->set( 'customer', $data );
+		}
```
coreymckrill: 🤔 I'm wondering if we should have a "doing it wrong" message or some other warning, rather than having it just fail silently here.

Another thought, would it make sense to bail at the top of the function if the session object is null, rather than doing the data processing (including some potential database calls) first?

albarin: Good suggestions @coreymckrill thanks!
I pushed the changes.

plugins/woocommerce/includes/data-stores/class-wc-customer-data-store-session.php
```diff
@@ -78,6 +78,15 @@ public function update( &$customer ) {
 	 * @param WC_Customer $customer Customer object.
 	 */
 	public function save_to_session( $customer ) {
+		if ( ! WC()->session ) {
+			wc_doing_it_wrong(
+				__METHOD__,
+				__( 'WC_Session is not available, customer data cannot be saved to session.', 'woocommerce' ),
```
coreymckrill, line 84: The only other suggestion I have would be to hint at _why_ WC_Session isn't available. Based on [this](https://github.com/woocommerce/woocommerce/blob/b5eff69e16c05435ebcb30bd72dffd7864376bce/plugins/woocommerce/includes/wc-core-functions.php#L2638-L2657), I would guess that the `woocommerce_init` hook has to fire first...

coreymckrill (approved):
👍 Successfully prevents the fatal. One other optional suggestion below, but marking as approved because it's otherwise ready to go.

PR #54682, opened 2025-01-20
Product Gallery: Remove "off" option from thumbnails

### LINKED ISSUE #54513: Product Gallery: Remove Thumbails block off option
Remove the Thumbails block off option from the Thumbnails block

kmanijak:

Closes https://github.com/woocommerce/woocommerce/issues/54513

1.Go to the Editor > Single Product template.
3. Click on the Product Gallery (Beta) block
4. **Expected:** Make sure "Thumbnails" have no "Off" option
5. Click on the Thumbnails block
6. **Expected:** Make sure "Thumbnails" have no "Off" option
7. Remove Thumbnails block and try adding it again
8. **Expected:** Make sure "Thumbnails" block can be inserted in Gallery Area

Before | After
-- | --
image | image

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce-blocks/assets/js/blocks/product-gallery/inner-blocks/product-gallery-thumbnails/edit.tsx
```diff
@@ -33,7 +33,7 @@ export const Edit = ( { attributes, setAttributes, context }: EditProps ) => {
 	} );
 
 	const Placeholder = () => {
-		return context.thumbnailsPosition !== ThumbnailsPosition.OFF ? (
```
tjcafferkey, line 36: We can remove the import for `ThumbnailsPosition` now I think.

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/inner-blocks/product-gallery-thumbnails/block-settings/index.tsx
```diff
@@ -53,7 +49,6 @@ export const ProductGalleryThumbnailsBlockSettings = ( {
 	const maxNumberOfThumbnails = 8;
 	const minNumberOfThumbnails = 3;
 	const { productGalleryClientId } = context;
-	// @ts-expect-error @wordpress/block-editor/store types not provided
 	const { updateBlockAttributes } = useDispatch( blockEditorStore );
 
 	return (
```
tjcafferkey, line 54: The `<Icon>` components are being flagged by the linter for missing attributes. We could fix this I think by just adding them with undefined values like so:

```tsx
<Icon
	size={ 32 }
	icon={ thumbnailsPositionLeft }
	onPointerEnterCapture={ undefined }
	onPointerLeaveCapture={ undefined }
/>
```

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/inner-blocks/product-gallery-thumbnails/edit.tsx
```diff
@@ -51,7 +51,7 @@ export const Edit = ( { attributes, setAttributes, context }: EditProps ) => {
 					);
 				} ) }
 			</div>
-		) : null;
+		);
 	};
 
 	return (
```
tjcafferkey, line 51: We're passing `attributes` and `setAttributes` down as props to `<ProductGalleryThumbnailsBlockSettings />` but [I don't think this component expects these props](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce-blocks/assets/js/blocks/product-gallery/inner-blocks/product-gallery-thumbnails/block-settings/index.tsx#L50)

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryThumbnails.php
```diff
@@ -119,7 +119,7 @@ protected function should_display_view_all( $mode, $thumbnails_count, $product_g
 	 * @return string Rendered block type output.
```
tjcafferkey, line 121: This does not always return a `string` type. We either should do, or update this

tjcafferkey: Its the same case with the `inject_view_all` method

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryThumbnails.php
```diff
@@ -119,7 +119,7 @@ protected function should_display_view_all( $mode, $thumbnails_count, $product_g
...
 	 */
 	protected function render( $attributes, $content, $block ) {
-		if ( isset( $block->context['thumbnailsPosition'] ) && '' !== $block->context['thumbnailsPosition'] && 'off' !== $block->context['thumbnailsPosition'] ) {
+		if ( isset( $block->context['thumbnailsPosition'] ) && '' !== $block->context['thumbnailsPosition'] ) {
```
tjcafferkey: We should probably check `context` exists first before operating on it

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryThumbnails.php
tjcafferkey: When defining `$number_of_thumbnails` we assume `$block->context['thumbnailsNumberOfThumbnails']` is a number but we don't explicitly check this. We probably should also check its an expected data type

tjcafferkey: I don't think we sanitize the `$html` before returning it. We could use `wp_kses_post` or `esc_html` Im not sure which is a better fit here? Probably the former?

tjcafferkey (changes requested):
Thanks for this @kmanijak Ive just left some comments for changes!

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/inner-blocks/product-gallery-thumbnails/block-settings/index.tsx
kmanijak: I don't think that's the right approach. I'm against adding unnecessary properties to satisfy linter/TS. I realised in most places we use Icon component from `@wordpress/components` and not `@wordpress/icons` like in here. So I replaced them and the linter issues are gone 👍

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/inner-blocks/product-gallery-thumbnails/edit.tsx
kmanijak: Good catch!

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryThumbnails.php
kmanijak: Nice ones!

> This does not always return a string type. We either should do, or update this

Fixed

> Its the same case with the inject_view_all method

Fixed

> When defining $number_of_thumbnails we assume $block->context['thumbnailsNumberOfThumbnails'] is a number but we don't explicitly check this. We probably should also check its an expected data type

Fixed

> I don't think we sanitize the $html before returning it. We could use wp_kses_post or esc_html Im not sure which is a better fit here? Probably the former?

Yeah, I used `wp_kses_post`. We expect HTML there. Good catch!

https://github.com/woocommerce/woocommerce/pull/54682/commits/6eb2122ff512f4f5f6a8d89ddc39c79a43ec102f

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryThumbnails.php
```diff
@@ -119,7 +121,11 @@ protected function should_display_view_all( $mode, $thumbnails_count, $product_g
 	 * @return string Rendered block type output.
 	 */
 	protected function render( $attributes, $content, $block ) {
-		if ( isset( $block->context['thumbnailsPosition'] ) && '' !== $block->context['thumbnailsPosition'] && 'off' !== $block->context['thumbnailsPosition'] ) {
+		if ( ! isset( $block->context ) ) {
+			return '';
+		}
+
+		if ( isset( $block->context['thumbnailsPosition'] ) && '' !== $block->context['thumbnailsPosition'] ) {
```
tjcafferkey: This function still silently fails if context exists but `$block->context['thumbnailsPosition']` does not. We should do one of two things here:
1. Reverse this condition and early-return an empty string
2. Add an `else` which returns an empty string

Option 1 would be preferable.

tjcafferkey (changes requested):
Thanks for fixing those, just one final thing which wasn't properly addressed before we can approve and merge I think!

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryThumbnails.php
kmanijak: Good catch!

https://github.com/woocommerce/woocommerce/pull/54682/commits/f0bdf28af9ef1ece32b149026450928307d8de29

tjcafferkey (approved):

PR #54685, opened 2025-01-20
Update uses of `PAYMENT_STORE_KEY` to use the `StoreDescriptor`

### LINKED ISSUE #54427: Update uses of `PAYMENT_STORE_KEY` and use the `StoreDescriptor` instead.
After some `@wordpress/data` types were updated our codebase started to show errors with `useSelect` and `useDispatch`. The cure for this is to use the types that are now provided with `@wordpress/data` and to update uses of the string `wc/store/payment` (often imported as `PAYMENT_STORE_KEY` and use the `StoreDescriptor` (i.e. the result of calling `createReduxStore`) instead.

Please see [this example PR](https://github.com/woocommerce/woocommerce/pull/54376) with some changes for the `wc/store/store-notices`  store.

Where we would previously use:

```js
import { PAYMENT_STORE_KEY } from '@woocommerce/block-data';

const {... } = useDispatch(
		PAYMENT_STORE_KEY
	);

const { ... } = useSelect ( ( select ) => select( PAYMENT_STORE_KEY ).someSelector() );
```

we would now do it like so:

```js
import { paymentStore } from '@woocommerce/block-data';

const {... } = useDispatch(
		paymentStore
	);

const { ... } = useSelect ( ( select ) => select( paymentStore ).someSelector() );
```

opr:

This PR updates uses of `PAYMENT_STORE_KEY` to use the `paymentStore` store descriptor instead.

It also updates an instance of `core/notices` being used and uses the store descriptor of `core/notices` (`noticesStore`) instead.

Closes #54427

The main thing to test is that several different payment methods work correctly with the Cart/Checkout blocks. You should test both regular and express payment methods and ensure there are no JS errors during the checkout process.

#### `usePaymentMethodState` and `usePaymentMethodInterface`

1. Install and activate Stripe. Activate some other payment methods.
2. Go to the Checkout block, ensure you can switch between payment methods.
3. Check out using [one of Stripe's failed cards](https://docs.stripe.com/testing#declined-payments) .
4. Ensure the errors prevent you from checking out.
5. Ensure you can check out with a valid card.

#### `PaymentEventsProvider`

1. Edit `plugins/woocommerce-blocks/assets/js/extensions/payment-methods/bacs/index.js` to the following:

```js
/**
 * External dependencies
 */
import { registerPaymentMethod } from '@woocommerce/blocks-registry';
import { __ } from '@wordpress/i18n';
import { getPaymentMethodData } from '@woocommerce/settings';
import { decodeEntities } from '@wordpress/html-entities';
import { sanitizeHTML } from '@woocommerce/utils';
import { RawHTML } from '@wordpress/element';

/**
 * Internal dependencies
 */
import { PAYMENT_METHOD_NAME } from './constants';

const settings = getPaymentMethodData( 'bacs', {} );
const defaultLabel = __( 'Direct bank transfer', 'woocommerce' );
const label = decodeEntities( settings?.title || '' ) || defaultLabel;

/**
 * Content component
 */
const Content = ( { paymentStatus } ) => {
	console.log( paymentStatus );
	return <RawHTML>{ sanitizeHTML( settings.description || '' ) }</RawHTML>;
};

/**
 * Label component
 *
 * @param {*} props Props from payment API.
 */
const Label = ( props ) => {
	const { PaymentMethodLabel } = props.components;
	return <PaymentMethodLabel text={ label } />;
};

/**
 * Bank transfer (BACS) payment method config object.
 */
const bankTransferPaymentMethod = {
	name: PAYMENT_METHOD_NAME,
	label: <Label />,
	content: <Content />,
	edit: <Content />,
	canMakePayment: () => true,
	ariaLabel: label,
	supports: {
		features: settings?.supports ?? [],
	},
};

registerPaymentMethod( bankTransferPaymentMethod );
```

7. In WooCommerce -> Settings -> Payments, enable BACS.
8. Install and enable Stripe.
9. Open the JS Console.
10. Add an item to your cart and go to the Checkout block.
11. Check "preserve log" in the JS console.
12. Complete checkout, and ensure you see the status (logged in `bacs/index.js`) change. Don't worry about it changing to all statuses, just ensure it logs without errors and changes.

#### `useCheckoutSubmit`

1. Ensure you can check out using an express payment method. While the express method is open, ensure the "Place order" button at the bottom of the checkout form is disabled.
2. Pay using an express payment method on Cart block, ensure the express methods appear disabled while you're interacting with one.

#### `CheckoutEventsProvider` and `updatePaymentMethods`

1. Go to WooCommerce -> Settings -> Payments and enable BACS.
3. Edit `plugins/woocommerce-blocks/assets/js/extensions/payment-methods/bacs/index.js`
4. Change the `canMakePayment` method to

```js
( { cart } ) => {
		return cart.billingAddress.country === 'US';
	}
```

5. Load the Checkout block and change the billing country to US. 
6. Ensure you see the BACS method as an available payment method.
7. Change it to a different country, ensure it disappears.
8. Change it back to US and ensure you can check out using BACS.

#### `PaymentMethodCard`

1. Install and activate Stripe.
2. As a logged in user, add an item to your cart and go to the Checkout block.
3. Pay using Stripe, and ensure you can save the method to your account. 
4. Check the box to save the method to your account, uncheck it and verify the state can be toggled.
5. Check it again, and check out.
6. Re-add an item to your cart and go back to the Checkout block.
7. Verify the method you just paid with is saved and can be used to check out again.

#### Incompatible payment method notice

1. Install an incompatible payment method (e.g. https://wordpress.org/plugins/express-checkout/)
2. Go to the Checkout block in the editor.
3. Select the Checkout Payments block.
4. Ensure you see a warning in the sidebar about the incompatible payment method.

#### Suppressing notices while express payment is active

1. Install and activate Stripe and set up the express payment methods.
2. Add a notice to the site by pasting this code into your JS console: `wp.data.dispatch( wp.notices.store ).createErrorNotice( 'Test notice', { context: 'wc/checkout' } );`
3. Start paying using the express payment methods.
4. Ensure the notice disappears.
5. Exit the express payment method without paying.
6. Ensure the notice returns.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce-blocks/assets/js/blocks-registry/payment-methods/registry.ts
```diff
@@ -11,15 +11,20 @@ import type {
...
 /**
  * Internal dependencies
  */
 import { default as PaymentMethodConfig } from './payment-method-config';
 import { default as ExpressPaymentMethodConfig } from './express-payment-method-config';
 import { canMakePaymentExtensionsCallbacks } from './extensions-config';
-
-import { STORE_KEY as PAYMENT_STORE_KEY } from '../../data/payment/constants'; // Full path here because otherwise there's a circular dependency.
+import { STORE_KEY as PAYMENT_STORE_KEY } from '../../data/payment/constants';
+import type { PaymentStoreDescriptor } from '../../data/payment';
```
opr, line 27: We can't get `paymentStore` from `@woocommerce/block-data` here as it introduces a circular dependency. The registry is really tightly coupled with the payment data store, so untangling these to remove the circular dependency would be too much work. Instead, I am importing only the _type_ of the `paymentStore` and casting the type returned by `dispatch`. It's the same result, just a little messier. This is the only file I had to do this in.

senadir (changes requested):
I didn't go too deep into testing payments (but I did test core ones), there's a console error in Checkout on load:

image

senadir:
it seems this change caused a 35kb increase (gzipped, so like 120kb) in the bundle size, is this expected?

opr:
> it seems this change caused a 35kb increase (gzipped, so like 120kb) in the bundle size, is this expected?

Unfortunately, yes, it's because we are now importing the store descriptor (so, actions, selectors etc.) rather than just the data store key string.

senadir:
Shouldnt those be already on another global script instead of being duplicated? is it worth it to add this load to skip a TS warning?

opr:
> Shouldnt those be already on another global script instead of being duplicated? is it worth it to add this load to skip a TS warning?

@senadir Yeah you're right, in [9b5f562](https://github.com/woocommerce/woocommerce/pull/54685/commits/9b5f562ffa0da572f2b9ee33c5f8758bc021b7bf) I removed the dependency from express payment methods, and reverted to using the store key with type casting. I think this is fine, it's something we can't get around because of a circular dependency.

plugins/woocommerce-blocks/assets/js/base/utils/create-notice.ts
```diff
@@ -33,14 +39,16 @@ export const createNotice = (
 	options: Partial< NoticeOptions >
 ) => {
 	const noticeContext = options?.context;
-	const suppressNotices =
-		select( 'wc/store/payment' ).isExpressPaymentMethodActive();
+	const selectors = select(
+		'wc/store/payment'
+	) as CurriedSelectorsOf< PaymentStoreDescriptor >;
+	const suppressNotices = selectors.isExpressPaymentMethodActive();
```
senadir, line 45: So this is the fix? nice.

opr: Yeah, not ideal, it's much nicer to get it from the store descriptor, but given the circular dependencies we can't, and this is what we'd get back anyway.

We should avoid doing this _where possible_ because it's harder to manage if the types change in future.

senadir (approved):
Testing and looking good thank you Thomas

PR #54706, opened 2025-01-21
Update Webpack Terser Plugin, use a shared config in blocks webpack configs to reduce duplication

samueljseay:

This is a chore PR that also tries to reduce a little noise in the many webpack configs maintained within `@woocommerce/block-library`. The basic optimization settings are exactly the same in every config, but
we repeat them over and over and it adds to maintenance burden.

It's possible some other parts could also move to shared config, but this one stands out as the most impactful as repeating the Terser plugin declaration over and over adds so much noise to `webpack-configs.js`.

I wanted to do this in my current work where I'm introducing an additional modules Webpack build, but it felt like a good idea to split this out into a smaller PR.

I believe that the following should be enough, this change is low risk:

* a basic smoke test that blocks don't crash such as cart, checkout, mini cart and adding products to cart via the product collection block (the default on your shop page). Check the browser console for fatal errors

* Seeing CI pass

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
@woocommerce/block-library: Extract some shared config for optimization in webpack-configs.js. Update webpack terser plugin to latest.

</details>

<details>

#### Comment 

</details>


mikejolley (approved):
Smoke tested and built well. 🚢 ship it!

mikejolley:
@samueljseay I did note that `woocommerce-monorepo/pnpm-lock.yaml` was updated after my build locally. Is that a concern?

samueljseay:
Hmm, ideally that shouldn't happen 🤔 I'll rebase this against trunk and reinstall/regenerate pnpm lock and see if that still happens.

samueljseay:
@mikejolley it looks to be an issue with pnpm 9.15.0. It changes the lock on this branch every time I install. 9.1.5.4 doesn't seem to have this behaviour so I'll get a PR up to update the packageManager field.

samueljseay:
Turns out it's not an issue with this branch specifically or 9.15.0, and it's being worked on, so I'll merge this.

PR #54708, opened 2025-01-21
Remove fileSuffix API from `@woocommerce/block-library` shared webpack configs as its no longer used.

samueljseay:

`fileSuffix` is an option that used to be passed to some of the webpack configs. If you look at the git history, it was used  to handle some `legacy` suffix added to some files in some cases. But now with the shared configs we never pass this option and the extra string handling we've added for this unused option adds noise to an already cluttered and noisy file.

This tidy up is something I spotted to do while I'm working on a new script module Webpack build, but to keep changes small and isolated from the purpose of my other work I've submitted this PR to clean up this aspect.

I don't believe there is anything we need to test here that is not covered by E2E and CI. We're not changing file names we're just removing code that added an empty string to file names 😅 

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Remove unused `fileSuffix` variable from shared Webpack configs in `@woocommerce/block-library`

</details>

<details>

#### Comment 

</details>


wavvves (approved):
Nice catch Sam 👍🏼

PR #54729, opened 2025-01-21
Introducing product tax status constants

wjrosa:

This PR introduces constants for all product tax status identifiers. The idea is to make it easier to reference them, avoid typos, make the code stricter, make it easier to find status usages, centralize them, improve documentation, and many other reasons that should help developers create code related to them.

- Checkout and build this branch on your test environment (`dev/introducing-product-stock-status-constants`)
    - Store setup
    - Product creation
    - Common checkout
    - Shortcode checkout
    - My account page navigation
    - Order listing/details

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Introduces and implements product tax status constants in the WooCommerce plugin codebase.

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Enums/ProductTaxStatus.php
```diff
@@ -0,0 +1,31 @@
...
+class ProductTaxStatus {
+	/**
+	 * Tax status for products that are taxable.
+	 *
+	 * @var string
+	 */
+	const TAXABLE = 'taxable';
+
+	/**
+	 * Tax status for products that are shipping only.
```
barryhughes: This reads a little like it is only used with physical products (which I guess is partly true, but still feels a little misleading to me) ... might this be clearer?

```suggestion
	 * Indicates that only the shipping cost should be taxed, not the product itself.
```

plugins/woocommerce/includes/abstracts/abstract-wc-order.php
```diff
@@ -19,6 +19,7 @@
 use Automattic\WooCommerce\Utilities\ArrayUtil;
 use Automattic\WooCommerce\Utilities\NumberUtil;
 use Automattic\WooCommerce\Utilities\OrderUtil;
+use Automattic\WooCommerce\Enums\ProductTaxStatus;
```
barryhughes: My understanding is that we *generally* arrange `use` statements alphabetically. 

To that end, what do you think of moving this up a few lines (so it comes right after `Enums\OrderStatus`? That will also 'group' it nicely with the other enum classes we are importing.

It would be best to apply this same principle to all files we're touching.

barryhughes (commented):
Looking pretty good to me! Left a few notes (all minor stuff), but we should also bring the branch up-to-date with `trunk` and hopefully that will get CI running.

Beyond that, I'll leave this for one of the other reviewers to give final approval ✅

plugins/woocommerce/includes/abstracts/abstract-wc-order.php
wjrosa: Oh, right, I didn't know/remember this rule. I have updated all statements accordingly in https://github.com/woocommerce/woocommerce/pull/54729/commits/a32105b477caeed0fc38c65ae9e8a6f99f948f73

coreymckrill (approved):
👍 All looks good, smoke tested in WP Admin and with Smooth Generator.

coreymckrill:
@wjrosa Looks like this needs a rebase, and you may need to re-run the failing CI checks (nothing looks like an actual problem, from what I can tell). Otherwise I think this is good to go.

PR #54733, opened 2025-01-21
[NOX] Add tracking events for the WooPayments payment methods selection flow

### LINKED ISSUE #10199: Implement tracks events for payment methods selection flow
### Description

Implement the following tracking events on the new payment methods selection flow (note on naming, we may want to use a `wcpay` prefix rather than `wcadmin`:

1. `wcadmin_settings_payments_payment_methods_pageview` - when the payment methods page is viewed
2. `wcadmin_settings_payments_payment_methods_continue` - when “Continue” is clicked on the payment methods page. This will also contain context variables for which payment methods were selected, and which were deselected - this is similar to the events we fire in the WooCommerce Core profiler, such as `wcadmin_coreprofiler_store_extensions_continue`.

### Acceptance criteria

* Events are created and validated.

anu-rock:

Closes [#10199](https://github.com/Automattic/woocommerce-payments/issues/10199)

Add tracking events for the payment methods selection flow during WooPayments activation on the new Payment settings page: page view and selected/not-selected PMs.

##### Local

1. Run `localStorage.setItem( 'debug', 'wc-admin:tracks' );` on your site browser console. And look for verbose logs.
2. Enable the new Payment settings page by going to WP Admin > Tools > WCA Test Helper > Features and turning on the `reactify-classic-payments-settings` flag. (requires the WC Beta Tester plugin).
3. Go to WP Admin > WooCommerce > Settings > Payments. Deactivate WooPayments if already active.
4. Click the Enable button for WooPayments. It will redirect you to the payment method selection page.
5. Verify that the `wcpay_settings_payment_methods_pageview` event was fired.
6. Toggle a few PMs or keep the default selection. Click the Continue button.
7. Verify that the `wcpay_settings_payment_methods_continue` event was fired with a map of all payment methods and their selection state (true or false).

image

##### JurassicNinja

1. Create a new JN site with WooCommerce on this PR's live branch and the WooCommerce Beta Tester plugin (here is a [direct JN create link](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce-beta-tester&woocommerce&branches.woocommerce=update/10199-woopayments-recommended-pms-add-tracks))
2. Go to WP Admin > WooCommerce menu item. You will be taken to the WooCommerce profiler. Click "Skip guided setup" and choose `US` as the store's base location country.
3. Go to WooCommerce > Settings > Advanced > WooCommerce.com and turn the `Enable tracking` flag on.
4. Install [Tracks Vigilante](https://github.com/Automattic/tracks-chrome-extension) browser extension and use its log of events.
5. Go to the Payments -> Settings page, Install and Enable WooPayments.
6. Follow the steps in the Local section above. Instead of browser console, verify that correct events are recorded in Tracks Vigilante.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/admin/client/settings-payments/test/settings-payments-methods.test.tsx
```diff
@@ -0,0 +1,53 @@
+/**
```
dmallory42, line 1: Thank you for adding tests for this behaviour!

plugins/woocommerce/client/admin/client/settings-payments/index.tsx
```diff
@@ -194,7 +195,13 @@ const SettingsPaymentsMethods = () => {
 	// Retrieve wooPayments gateway
 	const wooPayments = getWooPaymentsFromProviders( providers );
 
-	const onClick = useCallback( () => {
+	const onPaymentMethodsContinueClick = useCallback( () => {
+		// Record the event along with payment methods selected
+		recordEvent(
+			'wcpay_settings_payment_methods_continue',
+			paymentMethodsState
```
dmallory42: Currently it looks like this will be an object containing the PMs and either `true` / `false` depending on if they were selected.

My suggestion for making the data a bit easier to analyse would be to do something similar to the [wcadmin_coreprofiler_store_extensions_installed_and_activated](https://mc.a8c.com/tracks/events/event.php?eventname=wcadmin_coreprofiler_store_extensions_installed_and_activated) event (see the `installed_extensions` param) - so we would end up with something like:

```js
recordEvent(
'wcpay_settings_payment_methods_continue',
{
    selected_payment_methods: [ 'card', apple_pay', ... ],
    deselected_payment_methods: [ 'google_pay' ],
} );
```

This would also make it easier when registering the event as the param names would be fixed instead of dynamic. 

We'd need to add a bit of logic to filter the payment methods state in order to generate the lists of selected and deselected payment methods. Also I'm not 100% sure on the naming of those fields, so feel free to suggest any improvements!

plugins/woocommerce/client/admin/client/settings-payments/index.tsx
```diff
@@ -216,6 +223,7 @@ const SettingsPaymentsMethods = () => {
 
 		if ( location.pathname === '/payment-methods' ) {
 			hideWooCommerceNavTab( 'none' );
+			recordEvent( 'wcpay_settings_payment_methods_pageview' );
```
dmallory42, line 238: I'm not 100% sure, but generally I see people pass in `{}` as the second parameter, even when there are no additional params we want to track.

dmallory42 (approved):
LGTM - thank you for working on this! I 🚀  tested and everything is working well. I made a couple of suggestions for improvements I think we should consider before merge, but approving as this is in a good state once those are addressed!

plugins/woocommerce/client/admin/client/settings-payments/index.tsx
anu-rock: So, I found a few places where it's used without a second param. I think I'll leave it as it is, unless passing `{}` has a desired effect that we'd be missing out on?

https://github.com/woocommerce/woocommerce/blob/ad07839bfbd604a22ed2c41a633752cbfe1fc9ea/plugins/woocommerce/client/admin/client/homescreen/mobile-app-modal/index.tsx#L72-L80

https://github.com/woocommerce/woocommerce/blob/b3e76c41b4fa7bec4bdabc16163a21bd59ac351f/packages/js/product-editor/src/blocks/product-fields/attributes/edit.tsx#L63-L102

https://github.com/woocommerce/woocommerce/blob/a16f11ad6ee01389bdb68ee499b85c676e283d36/plugins/woocommerce/client/admin/client/marketplace/components/my-subscriptions/subscriptions-expired-expiring-notice.tsx#L81-L87

plugins/woocommerce/client/admin/client/settings-payments/index.tsx
anu-rock: Love your suggestion. I'll refactor it to make it so.

plugins/woocommerce/client/admin/client/settings-payments/test/settings-payments-methods.test.tsx
anu-rock: I'm glad they are helpful.

plugins/woocommerce/client/admin/client/settings-payments/index.tsx
dmallory42: In that case, it's fine! I only had a cursory look and wanted to make sure we were following patterns used elsewhere but it seems like it is the same in other places.

plugins/woocommerce/client/admin/client/settings-payments/index.tsx
anu-rock: Alright, done in https://github.com/woocommerce/woocommerce/pull/54733/commits/8a69569927a5f86537fc0bc80dccfdce056aeff1 ✅

dmallory42: Looks good to me, thanks for making that update. Let's :shipit: !

plugins/woocommerce/client/admin/client/settings-payments/index.tsx
```diff
@@ -194,7 +197,21 @@ const SettingsPaymentsMethods = () => {
 	// Retrieve wooPayments gateway
 	const wooPayments = getWooPaymentsFromProviders( providers );
 
-	const onClick = useCallback( () => {
+	const onPaymentMethodsContinueClick = useCallback( () => {
+		// Record the event along with payment methods selected
+		recordEvent( 'wcpay_settings_payment_methods_continue', {
```
vladolaru, line 202: I think we should only pass string values for event props, not objects. Here is [an example](https://github.com/woocommerce/woocommerce/blob/5d1959c5f3305e1a702d4a12f97ad04dd825354a/plugins/woocommerce/client/admin/client/payments-welcome/index.tsx#L119-L119) of turning a list into a coma-separated string for an event props.

AFAIK, Tracks only handles single values event props. Even booleans get translated. See this Slack message p1714122766970679-slack-C03KTTK2YMA.

It is best not to rely on some Tracks ingestion system (that may or may not be there) to automatically convert things.

plugins/woocommerce/client/admin/client/settings-payments/index.tsx
```diff
@@ -194,7 +197,21 @@ const SettingsPaymentsMethods = () => {
...
+			selected_payment_methods: Object.keys( paymentMethodsState ).filter(
```
vladolaru: Same here.

plugins/woocommerce/client/admin/client/settings-payments/index.tsx
```diff
@@ -194,7 +197,21 @@ const SettingsPaymentsMethods = () => {
...
+				( paymentMethod ) =>
+					paymentMethodsState[ paymentMethod ] === true
+			),
+			deselected_payment_methods: Object.keys(
```
vladolaru: Same here.

vladolaru (commented):
Thanks for working on this, Anurag. I left a couple of comments related to the data type we use for event props.

plugins/woocommerce/client/admin/client/settings-payments/index.tsx
vladolaru: cc @dmallory42

dmallory42: Great catch, @vladolaru!

anu-rock: Great catch! I missed this nuance when looking at the usage of recordEvent in our codebase. Fixed in https://github.com/woocommerce/woocommerce/pull/54733/commits/d53a9fa0de9913dd9bdd8875a6550ea45b1ebfed.

Issue #54664, opened 2025-01-20
[E2E] (Continuation of #54152) Audit the rest of the tests tagged with SKIP_ON_WPCOM


rodelgc:
This is a continuation of https://github.com/woocommerce/woocommerce/issues/54152 to continue auditing the rest of the e2e tests tagged with SKIP_ON_WPCOM.

The first PR that audited the first batch of e2e tests tagged with SKIP_ON_WPCOM: https://github.com/woocommerce/woocommerce/pull/54150

rodelgc:
# E2E tests that really should be skipped on WPCOM:

`plugins/woocommerce/tests/e2e-pw/tests/merchant/settings-woo-com.spec.js`.
- WPCOM doesn't have a _Settings > Advanced > WooCommerce.com_ page, although Pressable does.
- Also, it's impossible to disable Analytics and Marketplace Suggestions in WPCOM whether through the UI, REST API, or WP CLI. They are the two settings that this page contains.

**WPCOM:**
<img width="1205" alt="Image" src="https://github.com/user-attachments/assets/07640a86-d717-49f8-a56a-df5bff62157f" />

**Pressable**
<img width="1002" alt="Image" src="https://github.com/user-attachments/assets/26b9affc-d12c-4e34-af00-d30113c068b9" />

rodelgc:
# E2E tests that should not be skipped on WPCOM, but can't be unskipped for now

`plugins/woocommerce/tests/e2e-pw/tests/merchant/launch-your-store.spec.js`
- There's a [bug](https://github.com/woocommerce/woocommerce/issues/54699) in the classic admin interface of WPCOM sites where the "Coming soon" badge fails to display on the classic style, although it displays fine on the default WPCOM navigation style.
- I created a [task](https://github.com/woocommerce/woocommerce/issues/54700) to unskip this test in the future once that bug is fixed.
- For now, let's keep the SKIP_ON_WPCOM tag on this test until it's unblocked by the fix on that bug.

rodelgc:
The following PRs are now submitted for review:

https://github.com/woocommerce/woocommerce/pull/54748
https://github.com/woocommerce/woocommerce/pull/54758
https://github.com/woocommerce/woocommerce/pull/54762
https://github.com/woocommerce/woocommerce/pull/54766
https://github.com/woocommerce/woocommerce/pull/54769
https://github.com/woocommerce/woocommerce/pull/54771

Once they're merged, we can already close this task.

[...]

PR #54748, opened 2025-01-22
Unskip "Marketing Overview page have relevant content" E2E test on WPCOM

rodelgc:

Partially closes project task https://github.com/woocommerce/woocommerce/issues/54664.

This PR makes the E2E test `admin-marketing/overview.spec.js > Marketing Overview page have relevant content` runnable on a WPCOM site where it was previously failing, and was thus skipped.

Cuase of failure on WPCOM: WPCOM sites don't have a `Discover more marketing tools` and `Learn about marketing a store` cards in their Marketing Overview page. 

Screenshot 2025-01-22 at 11 35 15 AM

To address this, the checking of these 2 cards will now be done conditionally, depending on whether the site being tested is a WPCOM site or not.

The PR contains the following changes:
- Removes the `SKIP_ON_WPCOM` tag from the `Marketing Overview page have relevant content` test.
- Adds a step to check if the site being tested is hosted on WPCOM  by querying the WPCOM public api.
- Move the assertions for `Discover more marketing tools` and `Learn about marketing a store` cards into a conditional if-block that will only get executed when the site isn't hosted in WPCOM.

The steps below assumes you had already installed all PNPM dependencies and built the WooCommerce plugin.

#### Test on wp-env

1. Clone this branch and start up the local test env.
    ```sh
    pnpm env:test
    ```
3. Run this test and make sure it passes
    ```sh
    pnpm test:e2e admin-marketing -g "Marketing Overview page have relevant content"
    ```
1. Make sure no other e2e tests in the CI checks failed due to the changes in this PR.
 
#### Test on WPCOM and Pressable

1. Run the test on our WPCOM and Pressable sites and make sure `Marketing Overview page have relevant content` pass:
    ```sh
    export E2E_ENV_KEY=<...>
    pnpm test:e2e:with-env default-wpcom admin-marketing
    
    # Then do the same in Pressable
    pnpm test:e2e:with-env default-pressable admin-marketing
    ```

### DIFF
--- a/plugins/woocommerce/tests/e2e-pw/tests/admin-marketing/overview.spec.js
+++ b/plugins/woocommerce/tests/e2e-pw/tests/admin-marketing/overview.spec.js
@@ -10,9 +10,7 @@ test.describe( 'Marketing page', () => {
 		page,
 	} ) => {
 		// See if this is a WPCOM site.
-		const base_hostname = new URL( baseURL ).hostname;
-		const pattern_domain_name = /\.(?:wordpress\.com|wpcomstaging\.com)$/;
-		const is_wpcom_site = pattern_domain_name.test( base_hostname );
+		const is_wpcom_site = process.env.IS_WPCOM.toLowerCase() === 'true';
 
 		// Go to the Marketing page.
 		await page.goto( 'wp-admin/admin.php?page=wc-admin&path=%2Fmarketing' );
--- a/plugins/woocommerce/tests/e2e-pw/tests/admin-marketing/overview.spec.js
+++ b/plugins/woocommerce/tests/e2e-pw/tests/admin-marketing/overview.spec.js
@@ -48,11 +48,10 @@ test.describe( 'Marketing page', () => {
 			await expect(
 				page.getByRole( 'tab', { name: 'CRM', exact: true } )
 			).toBeVisible();
+			await expect(
+				page.getByText( 'Learn about marketing a store' )
+			).toBeVisible();
 		}
-
-		await expect(
-			page.getByText( 'Learn about marketing a store' )
-		).toBeVisible();
 	} );
 
 	test(
--- a/plugins/woocommerce/tests/e2e-pw/tests/admin-marketing/overview.spec.js
+++ b/plugins/woocommerce/tests/e2e-pw/tests/admin-marketing/overview.spec.js
@@ -6,7 +6,6 @@ test.describe( 'Marketing page', () => {
 	test.use( { storageState: ADMIN_STATE_PATH } );
 
 	test( 'Marketing Overview page have relevant content', async ( {
-		baseURL,
 		page,
 	} ) => {
 		// See if this is a WPCOM site.
--- a/plugins/woocommerce/tests/e2e-pw/tests/admin-marketing/overview.spec.js
+++ b/plugins/woocommerce/tests/e2e-pw/tests/admin-marketing/overview.spec.js
@@ -5,24 +5,34 @@ const { ADMIN_STATE_PATH } = require( '../../playwright.config' );
 test.describe( 'Marketing page', () => {
 	test.use( { storageState: ADMIN_STATE_PATH } );
 
-	test(
-		'Marketing Overview page have relevant content',
-		{ tag: tags.SKIP_ON_WPCOM },
-		async ( { page } ) => {
-			// Go to the Marketing page.
-			await page.goto(
-				'wp-admin/admin.php?page=wc-admin&path=%2Fmarketing'
-			);
+	test( 'Marketing Overview page have relevant content', async ( {
+		baseURL,
+		page,
+		request,
+	} ) => {
+		// See if this is a WPCOM site.
+		const base_hostname = new URL( baseURL ).hostname;
+		const is_wpcom_site = (
+			await request.get(
+				`https://public-api.wordpress.com/wp/v2/sites/${ base_hostname }`
+			)
+		 ).ok();
 
-			// Heading should be overview
-			await expect(
-				page.getByRole( 'heading', { name: 'Overview' } )
-			).toBeVisible();
+		// Go to the Marketing page.
+		await page.goto( 'wp-admin/admin.php?page=wc-admin&path=%2Fmarketing' );
 
-			// Sections present
-			await expect(
-				page.getByText( 'Channels', { exact: true } )
-			).toBeVisible();
+		// Heading should be overview
+		await expect(
+			page.getByRole( 'heading', { name: 'Overview' } )
+		).toBeVisible();
+
+		// Sections present
+		await expect(
+			page.getByText( 'Channels', { exact: true } )
+		).toBeVisible();
+
+		// Check the 'Discover more marketing tools' card and its individual tabs only on non-WPCOM sites.
+		if ( ! is_wpcom_site ) {
 			await expect(
 				page.getByText( 'Discover more marketing tools' )
 			).toBeVisible();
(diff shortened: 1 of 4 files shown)


plugins/woocommerce/tests/e2e-pw/tests/admin-marketing/overview.spec.js
```diff
@@ -4,24 +4,34 @@ const { tags } = require( '../../fixtures/fixtures' );
...
+		page,
+		request,
+	} ) => {
+		// See if this is a WPCOM site.
+		const base_hostname = new URL( baseURL ).hostname;
+		const is_wpcom_site = (
+			await request.get(
+				`https://public-api.wordpress.com/wp/v2/sites/${ base_hostname }`
+			)
+		 ).ok();
```
rodelgc: Initially, I thought about using these approaches in determining whether it's a WPCOM site or not:
- Using Woo's GET System Status endpoint and checking if `wpcomsh` is listed in the mu plugins from the response.
- Matching `baseURL` with domain names specific to WPCOM.

But I think using the WPCOM public api is simpler and more reliable in this case. All public WPCOM sites can be queried using that public endpoint anyway, and the routes pointing to more sensitive resources like `/users` were protected by that public endpoint when I checked. So the risks from an exposed WPCOM test site URL in our Playwright trace files is low.

We must also be careful not to publish the trace files of test runs from external sites here on the public repo anyway.

Stojdza (approved):
Changes look good and the test is passing on CI. 

I have the same issue as [here](https://github.com/woocommerce/woocommerce/pull/54758#pullrequestreview-2568359427), with the latest nightly WC and with WC 9.6, so I guess it's something with my local setup. Approving it ✅

rodelgc:
Thanks @Stojdza.

> I have the same issue as https://github.com/woocommerce/woocommerce/pull/54758#pullrequestreview-2568359427,

I retested this PR and it's failing for a different reason. It failed on my end not because of the error you encountered, but because the `Learn about marketing a store' card was being checked even on WPCOM. 

Screenshot 2025-01-23 at 12 20 50 PM

The correct behavior is that the test should NOT check for the presence of that card in WPCOM. I just pushed commit 3a37116abad1b63a95cb9eeb6f0ec108db6e0dd1 to do just that. I re-ran the test on both WPCOM and Pressable after this commit, and they both passed on my end.

Screenshot 2025-01-23 at 12 26 55 PM

Screenshot 2025-01-23 at 12 22 02 PM

But it would be good if you could review again by running the tests again on wpcom and Pressable after this latest commit. Apologies for that failure slipping into the PR review. I remember force-pushing into this branch due to a messy rebase and trunk merge so some earlier commits I did (including that one involving skipping the 'Learn about marketing a store' card on WPCOM) got lost along the way.

plugins/woocommerce/tests/e2e-pw/tests/admin-marketing/overview.spec.js
```diff
@@ -4,24 +4,34 @@ const { tags } = require( '../../fixtures/fixtures' );
...
+			page.getByRole( 'heading', { name: 'Overview' } )
+		).toBeVisible();
+
+		// Sections present
+		await expect(
+			page.getByText( 'Channels', { exact: true } )
+		).toBeVisible();
+
+		// Check the 'Discover more marketing tools' and 'Learn about marketing a store' cards only on non-WPCOM sites.
+		if ( ! is_wpcom_site ) {
```
Stojdza, line 30: While testing, I found that all `expect`s after this line are not running on Pressable as well. Is it expected? Per your comment
> The correct behavior is that the test should NOT check for the presence of that card in WPCOM

I understood that it's expected to not run only on WPCOM. Checks are passing against wp-env locally, which is expected 👍 

Here is the screenshot from Playwright UI (on the left is the test run against wp-env, on the right against Pressable):

Screenshot 2025-01-23 at 18 36 18

rodelgc: Thank you so much catching this @Stojdza. Yeah it turns out that Pressable sites were being returned by that endpoint too. So on my end, the `Discover more marketing tools` and `Learn about marketing a store` elements were not being tested as well.

I had just pushed a new commit (d9ced6ce6aa076ce79796cbafafdc84ec09aeca3) to use the domain name as the basis for determining whether it's a WPCOM site or not. The checks are currently running. Once they pass, they're ready for your re-review.

Locally I tested the change. Results: The assertions were skipped on WPCOM as expected, and were executed on Pressable also as expected.
Screenshot 2025-01-27 at 4 47 22 PM
Screenshot 2025-01-27 at 4 45 33 PM

Please test one more time on your end to be extra sure. Thanks a lot!

plugins/woocommerce/tests/e2e-pw/tests/admin-marketing/overview.spec.js
rodelgc: As mentioned [here](https://github.com/woocommerce/woocommerce/pull/54748#discussion_r1930170572), we'll try to use the approach based on the hostname instead.

plugins/woocommerce/tests/e2e-pw/tests/admin-marketing/overview.spec.js
rodelgc: The check [Core e2e tests 1/6](https://github.com/woocommerce/woocommerce/actions/runs/12985251504/job/36239037666?pr=54748#logs) is failing probably due to the issue in monorepo discussed in p1738004413624539-slack-C03CPM3UXDJ, which is unrelated to the changes in this PR. I'll update this branch with trunk and see if the checks pass this time.

rodelgc: `Core e2e tests 1/6` continues to fail even after updating with trunk. After looking at the [CI runs in trunk](https://github.com/woocommerce/woocommerce/actions/workflows/ci.yml?query=branch%3Atrunk), it seems that shard 1/6 have been failing there too, so these failures aren't related to this PR. 

It seems though, that the shard 1/6 failures are related to https://github.com/woocommerce/woocommerce/pull/54771 and I think I know why. I'll resolve that on a separate PR and give a heads-up on our monorepo channel.

adimoldovan:
@rodelgc What do you think about using an environment variable instead of checking the URL to detect the env? That can be set in the playwright config. For example, the default config (which uses wp-env) [sets and env variable USE_WP_ENV](https://github.com/woocommerce/woocommerce/blob/394a401d61462289ad99542698eafb9f33aa35b2/plugins/woocommerce/tests/e2e-pw/envs/default/playwright.config.js#L3). Similarly we could set variables in the [default-pressable](https://github.com/woocommerce/woocommerce/blob/c7fb0a7309deed515c7f3f2433dff400c03e7eee/plugins/woocommerce/tests/e2e-pw/envs/default-pressable/playwright.config.js) and [default-wpcom](https://github.com/woocommerce/woocommerce/blob/c7fb0a7309deed515c7f3f2433dff400c03e7eee/plugins/woocommerce/tests/e2e-pw/envs/default-wpcom/playwright.config.js) configs. I think variables like `IS_WPCOM` and `IS_PRESSABLE` could be useful in other areas of the suite.

rodelgc:
Thanks a lot for taking a look @adimoldovan I pushed a few new changes and it's ready for review again. I added the `IS_WPCOM` and `IS_PRESSABLE` environment variables on the 2 PW config files you mentioned.

I verified if they would run fine on each environment (localhost, Pressable, WPCOM) and it seems they did:
Screenshot from 2025-01-31 12-45-18
Screenshot from 2025-01-31 12-49-22
Screenshot from 2025-01-31 12-51-49

plugins/woocommerce/tests/e2e-pw/tests/admin-marketing/overview.spec.js
rodelgc: Resolved ☝️ and updated this branch with the latest changes from trunk.

adimoldovan (approved):
> Thanks a lot for taking a look @adimoldovan I pushed a few new changes and it's ready for review again. I added the IS_WPCOM and IS_PRESSABLE environment variables on the 2 PW config files you mentioned.

Thanks for following the suggestion @rodelgc. Let's merge it!

Issue #54464, opened 2025-01-14
Slow Admin\API\Reports\Order\DataStore::get_all_statuses() query being run when not needed


prettyboymp:
## Issue

The query generated by `Admin\API\Reports\Order\DataStore::get_all_statuses()` takes a couple of seconds to run on sites with a lot of orders.  The query is cached, however, when the value gets kicked out of cache for any reason, the query is rerun.

The problem is that this data is loaded on every page within the admin even if it isn't needed so other pages that do not need the distinct order statuses may end up running this query, delaying the request.

query: `SELECT DISTINCT status FROM wp_wc_order_stats`

Other than forcing a different DB engine, this query probably can't be improved much itself. 

## Expected Solution

We should avoid loading these settings when the page being requested doesn't need them.  This way, if/when the cached statuses get knocked out of cache, the query will only be run on pages that actually need the status for reports.

sabali33:
@prettyboymp I can I take a look at this issue?

prettyboymp:
> [@prettyboymp](https://github.com/prettyboymp) I can I take a look at this issue?

@sabali33, that would be great!

sabali33:
@prettyboymp Thanks I am taking it up. 🙂

sabali33:
@prettyboymp Could you take a look this PR for this issue: https://github.com/woocommerce/woocommerce/pull/54759

joshuatf:
@prettyboymp did you notice the cache not being hit quite a bit?  I frequently reproduced this issue in testing.

I think as part of this issue we may want to investigate why the cache is frequently invalidated.

prettyboymp:
> [@prettyboymp](https://github.com/prettyboymp) did you notice the cache not being hit quite a bit? I frequently reproduced this issue in testing.
> 
> I think as part of this issue we may want to investigate why the cache is frequently invalidated.

From my testing it appears that an order being modified triggers the cache clearing.  So an active site, that would also have more data to process, would run into this more frequently.

joshuatf:
Marking this high priority.  The query itself took around 310ms on a larger site in testing, which is not the largest query time we've seen, but this query does hydrate settings on most admin pages, making it relatively high impact.

PR #54759, opened 2025-01-22
Fix/54464

sabali33:

This PR seeks to prevent Admin\API\Reports\Order\DataStore::get_all_statuses() from running on every WooCommerce admin page.

It does so by checking the current admin page before allowing this function to run from the Automattic\WooCommerce\Internal\Admin\Settings::add_component_settings function.

1. Create a WordPress site with the latest version and login to the dashboard area
2. Go to plugins to install and activate WooCommerce plugin.
3. Before this PR, if you visit any Woocomerce admin page including the Analytics, products, etc and run this code in the page browser console, `window.wc.wcSettings.getSetting('admin')` you should see the resulting object contains `unregisteredOrderStatuses
` property.
4. After this PR, if you repeat step 3 on any of the Analytics admin pages, you should see the `unregisteredOrderStatuses` property.

#### Significance

-   [ ] Patch
-   [x ] Minor
-   [ ] Major

#### Type


plugins/woocommerce/src/Internal/Admin/Settings.php
```diff
@@ -366,4 +369,9 @@ private function get_custom_settings( $settings ) {
 		}
 		return $settings;
 	}
+	private function is_analytics_settings_page(){
```
Konamiman: We require all the class methods to have a documentation comment, could you please add one here?

Konamiman (changes requested):

Konamiman:
Hi @sabali33, thanks for your contribution. Could you please add testing instructions to your pull request? In other words, 
how could we verify that your proposed change improves the site loading performance and doesn't break existing functionality? Thank you!

sabali33:
@Konamiman I have added testing instructions and a little more context. Thanks

plugins/woocommerce/src/Internal/Admin/Settings.php
```diff
@@ -212,7 +212,10 @@ public function add_component_settings( $settings ) {
 		$settings['woocommerceTranslation'] = __( 'WooCommerce', 'woocommerce' );
 		// We may have synced orders with a now-unregistered status.
 		// E.g An extension that added statuses is now inactive or removed.
-		$settings['unregisteredOrderStatuses'] = $this->get_unregistered_order_statuses();
+		 if($this->is_analytics_settings_page()){
```
joshuatf: Unfortunately, we can't limit this setting only to admin/analytics pages because `wc-admin` routes are loaded as an SPA and this data must be hydrated on all `wc-admin` routes.

We might be able to use the `PageController::is_admin_page()` check to mitigate how often this query is run.  It might be good to wrap a bunch of these other settings as well, but cc @moon0326 in case you know a comprehensive list of which settings are needed for `wc-admin` pages vs connected pages.

plugins/woocommerce/src/Internal/Admin/Settings.php
```diff
@@ -366,4 +369,15 @@ private function get_custom_settings( $settings ) {
 		}
 		return $settings;
 	}
+
+	/**
+	 * Check if current page is a analytics settings page
+	 * 
+	 * @return bool
+	 */
+	private function is_analytics_settings_page() {
```
joshuatf: We can remove this function based on above comment in favor of `PageController::is_admin_page()`.

joshuatf (changes requested):
Thanks for the PR, @sabali33!  I left a couple comments in the code though I think even with those comments it's worth noting that this is only a partial fix for #54464.  This query still needs to be run on a lot of pages, so we probably want to investigate a cached solution in addition to limiting how often it's run.

I also see that you responded to @Konamiman above, but I'm not seeing testing instructions in the PR description.  Could you please add instructions on how to test this PR?

plugins/woocommerce/src/Internal/Admin/Settings.php
sabali33: Sure that makes sense. I will use the ``PageController::is_admin_page()` to check.

sabali33:
@joshuatf Concerning the testing instructions, I did type it and have no idea why it didn't get published.
@Konamiman sorry for the confusion.

I have updated this PR and addressed the raised issues.

plugins/woocommerce/src/Internal/Admin/Settings.php
```diff
@@ -212,7 +212,10 @@ public function add_component_settings( $settings ) {
 		$settings['woocommerceTranslation'] = __( 'WooCommerce', 'woocommerce' );
 		// We may have synced orders with a now-unregistered status.
 		// E.g An extension that added statuses is now inactive or removed.
-		$settings['unregisteredOrderStatuses'] = $this->get_unregistered_order_statuses();
+		 if(PageController::is_admin_page()){
```
joshuatf: There's a mixture of tabs and spaces on this line.  Can you update this to only use tabs?  Also the linter is complaining about lack of spaces around the if statement parentheses.

```suggestion
		if( PageController::is_admin_page() ){
```

plugins/woocommerce/src/Internal/Admin/Settings.php
```diff
@@ -212,7 +212,10 @@ public function add_component_settings( $settings ) {
...
+			$settings['unregisteredOrderStatuses'] = $this->get_unregistered_order_statuses();
+		 }
```
joshuatf: Also have an extra space on this line as well.

```suggestion
		}
```

joshuatf (commented):
* Tested that `unregisteredOrderStatuses` was still available under `wcSettings.admin` on various admin pages and worked as expected ✅ 
* Tested that `unregisteredOrderStatuses` is no longer available on connected pages (e.g., Products), however other `wcSettings` still exist ✅ 
* Tests are mostly passing with a couple exceptions, but I believe they are flaky and unrelated 🟡 

Thanks for working on this, @sabali33!  Left a couple comments regarding linting issues, but I think after those are fixed this PR is good to go.

sabali33:
@joshuatf Thanks for the review. I have addressed the issues raised 🙂

Konamiman (approved):

Konamiman:
Hi @joshuatf, since you requested changes now the pull request can't be merged until you approve it, could you please do it?

joshuatf (approved):
Changes look good!  Thanks for taking the time to work on this, @sabali33 😄

PR #54766, opened 2025-01-22
Unskip on WPCOM: `shopper/checkout-block.spec.js`

rodelgc:

Part of project task https://github.com/woocommerce/woocommerce/issues/54664.

This PR fixes the E2E test `shopper/checkout-block.spec.js` to be compatible with WPCOM.

Cause of failure in WPCOM: In the beforeAll hook, the test was trying to edit a customer's  billing/shipping address using a hard-coded customer id `2`. This works in wp-env because the test customer will always have `2` as the ID, but on external environments it won't be, causing the test to fail due to the bad request.

To fix this, the PR did the following:
1. Added a new `users` setup file to retrieve the id of the admin and test customer and save them as an environment variables accessible to other tests.
    - Other tests, especially the API ones, also use the hard-coded IDs. So saving them as an environment variable will allow the api tests to access them too.
2. Changed the hard-coded customer ID with the one from the environment variable.

It took me a while to pinpoint which request was causing the failure because the beforeAll hook contains a lot of them. So I organized them into `test.step()`s to help with any future debugging on that part of the code.

1. Run the test a few times and make sure it passes on WPCOM:
    ```sh
    E2E_ENV_KEY=<...> pnpm test:e2e:with-env default-wpcom shopper/checkout-block.spec.js
    ```
2. Do the same in Pressable, a few multiple times as well.
    ```sh
    E2E_ENV_KEY=<...> pnpm test:e2e:with-env default-pressable shopper/checkout-block.spec.js
    ```
    

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/tests/e2e-pw/tests/shopper/checkout-block.spec.js
```diff
@@ -167,7 +167,7 @@ test.describe(
 				enabled: true,
 			} );
 			// make sure our customer user has a pre-defined billing/shipping address
-			await api.put( `customers/2`, {
+			await api.put( `customers/${ process.env.CUSTOMER_USER_ID }`, {
```
rodelgc: The main change in this PR: Replace the hardcoded customer ID with one that's obtained from querying the `/customers` endpoint in `global-setup`.

adimoldovan:
@rodelgc, thanks for tackling this. Unfortunately, with https://github.com/woocommerce/woocommerce/pull/54628 merged the `global-setup.js` file was removed and created a conflict on this one.

Although, I'm not sure about the approach here. We had similar issues in other tests and the general approach was to create a customer in a fixture via the API. Or if not, move the call to getting the customer ID in a fixture of this test.

rodelgc:
Thanks @adimoldovan I think your work on https://github.com/woocommerce/woocommerce/pull/54628 is actually a much-needed change because our `global-setup.js` had been growing larger. I'll take care of handling the merge conflict that resulted from it.

> create a customer in a fixture via the API

With the new `setup` files in place, I think this is the best approach. I think just getting the user id of the customer will be enough for now since the customer's already set up in the Pressable and WPCOM sites, which is consistent with the wp-env starting setup. If this approach turns out to add more complexity than expected, I'll fall back to your second suggestion: _"move the call to getting the customer ID in a fixture of this test"_. 

Will fix the conflict and apply the suggestions on my next commit.

plugins/woocommerce/tests/e2e-pw/tests/shopper/checkout-block.spec.js
```diff
@@ -73,123 +73,143 @@ test.describe(
...
-				request,
-				baseURL,
-				'woocommerce_checkout_address_2_field',
-				'optional'
-			);
-			// make sure the currency is USD
-			await api.put( 'settings/general/woocommerce_currency', {
-				value: 'USD',
+
+			await test.step( 'Disable Coming Soon option.', async () => {
```
adimoldovan: This should not be required here, it is already disabled [in the setup](https://github.com/woocommerce/woocommerce/blob/9c48ef110ce9e716076d14c63ec12c8a5693f6c2/plugins/woocommerce/tests/e2e-pw/fixtures/site.setup.js#L104) phase.

rodelgc: Thanks @adimoldovan I pushed a new set of commits just now. The change to remove the disabling of the 'Coming soon' mode is now included in c19919e5c0c668c521c2fec8b81d18cf8baa648f.

rodelgc:
Done pushing a new set of changes:
- Removing of the 'coming soon' step as mentioned [here](https://github.com/woocommerce/woocommerce/pull/54766#discussion_r1936512511).
- Adding of a `users.setup.js` file to save the user IDs of the test admin and customer users as environment variables that are accessible to all tests.
- I also took the opportunity to replace all `wcApi` instantiations with the `api` fixture in this test in b57dbac49f440aae344cbb77c0254466d6ba8fee.

~~When the checks pass, it would be ready again for review.~~
Checks have passed, ready again for review @adimoldovan

adimoldovan:
> Adding of a users.setup.js file to save the user IDs of the test admin and customer users as environment variables that are accessible to all tests.

@rodelgc ☝🏼 this is the part I'm not sure of, doing it as a global setup step. This setup will run for all tests while that ID is only used in one. My suggestion is to move this in the test setup, in the `beforeAll`, which already sets up a lot of stuff to this test. Whether you get the ID of the existing customer or create a new one it's up to you, but let's keep it in the spec.

> I also took the opportunity to replace all wcApi instantiations with the api fixture in this test

Nice! Thank you for doing this, in time we'll update them all.

rodelgc:
Thanks @adimoldovan I had just pushed a new commit to get the id of an existing customer (instead of creating a new one, since I think that's the more straightforward way 😅 ) from within `beforeAll`. I also reverted the change where I added the `users.setup.js`. When the checks pass, it'll be ready again for another review.

adimoldovan (approved):
Thanks @rodelgc!

Issue #29822, opened 2021-05-04
WC->is_rest_api_request() showing different behavior according to the permalink settings.


mskapusuz:
**Prerequisites (mark completed items with an [x]):**
- [x] I have have carried out troubleshooting steps and I believe I have found a bug.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

**Describe the bug**
WC()->is_rest_api_request() method is returning different values according to the Wordpress Permalink setting choices.

**Expected behavior**
[is_rest_api_request() method of WooCommerce](https://github.com/woocommerce/woocommerce/blob/9347c728e9cb0bcfde201cd708a34b2d0715326f/includes/class-woocommerce.php#L301-L319) class should return TRUE for all REST requests.

**Actual behavior**
[is_rest_api_request() method of WooCommerce](https://github.com/woocommerce/woocommerce/blob/9347c728e9cb0bcfde201cd708a34b2d0715326f/includes/class-woocommerce.php#L301-L319) class is returning FALSE when WP Permalink is set as **Plain**

[is_rest_api_request() method of WooCommerce](https://github.com/woocommerce/woocommerce/blob/9347c728e9cb0bcfde201cd708a34b2d0715326f/includes/class-woocommerce.php#L301-L319) class is returning TRUE when WP Permalink is set as **different from Plain**

Depends on this, WC()->is_request( 'frontend' ) returns TRUE if permalink option selected as Plain. Thus, WC is loading frontend classes (Session etc.) on REST API requests if WP permalink option is Plain.

**Steps to reproduce the bug (We need to be able to reproduce the bug in order to fix it.)**
Steps to reproduce the bug:

I have wrote a snippet to test it.

```
add_action( 'rest_api_init', function(){
	register_rest_route(
		'test',
		'/rest-api-permalink-behavior-test/',
		array(
			'methods'             => \WP_REST_Server::READABLE,
			'callback'            => function(){ 
				return new \WP_REST_Response(
					array(
						'code'   => 'success',
						'markup' => sprintf( 'the value of is_rest_api_request is: %s', json_encode( WC()->is_rest_api_request() ) ),
					),
					200
				);
			},
			'permission_callback' => '__return_true',
			'args'                => array(
				'product_id' => array(
					'validate_callback' => 'is_numeric',
				),
			),
		)
	);
} );
```

1. As a first Place the snippet to WP.

**Test Case for Permalink: _Plain_:**
[...]

PR #54792, opened 2025-01-22
[NOT REVIEW] Test `is_rest_api_request` function

gigitux:

### Summary
While working on unit tests for [PR #54787](https://github.com/woocommerce/woocommerce/pull/54787), I discovered inconsistent behavior in `is_rest_api_request`.  

- **Single Test**: Running the `RestApiTest` in isolation returns `false` (expected behavior).  
- **Full Suite**: Running the entire test suite causes `is_rest_api_request` to return `true` (incorrect behavior).  

### Steps to Reproduce  
1. Run the single test:
   ```sh
   pnpm --filter=@woocommerce/plugin-woocommerce test:unit:env --filter='RestApiTest'

Working on unit test for https://github.com/woocommerce/woocommerce/pull/54787, I found a weird behavior with `is_rest_api_request`. 

✅  Test passes, is_rest_api_request returns false.

2. Run the full suite:

pnpm --filter=@woocommerce/plugin-woocommerce test:unit:env

❌ Test fails, is_rest_api_request returns true. (You can check the unit test job of this PR too)


### Possible Cause


I think that there is a side effect in our unit test suite that changes the behavior of this function. 

I found [this issue](https://github.com/woocommerce/woocommerce/issues/29822), but I'm not sure how much it is related. At the same time, I'm wondering if we should migrate to [WordPress Core functions ](https://core.trac.wordpress.org/ticket/42061). There is even a todo in our codebase:


https://github.com/woocommerce/woocommerce/blob/2e65e46b8faa51d22ff5a86780758379cc647de5/plugins/woocommerce/includes/class-woocommerce.php#L504-L527

### DIFF
--- a/plugins/woocommerce/tests/php/src/Internal/RestApiTest.php
+++ b/plugins/woocommerce/tests/php/src/Internal/RestApiTest.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace Automattic\WooCommerce\Tests\Internal;
+
+use Automattic\WooCommerce\Internal\RestApiParameterUtil;
+
+/**
+ * Tests for the RestApiParameterUtil class.
+ * @package Automattic\WooCommerce\Tests\Internal
+ */
+class RestApiTest extends \WC_Unit_Test_Case {
+
+	/**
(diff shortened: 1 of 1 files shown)


gigitux:
Fixed this issue with https://github.com/woocommerce/woocommerce/pull/54801

PR #54799, opened 2025-01-23
Product Gallery Thumbnails: remove settings from container block

kmanijak:

Closes # .

3. Click on the Product Gallery (Beta) block
4. **Expected:** Make sure there's **no** "Thumbnails" settings like these:
image

6. Click on the Thumbnails block
7. **Expected:** Make sure there's **no** "Thumbnails" settings like those below:
8. Try:
    - changing position of Thumbnails by changing blocks order (see video below)
    - changing Gallery Area block layout (Row -> Stack and opposite)
    - changing the number of thumbnails
    - save and go to frontend in various moments
12. Expected: Make sure "Thumbnails" are displayed either horizontally (above and below large image) or vertically (on sides), adjust to different number of thumbnails and display similarly on the frontend

Note: This PR doesn't cover ALL of the cases of Gallery Area layout settings. Block may end up in some weird state in edge cases but should behave correctly with the default Row or Stack layouts.

Before | After
-- | --
https://github.com/user-attachments/assets/3fcfe2ce-5afd-4f3c-b0bc-399618523216 | https://github.com/user-attachments/assets/8daf35a2-4c6f-4995-8296-00dc9b9f2a12

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


tjcafferkey:
I don't know if this is something to fix in this PR but in this scenario of settings (thumbnails left, 6 thumbnails) the spacing above/below the thumbnails isn't symmetrical it seems. 

Edit: Ah, I see that its because I have 5 thumbnails but my setting says show 6. I assume the code is anticipating 6 thumbnails whereas we should change this assumption to be "maximum" of 6 with the possibility of less.

Screenshot 2025-01-30 at 09 57 37

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/inner-blocks/product-gallery-thumbnails/edit.tsx
```diff
@@ -2,66 +2,54 @@
...
-	context: ProductGalleryContext;
-}
-
-export const Edit = ( { context }: EditProps ) => {
+export const Edit = ( {
+	attributes,
+	setAttributes,
+}: BlockEditProps< ProductGalleryThumbnailsBlockAttributes > ) => {
 	const blockProps = useBlockProps( {
 		className: clsx(
```
tjcafferkey: We can improve this and remove the `clsx` dependency.
```suggestion
		className: `wc-block-product-gallery-thumbnails wc-block-product-gallery-thumbnails--number-of-thumbnails-${ attributes.numberOfThumbnails }`,
```

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/style.scss
```diff
@@ -1,17 +1,9 @@
...
-$large-image: "#{$gallery}-large-image";
-$thumbnails: "#{$gallery}-thumbnails";
-$next-previous: "#{$large-image}-next-previous";
-$next-previous-left-off: "#{$next-previous}-left--off";
-$next-previous-right-off: "#{$next-previous}-right--off";
-$outside-image-offset: 30px;
-$outside-image-max-width: calc(100% - (2 * $outside-image-offset));
 $thumbnails-gap: 15px;
-$default-number-of-thumbnails: 3;
+$thumbnails-border-width: 1px;
```
tjcafferkey, line 3: Nit pick, but this variable is only used once so I'm not sure it needs to be a variable at all. I'm fine either way.

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryThumbnails.php
```diff
@@ -120,7 +120,7 @@ protected function should_display_view_all( $thumbnails_count, $product_gallery_
 	 * @return string Rendered block type output.
 	 */
 	protected function render( $attributes, $content, $block ) {
```
tjcafferkey, line 122: Do you think we should early return an empty string if there are no gallery thumbnails (or one image)? E.g.

```
if ( ! $product_gallery_images || count( $product_gallery_images ) <= 1 ) {
    return '';
}
```

tjcafferkey (approved):
Looks good and tests well! Ive just left a few comments but I wouldn't consider them blocking.

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryThumbnails.php
kmanijak: Good catch!

kmanijak:
> I don't know if this is something to fix in this PR but in this scenario of settings (thumbnails left, 6 thumbnails) the spacing above/below the thumbnails isn't symmetrical it seems.
> Edit: Ah, I see that its because I have 5 thumbnails but my setting says show 6. I assume the code is anticipating 6 thumbnails whereas we should change this assumption to be "maximum" of 6 with the possibility of less.

Good suggestion! I implemented the change in https://github.com/woocommerce/woocommerce/pull/54799/commits/a7de60ff1bd2158c50bb7296226a2828e598d4af!

I also addressed other comments, let me know @tjcafferkey if you want to take a look again or I'm good to merge.

BTW in scope of this PR I also enabled browsing all images in "small" gallery, so basically removed the artificial limit by "number of thumbnails". I think it improves UX and makes Pager showing the correct information. Example: number of thumbnails is 3 but I'm allowed to see beyond that (6/8):

image

tjcafferkey:
@kmanijak I think we're good to merge, thanks for fixing all that! I am just going to note something for discussion that we can decide to do in a follow-up PR if we like.

One thing that I keep noticing is that the alignment of thumbnails look 'off' even though technically they are not. But it is because in both the editor and the frontend the vertical thumbnails (left/right) don't sit flush with the top of the image gallery because they take the Pager block into account when calculating their position.

Do you think these would look better if the top thumbnail was aligned with the top of the Large Image block in both editor and frontend?

Screenshot 2025-01-31 at 11 58 21

Screenshot 2025-01-31 at 12 00 41

kmanijak:
> One thing that I keep noticing is that the alignment of thumbnails look 'off' even though technically they are not. But it is because in both the editor and the frontend the vertical thumbnails (left/right) don't sit flush with the top of the image gallery because they take the Pager block into account when calculating their position.
> Do you think these would look better if the top thumbnail was aligned with the top of the Large Image block in both editor and frontend?

This can be controlled by user in the Gallery Area container settings:
- by default when inserting the block I set vertical alignment to top
- after switching Row to Stack and back to Row, alignment is set to center by default (it's core's Group block behaviour). It can be brought back to top alignment manually:

https://github.com/user-attachments/assets/f34bd0bf-a8b6-4821-9eb2-a4d55ec71207

I think that aligns with overall approach to set good defaults but leave it for manual customizations later. WDYT?

tjcafferkey:
>I think that aligns with overall approach to set good defaults but leave it for manual customizations later. WDYT?

Makes sense, thanks for the explanation 👍🏻 lets merge! 🚀

PR #54803, opened 2025-01-23
Product Gallery: Remove full dialog gallery and associated code

### LINKED ISSUE #54517: Product Gallery: Remove template-part for full screen gallery
Remove the template-part part and the complexity of Product Gallery rendering itself and implement Full page gallery as in https://www.allbirds.co.nz/products/womens-tree-runner-go

tjcafferkey:

Update the full-screen/dialog gallery by:

* Removing the template part.
* Removing code depending on 'mode' from the main gallery.
* Implement a simple vertical scroll image gallery instead which cannot be customized.

Closes https://github.com/woocommerce/woocommerce/issues/54517

1. Add the new Product Gallery block to your Single Product template
2. Load the frontend of a product. Click an image to enable the dialog. Ensure all images are rendered stacked.
3. Use the close button to close the gallery.
4. Navigate to a different image in the 'small' gallery version and click it to enable the dialog again.
5. Ensure your selected image is the focused element in the dialog gallery.
6. Ensure keyboard controls work as expected
7. Test at various viewport sizes, in various browsers
8. Ensure e2e tests pass.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce-blocks/tests/e2e/tests/product-gallery/product-gallery.block_theme.spec.ts
```diff
@@ -326,15 +367,9 @@ test.describe( `${ blockData.name }`, () => {
...
-				'.wc-block-product-gallery-dialog__body'
-			);
-
-			const popUpSelectedImageId = await getVisibleLargeImageId(
-				productGalleryPopUpContent.locator(
-					`[data-block-name="woocommerce/product-gallery-large-image"]`
-				)
-			);
+			// eslint-disable-next-line playwright/no-wait-for-timeout, no-restricted-syntax
+			await page.waitForTimeout( 300 );
```
tjcafferkey, line 330: I need to somehow avoid this. The test evaluates before the side effect runs which focuses on the image element in `dialogStateChange`

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryThumbnails.php
```diff
@@ -188,7 +188,7 @@ protected function render( $attributes, $content, $block ) {
 				</div>',
 				esc_attr( $classes_and_styles['classes'] ),
 				esc_attr( $classes_and_styles['styles'] ),
-				wp_kses_post( $html ),
```
kmanijak, line 191: Why removing `wp_kses_post`? I added it in recent PR on your request 😅

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/edit.tsx
```diff
@@ -88,49 +85,13 @@ const TEMPLATE: InnerBlockTemplate[] = [
...
-	useEffect( () => {
-		const mode = getMode( currentTemplateId, templateType );
-
-		setAttributes( {
-			...attributes,
-			mode,
-			productGalleryClientId: clientId,
-		} );
-		// Move the Thumbnails block to the correct above or below the Large Image based on the thumbnailsPosition attribute.
-		moveInnerBlocksToPosition( attributes, clientId );
```
kmanijak, line 125: Removing this breaks Thumbnails positioning. You should keep it for and I'll get rid of it in a Thumbnails cleanup

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/edit.tsx
```diff
@@ -88,49 +85,13 @@ const TEMPLATE: InnerBlockTemplate[] = [
...
-		};
-	}, [] );
-
-	useEffect( () => {
-		const mode = getMode( currentTemplateId, templateType );
-
-		setAttributes( {
-			...attributes,
-			mode,
-			productGalleryClientId: clientId,
```
kmanijak, line 122: `productGalleryClientId` is still in use in Thumbnails settings

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
```diff
@@ -226,6 +171,55 @@ const productGallery = {
...
+		},
+		onDialogKeyDown: ( event: KeyboardEvent ) => {
+			if ( event.code === 'Escape' ) {
+				actions.closeDialog();
+			}
+		},
+		openDialog: () => {
+			const context = getContext();
+			context.isDialogOpen = true;
+			const triggerElement = getElement()?.ref as HTMLElement;
```
kmanijak: Nitpick:

```suggestion
			const triggerElement = getElement()?.ref;
```

seems this type casting can be also removed in other places 🤔

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
```diff
@@ -226,6 +171,55 @@ const productGallery = {
...
+				const imgContext =
+					clickedElement.getAttribute( 'data-wc-context' );
+				if ( ! imgContext ) {
+					return null;
+				}
+				const imageId = JSON.parse( imgContext ).imageId;
+				context.selectedImageNumber =
+					getImageIndex( context, imageId ) + 1;
+				selectImage( getContext(), 'current' );
+			}
```
kmanijak: As we discussed elsewhere, I kind of don't like this logic in practice.

1. IMO we _should not_ close the dialog on image click hence this logic could be removed
2. IMO we _should_ close the dialog on overlay click. If that's ok with you, you could include this in this PR or create an issue (should be fairly easy, attaching the on click directive to dialog 🤞 )

plugins/woocommerce-blocks/tests/e2e/tests/product-gallery/product-gallery.block_theme.spec.ts
```diff
@@ -81,6 +81,47 @@ const getThumbnailImageIdByNth = async (
...
+
+	if ( ( await focusedImage.count() ) > 0 ) {
+		const imageContext = ( await focusedImage.getAttribute(
+			'data-wc-context'
+		) ) as string;
+		return JSON.parse( imageContext ).imageId;
+	}
+
+	return null;
+};
```
kmanijak: I'd see this in page object in `product-gallery.page.ts`

plugins/woocommerce-blocks/tests/e2e/tests/product-gallery/product-gallery.block_theme.spec.ts
```diff
@@ -81,6 +81,47 @@ const getThumbnailImageIdByNth = async (
...
+			return null;
+		}
+		const context = element.getAttribute( 'data-wc-context' );
+		if ( ! context ) {
+			return null;
+		}
+		return JSON.parse( context ).imageId;
+	} );
+};
+
```
kmanijak: Ditto

plugins/woocommerce-blocks/tests/e2e/tests/product-gallery/product-gallery.block_theme.spec.ts
kmanijak: I see, we're evaluating JS so yeah, it doesn't wait for stuff. I think we could await for dialog to be visible before running `getActiveElementImageId`. Did you try it?

kmanijak (changes requested):
It works great! 🚀 

I left some comments with some feedback and few issues I found!

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryThumbnails.php
tjcafferkey: Oh good catch, this was leftover I think. For some reason it removes `tabindex`. Ill need to fix that while keeping it in there I think

tjcafferkey:
@kmanijak I've addressed your comments, and I tried awaiting for the dialog to be visible which also didn't work RE the e2e test timeout 😕

kmanijak (approved):
It's a great improvement!

One piece of feedback I have is regarding some constraints in image sizes in the gallery. Right now, the images are displayed in their original size with the max-width. It works great in first iteration but I think we'll have to think of some constraints to avoid high res vertical image extend drastically the screen etc. But that's separate conversation.

ziofonta:
the vertical scroll image on the desktop view is really terrible in my opinion. Seems like someone put a sequence of images in the html code without any layout presentation. I discovered you have removed the lightbox with the last woocommerce update, which has disappointed some of our customers that preferred the ligthbox, so now am I forced to make a rollback to the previous woocommerce version just because of this strange update? The "old" lightbox should be at least an option one can choose in the gallery settings...

ellenbauer:
@ziofonta Thank you for your comment. We highly appreciate your feedback, and I understand your surprise and concern about this recent change in direction for the large image view. 

While the overlay image slideshow has been a part of the WooCommerce product gallery for a long time, our new product gallery specifically aims to improve the mobile experience.

On mobile having a vertical scroll creates a more natural, fluid browsing experience without interrupting the user's flow. On larger screens reducing the number of clicks users need to make to view large product images is also an advantage.

One improvement we will make is showing all images at full width. This gives users large image views through vertical scrolling rather than an overlay slideshow.

ziofonta:
Sorry but I disagree about the mobile experience. Any photo gallery app on any mobile device browse images through left to right swipe. one gesture allows you to have the next picture isolated and centered in the screen, while the simple vertical scroll has not the same result. For me it's a downgrade, both on desktop and mobile devices.

ellenbauer:
@ziofonta I completely understand your perspective. From my point of view on mobile the gallery on the actual Product Single page is already the fullscreen gallery that can be swiped and the new scrollable overlay then gives a quick access to all images in the gallery.

ziofonta:
@ellenbauer do you mean is it still possible to use the fullscreen gallery with swipe? I only can view the enlarged pictures in the scrollable version.

PR #54844, opened 2025-01-24
Correctly render plain text emails in email preview

### LINKED ISSUE #53950: Email type (html/plain) is ignored in email preview
Email preview always shows the HTML version, even if the email type is set to plaintext. 

Make sure the email type is respected and can also be live previewed when managing a specific email.

NeosinneR:

Closes #53950.

1. Enable `Email improvements` feature in **WooCommerce > Settings > Advanced > Features**.
2. Go to **WooCommerce > Settings > Emails**.
3. Click on `Manage` any email. 
4. Change the Email type to `Plain text`.
5. Verify that the email preview correctly rendered a plain text version of the email.

### DIFF
--- a/plugins/woocommerce/includes/emails/class-wc-email.php
+++ b/plugins/woocommerce/includes/emails/class-wc-email.php
@@ -502,7 +502,19 @@ class WC_Email extends WC_Settings_API {
 	 * @return string
 	 */
 	public function get_email_type() {
-		$email_type = $this->get_transient_or_value( 'email_type', $this->email_type );
+		$email_type = $this->email_type;
+		/**
+		 * This filter is documented in templates/emails/email-styles.php
+		 *
+		 * @since 9.6.0
+		 * @param bool $is_email_preview Whether the email is being previewed.
+		 */
+		$is_email_preview = apply_filters( 'woocommerce_is_email_preview', false );
+		// Transient is used for live email preview without saving the settings.
+		if ( $is_email_preview ) {
+			$transient  = get_transient( "woocommerce_{$this->id}_email_type" );
+			$email_type = $transient ? $transient : $email_type;
+		}
 		return $email_type && class_exists( 'DOMDocument' ) ? $email_type : 'plain';
 	}
 
@@ -1245,29 +1257,5 @@ class WC_Email extends WC_Settings_API {
 		}
 
 		return $option;
-	}
-
-	/**
-	 * Return transient or provided value for email preview.
-	 *
-	 * @param string $key Option key.
-	 * @param mixed  $value Value to return when transient is not set.
-	 */
-	private function get_transient_or_value( string $key, $value ) {
-		/**
-		 * This filter is documented in templates/emails/email-styles.php
-		 *
-		 * @since 9.6.0
-		 * @param bool $is_email_preview Whether the email is being previewed.
-		 */
-		$is_email_preview = apply_filters( 'woocommerce_is_email_preview', false );
-		if ( $is_email_preview ) {
-			$email_id  = $this->id;
-			$transient = get_transient( "woocommerce_{$email_id}_{$key}" );
-			if ( false !== $transient ) {
-				return $transient ? $transient : $value;
-			}
-		}
-		return $value;
 	}
 }
--- a/plugins/woocommerce/includes/emails/class-wc-email.php
+++ b/plugins/woocommerce/includes/emails/class-wc-email.php
@@ -502,7 +502,8 @@ class WC_Email extends WC_Settings_API {
 	 * @return string
 	 */
 	public function get_email_type() {
-		return $this->email_type && class_exists( 'DOMDocument' ) ? $this->email_type : 'plain';
+		$email_type = $this->get_transient_or_value( 'email_type', $this->email_type );
+		return $email_type && class_exists( 'DOMDocument' ) ? $email_type : 'plain';
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


plugins/woocommerce/includes/emails/class-wc-email.php
```diff
@@ -1245,4 +1246,28 @@ private function get_option_or_transient( string $key, $empty_value = null ) {
...
 		return $option;
 	}
+
+	/**
+	 * Return transient or provided value for email preview.
+	 *
+	 * @param string $key Option key.
+	 * @param mixed  $value Value to return when transient is not set.
+	 */
+	private function get_transient_or_value( string $key, $value ) {
```
costasovo: The method name is quite generic, and the class seems to be related to general email, not a preview, but the method is related only to email preview. Shouldn't we reflect that in the method name? Like `get_transient_for_preview_or_value` ?

plugins/woocommerce/includes/emails/class-wc-email.php
```diff
@@ -1245,4 +1246,28 @@ private function get_option_or_transient( string $key, $empty_value = null ) {
...
+		 *
+		 * @since 9.6.0
+		 * @param bool $is_email_preview Whether the email is being previewed.
+		 */
+		$is_email_preview = apply_filters( 'woocommerce_is_email_preview', false );
+		if ( $is_email_preview ) {
+			$email_id  = $this->id;
+			$transient = get_transient( "woocommerce_{$email_id}_{$key}" );
+			if ( false !== $transient ) {
+				return $transient ? $transient : $value;
```
costasovo: I'm nitpicking here, but the message comment says, `Value to return when transient is not set.` However, if the transient contains a falsy value like an empty string or zero, the method will return the `$value` even if the transient is set. So this prevents us from storing/reading false values. The  `get_transient` function returns `false` when it is not set, so we could just check if `false !== $transient` and if not return what get_transient returned. Can't we?

plugins/woocommerce/includes/emails/class-wc-email.php
```diff
@@ -502,7 +502,8 @@ public function get_attachments() {
 	 * @return string
 	 */
 	public function get_email_type() {
-		return $this->email_type && class_exists( 'DOMDocument' ) ? $this->email_type : 'plain';
+		$email_type = $this->get_transient_or_value( 'email_type', $this->email_type );
```
costasovo: I struggle to understand why we need to check the transient. Why can't we use $this->email_type. It should be stored when a user sets the setting for the email. I tried it locally without the transient and I didn't observe any difference. I checked the preview in Settings > Emails tab below the table and also on the manage page for the particular email.

costasovo (commented):
@NeosinneR Thank you for fixing the issue. It works as expected. I just have a couple of questions to better understand the code.

plugins/woocommerce/includes/emails/class-wc-email.php
NeosinneR: Good point. I removed this method altogether, as it's not used elsewhere, and tried to describe it with an inline comment in https://github.com/woocommerce/woocommerce/pull/54844/commits/d4ac5ef29a2c8a67bcbc5ec43090d5c58fd5dc62. What do you think?

plugins/woocommerce/includes/emails/class-wc-email.php
NeosinneR: With the refactor in https://github.com/woocommerce/woocommerce/pull/54844/commits/d4ac5ef29a2c8a67bcbc5ec43090d5c58fd5dc62, I also updated the comment. In this case, the value can't be falsy - the code was correct, the comment was misleading.

plugins/woocommerce/includes/emails/class-wc-email.php
NeosinneR: `$this->email_type` comes from database (`$this->get_option( 'email_type' )`). The transient is used for live preview when changing settings, but without saving them to database.

You can check this when managing a specific email. Without this change, when you change the email type from HTML to plain text, the live preview wouldn't change. With this change, you will see the change from HTML to plain text in live preview, without needing to save new settings.

costasovo: @NeosinneR Thank you for the explanation! In my previous testing, I was checking after I saved the value.

I tried the live preview. It works, but I accidentally ran into an issue. It may be an edge case, but perhaps you may find a simple solution. 
If you change the email type, but don't save the change and return to the email listing page, the preview at the bottom of the page will show an incorrect preview for that email. I think it is because the transient keeps the edited value.

### Replication steps
1) Go to Woo > Settings > Emails and click Manage for a random email
2) Switch email type to the other value
3) Return to Woo > Settings > Emails and confirm not saving changes
4) Switch the preview at the bottom to the email you've just edited and observe that the rendered preview doesn't match the saved email type.

Screenshot 2025-01-28 at 10 55 11

As it is an edge case, we could also consider not addressing it in this PR and fixing it later.
Also, opening the manage page for that email again fixes the issue.

NeosinneR: @costasovo previously, the manage email page would remove general email settings transients and that specific email transient. But the email styles setting page didn't remove specific email transients on page load, which caused this issue. 

In https://github.com/woocommerce/woocommerce/pull/54844/commits/bad2bbef6bfbb92a48cffe43a4487d4c5ce609ee I changed it, so all email settings transients are always removed on page load.

costasovo (approved):
@NeosinneR Great job! After the last change all works as expected.
There is a failing test. It seems unrelated I reran it and it failed again but for a different reason. Tryin one more time...

NeosinneR:
@costasovo This might be an issue from the `trunk` mentioned in p1738072709191129-slack-C03CPM3UXDJ but it's already merged. I'll handle it from here (will try rebase), and merge it once tests pass (rebasing won't discard your approval).

PR #54861, opened 2025-01-25
WC_Order: Ensure status is a string in the setter

### LINKED ISSUE #54855: [Fatal] PHP Fatal error:  Uncaught TypeError: Automattic\WooCommerce\Utilities\OrderUtil::remove_status_prefix(): Argument #1 ($status) must be of type string, array given, called in | OrderUtil.php:239
### Prerequisites

### Describe the bug

Issue witnessed in 9.5.2

stack trace

```
PHP Fatal error:  Uncaught TypeError: Automattic\WooCommerce\Utilities\OrderUtil::remove_status_prefix(): Argument #1 ($status) must be of type string, array given, called in /../plugins/woocommerce/9.5.2/includes/abstracts/abstract-wc-order.php on line 645 and defined in /../plugins/woocommerce/9.5.2/src/Utilities/OrderUtil.php:239
Stack trace:
#0 /../plugins/woocommerce/9.5.2/includes/abstracts/abstract-wc-order.php(645): Automattic\WooCommerce\Utilities\OrderUtil::remove_status_prefix(Array)
#1 /../plugins/woocommerce/9.5.2/includes/class-wc-order.php(304): WC_Abstract_Order->set_status(Array)
#2 /../plugins/woocommerce/9.5.2/includes/class-wc-order.php(390): WC_Order->set_status(Array, '', false)
#3 /srv/htdocs/wp-content/plugins/tinkoff-woocommerce/tinkoff/success.php(63): WC_Order->update_status(Array)
#4 {main}
  thrown in /../plugins/woocommerce/9.5.2/src/Utilities/OrderUtil.php on line 239
```

### Expected behavior

No fatals witnessed

### Actual behavior

Fatals witnessed

### Steps to reproduce

We don’t have steps to reproduce, but 71 Fatals of this type were observed on Jan 17 Nov 29, 2024 across 1 sites.
Hopefully, someone can dig in and apply a defensive check if appropriate. – To be updated after investigation

### WordPress Environment

N/A

### Isolating the problem

coreymckrill:

This simply ensures type safety when setting an order status, before using the `OrderUtil::remove_status_prefix`, which requires a string for its input. Necessary when using more modern code within older parts of the codebase that were written before type hinting was widely available in PHP.

This also updates the linebreaks and whitespace in two complex conditionals within the same method, for better readability.

Fixes #54855

1. You can replicate the reported issue using `wp shell` before checking out this branch:
    ```
    wp shell
    $o = wc_create_order()
    $o->update_status( array( 'asdf' ) )
    ```
2. This should give you a fatal error:
    `Uncaught TypeError: Automattic\WooCommerce\Utilities\OrderUtil::remove_status_prefix()`
4. Now check out this branch and repeat the commands. This time, you should get a PHP warning instead:
    `Warning: Array to string conversion in /var/www/html/wp-content/plugins/woocommerce/includes/abstracts/abstract-wc-order.php on line 653`

I debated whether to allow the PHP warning, and decided it was ok so that extension developers still get some kind of feedback when their code is doing something wrong. I could be convinced otherwise, though.

### DIFF
--- a/plugins/woocommerce/changelog/fix-54855-update-status-not-string
+++ b/plugins/woocommerce/changelog/fix-54855-update-status-not-string
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Avoid a fatal error when trying to set an order status and the given status is not a valid type
--- a/plugins/woocommerce/includes/abstracts/abstract-wc-order.php
+++ b/plugins/woocommerce/includes/abstracts/abstract-wc-order.php
@@ -650,19 +650,31 @@ abstract class WC_Abstract_Order extends WC_Abstract_Legacy_Order {
 	 */
 	public function set_status( $new_status ) {
 		$old_status = $this->get_status();
-		$new_status = OrderUtil::remove_status_prefix( $new_status );
+		$new_status = OrderUtil::remove_status_prefix( (string) $new_status );
 
 		$status_exceptions = array( OrderStatus::AUTO_DRAFT, OrderStatus::TRASH );
 
 		// If setting the status, ensure it's set to a valid status.
 		if ( true === $this->object_read ) {
 			// Only allow valid new status.
-			if ( ! in_array( 'wc-' . $new_status, $this->get_valid_statuses(), true ) && ! in_array( $new_status, $status_exceptions, true ) ) {
+			if (
+				! in_array( 'wc-' . $new_status, $this->get_valid_statuses(), true )
+				&& ! in_array( $new_status, $status_exceptions, true )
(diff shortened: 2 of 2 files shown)


naman03malhotra (approved):
Nicely done! I am receiving a notice with this fix as to fatal in `trunk`.
Before:
image

After:
image

coreymckrill:
Yeah, the notice is expected (or warning, depending on what version of PHP you are running). I think it's ok, it gives developers feedback, and it's obviously better than the fatal 😅

PR #54874, opened 2025-01-27
[settings] Enhance edits components with dynamic label and help text handling

### LINKED ISSUE #54826: [Settings] Improve Text Field Handling
## Description

https://github.com/woocommerce/woocommerce/pull/54041#issuecomment-2611216328

The current implementation for rendering settings fields has an issue where text fields without a desc property default to displaying the field's id as the label. We need to respect the placeholder property, if it is defined, for better usability.

Determine how to appropriately handle the desc_tip property to provide helpful tooltips or inline descriptions.
Improving this behavior will enhance the user experience and align with the goals of the DataForms integration for consistent and user-friendly rendering of settings fields.

## Acceptance Criteria
- When a text field lacks a `desc` property:
  - Don't render label
  - If a `placeholder` property is provided, it is respected and displayed appropriately.
- Define and implement a solution for handling the `desc_tip` property, ensuring tooltips or inline descriptions are rendered effectively.
- Ensure the changes are backward-compatible and do not introduce regressions in other field types.

chihsuan:

Closes https://github.com/woocommerce/woocommerce/issues/54826

This PR improves the rendering of settings fields, addressing issues related to labels, placeholders, and descriptions. It ensures text fields without a desc property are handled appropriately, respects the placeholder property, and shows the help text.

In classic settings, we render the help text as a tooltip. It's possible to do the same thing by modifying the edit components. However, since we want to reuse the upstream components in the future, at this moment, I don't want to customize the edit components too much. So I just pass the help text to the `help` prop to `*Control` components`. If we get the design requirement to render the help text as a tooltip, we can customize the edit components then but should also request the upstream components to support that.

Changes:

- Refactor checkbox, input, and select edit components to support optional help text
- Add utility function to determine label and help text based on setting configuration
- Update transformers to handle label and help text for various setting types

2. Turn on `settings` feature flag
3. `WooCommerce > Settings`
4. See that text inputs, checkboxes, and select inputs have labels and help text
5. Go to `Emails > Email sender options`
6. See `"From" name` and `"From" address` input fields don't have unexpected id labels 
7. See `Header image` input has placeholder text `N\A`

Screenshot 2025-01-27 at 17 13 21

Screenshot 2025-01-27 at 17 12 50

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/packages/js/settings-editor/src/components/checkbox-edit/index.tsx
+++ b/packages/js/settings-editor/src/components/checkbox-edit/index.tsx
@@ -10,25 +10,23 @@ import type { DataFormControlProps } from '@wordpress/dataviews';
  */
 import type { DataFormItem } from '../../types';
 
-export const CheckboxEdit = ( {
-	field,
-	onChange,
-	data,
-}: DataFormControlProps< DataFormItem > ) => {
-	const { id, getValue, label } = field;
-	const value = getValue( { item: data } );
-
-	return (
-		<CheckboxControl
-			__nextHasNoMarginBottom={ true }
-			id={ id }
-			label={ label }
-			checked={ value === 'yes' }
-			onChange={ ( checked ) => {
-				onChange( {
-					[ id ]: checked ? 'yes' : 'no',
-				} );
-			} }
-		/>
-	);
-};
+export const getCheckboxEdit =
+	( help?: React.ReactNode ) =>
+	( { field, onChange, data }: DataFormControlProps< DataFormItem > ) => {
+		const { id, getValue, label } = field;
+		const value = getValue( { item: data } );
+		return (
+			<CheckboxControl
+				__nextHasNoMarginBottom={ true }
+				id={ id }
+				label={ label }
+				help={ help }
+				checked={ value === 'yes' }
+				onChange={ ( checked ) => {
+					onChange( {
+						[ id ]: checked ? 'yes' : 'no',
+					} );
+				} }
+			/>
+		);
+	};
--- a/packages/js/settings-editor/src/components/checkbox-edit/index.tsx
+++ b/packages/js/settings-editor/src/components/checkbox-edit/index.tsx
@@ -13,7 +13,12 @@ import type { DataFormItem } from '../../types';
 export const getCheckboxEdit =
 	( help?: React.ReactNode ) =>
 	( { field, onChange, data }: DataFormControlProps< DataFormItem > ) => {
-		const { id, getValue, label } = field;
+		const { id, getValue } = field;
+
+		// DataForm will automatically use the id as the label if no label is provided so we conditionally set the label to undefined if it matches the id to avoid displaying it.
+		// We should contribute upstream to allow label to be optional.
+		const label = field.label === id ? undefined : field.label;
+
 		const value = getValue( { item: data } );
 		return (
 			<CheckboxControl
--- a/packages/js/settings-editor/src/hooks/utils/transformers.tsx
+++ b/packages/js/settings-editor/src/hooks/utils/transformers.tsx
@@ -24,6 +24,12 @@ export type DataItem = Record< string, BaseSettingsField[ 'value' ] >;
  *
  * @param setting The setting object containing description and tip information
  * @return Object with label and help text
+ *
+ * Cases:
+ * - desc_tip === true: description becomes help text, empty label
+ * - desc_tip is string: string becomes help text, description becomes label
+ * - desc_tip === false: empty help text, description becomes label
+ * - desc_tip undefined: empty help text, description becomes label
  */
 export const getLabelAndHelp = (
 	setting: BaseSettingsField | CheckboxSettingsField
--- a/packages/js/settings-editor/src/hooks/utils/transformers.tsx
+++ b/packages/js/settings-editor/src/hooks/utils/transformers.tsx
@@ -20,15 +20,26 @@ import { getSelectEdit } from '../../components/selectEdit';
 export type DataItem = Record< string, BaseSettingsField[ 'value' ] >;
 
 /**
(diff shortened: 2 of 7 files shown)


psealock (commented):
The `label` prop is now rendering the `id`. There is something internal to `*Control` components that use the `id` as the label when the label isn't present.

### Trunk
image

### This branch
image

I think the boolean is supplied as a string, so something like this works. Alternatively, we should `Boolean()` the value to avoid this.

```js
const helpText =
		description === 'true' || description === 'false' ? '' : description;
```

chihsuan:
> The `label` prop is now rendering the `id`. There is something internal to `*Control` components that use the `id` as the label when the label isn't present.
> 
> ### Trunk
> image
> ### This branch
> image
> I think the boolean is supplied as a string, so something like this works. Alternatively, we should `Boolean()` the value to avoid this.
> 

@psealock This is weird. I'm not seeing the same behavior on my end. The following is what I'm seeing on this branch:

Screenshot 2025-01-28 at 11 55 31


It should already be handled here:

https://github.com/woocommerce/woocommerce/blob/41ed7cad2cf6ca31c464be5b5d4100f07161dfc6/packages/js/settings-editor/src/components/inputEdit/index.tsx#L19-L22

Could you check if you're still seeing the issue? I can see that select controls are still using the `id` as the label though. I'll fix that.

packages/js/settings-editor/src/components/checkbox-edit/index.tsx
```diff
@@ -10,25 +10,28 @@ import type { DataFormControlProps } from '@wordpress/dataviews';
...
 
-export const CheckboxEdit = ( {
-	field,
-	onChange,
-	data,
-}: DataFormControlProps< DataFormItem > ) => {
-	const { id, getValue, label } = field;
-	const value = getValue( { item: data } );
+export const getCheckboxEdit =
+	( help?: React.ReactNode ) =>
```
ilyasfoo, line 14: A few questions:

1. I'm just curious, is this shape an established pattern you're following?
2. Instead of committing to positional arguments, would it be better if we use named parameter instead? I think it's better for the long term in case we expect the list of parameters to grow.

ilyasfoo (commented):
@chihsuan Testing well, but a few questions

packages/js/settings-editor/src/components/checkbox-edit/index.tsx
chihsuan: > 1. I'm just curious, is this shape an established pattern you're following?

Nope,  this is a temporary workaround. Ideally, we should not have to do this or even create our own Edits components but retrieve the required data from the fields, however, Gutenberg doesn't support this. We should contribute upstream to make this better.

> Instead of committing to positional arguments, would it be better if we use named parameter instead? I think it's better for the long term in case we expect the list of parameters to grow.

I don't really expect the list of parameters to grow. I think we can use positional arguments first and then convert to named parameters when the number of arguments grows. Does that sound good?

ilyasfoo: Sounds good!

ilyasfoo (approved):
Thanks, LGTM

packages/js/settings-editor/src/hooks/utils/transformers.tsx
```diff
@@ -11,12 +11,25 @@ import type { Field, FormField } from '@wordpress/dataviews';
...
-import { SelectEdit } from '../../components/selectEdit';
+import { getSelectEdit } from '../../components/selectEdit';
 
 export type DataItem = Record< string, BaseSettingsField[ 'value' ] >;
 
+/**
+ * Helper function to determine label and help text from setting
+ * If desc_tip is true, we should display the desc as the help text.
+ */
+const getLabelAndHelp = (
```
psealock: 👍🏽 For reusable function - text areas and surely a few others will make use as well

packages/js/settings-editor/src/hooks/utils/transformers.tsx
```diff
@@ -11,12 +11,25 @@ import type { Field, FormField } from '@wordpress/dataviews';
...
+	setting: BaseSettingsField | CheckboxSettingsField
+) => {
+	return {
+		label: setting.desc_tip === true ? '' : setting.desc,
+		help: setting.desc_tip === true ? setting.desc : setting.desc_tip,
+	};
+};
```
psealock, line 50: To make this function reusable it looks like the description can be `settings.desc || setting.description`.

https://github.com/woocommerce/woocommerce/blob/d05829296d382c8251db90ee0603b86cb9b295a9/plugins/woocommerce/includes/gateways/paypal/includes/settings-paypal.php#L122-L129

Also, should we explicitly handle a `false` value for `desc_tip`? I haven't found an example in our codebase besides some defaults, but it could exist out there.

psealock (commented):
Thanks @chihsuan, this is working as expected. I'm not sure what happened. Just a few comments for improving `getLabelAndHelp`.

image

packages/js/settings-editor/src/hooks/utils/transformers.tsx
chihsuan: Great suggestions. I've improved the function and added test cases for it. cf9e65c94b6bc1ab05fb852736a98e1a25f36af9

psealock (approved):
Thanks for the changes, looking great 🚢

psealock:
@chihsuan if you don't mind, I will merge this now with the updates. I'd like to use `getLabelAndHelp` in the textarea implementation as well as the upcoming radio and multiselect.

PR #54890, opened 2025-01-27
[Experimental] Render Add to Cart with Options template parts in the template editor

### LINKED ISSUE #54417: Blockified Add to Cart with Options block: make it possible to edit different layouts per product type
It should be possible to save different layouts per product type, probably following the _Synced patterns_ approach proposed [here](https://github.com/woocommerce/woocommerce/discussions/50484). This should be done in a way that we load the PHP template for 3rd-party product types.

Out of scope:

* Provide a way for extensions registering custom product types to update to the blockified approach (ie: registering their own synced pattern).

---

Requirements:
* https://github.com/woocommerce/woocommerce/issues/54431
* https://github.com/woocommerce/woocommerce/issues/54432
* https://github.com/woocommerce/woocommerce/issues/54433
* https://github.com/woocommerce/woocommerce/issues/54434

### LINKED ISSUE #54432: Blockified Add to Cart with Options block: render synced patterns in the editor
* https://github.com/woocommerce/woocommerce/issues/54431

We want to make it possible to edit the Add to Cart synced patterns specific to each product type directly from the Single Product template (and the Single Product block). In order to do so, we need to update the editor side of the _Add to Cart with Options_ block to render the synced patterns content. It also needs to allow edits and, on save, save the synced patterns as expected.

Aljullu:

Closes #54417.
Closes #54432.

This PR makes it so the Add to Cart with Options block renders the template parts in the editor, so they can be modified by users.

[Enregistrament de pantalla del 2025-01-24 18-51-29.webm](https://github.com/user-attachments/assets/8222ed38-a044-4ff3-b064-b26d373eb858)

1. Install and activate the WooCommerce Beta Tester plugin
2. Go to _WooCommerce Admin Test Helper_ > _Features_ and enable `experimental-blocks` and `blockified-add-to-cart`.
3. Go to _Appearance_ > _Editor_ > _Templates_ > _Single Product_, select the _Add to Cart with Options_ block and update it to the blockified one.
4. Switch between product types making some edits (ie: in Simple and Variable product types).
5. Save and go to the frontend. Verify the edits took place (you will need to visit products of those types).
6. Add a file to your theme under `parts/external-product-add-to-cart-with-options.html`. Write any content there. For example:

```HTML
<!-- wp:paragraph -->
<p>This is a template from the theme.</p>
<!-- /wp:paragraph -->
<!-- wp:woocommerce/product-button {"textAlign":"center","fontSize":"small"} -->
<div class="wp-block-woocommerce-product-button is-loading has-small-font-size"></div>
<!-- /wp:woocommerce/product-button -->
```

7. Go to _Appearance_ > _Editor_ > _Templates_ > _Single Product_ again.
8. Verify when you select the _External_ product type, the template from the theme is loaded, so the `This is a template from the theme.` paragraph block is visible.
9. Make some edits to the external add to cart form.
10. Save and verify they are applied in the frontend.


plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/edit/edit-template-part.tsx
```diff
@@ -0,0 +1,129 @@
...
+export const AddToCartWithOptionsEditTemplatePart = ( {
+	productType,
+}: {
+	productType: string;
+} ) => {
+	const currentTheme = useSelect(
+		( select ) => select( coreStore ).getCurrentTheme()?.stylesheet,
+		[]
+	);
+	const templatePartSlug = `${ productType }-product-add-to-cart-with-options`;
```
mdperez86: If these variables `templatePartSlug`, `themeTemplatePartId` and `wooCommerceTemplatePartId` are not going to be used elsewhere they could be moved within the `useSelect` callback function and use `productType` and `currentTheme` as deps, so while we scope them to the callback block we are also more clear about what are the real deps of the `useSelect` hook.

plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/edit/edit-template-part.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { useSelect } from '@wordpress/data';
+import { Spinner } from '@wordpress/components';
+import { useEntityBlockEditor, store as coreStore } from '@wordpress/core-data';
+import {
+	InnerBlocks,
+	useInnerBlocksProps,
+	useBlockProps,
+} from '@wordpress/block-editor';
+
+export type FeaturesKeys = 'isBlockifiedAddToCart';
```
mdperez86: This could be moved to a conts file as it already exists in `plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/edit/index.tsx`

mdperez86 (approved):
This is looking great and it works as described in the testing instructions.

I just left some comments that are not a blocker to this PR.
Nice job @Aljullu !!!

plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/edit/edit-template-part.tsx
Aljullu: Ups, yes, actually this wasn't used anywhere. I removed it in 977ab97498b0965ebb69cc0e71862359d8c669c3.

plugins/woocommerce-blocks/assets/js/blocks/add-to-cart-with-options/edit/edit-template-part.tsx
Aljullu: Yeah, good point. I moved these variables inside the `useSelect` hook in 02c0b65ea30ec2d9a395af481c2360539ed8b123.

Issue #46461, opened 2024-04-11
Prevent 'BlockTemplatesController::add_archive_product_to_eligible_for_fallback_templates' to leak into classic themes


Aljullu:
There is a method in `BlockTemplatesController` called `add_archive_product_to_eligible_for_fallback_templates` which seems to be leaking into classic themes. From my investigation, it doesn't have any impact in functionality, but we should make it so it only runs in block themes to avoid running unnecessary code and to prevent unexpected bugs.

The method can be found here:

https://github.com/woocommerce/woocommerce/blob/de54195c3ab3a091c3c406984622b61a6f8fa413/plugins/woocommerce/src/Blocks/BlockTemplatesController.php#L191-L215

PR #54906, opened 2025-01-28
Improve coming soon store page matching

### LINKED ISSUE #53129: Improve coming soon store page matching
### Issue description

There are still mismatches with coming soon store page detection. I think this is due to the combination of possible configuration by both WordPress and WooCommerce page setups.

If it's possible, we should improve matching by minimize the possible combination by either normalizing configuration (i.e: find out a way to determine the WordPress' mapping to pages, regardless of permalink settings), or by simply adding more test cases in order to catch edge cases.

For example, the cart page can viewed with the following:

#### To reproduce:

1. Go to `WooCommerce > Settings > Site visibility`
2. Set site to `coming soon` and `apply to store pages only`
3. Go to `Settings > Permalinks`
5. Set permalink settings to `Plain`
6. In an incognito browser, go to cart page
7. Observe it's showing coming soon
8. Add `&tag=test` to the URL
9. Observe it's showing the cart page

Additionally, @moon0326 was able to change permalink settings in tests using the following snippet. I think it'll be good to test store page matching under different permalink settings, at a minimum, 1 with query-based routing (plain), and 1 with path-based routing such as post name.

```
update_option('permalink_structure', 'structure goes here'):
$wp_rewrite->init();
```

### Acceptance criteria

1. Fix `cart` store page matching
1. Strengthen test cases by adding more possible combination of WP permalink settings and WooCommerce page setup

chihsuan:

Closes https://github.com/woocommerce/woocommerce/issues/53129

This PR is to improve store page detection logic. It's challenging to use url matching to detect store pages. I think it's better to use native WordPress methods (e.g. `is_page()`, `is_post_type_archive()`, `is_singular()`, etc.) to detect store pages, which are more reliable and easier to maintain, so I introduced a new method `is_current_page_store_page()` to determine if the current page is a store page.

The only downside is that we can only use this method when WP has set up the query, typically during or after the `parse_query` or `template_redirect` action hooks. This causes an issue when we need to show the coming soon page in non-FSE themes. In [this PR](https://github.com/woocommerce/woocommerce/pull/54624), I fixed the "Coming Soon Social Links" template part for classic themes by moving the initialization of the Block templates to the `after_setup_theme` hook. The initialization is only needed when we need to show the coming soon page in non-FSE themes, but since we can't use the `is_current_page_store_page()` method in `after_setup_theme`, I changed it to initialize the class when coming soon mode is active, regardless of the page. I think this's a good compromise since it [doesn't cause any impact on functionality when the block template](https://github.com/woocommerce/woocommerce/issues/46461) is initialized for non-coming soon pages.

Changes:

- Add `is_current_page_store_page()` method, which use WP native methods to detect store pages.
- Deprecate the `is_store_page()` method with a notice
- Add test cases for the new method

**1. Basic Store Page Detection**
1. Create a new JN site with this branch.
2. Install and activate `WooCommerce Smooth Generator` plugin.
3. Generate some products using `WooCommerce Smooth Generator` plugin.
4. Navigate to `WooCommerce -> Settings -> Site Visibility` and choose `Coming soon with Restrict to store pages only`.
5. Open the site in incognito mode and navigate to any store page (Shop, Cart, Checkout, Product, Category, etc)
6. Confirm that the coming soon page is shown.

**2. URL Pattern Collision Prevention**
1. Create a new page from `Pages -> Add a new page`
2. Set the page title to `Shopping URL` and confirm the permalink is set to `shopping-url`.
3. Open the page in an incognito session.
4. Confirm that the coming soon page is NOT shown.
5. Set the link to `products-demo` and save the page
6. Open the page in an incognito session.
7. Confirm that the coming soon page is NOT shown.

**3. Plain Permalink Structure with Query Parameters**
1. Go to `Settings > Permalinks`
2. Set permalink settings to `Plain`
3. In an incognito browser, go to Cart page
4. Confirm that the coming soon page is shown.
5. Add `&tag=test` to the URL and verify coming soon page still shows

**4. Shop Page Detection with Plain Permalinks**
1. Open an incognito session and navigate to the frontend.
2. Click `Shop` from the menu.
3. Confirm the coming soon page is rendered.

**5. Unassigned Shop Page Handling**
1. Go to /wp-admin/admin.php?page=wc-settings&tab=products
2. Unassign the shop page and save changes
3. Open a incognito window and go to the /shop page
4. Verify that the coming soon page is displayed

***

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Admin/WCAdminHelper.php
```diff
@@ -181,116 +148,35 @@ public static function is_store_page( $url = '' ) {
...
-						}
-					}
-				}
-			}
+		if (
+			// Check taxonomies.
+			is_tax( 'product_tag' ) ||
+			is_tax( 'product_cat' ) ||
+			is_tax( 'product_brand' ) ||
+			is_tax( 'product_variation' )
```
ilyasfoo: Could we use [`is_product_taxonomy`](https://github.com/woocommerce/woocommerce/blob/5d7d14cd7ab7c37a0834d7f1b4192bf2c8000b97/plugins/woocommerce/includes/wc-conditional-functions.php#L47) here instead? I think we can leverage this as it should also capture custom product taxonomies by extensions. It gives the following taxonomies in a fresh WC install:

```
wp> var_dump(get_object_taxonomies( 'product' ));
array(6) {
  [0]=>
  string(13) "product_brand"
  [1]=>
  string(12) "product_type"
  [2]=>
  string(18) "product_visibility"
  [3]=>
  string(11) "product_cat"
  [4]=>
  string(11) "product_tag"
  [5]=>
  string(22) "product_shipping_class"
}
```

However, I'm unsure why `product_variation` isnt listed. As far as I can tell, there's no product page that uses `product_variation` taxonomy. Let me know what you think!

plugins/woocommerce/src/Admin/WCAdminHelper.php
```diff
@@ -322,4 +220,21 @@ private static function get_normalized_url_path( $url ) {
...
+	 *
+	 * @internal
+	 * @link https://wordpress.stackexchange.com/a/274572
+	 * @param \WP $wp WordPress environment instance.
+	 */
+	private static function get_url_from_wp( \WP $wp ) {
+		// Initialize query vars if they haven't been set.
+		if ( empty( $wp->query_vars ) || empty( $wp->request ) ) {
+			$wp->parse_request();
+
```
ilyasfoo: Nit
```suggestion
```

plugins/woocommerce/src/Internal/ComingSoon/ComingSoonHelper.php
```diff
@@ -30,11 +30,9 @@ public function is_store_coming_soon(): bool {
 	}
 
 	/**
-	 * Returns true when the provided URL is behind a coming soon screen.
-	 *
-	 * @param string $url The URL to check.
+	 * Return true if the current page should be shown in coming soon mode.
 	 */
-	public function is_url_coming_soon( string $url ): bool {
```
ilyasfoo, line 37: `is_url_coming_soon` was previously a public function, do we need to deprecate it? I highly doubt it was used by 3pd though, no results from "all-plugins" repository

plugins/woocommerce/src/Internal/ComingSoon/ComingSoonRequestHandler.php
```diff
@@ -36,28 +36,30 @@ class ComingSoonRequestHandler {
 	 */
 	final public function init( ComingSoonHelper $coming_soon_helper ) {
 		$this->coming_soon_helper = $coming_soon_helper;
+		// Skip if the site is live.
+		if ( $this->coming_soon_helper->is_site_live() ) {
```
ilyasfoo, line 40: Nice optimization!

ilyasfoo (commented):
Amazing work here, @chihsuan, thank you!

moon0326 (approved):
Thank you so much for this PR! Looks much cleaner and it's easier to understand.

Tested various scenarios including the previous bugs with shop page detection. Everything worked as expected 👍 🚀

plugins/woocommerce/src/Internal/ComingSoon/ComingSoonHelper.php
chihsuan: Since it's under Internal namespace, I think we don't need to deprecate it. Thanks for checking!

plugins/woocommerce/src/Admin/WCAdminHelper.php
chihsuan: > Could we use [is_product_taxonomy](https://github.com/woocommerce/woocommerce/blob/5d7d14cd7ab7c37a0834d7f1b4192bf2c8000b97/plugins/woocommerce/includes/wc-conditional-functions.php#L47) here instead? I think we can leverage this as it should also capture custom product taxonomies by extensions. It gives the following taxonomies in a fresh WC install:

Yeah, that's a great suggestion! I've updated the PR to use is_product_taxonomy. a04b4eda6baeed7da572d87eca52d308ac8391be

> However, I'm unsure why product_variation isnt listed. As far as I can tell, there's no product page that uses product_variation taxonomy. Let me know what you think!

Looked into it, and product_variation is indeed not a taxonomy. I've removed it from the list.

ilyasfoo (approved):
Nice, thanks! LGTM 🚢

PR #54945, opened 2025-01-28
Enhance stalebot workflow to handle stale PRs and draft status changes

### LINKED ISSUE #54169: [Monorepo] Remove stale PR approvals automatically
Ref: p1735923523653149-slack-C03CPM3UXDJ

We made a conscious decision in the monorepo to turn off the requirement for re-approval of PRs after changes are introduced.  The rationale was that it greatly slowed down the developer experience if a PR would have to be re-reviewed/approved even when making a small change (like resolving a pnpm-lock conflict). 

There are a few edge cases where things could be tightened up, however.  

- Occasionally a PR will get opened, reviewed/approved and then it won't get merged immediately. In that case, approval should probably get auto-revoked after [30 days?] and a comment posted.  Approval can be re-requested and the PR can be merged then.
- A PR is moved back to draft after approval is obtained.  Perhaps additional bugs are addressed, or the scope of the feature is changed.  Currently, if a PR changes to draft and then back to ready for review, the approval is preserved. We should auto-remove it.

I think these cases could be addressed with the Stalebot -- run periodically and un-approve PRs that are either changed to draft, or that have not been merged and 30 days has elapsed since approval was obtained.

lanej0:
- Add support for removing stale PR approvals after 30 days
- Implement workflow to remove approvals when a PR is converted to draft
- Update workflow permissions and conditions
- Modify workflow name to reflect expanded functionality

This PR implements changes in the stalebot action that will remove the approval from pull requests:
- When a PR is approved and then moved back to draft status, the approval is removed.  This should happen instantly when converted to draft.
- When an approved PR is older than 30 days (the approval is older than 30 days, not the creation of the PR), the approval is removed.

Closes https://github.com/woocommerce/woocommerce/issues/54169

1. For the 30 day stale check, you can trigger this action manually by going to Actions, selecting the Process stale issues and PRs workflow and clicking Run workflow (set this branch as the target) (it will process PRs that match the criteria).  You should see `Dismissing stale approval on PR #${pr.number} from ${review.user.login}` for each PR. -- I only let it close a half dozen PRs in my test (see comment below).
2. For PRs being changed to draft, create a test PR and get it approved.  Convert it to draft after approval.  You should see the action run and the approval removed. This might only be testable post-merge.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Added stalebot workflows to remove approvals from PRs that are draft or over 30 days old.

</details>

<details>

#### Comment 

</details>


lanej0:
I did a partial run of the 30 day checker here: https://github.com/woocommerce/woocommerce/actions/runs/13021477955/job/36322864119

There was one bug that cropped up because of that, so one PR had it's approval erroneously removed.  But the others are all good!

Stojdza (commented):
✅ The code looks good to me but I would appreciate another 👀 from someone else, since I'm not fully familiar with stalebot. 

Looking at the PRs proccessed [here](https://github.com/woocommerce/woocommerce/actions/runs/13021477955/job/36322864119), I noticed PR https://github.com/woocommerce/woocommerce/pull/53060 that was approved last week, but it was in the draft status at that point. Stalebot removed approval, since it was also approved ~2 months ago. This is an edge case, but I wonder what should we do with PRs that are approved while in the draft status? 

Possible scenario:
- PR in the draft status
- After some time PR is approved
- After some time PR marked as Ready for review

Depending on when was the approval, stalebot may remove it. But before stalebot removes the approval, there's is a chance to merge approved PR?

lanej0:
> Looking at the PRs proccessed [here](https://github.com/woocommerce/woocommerce/actions/runs/13021477955/job/36322864119), I noticed PR https://github.com/woocommerce/woocommerce/pull/53060 that was approved last week, but it was in the draft status at that point. Stalebot removed approval, since it was also approved ~2 months ago. This is an edge case, but I wonder what should we do with PRs that are approved while in the draft status?

Good catch @Stojdza -- it was using the approval date from the first approval (because that's what the API returned first).  I've fixed that now so that it's looking at the latest approval.  That's what we want, because this is just a safeguard measure to ensure that PRs that haven't been looked at in >30 days aren't getting merged.

.github/workflows/stalebot.yml
```diff
@@ -50,3 +53,88 @@ jobs:
...
+                          }
+                        }
+                        
+                        // Process only if we found an approval
+                        if (latestApproval) {
+                          console.log(`PR #${pr.number} latest approval from ${latestApprovalReview.user.login} on ${latestApproval.toISOString()}`);
+                          
+                          if (latestApproval < thirtyDaysAgo) {
+                            console.log(`Dismissing stale approval on PR #${pr.number} from ${latestApprovalReview.user.login} (approved on ${latestApproval.toISOString()})`);
+                            await github.rest.pulls.dismissReview({
```
kalessil: Looks like we'll dismiss only the latest approval here. WDYTH about dismissing all approvals similar to `remove-draft-approvals`?

lanej0: That was something I hadn't considered, thanks for bringing that up @kalessil.  The latest changes should dismiss all approvals on a PR if the latest approval is more than 30 days old.

lanej0:
> [!IMPORTANT] 
> This is an AI-generated review by https://github.com/Automattic/pr-review-agent/ 

Hey there! I've taken a look at the changes to the stalebot workflow, and I have a few thoughts that might help improve it:

1. The new job `remove-draft-approvals` seems to have incorrect YAML indentation. It's currently at the same level as the `stale` job, but it's indented as if it's part of the previous step. This will cause the workflow to fail. The job should be aligned with the `stale` job, like this:

```yaml:.github/workflows/stalebot.yml
jobs:
    stale:
        # ... existing job ...

    remove-draft-approvals:  # <- Fix indentation here
        runs-on: ubuntu-20.04
        # ... rest of the job ...
```

2. For the stale PR approvals removal script, consider adding some rate limiting or batching. The current implementation makes API calls in tight loops (one call per review dismissal) which could hit GitHub API rate limits on repositories with many PRs. You might want to:
   - Add a small delay between dismissals
   - Or batch the dismissals into groups
   - Or add error handling for rate limit responses

3. The workflow now has two distinct responsibilities - managing stale issues and managing PR approvals. Consider splitting these into separate workflow files for better maintainability and to avoid having one workflow failure affect both processes.

The core functionality looks solid though - I like how you're handling both the 30-day stale approvals and the draft PR conversion cases. The logging is also thorough which will help with debugging.

lanej0:
Wanted to test out the PR review agent without pestering somebody else on their PR, so I ran it on my own!  That said:

> 1. The new job `remove-draft-approvals` seems to have incorrect YAML indentation. It's currently at the same level as the `stale` job, but it's indented as if it's part of the previous step. This will cause the workflow to fail. The job should be aligned with the `stale` job, like this:
> 
> ```miniyaml
> jobs:
>     stale:
>         # ... existing job ...
> 
>     remove-draft-approvals:  # <- Fix indentation here
>         runs-on: ubuntu-20.04
>         # ... rest of the job ...
> ```
> 

This appears to be a false positive.  `remove-draft-approvals` is a separate job, but within the same workflow.  I've double checked, and I think the indentation is right.

> 2. For the stale PR approvals removal script, consider adding some rate limiting or batching. The current implementation makes API calls in tight loops (one call per review dismissal) which could hit GitHub API rate limits on repositories with many PRs. You might want to:
>    
>    * Add a small delay between dismissals
>    * Or batch the dismissals into groups
>    * Or add error handling for rate limit responses

Solid suggestion, but it probably won't act on too many PRs each run, so I don't think we need to worry about rate-limiting.

> 3. The workflow now has two distinct responsibilities - managing stale issues and managing PR approvals. Consider splitting these into separate workflow files for better maintainability and to avoid having one workflow failure affect both processes.

True, but we've been trying to simplify and reduce the number of workflows in the monorepo so I don't think it's super crazy to group PRs and Issues together.

rodelgc (approved):
I tested this by creating a [new branch](https://github.com/woocommerce/woocommerce/tree/rodel/test-pull-54945) from this PR, and then:
- [disabling](https://github.com/woocommerce/woocommerce/commit/02e62a9317f5f8f831690695ecda9f6f63963065) the untouched steps in the `stale` job so that only the newly added `Remove stale PR approvals (over 30 days old)` would be tested when I run the workflow.
- [Remove part of code that dismisses reviews](https://github.com/woocommerce/woocommerce/commit/c838bf0cf21d27dd17d6d9ea19b84ddfd2e7e32b) and then [clarified](https://github.com/woocommerce/woocommerce/commit/de8d621cb6111a8803e35a869a8031abb66758ba) in the dismiss message that it's just a test, no reviews would be dismissed.
- Add a console.log() to [print number of PRs to be checked](https://github.com/woocommerce/woocommerce/commit/f67b0332660dfb0abffc68d3b979b86cc91271a1)
- [Disable the job `remove-draft-approvals`](https://github.com/woocommerce/woocommerce/commit/3815019cd7db725c7ff004a8c7c860dbaa652625) as I think there's no real way we could test it pre-merge. I think the best we can do on that job is to do a static code review, which looked good to me when I verified the usage of `dismissReview` with the octokit usage [doc](https://octokit.github.io/rest.js/v21/#usage). ✅ 

I then [ran](https://github.com/woocommerce/woocommerce/actions/runs/13197016589/job/36840438337#step:2:125) this workflow using that branch, then cancelled it after about a minute. It ran in readonly mode since I removed the part of the code responsible for the actual dismissing of approvals.

The logs looked good to me:

- ✅ The script was able to spot multiple approvals in a PR like `PR 49417` ([log](https://github.com/woocommerce/woocommerce/actions/runs/13197016589/job/36840438337#step:2:125)) which contained 3 approval reviews, and `PR 47217` ([log](https://github.com/woocommerce/woocommerce/actions/runs/13197016589/job/36840438337#step:2:148)) which contained 2 approval reviews.
- ✅ Picked a few PRs that the workflow said has 1 approval for dismissal, and verified manually if the review approval was indeed more than 30d old.
- ✅ Picked a few PRs that the workflow said it will keep the approval of, and verified manually if the review approval was indeed within 30d.
- ✅ I explored for ways on how a PR could be approved without getting a review comment, but it seems like it's not possible (at least in our repo) so I think we're good to rely on the review comments for determining the approval state of the PRs.
- ✅ I thought maybe we'll need a token that's more privileged than the default GITHUB_TOKEN for dismissing the reviews but it seems that github-bot was able to successfully dismiss them during your test run so we're good too on the token side of things.

Maybe just a little heads-up on the performance aspect: the run showed that there are [373 PRs to be looked at](https://github.com/woocommerce/woocommerce/actions/runs/13197016589/job/36840438337#step:2:62). The run processed around 38 PRs (almost a tenth of the total PRs to be checked) which took `1m 3s` to complete in readonly mode (just logging the PRs it encountered, not actually performing the dismissal of the review.) This means we'll be calling the `github.rest.pulls.listReviews` on approximately 300-400 open PRs everyday via the `schedule` event. I'm not sure myself if we'll hit any kind of limit from it, but maybe we could think of a few ways to reduce the number of PRs that this workflow would perform a `listReview` on?

Off the top of my head, and it's just me thinking out loud, we can probably shrink the number of PRs by:
- excluding PRs that were created within the past 30d by looking at their `created_at` field
- exclude PRs that aren't targeting `trunk`? This would also enable the workflow to skip PR's that are CFEs or PRRs, since they usually target release branches. Although I think it would be rare to have a CFE or PRR  pull request to be open for more than 30d.

If the workflow doesn't seem to cause any performance issues in CI after it's merged, then I guess there's no need to trim down the number of PRs.

I saw a minor improvement that could be made on the nested `if()`s that I'll mention on a separate review comment. I'm approving this PR anyway since that's just a minor change. I'm also fine with addressing that on a separate PR, your call. Thanks for the work on this @lanej0

.github/workflows/stalebot.yml
```diff
@@ -50,3 +53,93 @@ jobs:
...
+                              latestApproval = reviewDate;
+                            }
+                          }
+                        }
+                        
+                        // Process only if we found any approvals
+                        if (latestApproval) {
+                          console.log(`PR #${pr.number} latest approval on ${latestApproval.toISOString()}`);
+                          
+                          if (latestApproval < thirtyDaysAgo) {
```
rodelgc, line 96: I think we can safely combine these two if's via `if ( latestApproval && latestApproval < thirtyDaysAgo )` while achieving the same effect in terms of the logging of the PR number? This will save us one level of nested if's. This is more of a nit-pick though.

Issue #52970, opened 2024-11-19
[Settings] Custom views - Handle various action output


psealock:
Use output buffering to capture the output of various actions:

* [`woocommerce_settings_start`](https://github.com/woocommerce/woocommerce/blob/2d65567a6ad261a043a95b4f758f72ed190fd0de/plugins/woocommerce/includes/admin/class-wc-admin-settings.php#L142)
* [`woocommerce_before_settings_xxx`](https://github.com/woocommerce/woocommerce/blob/2d65567a6ad261a043a95b4f758f72ed190fd0de/plugins/woocommerce/includes/admin/views/html-admin-settings.php#L27)
* [`woocommerce_after_settings_xxx`](https://github.com/woocommerce/woocommerce/blob/2d65567a6ad261a043a95b4f758f72ed190fd0de/plugins/woocommerce/includes/admin/views/html-admin-settings.php#L63)
* [`woocommerce_settings_xxx_end`](https://github.com/woocommerce/woocommerce/blob/2d65567a6ad261a043a95b4f758f72ed190fd0de/plugins/woocommerce/includes/admin/class-wc-admin-settings.php#L308)
* [`woocommerce_settings_xxx_after`](https://github.com/woocommerce/woocommerce/blob/2d65567a6ad261a043a95b4f758f72ed190fd0de/plugins/woocommerce/includes/admin/class-wc-admin-settings.php#L312)
* [`woocommerce_settings_xxx` (title id)](https://github.com/woocommerce/woocommerce/blob/2d65567a6ad261a043a95b4f758f72ed190fd0de/plugins/woocommerce/includes/admin/class-wc-admin-settings.php#L292)

Pass the value along to Settings data in a way that can be rendered to the page, making sure the area where output is rendered is in the right place.

PR #54948, opened 2025-01-29
[Settings] Add support for `woocommerce_settings_start` hook

psealock:

Related to https://github.com/woocommerce/woocommerce/issues/52970

This PR adds support for the `woocommerce_settings_start` hook which places content above all Settings. I decided to change the shape of the data to something like this (if it makes sense). The hooks to be added as part of https://github.com/woocommerce/woocommerce/issues/52970 tend to sandwich content before and after.

```js
{
  start: CustomSettingsField,
  pages | sections | settings: SettingsPage | SettingsSection | SettingsField
  end: CustomSettingsField,
}
```

1. Add/Enable Gutenberg 19.8 or older
2. Add/Enable [Settings Tester](https://github.com/woocommerce/woocommerce-settings-migration-tester), making sure you have the latest commits. You can create a zip by running `npm run plugin-zip` from project root.
3. Visit multiple pages and note the banner injected by running the [`woocommerce_settings_start` hook from the Settings Tester](https://github.com/woocommerce/woocommerce-settings-migration-tester/blob/ad0aec65e1d50b283604a1d42335fa647836b71b/includes/Admin/Settings/LegacySettings.php#L28).
4. Turn on the Settings feature flag and ensure the banner is rendered in the correct place. Note that the new Settings has the banner below the tabs, while its above in the legacy screens. Since the secondary navigation will be moved to the sidebar (See designs in pfrEX1-NV-p2) I think this placement works for now.

image

image

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

Support woocommerce_settings_start hook in new, unreleased Settings

</details>


packages/js/settings-editor/src/legacy/content.tsx
```diff
@@ -36,15 +37,22 @@ const Form = ( { settings }: { settings: SettingsField[] } ) => {
...
 	const section = settingsPage.sections[ activeSection ];
 
 	if ( ! section ) {
 		return null;
 	}
 
-	return <Form settings={ section.settings } />;
+	return (
+		<>
+			<CustomView html={ settingsData.start.content } />
```
ilyasfoo: I think this is blocking build https://github.com/woocommerce/woocommerce/actions/runs/13065656219/job/36457497092?pr=54948
```
Error: ../../packages/js/settings-editor build:project:cjs: src/legacy/content.tsx(54,23): error TS18048: 'settingsData.start' is possibly 'undefined'.
```

ilyasfoo: @psealock

psealock: Thanks @ilyasfoo, fixed in https://github.com/woocommerce/woocommerce/pull/54948/commits/343d947b5d8ec6bbaa1c63817072628afc272932

packages/js/settings-editor/src/components/sidebar/index.tsx
```diff
@@ -19,15 +19,15 @@ const { Icon, ...icons } = IconPackage;
 
 const SidebarNavigationScreenContent = ( {
 	activePage,
-	settingsData,
+	pages,
 }: {
 	activePage: string;
-	settingsData: SettingsData;
+	pages: { [ key: string ]: SettingsPage };
```
chihsuan: A minor suggestion: It might be worth it to create a new type for the pages for reuse in other places.

plugins/woocommerce/includes/admin/settings/class-wc-settings-page.php
```diff
@@ -252,6 +248,20 @@ protected function get_section_settings_data( $section_id, $sections ) {
 			return $section_settings_data;
 		}
 
+		/**
+		 * Get the custom view object.
+		 *
+		 * @param string $content The content of the custom view.
+		 * @return array The custom view object.
+		 */
+		public function get_custom_view_object( $content ) {
```
chihsuan, line 257: nit: I'm not sure if it will be used by other classes. If it doesn't, we can convert it to a private or protected method.

plugins/woocommerce/src/Admin/Features/Settings/Init.php
```diff
@@ -172,8 +172,13 @@ public static function add_component_settings( $settings ) {
 			$settings['settingsScripts'][ $setting_page->get_id() ] = self::get_script_urls( $settings_scripts_handles );
 		}
 
-		$transformer              = new Transformer();
-		$settings['settingsData'] = $transformer->transform( $pages );
+		$transformer                       = new Transformer();
+		$settings['settingsData']['pages'] = $transformer->transform( $pages );
+		$start_hook_content                = $setting_pages[0]->get_custom_view( 'woocommerce_settings_start' );
```
chihsuan, line 177: I'm not familiar with the hook. Just curious, does this mean the `woocommerce_settings_start` hook always returns the same content for all pages?

chihsuan (approved):
Nice work! @psealock 👍  Confirmed the banner is rendered in the correct place. Just left a few minor comments but not a blocker.

ilyasfoo (approved):
Tests well, LGTM!

plugins/woocommerce/src/Admin/Features/Settings/Init.php
psealock: > Just curious, does this mean the `woocommerce_settings_start` hook always returns the same content for all pages?

Yes, thats right. Its added before any content and does not change based on query parameters.

https://github.com/woocommerce/woocommerce/blob/0135e4943791d9ba8a6281c0f453d6b9754c0225/plugins/woocommerce/includes/admin/class-wc-admin-settings.php#L142

packages/js/settings-editor/src/route.tsx
```diff
@@ -38,20 +38,19 @@ const NotFound = () => {
 /**
  * Default route when active page is not found.
  *
- * @param {string}       activePage   - The active page.
- * @param {settingsData} settingsData - The settings data.
- *
+ * @param {string}        activePage - The active page.
+ * @param {SettingsPages} pages      - The settings pages.
```
psealock: @chihsuan The linter is giving me this error even though `SettingsPages` is defined in `global.d.ts`. Do you know what is required to get eslint to respect globals?

```
42:1  error  The type 'SettingsPages' is undefined  jsdoc/no-undefined-types`
```

chihsuan: Ah, it looks like ESL int's JSDoc checker doesn't automatically recognize types defined in global.d.ts. We have to use import syntax or configure ESLint to recognize global types like:

```js
module.exports = {
  rules: {
    'jsdoc/no-undefined-types': ['error', {
      definedTypes: [
        // Add your global types here
        'SettingsPages',
        'SettingsPage',
        'SettingsData'
      ]
    }]
  }
}
```

I see that the other JSdoc params don't have the same issue because we use `settingsPage` instead of `SettingsPage`. I think JSDoc checker thinks `settingsPage` is a variable so doesn't throw any error but it's actually be considered as `any` type.

chihsuan: I think we can move the types from global.d.ts to the typings instead. It will need to import the types in the files that need them, but it's more explicit and can be imported if someone needs them when using the package. What do you think? I can make the change.

psealock: Oh wow, ok yeah thats not ideal. Thank you! I lost a good chunk of time with this one already.

>I think we can move the types from global.d.ts to the typings instead.

I'm trying to see if `typings/global.d.ts` can be added in the configuration somehow, but no luck so far.

If that doesn't work, I think its ok to be explicit and move the types.

PR #54956, opened 2025-01-29
Workflow to regenerate manifest and push to trunk upon PR merge

timur27:

Closes # .

1.
2.
3.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/.github/workflows/docs-manifest.yml
+++ b/.github/workflows/docs-manifest.yml
@@ -1,0 +1,50 @@
+name: Update Docs Manifest
+
+on:
+  pull_request_target:
+    types: [closed]
+    paths:
+      - 'docs/**'  # Only run when files in docs directory change
+
+jobs:
+  update-manifest:
+    # Only run if the PR was merged (not just closed)
+    if: github.event.pull_request.merged == true && github.repository_owner == 'woocommerce'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          ref: trunk
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16'
+
+      - name: Install pnpm
+        uses: pnpm/action-setup@v2
+
+      - name: Install dependencies
+        run: pnpm install
+
+      - name: Generate and commit manifest
+        run: |
+          git config --local user.email "github-actions[bot]@users.noreply.github.com"
+          git config --local user.name "github-actions[bot]"
+          
+          # Generate new manifest
+          pnpm utils md-docs create docs woocommerce -o docs/docs-manifest.json
+          
+          # Check if there are changes
+          if [[ -n "$(git status --porcelain docs/docs-manifest.json)" ]]; then
+            git add docs/docs-manifest.json
+            git commit -m "Update docs manifest after merging #${{ github.event.pull_request.number }}"
+            git push origin trunk
+          else
+            echo "No manifest changes needed"
+          fi 


.github/workflows/docs-manifest.yml
```diff
@@ -0,0 +1,50 @@
...
+          pnpm utils md-docs create docs woocommerce -o docs/docs-manifest.json
+          
+          # Check if there are changes
+          if [[ -n "$(git status --porcelain docs/docs-manifest.json)" ]]; then
+            git add docs/docs-manifest.json
+            git commit -m "Update docs manifest after merging #${{ github.event.pull_request.number }}"
+            git push origin trunk
+          else
+            echo "No manifest changes needed"
+          fi
```
timur27, line 50: I want to consult this before moving forward. I'm not sure about this change because it seems that we don't have automated direct pushes to `trunk` in this repository. 

Our goal is to automate the regeneration of the docs manifest. One option we're considering is to trigger this regeneration automatically after a pull request that modifies the documentation is merged. This approach aims to eliminate the manual regeneration process and reduce the frequent conflicts that occur. 

Another option for this workflow would be to create a follow up PRs with updated docs-manifests, but I think this doesn't solve our main problem (conflicts still possible) and creates an extra step where we have a PR that should be reviewed by someone with less context that previously.

timur27:
@samueljseay @psealock hey folks, sorry for the direct ping. I just wanted to quickly consult with you on one aspect related to the manifest generation tool, in which you seem to play a crucial role in developing. Could you please take a look at https://github.com/woocommerce/woocommerce/pull/54956/files#r1933459443? Thanks so much for your time 🙏

kalessil:
Hey @timur27 :wave: I understand the original problems, but I'm unsure if the approach is worth pursuing.

I'd propose an alternative (I didn't check if it's feasible, though) - removing the file from the repository and generating it on the external consumer side. Would this be feasible?

A bit more context for the proposed alternative: similar to lock files (pnpm and composer), we expect the conflicts to be solved in corresponding PRs and pass review (even if manifest changes review is nominal). Also, direct pushes for the trunk, in this case, don't feel right.

timur27:
Hi @kalessil, thanks for your time reviewing this! I agree, [I'm not sure about the approach either](https://github.com/woocommerce/woocommerce/pull/54956#discussion_r1933459443) The `md-docs` tool was initially designed to work with local files and be kept together with the documentation. I think your proposal makes sense in general, I just need to look into how much of a refactor it would be and prioritize accordingly. 

I'll keep this PR in draft for now and will get back to it a bit later. Maybe @samueljseay or @psealock will be able to add any extra thoughts meanwhile.

.github/workflows/docs-manifest.yml
psealock: > I'm not sure about this change because it seems that we don't have automated direct pushes to trunk in this repository.

trunk is a protected branch, so some configuration would be needed to allow this. Why not run this check as part of the regular checks? If any `*.md` files have changed run the manifest cli command and push changes directly to the PR.

The changelog scripts do something similar where any edit to the description potentially kicks off the script to add a changelog entry by pushing directly to the PR. If you're editing a markdown document that might be kind of annoying to have new commits to pull down after every push with new changes.

Can you fire the manifest script at a different time in the lifecycle? Before merge? On ready for review? If any of these suggestions don't pan out you could continue with the approach here but I think you'd need to allow github bot special permission to push to trunk and carefully think about the security implications of that.

timur27: Thanks for the response @psealock! Automate it into the PR flow makes sense for sure, I'm just trying to find ways to help less-tech users avoid solving manifest conflicts. This happens pretty often and avoid doing this would save the team some time.

I totally agree with the way of thinking in your comment. I think if avoiding the problem I described with this change won't be possible, I'll leave things as they are and automate manifest generation command only.

kalessil: > If any *.md files have changed run the manifest cli command and push changes directly to the PR.
> The changelog scripts do something similar where any edit to the description potentially kicks off the script to add a changelog entry by pushing directly to the PR ...

That is a good idea @psealock :+1:

> I'm just trying to find ways to help less-tech users avoid solving manifest conflicts

Potential conflicts are indeed something to consider @timur27, as reliably generating manifest implies the PR branch is up to date with the target branch. `changelog-auto-add.yml` in that regard is well off.

kalessil:
Closing as docs now migrated to Docusaurus and the manifest is no more.

PR #54961, opened 2025-01-29
Improve order details in plain text emails

### LINKED ISSUE #54465: Update plain text emails
Now that the HTML emails are modernized, we need to make a few small changes in plain text emails to improve order detail rendering. 

- Unify how order number and date is rendered
- Add string padding to align product prices and total prices.
- Replace `X` with `×` in item quantity.

NeosinneR:

Screenshot 2025-01-29 at 11 09 42

Improve order details in plain text emails:
- Unify with the HTML version how the order number and date are rendered.
- Add string padding to align product prices and total prices.
    - Note that while prices are aligned within products and totals, they might not be aligned together. This is because we use [`str_pad()`](https://www.php.net/manual/en/function.str-pad.php) instead of [`mb_str_pad()`](https://www.php.net/manual/en/function.mb-str-pad.php) which was added only in PHP 8.3, and would handle multi-byte strings correctly. I don't think it's worth writing a polyfill in this case. 
- Replace `X` with `×` in item quantity.

Closes #54465.

1. Enable `Email improvements` feature in **WooCommerce > Settings > Advanced > Features**.
2. Go to **WooCommerce > Settings > Emails**.
3. Manage any order-related email.
4. Change the email type to `Plain text`.
5. Observe changes in the email preview.

### DIFF
--- a/plugins/woocommerce/templates/emails/plain/email-order-details.php
+++ b/plugins/woocommerce/templates/emails/plain/email-order-details.php
@@ -12,15 +12,25 @@
  *
  * @see https://woocommerce.com/document/template-structure/
  * @package WooCommerce\Templates\Emails
- * @version 3.7.0
+ * @version 9.8.0
  */
+
+use Automattic\WooCommerce\Utilities\FeaturesUtil;
 
 defined( 'ABSPATH' ) || exit;
 
+$email_improvements_enabled = FeaturesUtil::feature_is_enabled( 'email_improvements' );
+
 do_action( 'woocommerce_email_before_order_table', $order, $sent_to_admin, $plain_text, $email );
 
-/* translators: %1$s: Order ID. %2$s: Order date */
-echo wp_kses_post( wc_strtoupper( sprintf( esc_html__( '[Order #%1$s] (%2$s)', 'woocommerce' ), $order->get_order_number(), wc_format_datetime( $order->get_date_created() ) ) ) ) . "\n";
+if ( $email_improvements_enabled ) {
+	/* translators: %1$s: Order ID. %2$s: Order date */
+	echo wp_kses_post( sprintf( esc_html__( 'Order #%1$s (%2$s)', 'woocommerce' ), $order->get_order_number(), wc_format_datetime( $order->get_date_created() ) ) ) . "\n";
+	echo "\n==========\n";
+} else {
+	/* translators: %1$s: Order ID. %2$s: Order date */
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/templates/emails/plain/email-order-items.php
```diff
@@ -30,14 +34,25 @@
...
-		echo wp_kses_post( apply_filters( 'woocommerce_order_item_name', $item->get_name(), $item, false ) );
-		if ( $show_sku && $sku ) {
-			echo ' (#' . $sku . ')';
+		if ( $email_improvements_enabled ) {
+			$product_name = apply_filters( 'woocommerce_order_item_name', $item->get_name(), $item, false );
+			$product_name .= ' × ' . apply_filters( 'woocommerce_email_order_item_quantity', $item->get_quantity(), $item );
+			echo wp_kses_post( str_pad( wp_kses_post( $product_name ), 40 ) );
+			echo ' ';
+			echo esc_html( str_pad( wp_kses( $order->get_formatted_line_subtotal( $item ), array() ), 20, ' ', STR_PAD_LEFT ) ) . "\n";
+			if ( $sku ) {
```
alex-mpoet: A check for `$show_sku` seems to be missing.

plugins/woocommerce/templates/emails/plain/email-order-details.php
```diff
@@ -53,7 +53,7 @@
 }
 
 if ( $order->get_customer_note() ) {
-	echo esc_html__( 'Note:', 'woocommerce' ) . "\t " . wp_kses( wptexturize( $order->get_customer_note() ), array() ) . "\n";
+	echo "\n" . esc_html__( 'Note:', 'woocommerce' ) . "\n" . wp_kses( wptexturize( $order->get_customer_note() ), array() ) . "\n";
```
alex-mpoet: This change is outside the feature flag, is that how it should be?

alex-mpoet (commented):
@NeosinneR The improved plain-text emails look cool, I like them! I only have a couple notes.

plugins/woocommerce/templates/emails/plain/email-order-items.php
NeosinneR: Yep. I remember removing this when testing it, and forgot to add it back. Fixed in https://github.com/woocommerce/woocommerce/pull/54961/commits/8b951468d9030d0eae6ea93ad653a38c50f58f88

plugins/woocommerce/templates/emails/plain/email-order-details.php
NeosinneR: I saw it as a minor change to just change it globally, but let's keep it clean. Added the feature check in https://github.com/woocommerce/woocommerce/pull/54961/commits/8b951468d9030d0eae6ea93ad653a38c50f58f88

alex-mpoet (approved):

PR #54979, opened 2025-01-29
Sync the @wordpress/data and interactivity API cart stores when there is a change in any of them

### LINKED ISSUE #54501: Create an event to sync wc data store and interactivity store when there is a change in each of them.
See epic for more info: https://github.com/woocommerce/woocommerce/issues/54259

luisherranz:

This is still very preliminary.

The idea of this pull request is to synchronize the store created with the `@wordpress/data` package with the new store of the Interactivity API. So that when there are changes in one, the other store reconciles itself by making a new request to the Store API to update its internal state.

There are several ways to synchronize both stores:

1.  Observe the state to monitor when it changes and trigger the event.
3. Create a middleware that emits the event when an action is emitted from a specific list of actions that modify the state.
4.  Trigger the event within the actions that modify the state.

For now, I have tried to make it work by synchronizing any change in the state since, if we can make it work like that, it would be the most solid way.

It is also necessary to find a mechanism to prevent infinite loops to ensure that the synchronization of one of the stores does not trigger the synchronization event of the other store. Right now, I'm using a variable that ignores any change in the state while a synchronization is taking place. However, I'm not sure that it's the best solution because that could cause other changes that occur in between to be ignored.

Finally, it is necessary to investigate what happens when the page is loaded, since both stores make a request to the Store API to update their state and, in theory, it would be good to ignore these changes. This also means that it would be good to start thinking about how to better organize the store of the Interactivity API in this regard because right now this initial synchronization is being delegated to the blocks, and maybe it's something it should do on its own.

Closes #54501.

Basically, adding products to the cart with the Product Button block and checking that they are correctly updated in the Minicart block or vice versa. But I will update these instructions when I move forward with the implementation and the Pull Request is ready to review.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce-blocks/assets/js/data/cart/index.ts
+++ b/plugins/woocommerce-blocks/assets/js/data/cart/index.ts
@@ -6,6 +6,7 @@ import {
 	subscribe,
 	createReduxStore,
 	dispatch as wpDispatch,
+	select,
 } from '@wordpress/data';
 import { controls as dataControls } from '@wordpress/data-controls';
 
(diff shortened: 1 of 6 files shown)


plugins/woocommerce-blocks/assets/js/data/cart/index.ts
```diff
@@ -63,6 +64,94 @@ window.addEventListener( 'load', () => {
...
+		const customEvent = event as CustomEvent< {
+			type: string;
+			id: number;
+		} >;
+		if ( customEvent.detail.type === 'from_iAPI' ) {
+			console.log(
+				`Cart sync received on the @wordpress/data store: iapi-${ customEvent.detail.id }`
+			);
+
+			// Todo: investigate how to avoid infinite loops without causing racing conditions.
```
samueljseay: @luisherranz I'm definitely concerned about how fragile this is.

I haven't really thought this through too much, but since you're quite well versed with interactivity API internals maybe you'll be able to spot immediate problems:

What if we modified the redux store since `@wordpress/data` has this [ability to implement a custom store](https://github.com/WordPress/gutenberg/blob/trunk/packages/data/README.md), and have the redux store access the interactivity API store to store/modify state? Maybe we'd need to adjust reducers to remove immutability pattern of state updates, which is.. a bit dubious (although maybe we could clone objects or "something" before they go in and out of the reducer so they work with existing patterns) but I'm really concerned with trying to implement custom bi-directional state updates that it could introduce problems. race conditions like you say here.

luisherranz: The idea of synchronizing the state directly seems less secure to me than reconciling the stores through a call to the Store API.

I'm going to give it a thought to see if I can improve the way to avoid the infinite loop, but keep in mind that it will be very difficult for it to happen since, in order to cause problems, the user would have to start an action in the Minicart block, close it, start an action in some block that uses the interactivity API, and that the fetch triggered by that action takes less than the fetch of the synchronization:

image

(and something similar in the other direction)

luisherranz: In the iAPI store, I have moved the synchronization inside the actions. I like it less in the sense that we have to remember to emit the event in each action so it's less decoupled, but this way we avoid both the synchronization in the optimistic updates (where it is not necessary to synchronize) and the race conditions.

In the Redux Store, I'm still using the state, but I've moved the ignore flag to inside the actions, after receiving the response from the Store API, to make it synchronous and eliminate race conditions. I have tried to add a middleware to the store, but I have not found a way to do it. In any case, the way this store is built, it seems that all the actions that modify the cart after a Store API request end up using the same thunk (`receiveCart`). We will have to do the same for other parts of the store if we need to synchronize them in the future, though.

@samueljseay Let me know what you think 🙂

PR #54984, opened 2025-01-29
Improve caching of wc_count_comments counts to avoid re-queries after every order note.

### LINKED ISSUE #54286: Query used to provide filtered comment counts is slow and the transient constantly cleared on active sites
We're seeing slow queries from the query run within `WC_Comments::wp_count_comments()`.  We're seeing some sites where these counts take a couple seconds to even minutes to run. The comment counts are cached, however that cache is cleared every time a new comment is inserted, regardless of type.  This creates a scenario for store admin where the slow query is run frequently while working within the admin so that any random admin page may have a spike in response time.

* The comment count is used frequently throughout the default WordPress admin.
* Order notes use the WordPress comment system and are inserted frequently for order actions.
* The insertion of the order notes cause the cached counts to be cleared even though they aren't included in the counts.

The query:

```
SELECT comment_approved, COUNT(*) AS num_comments
FROM wp_comments
LEFT JOIN wp_posts ON comment_post_ID = wp_posts.ID
WHERE comment_type NOT IN ('action_log', 'order_note', 'webhook_delivery') AND wp_posts.post_type NOT IN ('product')
GROUP BY comment_approved
```

Todo:

- [x] Determine steps to reproduce the slow queries.  How many comments before the query starts to slow down?
- [x] Determine what other scenarios, beside insert, cause the transient to be cleared.
- [ ] Determine if the `woo_idx_comment_type` index, specifically created for this use should be improved.  It doesn't currently have high enough cardinality to be effective.

Initial thoughts:  We should attempt to avoid clearing the transient for comment types that we normally exclude from the count.

prettyboymp:

The following changes are made in this PR:

1. Switches the comment count caching from the transient interface to wp_cache. This allows the wp_cache_incr/decr methods to be used and is more aligned with only using transients for much slower actions.
2. Updates the callback handling for hooks related to comment modification to increment/decrement the cache wc_count_comments values for comment and post types that aren't included in the counted results stored in cache.
3. Updates the caching callback from `wp_count_comments` to use `get_comments()` and the `comments_clauses` filters rather than a custom query. So other extensions can also filter the counts as needed.

Closes #54286.  

#### Setup:
1. A site running persistent object cache is required to see these improvements.
2. Use a plugin like Query Monitor to view queries.
3. Optionally add a call to `wp_cache_set_comments_last_changed();` in the theme, mu-plugins, etc where it will mimic the default cache busting for WP_Comments_Query when a new comment is created.  This just makes sure any caching issues aren't being covered up by the built in cache.

Note: You will come across a query like `SELECT COUNT(*) FROM wp_comments .. AND comment_type IN ('review', '', 'comment')`.  This is from the Product Reviews pending count bubble in the admin menu.  It is outside the scope of this PR and additional caching will need to be added for it in another PR - ref: #55015. 

#### Admin Menu Comments Bubble

1. Go to an admin page that doesn't include comment counts on it's own, e.g. like Admin -> Tools.
2. On the first visit, you may see queries like the below to count the comments by status, however, any repeat requests should not contain the queries as they will have already been cached
  ```
SELECT COUNT(*)
FROM wp_comments
LEFT JOIN wp_posts AS wp_posts_to_exclude_reviews
ON comment_post_ID = wp_posts_to_exclude_reviews.ID
WHERE ( comment_approved = '1' )
AND comment_type != 'order_note'
AND comment_type != 'webhook_delivery'
AND comment_type != 'action_log'
AND wp_posts_to_exclude_reviews.post_type NOT IN ('product')
  ```
3. Visit a post on the site and submit a comment.
4. Refresh the admin page and verify that the above query did not need to run again.
5. Open an order to edit and add and Order Note.
6. Again, refresh the admin page and verify that the above query did not need to run again.

#### Comments Admin

1. Visit the Comments Admin page.
2. Check that count queries are not needed to render the comment status filters at the top of the page.
3. Visit each comment status and verify that the item count for each page matches the count in the status filter menu.
4. Update the status of some comments, refresh the page, and verify that the counts are updated properly.
5. Go to a post on the front end and post a comment.  Verify that the comment count is updated.
6. Go to a product on the front end and post a review.  Verify that the review isn't filtered after a refresh on the product and that the comment count was not changed by the review.

Note: The item count next to the pagination is not currently cached.  This is outside the scope of this PR and may be addressed in the future.

#### Reviews Admin

1. Visit the Reviews Admin
2. Verify that the counts and reviews are displaying the correct numbers and are not modified by the additional filters.


plugins/woocommerce/includes/class-wc-comments.php
```diff
@@ -215,6 +215,48 @@ public static function clear_transients( $post_id ) {
...
+	 * Callback for 'wp_insert_comment' to delete the comment count cache if the comment is included in the count.
+	 *
+	 * @param int        $comment_id The comment ID.
+	 * @param WP_Comment $comment Comment object.
+	 *
+	 * @return void
+	 */
+	public static function delete_comments_count_cache_on_wp_insert_comment( $comment_id, $comment ) {
+		if ( ! self::is_comment_excluded_from_wp_comment_counts( $comment ) ) {
+			self::delete_comments_count_cache();
```
joshuatf: Is it possible to adjust the cache count on insert if the comment is qualified for the count (similar to https://github.com/woocommerce/woocommerce/pull/54034)?  Or do we have to remove the cache and recalculate this on these occassions?

plugins/woocommerce/includes/class-wc-comments.php
```diff
@@ -215,6 +215,48 @@ public static function clear_transients( $post_id ) {
...
+	 *
+	 * @param int|string $new_status The new comment status.
+	 * @param int|string $old_status The old comment status.
+	 * @param WP_Comment $comment    Comment object.
+	 *
+	 * @return void
+	 */
+	public static function delete_comments_count_cache_on_comment_status_change( $new_status, $old_status, $comment ) {
+		if ( ! self::is_comment_excluded_from_wp_comment_counts( $comment ) ) {
+			self::delete_comments_count_cache();
```
joshuatf: Same question as above.

joshuatf (commented):
Thanks for tackling this, Michael!  Looks good overall; was just curious if we were able to bump the cache numbers without deleting the cache.

prettyboymp:
I updated to increment/decrement the counts instead of clearing the cache.  We'll want to implement a job to occasionally refresh the cache to avoid any drift that may occur on systems that don't have an atomic way to increment/decrement.

prettyboymp:
@joshuatf , I switched this over to use incr/decr methods to avoid having to clear the cached counts.  I also went ahead and set a 3 Day timeout for the cache as a place holder to account for any potential drift until we start using jobs to regenerate the cached counts.

plugins/woocommerce/includes/class-wc-comments.php
```diff
@@ -17,6 +17,11 @@
  */
 class WC_Comments {
 
+	/**
+	 * @var string COMMENT_COUNT_CACHE_GROUP The cache group to use for comment counts.
+	 */
+	private CONST COMMENT_COUNT_CACHE_GROUP = 'wc_comment_counts';
```
joshuatf:
```suggestion
	private const COMMENT_COUNT_CACHE_GROUP = 'wc_comment_counts';
```

plugins/woocommerce/includes/class-wc-comments.php
```diff
@@ -45,8 +50,8 @@ public static function init() {
 		add_filter( 'wp_count_comments', array( __CLASS__, 'wp_count_comments' ), 10, 2 );
```
joshuatf, line 56: It took me a really long time to debug an issue I was having here that turned out to be Docket Cache.  [It also attempts to cache `wp_count_comment` stats](https://github.com/nawawi/docket-cache/blob/0d4482a099508c696b7c5f2314b3077cec6710d8/includes/src/Tweaks.php#L133-L179) and in doing so, overrides ours.

A quick fix is to use `PHP_INT_MAX` for the priority, but this only works because (usually) the `woocommerce/` plugin folder will alphabetically run after `docket-cache` since they have the same priority.

My initial thinking was that Docket Cache's approach was wrong, but after giving this more thought, our approach with caching comments and our query below only works if there are not other comment types or exclusionary rules applied to `wp_count_comments`, otherwise it's just a matter of whoever filters `wp_count_comments` last.

To allow other plugins to continue to extend this filter, ideally we would check if `$counts` passed to the `wp_count_comments` filter was not empty and subtract our counts that we would like to exclude as opposed to directly returning a set of included counts.  For example:

```php
$counts (
    [approved] => 3
    [moderated] => 0
    [spam] => 0
    [trash] => 0
    [post-trashed] => 0
    [all] => 3
    [total_comments] => 3
)
$product_counts (
    [approved] => 1
    [moderated] => 0
    [spam] => 0
    [trash] => 0
    [post-trashed] => 0
    [all] => 2
    [total_comments] => 2
)
$diff = array_intersect( $counts, array_diff( $counts, $product_counts ) );
return $diff;
```

However, the initial consumer of the filter would need to use something like `get_comment_count( 0 )` to populate the initial value.  Another option is just to filter the comment query so that counts are not included unless the comment type is specified:

```php
add_action( 'pre_get_comments', function( $query ) {
    if ( ! isset( $query->query_vars['type'] ) ) {
        $not_in = (array) $query->query_vars['type__not_in'];
        $not_in[] = 'review';
        $query->query_vars['type__not_in'] = $not_in;
    }
} );
```

But it seems like our existing logic has been around for some time and I couldn't find any issues related to this so maybe a quick patch related to Docket Cache is enough and the comment type count conflict is too much of an edge case.

joshuatf (commented):
This is testing really well overall and is really great to cut down on the number of times we see this query.

I ran into a rather tricky issue with Docket Cache that clashes with our own comments cache and went down a bit of a rabbit hole.  I left some comments around this, but it may be that a simple fix is in order.  I do want to at least get that plugin compatible since it's relatively popular outside of Redis and Memcached for object persistence.

Also, the testing instructions I think may be from before the cache increment updates, but here's what I tested and seemed to work well (please let me know if any of this is incorrect):

* ✅ Adding logging for cache hit/miss and cache is consistently hit after the first query.
* ✅ Leaving a comment increments the comment count in the dashboard without the need to re-query
* ✅ Adding order notes does not affect the comment count cache
* ✅ Adding product reviews does not affect the comment count cache
* ⚠️ Comment count is incorrectly overridden when Docket Cache is enabled

(Images are from Docket Cache enabled, note the missing pending comments despite a count existing)
Screenshot 2025-03-04 at 11 23 41 AM
Screenshot 2025-03-04 at 11 28 15 AM

plugins/woocommerce/includes/class-wc-comments.php
prettyboymp: Good catch on this. I addressed this in [0d40f40](https://github.com/woocommerce/woocommerce/pull/54984/commits/0d40f402ecfe45ae734b8dfa40ffcad1ae9c1d00).  Responding to some of this in the top level comments.

prettyboymp:
@joshuatf, I've made some adjustments based on your feedback.

I've updated the testing instructions to better reflect what this PR has turned into.

> It took me a really long time to debug an issue I was having here that turned out to be Docket Cache. [It also attempts to cache wp_count_comment stats](https://github.com/nawawi/docket-cache/blob/0d4482a099508c696b7c5f2314b3077cec6710d8/includes/src/Tweaks.php#L133-L179) and in doing so, overrides ours.
> 
> A quick fix is to use PHP_INT_MAX for the priority, but this only works because (usually) the woocommerce/ plugin folder will alphabetically run after docket-cache since they have the same priority.
> 
> My initial thinking was that Docket Cache's approach was wrong, but after giving this more thought, our approach with caching comments and our query below only works if there are not other comment types or exclusionary rules applied to wp_count_comments, otherwise it's just a matter of whoever filters wp_count_comments last.

I do think Docket Cache's approach is slightly wrong.  It is using a priority of `PHP_INT_MAX` and ignores any previously filtered `$stats` it receives.  It should probably only do one of these as this approach essentially means no other plugin can filter that hook.

> To allow other plugins to continue to extend this filter, ideally we would check if $counts passed to the wp_count_comments filter was not empty and subtract our counts that we would like to exclude as opposed to directly returning a set of included counts.

I looked into this, but didn't feel that we could confidently know the content of the `$stats` being passed in had already filtered some of the comments we were attempting to filter out and cause the numbers to be off.  My solution was to stop using a custom query, and switch to using the standard `get_comments()` like WordPress core already uses.  I then added the needed adjustments to the current `comments_clause` filters to make these properly filter out the comments we didn't want to be counted/shown outside of specific contexts. We were already running some of these filters, which is probably why the issue hasn't come up before with counts being off for a plugin like Docket Cache.

plugins/woocommerce/includes/class-wc-comments.php
```diff
@@ -215,80 +240,130 @@ public static function clear_transients( $post_id ) {
...
+			'orderby'                   => 'none',
+		);
+
+		foreach ( $stat_key_to_comment_query_status_mapping as $stat_key => $query_status ) {
+			// For simplicity, the cache key is by the comment status returned by wp_get_comment_status() and used by wp_transition_comment_status().
+			$cache_key = 'wc_count_comments_' . $comment_query_status_to_comment_status_mapping[ $query_status ];
+			$count     = wp_cache_get( $cache_key, self::COMMENT_COUNT_CACHE_GROUP );
+			if ( false === $count ) {
+				$count = get_comments( array_merge( $args, array( 'status' => $query_status ) ) );
+				wp_cache_set( $cache_key, $count, self::COMMENT_COUNT_CACHE_GROUP, 3 * DAY_IN_SECONDS );
```
laurendavissmith, line 355: Trivial but 3 * DAY_IN_SECONDS should be set to a constant

plugins/woocommerce/includes/class-wc-comments.php
```diff
@@ -215,80 +240,130 @@ public static function clear_transients( $post_id ) {
...
+			'update_comment_meta_cache' => false,
+			'orderby'                   => 'none',
+		);
+
+		foreach ( $stat_key_to_comment_query_status_mapping as $stat_key => $query_status ) {
+			// For simplicity, the cache key is by the comment status returned by wp_get_comment_status() and used by wp_transition_comment_status().
+			$cache_key = 'wc_count_comments_' . $comment_query_status_to_comment_status_mapping[ $query_status ];
+			$count     = wp_cache_get( $cache_key, self::COMMENT_COUNT_CACHE_GROUP );
+			if ( false === $count ) {
+				$count = get_comments( array_merge( $args, array( 'status' => $query_status ) ) );
```
laurendavissmith: Shouldn't we cast this to (int) in the event that it fails?

plugins/woocommerce/src/Internal/Admin/ProductReviews/ReviewsUtil.php
```diff
@@ -10,17 +10,52 @@ class ReviewsUtil {
...
+			}
+			if ( ! in_array( 'product', $post_type ) ) {
+				return $clauses;
+			}
+		}
+
+		/**
+		 * Any comment queries with these values are likely to be custom handling where we don't want to change default behavior.
+		 */
+		foreach ( [ 'ID', 'parent', 'parent__in', 'post_author__in', 'post_author', 'post_name', 'type', 'type__in', 'type__not_in', 'post_type__in' ] as $arg ) {
```
joshuatf: Should we also add `comment__in` and `comment__not_in` to this list of arguments?

plugins/woocommerce/src/Internal/Admin/ProductReviews/ReviewsUtil.php
```diff
@@ -10,17 +10,52 @@ class ReviewsUtil {
...
+			_prime_post_caches( $post_ids );
+			foreach ( $post_ids as $post_id ) {
+				if ( 'product' === get_post_type( $post_id ) ) {
+					return $clauses;
+				}
+			}
 		}
 
+		$clauses['join']  .= " LEFT JOIN {$wpdb->posts} AS wp_posts_to_exclude_reviews ON comment_post_ID = wp_posts_to_exclude_reviews.ID ";
+		$clauses['where'] .= ( $clauses['where'] ? ' AND ' : '' ) . " wp_posts_to_exclude_reviews.post_type NOT IN ('product') ";
```
joshuatf: I'm a little surprised we don't wrap the `$clauses['where']` with a `trim()` for safety, but this exact same logic is used in a bunch of other places across this repo and others without issue, so it may not be a likely scenario.

plugins/woocommerce/src/Internal/Admin/ProductReviews/ReviewsUtil.php
```diff
@@ -10,17 +10,52 @@ class ReviewsUtil {
...
+				return $clauses;
+			}
+		}
 
-		if ( isset( $current_screen->base ) && 'edit-comments' === $current_screen->base ) {
-			$clauses['join']  .= " LEFT JOIN {$wpdb->posts} AS wp_posts_to_exclude_reviews ON comment_post_ID = wp_posts_to_exclude_reviews.ID ";
-			$clauses['where'] .= ( $clauses['where'] ? ' AND ' : '' ) . " wp_posts_to_exclude_reviews.post_type NOT IN ('product') ";
+		if ( ! empty( $comment_query->query_vars['post__in'] ) ) {
+			$post_ids = wp_parse_id_list( $comment_query->query_vars['post__in'] );
+			_prime_post_caches( $post_ids );
```
joshuatf: Nice 💯

joshuatf (commented):
This is testing really well for me and I like the new solution!

* ✅ Admin comment counts are appearing correctly
* ✅ Admin product review counts are appearing correctly
* ✅ Counts for both post comments and product reviews are shown correctly on the frontend
* ✅ Cache is consistently hit and updated on comment status change

Screenshot 2025-03-10 at 3 46 10 PM
Screenshot 2025-03-10 at 3 44 26 PM

Added one comment about included some additional arguments for `WP_Comment_Query` where we should ignore our clauses, but otherwise this LGTM.

plugins/woocommerce/src/Internal/Admin/ProductReviews/ReviewsUtil.php
prettyboymp: Good call. I'm adding trim() to all the usages since comments are covered in this PR to prevent any future/current cases where another plugin adding a value like `' '` causing a failed query.

plugins/woocommerce/src/Internal/Admin/ProductReviews/ReviewsUtil.php
kalessil: Do we want to prime parts behind two additional parameters this function has (edit: or consider partial priming)?

plugins/woocommerce/includes/class-wc-comments.php
```diff
@@ -43,8 +43,11 @@ public static function init() {
 		add_filter( 'comments_clauses', array( __CLASS__, 'exclude_webhook_comments' ), 10, 1 );
 		add_filter( 'comment_feed_where', array( __CLASS__, 'exclude_webhook_comments_from_feed_where' ) );
 
+		// Secure potential remaining Action Logs.
+		add_filter( 'comments_clauses', array( __CLASS__, 'exclude_action_log_comments' ), 10, 2 );
+
 		// Exclude product reviews.
-		add_filter( 'comments_clauses', array( ReviewsUtil::class, 'comments_clauses_without_product_reviews' ), 10, 1 );
+		add_filter( 'comments_clauses', array( ReviewsUtil::class, 'comments_clauses_without_product_reviews' ), 10, 2 );
```
kalessil, line 53: Don't have full context here, but it seems that `comment_feed_where` handling was not added for the two new handlers of `comments_clauses`. If it's intentional, can we leave some comments on why we skipped them?

plugins/woocommerce/includes/class-wc-comments.php
```diff
@@ -230,7 +250,10 @@ public static function clear_transients( $post_id ) {
 	 */
 	public static function increment_comments_count_cache_on_wp_insert_comment( $comment_id, $comment ) {
 		if ( ! self::is_comment_excluded_from_wp_comment_counts( $comment ) ) {
-			wp_cache_incr( 'wc_count_comments_' . $comment->comment_approved, 1, self::COMMENT_COUNT_CACHE_GROUP );
+			$comment_status = wp_get_comment_status( $comment );
```
kalessil, line 256: `$comment_status = wp_get_comment_status( $comment );` good one!

plugins/woocommerce/includes/class-wc-comments.php
```diff
@@ -245,14 +268,7 @@ public static function increment_comments_count_cache_on_wp_insert_comment( $com
...
 		if ( ! self::is_comment_excluded_from_wp_comment_counts( $comment ) ) {
-			// WP Core uses different strings for the hooks than what is represented by the comment_approved value.
-			$old_status_map = [
-				'approved'   => '1',
-				'unapproved' => '0',
-			];
-			$old_status     = $old_status_map[ $old_status ] ?? $old_status;
-
-			wp_cache_incr( 'wc_count_comments_' . $comment->comment_approved, 1, self::COMMENT_COUNT_CACHE_GROUP );
+			wp_cache_incr( 'wc_count_comments_' . $new_status, 1, self::COMMENT_COUNT_CACHE_GROUP );
```
kalessil, line 274: Nit:

```php
    $is_incremented = false !== wp_cache_incr( 'wc_count_comments_' . $new_status, 1, self::COMMENT_COUNT_CACHE_GROUP );
    if ( $is_incremented ) {
        wp_cache_decr( 'wc_count_comments_' . $old_status, 1, self::COMMENT_COUNT_CACHE_GROUP );
    }
```

plugins/woocommerce/includes/class-wc-comments.php
```diff
@@ -280,76 +296,74 @@ public static function delete_comments_count_cache() {
...
+			$comment_statuses = array(
+				'approved',
+				'unapproved',
 				'spam',
 				'trash',
 				'post-trashed',
 			);
-			foreach ( $comment_approved_values as $approved_value ) {
-				wp_cache_delete( 'wc_count_comments_' . $approved_value, self::COMMENT_COUNT_CACHE_GROUP );
+			foreach ( $comment_statuses as $comment_status ) {
```
kalessil: Nit: `wp_cache_delete_multiple`

plugins/woocommerce/includes/class-wc-comments.php
```diff
@@ -215,80 +242,130 @@ public static function clear_transients( $post_id ) {
...
 	/**
 	 * Delete comments count cache whenever there is
 	 * new comment or the status of a comment changes. Cache
 	 * will be regenerated next time WC_Comments::wp_count_comments()
 	 * is called.
 	 */
 	public static function delete_comments_count_cache() {
-		delete_transient( 'wc_count_comments' );
+		if ( wp_cache_supports( 'flush_group' ) ) {
+			wp_cache_flush_group( self::COMMENT_COUNT_CACHE_GROUP );
```
kalessil: I'm also using this group in https://github.com/woocommerce/woocommerce/pull/56215/files#diff-389d36391ae130391874a998ea6e4a3ab4a6b902f223e6e7c1ba0c464579b03dR18 and I think we'll need to avoid `wp_cache_flush_group( self::COMMENT_COUNT_CACHE_GROUP )`.

Given my other comment prompting `wp_cache_delete_multiple` usage we should be good with just using it.

plugins/woocommerce/includes/class-wc-comments.php
prettyboymp: The two cached counts aren't reliant on each other.  `wp_cache_incr( 'wc_count_comments_' . $new_status );` will return `false` if it has fallen out of cache and regenerate the next request to retrieve it.
We still need to attempt to decrement `'wc_count_comments_'. $old_status`, otherwise, if it is still cached, it will be off by one since it won't regenerate.

kalessil: :+1:

plugins/woocommerce/src/Internal/Admin/ProductReviews/ReviewsUtil.php
prettyboymp: > Do we want to prime parts behind two additional parameters this function has (edit: or consider partial priming)?

Good catch on that.  We only needed the post object for these.

prettyboymp:
@kalessil updates are in [95d1af2](https://github.com/woocommerce/woocommerce/pull/54984/commits/95d1af2abc3278e85daf6265609de6841fe1ee14)

kalessil (approved):
LGTM!

:heavy_check_mark: Followed the testing instructions - works as intended
:heavy_check_mark: Tested bulk operations
:heavy_check_mark: Tested random status transitions for individual and multiple comments
:heavy_check_mark: Did product reviews smoke testing - not affected

PR #55003, opened 2025-01-30
[WIP] Add/product details hook extendibility

gigitux:

This is a POC for adding blocks to the Accordion Group block via Block Hooks **only** when it is a child of the Product Details block. Currently, this approach has some limitation.

### Technical Details

The idea is to inject a custom attribute (`__isProductDetailsChildren`) to the Accordion Group block only when the Accordion Group is a child of the Product Details block.

#### Add support for the custom attribute

This code ensures that the accordion group supports the `__isProductDetailsChildren` attribute.

https://github.com/woocommerce/woocommerce/blob/212b6c9ec50c5d0098611922c94a1dea6a1685c6/plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php#L16-L25

#### Inject the custom attribute

When the Product Details block is being rendered, the entire tree is traversed and the accordion group is marked via `__isProductDetailsChildren`:

https://github.com/woocommerce/woocommerce/blob/ebda124a59acbe2813255cd1ab1021a5f5101f8e/plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php#L143-L155

#### Block Hooks

A block is added to the Accordion Group via Block Hooks API via the `hooked_block_types`:

https://github.com/woocommerce/woocommerce/blob/212b6c9ec50c5d0098611922c94a1dea6a1685c6/plugins/woocommerce/src/Blocks/BlockTypesController.php#L66-L79

As I mentioned above, it is necessary that the block is added only when the `accordion-group` is a child of the Product Details.

I was convicted that `parsed_anchor_block` would have the attributes that I inject in the render function of the Product Details block

https://github.com/woocommerce/woocommerce/blob/ebda124a59acbe2813255cd1ab1021a5f5101f8e/plugins/woocommerce/src/Blocks/BlockTypesController.php#L81-L116

but I discovered that the entire block hooks logic run on the post content markup, so the custom attribute will not be available yet:

https://github.com/WordPress/wordpress-develop/blob/6b41380785d2a4d62524800afce23fd921fe214f/src/wp-includes/block-template-utils.php#L635-L639

### First constraint

This means that if we follow this approach (which appears to be the only option), we must ensure that:  

- In patterns, the HTML markup includes the `__isProductDetailsChildren` attribute.  
- In the editor, when the accordion group is added inside the Product Details block, the attribute is manually added via JavaScript.

### Second constraint

Checking the documention, it looks like that the supported positions are: before, after, first_child, or last_child. It looks like that I can specify a "specific index".

For instance, in the case of the Accordion Group by default, we will have three accordions: Description, Additional Info and Reviews. It seems like that an extender can't easily inject a new accordion item between the Description and the Additional Info accordion item.

1.
2.
3.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Blocks/BlockTypesController.php
```diff
@@ -63,6 +63,57 @@ protected function init() { // phpcs:ignore WooCommerce.Functions.InternalInject
...
+					);
+					return parse_blocks( $markup_html )[0];
+
+				}
+
+				return $parsed_hooked_block;
+			},
+			50,
+			5
+		);
```
gigitux: This is a POC to add an  `accordion-item` to the `accordion-group`

plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
```diff
@@ -13,6 +13,31 @@ class BlockifiedProductDetails extends AbstractBlock {
...
 
+	public function add_block_type_metadata( $metadata ) {
+		if ( $metadata['name'] === 'woocommerce/accordion-group' ) {
+			$metadata['attributes']['__isProductDetailsChildren'] = array(
+				'type'    => 'boolean',
+				'default' => false,
+			);
+			return $metadata;
+		}
+		return $metadata;
```
gigitux: Updating the attributes of `woocommerce/accordion-group` to support injecting a custom attribute, enabling filtering of instances that are children of the Product Details block.

plugins/woocommerce/src/Blocks/BlockTypesController.php
```diff
@@ -63,6 +63,57 @@ protected function init() { // phpcs:ignore WooCommerce.Functions.InternalInject
...
+				$context
+			) {
+
+				// Has the hooked block been suppressed by a previous filter?
+				if ( is_null( $parsed_hooked_block ) ) {
+					return $parsed_hooked_block;
+				}
+
+				// Only apply the updated attributes if the block is hooked after a Post Content block.
+				if ( 'woocommerce/accordion-group' === $parsed_anchor_block['blockName'] && 'last_child' === $relative_position ) {
```
gigitux: I want to check about the presence of the `__isProductDetailsChildren` to ensure that the accordion item is added in right accordion-group block.

tjcafferkey:
A few thoughts:

**Attribute meta**
I am not suggesting this is better, but we do have the option of setting the `__isProductDetailsChildren` property (to follow convention we add `"wcDescendantOf": "woocommerce/product-details"` on the blocks metadata key. For example look how `ignoredHookedBlocks` is tracked on blocks:

```html
<!-- wp:navigation {
    "ref": 4,
    "overlayBackgroundColor": "base",
    "overlayTextColor": "contrast",
    "metadata": {
        "ignoredHookedBlocks": [
            "woocommerce/customer-account"
        ]
    },
    "layout": {
        "type": "flex",
        "justifyContent": "right", 
        "flexWrap": "wrap"
    }
} /-->
```

**Inner Blocks template**

To ensure this attribute is always set on the Accordion block every time the user inserts the Product Details block can we not use an inner blocks template for the Product Details block and define its usage here, [similar to how the Product Collection does this](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce-blocks/assets/js/blocks/product-collection/constants.ts#L121-L164)?

We could also hook into `rest_pre_insert_wp_template` ([ref](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/rest-api/endpoints/class-wp-rest-templates-controller.php#L656)) to check the user hasn't deleted it by accident, and re-add it if so. For context this is where/how we add the `ignoredHookedBlocksMeta` attribute on blocks.

**Updating the attribute value**

I don't feel like we should be dealing with HTML markup to update attributes. Can we not do something [similar to CustomerAccount.php](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Blocks/BlockTypes/CustomerAccount.php#L68-L84)?

gigitux:
Thanks, @tjcafferkey, for the review and for the chat about this! I’ll share some final thoughts after our discussion.

> **Attribute meta**  

Agreed! I'll update it.  

> **Inner Blocks template**  

My main concern was theme developers creating patterns—if they don’t include the custom attribute in the markup, extensibility won't work. As you noted, Core relies on metadata like `ignoredHookedBlocks` for proper extensibility. So, it looks like a false concern 😁.

Instead, for Site Editor use case:
- when the user removes and re-adds the Product Details block, we can rely on the inner blocks template approach to ensure that the attribute is present.
- when the user removes and re-add the Accordion group inside the Product Details block, we can add some custom JS code to ensure that the attribute is always present.

> **Updating the attribute value**  

After the foundational PR, we may introduce utility functions to simplify this for third-party developers.  



In the following days, I will update the PR and ask you for another review!

gigitux:
Thanks for your feedback! After further investigation, I discovered that this approach isn't viable. For more details, you can check this GitHub Discussion: https://github.com/woocommerce/woocommerce/discussions/55244

PR #54934, opened 2025-01-28
[e2e tests] Fix Playwright projects dependencies

adimoldovan:
### Changes proposed in this Pull Request:

Fixes running e2e tests with config files different than the default one.
https://github.com/woocommerce/woocommerce/pull/54628 introduced projects dependencies but broke the projects setup in the env config files. As a result, running tests for environments other than default ran all the tests instead of only the ones configured to run in the environment config file.

Other changes:
- added convenience npm scripts to run the most common e2e tests suites
- improved the `run-tests-with-env.sh` script for more useful and prettier console logging

### How to test the changes in this Pull Request:

Run the tests for different environments. Make sure only the tests configured in the environment playwright config are running.

Examples:

Should run all the tests in all the projects defined in the default-wpcom playwright config file
```
pnpm test:e2e:wpcom
```

Should only run the tests in the `e2e-wpcom` project defined in the default-wpcom playwright config file
```
pnpm test:e2e:wpcom --project=e2e-wpcom
```

Tested with (any test failures are not related to this PR):
- [gutenberg jobs](https://github.com/woocommerce/woocommerce/actions/runs/13017301226)
- [e2e hpos disabled](https://github.com/woocommerce/woocommerce/actions/runs/13017289720)
- [api tests hpos disabled](https://github.com/woocommerce/woocommerce/actions/runs/13017478293)
- [pressable](https://github.com/woocommerce/woocommerce/actions/runs/13017270848)
- [wpcom](https://github.com/woocommerce/woocommerce/actions/runs/13017265403)
- [woocommerce-payments](https://github.com/woocommerce/woocommerce/actions/runs/13015756541)
- [woocommerce-paypal-payments](https://github.com/woocommerce/woocommerce/actions/runs/13015767061)
- [services](https://github.com/woocommerce/woocommerce/actions/runs/13015776237)

PR #55013, opened 2025-01-30
[e2e tests] Remove left over Playwright project from main config file

adimoldovan:

Remove the HPOS disabled Playwright project from the main config file. It was supposed to be removed by https://github.com/woocommerce/woocommerce/pull/54934 but it was missed.

Not much to test, if CI is green we should be good.


Stojdza (approved):
LGTM 👍

PR #9606, opened 2015-11-15
Shop order search allowed via payment method

shivapoudel:
If shop manager wants to search order via payment methods then they can simply search the payment method :P

PR #55023, opened 2025-01-30
Prevent quantity change notice showing on mini cart when adding items

### LINKED ISSUE #42227: "The quantity of "product" was changed to 6" notice overlaps the mini cart on mobile and desktop
## Describe the bug

From the catalog page, using the Products (beta) block, when adding a product to the cart that was already in the cart, a notice banner pops up saying that the quantity of the product was changed. Unfortunately it overlaps with the mini cart.

## To reproduce

1. Enable the catalog blockified template
2. Add the Products (beta) block
3. Enable the mini cart to auto-open after add to cart
4. Add a product to the cart
5. close the mini cart
6. add the same product to the cart again
7. A notice is shown both on mobile and on desktop overlapping the mini cart

## Expected behavior
No matter the device, there shouldn't be any notice showed in the context of the Mini-Cart block when users modify the cart contents.

## Screenshots

https://github.com/woocommerce/woocommerce-blocks/assets/47855768/ab0d7568-d6f8-490a-a52a-c7d60ffd1635

https://github.com/woocommerce/woocommerce-blocks/assets/47855768/d6c35f26-446a-4131-8ec7-7be45c709159

opr:

- Add a new part of the cart data store called `productsPendingAdd`, this will be used to track whether the cart is currently in the process of adding an item. We have to use `product` here rather than `item` as a product only becomes an item _after_ it has been added to the cart.
- Update `notifyIfQuantityChanged` so it checks which items are pending add, if the product Id is being added, skip announcing the change.
- Removed two unused action types: `TRIGGER_ADDED_TO_CART_EVENT`, and `TRIGGER_ADDING_TO_CART_EVENT`
- Create thunks to handle dispatching the added to cart and adding to cart events, I did this because it seems smart to fire off "side effects" (setting the pending products, and dispatching the events) all together in one place.
- Added E2E and unit tests to cover these changes.

Note: the notices will still show if item quantity changes _unexpectedly_ - see https://github.com/woocommerce/woocommerce/issues/42227#issuecomment-2580942196

> These notices are only supposed to display if the quantity of an item updates unexpectedly. The way they always display in the Mini-cart is a bug.
>
> To illustrate the desired use of these, I have an example: Imagine a merchant has a rule requiring customers to purchase item A in pairs with item B (2:1 ratio). If a shopper has:
>
> 10 units of item A
> 5 units of item B
>
> When they increase item B to 6 units, the system automatically adjusts item A to 12 units to maintain the required ratio. Since the shopper didn't manually change item A's quantity, the notice alerts them to this automatic adjustment.

Closes #42227

1. Add a Product Collection Block to a page on your site.
2. Go to the site editor and add the mini cart to the same page if using a classic theme, or the header template part if using a block theme.
3. In the mini cart settings, enable the `Open drawer when adding` setting.
4. Go to the front end of the site and visit the page with the Product Collection Block and add an item to your cart. 
5. Verify the mini cart opens and there is no notice saying "The quantity of X was changed to Y"
6. Close the mini cart and add the same item, verify the same.

#### Developer testing

1. Go to the Product Collection Block on the front end. Add an item to your cart. Remember the ID of the item you just added.
2. **Do not** reload the page after the next step, it is not needed and will stop the test working correctly.
3. Add this snippet to your site. Change the `$product_id` to the one you already have in your cart.
4. Add a different item to the cart.
5. When the mini cart opens, ensure you **do** see the "The quantity of X was changed to Y" notice.

```php
add_action('woocommerce_before_calculate_totals', 'sync_cart_quantities', 20, 1);

function sync_cart_quantities(WC_Cart $cart) {
	$product_id = 623;

	foreach ( $cart->get_cart_contents() as $cart_item_key => $cart_item) {
		if ($cart_item['product_id'] == $product_id) {
			$cart->set_quantity($cart_item['key'], 10, false);
		}
	}
	return $cart;
}
```

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce-blocks/assets/js/data/cart/default-state.ts
```diff
@@ -29,10 +29,12 @@ import {
 
 const EMPTY_PENDING_QUANTITY: [] = [];
 const EMPTY_PENDING_DELETE: [] = [];
+const EMPTY_PENDING_ADD: [] = [];
 
 export interface CartState {
 	cartItemsPendingQuantity: string[];
 	cartItemsPendingDelete: string[];
+	productsPendingAdd: number[];
```
opr, line 37: Cart item keys are strings, but during add to cart, we add the _product id_ which is a number

plugins/woocommerce-blocks/assets/js/data/cart/default-state.ts
```diff
@@ -57,6 +59,7 @@ Object.keys( defaultFields ).forEach( ( key ) => {
 export const defaultCartState: CartState = {
 	cartItemsPendingQuantity: EMPTY_PENDING_QUANTITY,
 	cartItemsPendingDelete: EMPTY_PENDING_DELETE,
+	productsPendingAdd: EMPTY_PENDING_ADD,
```
opr, line 62: Again, differentiating between _cart items_ (what products become when they are _in_ the cart) vs. _products_ (a product in the store that shoppers add to their cart)

plugins/woocommerce-blocks/assets/js/data/cart/action-types.ts
nielslange, line 1: Kudos for removing the following entries:
```ts
TRIGGER_ADDING_TO_CART_EVENT: 'TRIGGER_ADDING_TO_CART_EVENT',
TRIGGER_ADDED_TO_CART_EVENT: 'TRIGGER_ADDED_TO_CART_EVENT',
```

While checking the other entries, I noticed that the following one is also no longer used:
```ts
SET_FULL_SHIPPING_ADDRESS_PUSHED: 'SET_FULL_SHIPPING_ADDRESS_PUSHED',
```
We removed the corresponding actions, selectors and reducer 2 years ago with #9606.

plugins/woocommerce-blocks/assets/js/data/cart/test/notify-quantity-changes.ts
```diff
@@ -95,6 +95,18 @@ describe( 'notifyQuantityChanges', () => {
 		} );
 		expect( mockedCreateInfoNotice ).not.toHaveBeenCalled();
 	} );
+	it( 'does not show notices when the item is being added', () => {
+		const { oldCart, newCart } = getFreshCarts();
+		newCart.items[ 0 ].quantity = 5;
+		newCart.items[ 0 ].quantity_limits.maximum = 10;
```
nielslange: Given that we're not testing against the maximum allowed quantity limit, do we need this line here?

nielslange (approved):
Thanks for working on this, @opr. I've successfully tested the changes and the code looks good. I left a few notes, but nothing critical. Thus, I'm pre-approving this PR.

plugins/woocommerce-blocks/assets/js/data/cart/test/notify-quantity-changes.ts
opr: Good point, done!

plugins/woocommerce-blocks/assets/js/data/cart/action-types.ts
opr: Nice spot, done!

PR #55024, opened 2025-01-30
[Experimental] Fix /wp-json/wc/store/v1/products/{productId} API to return the grouped products ids when the product type is grouped

### LINKED ISSUE #54784: Fix /wp-json/wc/store/v1/products/{productId} API to return the grouped products ids when the product type is grouped
**Description**

The Grouped Product Selector block from the Add to Cart uses the `product` from the `/wp-json/wc/store/v1/products/{productId}` endpoint to render the grouped products. The problem is that this endpoint does not returns the grouped products ids or the product children.

**ACs** 

- [x] Fix `/wp-json/wc/store/v1/products/{productId}` API to return the grouped products ids when the product type is grouped.
- [x] Change the Grouped Product Selector -> Product Item Template block to use the grouped product ids returned by the endpoint to render them in the Add to Cart block.
- [x] Make sure this works correctly for both SingleProductBlock and SingleProductTemplate

mdperez86:

Closes #54784

1. Install and activate the WooCommerce Beta Tester plugin
2. Go to WooCommerce Admin Test Helper > Features and enable experimental-blocks and blockified-add-to-cart.
3. Go to Appearance > Editor > Templates > Single Product, select the Add to Cart with Options block and update it to the blockified one.
4. Select the Add to Cart block and choose the Grouped product type from the product type selector.
5. Since in this template the add to cart block sets a mock product then the grouped products listed in the form are the first 3 returned by the products endpoint.
6. Then create a new post
7. Add the single product block
8. Select a grouped product
9. Make sure the grouped products listed in the form are the right ones associated with the parent product.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/src/StoreApi/Schemas/V1/ProductSchema.php
+++ b/plugins/woocommerce/src/StoreApi/Schemas/V1/ProductSchema.php
@@ -362,6 +362,17 @@ class ProductSchema extends AbstractSchema {
 					],
 				],
 			],
+			'grouped_products'    => [
+				'description' => __( 'List of grouped product IDs, if applicable.', 'woocommerce' ),
+				'type'        => 'array',
+				'context'     => [ 'view', 'edit' ],
+				'items'       => [
+					'description' => __( 'List of grouped product ids.', 'woocommerce' ),
+					'type'        => 'integer',
+					'context'     => [ 'view', 'edit' ],
+					'readonly'    => true,
+				],
+			],
 			'has_options'         => [
 				'description' => __( 'Does the product have additional options before it can be added to the cart?', 'woocommerce' ),
 				'type'        => 'boolean',
@@ -496,6 +507,7 @@ class ProductSchema extends AbstractSchema {
 			'tags'                => $this->get_term_list( $product, 'product_tag' ),
 			'attributes'          => $this->get_attributes( $product ),
 			'variations'          => $this->get_variations( $product ),
+			'grouped_products'    => $this->get_grouped_products( $product ),
 			'has_options'         => $product->has_options(),
 			'is_purchasable'      => $product->is_purchasable(),
 			'is_in_stock'         => $product->is_in_stock(),
@@ -689,6 +701,16 @@ class ProductSchema extends AbstractSchema {
 	}
 
 	/**
+	 * Get grouped product IDs.
+	 *
+	 * @param \WC_Product $product Product instance.
+	 * @return array
+	 */
+	protected function get_grouped_products( \WC_Product $product ) {
+		return $product->is_type( ProductType::GROUPED ) ? $product->get_visible_child_ids() : [];
+	}
+
+	/**
 	 * Get list of product attributes and attribute terms.
 	 *
 	 * @param \WC_Product $product Product instance.
@@ -826,7 +848,7 @@ class ProductSchema extends AbstractSchema {
 		}
 
 		if ( $product->is_type( ProductType::GROUPED ) ) {
-			$children       = array_filter( array_map( 'wc_get_product', $product->get_children() ), 'wc_products_array_filter_visible_grouped' );
+			$children       = $product->get_visible_children();
 			$price_function = 'incl' === $tax_display_mode ? 'wc_get_price_including_tax' : 'wc_get_price_excluding_tax';
 
 			foreach ( $children as $child ) {
--- a/plugins/woocommerce/src/StoreApi/Schemas/V1/ProductSchema.php
+++ b/plugins/woocommerce/src/StoreApi/Schemas/V1/ProductSchema.php
@@ -707,7 +707,15 @@ class ProductSchema extends AbstractSchema {
 	 * @return array
 	 */
 	protected function get_grouped_products( \WC_Product $product ) {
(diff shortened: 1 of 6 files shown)


Aljullu (commented):
This looks good to me. I'm a bit wary of introducing new methods into `WC_Product_Grouped`. For example, I think `get_visible_child_ids()` would be better implemented at the consumer level instead of becoming a method in the product class, but not a strong opinion. I think before merging it might be good to reach out to Proton in case they want to take a look given that this is a class that impacts several parts of WooCommerce. But overall, the approach taken in this PR looks good to me. :+1:

mdperez86:
> This looks good to me. I'm a bit wary of introducing new methods into `WC_Product_Grouped`. For example, I think `get_visible_child_ids()` would be better implemented at the consumer level instead of becoming a method in the product class, but not a strong opinion. I think before merging it might be good to reach out to Proton in case they want to take a look given that this is a class that impacts several parts of WooCommerce. But overall, the approach taken in this PR looks good to me. 👍

I was following a similar pattern made in -> https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/class-wc-product-variable.php#L245

plugins/woocommerce/src/StoreApi/Schemas/V1/ProductSchema.php
```diff
@@ -362,6 +362,17 @@ public function get_properties() {
 					],
 				],
 			],
+			'grouped_products'    => [
```
Aljullu, line 365: Wondering if this property should be called `children` instead of `grouped_products`. I understand that the Grouped product is the parent product and the children are the individual products contained by the grouped product, no?

Aljullu (commented):
The code looks good to me. I left one minor comment about changing the property name in the endpoint, but not a big deal.

> I was following a similar pattern made in -> https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/class-wc-product-variable.php#L245

I wasn't aware of that, thanks for sharing!

In that case, I agree it makes sense having `get_visible_children()` as a method in the `WC_Product_Grouped` to be aligned with `WC_Product_Variable`. I'm not sure about `get_visible_child_ids()`, though. I feel that if we add this method, somebody in the future might want a similar method to retrieve the children's names, prices, etc. so we would end up with many methods all returning the same data in a different shape. On the other hand, using `get_visible_children()` and getting the IDs from there should be quite straightforward for consumers. What do you think?

---

Btw, I noticed that the endpoint returns the products in this form `"grouped_products":{"1":66,"2":67},`, which JS considers to be an object, not an array. This causes the child products not to be shown when selecting a Grouped product in the _Single Product_ block in a post. Can you reproduce it too?

imatge

albarin:
@mdperez86 @Aljullu
After talking a bit about it with the team, we agree with what @Aljullu was saying:
- It's okay to have the `get_visible_children` (similar to what the `WC_Product_Variable` class has).
- It's preferred not to have the `get_visible_child_ids` in the `WC_Product_Grouped` class.

Thank you both!

Aljullu (commented):
> Btw, I noticed that the endpoint returns the products in this form `"grouped_products":{"1":66,"2":67},`, which JS considers to be an object, not an array. This causes the child products not to be shown when selecting a Grouped product in the _Single Product_ block in a post. Can you reproduce it too?

Strangely, I can only reproduce this with one Grouped product in my testing store, but can't reproduce it with newly created ones, so it's likely I had something broken on my dev site, sorry for the confusion!

Once the comment from @albarin has been addressed, this LGTM.

mdperez86:
> In that case, I agree it makes sense having `get_visible_children()` as a method in the `WC_Product_Grouped` to be aligned with `WC_Product_Variable`. I'm not sure about `get_visible_child_ids()`, though. I feel that if we add this method, somebody in the future might want a similar method to retrieve the children's names, prices, etc. so we would end up with many methods all returning the same data in a different shape. On the other hand, using `get_visible_children()` and getting the IDs from there should be quite straightforward for consumers. What do you think?

done -> [da7e70d](https://github.com/woocommerce/woocommerce/pull/55024/commits/da7e70dfc49fb3685f52e71a4835c60e8756d547)

Aljullu (approved):
LGTM, good job, @mdperez86! :raised_hands:

plugins/woocommerce/tests/php/src/Blocks/StoreApi/Routes/Products.php
```diff
@@ -66,12 +67,31 @@ public function test_get_item() {
...
 	}
 
+	/**
+	 * Test get grouped product.
+	 */
+	public function test_grouped_product() {
+		$response = rest_get_server()->dispatch( new \WP_REST_Request( 'GET', '/wc/store/v1/products/' . $this->products[2]->get_id() ) );
+		$data     = $response->get_data();
+
+		$grouped_product_ids = $this->products[2]->get_visible_child_ids();
```
Aljullu: This test will need to be updated now that `get_visible_child_ids()` has been removed.

PR #55028, opened 2025-01-31
Fix additional checkout field not saved through edit account page form

### LINKED ISSUE #52066: woocommerce_register_additional_checkout_field : "contact" additional field not saved through "edit-account" page form.
### Prerequisites

### Describe the bug

When an additional field is added with "woocommerce_register_additional_checkout_field" **in the "contact" location** this field can't be modified through the "my account -> edit account" page. Only **field edition through order checkout works**.

### Expected behavior

When validating the "my account -> edit account" form page, the field should be modified in db, as suggested by the message on form validation.

### Actual behavior

The field keeps its old value, is never written in db.

### Steps to reproduce

Followed the (awesome) doc [here](https://developer.woocommerce.com/docs/cart-and-checkout-additional-checkout-fields/#3-using-the-api):

Add this code to the theme's functions.php (or everywhere else is the same):
```
add_action(
	'woocommerce_init',
	function() {
			array(
				'id'            => 'mynamespace/new_field_id',
				'label'         => 'new field',
				'location'      => 'contact',
			),
		);
	}
);
```
It will make the field input to appear in the "my account" page and in the order checkout, but it is read-only through the my-account page.
Add the same field to the "address" location works.

```
### WordPress Environment ###

WC Version: 9.3.3
Legacy REST API Package Version: L’extension Legacy REST API n’est pas installée sur ce site.
Action Scheduler Version: ✔ 3.8.1
Log Directory Writable: ❌ Pour autoriser la connexion
faites en sorte que /var/www/wptest/wp-content/uploads/wc-logs/ soit accessible en écriture.

WP Version: 6.6.2
WP Multisite: –
WP Memory Limit: 256 Mo
WP Debug Mode: –
WP Cron: ✔
Language: fr_FR
External object cache: –

### Server Environment ###

Server Info: Apache/2.4.41 (Ubuntu)
PHP Version: 8.2.4
PHP Post Max Size: 100 Mo
PHP Time Limit: 30
PHP Max Input Vars: 1000
cURL Version: 7.68.0
OpenSSL/1.1.1f

SUHOSIN Installed: –
MySQL Version: 10.3.39-MariaDB-0ubuntu0.20.04.2
Max Upload Size: 100 Mo
Default Timezone is UTC: ✔
fsockopen/cURL: ✔
SoapClient: ❌ Votre serveur n’a pas la classe SoapClient activée - certaines extensions de passerelle utilisant SOAP peuvent ne pas fonctionner correctement.
DOMDocument: ✔
GZip: ✔
Multibyte String: ✔
Remote Post: ✔
Remote Get: ✔

### Database ###

[REDACTED]

### Post Type Counts ###

attachment: 1
page: 7
post: 2
revision: 1
wp_navigation: 1

### Security ###

Secure connection (HTTPS): ❌
					Votre boutique n’utilise pas HTTPS. En savoir plus sur HTTPS et les certificats SSL.
Hide errors from visitors: ✔

### Active Plugins (1) ###

WooCommerce: par Automattic – 9.3.3

### Inactive Plugins (2) ###

Akismet Anti-spam: Spam Protection: par Automattic - Anti-spam Team – 5.3.3
Hello Dolly: par Matt Mullenweg – 1.7.2

### Settings ###

Legacy API Enabled: –
Force SSL: –
Currency: EUR (€)
Thousand Separator: 
Decimal Separator: ,
Number of Decimals: 2
grouped (grouped)
simple (simple)
variable (variable)

featured (featured)
outofstock (outofstock)
rated-1 (rated-1)
rated-2 (rated-2)
rated-3 (rated-3)
rated-4 (rated-4)
rated-5 (rated-5)

HPOS feature enabled: ✔
HPOS data sync enabled: –

### Logging ###

Enabled: ✔
Retention period: 30 jours
Level threshold: –
Log directory size: 0 o

### WC Pages ###

Base de la boutique: #7 - /index.php/boutique/
Panier: #8 - /index.php/panier/ -  Contient le bloc woocommerce/cart
Commander: #9 - /index.php/commander/ -  Contient le bloc woocommerce/checkout
Mon compte: #10 - /index.php/mon-compte/
Conditions générales de vente et d’utilisation: ❌ Page non définie

### Theme ###

Name: Twenty Twenty-Four
Version: 1.2
Author URL: https://fr.wordpress.org
Child Theme: ❌ – Si vous modifiez WooCommerce sur un thème parent que vous n‘avez pas construit personnellement
nous vous recommandons d‘utiliser un thème enfant. Voir : Comment créer un thème enfant

WooCommerce Support: ❌ Non déclaré

### Templates ###

Overrides: /var/www/wptest/wp-content/plugins/woocommerce/templates/block-notices/error.php
/var/www/wptest/wp-content/plugins/woocommerce/templates/block-notices/notice.php
/var/www/wptest/wp-content/plugins/woocommerce/templates/block-notices/success.php


### Admin ###

analytics
product-block-editor
coupons
core-profiler
customize-store
customer-effort-score-tracks
import-products-task
shipping-smart-defaults
shipping-setting-tour
homescreen
marketing
mobile-app-banner
navigation
onboarding
onboarding-tasks
product-custom-fields
remote-inbox-notifications
remote-free-extensions
payment-gateway-suggestions
printful
shipping-label-banner
subscriptions
store-alerts
transient-notices
woo-mobile-welcome
wc-pay-promotion
wc-pay-welcome-page
launch-your-store

Disabled Features: experimental-blocks
minified-js
product-pre-publish-modal
settings
blueprint

Daily Cron: ✔ Next scheduled: 2024-10-17 17:18:49 +02:00
Options: ✔
Notes: 55
Onboarding: completed

### Action Scheduler ###

Terminé: 2
Oldest: 2024-10-16 19:19:19 +0200
Newest: 2024-10-16 19:19:59 +0200

En attente: 1
Oldest: 2024-10-17 19:19:20 +0200
Newest: 2024-10-17 19:19:20 +0200


Generated at: 2024-10-16 19:22:32 +02:00
```
</details>

### Isolating the problem

wavvves:

Fixes a bug causing additional fields not being persisted when edited through My Account. Also introduces some defensive coding around creating customer objects from user IDs to safeguard against thrown data store exceptions.

Closes #52066 .

1. Add the following snippet
```php
add_action(
	'woocommerce_init',
	function () {
			array(
				'id'       => 'mynamespace/new_field_id',
				'label'    => 'new field',
				'location' => 'contact',
				'required' => false,
                                 'validate_callback' => function () {
					return new WP_Error(
						'generic_validation_error',
						'Generic validation error message'
					);
				},
			),
		);
			array(
				'id'       => 'mynamespace/another_field_id',
				'label'    => 'another field',
				'location' => 'contact',
				'required' => false
			),
		);
	}
);
```
2. Place an order using the checkout block and ensure the declared fields show, can be filled, and are optional.
3. Submit the order with first blank and second filled. Verify they appear on the order confirmation page with the correct value.
4. Go to My Account > Account Details and verify the value submitted during order is pre-filled.
5. Edit both fields and verify the value persists after saving. Change the first one back to blank and save. Verify it worked.
6. Change the snippet above to state `'required' => true` on the second field. Back on account details verify the changed field is now marked as required.
7. Edit the field to be empty and attempt to save.
8. Verify an error notice is shown stating the field is required.
9. Smoke test editing all fields in that screen.
10. Save values to both, and verify that on the checkout block for a new order those values are pre-filled.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
```diff
@@ -217,10 +227,11 @@ public function save_account_form_fields( $user_id ) {
 		$location_validation = $this->checkout_fields_controller->validate_fields_for_location( $field_values, 'contact', 'other' );
 
 		if ( is_wp_error( $location_validation ) && $location_validation->has_errors() ) {
-			wc_add_notice( $location_validation->get_error_message(), 'error' );
+			foreach ( $location_validation->get_error_messages() as $error_message ) {
+				wc_add_notice( $error_message, 'error' );
```
wavvves: @opr Didn't quite figure this one out, do we still want to save customer data after this?

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
```diff
@@ -185,23 +185,33 @@ public function edit_account_form_fields() {
...
+				'notice'
+			);
+			return;
+		}
+
 		$additional_fields = $this->checkout_fields_controller->get_fields_for_location( 'contact' );
-		$field_values      = array();
 
-		foreach ( array_keys( $additional_fields ) as $key ) {
-			$post_key = CheckoutFields::get_group_key( 'other' ) . $key;
```
wavvves, line 195: Bug source was this post_key. The field name on the form isn't prefixed with the group.

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
```diff
@@ -185,23 +185,33 @@ public function edit_account_form_fields() {
...
+			$field_value = wc_clean( wp_unslash( $_POST[ $key ] ?? '' ) ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
+
+			$field_value = $this->checkout_fields_controller->sanitize_field( $key, $field_value );
 			$validation  = $this->checkout_fields_controller->validate_field( $key, $field_value );
 
 			if ( is_wp_error( $validation ) && $validation->has_errors() ) {
-				wc_add_notice( $validation->get_error_message(), 'error' );
+				foreach ( $validation->get_error_messages() as $error_message ) {
+					wc_add_notice( $error_message, 'error' );
+				}
```
wavvves: I know messages come condensed,  but since these are using callbacks that can be custom-defined, we should account for `WP_Error` possibly being used to the full extent, or else risk suppressing error messages. Right?

plugins/woocommerce/includes/class-wc-form-handler.php
```diff
@@ -337,9 +339,9 @@ public static function save_account_details() {
 			wp_update_user( $user );
 
 			// Update customer object to keep data in sync.
-			$customer = new WC_Customer( $user->ID );
```
wavvves, line 340: These can throw data store-related exceptions or WC_Data_Exception validation errors.

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
opr: If there's a validation error, then no

opr (changes requested):
Thanks for the PR Paulo, the main bug is fixed, but unfortunately it fails when trying to save an optional field with an empty value.

Also, please would you write some tests to cover this scenario (Saving optional and required contact fields in the My account area).

wavvves:
> Thanks for the PR Paulo, the main bug is fixed, but unfortunately it fails when trying to save an optional field with an empty value.
> 
> Also, please would you write some tests to cover this scenario (Saving optional and required contact fields in the My account area).

Strangely, I tried saving an empty optional field and it worked. What am I missing? 🤔

I added the tests, good to review again @opr  👍🏼

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
```diff
@@ -186,42 +186,55 @@ public function edit_account_form_fields() {
...
-				wc_add_notice( $validation->get_error_message(), 'error' );
-				continue;
+			$field_values = array();
+
+			foreach ( $additional_fields as $key => $field_data ) {
+				// We can't skip, field might be required.
+				$field_value = wc_clean( wp_unslash( $_POST[ $key ] ?? '' ) ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
+
+				$field_value = $this->checkout_fields_controller->sanitize_field( $key, $field_value );
+				$validation  = $this->checkout_fields_controller->validate_field( $key, $field_value );
```
opr: I think here we should skip this if the field is optional and does not have a value.

plugins/woocommerce/tests/php/src/Blocks/Domain/Services/CheckoutFieldsFrontendTest.php
```diff
@@ -0,0 +1,263 @@
...
+
+		$this->assertCount( 1, $mocked_messages );
+		$this->assertEquals( __( 'Unable to save customer additional fields: ', 'woocommerce' ) . $hash, $mocked_messages[0]['message'] );
+		$this->assertEquals( 'notice', $mocked_messages[0]['type'] );
+	}
+
+	/**
+	 * @testDox Contact additional field validation error
+	 */
+	public function test_save_account_form_fields_contact_generic_validation_error() {
```
opr, line 125: I think we should add another test where we check that an optional field with validation, submitted with an empty value can be saved correctly.

opr (changes requested):
@wavvves thanks for the changes, the tests look good, but there is one case to add.

The field I tested had additional validation too, which I think should only run if the field has a value, it doesn't run on checkout if there's no value. Try with the below.

```php
		array(
			'id'                => 'plugin-namespace/alt-email',
			'label'             => 'Alternative Email',
			'required'          => false,
			'sanitize_callback' => function ( $field_value ) {
				return sanitize_email( $field_value );
			},
			'validate_callback' => function ( $field_value ) {
				if ( ! is_email( $field_value ) ) {
					return new \WP_Error( 'invalid_alt_email', 'Please ensure your alternative email matches the correct format.' );
				}
			},
		)
	);
```

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
wavvves: I get it now, validation on optional should not run on empty submissions. Nice catch!

wavvves: Submitted changes and ready for another review 👍🏼

plugins/woocommerce/tests/php/src/Blocks/Domain/Services/CheckoutFieldsFrontendTest.php
wavvves: Added 👍🏼

opr (changes requested):
Thanks, it doesn't error anymore, but submitting an empty value for the optional field doesn't delete it, it just defaults back to the previous value, are you seeing the same?

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
```diff
@@ -186,42 +186,60 @@ public function edit_account_form_fields() {
...
-				wc_add_notice( $validation->get_error_message(), 'error' );
-				continue;
+			$field_values = array();
+
+			foreach ( $additional_fields as $key => $field_data ) {
+				// We can't skip, field might be required.
+				$field_value = trim( wc_clean( wp_unslash( $_POST[ $key ] ?? '' ) ) ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
+
+				if ( 0 === strlen( $field_value ) && ! ( $field_data['required'] ?? false ) ) {
+					continue; // Skip, otherwise validation might prevent saving empty optional fields.
```
opr, line 203: I think this would be the solution for saving empty optional fields.
```suggestion
					$field_values[ $key ] = '';
					continue; // Skip, otherwise validation might prevent saving empty optional fields.
```

wavvves: Thanks @opr, indeed this was missing. I also updated tests to check for saved values wherever I saw fit, so this would be caught!

plugins/woocommerce/tests/php/src/Blocks/Domain/Services/CheckoutFieldsFrontendTest.php
```diff
@@ -0,0 +1,281 @@
...
+
+		$this->assertCount( 1, $mocked_messages );
+		$this->assertEquals( 'Generic validation error message', $mocked_messages[0]['message'] );
+		$this->assertEquals( 'error', $mocked_messages[0]['type'] );
+
+		$value = $this->controller->get_field_from_object( 'mynamespace/generic_validation_error', new WC_Customer( 1 ) );
+		$this->assertEquals( '', $value );
+
+		$_POST['mynamespace/generic_validation_error'] = '';
+		$mocked_messages                               = [];
```
opr, line 156: It might be a good idea to save this with a valid value first, then save it again with an empty value. This is so verify that going from valid -> empty works correctly. Currently we're only testing empty -> empty works, right?

opr (changes requested):
I am testing this again and this setup isn't working:

- Have an optional field with custom validation
- Fill with invalid data
- Expect invalid data _not_ to be saved

In the above, I see two errors for the same thing and my invalid data is saved to the field.
image

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
```diff
@@ -186,42 +186,61 @@ public function edit_account_form_fields() {
...
+			$additional_fields = $this->checkout_fields_controller->get_fields_for_location( 'contact' );
 
-			if ( is_wp_error( $validation ) && $validation->has_errors() ) {
-				wc_add_notice( $validation->get_error_message(), 'error' );
-				continue;
+			$field_values = array();
+
+			foreach ( $additional_fields as $key => $field_data ) {
+				// We can't skip, field might be required.
+				$field_value = trim( wc_clean( wp_unslash( $_POST[ $key ] ?? '' ) ) ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
```
mikejolley: `wc_clean` uses `sanitize_text_field`. The later already trims for us. This is not necccessary.

mikejolley: `$_POST[ $key ]` is wrong. It doesn't match the form field because the form field `name` is prefixed with the `group`. e.g. `CheckoutFields::get_group_key( $group ) . $key`

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
```diff
@@ -186,42 +186,61 @@ public function edit_account_form_fields() {
...
+
+				if ( 0 === strlen( $field_value ) && ! ( $field_data['required'] ?? false ) ) {
+					$field_values[ $key ] = '';
+					continue; // Skip, otherwise validation might prevent saving empty optional fields.
+				}
+
+				$validation = $this->checkout_fields_controller->validate_field( $key, $field_value );
+				// Sanitize should always be after validation to ensure char stripping doesn't affect validation.
+				$field_value = $this->checkout_fields_controller->sanitize_field( $key, $field_value );
```
mikejolley, line 208: Why bother with sanitization before checking if validation worked? I suspect this approach is incorrect also. Validation runs after sanitize in API.

plugins/woocommerce/includes/class-wc-form-handler.php
```diff
@@ -5,6 +5,8 @@
  * @package WooCommerce\Classes\
  */
 
+use Automattic\WooCommerce\Blocks\Domain\Services\CheckoutFields;
+use Automattic\WooCommerce\Blocks\Package;
```
mikejolley, line 9: These 2 imports are unused

mikejolley:
Testing instructions have issues; `validate_callback` on the first field should trigger an error whenever you try to set a value. The test instructions tell you to fill this field which won't work. I'm fixing this in my PR.

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
wavvves: I think validation should run before sanitization, as validation will expose errors whereas sanitizing just ensures data is safe to use/store. There seems to be some  cases around:
https://github.com/woocommerce/woocommerce/blob/7d74306c34520d5683d4a76e761146c5557cff75/plugins/woocommerce/includes/class-wc-checkout.php#L898
Also, developer.wordpress.org [states it should be done as early as possible](https://developer.wordpress.org/apis/security/data-validation/), and escape as late as possible, not being too specific where would sanitize land 🤷
If we're running validation to check allowed chars after sanitizing those chars, will it not always pass? I had issues with this recently after sanitizing first 😅

plugins/woocommerce/includes/class-wc-form-handler.php
wavvves: Strangely my IDE doesn't flag the Package one, but I don't see where it's being used 🤷

plugins/woocommerce/tests/php/src/Blocks/Domain/Services/CheckoutFieldsFrontendTest.php
wavvves: cc/ @mikejolley since this was merged to your PR and going to be closed now.

wavvves:
Closing this and it was incorporated into https://github.com/woocommerce/woocommerce/pull/55047

PR #55042, opened 2025-01-31
Product Gallery: simplify and refactor iAPI 

kmanijak:

This PR simplified the Product Gallery frontend logic by:
- replacing `visibleImagesIds` and `dialogVisibleImagesIds` by single set `imageIds`
- removing `elementThatTriggeredDialogOpening` as the focus seems to be preserved after opening and closing the gallery
- when possible, switching from direct `context` access to `state` usage
- cleaning up some actions and context

This PR is not meant to introduce any logic changes.

1. CI should pass
2. Smoke test Product Gallery block:
    - Go to the Editor > Single Product template.
    - Replace the Product Image Gallery block with the Product Gallery (beta) block. 
    - Go to the frontend and play with the gallery, especially on the variable product with multiple images

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
```diff
@@ -30,72 +28,38 @@ const getContext = ( ns?: string ) =>
...
 const disableArrows = (
 	context: ProductGalleryContext,
 	newImageNumber: number
 ) => {
-	const imagesIds = getCurrentImages( context );
 	// One-based index so it ranges from 1 to imagesIds.length.
 	context.disableLeft = newImageNumber === 1;
-	context.disableRight = newImageNumber === imagesIds.length;
+	context.disableRight = newImageNumber === context.imageIds.length;
 };
```
tjcafferkey: I don't like that we're mutating `context` here. Can we make this function return new values and assign them correctly. We also wont need to pass `context`, only `newImageNumber` and `numberOfTotalImages`

tjcafferkey:
```ts
type ArrowState = Pick<ProductGalleryContext, 'disableLeft' | 'disableRight'>;

const getArrowState = (imageNumber: number, totalImages: number): ArrowState => ({
    disableLeft: imageNumber === 1,
    disableRight: imageNumber === totalImages
});
```

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
```diff
@@ -30,72 +28,38 @@ const getContext = ( ns?: string ) =>
...
 
 const selectImage = (
```
tjcafferkey: I think we can optimize this function. I was already planning on a refactor and had something in mind in conjunction with my `getArrowState` updates above:

```ts
type ArrowState = Pick<ProductGalleryContext, 'disableLeft' | 'disableRight'>;

const getArrowState = (imageNumber: number, totalImages: number): ArrowState => ({
    disableLeft: imageNumber === 1,
    disableRight: imageNumber === totalImages
});

const selectImage = (
    context: ProductGalleryContext,
    type: 'prev' | 'next' | 'current' | 'first'
): void => {
    const { selectedImageNumber, imageIds } = state;
    
    const newImageNumber = {
        first: () => 1,
        current: () => getImageIndex() + 1,
        prev: () => Math.max(1, selectedImageNumber - 1),
        next: () => Math.min(imageIds.length, selectedImageNumber + 1),
    }[type]();

    const arrowState = getArrowState(newImageNumber, context.imageIds.length);
    
    Object.assign(context, {
        selectedImageNumber: newImageNumber,
        ...arrowState
    });
};
```

tjcafferkey: Really we should go one further and prevent this function from updating `context` itself since its essentially a utility function. It should return a new object which the action should use to set on `context` in itself:

```ts
type ArrowStateArgs = Pick<ProductGalleryContext, 'disableLeft' | 'disableRight'>;

const getArrowState = (imageNumber: number, totalImages: number): ArrowStateArgs => ({
    disableLeft: imageNumber === 1,
    disableRight: imageNumber === totalImages
});

type SelectImageArgs = ArrowStateArgs & Pick<ProductGalleryContext, 'selectedImageNumber'>;

const selectImage = (
    type: 'prev' | 'next' | 'current' | 'first'
): SelectImageArgs => {
    const { selectedImageNumber, imageIds } = state;
    
    const newImageNumber = {
        first: () => 1,
        current: () => getImageIndex() + 1,
        prev: () => Math.max(1, selectedImageNumber - 1),
        next: () => Math.min(imageIds.length, selectedImageNumber + 1),
    }[type]();

    return {
        selectedImageNumber: newImageNumber,
        ...getArrowState(newImageNumber, imageIds.length)
    };
};
```

tjcafferkey:
I've just left some comments around the `selectImage` function, but also I'm actually unsure on removing the image focus functionality when opening the dialog. I think for stores with a small image set its fine, but I'm thinking about products that may have even 12+ images it then becomes a bit tedious to scroll up/down and find the image you clicking on to zoom.

I understand that it simplifies the code but perhaps we can find ways to optimize the code so its easier to read/reason with?

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
kmanijak: I've addressed your feedback a but differently:
- made `selectImage` more generic and accept specific `newImageNumber`
- considering we have dedicated actions for next/prev/current images, they're responsible for calculating new image number and provide it to generic `selectImage`

Seems much cleaner IMO.
https://github.com/woocommerce/woocommerce/pull/55042/commits/675bd14e6fe5ca2b982e25ed21473f0f758da0a6
https://github.com/woocommerce/woocommerce/pull/55042/commits/4261c2116e722f25c7e9bba18477fa1c08d2634f

Let me know what you think 🙌

kmanijak:
> I've just left some comments around the selectImage function, but also I'm actually unsure on removing the image focus functionality when opening the dialog. I think for stores with a small image set its fine, but I'm thinking about products that may have even 12+ images it then becomes a bit tedious to scroll up/down and find the image you clicking on to zoom.
> I understand that it simplifies the code but perhaps we can find ways to optimize the code so its easier to read/reason with?

As explained elsewhere I didn’t remove the focus in dialog when opening it. The element that triggered opening the dialog was kept in context to regain the focus AFTER closing the dialog and that's what I removed, since the focus is regained on gallery anyway.

Feel free to re-review again 🙌  The biggest part I don't like is [in here](https://github.com/woocommerce/woocommerce/pull/55042/files#diff-4ac5eb03f43c1cd1490c70c631d0dcb3e7dcb81d5cfefa4b8025a3d526454e0aL264) that there's a different `context` in `watchForChangesOnAddToCartForm` callback, probably because of the block this directive is attached to. I don't know how to address it so it could use `selectImage` action. But I don't consider it the highest priority atm.

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
```diff
@@ -30,92 +28,41 @@ const getContext = ( ns?: string ) =>
...
- * - on-page gallery - may display subset of images.
- * - dialog gallery - displays all of the images.
- * Function returns images per current context.
- */
-const getCurrentImages = ( context: ProductGalleryContext ) => {
-	const { isDialogOpen } = context;
-	return context[
-		isDialogOpen ? 'dialogVisibleImagesIds' : 'visibleImagesIds'
-	];
+const getImageIndex = (): number => {
```
tjcafferkey: Feels like this should belong on the state object with the other getters? What do you think?

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
```diff
@@ -30,92 +28,41 @@ const getContext = ( ns?: string ) =>
...
-	const imagesIds = getCurrentImages( context );
-	return imagesIds.indexOf( imageId );
-};
-
-const disableArrows = (
-	context: ProductGalleryContext,
-	newImageNumber: number
-) => {
-	const imagesIds = getCurrentImages( context );
+const getArrowsState = ( imageNumber: number, totalImages: number ) => ( {
```
tjcafferkey, line 31: Same here. Feels like this should belong on the state object with the other getters? What do you think? These are calculating derived values from state. We could then remove the `totalImages` param

tjcafferkey (commented):
This is a nice improvement. Ive just left a couple of comments before approving

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
kmanijak: I did that but then, I couldn't use it in another getter as it's been telling me there's circular dependency... Which I don't get as we derive state from other state properties in some other getters. I left it in such a state: https://github.com/woocommerce/woocommerce/pull/55042/commits/290fa5f02852d76335fee9e2535acf255f36bb9b

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
kmanijak: We can't do that. Getter cannot accept any arguments and we still need to provide "new image number" to calculate the state.

kmanijak:
> This is a nice improvement. Ive just left a couple of comments before approving

@tjcafferkey, I addressed and answered to your comments. Also, I just realised there's more getters to remove so I remove them. https://github.com/woocommerce/woocommerce/pull/55042/commits/fb9bf7734cfe65ad5c5b2b2a2f4f7cfc99ffa3c9

I think we could go ship it and iterate further.

tjcafferkey (approved):
LGTM thanks for replying to the feedback!

PR #55047, opened 2025-01-31
[Experimental] Add additional field validation rule parsing in Store API and frontend forms

### LINKED ISSUE #52066: woocommerce_register_additional_checkout_field : "contact" additional field not saved through "edit-account" page form.

### LINKED ISSUE #54479: Add validation parsing for additional fields in PHP

mikejolley:

Adds handling for validation rules for additional fields for the Store API, and for the account pages. Some document changes were needed to handle contact vs order fields which share the same group.

Closes #54479

Incorporates fixes from https://github.com/woocommerce/woocommerce/pull/55028

Fixes a bug causing additional fields not being persisted when edited through My Account. Also introduces some defensive coding around creating customer objects from user IDs to safeguard against thrown data store exceptions.

Closes #52066

**This feature is behind a feature flag.** Developers can test the following using a dev build `pnpm --filter=@woocommerce/plugin-woocommerce watch:build`

You'll need some fields setup with rules. Here is what I've been using:

<details>

```php
add_action(
	'woocommerce_init',
	function () {
			array(
				'id'       => 'namespace/vat-number',
				'label'    => 'VAT Number',
				'location' => 'address',
				'required' => true,
				'rules'    => array(
					'visible'    => array(
						'cart' => array(
							'properties' => array(
								'billing_address' => array(
									'properties' => array(
										'country' => array(
											'type' => 'string',
											'enum' => WC()->countries->get_european_union_countries( 'eu_vat' ),
										),
									),
								),
							),
						),
					),
					'validation' => array(
						'type'    => 'string',
						'pattern' => '^[A-Z]{2}[0-9A-Z]{2,12}$',
					),
				),
			)
		);
	}
);

add_action( 'woocommerce_init', 'custom_fields_tester_register_custom_checkout_fields' );

function custom_fields_tester_register_custom_checkout_fields() {

	if ( ! function_exists( 'woocommerce_register_additional_checkout_field' ) ) {
		return;
	}

		array(
			'id'       => 'first-plugin-namespace/truck-size-ok',
			'label'    => 'Can a truck fit down your road?',
			'location' => 'address',
			'type'     => 'checkbox',
		)
	);

		array(
			'id'                => 'plugin-namespace/alt-email',
			'label'             => 'Alternative Email',
			'required'          => true,
			'sanitize_callback' => function ( $field_value ) {
				return sanitize_email( $field_value );
			},
			'rules'             => array(
				'validation' => array(
					'type'   => 'string',
					'format' => 'email',
					'not'    => array(
						'const' => array( '$data' => '/customer/billing_address/email' ),
					),
				),
			),
		)
	);

		array(
			'location' => 'contact',
			'required' => true,
			'type'     => 'select',
			'options'  => array(
				array(
					'label' => 'Director',
					'value' => 'director',
				),
				array(
					'label' => 'Engineering',
					'value' => 'engineering',
				),
				array(
				),
				array(
					'label' => 'Other',
					'value' => 'other',
				),
			),
		)
	);

		array(
			'id'       => 'plugin-namespace/mailing-list',
			'label'    => 'Sign up to our mailing list?',
			'location' => 'contact',
			'type'     => 'checkbox',
		)
	);

		array(
			'id'                => 'plugin-namespace/gov-id',
			'label'             => 'Government ID',
			'location'          => 'address',
			'required'          => true,
			'sanitize_callback' => function ( $field_value ) {
				return str_replace( ' ', '', $field_value );
			},
			'rules'             => array(
				'validation' => array(
					'type'    => 'string',
					'pattern' => '^[A-Z0-9]{5}$',
				),
			),
		),
	);

		array(
			'id'       => 'plugin-namespace/confirm-gov-id',
			'label'    => 'Confirm Government ID',
			'location' => 'address',
			'type'     => 'text',
			'required' => true,
			'rules'    => array(
				'validation' => array(
					'type'  => 'string',
					'const' => array( '$data' => '1/plugin-namespace~1gov-id' ),
				),
			),
		),
	);

		array(
			'id'       => 'plugin-namespace/contact-about-order',
			'label'    => 'Can we contact you about your order?',
			'location' => 'address',
			'type'     => 'checkbox',
		),
	);

		array(
			'id'       => 'plugin-namespace/preferred-contact-time',
			'label'    => 'Preferred time to contact (Morning, Afternoon, or Evening)',
			'location' => 'address',
			'type'     => 'select',
			'options'  => array(
				array(
					'label' => 'Morning',
					'value' => 'morning',
				),
				array(
					'label' => 'Afternoon',
					'value' => 'afternoon',
				),
				array(
					'label' => 'Evening',
					'value' => 'evening',
				),
			),
			'rules'    => array(
				'required' => array(
					'customer' => array(
						'properties' => array(
							'address' => array(
								'properties' => array(
									'plugin-namespace/contact-about-order' => array( 'const' => true ),
								),
							),
						),
					),
				),
			),
		),
	);

		array(
			'id'       => 'plugin-namespace/leave-on-porch',
			'label'    => __( 'Please leave my package on the porch if I\'m not home', 'woocommerce' ),
			'location' => 'order',
			'type'     => 'checkbox',
		),
	);

		array(
			'id'       => 'plugin-namespace/location-on-porch',
			'label'    => __( 'Describe where we should hide the parcel', 'woocommerce' ),
			'location' => 'order',
			'type'     => 'text',
			'rules'    => array(
				'visible' => array(
					'checkout' => array(
						'properties' => array(
							'additional_fields' => array(
								'properties' => array(
									'plugin-namespace/leave-on-porch' => array( 'const' => true ),
								),
							),
						),
					),
				),
			),
		)
	);
		array(
			'id'       => 'plugin-namespace/leave-with-neighbor',
			'label'    => __( 'Which neighbor should we leave it with if unable to hide?', 'woocommerce' ),
			'location' => 'order',
			'type'     => 'select',
			'options'  => array(
				array(
					'label' => 'Neighbor to the left',
					'value' => 'left',
				),
				array(
					'label' => 'Neighbor to the right',
					'value' => 'right',
				),
				array(
					'label' => 'Neighbor across the road',
					'value' => 'across',
				),
				array(
					'label' => 'Do not leave with a neighbor',
					'value' => 'none',
				),
			),
			'rules'    => array(
				'visible' => array(
					'checkout' => array(
						'properties' => array(
							'additional_fields' => array(
								'properties' => array(
									'plugin-namespace/leave-on-porch' => array( 'const' => true ),
								),
							),
						),
					),
				),
			),
		)
	);
}
```

</details>

For testing API POSTs to checkout, you can use this data:

<details>

```json
{
  "shipping_address": {
    "first_name": "John",
    "last_name": "Doe",
    "company": "",
    "address_1": "30 Test Road",
    "address_2": "",
    "city": "Testville",
    "state": "CA",
    "postcode": "90210",
    "country": "US",
    "phone": "",
		"namespace/vat-number": "GB232457280",
		"plugin-namespace/gov-id": "12345",
		"plugin-namespace/confirm-gov-id": "12345",
		"plugin-namespace/contact-about-order": false
	},
  "billing_address": {
    "first_name": "Bob",
    "last_name": "Doe",
    "company": "",
    "address_1": "30 Test road",
    "address_2": "",
    "city": "Testville",
    "state": "CA",
    "postcode": "90210",
    "country": "US",
    "phone": "",
    "email": "test@test.com",
    "phone": "",
		"namespace/vat-number": "GB232457280",
		"plugin-namespace/gov-id": "12345",
		"plugin-namespace/confirm-gov-id": "12345",
		"plugin-namespace/contact-about-order": false
  },
  "customer_note": "Hello",
  "create_account": false,
  "payment_method": "bacs",
	"additional_fields": {
		"plugin-namespace\/alt-email": "te2st@test.com",
		"plugin-namespace\/job-function": "engineering"
	}
}
```

</details>

1. Add something to cart via Store API and POST to checkout using the samples ^ above. Trigger validation errors by providing invalid VAT numbers, GOV IDs, setting `plugin-namespace/contact-about-order` to true, or `plugin-namespace\/alt-email` as your billing email. 
2. Add something to cart on the store and repeat the above tests on the actual checkout. Checkout successfully with valida data.
3. Go to My Account > Edit Account and edit contact fields. Confirm validation works. The alt email one is good to test here to confirm the document object has the billing email for context.
4. Go to My Account > Address address and edit billing/shipping fields. Ensure you test invalid and valid data and notices are shown based on that.

Please also check regressions by using the production build `pnpm --filter=@woocommerce/plugin-woocommerce build` and confirming you can checkout even with fields/rules present.

FOR GLOBALSTEP TESTING:

To test the fixes from #55028, add the following snippet:

<details>

```php
add_action(
	'woocommerce_init',
	function () {
			array(
				'id'       => 'mynamespace/new_field_id',
				'label'    => 'new field',
				'location' => 'contact',
				'required' => false,
                                 'validate_callback' => function () {
					return new WP_Error(
						'generic_validation_error',
						'Generic validation error message'
					);
				},
			),
		);
			array(
				'id'       => 'mynamespace/another_field_id',
				'label'    => 'another field',
				'location' => 'contact',
				'required' => false
			),
		);
	}
);
```
</details>

1. Place an order using the checkout block and ensure the declared fields show, can be filled, and are optional.
2. Submit the order with first blank and second filled. Verify they appear on the order confirmation page with the correct value.
3. Go to My Account > Account Details and verify the value submitted during order is pre-filled.
4. Edit field 2 again and verify the value persists after saving. 
5. Change the first field and save. Generic error message shown. Field value is not persisted.
6. Change the snippet above to state 'required' => true on the second field. Back on account details verify the changed field is now marked as required.
7. Edit the field to be empty and attempt to save.
8. Verify an error notice is shown stating the field is required.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/includes/class-wc-form-handler.php
```diff
@@ -132,7 +132,7 @@ public static function save_address() {
 			// Validation: Required fields.
 			if ( ! empty( $field['required'] ) && empty( $value ) ) {
 				/* translators: %s: Field name. */
-				wc_add_notice( sprintf( __( '%s is a required field.', 'woocommerce' ), $field['label'] ), 'error', array( 'id' => $key ) );
```
mikejolley: We had a mixture of `%s is a required field` and `%s is a required field.` so I standardised on one.

plugins/woocommerce/src/Blocks/BlockTypes/Checkout.php
```diff
@@ -165,10 +166,8 @@ protected function get_block_type_script( $key = null ) {
 
 		if ( Features::is_enabled( 'experimental-blocks' ) ) {
 			$checkout_fields = Package::container()->get( CheckoutFields::class );
-			$checkout_schema = Package::container()->get( CheckoutFieldsSchema::class );
-
 			// Load schema parser asynchronously if we need it.
-			if ( $checkout_schema->has_valid_schema( $checkout_fields->get_additional_fields() ) ) {
+			if ( Validation::has_field_with_schema( $checkout_fields->get_additional_fields() ) ) {
```
mikejolley: Renamed for clarity.

plugins/woocommerce/src/Blocks/Domain/Bootstrap.php
```diff
@@ -299,17 +298,10 @@ function ( Container $container ) {
...
-			function () {
-				return new CheckoutFieldsSchema();
-			}
-		);
 		$this->container->register(
 			CheckoutFields::class,
 			function ( Container $container ) {
-				$schema = $container->get( CheckoutFieldsSchema::class );
-				return new CheckoutFields( $container->get( AssetDataRegistry::class ), $schema );
+				return new CheckoutFields( $container->get( AssetDataRegistry::class ) );
```
mikejolley, line 300: I moved the validation functions into a handler so we no longer need this dependency.

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFields.php
```diff
@@ -206,35 +199,25 @@ public function register_checkout_field( $options ) {
...
+				'optionalLabel'              => sprintf( __( '%s (optional)', 'woocommerce' ), $options['label'] ),
+				'location'                   => '',
+				'type'                       => 'text',
+				'hidden'                     => false,
+				'required'                   => false,
+				'attributes'                 => [],
+				'show_in_order_confirmation' => true,
+				'sanitize_callback'          => array( $this, 'default_sanitize_callback' ),
+				'validate_callback'          => array( $this, 'default_validate_callback' ),
+				'rules'                      => [],
```
mikejolley, line 214: Having this defined won't impact production.

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFields.php
```diff
@@ -911,27 +996,6 @@ public function validate_fields_for_location( $fields, $location, $group = 'othe
...
-			 * Pass an error object to allow validation of an additional field.
-			 *
-			 * @param WP_Error $errors  A WP_Error object that extensions may add errors to.
-			 * @param mixed    $fields  List of fields (key value pairs) in this location.
-			 * @param string   $group   The group of this location (shipping|billing|other).
-			 *
-			 * @since 8.6.0
-			 * @deprecated 8.9.0 Use woocommerce_blocks_validate_location_order_fields instead.
-			 */
-			wc_do_deprecated_action( 'woocommerce_blocks_validate_location_additional_fields', array( $errors, $fields, $group ), '8.9.0', 'woocommerce_blocks_validate_location_additional_fields', 'This action has been graduated, use woocommerce_blocks_validate_location_additional_fields instead.' );
```
mikejolley, line 942: This was deprecated but did not not exist pre-deprecation.

plugins/woocommerce/src/StoreApi/Authentication.php
```diff
@@ -22,6 +23,21 @@ public function init() {
 		add_filter( 'rest_pre_serve_request', array( $this, 'send_cors_headers' ), 10, 3 );
 		add_filter( 'rest_allowed_cors_headers', array( $this, 'allowed_cors_headers' ) );
 
+		// If cart has a valid token, override the core session class.
```
mikejolley: These changes were necessary to load cart data early in the request!

mikejolley:
I've manually picked on @senadir to review this since he's familiar with project.

@wavvves Would you mind checking the changes to Cart Token and Authentication class? I've moved the logic earlier during `rest_api_init`. I don't think this is a problem but would like your opinion. This was necessary so the cart was accessible while the request is sanitized/validated.

@opr your e2e and unit tests for additional fields saved a lot of time here thank you. I had to update one test because the error message response is slightly different for address fields—the error is surfaced during `validate_callback` instead of later in the request. Should be ok but just in case you want to confirm.

plugins/woocommerce/src/StoreApi/Authentication.php
mikejolley: cc @wavvves

plugins/woocommerce/tests/php/src/Blocks/StoreApi/Routes/AdditionalFields.php
```diff
@@ -1891,7 +1891,7 @@ public function test_place_order_required_address_field() {
 		$data     = $response->get_data();
 
 		$this->assertEquals( 400, $response->get_status(), print_r( $data, true ) );
-		$this->assertEquals( \sprintf( 'There was a problem with the provided shipping address: %s is required', $label ), $data['message'], print_r( $data, true ) );
+		$this->assertEquals( \sprintf( 'There was a problem with the provided shipping address: %s is required', $label ), $data['data']['params']['shipping_address'], print_r( $data, true ) );
```
mikejolley: cc @opr

opr: I think this is fine, as long as the error message gets to the front end ok

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
```diff
@@ -145,7 +146,7 @@ public function edit_account_form_required_fields( $fields ) {
 
 		foreach ( $additional_fields as $key => $field ) {
 			if ( ! empty( $field['required'] ) ) {
-				$fields[ $key ] = $field['label'];
+				$fields[ CheckoutFields::get_group_key( 'other' ) . $key ] = $field['label'];
```
senadir: Was this like always wrong?

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
```diff
@@ -186,41 +187,41 @@ public function edit_account_form_fields() {
...
+
+		// Generate document object based on POSTed values.
+		$document_object = null;
+
+		if ( Features::is_enabled( 'experimental-blocks' ) ) {
+			$document_object = new DocumentObject(
+				[
+					'customer' => [
+						'additional_fields' => $additional_fields,
+					],
```
senadir: This is the contact fields? did we do that as well on Checkout?

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsSchema/Validation.php
```diff
@@ -0,0 +1,158 @@
...
+				'properties' => [
+					$secondary_key => [
+						'properties' => [
+							$field_id => $field_schema,
+						],
+					],
+				],
+			],
+		];
+	}
```
senadir, line 57: So this means contact fields would be either on checkout or customer or what?

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsSchema/Validation.php
```diff
@@ -0,0 +1,158 @@
...
+		);
+
+		if ( ! $result->hasError() ) {
+			return true;
+		}
+
+		// Generate useful error message.
+		$formatter = new ErrorFormatter();
+		$formatted = $formatter->formatFlat( $result->error() );
+		$errors    = implode( ', ', count( $formatted ) > 2 ? array_slice( $formatted, 2 ) : $formatted );
```
senadir: As discussed let's not return this, have a follow to parse over errors in exchange for hardcoded, translatable errors.

As a first pass we should either return the passed error or a generic error.

plugins/woocommerce/src/StoreApi/Routes/V1/Checkout.php
```diff
@@ -177,58 +177,138 @@ protected function get_route_response( \WP_REST_Request $request ) {
...
+				'group'    => 'other',
+				'location' => 'contact',
+				'param'    => 'additional_fields',
+			],
+			'order'            => [
+				'group'    => 'other',
+				'location' => 'order',
+				'param'    => 'additional_fields',
+			],
+		];
```
senadir: Do you think those should live in a trait or somewhere else? they're majorly distracting the flow.

plugins/woocommerce/src/StoreApi/StoreApi.php
```diff
@@ -37,7 +38,22 @@ function () {
...
+					add_filter(
+						'woocommerce_session_handler',
+						function () {
+							return SessionHandler::class;
+						}
+					);
+				}
+
+				$cart_controller = new CartController();
+				$cart_controller->load_cart();
```
senadir: Even for products response 🤔 Dunno why I feel like this will bite us in the ass later, something something cache

senadir (commented):
Looking good Mike. I left some questions and the only change we need is the errors ones given those make it to frontend.

Alternatively, we can not translate them given we have the same error client side (by design)

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
mikejolley: Yes; saving was not possible here.

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
mikejolley: Yes checkout uses this key for contact fields too. https://github.com/woocommerce/woocommerce/blob/0096bc5021df4e63862d709e66c40b22b622ebaf/plugins/woocommerce/src/StoreApi/Routes/V1/Checkout.php#L192

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsSchema/Validation.php
mikejolley: Contact fields on `customer`.
Order fields on `checkout`.

plugins/woocommerce/src/StoreApi/Routes/V1/Checkout.php
mikejolley: It's not shared. Its very specific to this endpoint. So to that end, probably not? I tried to reuse things but it didn't work out. Frontend forms and this route need different handling to account for the fields being shown (groups and areas) and data being fed to the document object.

plugins/woocommerce/src/StoreApi/StoreApi.php
mikejolley: You're going to need to explain your logic—it's not clear if you're expecting me to get rid of this or not. Fact is we need the cart loaded early, otherwise we cannot validate during `validate_callback`.

senadir: It means we're now loading cart for all Store API endpoints, I'm raising that this might be a problem in the future, I get that we need to do this this early.

mikejolley: How might it be a problem? Performance? I think given Store API is for the Store, cart access is a given. We do limit this to Store API requests at least.

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsSchema/Validation.php
mikejolley: https://github.com/woocommerce/woocommerce/pull/55047/commits/b28531a4a696064638d233927838205cbf53ff7f

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFields.php
```diff
@@ -369,13 +368,45 @@ private function validate_options( &$options ) {
...
+	 */
+	private function process_field_options( $field_data, $options ) {
+		if ( 'checkbox' === $field_data['type'] ) {
+			$field_data = $this->process_checkbox_field( $field_data, $options );
+		} elseif ( 'select' === $field_data['type'] ) {
+			$field_data = $this->process_select_field( $field_data, $options );
+		}
+		// If the field has conditional required rules, we need to set the required property to false so it can be evaluated.
+		if ( Features::is_enabled( 'experimental-blocks' ) && ! empty( $field_data['rules']['required'] ) ) {
+			$field_data['required'] = false;
```
wavvves, line 411: This seems hacky 🤔

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFields.php
```diff
@@ -773,32 +804,71 @@ public function sanitize_field( $field_key, $field_value ) {
...
+			if ( ! $field ) {
+				return $errors;
+			}
+
+			// Validate required fields.
+			$is_required = $this->is_required_field( $field, $document_object, $document_object_context );
 
-				if ( is_wp_error( $validation ) ) {
-					$errors->merge_from( $validation );
+			if ( ( '' === $field_value || is_null( $field_value ) ) && $is_required ) {
```
wavvves: What if the field is required but no entry in the posted data? would this be skipped and the field treated like an optional field with empty data submitted?

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
```diff
@@ -145,7 +146,7 @@ public function edit_account_form_required_fields( $fields ) {
 
 		foreach ( $additional_fields as $key => $field ) {
 			if ( ! empty( $field['required'] ) ) {
```
wavvves: Instead of using a field as an array, can we enforce types via a field object and start checking this with `$field->is_required()` or `$field->required`? I know it's an extra effort, but I'm worried about these empty() calls being banalized until they wrongly evaluate data again like an issue we had in the past. I never liked empty() much, and practically only use it to skip empty arrays that can be null|false (otherwise I'll use count). Empty is not good on performance and its decision process is not cool for all contexts :/

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
```diff
@@ -186,41 +187,41 @@ public function edit_account_form_fields() {
...
-			$validation  = $this->checkout_fields_controller->validate_field( $key, $field_value );
+		// Validate individual fields agains the document object.
+		foreach ( $additional_fields as $field_key => $field_value ) {
+			$validation_result = $this->checkout_fields_controller->validate_field( $field_key, $field_value, $document_object, 'contact' );
 
-			if ( is_wp_error( $validation ) && $validation->has_errors() ) {
-				wc_add_notice( $validation->get_error_message(), 'error' );
-				continue;
+			if ( is_wp_error( $validation_result ) && $validation_result->has_errors() ) {
+				wc_add_notice( $validation_result->get_error_message(), 'error', array( 'id' => $field_key ) );
```
wavvves: I know error messages get condensed, but since validate and sanitize are probably going to be user defined should we account for full extent WP_Error usage? eg:

```php
if ( is_wp_error( $validation ) && $validation->has_errors() ) {
    foreach ( $validation->get_error_messages() as $error_message ) {
        wc_add_notice( $error_message, 'error' );
    }
}
```

wavvves: Also this is going to popup a lot, so a method for adding notices would be justified, or maybe incorporate it in the validate/sanitize invoker

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
```diff
@@ -186,41 +187,41 @@ public function edit_account_form_fields() {
...
+			} else {
```
wavvves: We should use `continue;` here instead of `else`. `elses` are always avoidable, not clean code friendly, and a sign that code structure and returns leading to it can be greatly improved if `else` seems not avoidable.

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
```diff
@@ -186,41 +187,41 @@ public function edit_account_form_fields() {
...
+		// Validate all fields for this location (this runs custom validation callbacks).
+		$location_validation = $this->checkout_fields_controller->validate_fields_for_location( $additional_fields, 'contact', 'other' );
 
 		if ( is_wp_error( $location_validation ) && $location_validation->has_errors() ) {
 			wc_add_notice( $location_validation->get_error_message(), 'error' );
+			return;
 		}
 
-		// phpcs:enable WordPress.Security.NonceVerification.Missing
 		$customer->save();
```
wavvves: I had this doubt on my PR, this is the second customer save on the request. If we persisted the fields, what else is this saving?

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
```diff
@@ -273,39 +274,74 @@ public function edit_address_fields( $address, $address_type ) {
...
+		foreach ( $additional_fields as $field_key => $field_value ) {
+			$validation_result = $this->checkout_fields_controller->validate_field( $field_key, $field_value, $document_object, $address_type . '_address' );
 
-			if ( ! isset( $_POST[ $post_key ] ) ) {
-				continue;
+			if ( is_wp_error( $validation_result ) && $validation_result->has_errors() ) {
+				wc_add_notice( $validation_result->get_error_message(), 'error', array( 'id' => $field_key ) );
+			} else {
+				$this->checkout_fields_controller->persist_field_for_customer( $field_key, $field_value, $customer, $address_type );
 			}
```
wavvves: Same as the above error handling and `continue;` suggestions.

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
wavvves: Wait I might be wrong about this, is `WP_Error::get_error_message()` responsible for condensing `WP_Error::get_error_messages()` into bullet points? If so just hit the shame bell 😖

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsSchema/Validation.php
```diff
@@ -0,0 +1,153 @@
...
+	public static function has_field_schema( $fields ) {
+		$return = false;
+
+		foreach ( $fields as $field ) {
+			if (
+				! empty( $field['rules'] ) && is_array( $field['rules'] ) &&
+				(
+					! empty( $field['rules']['required'] ) ||
+					! empty( $field['rules']['hidden'] ) ||
+					! empty( $field['rules']['validation'] )
```
wavvves, line 102: Again field object enforced types would be nice. A field->has_rules() using count() would be preferable.
We check a lot for stuff on fields, empty is not cool, and this leads to each dev making their own check preference which can lead to bugs 🫠

plugins/woocommerce/src/StoreApi/Authentication.php
```diff
@@ -22,6 +23,21 @@ public function init() {
...
+			if ( $has_valid_token ) {
+				// Overrides the core session class.
+				add_filter(
+					'woocommerce_session_handler',
+					function () {
+						return SessionHandler::class;
+					}
+				);
+			}
+		}
```
wavvves: Since we have a get_cart_token method here now we can make this cleaner:

```suggestion
        // Validate returns early if no token, so no perf penalty
		if ( JsonWebToken::validate( $this->get_cart_token(), $this->get_cart_token_secret() ) ) {
			// Overrides the core session class.
			add_filter(
				'woocommerce_session_handler',
				function () {
					return SessionHandler::class;
				}
			);
		}
```

plugins/woocommerce/src/StoreApi/Authentication.php
```diff
@@ -69,7 +85,7 @@ public function send_cors_headers( $value, $result, $request ) {
 
 		// Allow preflight requests, certain http origins, and any origin if a cart token is present. Preflight requests
 		// are allowed because we'll be unable to validate cart token headers at that point.
-		if ( $this->is_preflight() || $this->has_valid_cart_token( $request ) || is_allowed_http_origin( $origin ) ) {
+		if ( $this->is_preflight() || $this->has_valid_cart_token( $request->get_header( 'Cart-Token' ) ) || is_allowed_http_origin( $origin ) ) {
```
wavvves: Not sure why I added the redundant has_valid_cart_token method, but it probably can get deleted and replaced with validate. Also better than making it public, the last thing we'd want would be third parties using a redundant method, that would be a pet to maintain forever.

```suggestion
		if ( $this->is_preflight() || JsonWebToken::validate( $this->get_cart_token(), $this->get_cart_token_secret() ) || is_allowed_http_origin( $origin ) ) {
```

These conditions can also be reordered by performance.

plugins/woocommerce/src/StoreApi/Authentication.php
```diff
@@ -94,15 +110,25 @@ protected function is_preflight() {
...
 	 */
-	protected function has_valid_cart_token( \WP_REST_Request $request ) {
-		$cart_token = $request->get_header( 'Cart-Token' );
-
+	public function has_valid_cart_token( $cart_token = null ) {
+		if ( ! $cart_token ) {
+			$cart_token = $this->get_cart_token();
+		}
 		return $cart_token && JsonWebToken::validate( $cart_token, $this->get_cart_token_secret() );
 	}
```
wavvves: Yeah, let's kill this one..

plugins/woocommerce/src/StoreApi/Authentication.php
wavvves: I don't think there's a problem, actually believe this cart token thing is better suited to be here on Auth than the Abstract route 👍🏼

wavvves (commented):
I left some nagging comments, but a few might be useful 👍🏼 
This PR touches a lot, are there already Unit Tests for all the changed methods?

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFields.php
mikejolley: This is mainly to prevent someone miss-setting `required` at top level as well as in the rules. We're going to move the rules top level which will mitigate this in a follow up.

plugins/woocommerce/src/StoreApi/Authentication.php
mikejolley: https://github.com/woocommerce/woocommerce/pull/55047/commits/84cf630021cd8724c4b437e29243433d8d95f364

plugins/woocommerce/src/StoreApi/Authentication.php
mikejolley: https://github.com/woocommerce/woocommerce/pull/55047/commits/84cf630021cd8724c4b437e29243433d8d95f364

plugins/woocommerce/src/StoreApi/Authentication.php
mikejolley: https://github.com/woocommerce/woocommerce/pull/55047/commits/84cf630021cd8724c4b437e29243433d8d95f364

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
mikejolley: I'm gonna skip this one. I agree it would be useful, but `get_fields_for_location` is pre-existing and returns arrays. I'd rather not increase scope further with a refactor of those existing methods.

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsSchema/Validation.php
mikejolley: https://github.com/woocommerce/woocommerce/pull/55047#discussion_r1949135560

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
mikejolley: > This will get the first message available for the code. If no code is given then the first code available will be used.

We only want one validation message at a time because some overlap, and some would be duplicated. e.g. required field + not matching format.

As for custom error messages, thats coming in a follow up and needs schema. Nadir and I already talked about it.

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
mikejolley: https://github.com/woocommerce/woocommerce/pull/55047/commits/032c2f7717160d49a84e2aa290fcd7143453958c

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
mikejolley: https://github.com/woocommerce/woocommerce/pull/55047/commits/032c2f7717160d49a84e2aa290fcd7143453958c

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFieldsFrontend.php
mikejolley: The customer object is generated at the start of the method so it saves valid additional fields unless a "location" returns an error message. Invalid fields are skipped.

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutFields.php
mikejolley: It's down to the caller. The callers in this case iterate over fields rather than POST, so its not an issue.

plugins/woocommerce/src/StoreApi/Routes/V1/Checkout.php
```diff
@@ -175,58 +177,158 @@ protected function get_route_response( \WP_REST_Request $request ) {
...
+						if ( 'shipping_address' === $context ) {
+							/* translators: %s: is the field error message */
+							$error_message = sprintf( __( 'There was a problem with the provided shipping address: %s', 'woocommerce' ), $error_message );
+						} elseif ( 'billing_address' === $context ) {
+							/* translators: %s: is the field error message */
+							$error_message = sprintf( __( 'There was a problem with the provided billing address: %s', 'woocommerce' ), $error_message );
+						}
+						$errors->add( 'woocommerce_required_checkout_field', $error_message );
+					}
+					break;
```
mikejolley: Continue

plugins/woocommerce/src/StoreApi/Routes/V1/Checkout.php
```diff
@@ -175,58 +177,158 @@ protected function get_route_response( \WP_REST_Request $request ) {
...
+						$valid_check->add_data(
+							array(
+								'location' => $context_data['location'],
+								'key'      => $field_key,
+							),
+							$code
+						);
+					}
+					$errors->merge_from( $valid_check );
+					break;
```
mikejolley, line 293: continue

plugins/woocommerce/src/StoreApi/Routes/V1/Checkout.php
```diff
@@ -175,58 +177,158 @@ protected function get_route_response( \WP_REST_Request $request ) {
...
+	 * This runs after individual field validation_callbacks have been called.
 	 *
-	 * @throws RouteException When a required additional field is missing.
+	 * @param \WP_REST_Request $request Request object.
+	 * @return true|\WP_Error
 	 */
-	public function validate_required_additional_fields( \WP_REST_Request $request ) {
-		$document_object = null;
+	public function validate_callback( $request ) {
+		$sanitized_request = clone $request;
```
mikejolley, line 193: Comment

plugins/woocommerce/src/StoreApi/Routes/V1/Checkout.php
mikejolley: Continue

senadir (approved):
looking good, ship it

veljkho:
I added the heading `FOR GLOBALSTEP TESTING:` for the second part of the testing instructions, as I will delegate this PR for GS to test.

PR #52871, opened 2024-11-15
Fix "missing 'price' property in 'offers'" error in Google Search Console

Aljullu:
### Changes proposed in this Pull Request:

In https://github.com/woocommerce/woocommerce/pull/52105 we updated the structured data of sale prices. That was done by moving the sale price and the regular price under `priceSpecification`, which allowed for both prices to be listed in the metadata. That was an SEO enhancement.

Even though the code from that PR passed the validator checks, after testing it in an online store, it caused Google Search Console to output an error and a warning:

Missing 'price' in 'offers' | Missing 'priceValidUntil' in 'offers'
--- | ---
Captura de pantalla de 2024-11-15 11-02-52 | Captura de pantalla de 2024-11-15 11-02-44

The first one is specially serious, as it might cause the products to be delisted from Google enriched results.

To be honest, I'm not 100% sure why the errors appear, but I think:

* The first one was caused by `priceCurrency` still being under `offers`. I think in some way Google expects that if there is a `priceCurrency` property, there must be a `price` property at the same level. I made it so `priceCurrency` is only added if we detect a `price`, `lowPrice` or `highPrice` properties.
*  The second one seems to indicate that even though we added `validThrough` to the specific `priceSpecification`s, we still needed to keep `priceValidUntil` in the parent `offers`. I added it back.

I also made sure that we don't include `priceType` in the current offer. As this is what's advised by Google. And also made sure that in Grouped products there is no chance that the sale and regular price is the same. (I left comments inline explaining everything).

I tested it in a live store and Google Search Console seems to be fine again.

cc @zhyian in case you also want to take a look. :slightly_smiling_face: 

Note: I'm targeting `release/9.5` as I think this deserves a CFE so we don't ship a regression in WC 9.5.

### How to test the changes in this Pull Request:

1. Create an online store that can be crawled by Google Search Console. Create several products: simple, variable, grouped, on sale, etc.
2. Inspect all product pages using the Search Console URL [Inspector Tool](https://support.google.com/webmasters/answer/9012289?hl=en).
3. Verify there are no errors in any of them.

PR #55061, opened 2025-02-03
Add sale price before regular price in structured data

### LINKED ISSUE #55043: Yandex only displaying the regular price when sale price is set after 9.5.0 update
### Prerequisites

### Describe the bug

As per comment in #52105, after the 9.5.0 release and #52871, Yandex stopped displaying the sale price in search results and is only displaying the full price – I'm assuming because priceType is not set to "http://schema.org/SalePrice" in the schema and it's only considering the top entry. I understand this may be Google's recommended format but might need to revise Yandex's rules.

Image

### Expected behavior

Ideally both the sale and the full price are set, with the full price crossed out, if not the sale price only.

### Actual behavior

Only the full price is displayed when the sale price is set.

### Steps to reproduce

1. Create a product
2. Add both the full price and a sale price
3. priceType is missing "http://schema.org/SalePrice"

### WordPress Environment

Fresh Playground install with no plugins.

### Isolating the problem

Aljullu:

See https://github.com/woocommerce/woocommerce/issues/55043 for more details. For some reason, it seems like Yandex didn't like the changes we introduced in https://github.com/woocommerce/woocommerce/pull/52105 and https://github.com/woocommerce/woocommerce/pull/52871. I suspect it might be because when there is a regular price and a sale price, we print the sale price as the second one.

Printing the sale price first, would also make us aligned with Google's example:

https://developers.google.com/search/docs/appearance/structured-data/merchant-listing#sale-pricing-example

Closes #55043.

1. Create a simple product with a sale. Go to the frontend and copy and paste the page source code.
2. Paste it in https://search.google.com/test/rich-results and verify the price info is correct (you need to click on Product Snippet to see the product data; you might see some errors due to missing reviews, shipping details, hasMerchantReturnPolicy or the image not being accessible by Google, but that's unrelated :slightly_smiling_face: ): 
imatge
3. Paste it in https://validator.schema.org/ and verify the price info is also correct:
imatge
4. Repeat steps 1-3 with a Grouped product on sale. Keep in mind that you will need the sale price of one of the products inside the grouped product to be the lowest. For example, if product A has price $18, product B (the one on sale) needs to be lower than $18 (ie ~$20~ $17).

imatge

### DIFF
--- a/plugins/woocommerce/changelog/fix-55043-sale-price-before-regular-price
+++ b/plugins/woocommerce/changelog/fix-55043-sale-price-before-regular-price
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Add sale price before regular price in structured data to better SEO
--- a/plugins/woocommerce/includes/class-wc-structured-data.php
+++ b/plugins/woocommerce/includes/class-wc-structured-data.php
@@ -336,12 +336,17 @@ class WC_Structured_Data {
 						$sale_price_valid_until = gmdate( 'Y-m-d', $product->get_date_on_sale_to()->getTimestamp() );
 					}
 
-					$markup_offer['priceSpecification'][] = array(
-						'@type'                 => 'UnitPriceSpecification',
-						'price'                 => wc_format_decimal( $min_sale_price, wc_get_price_decimals() ),
-						'priceCurrency'         => $currency,
-						'valueAddedTaxIncluded' => wc_prices_include_tax(),
-						'validThrough'          => $sale_price_valid_until ?? $price_valid_until,
+					// We add the sale price to the top of the array so it's the first offer.
+					// See https://github.com/woocommerce/woocommerce/issues/55043.
+					array_unshift(
+						$markup_offer['priceSpecification'],
+						array(
+							'@type'                 => 'UnitPriceSpecification',
+							'price'                 => wc_format_decimal( $min_sale_price, wc_get_price_decimals() ),
+							'priceCurrency'         => $currency,
+							'valueAddedTaxIncluded' => wc_prices_include_tax(),
+							'validThrough'          => $sale_price_valid_until ?? $price_valid_until,
+						)
 					);
 				}
 			} else {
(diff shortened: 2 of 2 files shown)


gigitux (approved):
LGTM! Thanks for working on this! 🙇 

I would suggest updating the testing instructions for external testers to clarify that, for Google Search, they need to click on ‘Product Snippet’ to view product information.
Additionally, for Grouped Products, consider specifying that the expected behavior is to display the sale price of the lowest-priced product.

Aljullu:
Thanks for the review, @gigitux!

> I would suggest updating the testing instructions for external testers to clarify that, for Google Search, they need to click on ‘Product Snippet’ to view product information.
> Additionally, for Grouped Products, consider specifying that the expected behavior is to display the sale price of the lowest-priced product.

Good points, I updated this in the PR testing steps. :+1:

ivantipov:
Can confirm the sale price is now being displayed, with or without the SalePrice entry – though not both as my client says it used to do and as you can see in an example here:

https://www.ozon.ru/product/apple-smartfon-iphone-16-pro-esim-sim-8-256-gb-seryy-1687844335/?campaignId=527

Capture d’écran 2025-02-12 à 13 49 37

Schema for that page only shows the sale price.

Aljullu:
Thanks for checking, @ivantipov! Happy to hear that the correct price is showing up!

> Schema for that page only shows the sale price.

I checked the page you sent and I'm a bit confused, I would expect `priceSpecification` to be there with an array of prices. Can you share what WooCommerce version that store is using and what type of product? Also, my understanding is that you implemented the changes from this PR manually? Thanks in advance!

ivantipov:
Hi @Aljullu, this is not a WooCommerce store, just an example of both prices being shown in a search result and the specification needed for that. My client claims this was also the case in his store before 9.5.0 – now it's just the sale price. I did add the fix manually to that store, yes.

Aljullu:
Thanks for coming back to us, @ivantipov! I created another PR (https://github.com/woocommerce/woocommerce/pull/55561) adding back the `priceType` property to structured data. I would love if you could give it a try, as we haven't been able to reproduce the issue yet. Given that we can't be sure whether that PR indeed fixes the issue, we will need somebody test it and confirm it works for them. :pray: 

In case you want to try it, you can find the updated [`class-wc-structured-data.php` file](https://github.com/woocommerce/woocommerce/blob/fix/55043-list-price-structured-data/plugins/woocommerce/includes/class-wc-structured-data.php) here. Thanks in advance!

PR #55062, opened 2025-02-03
[e2e tests] Improve create-shipping-spec

adimoldovan:

Some tweaks in `create-shipping-zones.spec.js` that make a faster and more robust spec.

- ensure some general settings are as expected by the tests (e.g. decimal settings)
- relaxed some locators in assertions. We don't need to check for strong and other css classes - it was failing in some environments
- added some todo comments to follow-up on the validity of the tests using conditionals
- use the api instance from fixtures instead of creating a new one each times throughout the spec
- updated some locators to make my IDE happy, it was showing errors for invalid selectors
- removed some extra page reloads. If the is an issue there with the content not being updated in time we need to fix it in other ways, not by forcing a page reload
- removed a code block waiting for a tour. this added an extra 5s execution time in an unexpected place inside the spec. If such a tour will cause issues in the future we'll solve it by using Playwright's [addLocatorHandler](https://playwright.dev/docs/api/class-page#page-add-locator-handler) instead (which doesn't wait unnecessarily)

Ran the spec many times on repeat - they passed every time. Tested on local wp-env and JN multisite.

```
pnpm test:e2e create-shipping-zones.spec.js
```

### DIFF
--- a/plugins/woocommerce/changelog/e2e-improve-create-shipping-zones-spec
+++ b/plugins/woocommerce/changelog/e2e-improve-create-shipping-zones-spec
@@ -1,0 +1,4 @@
+Significance: patch
+Type: dev
+
+E2E tests: improve create-shipping-zones spec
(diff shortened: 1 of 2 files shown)


lanej0 (approved):
Ran it locally 25 times, and all good.

I agree with your TODO comments.  We should definitely prioritize removing those conditionals.

Issue #49695, opened 2024-07-18
[E2E] Programmatically install WC package on the externally hosted sites 


Stojdza:
In order to run tests against external environments (Pressable, WPCOM) using the latest package (Beta, RC), we should install WC package there before automated tests start. 

More context: pdV5qK-wB-p2

kalessil:
@Stojdza: assigning to Solaris team, but feel free to assign to Vortex if we are expected to work on it.

Stojdza:
Pressable API for WP CLI commands - https://my.pressable.com/documentation/api/v1#wp-cli-commands

juliaamosova:
@lanej0 is AFK this week and will be back to work on Oct 21st. He will pick up this issue then. 

I think the effort here is not too big, but the improvement is a great one even if we run tests not too often. Suppose somebody not close to the release work wants to run tests, for example. In that case, they will need to know the site URL, find the login details, log in, look for the latest beta/RC, etc. - too many steps compared to just clicking on the button to initiate a test run. It will be great to finalize this work now before moving away to something else.

lanej0:
It's a little tricky just based on the wide variety of web hosts out there.  We may not have shell access/WP CLI access with some (Bluehost, Hostgator) for example.  I've got this task back on my plate, and I'll try to work out a broadly-applicable solution.

lanej0:
I made some updates to the helper plugin I made for test site cleanup.  You can now pass it a URL of a version to install through an API endpoint and it will update/install that copy of WooCommerce.

The linked PR above lays the groundwork.  Then we'll just need to update the environment setup scripts to call the API to update WC.

rodelgc:
Hi @lanej0 just to be clear: we'll have to wait for [Update to e2e helper plugin to allow WC updates #52350](https://github.com/woocommerce/woocommerce/pull/52350) to be completed first before being able to proceed with this task? Thanks :)

Stojdza:
Following up on the conversation from Slack with @lanej0 - I'm assigning myself to this; I will start working on it next week.

Stojdza:
This issue is part of [the project](https://github.com/orgs/woocommerce/projects/286/views/1) that is in progress and will be done as part of it. I'll lower the priority to normal to save some time during the sweeps.

Stojdza:
With [this PR being](https://github.com/woocommerce/woocommerce/pull/54103) merged, we can now utilize WC Beta Tester to install desired version of WC on the website where we're running tests.

[...]

PR #55063, opened 2025-02-03
[e2e] Install specific version of WC using WC Beta Tester

Stojdza:

This PR aims to introduce the option to[ install a specified version of WC during test execution](https://github.com/woocommerce/woocommerce/issues/49695) using WC Beta Tester API. If enabled, by default, it will install the latest version of WC (latest final, Beta, or RC). 

More info on how WC Beta Tester can be used to install a specific version of WC can be found in this PR https://github.com/woocommerce/woocommerce/pull/54103.

This functionality will be used for running tests on external environments where installing WC using WC Beta Tester is enabled by default. For local execution and `wp-env`, there should be no changes. 

Part of https://github.com/woocommerce/woocommerce/issues/49695

Besides the new logic for installing WC using the WC Beta Tester, this PR enables installation only for the Pressable website. 
1. Log in to the Pressable website and check the installed version of WooCommerce. If not, install and activate the latest stable version (we need this for testing).
2. Checkout this branch and run `pnpm test:e2e:pressable basic.spec.js` and wait for the tests to finish. Was it successful?
3. Go to `/wp-admin/plugins.php` page on the Pressable website. The latest release, in this case `[9.6.1](https://github.com/woocommerce/woocommerce/releases/tag/9.6.1)`, should be installed and activated
4. Now run `WC_VERSION=9.6.0-rc.1 pnpm test:e2e:pressable basic.spec.js --ui` and wait for the tests to finish. Was it successful?
5. Go to `/wp-admin/plugins.php` page on the Pressable website. WC `9.6.0-rc.1` should be installed and activated. 
6. Go to `/wp-admin/plugins.php` and deactivate WC (make sure that no WC plugin is active)
7.  Now run `WC_VERSION=9.5.2 pnpm test:e2e:pressable basic.spec.js --ui` and wait for the tests to finish. Was it successful?

Additional testing and consideration:
1. Is CI green? These changes should not affect CI execution and execution with `wp-env`
2. Take a look at `site.setup.js` file, does the new fixture `wcbtApi` makes sense?
3. I find running command with `WC_VERSION=9.5.2 pnpm test:e2e:...` really handy. That way, I could easily run tests against previous versions of WC if I need to check or compare something. Is this something we can use for `wp-env` as well? In that case, we would need to run `wp-env` with WC Beta Tester installed. 

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Install a specific version of WC using WC Beta Tester.

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/tests/e2e-pw/fixtures/wc-install.setup.js
```diff
@@ -0,0 +1,98 @@
+/**
+ * Internal dependencies
+ */
+import { test as setup } from './fixtures';
+import { ADMIN_STATE_PATH } from '../playwright.config';
+
+setup.use( { storageState: ADMIN_STATE_PATH } );
+
+setup( 'install WC using WC Beta Tester', async ( { page } ) => {
+	if ( process.env.INSTALL_WC === 'true' ) {
```
adimoldovan: May I suggest to use `setup.skip()` instead of this conditional? With [the skip option](https://playwright.dev/docs/api/class-test#test-skip) the setup script will be reported as such and it will be more clear that the plugins was not installed.

Something in the lines of:
```
setup.skip( !process.env.INSTALL_WC, 'Skipping installing the plugin, INSTALL_WC not found');
```

adimoldovan:
@Stojdza Thanks for working on this! We're getting closer to having the full flow automated. 

The first thing I noticed is the use of the browser and `page.evaluate`. Those endpoints should be available to call as a regular api, no browser should be needed. Maybe we should add a new [fixture](https://github.com/woocommerce/woocommerce/blob/1e4361c10153f19c4fa89d1e79971188055bc470/plugins/woocommerce/tests/e2e-pw/fixtures/fixtures.js#L23), `wcbtApi`, just like we have for `wcAdminApi`, that uses the `wc-admin-test-helper` namespace.

plugins/woocommerce/tests/e2e-pw/fixtures/wc-install.setup.js
Stojdza: Thanks Adi! This is nice and clean 👍 Implemented it here [b720af5](https://github.com/woocommerce/woocommerce/pull/55063/commits/b720af509827e87a5430e0ae656112afbc923c49)

Stojdza: This is now part of [`site.setup.js` fixture](https://github.com/woocommerce/woocommerce/pull/55063/files#diff-f6a4141a21bfc84664a41827dcf301bc2facc6822d2d1d56ec5eaf6ce6bb5188R109-R112).

Stojdza:
@adimoldovan, thanks for pushing me in this direction! 🙌 

> Those endpoints should be available to call as a regular api, no browser should be needed. Maybe we should add a new [fixture](https://github.com/woocommerce/woocommerce/blob/1e4361c10153f19c4fa89d1e79971188055bc470/plugins/woocommerce/tests/e2e-pw/fixtures/fixtures.js#L23), wcbtApi, just like we have for wcAdminApi, that uses the wc-admin-test-helper namespace.

Because of [this comment](https://github.com/woocommerce/woocommerce/pull/54103#issuecomment-2569027694) and how we tested the functionality so far, I thought it was mandatory to load the page first (similar to what we're doing in `token.setup.js`). But thanks to your proposal, I saw it can be done using the browser, which is faster and more reliable. 

In the last few commits, I addressed the feedback and updated the PR description. When you have a time, can you take one more look? 

There is a case when this solution won't work on the external environments - if there's no WC installed on the website, `token.setup.js` will fail. With this "only API" approach, we need a token to install the desired WC version; and we need WC installed to set up a token 🔁 I'm working on a solution at the moment, and I can push it here or in the follow-up PR. It's a case that won't happen often, but I would rather have it done as robust as possible.

adimoldovan:
Thanks for applying the suggestions @Stojdza.

> There is a case when this solution won't work on the external environments - if there's no WC installed on the website, `token.setup.js` will fail. With this "only API" approach, we need a token to install the desired WC version; and we need WC installed to set up a token 🔁 

We don't need the token to install the desired WC version, `wcbtApi` is using basic authentication and we have the user and password at that point.

But still, `token.setup.js` will fail if it runs without WC installed, the same as all the other steps in the setup that rely on the plugins being installed.

The install needs to be its own setup file, and it should run before everything else. With the current structure, the global authentication `auth.setup.js` which is the first one running should depend on (the new) `install-wc.setup.js`.

Stojdza:
Thanks for the suggestion @adimoldovan 👍 I added [the changes](https://github.com/woocommerce/woocommerce/pull/55063/commits/14cecaecd597b39d5ed5931cda0c409736b37e4d), and while tests and WC installation step are passing locally against Pressable and wp-env, they are failing on GH with [this error](https://github.com/woocommerce/woocommerce/actions/runs/13198620538/job/36845435821?pr=55063). Looking at why is this happening 👀

plugins/woocommerce/tests/e2e-pw/fixtures/fixtures.js
```diff
@@ -51,6 +51,88 @@ exports.test = base.test.extend( {
...
+					'./wp-json/wc-admin-test-helper/live-branches/activate/v1',
+					{
+						method: 'POST',
+						data: {
+							version: resolvedVersion,
+						},
+					}
+				);
+			},
+		} );
```
adimoldovan: You don't need to implement all these methods here (not even sure that works?), all you need is to use the `wcbtApi`, to expose it to wherever the fixture is used.

The code from the functions `deactivateWC`, `installLatest`, etc. should be in the `install-ws.setup.js` only. I don't think we'll need this anywhere else, so I don't think extracting the logic in public functions is required.

```suggestion
		await use( wcbtApi );
```

plugins/woocommerce/tests/e2e-pw/playwright.config.js
```diff
@@ -62,10 +62,16 @@ if ( process.env.CI ) {
 }
 
 export const setupProjects = [
+	{
+		name: 'install wc with wc beta tester',
```
adimoldovan: Nitpick: I would go with a shorter title, like "install wc". I don't see it relevant to specify it is using the beta tester.

plugins/woocommerce/tests/e2e-pw/fixtures/install-wc.setup.js
```diff
@@ -0,0 +1,68 @@
+/**
+ * Internal dependencies
+ */
+import { test as setup } from './fixtures';
+import { ADMIN_STATE_PATH } from '../playwright.config';
+
+setup.use( { storageState: ADMIN_STATE_PATH } );
```
adimoldovan: This is not needed and it's the cause of the failure you mentioned.

```suggestion
```

adimoldovan (changes requested):
>...they are failing on GH with [this error](https://github.com/woocommerce/woocommerce/actions/runs/13198620538/job/36845435821?pr=55063).

That's caused by the `setup.use( { storageState: ADMIN_STATE_PATH } );` in the `install-wc.setup.js`. That requires the `./state/admin.json` file to existst, but that is only created by the authentication setup step. You need to delete that part, is is not required here. I also left a code suggestion for this.

> and while tests and WC installation step are passing locally

Deleting the `./state/admin.json` before will cause a failure locally as well. You have that file from previous runs, that's why it's not failing.

plugins/woocommerce/tests/e2e-pw/playwright.config.js
Stojdza: The shorter the better 👍 I went with beta tester part because I was thinking about someone who might run the tests (let's say against wp-env) and see the step "install wc" skipped, and might think why WC is skipped and not installed

plugins/woocommerce/tests/e2e-pw/fixtures/fixtures.js
Stojdza: Done in [de3a60a](https://github.com/woocommerce/woocommerce/pull/55063/commits/de3a60a5e850f95db30cd9b6eacb7a0648c28fea)

plugins/woocommerce/tests/e2e-pw/playwright.config.js
Stojdza: Done - [225b1ac](https://github.com/woocommerce/woocommerce/pull/55063/commits/225b1ac3d70428ca4c52f7a3047b6c273f9098bf)

Stojdza:
Feedback addressed. Additionally I updated dependencies for [Pressable](https://github.com/woocommerce/woocommerce/pull/55063/commits/070cadd1254848009e0b8bb3960ac026e9b56c10) and [WPCOM](https://github.com/woocommerce/woocommerce/pull/55063/commits/42cf1c71445ae5cca39117e15cce7772b6cac0d1), since site reset was failing after these changes. 

[Here](https://github.com/woocommerce/woocommerce/actions/runs/13204482197) is the on-demand job with the latest changes. For Pressable, `[install wc] › ../fixtures/install-wc.setup.js:6:6 › Install WC using WC Beta Tester (6.0s)` was successfully executed (also verified WC version on the website), and for WPCOM this step was skipped.

rodelgc:
Hello @Stojdza I'm not sure if it's related to the changes in this PR, but I'm seeing multiple WooCommerce versions on the Pressable site.

Screenshot 2025-02-07 at 10 50 25 PM

Stojdza:
@rodelgc 👋 Yes, it was me, testing. With these changes, we don't delete plugins. We can activate/deactivate them and install them. Feel free to remove any of these if needed. Thanks!

adimoldovan:
> Additionally I updated dependencies for [Pressable](https://github.com/woocommerce/woocommerce/pull/55063/commits/070cadd1254848009e0b8bb3960ac026e9b56c10) and [WPCOM](https://github.com/woocommerce/woocommerce/pull/55063/commits/42cf1c71445ae5cca39117e15cce7772b6cac0d1), since site reset was failing after these changes.

@Stojdza  I'm not seeing why this is needed. Both site-setup and reset should work fine. Can you point me to such a failure?

plugins/woocommerce/tests/e2e-pw/envs/default-pressable/playwright.config.js
```diff
@@ -2,6 +2,7 @@ let config = require( '../../playwright.config.js' );
 const { tags } = require( '../../fixtures/fixtures' );
 
 process.env.IS_PRESSABLE = 'true';
+process.env.INSTALL_WC = 'true';
```
adimoldovan, line 5: This is missing in the wpcom config. Is it intentional?

plugins/woocommerce/tests/e2e-pw/fixtures/install-wc.setup.js
```diff
@@ -0,0 +1,114 @@
...
+		console.error( 'Error deactivating WooCommerce:', err );
+	}
+
+	const wcVersion = process.env.WC_VERSION || 'latest';
+	let resolvedVersion = '';
+
+	// Install WC
+	if ( wcVersion === 'latest' ) {
+		try {
+			const latestResponse = await wcbtApi.fetch(
```
adimoldovan: Shouldn't we first check if the latest version is already installed and skip the installation if so?

Stojdza:
> I'm not seeing why this is needed. Both site-setup and reset should work fine. Can you point me to such a failure?

[Here](https://github.com/woocommerce/woocommerce/actions/runs/13203501453/job/36862124874#step:9:121) it is. `[reset]` was executed right after `[install wc]`. After I changed the order it passed, but I'm open to other suggestions.

adimoldovan:
> > I'm not seeing why this is needed. Both site-setup and reset should work fine. Can you point me to such a failure?
> 
> [Here](https://github.com/woocommerce/woocommerce/actions/runs/13203501453/job/36862124874#step:9:121) it is. `[reset]` was executed right after `[install wc]`. After I changed the order it passed, but I'm open to other suggestions.

Ah, I see. It was actually requiring global authentication, because it was missing the `.state/admin.json` file. BUT, it should not need that file, the `setup.use( { storageState: ADMIN_STATE_PATH } );` was probably a left-over and should be removed. Can you please remove it, and if you revert the dependencies changes I'm confident everything will work.

plugins/woocommerce/tests/e2e-pw/envs/default-pressable/playwright.config.js
Stojdza: Yes, because WC on our WPCOM plan cannot be deactivated. I shared this in the Slack update here: p1736022424909529-slack-C0833T6CFHP, and @rodelgc confirmed it in the following message. In this PR, I focused on implementing the functionality and the Pressable website.

plugins/woocommerce/tests/e2e-pw/fixtures/install-wc.setup.js
Stojdza: It's not that I didn't think about it, but installing and activating WC via API was fast and I decided to go without checking if the specific WC version is already installed and activated. Additionally, we wouldn't run tests against external envs very often, so that was one more reason to skip the check. 

Anyway, I implemented it here [c9da3bd](https://github.com/woocommerce/woocommerce/pull/55063/commits/c9da3bddf46ed1cd95c55d86d328cec514df28fd). It took me longer than expected, but it should work now. When testing, please consider these scenarios and let me know your thoughts:

- Install the latest release with `pnpm test:e2e:pressable basic.spec.js` when:
   - the latest is already installed and activated
   - other version of WC is installed and activated
   - no WC is activated
- Install specific WC with `WC_VERSION=9.6.0-rc.1 pnpm test:e2e:pressable basic.spec.js` when:
   - the same version is already installed and activated
   - the latest is already installed and activated
   - no WC is activated

Thanks 🙌

plugins/woocommerce/tests/e2e-pw/envs/default-pressable/playwright.config.js
```diff
@@ -16,6 +16,7 @@ config = {
 			name: 'reset',
 			testDir: `${ config.TESTS_ROOT_PATH }/fixtures`,
 			testMatch: 'reset.setup.js',
+			dependencies: [ 'site setup' ],
```
Stojdza: @adimoldovan, I added this dependency because reset was returning `500` without it ([example](https://github.com/woocommerce/woocommerce/actions/runs/13266737356/job/37036060223)). This way, it will wait for `setupProjects` to finish first. Please, let me know if there's a better idea on how to solve it. [Here](https://github.com/woocommerce/woocommerce/actions/runs/13287596105/job/37101904978) is the successful site reset when this change was added.

plugins/woocommerce/tests/e2e-pw/fixtures/install-wc.setup.js
```diff
@@ -0,0 +1,144 @@
...
+				`Failed to install latest WC: ${ latestResponse.status() } ${ await latestResponse.text() }`
+			);
+		}
+
+		resolvedVersion = ( await latestResponse.json() )?.version || '';
+
+		setup.skip(
+			resolvedVersion === activatedWcVersion,
+			'Skip installing WC: The latest version is already installed and activated.'
+		);
```
adimoldovan: I would no use `setup.skip` at this point, it already executed half of the setup and leads to confusion. The step was marked as skipped while the log was telling me `INSTALL_WC is enabled. Running installation script...`

```suggestion
		if(resolvedVersion === activatedWcVersion) {
			console.log('Required version is already installed and active.');
			return;
		}
```

plugins/woocommerce/tests/e2e-pw/fixtures/install-wc.setup.js
```diff
@@ -0,0 +1,144 @@
...
+				err
+			);
+		}
+	} else {
+		console.error(
+			'Error: resolvedVersion is undefined. Skipping activation.'
+		);
+	}
+
+	console.log( 'Installing with WC Beta Tester is finished.' );
```
adimoldovan: Can we also add a check that the plugin version is actually installed and active, similar to the initial `wp-json/wp/v2/plugins` check.

adimoldovan (changes requested):
Thanks for the updates @Stojdza.

I left comments on code.

As a general comment, for a follow-up, not necessarily in this PR, we could address the fact that the deactivated versions are not deleted and different versions are piling up. We could [delete them](https://developer.wordpress.org/rest-api/reference/plugins/#delete-a-plugin).

plugins/woocommerce/tests/e2e-pw/envs/default-pressable/playwright.config.js
adimoldovan: I could also reproduce this in my tests, and my best guess is that it happens because the plugin is not yet ready. Delaying the reset call helps and I'm not opposed to that, I'm only opposing it running after the site-setup, because it can potentially remove things that are set by the site-setup. In my mind it should be reset, then authentication, then site-setup.
In my testing I found that adding a simple check via the API that the correct plugin version is actually installed and active as mentioned [here](https://github.com/woocommerce/woocommerce/pull/55063#discussion_r1954300655) added enough delay for the reset to work. If not, I would be ok with adding 'global authentication' as a dependency for reset.

plugins/woocommerce/tests/e2e-pw/fixtures/install-wc.setup.js
Stojdza: Added [49422a7](https://github.com/woocommerce/woocommerce/pull/55063/commits/49422a742b5ef7b4c9cadcda3fefe7f5a3758203)

plugins/woocommerce/tests/e2e-pw/envs/default-pressable/playwright.config.js
Stojdza: After checking WC version at the end of the script [49422a7](https://github.com/woocommerce/woocommerce/pull/55063/commits/49422a742b5ef7b4c9cadcda3fefe7f5a3758203) I also reverted this change  [49422a7](https://github.com/woocommerce/woocommerce/pull/55063/commits/49422a742b5ef7b4c9cadcda3fefe7f5a3758203). Reset works fine now 👍

plugins/woocommerce/tests/e2e-pw/fixtures/install-wc.setup.js
Stojdza: You're right. Logging was confusing indeed. I removed `setup.skip` from two places and improved logging here [c15600c](https://github.com/woocommerce/woocommerce/pull/55063/commits/c15600c9d45e2c1ff89521e4bc74fb5b4eda6103)

Stojdza:
> As a general comment, for a follow-up, not necessarily in this PR, we could address the fact that the deactivated versions are not deleted and different versions are piling up. We could [delete them](https://developer.wordpress.org/rest-api/reference/plugins/#delete-a-plugin).

I agree. Let's address this in a separate PR 👍

adimoldovan (approved):
Thanks for addressing all the suggestions @Stojdza! 🚀

PR #55079, opened 2025-02-04
[Settings] Add support for radio type

### LINKED ISSUE #54590: [Settings] Handle `radio` type input
`radio` inputs can use `RadioControl` from Gutenberg, see https://wordpress.github.io/gutenberg/?path=/docs/components-radiocontrol--docs

https://github.com/woocommerce/woocommerce/blob/7f513fe2e50a31bfb0687e55dce9ae8c857976cc/plugins/woocommerce/includes/admin/class-wc-admin-settings.php#L457

psealock:

Closes https://github.com/woocommerce/woocommerce/issues/54590

Add support for `radio` input types and make use of `getLabelAndHelp` for `textarea` inputs.

Note on labels: Gutenberg uses the `id` when the label is blank which means we are rendering this value. Given that we have no label when `desc_tip` is true, this means we'll render the id each time.

https://github.com/woocommerce/woocommerce/blob/trunk/packages/js/settings-editor/src/hooks/utils/transformers.tsx#L39-L44

I'm not entirely sure what to do here, but I think we should handle it on a component to component basis in a follow up because some inputs require a label but others like a fieldset of radios require a legend. We should await clearer direction from the Core Settings discussions and either handle this upstream or in Woo.

1. Add/enable Gutenberg 19.8 or older
2. Toggle on the settings feature flag
3. WooCommerce > Settings > Shipping > Shipping Settings and see the Shipping destination radio group. Make sure it renders well. See the setting id as the label and the note above.

image

4. WooCommerce > Settings > Accounts & Privacy and see the text areas rendered.

image

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

Add support for radio types in new, unreleased Settings.

</details>


chihsuan (approved):
LGTM and tested well! 👍 

> I'm not entirely sure what to do here, but I think we should handle it on a component to component basis in a follow up because some inputs require a label but others like a fieldset of radios require a legend. We should await clearer direction from the Core Settings discussions and either handle this upstream or in Woo.

 I handled this for the inputEdit here. IMO, we should update upstream to allow label to be optional. It shouldn't use id when label is empty.
https://github.com/woocommerce/woocommerce/blob/abb8aa24ee9f4927fcd6f8b70ba0a797de953fe7/packages/js/settings-editor/src/components/inputEdit/index.tsx#L19-L21

ilyasfoo (approved):
LGTM! 🚢

Issue #54259, opened 2025-01-08
Migrate WooCommerce to the published version of `@wordpress/interactivity`


samueljseay:
- [ ] The goal of this is to remove the privately incubated version of `@wordpress/interactivity` within WooCommerce and move to using the publically published npm module. This will require some work to support ESM modules within our existing interactivity components.

More detailed issue tickets TBD.

luisherranz:
Once we transition to using the package `@wordpress/interactivity`, we should take a look at this bug to see if it is resolved or if it is still present:

- https://github.com/woocommerce/woocommerce/issues/52796

samueljseay:
I'm descoping a few of the issues here by removing them from the epic, but I won't close the ones we still want to do later. I'll create a new epic for the leftover work that can be reprioritized in a follow up project.

PR #55080, opened 2025-02-04
Migrate released interactivity blocks to `@wordpress/interactivity`

### LINKED ISSUE #54208: Move imports used in interactivity API code to a modules build
Description:

For the modules that are currently being used in Interactivity blocks and components within WooCommerce we need to move them to support ESM so that we can move to the latest version of Interactivity API.

This will require creating a separate Webpack bundle to build these with the latest Wordpress DEWP and to remove incompatibilities with ESM.

### LINKED ISSUE #54501: Create an event to sync wc data store and interactivity store when there is a change in each of them.

### LINKED ISSUE #54502: Stop reading from the wp.data store and move logic to the iAPI store.

samueljseay:

This feature branch encompasses the minimal changes needed to migrate existing (released) blocks using the interactivity API to using `@wordpress/interactivity`. 

Changes:

Blocks affected:

`woocommerce/store-notices`
`woocommerce/product-button`
`woocommerce/product-collection`
`woocommerce/product-gallery`

And also the product filters blocks

* Migrate imports to `@wordpress/interactivity`
* Migrate directives from `wc` to `wp`
* Create a new script modules build for those blocks
* Implement basic support for interactivity powered notices within `woocommerce/store-notices` block
* Utilize iapi powered store notices within `woocommerce/product-button`
* Also brings in https://github.com/woocommerce/woocommerce/pull/54979 to sync iAPI store with cart store

Closes https://github.com/woocommerce/woocommerce/issues/54208
Closes https://github.com/woocommerce/woocommerce/issues/54501
Closes https://github.com/woocommerce/woocommerce/issues/54502

1. Test the affected blocks especially in frontend to confirm that they function as before
2. Test the new client-side store notices
3. Test that mini cart stays up to date with cart changes

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

Migrate iAPI blocks to `@wordpress/interactivity`

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
```diff
@@ -74,6 +70,7 @@ public function dequeue_add_to_cart_scripts() {
 	 * @return string Rendered block type output.
 	 */
 	protected function render( $attributes, $content, $block ) {
+		wp_enqueue_script_module( 'woocommerce/product-button' );
```
samueljseay: I better move this further down past the early return

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
```diff
@@ -178,31 +177,22 @@ protected function render( $attributes, $content, $block ) {
...
-		$button_directives = '
-			data-wc-on--click="actions.addToCart"
-			data-wc-class--loading="context.isLoading"
-		';
-
-		$anchor_directive = '
-			data-wc-on--click="woocommerce/product-collection::actions.viewProduct"
-		';
+		$button_directives = 'data-wp-on--click="actions.addCartItem"';
+		$anchor_directive  = 'data-wp-on--click="woocommerce/product-collection::actions.viewProduct"';
```
luisherranz, line 200: This directive is also injected [from here](https://github.com/woocommerce/woocommerce/blob/04c998dc520c8112288c4babb67a6adfb88146bc/plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/Controller.php#L155), so maybe we are repeating it?

gigitux:
Here are the results from my recent tests:

I created two fresh sites:
- Site 1: Runs the latest stable WooCommerce version.
- Site 2: Uses a WooCommerce build from this branch.

### ⚠️ Product Button

It looks like that there is a regression regarding the Product Button. On Site 2 for a second, I see only a dark button without any text inside. This doesn't happen on Site 1.


| Stable | Interactivity API branch |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/d775a60a-5d48-48b7-8328-57f539918495"></video>| <video src="https://github.com/user-attachments/assets/8b84874f-2c5a-4218-9561-03349be3aa1b"> </video>| 


### 👍 Product Collection

Pagination works as expected.

### 👍  Product Gallery

Everything works as expected.


### Store notices

I think that there isn't any to test yet mostly because we decide to pivot to [this approach](https://github.com/woocommerce/woocommerce/pull/55717), right?

samueljseay:
> I think that there isn't any to test yet mostly because we decide to pivot to https://github.com/woocommerce/woocommerce/pull/55717, right?

Yep that's right.

Will you look at the filters? I noticed some behaviour that looked odd, but Roy is not around to confirm with for another couple days.

> It looks like that there is a regression regarding the Product Button. On Site 2 for a second, I see only a dark button without any text inside. This doesn't happen on Site 1.

@gigitux Is this a local test or JN or other? Are you sure you rebuilt properly? It looks like the item is not being added to cart as well as the flash. No animation etc. Which is not something I can repro on the branch right now. Can you also check console errors for me? Looks like something very wrong :D I'm sure you know this, but, running a watch build while swapping branches won't work, you'll need to restart the build.

gigitux:
> Will you look at the filters? I noticed some behaviour that looked odd, but Roy is not around to confirm with for another couple days.
> 
> >

I can take a look at them if needed.

>@gigitux Is this a local test or JN or other? Are you sure you rebuilt properly? It looks like the item is not being added to cart as well as the flash. No animation etc. Which is not something I can repro on the branch right now. Can you also check console errors for me? Looks like something very wrong :D I'm sure you know this, but, running a watch build while swapping branches won't work, you'll need to restart the build.

I built a dedicated zip and uploaded it on a JN site, so I don't think that it is a build issue. I tried again with another JN site and I'm able to reproduce it. Can you reproduce it visiting https://totally-maximum-bee.jurassic.ninja/shop/ ?

luisherranz:
>  Can you reproduce it visiting https://totally-maximum-bee.jurassic.ninja/shop/ ?

That is working well on my computer, both in Chrome and Safari.

gigitux:
> > Can you reproduce it visiting https://totally-maximum-bee.jurassic.ninja/shop/ ?
> 
> That is working well on my computer, both in Chrome and Safari.

Don't you see any difference with trunk? On trunk, the text is immediately present, while on this branch for a second I see the button without text. I tested with Edge and I'm able to see this difference:

- https://real-weaver.jurassic.ninja/shop/ -> WooCommerce trunk branch
- https://totally-maximum-bee.jurassic.ninja/shop/ -> WooCommerce dev/migrate-to-wp-interactivity-54259 branch


| Trunk | Interactivity API  |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/81616e42-3870-4770-99ba-0eda1a8db522" /> | <video src="https://github.com/user-attachments/assets/c23585e3-4535-4958-baa0-e64ce133e04d" /> |

samueljseay:
> Can you reproduce it visiting https://totally-maximum-bee.jurassic.ninja/shop/ ?

@gigitux Works for me too.  I tested Chrome and Firefox too just in case.

samueljseay:
@gigitux oh I see what you mean yup. Your initial screen recording looked really broken so it was a distraction. Yes I see the brief flash with nothing in the button

gigitux:
> @gigitux oh I see what you mean yup. Your initial screen recording looked really broken so it was a distraction. Yes I see the brief flash with nothing in the button

Sorry for the confusion! I'm glad that you're able to replicate it! 


Also, I noticed that a spinner appears on the trunk when the user clicks the button, but it is no longer visible on this branch: is this a desired change?

samueljseay:
@gigitux that's a good question regarding the spinner. I'll find out. I like the new animation better but I don't know if we got sign off to change that. 

EDIT: The reason it's changed is because we optimistically update how many items are in the cart and that was a way to animate reverting the state if it failed. It would be weird to flash back to the old number. I think I probably need to ask more broadly about the animation change though.

@luisherranz regarding the issue with the text being empty, it seems like it should be initially filled by `$initial_product_text` but it seems like it gets wiped out before render. Any chance you could take a look at that? Even if I try render some static text in the span it is gone at render.

gigitux:
> @gigitux that's a good question regarding the spinner. I'll find out. I like the new animation better but I don't know if we got sign off to change that.
> 
> EDIT: The reason it's changed is because we optimistically update how many items are in the cart and that was a way to animate reverting the state if it failed. It would be weird to flash back to the old number. I think I probably need to ask more broadly about the animation change though.

I'm fine with this change. Thanks for the clarification!

Regarding the filters block, in some cases (reproducible by changing the max price via the input) the text is duplicated, as shown in the video below:  

https://github.com/user-attachments/assets/8f28ebb7-b5a0-4f57-bf86-7f0e84385556

If correct, @luisherranz is already aware of this bug. However, since the Filters Block is still experimental, I think that this shouldn't block the current PR.

luisherranz:
> If correct, @luisherranz is already aware of this bug. However, since the Filters Block is still experimental, I think that this shouldn't block the current PR.

Yes, it is documented here:

- https://github.com/woocommerce/woocommerce/issues/54259#issuecomment-2606896670

> it seems like it should be initially filled by $initial_product_text but it seems like it gets wiped out before render. Any chance you could take a look at that?

I'll take a look at that 🙂

luisherranz:
> it seems like it should be initially filled by $initial_product_text but it seems like it gets wiped out before render. Any chance you could take a look at that?

Fixed in https://github.com/woocommerce/woocommerce/pull/55080/commits/68d4c16ebeb7aa0899bd7428775c588a3f3318c4.

samueljseay:
@luisherranz that's great, thank you. I'm curious why does server rendered text get wiped out before client side state is rendered? I expected it would work to have some server rendered text in there first.

luisherranz:
When the `data-wp-text` directive is evaluated on the server, it looks at the value that exists in `state.addToCardText` and updates the HTML with the value it finds. If that value is falsy, then it doesn't put any text.

The complexity here comes from the fact that the same HTML is being used for buttons that end up with `data-wp-text` and buttons that end up needing manual SSR, but when using the interactivity API, it's not necessary to do the SSR manually. You need to populate the state and derived state correctly, though.

More about Server Side Rendering in the Interactivity API in [this guide](https://developer.wordpress.org/block-editor/reference-guides/interactivity-api/core-concepts/server-side-rendering/).

samueljseay:
@luisherranz Thanks that makes sense

roykho:
Two issues I am noticing after the migration.

- In a mobile device (not dev tools), I can no longer tap "Filter products" to bring up the filters overlay.
- Once in the overlay, after selecting a filter such as the attribute filter "gray", the overlay closes itself instead of waiting for user to click/tap on "Apply" button. This is also what's causing the e2e test to fail.

I will take a look at it tomorrow but if you know/have any insights that would be great.

samueljseay:
@roykho Thanks for that, do you wanna block merge till we get that sorted?

In the meantime I will also look and see if there is an obvious issue or fix

samueljseay:
> In a mobile device (not dev tools), I can no longer tap "Filter products" to bring up the filters overlay.

@roykho I cannot reproduce this. On an Android device running Chromium browser opening and closing the overlay works perfectly for me.

> Once in the overlay, after selecting a filter such as the attribute filter "gray", the overlay closes itself instead of waiting 

for user to click/tap on "Apply" button. This is also what's causing the e2e test to fail.
This one I can reproduce

samueljseay:
@roykho I just wanted to note that if you follow: p1740583588057979/1740556868.896719-slack-C05GY7MD3ST and disable the router region in ProductFilters, the overlay does not auto close (and it will also fix the other bugs with duplicated filters).

So it looks like it would be related to the bug mentioned there.

@luisherranz do you have a suggestion of a path forward here?

roykho:
> Thanks for that, do you wanna block merge till we get that sorted?

@samueljseay I don't think we need to block it.

> I cannot reproduce this. On an Android device running Chromium browser opening and closing the overlay works perfectly for me.

@samueljseay I reproduced it on iPhone Safari. I tested trunk and it works there.

> I just wanted to note that if you follow: p1740583588057979/1740556868.896719-slack-C05GY7MD3ST and disable the router region in ProductFilters, the overlay does not auto close (and it will also fix the other bugs with duplicated filters).

@samueljseay but isn't this issue only if you have the filters inside the Product Collection block? In my testing case, it is in the Product Catalog template outside of the Product Collection block.

roykho:
@samueljseay - I just tested the navigation block in iPhone Safari and Chrome and tapping it also doesn't work for me. I wonder if this is the same issue I am seeing with the Product Filters.

samueljseay:
@roykho hmm interesting. I have only tested the filters nested inside product collection so maybe that's the difference?

samueljseay:
@roykho Chrome on iOS is actually just webkit webview so this could be a safari problem? 🤔  i will see if i can test with safari on my iPad

samueljseay:
I am EOD but tomorrow I'm thinking we'll be close to merge this, I just need to resolve conflicts and add a couple more changelog items. 

I want to see if I can repro the (possibly just safari) issue. We should maybe track it in an issue @roykho ?

roykho:
> In a mobile device (not dev tools), I can no longer tap "Filter products" to bring up the filters overlay.

@samueljseay I tested again back to back (trunk/this branch), I can't replicate it anymore. Strange!

So I think we're good to go.

luisherranz (approved):
Awesome work, folks 🙂👏

Issue #55082, opened 2025-02-04
[e2e tests] Adapt users management tests to multisite setups


adimoldovan:
The user management flows are a bit different on multisite installations and all the e2e tests covering creation or management of users are failing.
We should adapt these tests to work on a multisite environment as well.

Issue #55066, opened 2025-02-03
[e2e tests] Update core-profiler.spec.js to work on multisite setups


adimoldovan:
`core-profiler.spec.js` fails in multisite setups because because different permissions being required for such setup.

1. The tests that are visiting the plugins page need to run with a super-admin instead of just admin. By default the admin cannot see the plugins page. Or, the plugins page can be enabled via the Network Settings -> Enable administration menus -> Enable menus Plugins.

2. Once the plugins page access issue is solved, the part that deletes the plugins fails, because you cannot delete the plugin at site level. We can either update the URL and go to network level plugins page or skip this part for multisites.

PR #55086, opened 2025-02-04
[e2e tests] Skip tests in multisite setups

adimoldovan:

Some tests are not expected to work with multisite installations, because the flows are a bit different. Skip those tests, using the `IS_MULTISITE` environment variable.

Related:

- https://github.com/woocommerce/woocommerce/issues/55066
- https://github.com/woocommerce/woocommerce/issues/55082

Run the affected specs with the IS_MULTISITE env variable set. The tests should be skipped.

```
IS_MULTISITE=1 pnpm test:e2e core-profiler.spec.js users-create.spec.js users-manage.spec.js account-email-receiving.spec.js
```


lanej0:
> [!IMPORTANT] 
> This is an AI-generated review by https://github.com/Automattic/pr-review-agent/ 

Hey there! I've taken a look at the changes to skip certain tests in multisite setups. The approach looks good overall, but I have a couple of suggestions that might help improve the implementation:

1. Consider adding a brief comment in each skipped test explaining why it doesn't work in multisite. This would help other developers understand the limitations without having to dig through the related issues. For example:

```js
test.skip(
    process.env.IS_MULTISITE,
    'Test not working on a multisite setup - user management flows differ in multisite'
);
```

2. For better maintainability, you might want to consider creating a helper function for the multisite skip condition, something like:

```js
const skipInMultisite = (reason) => {
    return {
        skip: process.env.IS_MULTISITE,
        reason: `Test not working on a multisite setup - ${reason}`
    };
};
```

Then you could use it like:
```js
test.skip(skipInMultisite('user management flows differ in multisite'));
```

This would make the skip conditions more consistent across tests and make it easier to modify the skip behavior in the future if needed.

The testing instructions look good and clear. I can see that running `IS_MULTISITE=1 pnpm test:e2e core-profiler.spec.js users-create.spec.js users-manage.spec.js account-email-receiving.spec.js` will properly test the changes.

lanej0 (approved):
Looks good.

Small nitpick, but it might be good to add a comment with these skips to keep try of why (or why we think) a test doesn't work on a multisite setup. Just to help us keep track of what's going on, and if the problem could be addressed in the future.

adimoldovan:
> Looks good.
> 
> Small nitpick, but it might be good to add a comment with these skips to keep try of why (or why we think) a test doesn't work on a multisite setup. Just to help us keep track of what's going on, and if the problem could be addressed in the future.

Good point. Updated the skip reason with links to Github issues in [b6a0009](https://github.com/woocommerce/woocommerce/pull/55086/commits/b6a00097c05251694341b75ddff336dcdcf5fa02)

PR #55088, opened 2025-02-04
Update Coming Soon templates with improved styling and layout

### LINKED ISSUE #53981: Improve style consistency for coming soon templates
### Prerequisites

### Describe the bug

The styles for WooCommerce "Coming Soon" templates are inconsistent across different themes. This is primarily because each theme has its own set of custom styles, which can override or conflict with the template styles.

Ensuring complete consistency across all themes is challenging because:
1. Each theme includes unique styles that can conflict with or override default template styles.
2. Forcing uniform styles could make it difficult for users to customize the appearance to match their site's design.

However, we can aim to make the styles more consistent while allowing customization. This will improve the user experience by reducing discrepancies between themes and providing a more unified look out of the box.

### Expected behavior

The styles should be more consistent with the default template styles when using other themes.

### Actual behavior

The styles are inconsistent with the default template styles when using other themes.

### Steps to reproduce

1. Create a new WooCommerce site
2. Enable `coming-soon-newsletter-template` feature flag.
3. Use other block themes to see if the templates are consistent.

### WordPress Environment

n/a

### Isolating the problem

chihsuan:

Closes https://github.com/woocommerce/woocommerce/issues/53981

Before this PR, we have hardcoded styles in `entire-site.scss` for the coming soon page template. This enables us to have consistent styles across all themes. However, this also makes it hard for users or themes to customize the coming soon page styles, and it's not a common practice in Gutenberg. We should use block settings to customize the styles except for styles that are not possible to achieve with block settings. This PR is to improve the coming soon page template styles for better theme compatibility.

These changes will allow users or themes to customize the coming soon page styles more easily without having to override the hardcoded styles. The templates might look different on different themes, and some themes might not be compatible with our templates. However, I think we should let themes support our templates instead of forcing the templates to fit all themes.

The changes include:

1. Remove the hardcoded styles from `entire-site.scss`.
2. Remove `:has()` selector to optimize the performance. This can be removed now because we don't need to have reset styles across all themes.
3. Some layout adjustments to improve the styles.
4. Fix the login/logout button font-family so it can dynamically changed based on the theme font-family.

1. Create a fresh site with this branch
2. Enable coming soon mode for entire site
3. Open an incognito window and go to the site frontend
4. Check if the coming soon page is displayed correctly (different window size, different browser, different device)
5. Try using different themes and check if the coming soon page is displayed correctly. Some themes might not support our templates.

Themes that are tested well:

- TT3
- TT4
- TT5
- Storefront
- FSE Empire
- Shuttle Restaurant
- Shuttle

Screenshot 2025-02-04 at 17 49 51

Screenshot 2025-02-04 at 17 50 12

Themes that have some issues because they have unique styles that cause the layout issues:

- Next Bakery
- TT2
- TT

For devs, please confirm that the issues are expected because of the unique styles of the themes.

For example:
Screenshot 2025-02-04 at 17 04 35
Next Bakery has wide margins for .alignwide.

6. Enable `coming-soon-newsletter-template` feature flag
7. Go to coming soon template editor page `/site-editor.php?postId=woocommerce%2Fwoocommerce%2F%2Fcoming-soon&postType=wp_template&canvas=edit`
8. Confirm templates look good in `Sidebar > Templates > Design`
9. Select a different template from `Sidebar > Templates > Design`
10. Confirm templates look good in frontend (at least compatible with the above themes)

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


psealock:
With TT3 'm seeing some odd header space that isn't there on `trunk`. This is on my local instance since JN isn't working right now, not sure if that has something to do with it. Can you reproduce?

image

chihsuan:
> With TT3 'm seeing some odd header space that isn't there on `trunk`. This is on my local instance since JN isn't working right now, not sure if that has something to do with it. Can you reproduce?
> 
> image

Thanks for catching that! @psealock  I can reproduce it.

The issue is that when a theme has `useRootPaddingAwareAlignments` enabled in theme.json, it will add a custom padding to the site blocks element. On the trunk, we set the body background color to the coming soon color, so it's not visible. I've fixed it in d6c228c70df58995ef44de430c42b296b796a9ca. I had to use the `:has` selector, which explicitly targets the element, so it shouldn't cause performance issues.

psealock:
This is looking good @chihsuan, I've poked around with a few more themes and noticed the known issue with Twenty twenty-two. Is this an acceptable trade off or is there a plan in place to handle these cases? Or is it already something we see on `trunk`?

image

Emoza ✅ 
Classic WooCommerce ✅ 
Lights Store ✅

chihsuan:
> This is looking good @chihsuan, I've poked around with a few more themes and noticed the known issue with Twenty twenty-two. Is this an acceptable trade off or is there a plan in place to handle these cases? Or is it already something we see on `trunk`?
> 

@psealock I think it's an acceptable trade-off, but open to opinions. 🙏  The issue is that Twenty twenty-two has [custom styles that add padding around the root container](https://github.com/WordPress/twentytwentytwo/blob/trunk/style.css#L86-L108). Although it's possible to fix it by adding our custom styles to override it, I hesitate to do that because it's more like a theme-specific issue.

Any theme can have any custom styles that break the layout. The one way to ensure that the layout is not broken is by using `!important`, but it's not a good solution. In that case, users or theme developers will be unable to customize the styles.

Screenshot 2025-02-06 at 11 20 29

plugins/woocommerce/patterns/page-coming-soon-default.php
```diff
@@ -17,41 +17,40 @@
...
-<div class="wp-block-group woocommerce-coming-soon-banner-container"><!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"10px","bottom":"14px"}}},"className":"woocommerce-coming-soon-header","layout":{"type":"constrained"}} -->
-<div class="wp-block-group alignwide woocommerce-coming-soon-header has-background" style="padding-top:10px;padding-bottom:14px"><!-- wp:group {"align":"wide","layout":{"type":"flex","justifyContent":"space-between","flexWrap":"wrap"}} -->
+<div class="wp-block-woocommerce-coming-soon woocommerce-coming-soon-entire-site has-background" style="background-color:#bea0f2"><!-- wp:cover {"customOverlayColor":"transparent","isUserOverlayColor":true,"minHeight":100,"minHeightUnit":"vh","className":"coming-soon-is-vertically-aligned-center coming-soon-cover","style":{"spacing":{"padding":{"top":"0px","bottom":"0px","left":"24px","right":"24px"}}},"layout":{"type":"constrained","wideSize":"1280px"}} -->
```
ilyasfoo: I'm not sure if it's feasible, but could we have all new blocks to inherit the color from coming soon block instead? Currently all new blocks will inherit white color:

image

ilyasfoo (commented):
Thanks for cleaning up, @chihsuan! However, I think this is causing regression on preview height:

image

image

plugins/woocommerce/patterns/page-coming-soon-default.php
chihsuan: Yes, I think it should be possible to do. I'll fix it.

chihsuan:
> However, I think this is causing regression on preview height:

@ilyasfoo What theme were you using? I created a fresh site with this branch, and I can't reproduce it with The Twenty-Four theme.

plugins/woocommerce/patterns/page-coming-soon-default.php
chihsuan: Ok, fixed!

https://github.com/user-attachments/assets/6165cf71-c219-47bd-b8b0-1da443fec62e

Note: The issue is that the "Cover" block adds a default color style (white or black, depending on the `isDark` attribute) when the color is not set, so the color setting in the coming soon block is not applied. The block editor doesn't support the `inherit` value for the color setting in the editor UI, so we need to manually set the color to inherit for the Cover block.

ilyasfoo: Nice! This works amazingly 👍

ilyasfoo (approved):
> @ilyasfoo What theme were you using? I created a fresh site with this branch, and I can't reproduce it with The Twenty-Four theme.

Odd, I couldn't reproduce it anymore in a fresh site. I have no idea what happened 😅 

Tests well, LGTM! Thanks!

psealock (approved):
>@psealock I think it's an acceptable trade-off, but open to opinions. 🙏 The issue is that Twenty twenty-two has [custom styles that add padding around the root container](https://github.com/WordPress/twentytwentytwo/blob/trunk/style.css#L86-L108). Although it's possible to fix it by adding our custom styles to override it, I hesitate to do that because it's more like a theme-specific issue.

>Any theme can have any custom styles that break the layout. The one way to ensure that the layout is not broken is by using !important, but it's not a good solution. In that case, users or theme developers will be unable to customize the styles.

Makes sense to me, thanks ✅

Issue #54662, opened 2025-01-20
[API] Audit tests that were tagged with `SKIP_ON_PRESSABLE` and `SKIP_ON_WPCOM`


rodelgc:
API tests in `trunk` that are currently skipped in WPCOM and Pressable, and the PRs that address them:

| Test                            | PR     |
| ------------------------------- | ------ |
| `coupons.test.js`               | #55160 |
| `customers-crud.test.js`        | #55162 |
| `order-complex.test.js`         | #55429 |
| `orders.test.js`                | #55430 |
| `payment-gateways-crud.test.js` | #55163 |
| `product-list.test.js`          | #55359 |
| `products-crud.test.js`         | #55358 |
| `settings-crud.test.js`         | #55357 |
| `system-status-crud.test.js`    | #55168 |

rodelgc:
First batch of PRs that are part of this task are now submitted for review:

https://github.com/woocommerce/woocommerce/pull/55160
https://github.com/woocommerce/woocommerce/pull/55162
https://github.com/woocommerce/woocommerce/pull/55163
https://github.com/woocommerce/woocommerce/pull/55168

rodelgc:
Second and final batch of PRs ready for review:

[[API tests] Unskip settings-crud test on WPCOM and Pressable #55357](https://github.com/woocommerce/woocommerce/pull/55357)
[[API tests] Unskip products-crud on WPCOM and Pressable. #55358](https://github.com/woocommerce/woocommerce/pull/55358)
[[API tests] Unskip product-list on WPCOM and Pressable #55359](https://github.com/woocommerce/woocommerce/pull/55359)
[[API tests] Unskip order-complex.test.js on WPCOM and Pressable. #55429](https://github.com/woocommerce/woocommerce/pull/55429)
[[API tests] Unskip orders.test.js on WPCOM and Pressable. #55430](https://github.com/woocommerce/woocommerce/pull/55430)

Stojdza:
@rodelgc 👋 I see all PRs mentioned in this issue are merged. Are we good to mark this issue as done?

Issue #50758, opened 2024-08-19
[e2e tests] Remove the conversion of cart and checkout blocks to classic in global setup


adimoldovan:
More context: p1723638123619019-slack-C01BWDDTGKX

The Core e2e tests have a global setup step that convert the default cart and checkout to their shortcode version. The blocks version is the default since 8.3.
We should remove this step and base our e2e tests on the defaults.
We should add new tests to cover the functionality of classic Cart and Checkout.

adimoldovan:
Bumped to high prio, following this conversation: p1733418038012499-slack-C01BWDDTGKX

Stojdza:
There has been no progress on this in the last two weeks - keeping it a high priority.

juliaamosova:
We are going to keep the priority high but work on it as soon as someone on the tam completes one of the current projects.

rodelgc:
Surfacing here what Adi and I discussed in this [thread](https://github.com/woocommerce/woocommerce/pull/55089#discussion_r1951170717) that when implementing this task, and if we are still verifying the individual properties of `Local pickup` in `payment-gateways-crud.test.js` by then, then we'll need to change `local_pickup` with `pickup_location` in that API test.

PR #55089, opened 2025-02-04
[api tests] Fix tests for external environments and multisite

adimoldovan:

Fixes the `payment-gateways-crud.test.js` API tests to work with external environments. Split the assertion into smaller ones to be easier to spot what's actually different in case of failure.

Relates to https://github.com/woocommerce/woocommerce/issues/54662

Run the `payment-gateways-crud.test.js` spec in different envs.


plugins/woocommerce/tests/e2e-pw/tests/api-tests/payment-gateways/payment-gateways-crud.test.js
```diff
@@ -1,180 +1,177 @@
...
+					options: expect.objectContaining( {
+						'Flat rate': {
+							flat_rate: 'Any &quot;Flat rate&quot; method',
+						},
+						'Free shipping': {
+							free_shipping:
+								'Any &quot;Free shipping&quot; method',
+						},
+						'Local pickup': {
+							local_pickup: 'Any &quot;Local pickup&quot; method',
```
rodelgc, line 158: In #55163, I listed there a few environments where `pickup_location` was shown as the key name rather than `local_pickup`. I wasn't able to pinpoint exactly what's causing this weird variation. By any chance, did you happen to discover the cause of this inconsistency? If so, we can probably add a step in beforeAll to make sure that it always shows up as `local_pickup`, then close my #55163 in favor of this PR.

adimoldovan: I ran the test in all our currently supported environments and it passed: wp-env, default-wpcom, default-pressable, plus JN and Local multisite. This is the default value, which is what the test is checking. Any small update to the env (e.g. adding a shipping zone) will make the test fail - makes me wonder if we want this test after all.
I find it strange that you got such different results. I also [responded](https://github.com/woocommerce/woocommerce/pull/55163#issuecomment-2639074150) in your PR.

rodelgc: I think I finally figured out what's causing the `local_pickup` and `pickup_location` mystery. If the checkout page is either the shortcode one or the Classic Checkout one, then the field name would be `local_pickup`, which is the default in wp-env. 

But if the checkout page is using the Checkout Block, the field name would be `pickup_location`. Let me know if it's the behavior you experience as well even on wp-env.

I'm not sure how to proceed from here though. Honestly, I don't find my solution in https://github.com/woocommerce/woocommerce/pull/55163 to be an elegant one because it felt like a bit of an overkill to have a custom matcher that will only be used once in the code.

So far what I'm thinking is maybe to make sure the checkout page is using the shortcode version prior to running this test. However, I already forgot whether we run the API tests with 4 workers or just 1. If we're using more than 1 worker for the api tests, two parallel tests that rely on both versions of the checkout page might clash with each other.

Another approach is to query the checkout page first if it's currently shortcode or block, then set the key name based on that. Third approach could be `Local pickup: expect.any( Object )` 😄 

> makes me wonder if we want this test after all.

The main goal of this test is to verify the payment gateways, so maybe it makes sense if we loosen up the assertions on some of the object properties like those already related to shipping or something else?

adimoldovan: @rodelgc 

> Let me know if it's the behavior you experience as well even on wp-env.

Indeed, I experienced the same behaviour. Nice catch!

> I'm not sure how to proceed from here though. Honestly, I don't find my solution in https://github.com/woocommerce/woocommerce/pull/55163 to be an elegant one because it felt like a bit of an overkill to have a custom matcher that will only be used once in the code.

Given that we're [converting the checkout page to shortcode in a global setup](https://github.com/woocommerce/woocommerce/blob/854137c204d0a6695500cb97ee4e9e4717ad8527/plugins/woocommerce/tests/e2e-pw/fixtures/site.setup.js#L66), I say we're good to proceed with `local_pickup` here. We're expecting the default value, and that's the default value. With [50758](https://github.com/woocommerce/woocommerce/issues/50758) we'll have to change this. 

> Another approach is to query the checkout page first if it's currently shortcode or block, then set the key name based on that.

This is exactly what I would avoid. Any other field from the response can be changed by some setup update in the site. If we're running this on a site that's not freshly set up it will definitely fail. Hence my question about what we want to test here. The default setup? The fact that the endpoint returns something? The same question can be asked about many more tests in the API suite though, because they work fine in wp-env in CI, but in the expanding list of environments we cannot say the same. But we should take this discussion outside of this PR.

rodelgc: > Given that we're [converting the checkout page to shortcode in a global setup](https://github.com/woocommerce/woocommerce/blob/854137c204d0a6695500cb97ee4e9e4717ad8527/plugins/woocommerce/tests/e2e-pw/fixtures/site.setup.js#L66). I say we're good to proceed with local_pickup here.

I see, yeah that makes sense. I briefly searched the api tests just now to see if we have any that alters the checkout page and we don't seem to have one so yeah, we're covered here. I'll approve this one and close https://github.com/woocommerce/woocommerce/pull/55163.

> But we should take this discussion outside of this PR.

Agree! And thanks for raising those good points.

rodelgc (approved):

PR #55110, opened 2025-02-04
Fixing `shipping` endpoint response

lbonomo:
Shipping endpoint don't return `X-WP-Total` header.

```bash
$ wp wc shipping_method list --user=admin --format=count
Warning: Undefined array key "X-WP-Total" in /var/www/html/wp-content/plugins/woocommerce/includes/cli/class-wc-cli-rest-command.php on line 239
```

1. Go to WC > Settings > Shipping zones.
2. If you only have the default "Rest of the world" zone, make sure to create at least one more associated to a few regions.
3. Make sure to enable a few shipping methods for said zone.
4. Take note of this new zone ID (look at the `zone_id` parameter in the URL while editing it).
5. Run the following commands using WP-CLI. No error should be logged or printed while doing so.
   ```
   wp> wp wc shipping_method list --user=admin --format=count
   wp> wp wc shipping_zone list --user=admin --format=count
   wp> wp wc shipping_zone_location list <YOUR_ZONE_ID> --user=admin --format=count
   wp> wp wc shipping_zone_method list <YOUR_ZONE_ID> --user=admin --format=count
   ```

   In particular, error `Warning: Undefined array key "X-WP-Total"` should not appear.
6. (Optional)
   1. Use the request software of your choice to make GET requests to:
      - `<yoursite>/wp-json/wc/v3/shipping_methods/`
      - `<yoursite>/wp-json/wc/v3/shipping/zones/`
      - `<yoursite>/wp-json/wc/v3/shipping/zones/<YOUR_ZONE_ID>/locations`
      - `<yoursite>/wp-json/wc/v3/shipping/zones/<YOUR_ZONE_ID>/methods`

      Make sure the requests work correctly and that, in all cases, headers `X-WP-Total` (reflecting the number of items returned) and `X-WP-TotalPages` are set.


plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-shipping-methods-v2-controller.php
```diff
@@ -101,14 +101,19 @@ public function get_item_permissions_check( $request ) {
 	 * @return WP_Error|WP_REST_Response
 	 */
 	public function get_items( $request ) {
+
```
jorgeatorres: This empty line produces a linting error. See https://github.com/woocommerce/woocommerce/actions/runs/13146532631/job/36987499029?pr=55110 for the details.

plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-shipping-methods-v2-controller.php
```diff
@@ -101,14 +101,19 @@ public function get_item_permissions_check( $request ) {
...
-			$response[] = $method;
+			$method = $this->prepare_item_for_response( $shipping_method, $request );
+			$method = $this->prepare_response_for_collection( $method );
+			$data[] = $method;
 		}
-		return rest_ensure_response( $response );
+
+		$total    = count( $data );
+		$response = rest_ensure_response( $data );
+		$response->header( 'X-WP-Total', $total );
```
jorgeatorres: It'd be a good opportunity to [also set `X-WP-TotalPages`](https://developer.wordpress.org/rest-api/using-the-rest-api/pagination/) for this and the other requests (obviously set to `1`). Even though this isn't used by the WP-CLI tool, it could be useful to proper REST requests.

jorgeatorres (changes requested):
Hi @lbonomo!

Thanks for your contribution and apologies for the delay in getting back to you. This works great. There are only a few details before we can merge:

- You need to include a changelog entry with this PR. You can create one from your copy of the repo by running `pnpm --filter='@woocommerce/plugin-woocommerce' changelog add` and then committing the resulting file.
- Some minor feedback left inline as part of the review.

Thanks again for contributing and looking forward to your response.

plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-shipping-methods-v2-controller.php
lbonomo: Done

plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-shipping-methods-v2-controller.php
lbonomo: Done

jorgeatorres (approved):
Hey @lbonomo!

Thanks for updating your PR. Things look good to me and so I'm approving.

I made some edits to your code to account for cases where `$total` might be 0. In this case, we still want `X-WP-TotalPages` to be 0 (not 1). This is something I overlooked during my prior review, but decided to address it on my own rather than asking you as the changes were tiny.

Thanks again for contributing! 🥇 Hope to see more of you around here. ¡Gracias!

PR #55115, opened 2025-02-05
[Settings] support for 'info' settings type

### LINKED ISSUE #54585: [Settings] Handle type 'info'
Add support for Setting type `info`. We don't have many examples of usage, but its handled as additional text The info field type is non-interactive and serves as a visual or textual element to inform users.

https://github.com/woocommerce/woocommerce/blob/7f513fe2e50a31bfb0687e55dce9ae8c857976cc/plugins/woocommerce/includes/admin/class-wc-admin-settings.php#L296-L303

The `text` property is passed through `wp_kses_post` so ensure we support rendering HTML, such as links.

chihsuan:

Closes https://github.com/woocommerce/woocommerce/issues/54585

Add support for Setting type `info`. It's used to display non-interactive information in the settings page.

For the classic settings page, it's rendered as:

https://github.com/woocommerce/woocommerce/blob/7f513fe2e50a31bfb0687e55dce9ae8c857976cc/plugins/woocommerce/includes/admin/class-wc-admin-settings.php#L296-L303

In this PR, I try to render it in the same way as classic settings page; However, it's not fully aligned:

-   We don't render content in `<tr>`, `<td>` tags since we're rendering it as a regular dataform field.
-   The `row_class` wraps the value instead of the entire content (title + value).

Additionally, I replaced the `classnames` with `clsx` pdnLyh-6ic-p2

1. Add/Enable Gutenberg 19.8 or older
2. Add/Enable [Settings Tester](https://github.com/woocommerce/woocommerce-settings-migration-tester), making sure you have the latest commits. You can create a zip by running `npm run plugin-zip` from project root.
3. Add/Enable `Code Snippets` and add the following snippet:

```php

add_filter( 'settings_tester_get_settings_settings-tester', function( $settings ) {
	$settings[] = array(
		'title' => __( 'Info Message', 'settings-tester' ),
		'type'  => 'title',
		'desc'  => '',
		'id'    => 'settings_tester_info_message',
	);

	$settings[] = array(
		'type' => 'info',
		'title' => __( 'Info', 'settings-tester' ),
		'text'  => implode( "\n\n", array(
			'This is a info message.',
			'<code>This is a code block.</code><a href="#">This is a link.</a><strong>This is a strong text.</strong>',
		) ),
		'css'   => 'background-color: yellow;',
		'row_class' => 'settings-tester-info-message',
	);

	$settings[] = array(
		'type' => 'sectionend',
		'id'   => 'settings_tester_info_message',
	);
	return $settings;
} );
```

4. Turn on the Settings feature flag
5. Go to `/wp-admin/admin.php?page=wc-settings&tab=settings-tester`
6. Confirm that the Info field is rendered with yellow background and class name is `settings-tester-info-message`

Screenshot 2025-02-05 at 14 56 00

Screenshot 2025-02-05 at 14 55 56

Screenshot 2025-02-05 at 14 57 30

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

support for 'info' settings type 
</details>


plugins/woocommerce/src/Admin/Features/Settings/Transformer.php
```diff
@@ -120,6 +120,17 @@ private function process_setting( array $setting, array &$transformed_settings )
 				$this->handle_checkbox_setting( $setting, $transformed_settings );
 				break;
 
+			case 'info':
+				if ( ! empty( $setting['text'] ) ) {
+					$setting['text'] = wp_kses_post( wpautop( wptexturize( $setting['text'] ) ) );
+				}
+				if ( ! empty( $setting['row_class'] ) && substr( $setting['row_class'], 0, 16 ) !== 'wc-settings-row-' ) {
+					$setting['row_class'] = 'wc-settings-row-' . $setting['row_class'];
+				}
```
chihsuan, line 129: https://github.com/woocommerce/woocommerce/blob/a60e66c2c14fc5effb5c3c967d386796ade6d082/plugins/woocommerce/includes/admin/class-wc-admin-settings.php#L253-L255

psealock (approved):
This is all looking great, thanks @chihsuan. I like the improvements to `clsx` and `sanitizeHTML`.

One thing to note is we haven't been supporting CSS properties in any of the types, but maybe this should be an exception? If we add it here, the CSS manipulation functions added could go directly into the utils folder. What do you think? Approving in either case.

chihsuan:
> One thing to note is we haven't been supporting CSS properties in any of the types, but maybe this should be an exception?

Yeah, I think it's a good idea to add CSS support to the info type. 

>  If we add it here, the CSS manipulation functions added could go directly into the utils folder. 

Sounds good. I'll move the CSS manipulation functions to the utils folder.

chihsuan:
> Sounds good. I'll move the CSS manipulation functions to the utils folder.

Done in f040076ce3be0e0cde788a38ae0693fbd327328d. I also added some tests for it.

packages/js/settings-editor/src/utils/index.ts
```diff
@@ -2,11 +2,86 @@
...
+ * @param css - CSS string in format "property: value; property2: value2"
+ * @return Object with camelCase keys and corresponding values
+ */
+/**
+ * Converts a CSS string to a React style object.
+ * Handles conversion from kebab-case to camelCase for CSS properties.
+ *
+ * @param css - CSS string in format "property: value; property2: value2"
+ * @return Object with camelCase keys and corresponding values
+ */
```
ilyasfoo: Duplicate
```suggestion
```

packages/js/settings-editor/src/utils/index.ts
```diff
@@ -2,11 +2,86 @@
...
+export const cssStringToObject = ( css: string ): Record< string, string > => {
```
ilyasfoo: I think trying to convert CSS strings to react object ourselves is risky. The code can be more complicated than expected. For example, ChatGPT found the following issues. IMO if there's a library that already handle this, it'll be great to leverage it.

#### Not handling special vendor prefix

[example](https://github.com/user-attachments/assets/7b6b3768-695c-4ea1-a560-1d7c266c2263)

```js
const testCSS = `
	-webkit-transform: rotate(45deg);
	-moz-transform: rotate(45deg);
	-o-transform: rotate(45deg);
	-ms-transform: rotate(45deg);
	transform: rotate(45deg);
	-webkit-user-select: none;
	-moz-user-select: none;
	-ms-user-select: none;
	user-select: none;
`;

console.log(cssStringToObject(testCSS));
```

ChatGPT suggested the following code (untested):

```js
// Handle all known vendor prefixes
if (camelKey.startsWith('Webkit')) camelKey = 'Webkit' + camelKey.slice(6);
if (camelKey.startsWith('Moz')) camelKey = 'Moz' + camelKey.slice(3);
if (camelKey.startsWith('O')) camelKey = 'O' + camelKey.slice(1);
if (camelKey.startsWith('Ms')) camelKey = 'ms' + camelKey.slice(2); // React uses lowercase `ms`
```

#### Not stripping `!important`

[AFAIK react doesn't support `!important`](https://github.com/facebook/react/issues/1881#issue-38233034), we'll need to strip it

#### Strings with colons:

[example](https://github.com/user-attachments/assets/968656f0-8cc9-4b42-8649-cfaaa25d07c1)

```
console.log(cssStringToObject('content: "Hello: World";'));
```

ilyasfoo (commented):
@chihsuan Tests well, but I have a comment on CSS function:

packages/js/settings-editor/src/utils/index.ts
adrianduffell: >  if there's a library that already handle this

Maybe have the browser do it?  🤔 

```
function cssStringToObject(cssString) {
    const el = document.createElement("div");
    el.style.cssText = cssString;
    document.body.appendChild(el); 
    const styles = el.style;
    document.body.removeChild(el); 
    return styles;
}
```

chihsuan: > Maybe have the browser do it? 🤔

Thanks for the suggestion! I'll check it out!

chihsuan: > ```
> function cssStringToObject(cssString) {
>     const el = document.createElement("div");
>     el.style.cssText = cssString;
>     document.body.appendChild(el); 
>     const styles = el.style;
>     document.body.removeChild(el); 
>     return styles;
> }
> ```

It's possible to use it. But the function returns a [CSSStyleDeclaration](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration) object, which is not a plain object, so we need to convert it to a plain object. It doesn't provide a native way to do this, so we need to loop and filter.

After some research, I found that we can use `ref` to set the style directly for the div element. It's compatible with the current implementation and relatively simple. Let me know your thoughts.

d84c03898251cd73bd03cf623dcca52ddb25f4f4

ilyasfoo (approved):
> After some research, I found that we can use ref to set the style directly for the div element. It's compatible with the current implementation and relatively simple. Let me know your thoughts.

Awesome, this looks more reliable IMO. Tests well, thanks!

PR #55118, opened 2025-02-05
Migrate blueprint feature flag to core feature system

### LINKED ISSUE #55114: Blueprint - migrate feature flag to core
## Issue Description

Migrate the Blueprint feature flag from WCA to the core feature system. This will allow it to be disabled via UI under advanced settings. Additionally, enable the feature by default.

## Acceptance Criteria

* List Blueprint in Settings > Advanced > Features
* Use the name "Blueprint (Beta)"
* Use ID/keys of "blueprint" to ensure we can remove the beta label in the future.
* Use the description "Enable blueprint to import and export settings in bulk"
* Enable the feature by default

chihsuan:

Closes https://github.com/woocommerce/woocommerce/issues/55114

This PR migrates the Blueprint feature flag from WCA to the core feature system.

> * List Blueprint in Settings > Advanced > Features
> * Use the name "Blueprint (beta)"
> * Use ID/keys of "blueprint" to ensure we can remove the beta label in the future.
> * Use the description "Enable blueprint to import and export settings in bulk"
> * Enable the feature by default

1. Create a fresh site with this branch
2. Go to `Settings > Advanced > Features`
3. See "Blueprint (beta)" in the list and it's disabled by default
4. Turn on the feature and go to `Settings > Advanced > Blueprint`
5. See the new Blueprint page
6. Try to export settings, and it should work
7. For devs, you should see blueprint cli commands available by running `wp wc blueprint`
8. Disable the feature, and blueprint page should not be accessible

Screenshot 2025-02-05 at 16 01 24

```bash
> wp wc blueprint
usage: wp wc blueprint export <save-to> [--steps=<steps>] [--format=<format>]
   or: wp wc blueprint import <schema-path> [--show-messages=<show-messages>]

See 'wp help wc blueprint <command>' for more information on a specific command.
```

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/admin/config/core.json
+++ b/plugins/woocommerce/client/admin/config/core.json
@@ -38,7 +38,6 @@
 		"async-product-editor-category-field": false,
 		"launch-your-store": true,
 		"product-editor-template-system": false,
-		"blueprint": false,
 		"reactify-classic-payments-settings": false,
 		"use-wp-horizon": false,
 		"add-to-cart-with-options-stepper-layout": true,
--- a/plugins/woocommerce/includes/class-wc-cli.php
+++ b/plugins/woocommerce/includes/class-wc-cli.php
@@ -53,7 +53,7 @@ class WC_CLI {
 		$cli_runner = wc_get_container()->get( ProductAttributesLookupCLIRunner::class );
 		WP_CLI::add_hook( 'after_wp_load', fn() => \WP_CLI::add_command( 'wc palt', $cli_runner ) );
 
-		if ( class_exists( \Automattic\WooCommerce\Blueprint\Cli::class ) ) {
+		if ( FeaturesUtil::feature_is_enabled( 'blueprint' ) && class_exists( \Automattic\WooCommerce\Blueprint\Cli::class ) ) {
 			WP_CLI::add_hook( 'after_wp_load', 'Automattic\WooCommerce\Blueprint\Cli::register_commands' );
 		}
 	}
@@ -62,10 +62,6 @@ class WC_CLI {
 	 * Include Blueprint CLI if it's available.
 	 */
 	private function maybe_include_blueprint_cli() {
-		if ( ! function_exists( 'wc_admin_get_feature_config' ) ) {
-			require_once WC_ABSPATH . 'includes/react-admin/feature-config.php';
-		}
-
 		if ( FeaturesUtil::feature_is_enabled( 'blueprint' ) ) {
 			require_once dirname( WC_PLUGIN_FILE ) . '/vendor/woocommerce/blueprint/src/Cli.php';
 		}
--- a/plugins/woocommerce/includes/class-wc-cli.php
+++ b/plugins/woocommerce/includes/class-wc-cli.php
@@ -8,6 +8,7 @@
 
 use Automattic\WooCommerce\Database\Migrations\CustomOrderTable\CLIRunner as CustomOrdersTableCLIRunner;
 use Automattic\WooCommerce\Internal\ProductAttributesLookup\CLIRunner as ProductAttributesLookupCLIRunner;
+use Automattic\WooCommerce\Utilities\FeaturesUtil;
 
 defined( 'ABSPATH' ) || exit;
 
@@ -65,8 +66,7 @@ class WC_CLI {
 			require_once WC_ABSPATH . 'includes/react-admin/feature-config.php';
 		}
 
-		$features = wc_admin_get_feature_config();
-		if ( isset( $features['blueprint'] ) ) {
+		if ( FeaturesUtil::feature_is_enabled( 'blueprint' ) ) {
 			require_once dirname( WC_PLUGIN_FILE ) . '/vendor/woocommerce/blueprint/src/Cli.php';
 		}
 	}
--- a/plugins/woocommerce/src/Admin/Features/Features.php
+++ b/plugins/woocommerce/src/Admin/Features/Features.php
@@ -8,6 +8,7 @@ namespace Automattic\WooCommerce\Admin\Features;
 use Automattic\WooCommerce\Admin\PageController;
 use Automattic\WooCommerce\Internal\Admin\Loader;
 use Automattic\WooCommerce\Internal\Admin\WCAdminAssets;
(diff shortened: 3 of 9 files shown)


plugins/woocommerce/src/Admin/Features/Features.php
```diff
@@ -146,6 +147,10 @@ public static function load_features() {
 				new $feature_class();
 			}
 		}
+
+		if ( FeaturesUtil::feature_is_enabled( 'blueprint' ) ) {
+			new \Automattic\WooCommerce\Admin\Features\Blueprint\Init();
+		}
```
chihsuan, line 153: WCA feature flag system automatically initializes the feature class when the feature is enabled but core system doesn't do this so we need to manually initialize the feature class. I think we should initialize the feature class in the same place as WCA does so it's easier to maintain. Open to suggestions.

adrianduffell: Another option I see is to initialize it with the [DI container in the WooCommerce class](https://github.com/woocommerce/woocommerce/blob/ccc6eb86c76bafdb642f09c5f694e801dee9e414/plugins/woocommerce/includes/class-woocommerce.php#L314-L315). I'm good with what you have though.

plugins/woocommerce/src/Internal/Features/FeaturesController.php
```diff
@@ -312,6 +312,26 @@ private function get_feature_definitions() {
 						'woocommerce'
 					),
 				),
+				'blueprint'              => array(
+					'name'               => __( 'Blueprint (Beta)', 'woocommerce' ),
```
adrianduffell:
```suggestion
					'name'               => __( 'Blueprint (beta)', 'woocommerce' ),
```

Apologies, could we make this sentence case?

adrianduffell (approved):
Thanks @chihsuan, LGTM and tested well! I left a couple suggestions.

Pre-approving! 🚀

plugins/woocommerce/src/Admin/Features/Features.php
chihsuan: Thanks for the suggestion! 🙌   On second thought, I think it's better to keep it as is so we only initialize it in the admin context.

plugins/woocommerce/includes/class-wc-cli.php
```diff
@@ -65,8 +66,7 @@ private function maybe_include_blueprint_cli() {
 			require_once WC_ABSPATH . 'includes/react-admin/feature-config.php';
 		}
 
-		$features = wc_admin_get_feature_config();
-		if ( isset( $features['blueprint'] ) ) {
+		if ( FeaturesUtil::feature_is_enabled( 'blueprint' ) ) {
```
ilyasfoo, line 65: Previously this checks existence of the feature, but now it's checking if it's enabled. I think the change intended, but just making sure

ilyasfoo: May not be related to this PR, but is it intended for the cli command to be available despite disabling the feature?

image

ilyasfoo (approved):
Tests well! I have minor comments but pre-approving

plugins/woocommerce/client/admin/config/core.json
```diff
@@ -38,7 +38,6 @@
 		"async-product-editor-category-field": false,
 		"launch-your-store": true,
 		"product-editor-template-system": false,
-		"blueprint": false,
```
ilyasfoo, line 41: @chihsuan  1 more thing to confirm, previously blueprint is disabled by default in production build, but now I assume it's not (essentially launching the feature in the next version)?

chihsuan: Yes, I think this is what we want to do according to the issue description.  @adrianduffell To confirm, do we still want to turn it on by default? As you just mentioned we may push back the blueprints MVP date.

plugins/woocommerce/includes/class-wc-cli.php
chihsuan: > May not be related to this PR, but is it intended for the cli command to be available despite disabling the feature?

That's a good question! I didn't check this. I thought it should be disabled when feature flag is off. @moon0326 Could you confirm if this is the intended behavior?

plugins/woocommerce/client/admin/config/core.json
adrianduffell: @chihsuan It's looking likely that we will push the release date back, I'll know later today. If needed, You could merge it with it disabled by default for now.

plugins/woocommerce/includes/class-wc-cli.php
adrianduffell: Wonder if we should actually put the CLI command under a separate feature flag (non-UI) and release it later on? I'm thinking the command adds a bit too much for us to handle in MVP 🤔

adrianduffell: >  I thought it should be disabled when feature flag is off

@chihsuan, yeah that's correct - it should be controlled by a feature flag

plugins/woocommerce/client/admin/config/core.json
chihsuan: Okay, I'll update the PR to disable it by default. Thanks for confirming!

chihsuan: Updated in 645467c4fbb04f6371f94a21632152d4b01b2cab.

plugins/woocommerce/includes/class-wc-cli.php
chihsuan: Thanks, I've fixed conditional checks for Blueprint CLI in 67bf137e3e92ef63dcf2f9ef1d446a5caf9b01b0.

> Wonder if we should actually put the CLI command under a separate feature flag (non-UI) and release it later on?

IMO it's fine to release it together because it seems that CLI and Blueprint REST API share most of the code and CLI only provides two commands (import/export). cc @moon0326 since you have the best understanding of the code. 🙏 

I'll merge this PR first and we can do it in a follow-up PR if we want to change this.

PR #55129, opened 2025-02-05
Add `$args` to the `woocommerce_hpos_pre_query` filter

### LINKED ISSUE #55126: [Enhancement]: Add OrdersTableQuery::args as a parameter of the woocommerce_hpos_pre_query filter

Currently, people using the `woocommerce_hpos_pre_query` can't access the initial $args passed to the OrdersTableQuery class constructor, as that is a private attribute. Although we do have access to the final SQL, having access to the args array would make it much easier to build, for example, an equivalent Elasticsearch query of the request being run.

If I'm not missing any obvious detail, this should be as simple as changing
```php
		 * @param array|null $order_data {
		 *     An array of order data.
		 *     @type int[] $orders        Return an array of order IDs data to short-circuit the HPOS query,
		 *                                or null to allow HPOS to run its normal query.
		 *     @type int   $found_orders  The number of orders found.
		 *     @type int   $max_num_pages The number of pages.
		 * }
		 * @param OrdersTableQuery   $query The OrdersTableQuery instance.
		 * @param string             $sql The OrdersTableQuery instance.
		 */
		$pre_query = apply_filters( 'woocommerce_hpos_pre_query', null, $this, $this->sql );
```

to something like this:

```php
		 * @param array|null $order_data {
		 *     An array of order data.
		 *     @type int[] $orders        Return an array of order IDs data to short-circuit the HPOS query,
		 *                                or null to allow HPOS to run its normal query.
		 *     @type int   $found_orders  The number of orders found.
		 *     @type int   $max_num_pages The number of pages.
		 * }
		 * @param OrdersTableQuery   $query The OrdersTableQuery instance.
		 * @param string             $sql   Fully built SQL query.
		 * @param array              $args  Query arguments.
		 */
		$pre_query = apply_filters( 'woocommerce_hpos_pre_query', null, $this, $this->sql, $this->args );
```

_No response_

### Additional context

_No response_

felipeelia:

Add OrdersTableQuery::args as a parameter of the woocommerce_hpos_pre_query filter.

Currently, people using the `woocommerce_hpos_pre_query` can't access the initial $args passed to the OrdersTableQuery class constructor, as that is a private attribute. Although we do have access to the final SQL, having access to the args array would make it much easier to build, for example, an equivalent Elasticsearch query of the request being run.

Closes #55126

1.
2.
3.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


vikrampm1:
@naman03malhotra This PR is ready for your review. Please review as soon as you're able, and merge it once it's good to go!

naman03malhotra (commented):
👋 Thanks for this PR. Can you please provide and example of the use-case how this change will be leveraged by the users?

felipeelia:
Hi @naman03malhotra. The main reason why we need that extra argument is https://github.com/10up/ElasticPress/issues/3476.

In summary, the `woocommerce_hpos_pre_query` was introduced so developers could avoid the regular SQL query and fetch orders in a different way (for ElasticPress, that would be an Elasticsearch query.)

It turns out that when you use that filter, you really don't have access to the arguments passed, as `args` is a private attribute. Having access only to the final SQL query generated is not that useful, as we would have to parse the entire query.

Let me know if that is context enough and if you need any extra info. Thanks!

felipeelia:
Hello again @naman03malhotra! Sorry for the ping, but is there any additional context we can provide to help move this one forward? We'd love to have this added before we release the new version of ElasticPress :) Thanks in advance!

naman03malhotra (commented):
This change looks innocent to me and the added $args parameter offers enhanced flexibility, however we had cases in the past where it was hard to use to change/rollback when we extended the params to public.

Also, for instance, any code written to explicitly expect only three arguments might break or trigger unexpected behavior—such as callbacks that use func_num_args() or func_get_args() and assume a fixed count, though rarely seen. Additionally, exposing the internal $args array risks external code becoming tightly coupled to WooCommerce’s internal query structure, which could lead to compatibility issues if that structure ever changes, adding additonal maintainence for us.

As this is a critical component, I'd like to get a second opinion on this change, I'll request an additonal review from my team to make sure we are shipping a stable chanage.

jorgeatorres (commented):
Hey @naman03malhotra! Thanks for the ping! And thank you @felipeelia for your contribution. I don't see anything wrong with sharing the args here as part of the filter. Luckily, this is just an array and not a more complicated structure (such as the `OrdersTableQuery` instance itself, which we're already passing as an argument).

That said, precisely because we're already passing the query instance, I wonder if we should _instead_ add a method to this class such as `get_args()` (or similar) instead of adding a new parameter to the filter.

What are your thoughts @naman03malhotra @felipeelia?

felipeelia:
I'm okay either way @jorgeatorres. IDK if I'll have time to update the PR (especially tests) if we decide to have a method instead, but it would certainly help us with our needs :)

jorgeatorres:
> I'm okay either way @jorgeatorres. IDK if I'll have time to update the PR (especially tests) if we decide to have a method instead, but it would certainly help us with our needs :)

No worries @felipeelia. Let's see what @naman03malhotra thinks and we'll either go with your current approach or we can update things ourselves. Thanks for your patience!

naman03malhotra (commented):
> I wonder if we should instead add a method to this class such as get_args() (or similar) instead of adding a new parameter to the filter.

Thank for the review Jorge. I would personally vote for this approach, if this isn't big of a change.

felipeelia:
@naman03malhotra @jorgeatorres I've updated the PR adding a new get_args method. As I've written that (super small) unit test blindly, I hope it doesn't require much change from you guys. Let me know if there is anything else I can do to help move this along!

jorgeatorres:
Hi @felipeelia!

I've renamed the method to `get_query_args()` and made it return the _original_ query vars used to initialize the object, instead of the query args after processing (`$args`).
The reason for this is that I'm a bit wary of exposing `$args` itself as it can contain metaqueries or other variables that were filled in internally and not part of the original query. If code starts using those variables to "figure out" what the query was, that might break if we later change the internal implementation.

Does `get_query_args()` as implemented in my code work for your use case?

felipeelia:
I think it can go both ways, @jorgeatorres 

Although it would be better to have the initial args in some cases, we could also need the fully changed args to use default values applied throughout the implementation. For example, let's say args does not contain an orderby parameter but the code injects a default orderby later, it would be good to get that info as well to keep parity between functionalities. Or, as another example, $args will have the sanitized status values, while $query_args will not.

I'd say having get_query_args already unblocks us but if we could have access to both versions of the attribute, that would be great.

naman03malhotra (commented):
>. For example, let's say args does not contain an orderby parameter but the code injects a default orderby later, it would be good to get that info as well to keep parity between functionalities. Or, as another example, $args will have the sanitized status values, while $query_args will not.

Valid point, but let's go with the current implemented approach, if we came across a usecase where the original args are needed, we can raise another PR for that, wdyt?

felipeelia:
wfm @naman03malhotra !

plugins/woocommerce/tests/php/src/Internal/DataStores/Orders/OrdersTableQueryTests.php
```diff
@@ -597,4 +597,16 @@ public function test_query_s_filters_customer_email() {
 		$query = new OrdersTableQuery( $query_args );
 		$this->assertEqualsCanonicalizing( array( $orders[0] ), $query->orders );
 	}
+
+	/**
+	 * @testDox The get_args method returns the initial args sent.
+	 */
+	public function test_get_args() {
```
naman03malhotra:
```suggestion
	public function test_get_query_args() {
```

plugins/woocommerce/tests/php/src/Internal/DataStores/Orders/OrdersTableQueryTests.php
```diff
@@ -597,4 +597,16 @@ public function test_query_s_filters_customer_email() {
 		$query = new OrdersTableQuery( $query_args );
 		$this->assertEqualsCanonicalizing( array( $orders[0] ), $query->orders );
 	}
+
+	/**
+	 * @testDox The get_args method returns the initial args sent.
```
naman03malhotra:
```suggestion
	 * @testDox The get_query_args method returns the initial args sent.
```

naman03malhotra (approved):
Approving! left a comment for minor changes with naming.

plugins/woocommerce/tests/php/src/Internal/DataStores/Orders/OrdersTableQueryTests.php
felipeelia: Ah, good catch, thanks! Changed in 50cd2a9

plugins/woocommerce/tests/php/src/Internal/DataStores/Orders/OrdersTableQueryTests.php
felipeelia: Also changed in 50cd2a9 :)

jeffpaul:
@naman03malhotra if this is approved, are you good to get this merged in (and tagged to a specific Woo release)?

Issue #54438, opened 2025-01-13
Code Snippets Overhaul


shaybanshee:
### Code Snippets Testing and Updates

[Code Snippets](https://developer.woocommerce.com/docs/category/code-snippets/) can be viewed on our docs page, under the Getting Started section. 

When we migrated these code snippets over from WooCommerce.com, we did not have an opportunity to test or organize them into a more cohesive segmentation. We also have no way of tracking the original writer of the code snippet, hence the crowd sourced approach I am taking to get these tested and updated.

## Overview of Code Snippets

- Code snippets should not be a part of the Getting Started hierarchy.
- Code snippets are a small, reusable piece of PHP code that can be added to your WooCommerce store to customize its functionality - they should not be cumbersome, and are meant for specific customization use cases for mid-advanced users. 
- Code snippets and customization are not supported by our HE team, so the responsibility to maintaining and updating the best practices and relevance of these snippets falls to internal Woo developers, and the community. 
- The markdown files for the code snippets are accessible to edit in the [/docs](https://github.com/woocommerce/woocommerce/tree/trunk/docs/code-snippets) folder.
- In order to update, the docs [contribution process](https://developer.woocommerce.com/docs/contributing-technical-documentation/) needs to be followed which is subject to change soon. 

## Goals 

1. Each snippet should be tested to see if it is still working with WooCommerce 9.5.2(or latest version)
2. Snippets should be relevant and written with best practices in mind
3. Snippets need to be revisited to be maintained
4. Snippets will be reorganized for the mid to advanced developer segment

shaybanshee:
@nielslange Here is the overview; what do you think the best workflow is? My thought is, if you can confirm the functionality of these snippets, I can update any language, typos, etc. thereafter and work on the file organization?

jessepearson:
@shaybanshee Is there anything else we need to do after tagging you as a reviewer for the PRs with the doc updates?

PR #55159, opened 2025-02-05
Update change-a-currency-symbol.md

### LINKED ISSUE #55158: Code Snippets Overhaul: Changing a Currency Symbol
| Review | Source |
| --- | --- |
| https://developer.woocommerce.com/docs/change-a-currency-symbol/ | https://github.com/woocommerce/woocommerce/blob/trunk/docs/code-snippets/change-a-currency-symbol.md |

barryhughes:
As part of our [code snippets review](https://github.com/woocommerce/woocommerce/issues/54438), I'm updating the *Change a Currency Symbol* post to add some additional context as to why this might be needed.

The snippet itself remains unchanged (and I have verified it continues to work). Even so, if you wish, feel free to test and review it as part of this code review.

Fixes #55158.

### DIFF
--- a/docs/code-snippets/change-a-currency-symbol.md
+++ b/docs/code-snippets/change-a-currency-symbol.md
@@ -3,9 +3,9 @@ post_title: Change a currency symbol
 tags: code-snippet
 ---
 
-See the [currency list](https://woocommerce.github.io/code-reference/files/woocommerce-includes-wc-core-functions.html#source-view.475) for reference on currency codes.
+In WooCommerce, each currency is associated with a code and a symbol. For example, the Australian dollar has the code `AUD` and the symbol `$` (if you are interested, you can see a full list of codes and symbols in the [source code](https://woocommerce.github.io/code-reference/files/woocommerce-includes-wc-core-functions.html#source-view.662)). 
 
-Add this code to your child theme's `functions.php` file or via a plugin that allows custom functions to be added, such as the [Code Snippets](https://wordpress.org/plugins/code-snippets/) plugin. Avoid adding custom code directly to your parent theme's functions.php file, as this will be wiped entirely when you update the theme.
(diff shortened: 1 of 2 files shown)


docs/code-snippets/change-a-currency-symbol.md
```diff
@@ -3,9 +3,9 @@ post_title: Change a currency symbol
 tags: code-snippet
 ---
 
-See the [currency list](https://woocommerce.github.io/code-reference/files/woocommerce-includes-wc-core-functions.html#source-view.475) for reference on currency codes.
+In WooCommerce, each currency is associated with a code and a symbol. For example, the Australian dollar has the code `AUD` and the symbol `$` (if you are interested, you can see a full list of codes and symbols in the [source code](https://woocommerce.github.io/code-reference/files/woocommerce-includes-wc-core-functions.html#source-view.662)).
```
barryhughes: I updated the link, because it seemed to be pointing to the wrong section of the source ... but, the line number will probably continue to fall out of sync with this reference. Might it be better to link to a specific commit or tag, such as:

https://github.com/woocommerce/woocommerce/blob/9.6.1/plugins/woocommerce/includes/wc-core-functions.php#L682

Or, to re-word my question, is it a requirement/expectation to use https://woocommerce.github.io/code-reference/ when linking to the source?

shaybanshee: It is definitely not a requirement - and something that we will eventually sunset, so the sooner we update the better!

shaybanshee (approved):

docs/code-snippets/change-a-currency-symbol.md
barryhughes:
```suggestion
In WooCommerce, each currency is associated with a code and a symbol. For example, the Australian dollar has the code `AUD` and the symbol `$` (if you are interested, you can see a full list of codes and symbols in the [source code](https://github.com/woocommerce/woocommerce/blob/9.6.1/plugins/woocommerce/includes/wc-core-functions.php#L682)). 
```

PR #55162, opened 2025-02-06
[API Tests] Unskip api test `customers-crud.test.js` on WPCOM and Pressable

rodelgc:

Part of #54662.

This PR makes the api test `customers-crud.test.js` runnable on WPCOM and Pressable, and also on multisites, by applying these changes:
- Replace hard-coded user IDs with actual ones
- Remove special characters in usernames when creating a new user. Multisites don't like special characters in usernames.
- On multisites, skip tests on customer deletion due to a currently existing bug on multisite environments.

1. Make sure all api tests pass in CI.
2. Test on WPCOM and Pressable, making sure that the api test `customers-crud.test.js` pass, with no tests skipped:
    ```sh
    export E2E_ENV_KEY="..."
    
    # Run on WPCOM
    pnpm test:e2e:wpcom api-wpcom customers-crud.test.js
    
    # Run on Pressable
    pnpm test:e2e:pressable api-pressable customers-crud.test.js
    ```
4. Set up a multisite either on JN or your local machine, and run `customers-crud.test.js`against it. Make sure the individual tests pass, but these two tests should be skipped:
    - `can permanently delete an customer`
    - `can batch delete customers`

   ```sh
   export BASE_URL="https://your.multisite-url.com"
   # Set other important environment variables like the admin and customer credentials in your multisite
   pnpm test:api customers-crud.test.js
   ```

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/dev-e2e-api-unskip-customers-crud
+++ b/plugins/woocommerce/changelog/dev-e2e-api-unskip-customers-crud
@@ -1,0 +1,4 @@
+Significance: patch
+Type: dev
+
+Comment: Unskip the API test `customers-crud.test.js` on WPCOM and Pressable.
--- a/plugins/woocommerce/tests/e2e-pw/fixtures/site.setup.js
+++ b/plugins/woocommerce/tests/e2e-pw/fixtures/site.setup.js
@@ -104,3 +104,9 @@ setup( 'convert Cart and Checkout pages to shortcode', async ( { wpApi } ) => {
 setup( 'disable coming soon', async ( { baseURL } ) => {
 	await setComingSoon( { baseURL, enabled: 'no' } );
 } );
+
+setup( 'determine if multisite', async ( { api } ) => {
+	const response = await api.get( 'system_status' );
+	const { environment } = response.data;
+	process.env.IS_MULTISITE = environment.wp_multisite;
+} );
--- a/plugins/woocommerce/tests/e2e-pw/tests/api-tests/customers/customers-crud.test.js
+++ b/plugins/woocommerce/tests/e2e-pw/tests/api-tests/customers/customers-crud.test.js
@@ -3,22 +3,22 @@ const { admin } = require( '../../../test-data/data' );
 
 test.describe( 'Customers API tests: CRUD', () => {
 	let adminId;
-		let customerId;
-		let subscriberUserId;
-		let subscriberUserCreatedDuringTests = false;
+	let customerId;
+	let subscriberUserId;
+	let subscriberUserCreatedDuringTests = false;
 
-		test.beforeAll( async ( { request } ) => {
-			// Call the API to return all users and determine if a
-			// subscriber user has been created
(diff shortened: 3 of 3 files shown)


adimoldovan:
Thanks for these updates @rodelgc. 

Please also check other tests skipped for multisite, based on a IS_MULTISITE env variable, introduced by https://github.com/woocommerce/woocommerce/pull/55086. I like your approach of checking for multisite via the API, but we should be consistent. Maybe combine them and set the  IS_MULTISITE in a setup fixture based on environment.wp_multisite? This will make is available in all specs.

> Remove special characters in usernames when creating a new user. Multisites don't like special characters in usernames

I'm not sure I understand this one.

adimoldovan (changes requested):
I realised I only left a comment instead of using the `review changes` and this keeps popping up in our alerts. 
See [the comment](https://github.com/woocommerce/woocommerce/pull/55162#issuecomment-2639093862).

rodelgc:
Thanks @adimoldovan. I pushed a few changes that sets the `process.env.IS_MULTISITE` variable from within the `site.setup.js` file. The `customers-crud.test.js` test then uses this variable instead of the old `isMultisite` one.

> I'm not sure I understand this one.

In the UI, multisites complain if the username of the customer contains a special character:

Screenshot 2025-02-11 at 9 41 16 PM

However, the "Create customer" API endpoint allows special characters. For example, I was able to create these customers using the REST API on the very same multisite:
Screenshot 2025-02-11 at 9 59 43 PM

Because of this inconsistency between the UI and the API, I went with the safer username (the one that doesn't have special characters).

adimoldovan:
> In the UI, multisites complain if the username of the customer contains a special character:

@rodelgc This must be a bug:
- I checked and I could successfully create `customer_123` in a multisite, but using a super-admin. If I tried that with an site admin it failed indeed. I don't think the username format is at fault here and it could be a permissions issue instead.
- As long as the format is a valid WordPress username one, I see no reason to throw this error.
- It works via the API, so the inconsistency alone is a bug.

Leaving the bug aside, I see no harm in updating the tests to pass for now. I would however investigate this further, report the bug and update the tests back when the bug is fixed.

adimoldovan (approved):
👍🏼 
Updating the username format is not required, because it doesn't affect the API. Still, I see no harm either, so we should 
merge this.

@rodelgc, thanks for this update. Please follow-up on the user creation issue, as mentioned [here](https://github.com/woocommerce/woocommerce/pull/55162#issuecomment-2650960283)

rodelgc:
Thanks for checking @adimoldovan ! Yeah I agree it would fall as a permissions bug. I'll double check the private repo to confirm if my memory is correct that some user permissions are currently being fixed there, and will raise a new bug issue if there's none or if there's no issue there that's related to this.

Issue #53214, opened 2024-11-26
[api-core-tests] Split big assertions into smaller and more manageable ones


adimoldovan:
We have some api tests that are checking some enormous object and one small difference that can cause a failure is very hard to find. We should split these into more assertions that check smaller objects to make tests more deterministic.

Example: https://github.com/woocommerce/woocommerce/blob/dc2ef16d719f45be5e14762f8a41ac3aa436e7dd/plugins/woocommerce/tests/e2e-pw/tests/api-tests/system-status/system-status-crud.test.js#L99

PR #55168, opened 2025-02-06
[API Tests] Unskip `system-status-crud.test.js` on WPCOM

rodelgc:

Part of #54662.

This PR updates the test `can view all system status items` to make it compatible with WPCOM, Pressable, and multisites. Other tests were untouched. 

The failures were due to schema differences of the endpoint responses between wp-env and WPCOM. The failures were scattered across different parts and assertions in this test. As mentioned in #53214, it had been difficult to troubleshoot these scattered failures so as much as I wanted to make the changes minimal, I ended up needing to restructure this test. 

These changes are:
- Organized the contents of this test into `test.step()`s and also added explicit messages in some of the `expect()` calls to make them more verbose and easier to identify failures.
    - In certain parts, I nested the test steps so I had to disable the `playwright/no-nested-step` rule to prevent eslint warnings. I limited the nesting to two levels only to hopefully maintain readability.
- Put all [`schemas`](https://github.com/woocommerce/woocommerce/blob/970818f4052b9f2493e03d8115ee9a85fac6c7f5/plugins/woocommerce/tests/e2e-pw/tests/api-tests/system-status/system-status-crud.test.js#L72) in a single place to decouple them from the assertions, which also helps make individual `test.step()`s to be shorter.
- Created a custom matcher [`expect.anyOf()`](https://github.com/woocommerce/woocommerce/blob/970818f4052b9f2493e03d8115ee9a85fac6c7f5/plugins/woocommerce/tests/e2e-pw/tests/api-tests/system-status/system-status-crud.test.js#L12) to relax some assertions on object properties that have different data types between wp-env and WPCOM.
- Keys in the `database.database_tables.woocommerce` object respect the database prefix. This is why the assertions for this object are failing on multisites -- multisites have a db prefix like `wp_2_` instead of the default `wp_`. To address this, I created the [`getExpectedWooCommerceTables()`](https://github.com/woocommerce/woocommerce/blob/970818f4052b9f2493e03d8115ee9a85fac6c7f5/plugins/woocommerce/tests/e2e-pw/tests/api-tests/system-status/system-status-crud.test.js#L235) function to return the correctly db-prefixed key names.

There were a few oddities that needed special handling:
- The properties `environment.remote_get_response` and `environment.remote_post_response`:
    - In WPCOM they are numbers. But in wpenv they are strings -- but here's the weird thing: they're boolean (with value `false`) if the wp-env is newly launched. They only become strings on succeeding runs of this test :shrug:. To address this, I set the assertion of those 2 properties to be of type boolean, string, or number.
- The property `environment.external_object_cache`:
    - It is `null` on wp-env, but boolean in other environments not just in WPCOM, since hosts can use an external object cache. The weird thing here is that in JavaScript, the `typeof null` evaluates to `object` instead of `null` as mentioned [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#typeof_null). To handle this, I placed it on its own separate assertion.
- The `database.database_tables.other` object:
    - Some keys in this object follow the database prefix like `wp_2_wc_admin_notes`. Unfortunately some don't, like `wp_usermeta` :man_shrugging:  To handle this I had to rely on their [suffixes](https://github.com/woocommerce/woocommerce/blob/970818f4052b9f2493e03d8115ee9a85fac6c7f5/plugins/woocommerce/tests/e2e-pw/tests/api-tests/system-status/system-status-crud.test.js#L200) instead of their full key name.

1. Make sure  `system-status-crud.test.js` pass together with all API tests in CI.
2. Test it on WPCOM, Pressable, and Multisite. Inspect their HTML reports and see if there's anything unusual, or if the format is confusing:
    ```sh
    # Test on WPCOM and Pressable
    export E2E_ENV_KEY="..."
    pnpm test:e2e:wpcom api-wpcom system-status-crud.test.js
    pnpm test:e2e:pressable api-pressable system-status-crud.test.js
    
    # Test on multisite
    export BASE_URL=https://your.multisite.url
    # Set other variables here like admin and customer credentials of your multisite
    pnpm test:api system-status-crud.test.js
    ```

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/dev-e2e-api-unskip-wpcom-system-status-crud-2
+++ b/plugins/woocommerce/changelog/dev-e2e-api-unskip-wpcom-system-status-crud-2
@@ -1,0 +1,4 @@
+Significance: patch
+Type: dev
+
+Unskip api test `system-status-crud.test.js` on WPCOM, Pressable.
--- a/plugins/woocommerce/tests/e2e-pw/tests/api-tests/system-status/system-status-crud.test.js
+++ b/plugins/woocommerce/tests/e2e-pw/tests/api-tests/system-status/system-status-crud.test.js
@@ -252,7 +252,7 @@ const getExpectedOtherTables = ( dbPrefix ) => {
 
 /* eslint-disable playwright/no-nested-step */
 test.describe( 'System Status API tests', () => {
-	test.only( 'can view all system status items', async ( { request } ) => {
+	test( 'can view all system status items', async ( { request } ) => {
 		let responseJSON,
 			databasePrefix,
 			databaseSize,
--- a/plugins/woocommerce/tests/e2e-pw/tests/api-tests/system-status/system-status-crud.test.js
+++ b/plugins/woocommerce/tests/e2e-pw/tests/api-tests/system-status/system-status-crud.test.js
@@ -31,15 +31,7 @@ const schemas = {
 		{ field: 'gzip_enabled', type: any( Boolean ) },
 		{ field: 'mbstring_enabled', type: any( Boolean ) },
 		{ field: 'remote_post_successful', type: any( Boolean ) },
-		{
-			field: 'remote_post_response',
-			type: process.env.IS_WPCOM ? any( Number ) : any( String ),
-		},
 		{ field: 'remote_get_successful', type: any( Boolean ) },
-		{
-			field: 'remote_get_response',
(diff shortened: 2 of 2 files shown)


adimoldovan (approved):
👍🏼

PR #55180, opened 2025-02-06
[e2e tests] Update gitignore to exclude env paths containing _local

adimoldovan:

Update .gitignore file to ignore anything under a directory in envs starting with `_local`. This is useful to define more local environments that you don't want to be pushed to remote.

Create a directory `plugins/woocommerce/tests/e2e-pw/envs/_local-abc` and add any file under it. Check git status and make sure the file is not marked as new.


Stojdza (approved):
LGTM

Issue #54805, opened 2025-01-23
Cash on Delivery payment method not present for guest users upon visiting checkout for the first time


veljkho:
### Prerequisites

- [x] I have carried out troubleshooting steps and I believe I have found a bug.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

As per the mass test execution against WooCommerce 9.7 Beta 1 build, we believe that we have found an issue in the build since the issue is not happening on < 9.7.0-beta.1

After enabling the COD payment method and as anonymous/guest visiting the checkout page for the first time there's error message "There are no payment methods available." instead of COD enabled and eligibility to purchase.

If I log in to admin, I can see "Cash on Delivery" is enabled in Woo settings, and there are no special rules to use it, so it should have been available. This works fine on earlier versions of Woo < 9.7.0-beta.1

More discussion here: p1737542593023539-slack-C03R7A7D1DW

### Expected behavior

COD method enabled and no error message

### Actual behavior

Error message present with no payment methods present (COD was enabled in the settings)

### Steps to reproduce

1. We enable the cash-on-delivery payment method
2. We create a test product
3. As anonymous (private window, guest flow), we go to "/shop"
4. We visit the product page
5. We add it to cart
6. We go to "/checkout"
7. It shows, "There are no payment methods available."
8. We try to place the order anyway, it refreshes the page and shows "No payment method provided"
9. But now the "Cash on Delivery" payment method shows up and is selectable

### WordPress Environment

N/A

### Isolating the problem

- [x] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [x] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/products/storefront/).
- [ ] I can reproduce this bug consistently using the steps above.

nielslange:
Hello @veljkho 👋

Thanks for reporting this issue. On my end, I'm unable to reproduce it using the WooCommerce 9.7 Beta 1 release. When visiting the checkout page, using the Checkout block, as a guest, using an incognito window, I do see COD as a payment option. Did this issue happen on your end when using a fresh installation with the WooCommerce 9.7 Beta 1 release, or on an existing site after upgrading from an earlier version of WooCommerce?

veljkho:
Hi @nielslange 

[...]

PR #53611, opened 2024-12-10
Cart Data Store: Prevent unnecessary requests to cart API when cart is empty (for the mini cart block)

mikejolley:
### Submission Review Guidelines:

-   I have followed the [WooCommerce Contributing Guidelines](https://github.com/woocommerce/woocommerce/blob/trunk/.github/CONTRIBUTING.md) and the [WordPress Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/).
-   I have checked to ensure there aren't other open [Pull Requests](https://github.com/woocommerce/woocommerce/pulls) for the same update/change.
-   I have reviewed my code for [security best practices](https://developer.wordpress.org/apis/security/).
-   Following the above guidelines will result in quick merges and clear and detailed feedback when appropriate.

### Changes proposed in this Pull Request:

This is an updated approach to prevent API requests on load when using the mini-cart block. This supersedes #51981 

1. Removes the mini cart specific caching (local storage) and update logic
2. Makes the mini cart use the cart data store
3. Adds persistent caching to cart data store
4. Resolves early without an API request if the cart is empty (no cookie) or the local storage contains the current cart which matches the current cookie. Otherwise fires an API request as normal.

This should mitigate page caching issues with stale data without the extra request to the server when its not needed.

Closes #50793
Closes #51981

### How to test the changes in this Pull Request:

1. Add the Mini Cart Block to a page.
2. Ensure it renders the correct quantity in cart on page load. Add some items to the cart to test this.
3. With a new session and empty cart, view the page with the browser network inspector open. Ensure no XHR requests are made.
4. Add `?add-to-cart=16` (use a valid product ID) to the URL. Ensure mini cart is updated.
5. Open the shop page on another tab and add something to the cart. Refresh mini cart. It updates.

### Changelog entry

-   [ ] Automatically create a changelog entry from the details below.

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Details</summary>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

-   [ ] Fix - Fixes an existing bug
-   [ ] Add - Adds functionality
-   [ ] Update - Update existing functionality
-   [ ] Dev - Development related task
-   [ ] Tweak - A minor adjustment to the codebase
-   [ ] Performance - Address performance issues
-   [ ] Enhancement - Improvement to existing functionality

#### Message 

</details>

<details>

<summary>Changelog Entry Comment</summary>

[...]

Issue #53903, opened 2024-12-19
In checkout block, "no shipping options available" is briefly shown after adding a correct shipping address


tomxygen:
### Prerequisites

- [ ] I have carried out troubleshooting steps and I believe I have found a bug.
- [ ] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

When shoppers reach the checkout block and add a correct shipping address, the warning message "no shipping options available" is shown for a moment.

See the attached video.

https://github.com/user-attachments/assets/9673bd74-8b7a-4a70-ac36-1016462eb809

### Expected behavior

The message shouldn't be displayed.
The loading spinner should be the only visibile thing.

### Actual behavior

-

### Steps to reproduce

1. enable shipping options for a specific location
2. add a product to the cart and go to checkout
3. add a valid shipping address for the location.

### WordPress Environment

<details><summary>System Status Report</summary>

```
### WordPress Environment ###

WordPress address (URL): [Redacted]
Site address (URL): [Redacted]
WC Version: 9.6.0-dev
Legacy REST API Package Version: Il plugin API REST legacy non è installato in questo sito.
Action Scheduler Version: ✔ 3.9.0
Log Directory Writable: ✔
WP Version: 6.7.1
WP Multisite: –
WP Memory Limit: 1 GB
WP Debug Mode: –
WP Cron: ✔
Language: it_IT
External object cache: –

### Server Environment ###

Server Info: Apache
PHP Version: 8.3.14
PHP Post Max Size: 100 MB
PHP Time Limit: 30
PHP Max Input Vars: 1000000
cURL Version: 7.76.1
OpenSSL/3.2.2

SUHOSIN Installed: –
MySQL Version: 10.11.10-MariaDB
Max Upload Size: 100 MB
Default Timezone is UTC: ✔
fsockopen/cURL: ✔
SoapClient: ✔
DOMDocument: ✔
GZip: ✔
Multibyte String: ✔
Remote Post: ✔
Remote Get: ✔

### Database ###

[REDACTED]

### Post Type Counts ###

attachment: 39
mailpoet_page: 4
page: 15
post: 3
product: 26
product_variation: 10
revision: 48
shop_order_placehold: 16
wp_font_face: 36
wp_font_family: 12
wp_global_styles: 1
wp_navigation: 2
wp_template: 4
wp_template_part: 4

### Security ###

Secure connection (HTTPS): ✔
Hide errors from visitors: ✔

[...]

PR #55191, opened 2025-02-06
[Checkout] Prevent a delay in rendering checkout payment methods on first load.

mikejolley:

Fixes a delay with cart data hydration that causes payment methods to be missing during first render. With this fix in place, data is immediately available.

This issue was introduced with https://github.com/woocommerce/woocommerce/pull/53611 which included some updated logic in `getCartData` to update the cart hash from the cart data response. A promise was used to catch errors.

This promise is the reason for delay; we believe that due to the promise, the payment methods are rendered empty, and from that point, updating payment methods is debounced. Making this synchronous will ensure cart data exists before the first render, which avoids the debounce, and prevents the error state being displayed to the customer.

To see the original issue without this PR, add something to cart, go to the checkout page, zoom out so you can see the payment methods without scrolling, then refresh. You'll see a red notice for a short while before payment methods are shown. 

| Before | After |
|--------|-------|
| Screen Recording 2025-02-06 at 13 05 59  | Screen Recording 2025-02-06 at 13 04 30 |

This seems to be related to the below issue, although I couldn't replicate them being missing indefinitely, only after a short delay. Someone who can reproduce the issue should test after this PR is in place before closing.

Ref: #54805

1. Add item to cart
2. Go to checkout page
3. Zoom out so you can see payment methods
4. Refresh page
5. Payment methods will be visible without first seeing a "no payment methods exist" error.
6. Checkout as normal

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce-blocks/assets/js/data/cart/resolvers.ts
+++ b/plugins/woocommerce-blocks/assets/js/data/cart/resolvers.ts
@@ -31,22 +31,20 @@ export const getCartData =
 			apiFetch.setCartHash( response?.headers );
 		}
 
-		response
-			.json()
-			.then( function ( cartData: CartResponse ) {
-				const { receiveCart, receiveError } = dispatch;
+		try {
+			const cartData: CartResponse = await response.json();
+			const { receiveCart, receiveError } = dispatch;
 
-				if ( ! cartData ) {
-					receiveError( CART_API_ERROR );
-					return;
-				}
-
-				receiveCart( cartData );
-			} )
-			.catch( () => {
-				const { receiveError } = dispatch;
+			if ( ! cartData ) {
 				receiveError( CART_API_ERROR );
-			} );
+				return;
+			}
+
+			receiveCart( cartData );
+		} catch ( error ) {
+			const { receiveError } = dispatch;
+			receiveError( CART_API_ERROR );
+		}
 	};
 
 /**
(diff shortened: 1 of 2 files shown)


opr (approved):
Working for me and the reasoning makes sense:

Current flow:
- Payment methods load empty
- Resolver returns while Promise is still pending
- Checkout renders due to resolver finishing. 
- Renders with error due to missing payment methods
- Promise resolves and payment methods are updated after a debounce.
- Error disappears.

Flow after this change:

- Payment methods load empty
- Resolver waits for Promise to resolve
- Checkout renders when resolver has finished
- Renders with no error because payment methods are present

pierorocca:
https://github.com/woocommerce/woocommerce/issues/53903 is I believe related. Apologies I should have changed this to a bug rather than keeping it as an enhancement.

opr:
> #53903 is I believe related. Apologies I should have changed this to a bug rather than keeping it as an enhancement.

Thanks Piero, this bug is related in terms of the experience, but it's not caused by the same thing and will need a different PR to fix. We should look into #53903 in a separate effort.

pierorocca:
@opr my bad. This is related to payments options while the other is shipping.

PR #55031, opened 2025-01-31
[e2e tests] Restructure project - move orders specs

adimoldovan:
### Changes proposed in this Pull Request:

Part of pdV5qK-Ie-p2
More context: pcr5Cj-70A-p2

Move all orders related tests into the new `order` folder.

### How to test the changes in this Pull Request:

Run the tests in the `tests/order` folder. They should all pass.

PR #55196, opened 2025-02-06
[e2e tests] Restructure projects - move cart and checkout specs

adimoldovan:

Part of pdV5qK-Ie-p2
More context: pcr5Cj-70A-p2

Move all cart and checkout related tests into the new `cart` and `checkout` folders.
Also moved on order spec that was missed in [55031](https://github.com/woocommerce/woocommerce/pull/55031)

Simply moving the tests and changing the order they run in led to some failures. I had to update to following to fix that:

- added an `afterAll` hook to reset the coming soon mode back to 'no' in core profile tests. These were enabling the coming soon mode breaking any subsequent shopper tests (there were non before this PR)
- reset `woocommerce_default_country` to its initial value after test in `payment.spec.js`. This was messing up the cart and checkout tests that were relying on the default "US:CA" value.

Run the tests in the `tests/cart` and `tests/checkout` folders. They should all pass.

```
pnpm test:e2e tests/cart tests/checkout
```


Stojdza (approved):
Looks good and tests are passing locally for me 👍

PR #55197, opened 2025-02-06
Fix: Ensure 'view' action is excluded by key instead of translated name (#54879)

### LINKED ISSUE #54879: View action in Order Details page for non-EN stores after WC 9.6.0
I've updated to *WC 9.6.0* and now have the _'View'_ action in the *Order Details* page which has no sense (it's a link to the current page).

This is because in #52756 is used `array_filter` to exclude the action with the value 'View' (note: without using `gettext`) for the key 'name'. 

```
$actions = array_filter(
	wc_get_account_orders_actions( $order ),
	function ( $action ) {
		return 'View' !== $action['name'];
	}
);
```

It therefore _only applies to English_, not to other languages.

Even using the gettext value `__( 'View', 'woocommerce' )` is not the best solution, since 'gettext' may have been filtered and that string overwritten.

Why not simply *`unset` the key 'view' from the actions array*?

```
$actions = wc_get_account_orders_actions( $order );
unset( $actions['view'] );
```

This should be enough to *exclude View action / button from Order Details page*.

Thanks.

_Originally posted by @dademaru in https://github.com/woocommerce/woocommerce/issues/52756#issuecomment-2609256530_

Daan001:

- Fixes the issue where the 'View' action (translated as "Bekijken" in Dutch) was incorrectly excluded due to filtering based on the action name rather than the array key.  
- Excludes the 'view' action from the $actions array by filtering based on the key, ensuring it is correctly excluded, regardless of language.  

Closes #54879.

1. Switch your WordPress installation to Dutch (or any language where "View" is translated).  
2. Complete an order as a guest.  
3. After the order is completed, ensure the "View" action is correctly excluded from the available actions on the Order Details page.

#### Screenshots

<table>
<tr>
<td valign="top">Before:
<br><br>
Screenshot 2025-02-11 at 11 16 15
</td>
<td valign="top">After:
<br><br>
Screenshot 2025-02-11 at 11 18 30
</td>
</tr>
</table>

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 
Fixes exclusion of 'View' action by key instead of translated name in the `order-details.php` template.

</details>

### DIFF
--- a/plugins/woocommerce/changelog/55197-trunk
+++ b/plugins/woocommerce/changelog/55197-trunk
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fixes exclusion of 'View' action by key instead of translated name in the `order-details.php` template.
--- a/plugins/woocommerce/templates/order/order-details.php
+++ b/plugins/woocommerce/templates/order/order-details.php
@@ -12,7 +12,7 @@
  *
  * @see     https://woocommerce.com/document/template-structure/
  * @package WooCommerce\Templates
- * @version 9.6.0
+ * @version 9.8.0
  *
  * @var bool $show_downloads Controls whether the downloads table should be rendered.
  */
--- a/plugins/woocommerce/templates/order/order-details.php
+++ b/plugins/woocommerce/templates/order/order-details.php
@@ -32,9 +32,10 @@ $show_purchase_note = $order->has_status( apply_filters( 'woocommerce_purchase_n
 $downloads          = $order->get_downloadable_items();
 $actions            = array_filter(
 	wc_get_account_orders_actions( $order ),
-	function ( $action ) {
-		return 'View' !== $action['name'];
-	}
+	function ( $key ) {
+		return 'view' !== $key;
+	},
+	ARRAY_FILTER_USE_KEY
 );
 
 // We make sure the order belongs to the user. This will also be true if the user is a guest, and the order belongs to a guest (userID === 0).


nielslange (changes requested):
Thanks for creating this PR, @Daan001. We appreciate your contribution. A couple of things need attention, though. 

First, it looks like the changelog entry is missing from the PR description. This entry is crucial as it documents the changes introduced. You can manually add a changelog entry by running:

```sh
pnpm --filter='@woocommerce/plugin-woocommerce' changelog add
```

Additionally, the CI checks indicate that the version in the changed template file needs to be bumped to `9.8.0`. This version bump is essential for themes that override this template to identify that there has been a change.

Lastly, I see that you've pushed changes directly to `trunk` within your fork. For better management, it would be beneficial to use a separate branch, like `fix/excluded-by-key`, in future contributions. It's okay for now, but using a feature branch next time would be helpful.

Daan001:
Hi @nielslange,

Thanks for the feedback! As this is my first PR, I really appreciate the guidance.

I’ve made the following updates:

- Added the changelog entry.
- Bumped the version to 9.8.0.
- I wasn’t able to use a new branch for this PR, but I’ll ensure to do so for future contributions.

Looking forward to your feedback!

nielslange (approved):
Thanks for applying the changes, and congratulations on your first PR, @Daan001! 👏 I've successfully tested the updates and added before-and-after screenshots. These will be helpful when testing the release that includes this PR, as they provide a quick visual reference for what changed.

Thanks again for your contribution! 🙌

PR #55208, opened 2025-02-06
Prevent possible fatal when calling `woocommerce_output_all_notices()`

### LINKED ISSUE #55101: [Fatal] PHP Fatal error:  Uncaught Error: Call to undefined function wc_print_notices() in | wc-template-functions.php:4097
### Prerequisites

### Describe the bug

witnessed in `9.6.0`

[link to code](https://github.com/woocommerce/woocommerce/blob/6f5032a0277a0eb006a442407d9d09c0c370b8be/plugins/woocommerce/includes/wc-template-functions.php#L4097)

```
PHP Fatal error:  Uncaught Error: Call to undefined function wc_print_notices() in /../plugins/woocommerce/9.6.0/includes/wc-template-functions.php:4097
Stack trace:
#0 /../wp-content/plugins/elementor-pro/modules/woocommerce/widgets/notices.php(121): woocommerce_output_all_notices()
#1 /../wp-content/plugins/elementor/includes/base/controls-stack.php(2366): ElementorPro\Modules\Woocommerce\Widgets\Notices->render()
#2 /../wp-content/plugins/elementor/includes/base/widget-base.php(636): Elementor\Controls_Stack->render_by_mode()
#3 /../wp-content/plugins/elementor/includes/base/widget-base.php(775): Elementor\Widget_Base->render_content()
#4 /../wp-content/plugins/elementor/includes/base/element-base.php(483): Elementor\Widget_Base->print_content()
#5 /../wp-content/plugins/elementor/modules/element-cache/module.php(73): Elementor\Element_Base->print_element()
#6 /../core/6.7.1/wp-includes/shortcodes.php(434): Elementor\Modules\ElementCache\Module->Elementor\Modules\ElementCache\{closure}(Array, '', 'elementor-eleme...')
#7 [internal function]: do_shortcode_tag(Array)
#8 /../core/6.7.1/wp-includes/shortcodes.php(273): preg_replace_callback('/\\[(\\[?)(elemen...', 'do_shortcode_ta...', '<div class="ele...')
#9 /../wp-content/plugins/elementor/core/base/document.php(1870): do_shortcode('<div class="ele...')
#10 /../wp-content/plugins/elementor/core/base/document.php(1204): Elementor\Core\Base\Document->print_elements(Array)
#11 /../wp-content/plugins/elementor/includes/frontend.php(1197): Elementor\Core\Base\Document->print_elements_with_wrapper(Array)
#12 /../wp-content/plugins/elementor/includes/frontend.php(1092): Elementor\Frontend->get_builder_content(4403)
#13 /../core/6.7.1/wp-includes/class-wp-hook.php(324): Elementor\Frontend->apply_builder_in_content('<style>\n\t\t\t.woo...')
#14 /../core/6.7.1/wp-includes/plugin.php(205): WP_Hook->apply_filters('<style>\n\t\t\t.woo...', Array)
#15 /../plugins/wordpress-seo/24.3/src/builders/indexable-link-builder.php(140): apply_filters('the_content', '<style>\n\t\t\t.woo...')
#16 /../plugins/wordpress-seo/24.3/src/actions/indexing/abstract-link-indexing-action.php(76): Yoast\WP\SEO\Builders\Indexable_Link_Builder->build(Object(Yoast\WP\SEO\Models\Indexable), '<style>\n\t\t\t.woo...')
#17 /../plugins/wordpress-seo/24.3/src/integrations/admin/background-indexing-integration.php(163): Yoast\WP\SEO\Actions\Indexing\Abstract_Link_Indexing_Action->index()
#18 /../core/6.7.1/wp-includes/class-wp-hook.php(324): Yoast\WP\SEO\Integrations\Admin\Background_Indexing_Integration->index()
#19 /../core/6.7.1/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters('', Array)
#20 /../core/6.7.1/wp-includes/plugin.php(565): WP_Hook->do_action(Array)
#21 /../core/6.7.1/wp-cron.php(191): do_action_ref_array('wpseo_indexable...', Array)
#22 {main}
  thrown in /../plugins/woocommerce/9.6.0/includes/wc-template-functions.php on line 4097
``` 

### Expected behavior

no fatals expected

### Actual behavior

fatal errors witnessed

### Steps to reproduce

We don't have steps to reproduce, but observed 199 instances of this problem on Fri Jan 31, 2025 across 3 sites. Hopefully, someone can dig in and apply a defensive check if appropriate. - To be updated after investigation.

### WordPress Environment

N/A

### Isolating the problem

jorgeatorres:
in the wrong context

Under normal circumstances, functions in `wc-template-functions.php` are made available [everywhere](https://github.com/woocommerce/woocommerce/blob/ffff59f12979d9debea531db0cd99efbce4de94e/plugins/woocommerce/includes/class-woocommerce.php#L289), but `woocommerce_output_all_notices()`, a function defined in this file, depends on `wc_print_notices()` which comes from a file that is only included on [frontend requests](https://github.com/woocommerce/woocommerce/blob/ffff59f12979d9debea531db0cd99efbce4de94e/plugins/woocommerce/includes/class-woocommerce.php#L810).

Though unlikely, this means that `woocommerce_output_all_notices()` could end up being called in a context where `wc_print_notices()` is not defined.

This PR adds a defensive check to prevent this from bubbling up into a fatal error.

Closes #55101.

1. Drop the following PHP as a .php file inside `wp-content/mu-plugins` or run it as a code snippet in a different way:
   ```php
   <?php
   add_action(
	   'admin_footer',
	   function() {
		   woocommerce_output_all_notices();
	   }
   );
   ```
1. Visit the admin of your site.
1. No error should be logged.
1. (Optional) Check out `trunk` and follow the instructions above. This time, a fatal will be triggered (check your logs):
   > `PHP Fatal error:  Uncaught Error: Call to undefined function wc_print_notices() in /[...]/woocommerce/plugins/woocommerce/includes/wc-template-functions.php:4150`

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/includes/wc-template-functions.php
+++ b/plugins/woocommerce/includes/wc-template-functions.php
@@ -4137,6 +4137,10 @@ function wc_get_cart_undo_url( $cart_item_key ) {
  * @since 3.5.0
  */
 function woocommerce_output_all_notices() {
+	if ( ! function_exists( 'wc_print_notices' ) ) {
+		return;
+	}
+
 	echo '<div class="woocommerce-notices-wrapper">';
 	wc_print_notices();
 	echo '</div>';
--- a/plugins/woocommerce/includes/wc-template-functions.php
+++ b/plugins/woocommerce/includes/wc-template-functions.php
@@ -4138,6 +4138,11 @@ function wc_get_cart_undo_url( $cart_item_key ) {
  */
 function woocommerce_output_all_notices() {
 	if ( ! function_exists( 'wc_print_notices' ) ) {
+		wc_doing_it_wrong(
+			__FUNCTION__,
+			'Function should only be used during frontend requests.',
+			'9.8.0'
+		);
 		return;
 	}
 
--- a/plugins/woocommerce/changelog/fix-55101
+++ b/plugins/woocommerce/changelog/fix-55101
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Prevent fatal error when outputting WC notices in an invalid context.


plugins/woocommerce/includes/wc-template-functions.php
```diff
@@ -4137,6 +4137,10 @@ function wc_get_cart_undo_url( $cart_item_key ) {
  * @since 3.5.0
  */
 function woocommerce_output_all_notices() {
+	if ( ! function_exists( 'wc_print_notices' ) ) {
+		return;
```
coreymckrill, line 4146:
```suggestion
		include_once WC_ABSPATH . 'includes/wc-notice-functions.php';
```

What if we just loaded the file here when necessary?

coreymckrill: Nevermind, I see that the function still wouldn't do anything since it requires that the session object is initialized first. Returning as you put here is probably the best option.

Might be worth adding a note in the doc block for this function that it is only intended for front end use. And maybe even a `_doing_it_wrong()`?

jorgeatorres: Yeah, I thought about including the file, but decided against it because it seems like this is an indication that something's off and we shouldn't just bring everything in (and there's also sessions and stuff as you mentioned).

> Might be worth adding a note in the doc block for this function that it is only intended for front end use. And maybe even a _doing_it_wrong()?

Given this isn't really interacting with anything (i.e. it's not adding to the cart or something like that) I didn't think it was worth it, but guess better safe than sorry.
I've added one in commit e74d19ddf04480238e0209d4d567657b07f79546. Let me know what you think.

coreymckrill: 👍 That looks good. I was wondering if the message string should be translatable, but it looks like we are not consistent one way or the other in the codebase, and it would only be seen when WP_DEBUG is turned on anyway, so I think it's fine to not do it here.

coreymckrill (approved):
👍 Tests well.

PR #55237, opened 2025-02-07
Product Gallery: optimise zoom in so styles are only applied to selected image + unify context usage

### LINKED ISSUE #54966: Product Gallery: optimise zoom in
**Background:**
Product Gallery has Zoom on hover feature. It’s not optimal, as all of the images (also those invisible to user) are zoomed in which is not performant.

https://github.com/user-attachments/assets/a2bc9348-7872-48d0-8138-b294d852ac4c

**Goal:**
Change the behaviour to make zoom on click, not on hover

kmanijak:

This PR includes two types of changes:
1. Optimise zoom on hover. Previously the `transform` styles were applied to all of the images in the gallery. After this PR they're applied only to the selected image. Changes in [this file](https://github.com/woocommerce/woocommerce/pull/55237/files#diff-2c862b20e50e9058f66a2c2cb1052b226fe73590d00e6b92448ce3a61a16fbb6) are responsible for that.
2. The rest of changes are to unify the usage of iAPI context. There's no point of writing getters for context properties. It should be only used to derive state from context so as an example:

```
		// DON'T
		get disableLeft() {
			return getContext().disableLeft;
		},

		// DO
                 get thumbnailTabIndex(): string {
			return state.isSelected ? '0' : '-1';
		},
```

So the rest of changes are unifying it.

Closes https://github.com/woocommerce/woocommerce/issues/54966

3. Go to the frontend
4. Open dev tools by inspecting the large image

Before | After
-- | --
Take a look like `style` attribute is applied also to other `img` elements, not visible on the page | Only visible images get the updated style
zoom-bvefore | zoom-afterdf

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryLargeImageNextPrevious.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryLargeImageNextPrevious.php
@@ -122,7 +122,7 @@ class ProductGalleryLargeImageNextPrevious extends AbstractBlock {
 
 		return sprintf(
 			'<button
-				data-wc-bind--disabled="state.disable%1$s"
+				data-wc-bind--disabled="context.disable%1$s"
 				class="wc-block-product-gallery-large-image-next-previous--button wc-block-product-gallery-large-image-next-previous-%2$s"
 			>
 				<svg xmlns="http://www.w3.org/2000/svg" width="49" height="48" viewBox="0 0 49 48" fill="none">
--- a/plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryLargeImageNextPrevious.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryLargeImageNextPrevious.php
@@ -116,10 +116,9 @@ class ProductGalleryLargeImageNextPrevious extends AbstractBlock {
 		$button_side_class         = 'left';
 		$button_disabled_directive = 'context.disableLeft';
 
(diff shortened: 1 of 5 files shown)


plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryLargeImageNextPrevious.php
```diff
@@ -122,7 +122,7 @@ protected function get_button( $button_type ) {
 
 		return sprintf(
 			'<button
-				data-wc-bind--disabled="state.disable%1$s"
+				data-wc-bind--disabled="context.disable%1$s"
```
tjcafferkey: Not a blocker because its not your code, but I really dislike this pattern because its so hard to lookup where values are being used via the IDE search functionality.

tjcafferkey (approved):
LGTM love that we're removing unnecessary selectors!

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryLargeImageNextPrevious.php
kmanijak: Great catch! Fixed in https://github.com/woocommerce/woocommerce/pull/55237/commits/03953f9afcc906faaae2421aca1eee9182e1588d

PR #55242, opened 2025-02-07
[dev] Monorepo: move blocks json manifest generation into core build steps

samueljseay:

Running the blocks json to php job and the manifest validation job are not needed during build. they're really things that only need to happen at production. They're massively slowing down incremental builds in `@woocommerce/block-library`. They also don't need to be considered by wireit at all since we run them once per build and only when building a release zip are they needed.

We need to test that build-zip can work, that the end result is a valid manifest and that if the manifest is invalid the build zip job fails.

I don't think its necessary to test that these jobs are no longer run in dev.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Monorepo: improve blocks watch turnaround time

</details>

<details>

#### Comment 

</details>


kalessil:
Converting this PR to a draft - will push some changes here.

PS: I like the direction overall and need to check a few things around the impact on CI and co.

plugins/woocommerce-blocks/package.json
```diff
@@ -48,10 +48,8 @@
 		"changelog": "XDEBUG_MODE=off composer install --quiet && composer exec -- changelogger",
 		"build": "pnpm --if-present --workspace-concurrency=Infinity --stream --filter=\"$npm_package_name...\" '/^build:project:.*$/'",
 		"build:project": "pnpm --if-present '/^build:project:.*$/'",
-		"build:project:validate-blocks-manifests": "wireit",
 		"build:project:bundle": "wireit",
-		"build:project:blocks-json-to-php": "wireit",
-		"build:check-assets": "rimraf build/* && cross-env ASSET_CHECK=true BABEL_ENV=default NODE_ENV=production webpack",
+		"build:check-assets": "npx ajv validate -s bin/block.json-validation-schema.json -d assets/js/**/block.json && rimraf build/* && cross-env ASSET_CHECK=true BABEL_ENV=default NODE_ENV=production webpack",
```
kalessil, line 52: @samueljseay @mikejolley: sanity check, `build:check-assets` script in blocks seems unused (introduced in #5634). Can you confirm I've mot missed any usages?

plugins/woocommerce-blocks/package.json
```diff
@@ -393,25 +390,21 @@
 				"tsconfig.base.json",
 				"babel.config.js",
 				"assets",
-				"packages"
+				"!assets/**/test/**",
```
kalessil: It needs testing and refinement; if it works as intended, wireit will make fewer file hashing loops.

plugins/woocommerce-blocks/webpack.config.js
```diff
@@ -30,7 +30,11 @@ const sharedConfig = {
 		timings: true,
 	},
 	watchOptions: {
-		ignored: /node_modules/,
+		ignored: [
```
kalessil: It needs testing and refinement - WIP

plugins/woocommerce-blocks/package.json
ObliviousHarmony: Something to keep in mind when working with `files` is that the order of the glob patterns can have an impact on performance. The underlying `fast-glob` package does not assign any significance to the order of the patterns but `wireit` does. [They get around this limitation by separating the globs into new groups each time an inclusion pattern follows an exclusion pattern](https://github.com/google/wireit/blob/3ea9967d0e7a6fbb1a381a8bc222691a6cd5eef1/src/util/glob.ts#L118-L195). We try to optimize this process by bundling together all of the exclusion patterns _after_ the inclusion patterns.

plugins/woocommerce-blocks/package.json
ObliviousHarmony: The `package.json` files for Blocks & Admin still have some crust from iteration in their respective repositories before they were merged. I didn't want to break any workflows and so they were never removed but things like this should be cleaned up.

plugins/woocommerce/package.json
```diff
@@ -807,12 +808,13 @@
...
 			"files": [
-				"packages/**/*.php"
+				"packages/action-scheduler/action-scheduler.php",
+				"packages/woocommerce-admin/woocommerce-admin.php"
 			],
 			"output": []
 		},
 		"watch:build:project:copy-assets": {
-			"command": "nodemon --exec \"pnpm run\" build:project",
+			"command": "nodemon --exec \"pnpm '/^build:project:copy-assets:.*$/'\"",
```
ObliviousHarmony: I'm worried that this might be a little too brittle since it relies on the asset copy operation always being `build:project:copy-assets:.*` I _do_ like the premise, however, how would you feel about checking `NODE_ENV` instead and only doing this in a production build?

ObliviousHarmony (commented):
This is a great solution and I really like the way you leaned into making the `build:project:copy-assets:blocks` fail if the manifest generation fails.

plugins/woocommerce-blocks/package.json
kalessil: Gotcha, many thanks @ObliviousHarmony! I'll drop this command, then (probably with ASSET_CHECK if no other usages).

plugins/woocommerce-blocks/package.json
kalessil: Good hint, thanks! I'll dive deeper into it as get to refining those filters.

plugins/woocommerce-blocks/webpack.config.js
```diff
@@ -37,18 +42,40 @@ const sharedConfig = {
 
 const CartAndCheckoutFrontendConfig = {
 	...sharedConfig,
+	cache: {
```
kalessil: Rought approach ATM, but good results:
- watch startup faster after the cache is populated (even if folks drop node modules, it'll repopulate)
- RAM usage drops ~500mb

Some nuances to address:
- CI-caching impact
- There are ~6 repeating sass-warning surfacing in the console - likely deprecation warnings on our styles
- perhaps dropping babel caching if no impact to build perfromance

kalessil: existing FS cache (if not exists numbers are similar to the second set of timings)

```
WP directives assets build completed (6.828s)
Payment Method Extensions assets build completed (7.107s)
Site Editor assets build completed (7.138s)
Experimental Extensions assets build completed (7.68s)
Core assets build completed (8.378s)
Cart & Checkout Frontend assets build completed (11.103s)
Frontend assets build completed (11.909s)
Main assets build completed (17.964s)
Styles assets build completed (18.22s)
```

vs default in-memory cache

```
WP directives assets build completed (7.316s)
Site Editor assets build completed (8.73s)
Payment Method Extensions assets build completed (9.089s)
Experimental Extensions assets build completed (16.114s)
Core assets build completed (16.204s)
Cart & Checkout Frontend assets build completed (37.454s)
Frontend assets build completed (43.341s)
Main assets build completed (56.481s)
Styles assets build completed (62.869s)
```

plugins/woocommerce-blocks/package.json
kalessil: I'll revert this for now and explore in another PR.

plugins/woocommerce/package.json
kalessil: Well, watch, by definition, is a non-production context, so I'm intentionally narrowing this down.
`NODE_ENV` would be added complexity; it would be easier to revert to `build:project` - changes on lines 811-812 making sure `wireit` will doing minimal work anyway,

.github/actions/setup-woocommerce-monorepo/action.yml
```diff
@@ -58,7 +58,7 @@ runs:
               key: "${{ runner.os }}-pnpm-${{ inputs.pull-package-deps }}-build:${{ ( inputs.build-type == 'backend' && 'backend' ) || 'full' }}-${{ hashFiles( 'pnpm-lock.yaml' ) }}"
               restore-keys: "${{ runner.os }}-pnpm-${{ inputs.pull-package-deps }}-build:${{ ( inputs.build-type == 'backend' && 'backend' ) || 'full' }}-"
         - name: 'Cache: node cache'
-          if: ${{ inputs.pull-package-deps != 'false' && inputs.build-type != 'backend' }}
+          if: ${{ inputs.pull-package-deps != 'false' && inputs.build-type != 'backend' && ! startsWith( inputs.pull-package-deps, '@woocommerce/block-library') }}
```
kalessil: Temporary change to not affect CI caching.

plugins/woocommerce-blocks/package.json
kalessil: To be followed up in https://github.com/woocommerce/woocommerce/pull/55325

plugins/woocommerce/package.json
kalessil: Updated.

plugins/woocommerce-blocks/package.json
kalessil: To be followed up in https://github.com/woocommerce/woocommerce/pull/55325

PR #55246, opened 2025-02-07
Product Gallery: Refactor & Performance Improvements

### LINKED ISSUE #55092: Product Gallery: Investigate performance improvements
If we have a gallery containing 10 images, we load in all of them even if they're not in view. Lets investigate ways in which we can improve upon this.

tjcafferkey:

This PR refactors how we render the `ProductGalleryLargeImage` markup. Instead of generating the markup including all of the image sources on the server we opt to return the src paths instead and use the Interactivity API directives to render them via a template loop instead.

This second change delays loading large images until the user interacts with the gallery. The thumbnail block preloads all required images into cache. Without thumbnails, only the first two images load initially, as users navigate one at a time, with more loading on interaction. If thumbnails later use smaller, separate images, this logic will need adjustment, as users could jump to an unloaded image by clicking a thumbnail.

Closes https://github.com/woocommerce/woocommerce/issues/55092

1. Load the product page and check only visible images (in the thumbnail block) are loaded, but ensure there are more.
2. Now interact with the gallery with your network tab open. Check that the rest of the images are downloaded.
3. Now remove the thumbnail block and check only two images are downloaded initially.
4. Repeat step 2
5. Regression test the zoom feature to ensure everything works as expected as well as various thumbnail settings.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryLargeImage.php
```diff
@@ -214,4 +208,51 @@ private function get_open_dialog_directives( $block_context ) {
 	protected function get_block_type_script( $key = null ) {
 		return null;
 	}
+
+	/**
+	 * Returns an array of allowed HTML for wp_kses that includes directives.
+	 *
+	 * @return array
+	 */
+	private function get_allowed_directives_html() {
```
tjcafferkey: Is this necessary? We should sanitize the HTML I think but we need to ensure relevant attribute data isn't removed. But this means maintaining a list of these attributes too.

plugins/woocommerce/src/Blocks/Utils/ProductGalleryUtils.php
```diff
@@ -55,13 +95,8 @@ public static function get_product_gallery_images( $post_id, $size = 'full', $at
 					$product_image_html_processor = new \WP_HTML_Tag_Processor( $product_image_html );
 					$product_image_html_processor->next_tag( 'img' );
 					$product_image_html_processor->set_attribute(
```
tjcafferkey, line 97: Creating context for the sake of storing an image ID felt unnecessary. I've removed this and instead store this as a data attribute on the element. We need this when we trigger an event on the image, so we have direct access to this via `event.target.getAttribute`. It also removes the confusion of contexts being merged at various levels.

plugins/woocommerce/src/Blocks/Utils/ProductGalleryUtils.php
```diff
@@ -8,6 +8,46 @@
...
+	 *
+	 * @param \WC_Product $product The product object to retrieve the gallery images for.
+	 * @return array An array of image data for the product gallery.
+	 */
+	public static function get_product_gallery_image_data( $product ) {
+		$image_data = array();
+		$image_ids  = self::get_product_gallery_image_ids( $product );
+
+		foreach ( $image_ids as $image_id ) {
+			if ( '0' === $image_id ) {
```
kmanijak, line 22: Is it guaranteed placeholder image has id 0?

plugins/woocommerce-blocks/tests/e2e/tests/product-gallery/inner-blocks/product-gallery-large-image/product-gallery-large-image.block_theme.spec.ts
```diff
@@ -99,7 +99,8 @@ test.describe( `${ blockData.name }`, () => {
 					( el ) => el.style
 				);
 
-				expect( style.transform ).toBe( 'scale(1)' );
+				expect( style.transform ).not.toBe( 'scale(1.3)' );
+				expect( style.transform ).toBe( '' );
```
kmanijak: The second expect "includes" the first one

```suggestion
				expect( style.transform ).toBe( '' );
```

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
```diff
@@ -34,19 +41,66 @@ const getArrowsState = ( imageNumber: number, totalImages: number ) => ( {
...
 const productGallery = {
 	state: {
 		get isSelected() {
-			const { selectedImageNumber, imageIds, imageId } = getContext();
-			return selectedImageNumber === imageIds.indexOf( imageId ) + 1;
+			const { selectedImageNumber, imageIds, selectedImageId } =
+				getContext();
+			return (
+				selectedImageNumber === imageIds.indexOf( selectedImageId ) + 1
+			);
```
kmanijak: That logic returns true for every image now. Previously it relied on _individual_ `imageId` from each image. Now, it relies on `selectedImageNumber` and `selectedImageId` which is the same no matter in which image context you run this state.

What I mean is, let's say:
- `selectedImageNumber` is `2`
- `selectedImageId` is `123`

`isSelected` is fired in context of image:
- selected image: `123` -> condition is true
- different image: `456` -> condition is also true.

With the changes you applied, I guess this should be based on `target.getAttribute...` to get image ID.

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
```diff
@@ -34,19 +41,66 @@ const getArrowsState = ( imageNumber: number, totalImages: number ) => ( {
...
 		},
 		get imageIndex(): number {
-			const { imageIds, imageId } = getContext();
-			return imageIds.indexOf( imageId );
+			const { imageIds, selectedImageId } = getContext();
+			return imageIds.indexOf( selectedImageId );
```
kmanijak, line 94: Similar to the above `imageIndex` will always return index of currently selected image.

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
kmanijak: Ok, I see you removed most usages of `isSelected` except for `thumbnailTabIndex`. This means all of the images get `0` while according to previous logic only selected one should get `0` and others `-1`.

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
kmanijak: I checked and seems this state is no longer used anywhere. It can be removed completely.

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryLargeImage.php
kmanijak: Based on my knowledge that should no longer be necessary as the directives should be allowed in sanitization since WP 6.5: https://core.trac.wordpress.org/ticket/61052. If it's needed then it's probably a bug.

kmanijak (changes requested):
Thanks for the effort and performance improvement! In basic scenario images are loading as necessary! However, I found some issues here and there:

1. For some reason the arrows no longer render for me on this branch:

image

2. Users can use arrows to change images. They don't get loaded when they do. FWIW I created an issue that Pager and Arrows don't react to this as well: https://github.com/woocommerce/woocommerce/issues/55952

https://github.com/user-attachments/assets/f0d665e9-a82e-4123-aaf2-b7090b97a27c

3. I feel like introducing `selectedImageId` is a duplication of `selectedImageNumber`. One can be easily derived from the other. How do you feel about switching the logic fully to use only one of them? I don't mind removing `selectedImageNumber` in favour of `selectedImageId`.

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryLargeImage.php
tjcafferkey: https://github.com/woocommerce/woocommerce/pull/55246/commits/dc87f25bb51ef91f37ba0322853247b5294d80d8

plugins/woocommerce/src/Blocks/Utils/ProductGalleryUtils.php
tjcafferkey: `get_product_gallery_image_ids` sets the `id` as `0` so we can guarantee it I think yes.

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
tjcafferkey: Resolved in https://github.com/woocommerce/woocommerce/pull/55246/commits/b2ebac9617f41f1e6cc13926a2ec811fee75b237

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
tjcafferkey: Resolved in https://github.com/woocommerce/woocommerce/pull/55246/commits/b2ebac9617f41f1e6cc13926a2ec811fee75b237

tjcafferkey:
>I don't mind removing selectedImageNumber in favour of selectedImageId.

This is a good idea, I will make the update on monday @kmanijak

cpapazoglou:
Great work on the Product Gallery performance improvements! 👏

While reviewing the test coverage, I noticed some areas where we could strengthen the test suite to ensure the performance improvements remain stable:

### Current Coverage
✅ Basic functionality (thumbnails, navigation, pager)
✅ Touch interactions
✅ Basic image loading checks

### Suggested Additional Test Coverage

1. **Lazy Loading Behavior**
```typescript
test('should initially load only first two images', async () => {
    // Verify first two images have src attributes
    // Verify subsequent images have empty src
});
test('should load more images after user interaction', async () => {
    // Verify images load progressively as user navigates
});
```
2. **Performance Metrics**
```typescript
test('should meet performance benchmarks', async () => {
    // Network throttling setup
    // Initial load time measurement
    // Network request count verification
});
```
3. **Edge Cases**
```typescript
test('should handle large galleries gracefully', async () => {
    // Memory usage monitoring
    // Smooth scrolling verification
});

test('should handle network failures gracefully', async () => {
    // Simulate failed image loads
    // Verify error states
});
```
4. **Network Conditions**
- Add tests with simulated slow connections
- Verify behavior under poor network conditions
- Test bandwidth usage optimization
 
These additions would help ensure:
- The lazy loading implementation works as intended
- Performance improvements are measurable and stable
- Edge cases are handled gracefully
- The gallery remains responsive under various conditions
Let me know if you'd like help implementing any of these test cases! 🚀

Co-authored with Cursor during the Cursor Workshop ☺️

tjcafferkey:
@kmanijak this should be ready for another review now, I've just pushed a linting fix but all tests should pass also.

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
```diff
@@ -20,35 +19,122 @@ export interface ProductGalleryContext {
...
-			return state.isSelected ? '0' : '-1';
+		/**
+		 * The processed image data.
+		 *
+		 * @return {Object} The processed image data.
+		 */
+		get processedImageData() {
+			// The thumbnail block preloads all required images into cache. Without thumbnails, only the first two images load initially,
+			// as users navigate one at a time, with more loading on interaction. If thumbnails later use smaller, separate images, this
+			// logic will need adjustment, as users could jump to an unloaded image by clicking a thumbnail.
```
kmanijak, line 104: > If thumbnails later use smaller, separate images, this
> logic will need adjustment, as users could jump to an unloaded image by clicking a thumbnail.

This can happen now if the merchant sets number of Thumbnails bigger than 3.

One idea would be to treat `mouseenter` as an interaction that triggers loading images. We use this strategy in Product Collection pagination. Hovering on the page triggers preloading the page.

I think it could be a good strategy here as well.

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
```diff
@@ -57,22 +143,44 @@ const productGallery = {
 				context.imageIds.length
 			);
 
-			context.selectedImageNumber = newImageNumber;
+			actions.userHasInteracted();
```
kmanijak, line 148: Please rebase with `trunk` so this PR is included: https://github.com/woocommerce/woocommerce/pull/55976. At the moment changing images using keyboard arrows doesn't load the images.

https://github.com/user-attachments/assets/8e9b5160-a4cc-4a44-8432-1ebbcd2d40ed

plugins/woocommerce-blocks/tests/e2e/tests/product-gallery/product-gallery.block_theme.spec.ts
```diff
@@ -36,20 +36,34 @@ const test = base.extend< { pageObject: ProductGalleryPage } >( {
 	},
 } );
 
+/**
+ * Get the image id from the image element. Sometimes the image id is in the data-image-id attribute, sometimes in the id attribute.
```
kmanijak: Do you know the reason why it's either this or that? I'm just curious, I don't think I caught it from a code elsewhere 🤔

kmanijak (commented):
WooCommerce PR #55246 Testing Report

**Basic Scenario**
- Loaded the page with the network tab open.
- Verified that interacting with the Product Gallery block loads more than the first two images.
- ✅ Images load correctly.

**Thumbnail Click Testing**
- Clicked one of the last thumbnails.
- ✅ Image loads correctly.

**Arrow Click Testing**
- Used the arrow buttons to navigate through the gallery.
- Verified that clicking arrows is treated as an interaction.
- ✅ Images load correctly.

**Keyboard Enter Key Testing**
- Focused on one of the last thumbnails and pressed Enter.
- ✅ Image loads correctly.

**Keyboard Arrow Key Testing**
- Focused on the gallery and used keyboard arrow keys to navigate.
- ❌ Images do not load correctly.
- I included a video in an inline comment

**Variable Product Testing**
- Navigated to a variable product page.
- Changed the product variation (e.g., switched to "White" color).
- Verified that the image updates correctly when selecting a new variation.
- ✅ Image updates as expected.

**Regression Testing**
- Verified that the zoom on hover functionality works correctly.
- ✅ Zoom on hover functions as expected.

**Opened the full-page gallery and ensured it operates correctly.**
- ✅ Full-page gallery opens and functions as expected.

While submitting this review I see you actually merged `trunk` so I'll verify the keyboard arrows scenario once again.

plugins/woocommerce-blocks/tests/e2e/tests/product-gallery/product-gallery.block_theme.spec.ts
tjcafferkey: `id` needs to be semantically unique so we can't have the large image with `id="32"` and then its corresponding thumbnail with the same `id`.

I can update them to be streamlined and both use `data-image-id` I think that might be better to be honest.

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/frontend.tsx
tjcafferkey: >This can happen now if the merchant sets number of Thumbnails bigger than 3.

If the user has 10 images, and sets the thumbnails to be 5 then 5 images are loaded in, no?

plugins/woocommerce-blocks/tests/e2e/tests/product-gallery/product-gallery.block_theme.spec.ts
kmanijak: > id needs to be semantically unique so we can't have the large image with id="32" and then its corresponding thumbnail with the same id.

Ah right, makes sens! Thanks for explaining!

> I can update them to be streamlined and both use data-image-id I think that might be better to be honest.

Agree!

tjcafferkey: Done in https://github.com/woocommerce/woocommerce/pull/55246/commits/7854ac6a434d60817f67e9de69b500a632f7ee03

kmanijak: Remove or adjust the comment! 😆

plugins/woocommerce-blocks/tests/e2e/tests/product-gallery/product-gallery.page.ts
```diff
@@ -69,17 +69,31 @@ export class ProductGalleryPage {
...
+				return imageId;
 			}
-			return JSON.parse( context ).imageId;
+
+			const dataImageId = element.getAttribute( 'data-image-id' );
+			if ( dataImageId ) {
+				return dataImageId;
+			}
+
+			return null;
```
kmanijak: And update this helper as well 😄

kmanijak (changes requested):
There must be something wrong with the recent refactor, the images don't change when clicking the arrows. Pager gets updated, also when opening the full page gallery it works fine but images don't scroll.

I checked moving back 3 commits and it works just fine. With the newest changes it doesn't:

https://github.com/user-attachments/assets/81270ca8-da2b-4b2e-aee8-980a164a78c4

plugins/woocommerce-blocks/tests/e2e/tests/product-gallery/product-gallery.page.ts
tjcafferkey: https://github.com/woocommerce/woocommerce/pull/55246/commits/4b0bb9d9314e98dbc62458b41236086d7d9b152a

tjcafferkey: https://github.com/woocommerce/woocommerce/pull/55246/commits/848f5253553adcef0f343c09413e517f3dce977c

tjcafferkey:
@kmanijak I've fixed the issue with the image in https://github.com/woocommerce/woocommerce/pull/55246/commits/cfe47ce2ce6e389c10b50790ca8ec25c512c1738 and I can explain why the tests were passing.

The logic was working correctly, so when you changed image all the of 'active' classes were assigned correctly, it is these active classes which the E2E tests to query the elements for tests.

However, the actual functionality to scroll into view wasn't working because we didnt update the query selector.

Do you think we can follow up in this PR with an E2E test for this?

kmanijak (approved):
I followed the scenarios from the [other comment](https://github.com/woocommerce/woocommerce/pull/55246#pullrequestreview-2656334243).

All works fine, thanks for patience Tom! 🚢

PR #55250, opened 2025-02-07
[Experimental] Fix add to cart action url to be the same as the current url

### LINKED ISSUE #53284: [Enhancement]: Find a way to identify when a block is descendant of the SingleProduct block
The current method of identifying whether a block is a descendant of the SingleProduct block is [not ideal](https://github.com/woocommerce/woocommerce/pull/53215/files#r1861908953). This approach could be improved for better maintainability and reliability.

Let's try to find a better way to [handle is-descendent-of-single-product-block](https://github.com/woocommerce/woocommerce/blob/35a975865fe1bfa1e7aaabc5a2a02a6578e545f5/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php#L161).

mdperez86:

Closes #53284

1. Make sure `Redirect to the cart page after successful addition` is unchecked under `/wp-admin/admin.php?page=wc-settings&tab=products`
2. Create a new post
5. Add the single product block
6. Select a simple product and save the post
7. Go to view the post created
8. The simple product should be shown in the frontend
9. Select 1 and click add to cart
10. Notice that the form does not perform a new redirect and it remains in the same page.
11. The same should also happens when trying to add a product cart form the single product page.
12. If the check `Redirect to the cart page after successful addition` is checked
13. Then a redirection to the cart page should happens after adding a product to the cart

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartForm.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartForm.php
@@ -18,25 +18,6 @@ class AddToCartForm extends AbstractBlock {
 	protected $block_name = 'add-to-cart-form';
 
 	/**
-	 * Initializes the AddToCartForm block and hooks into the `wc_add_to_cart_message_html` filter
-	 * to prevent displaying the Cart Notice when the block is inside the Single Product block
-	 * and the Add to Cart button is clicked.
-	 *
-	 * It also hooks into the `woocommerce_add_to_cart_redirect` filter to prevent redirecting
-	 * to another page when the block is inside the Single Product block and the Add to Cart button
-	 * is clicked.
-	 *
-	 * @return void
-	 */
-	protected function initialize() {
-		parent::initialize();
-		// Scope the excecution of this hook to be only when the add to cart form is submitted.
-		if ( isset( $_SERVER['REQUEST_METHOD'] ) && 'POST' === $_SERVER['REQUEST_METHOD'] ) {
-			add_filter( 'woocommerce_add_to_cart_redirect', array( $this, 'add_to_cart_redirect_filter' ), 10, 1 );
-		}
-	}
-
-	/**
 	 * Get the block's attributes.
 	 *
 	 * @param array $attributes Block attributes. Default empty array.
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/src/Blocks/BlockTypes/AddToCartForm.php
```diff
@@ -155,6 +138,14 @@ protected function render( $attributes, $content, $block ) {
 		$is_external_product_with_url = $product instanceof \WC_Product_External && $product->get_product_url();
 		$is_stepper_style             = 'stepper' === $attributes['quantitySelectorStyle'] && ! $product->is_sold_individually() && Features::is_enabled( 'add-to-cart-with-options-stepper-layout' );
 
+		add_filter( 'woocommerce_add_to_cart_form_action', function ( $url ) use ( $is_descendent_of_single_product_block ) {
+			if ( $is_descendent_of_single_product_block ) {
```
Aljullu: Could this `if` be around the `add_filter()` instead of inside? This way we avoid registering the filter unnecessarily. Ie:

```PHP
if ( $is_descendent_of_single_product_block ) {
  add_filter( ... );
}
```

Also, not a strong opinion, but wondering if we should maybe add a `remove_filter()` after we rendered the template? Just to make sure we leave things as they were before the block was rendered.

Aljullu (commented):
The approach looks good to me and it works as expected, thanks for investigating this, @mdperez86!

I left one comment about the way we add the filter, but not a big deal. I think we should also remove `[Experimental]` from the PR title, as this PR impacts a block available in production builds.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartForm.php
mdperez86: done -> [5a1ec50](https://github.com/woocommerce/woocommerce/pull/55250/commits/5a1ec50fd64982fbabba2669c48d73052cb8eebd)

peterfabian (approved):
This fixes the issue we saw in Gift Cards, so 👍  from me.

Aljullu (approved):
LGTM as well. Really nice cleanup. :clap:

PanosSynetos:
Awesome teamwork here! Thanks for the prompt fix and the reviews

PR #55254, opened 2025-02-07
[Experimental] Add additional field validation rule parsing in Checkout block

### LINKED ISSUE #54478: Add validation parsing for additional fields in JS

senadir:

Closes https://github.com/woocommerce/woocommerce/issues/54478

Use the following fields to test:
```php
add_action( 'woocommerce_init', function() {
		array(
			'id'       => 'plugin-namespace/pickup-window',
			'label'    => 'Pickup Window',
			'location' => 'address',
			'type'     => 'text',
			'rules'    => array(
				'validation' => array(
					'allOf'        => array(
						array(
							'type'         => 'string',
							'pattern'      => '^[0-9]{2}:[0-9]{2}-[0-9]{2}:[0-9]{2}$',
							'errorMessage' => 'Must be in HH:MM-HH:MM format',
						),
						array(
							'type'         => 'string',
							'pattern'      => '^([01][0-9]|2[0-3]):[0-5][0-9]-([01][0-9]|2[0-3]):[0-5][0-9]$',
							'errorMessage' => 'Hours must be 00-23, minutes must be 00-59',
						),
						array(
							'type'         => 'string',
							'not'          => array(
								'enum' => array( '00:00-23:59' ),
							),
							'errorMessage' => 'Cannot specify full day range (00:00-23:59)',
						),
					),
					'errorMessage' => 'Please enter a valid pickup window in the format HH:MM-HH:MM',
				),
				'required'   => array(
					'cart'    => array(
						'properties' => array(
							'prefers_collection' => array(
								'type'  => 'boolean',
								'const' => true,
							),
						),
					),
				),
			),
		),
	);

		array(
			'id'       => 'plugin-namespace/confirm-pickup-window',
			'label'    => 'Confirm Pickup Window',
			'location' => 'address',
			'type'     => 'text',
			'rules'    => array(
				'validation' => array(
					'type'         => 'string',
					'const'        => array( '$data' => '1/plugin-namespace~1pickup-window' ),
					'errorMessage' => 'Please confirm your pickup window',
				),
				'required'   => array(
					'cart'    => array(
						'properties' => array(
							'prefers_collection' => array(
								'type'  => 'boolean',
								'const' => true,
							),
						),
					),
					'address' => array(
						'properties' => array(
							'pickup_window' => array(
								'type' => 'string',
								'not'  => array(
									'const' => '',
								),
							),
						),
					),
				),
			),
		),
	);
}, 10, 1 );

```
1. Enable Local Pickup and have shipping rates.
2. In Checkout, select regular shipping method.
3. In shipping and billing address confirm that Pickup and confirm Pickup are both optional.
4.  Focus the fields without typing and then blur them out, **ensure no error is shown.**
5. Switch to local pickup, ensure the fields are now required.
6. Focus the fields and unfocus them, error should be shown.
7. Start typing in the first field, type `12:00-12:00`, type the same value in the other field. There should be no validation errors.
8. Modify the first field to something else, an error should surface in the second field.
9. Change the first field to `00:00-23:59` and blur our, make sure you get the custom error `Cannot specify full day range (00:00-23:59)`.
10. Fix both fields, submit the order, you should be able to submit.

### Regression testing
1. For country you know the postcode pattern of, type a correct postcode. There should be no errors.
2. Change the code to a non-working one, you should see an error.
3. Change email to a wrong email, you should see an error on blur.
4. Enable account creation post and during Checkout with manual password.
5. Check "create an account" in checkout.
6. Focus and unfocus the password field in Checkout, you should see an error.
7. Start typing a short password and blur, you should see an error.
8. type a long strong password and blur, there should be no error.
9. Don't create account and place the order.
10. In the thank you page, start typing a short password
11. you should see an error and you can't submit the account
12. fix it to a longer account and you can submit the account.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 
This is behind a feature flag.

</details>


plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/form.tsx
```diff
@@ -81,6 +86,46 @@ const Form = < T extends AddressFormValues | ContactFormValues >( {
 		Record< string, ValidatedTextInputHandle | null >
 	>( {} );
 
+	const { errors, previousErrors } = useFormValidation(
+		formFields,
+		addressType
+	);
+
+	useEffect( () => {
```
senadir, line 97: The typefest happening here can be avoidable if I sit down and refactor our form types

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/form.tsx
```diff
@@ -81,6 +86,46 @@ const Form = < T extends AddressFormValues | ContactFormValues >( {
...
+				inputRef?.setErrorMessage( error );
+
+				dispatch( validationStore ).setValidationErrors( {
+					[ `${ addressType }_${ key }` ]: {
+						message: error,
+						hidden:
+							!! inputRef?.isFocused() ||
+							values[ key as keyof T ] === '',
+					},
+				} );
```
senadir: We need to both add a validity error and data store error, but I think if the field is focused we can just do setErrorMessage

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/form.tsx
```diff
@@ -370,15 +415,6 @@ const Form = < T extends AddressFormValues | ContactFormValues >( {
...
 						} }
-						customValidation={ ( inputObject: HTMLInputElement ) =>
-							customValidationHandler(
-								inputObject,
-								field.key,
-								objectHasProp( values, 'country' )
-									? values.country
-									: ''
-							)
-						}
```
senadir, line 384: This is now part of formValidation

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/use-form-fields.ts
```diff
@@ -43,12 +48,26 @@ export const useFormFields = (
...
+							key === 'customer'
+					)
+				) {
+					schema = {
+						type: 'object',
+						properties: defaultConfig.rules.required,
+					};
+				} else {
+					schema = defaultConfig.rules.required;
+				}
```
senadir, line 61: We now support if you pass the whole object with a header or just the inside part of it

senadir:
pushing to see what breaks

mikejolley:
@senadir if this is ready to review it needs testing instructions.

senadir:
I will let you know when it's ready for testing, I pushed it to see what tests fail

plugins/woocommerce-blocks/packages/components/text-input/validated-text-input.tsx
```diff
@@ -117,29 +121,39 @@ const ValidatedTextInput = forwardRef<
...
+				);
+
+				if ( validityMessage ) {
+					setValidationErrors( {
+						[ errorIdString ]: {
+							message: validityMessage,
+							hidden: errorsHidden,
+						},
+					} );
+				}
```
senadir, line 148: This was setting errors even when `validityMessage` was empty string.

plugins/woocommerce-blocks/packages/components/text-input/test/validated-text-input.tsx
```diff
@@ -200,8 +206,8 @@ describe( 'ValidatedTextInput', () => {
 
 		await expect(
 			select( validationStore ).getValidationError( 'test-input' )
-		).not.toBe( undefined );
```
senadir, line 203: This was passing because with the bug in `plugins/woocommerce-blocks/packages/components/text-input/validated-text-input.tsx`, the error was `{ message: "", hidden: true }`. There was no error to set simply because the field has no browser error to set (it has a value) and `customValidation` should be the be the setting the error, but it didn't.

plugins/woocommerce-blocks/assets/js/settings/shared/default-fields.ts
```diff
@@ -48,9 +48,9 @@ export interface FormField {
 	attributes?: Record< keyof CustomFieldAttributes, string >;
 	// The rules for the field.
 	rules?: {
-		required?: JSONSchemaType< DocumentSchema >;
```
senadir, line 51: There is no such thing as DocumentSchema type :)

plugins/woocommerce-blocks/assets/js/blocks/checkout/inner-blocks/checkout-contact-information-block/create-password.tsx
```diff
@@ -15,7 +15,7 @@ const CreatePassword = () => {
 		return {
 			customerPassword: store.getCustomerPassword(),
 		};
-	} );
+	}, [] );
```
senadir, line 18: Just good hygiene

plugins/woocommerce-blocks/packages/components/text-input/validated-text-input.tsx
```diff
@@ -117,29 +121,39 @@ const ValidatedTextInput = forwardRef<
 
 				if (
 					inputObject.checkValidity() &&
-					customValidationRef.current( inputObject )
+					customValidationRef.current( inputObject ) &&
+					errorsHidden
```
senadir, line 125: We only clear errors if we're showing errors, this seems to now fixes everything.

plugins/woocommerce-blocks/packages/components/text-input/validated-text-input.tsx
```diff
@@ -117,29 +121,39 @@ const ValidatedTextInput = forwardRef<
...
-							label,
-							inputObject,
-							customValidityMessage
-						),
-						hidden: errorsHidden,
-					},
-				} );
+				if ( ! errorsHidden ) {
+					showValidationError( errorIdString );
+				}
```
senadir, line 133: Showing errors on blur instead of relaying on creating a new error

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/use-form-validation.ts
```diff
@@ -0,0 +1,270 @@
...
+		}
+	}
+
+	const customValidation =
+		values &&
+		formFields
+			.map( ( field ) => {
+				if ( schemaErrorsMap[ field.key ] ) {
+					return [ field.key, schemaErrorsMap[ field.key ] ];
+				}
```
senadir: We're merging schema errors with customValidation errors.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/use-form-validation.ts
```diff
@@ -0,0 +1,270 @@
...
+import { isEmail } from '@wordpress/url';
+
+const getFieldLabelCasing = ( fieldLabel: string ): string => {
+	const localeData = getLocaleData();
+	const shouldKeepOriginalCase = [ 'de', 'de_AT', 'de_CH' ].includes(
+		localeData?.[ '' ]?.lang ?? 'en'
+	);
+
+	return shouldKeepOriginalCase ? fieldLabel : fieldLabel.toLowerCase();
+};
```
senadir: Copy paste from other places

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/form.tsx
mikejolley: Yeah I was gonna say, this looks like it needs some type of refactor or inferred type.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/form.tsx
```diff
@@ -81,6 +86,49 @@ const Form = < T extends AddressFormValues | ContactFormValues >( {
...
+		);
+
+		if ( previousErrors ) {
+			Object.entries(
+				previousErrors as Record< keyof FormFields, string >
+			).forEach( ( [ key ] ) => {
+				const inputRef = inputsRef.current[ key ];
+				// If error was previously set but is now cleared
+				if ( ! errors[ key as keyof FormFields ] ) {
+					dispatch( validationStore ).clearValidationError(
```
mikejolley, line 120: I think my main concern here is the adding and clearing of notices happens on each render, which is triggered by any field change (typing). Is this a problem and what alternatives are there if any?

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/use-form-fields.ts
```diff
@@ -24,7 +24,12 @@ export const useFormFields = (
 	// Default fields from settings.
 	defaultFields: FormFields,
 	// Form type, can be billing, shipping, contact, additional-information, or calculator.
-	formType: string,
+	formType:
+		| 'billing'
+		| 'shipping'
+		| 'contact'
+		| 'additional-information'
```
mikejolley: This might be a valid but its another list of keys similar to those defined in the Checkout Fields classes but slightly different. For example, the PHP side has locations for billing, shipping, contact, **order**. `additional-information` was deprecated on the PHP side.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/use-form-fields.ts
```diff
@@ -62,6 +81,7 @@ export const useFormFields = (
 			if (
 				defaultConfig.rules.hidden &&
 				typeof defaultConfig.rules.hidden === 'object' &&
+				! Array.isArray( defaultConfig.rules.hidden ) &&
 				Object.keys( defaultConfig.rules.hidden ).length > 0
```
mikejolley: Whats all this then? Same question for required above. Why are all these conditionals necessary? If this is necessary, could you make a helper that works for all rule types?

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/use-form-validation.ts
```diff
@@ -0,0 +1,269 @@
...
+	usePrevious,
+} from '@woocommerce/base-hooks';
+import type { JSONSchemaType, ErrorObject } from 'ajv';
+import { __, sprintf, getLocaleData } from '@wordpress/i18n';
+import { useRef } from '@wordpress/element';
+import fastDeepEqual from 'fast-deep-equal/es6';
+import { isPostcode } from '@woocommerce/blocks-checkout';
+import { isEmail } from '@wordpress/url';
+
+const getFieldLabelCasing = ( fieldLabel: string ): string => {
```
mikejolley: Can we add a doc block why this is needed? It seems like it might be a pain to maintain if its needed.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/use-form-validation.ts
```diff
@@ -0,0 +1,269 @@
...
+const getFieldKey = (
+	instancePath: ErrorObject[ 'instancePath' ]
+): keyof FormFields => {
+	return instancePath
+		.split( '/' )
+		.pop()
+		?.replace( '~1', '/' ) as keyof FormFields;
+};
+
+const getErrorsMap = (
```
mikejolley, line 49: FWIW I think it would be fine to leave out error handling and use generic errors until we finalise how we want to handle them on both PHP and JS sides for consistency.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/use-form-validation.ts
```diff
@@ -0,0 +1,269 @@
...
+			values = {} as Record< keyof FormFields, string >;
+			break;
+	}
+
+	const partialSchema = formFields.reduce( ( acc, field ) => {
+		if (
+			! field.hidden &&
+			typeof field.rules?.validation === 'object' &&
+			! Array.isArray( field.rules.validation ) &&
+			( field.required || values[ field.key as keyof typeof values ] )
```
mikejolley: This type of logic again is repeated a lot throughout and should be abstracted.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/use-form-validation.ts
```diff
@@ -0,0 +1,269 @@
...
+		)
+	) {
+		currentResults.current = Object.fromEntries(
+			customValidation as [ keyof FormFields, string ][]
+		) as Record< keyof FormFields, string >;
+	}
+
+	return {
+		errors: currentResults.current,
+		previousErrors,
```
mikejolley, line 261: Should we return `previousErrors` here or allow the consumer of this hook to do it if the use case demands it?

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/utils.ts
```diff
@@ -33,11 +33,11 @@ export const createFieldProps = (
...
 	label: ( field?.required ? field?.label : field?.optionalLabel ) || '',
-	autoCapitalize: field?.autocapitalize,
-	autoComplete: field?.autocomplete,
-	errorMessage: field?.errorMessage,
-	required: field?.required,
-	placeholder: field?.placeholder,
+	autoCapitalize: field?.autocapitalize || '',
+	autoComplete: field?.autocomplete || '',
+	errorMessage: field?.errorMessage || '',
+	required: field?.required || false,
```
mikejolley: In HTML, `false` props need to be omitted. Is this handled somewhere?

plugins/woocommerce-blocks/assets/js/base/hooks/use-schema-parser.ts
```diff
@@ -169,9 +193,13 @@ export interface DocumentObject {
...
-				billing_address: CoreAddress;
-				shipping_address: CoreAddress;
-				address: CoreAddress;
+				billing_address: BillingAddress;
+				shipping_address: ShippingAddress;
+				address: T extends 'billing'
+					? BillingAddress
+					: T extends 'shipping'
+					? ShippingAddress
+					: undefined;
```
mikejolley: I think we went with `null` on php side if it matters.

plugins/woocommerce-blocks/assets/js/settings/shared/default-fields.ts
```diff
@@ -100,6 +100,7 @@ export interface CoreAddress {
 	state: string;
 	postcode: string;
 	phone: string;
+	email?: string;
```
mikejolley: What did this fix? Its not common to both address types.

plugins/woocommerce-blocks/packages/components/text-input/validated-text-input.tsx
mikejolley: Can you elaborate here?

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/form.tsx
senadir: We can do 2 things:
- Debounce validation (I think will result in a janky UI).
- Only run validation and add errors if no field is focused (will need some trigger somewhere).

I might proceed with debounce once the UX is good.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/use-form-fields.ts
senadir: will switch up to `order`

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/use-form-fields.ts
senadir: The extra change was needed to account for an empty array that comes from the server when no rules are set, I can simplify this set of rules thou.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/use-form-validation.ts
senadir: I first tried to centralize it to a single package but causes circular deps issue, will add a comment, this is used in like 4 or 5 places, so not good.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/use-form-validation.ts
senadir: Only handling I'm doing is for pattern which I can remove.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/use-form-validation.ts
senadir: I agree I think it can be managed at consumer level

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/utils.ts
senadir: I might undo this change. I think `required` is handled at input component level

plugins/woocommerce-blocks/assets/js/settings/shared/default-fields.ts
senadir: I can't recall, let's see if type fixing fixes that.

plugins/woocommerce-blocks/packages/components/text-input/validated-text-input.tsx
senadir: Previously this hook would clear up all errors and generate news ones with hidden or not based on `errorsHidden`, this didn't reveal errors coming from other places, so the fix above fixes that.

mikejolley (commented):
In testing it worked for the most part. I noticed 2 oddities but didn't look into the cause.

I noticed that when the "pickup window" fields are optional, if you enter one, the confirmation one is not made required. The rules are therefore "AND" rather than "OR". How would you fix this as the implementor of the field?

I noticed that if you submit a form, all errors are visible even for fields not yet shown. I haven't confirmed if this happens in trunk, but you can reproduce it by:

1. Select local pickup
2. Choose a pickup window. Do not confirm it.
3. Place order. Validation errors are shown.
4. Toggle to shipping.
5. Toggle to ship to a different address
6. Billing form is open with errors for all fields visible.

senadir:
For the `AND`/`OR` thing, using `anyOf` seems to work:
```php
	array(
		'id'       => 'plugin-namespace/confirm-pickup-window',
		'label'    => 'Confirm Pickup Window',
		'location' => 'address',
		'type'     => 'text',
		'rules'    => array(
			'validation' => array(
				'type'         => 'string',
				'const'        => array( '$data' => '1/plugin-namespace~1pickup-window' ),
				'errorMessage' => 'Please confirm your pickup window',
			),
			'required'   => array(
				'type'       => 'object',
				'properties' => array(
					'anyOf' => array(
						'cart'    => array(
							'properties' => array(
								'prefers_collection' => array(
									'type'  => 'boolean',
									'const' => true,
								),
							),
						),
						'address' => array(
							'properties' => array(
								'pickup_window' => array(
									'type' => 'string',
									'not'  => array(
										'const' => '',
									),
								),
							),
						),
					),
				),
			),
		),
	),
);
```

plugins/woocommerce-blocks/assets/js/base/context/event-emit/utils.ts
```diff
@@ -35,7 +35,7 @@ export enum noticeContexts {
 	BILLING_ADDRESS = 'wc/checkout/billing-address',
 	SHIPPING_METHODS = 'wc/checkout/shipping-methods',
 	CHECKOUT_ACTIONS = 'wc/checkout/checkout-actions',
-	ORDER_INFORMATION = 'wc/checkout/additional-information',
+	ORDER_INFORMATION = 'wc/checkout/order-information',
```
mikejolley, line 38: What is the consequence of this change?

plugins/woocommerce/src/Blocks/Utils/CartCheckoutUtils.php
```diff
@@ -288,9 +288,23 @@ public static function is_overriden_by_custom_template_content( string $block ):
 	public static function get_country_data() {
 		$billing_countries  = WC()->countries->get_allowed_countries();
 		$shipping_countries = WC()->countries->get_shipping_countries();
-		$country_locales    = wc()->countries->get_country_locale();
 		$country_states     = wc()->countries->get_states();
 		$all_countries      = self::deep_sort_with_accents( array_unique( array_merge( $billing_countries, $shipping_countries ) ) );
+		$country_locales    = array_map(
```
mikejolley, line 293: What were these changes for?

mikejolley (commented):
Looked over code a few times now so should be ok. I left 2 questions.

There is a bug however. Reproduce as follows:

1. Select shipping
2. Enter invalid pickup time and confirm time
3. Toggle Local pickup
4. Enter valid times
5. Place order
6. See error coming from shipping fields which is unresolvable

plugins/woocommerce-blocks/assets/js/base/context/event-emit/utils.ts
senadir: Nothing really, this is client side only it seems.

plugins/woocommerce/src/Blocks/Utils/CartCheckoutUtils.php
senadir: changed priority to index on the server, no point is processing that manually on the client.

senadir:
I can't reproduce that issue with the following corrected fields:

```php
add_action( 'woocommerce_init', function() {
	array(
		'id'       => 'plugin-namespace/pickup-window-2',
		'label'    => 'Pickup Window',
		'location' => 'address',
		'type'     => 'text',
		'rules'    => array(
			'validation' => array(
				'allOf'        => array(
					array(
						'type'         => 'string',
						'pattern'      => '^[0-9]{2}:[0-9]{2}-[0-9]{2}:[0-9]{2}$',
						'errorMessage' => 'Must be in HH:MM-HH:MM format',
					),
					array(
						'type'         => 'string',
						'pattern'      => '^([01][0-9]|2[0-3]):[0-5][0-9]-([01][0-9]|2[0-3]):[0-5][0-9]$',
						'errorMessage' => 'Hours must be 00-23, minutes must be 00-59',
					),
					array(
						'type'         => 'string',
						'not'          => array(
							'enum' => array( '00:00-23:59' ),
						),
						'errorMessage' => 'Cannot specify full day range (00:00-23:59)',
					),
				),
				'errorMessage' => 'Please enter a valid pickup window in the format HH:MM-HH:MM',
			),
			'required'   => array(
				'cart'    => array(
					'properties' => array(
						'prefers_collection' => array(
							'type'  => 'boolean',
							'const' => true,
						),
					),
				),
			),
		),
	),
);

	array(
		'id'       => 'plugin-namespace/confirm-pickup-window-2',
		'label'    => 'Confirm Pickup Window',
		'location' => 'address',
		'type'     => 'text',
		'rules'    => array(
			'validation' => array(
				'type'         => 'string',
				'const'        => array( '$data' => '1/plugin-namespace~1pickup-window-2' ),
				'errorMessage' => 'Please confirm your pickup window',
			),
			'required'   => array(
				'type'       => 'object',
				'properties' => array(
					'anyOf' => array(
						'cart'    => array(
							'properties' => array(
								'prefers_collection' => array(
									'type'  => 'boolean',
									'const' => true,
								),
							),
						),
						'address' => array(
							'properties' => array(
								'pickup_window' => array(
									'type' => 'string',
									'not'  => array(
										'const' => '',
									),
								),
							),
						),
					),
				),
			),
		),
	),
);
}, 10, 1 );
```


https://github.com/user-attachments/assets/fe178964-9c63-47dd-bbfe-66a9dc67abf5

mikejolley:
@senadir I think we're good here. I made some fix commits. Should we get someone else to take a look? This should merge before https://github.com/woocommerce/woocommerce/pull/55879

mikejolley (approved):

plugins/woocommerce/src/Blocks/Utils/CartCheckoutUtils.php
Tofandel: This is a breaking change. This is a public function and changing the schema of it's return is a big no no. And of course this is not under the feature flag

mikejolley: @Tofandel This populates data for our client. The actual country class, which 3rd parties would be using, remains unchanged. Are you using this or just pointing it out?

PR #55263, opened 2025-02-08
Add defensive check after filter for cart items

### LINKED ISSUE #55103: [Fatal] PHP Fatal error:  Uncaught Error: Call to a member function get_tax_class() on null in | class-wc-cart-totals.php:232
### Prerequisites

### Describe the bug

witnessed in `9.6.0`

[link to code](https://github.com/woocommerce/woocommerce/blob/6f5032a0277a0eb006a442407d9d09c0c370b8be/plugins/woocommerce/includes/class-wc-cart-totals.php#L232)

```
PHP Fatal error:  Uncaught Error: Call to a member function get_tax_class() on null in /../wp-content/plugins/woocommerce/includes/class-wc-cart-totals.php:232
Stack trace:
#0 /../wp-content/plugins/woocommerce/includes/class-wc-cart-totals.php(658): WC_Cart_Totals->get_items_from_cart()
#1 /../wp-content/plugins/woocommerce/includes/class-wc-cart-totals.php(150): WC_Cart_Totals->calculate_item_totals()
#2 /../wp-content/plugins/woocommerce/includes/class-wc-cart-totals.php(141): WC_Cart_Totals->calculate()
#3 /../wp-content/plugins/woocommerce/includes/class-wc-cart.php(1409): WC_Cart_Totals->__construct(Object(WC_Cart))
#4 /../wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-button/src/Endpoint/SimulateCartEndpoint.php(87): WC_Cart->calculate_totals()
#5 /../wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-button/src/Endpoint/AbstractCartEndpoint.php(73): WooCommerce\PayPalCommerce\Button\Endpoint\SimulateCartEndpoint->handle_data()
#6 /../wp-content/plugins/woocommerce-paypal-payments/modules/ppcp-button/src/ButtonModule.php(140): WooCommerce\PayPalCommerce\Button\Endpoint\AbstractCartEndpoint->handle_request()
#7 /../core/6.7.1/wp-includes/class-wp-hook.php(324): WooCommerce\PayPalCommerce\Button\ButtonModule::WooCommerce\PayPalCommerce\Button\{closure}('')
#8 /../core/6.7.1/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters('', Array)
#9 /../core/6.7.1/wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#10 /../wp-content/plugins/woocommerce/includes/class-wc-ajax.php(96): do_action('wc_ajax_ppc-sim...')
#11 /../core/6.7.1/wp-includes/class-wp-hook.php(324): WC_AJAX::do_wc_ajax('')
#12 /../core/6.7.1/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters(false, Array)
#13 /../core/6.7.1/wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#14 /../core/6.7.1/wp-includes/template-loader.php(13): do_action('template_redire...')
#15 /../core/6.7.1/wp-blog-header.php(19): require_once('/../core...')
#16 /../core/6.7.1/index.php(17): require('/../core...')
#17 {main}
  thrown in /../wp-content/plugins/woocommerce/includes/class-wc-cart-totals.php on line 232
``` 

### Expected behavior

no fatals witnessed

### Actual behavior

fatals were witnessed

### Steps to reproduce

We don't have steps to reproduce, but observed 134 instances of this problem on Fri Jan 31, 2025 across 8 sites. Hopefully, someone can dig in and apply a defensive check if appropriate. - To be updated after investigation.

### WordPress Environment

N/A

### Isolating the problem

coreymckrill:

If an extension uses the `woocommerce_get_cart_item_from_session` filter and modifies the array of cart item data, it can cause a fatal error if the modified data does not include the specific `data` key containing an instance of the cart item's corresponding product object. This simply ensures that even if an extension is using that filter incorrectly, it will not cause a fatal error.

Fixes #55103

1. Before checking out this branch, visit the shop page on your test site. Add a product to the cart, and then go to the Cart page.
2. Add this test snippet to your site:
    ```php
	add_filter(
		'woocommerce_get_cart_item_from_session',
		function ( $session_data, $values ) {
			return $values;
		},
		10,
		2
	);
    ```
3. Refresh the cart page. Now you should see the fatal error reported in #55103.
4. Now check out this branch. Refresh the cart page again. The error should be gone, and the cart page should load successfully, and show that your cart has the product that you originally added.

### DIFF
--- a/plugins/woocommerce/includes/class-wc-cart-session.php
+++ b/plugins/woocommerce/includes/class-wc-cart-session.php
@@ -218,7 +218,14 @@ final class WC_Cart_Session {
 				 */
 				$cart_contents[ $key ] = apply_filters( 'woocommerce_get_cart_item_from_session', $session_data, $values, $key );
 				if ( ! isset( $cart_contents[ $key ]['data'] ) || ! $cart_contents[ $key ]['data'] instanceof WC_Product ) {
-					// If the cart contents is missing the product object after filtering, add it back in.
+					// If the cart contents is missing the product object after filtering, something is wrong.
+					wc_doing_it_wrong(
+						__METHOD__,
+						'When filtering cart items with woocommerce_get_cart_item_from_session, each item must have a data key containing a product object.',
+						'9.8.0'
+					);
+
+					// Add the product back in.
 					$cart_contents[ $key ]['data'] = $product;
 				}
 
--- a/plugins/woocommerce/includes/class-wc-cart-session.php
+++ b/plugins/woocommerce/includes/class-wc-cart-session.php
@@ -207,7 +207,20 @@ final class WC_Cart_Session {
 					)
 				);
 
+				/**
+				 * Filter to modify or add session data to the cart contents.
+				 *
+				 * @since 3.2.0
+				 *
+				 * @param array  $session_data Data for an item in the cart.
+				 * @param array  $values       Data for an item in the cart, without the product object.
+				 * @param string $key          The cart item hash.
+				 */
 				$cart_contents[ $key ] = apply_filters( 'woocommerce_get_cart_item_from_session', $session_data, $values, $key );
+				if ( ! isset( $cart_contents[ $key ]['data'] ) || ! $cart_contents[ $key ]['data'] instanceof WC_Product ) {
+					// If the cart contents is missing the product object after filtering, add it back in.
+					$cart_contents[ $key ]['data'] = $product;
+				}
 
 				// Add to cart right away so the product is visible in woocommerce_get_cart_item_from_session hook.
 				$this->cart->set_cart_contents( $cart_contents );
--- a/plugins/woocommerce/changelog/fix-55103-null-cart-item
+++ b/plugins/woocommerce/changelog/fix-55103-null-cart-item
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Add defensive check to ensure the product object exists in items in the cart session


naman03malhotra (commented):
Thank you for working on this. As expected, it works well as per the instructions. But I am thinking of cases where it can introduce unintended behavior, pardon me if I am just overthinking this.

However, I am brainstorming the cases where the fix can cause other problems, for example, a 3PD intentionally modifying or removing ['data'] key for valid reasons in that case, we are silently overriding the intended behavior.

I am wondering if by adding additional logging we can provide more information to the users for this behavior. Or we can find other way to catch it and throw a warning.

coreymckrill:
> However, I am brainstorming the cases where the fix can cause other problems, for example, a 3PD intentionally modifying or removing ['data'] key for valid reasons in that case, we are silently overriding the intended behavior.

@naman03malhotra I think this is a legitimate concern, and I thought about this too. What I discovered is that there are numerous places in the code, besides this one, where there is an assumption that the `['data']` key exists and that it is a product object. So a 3PD could have a legitimate reason to _replace_ the contents of `['data']` with their own object, but it would still have to extend the basic product object in order to have all the necessary methods, and in that case, it would not get overwritten by our defensive check. If `['data']` contains some value that is not like a product, or is missing completely, I am therefore assuming it's not intentional.

> I am wondering if by adding additional logging we can provide more information to the users for this behavior.

I'm not sure what other logging would be useful here, but I'm open to ideas.

> Or we can find other way to catch it and throw a warning.

We could try and add validation logic at every place in the code that assumes `['data']` is a product object, but I think it would be difficult to find them all and decide on a case-by-case basis what should happen if the object doesn't exist, and I think we would end up with the application generally being in an unexpected/unstable state.

One alternative to adding the missing product object back in would be to instead skip that item and not add it to the cart at all. I tested this as well, and it does avoid the fatal error, but it also means the cart has silently changed and will be in an unexpected state.

🤷 There probably isn't a perfect answer. What do you think?

plugins/woocommerce/includes/class-wc-cart-session.php
```diff
@@ -207,7 +207,20 @@ public function get_cart_from_session() {
...
+				 *
+				 * @param array  $session_data Data for an item in the cart.
+				 * @param array  $values       Data for an item in the cart, without the product object.
+				 * @param string $key          The cart item hash.
+				 */
 				$cart_contents[ $key ] = apply_filters( 'woocommerce_get_cart_item_from_session', $session_data, $values, $key );
+				if ( ! isset( $cart_contents[ $key ]['data'] ) || ! $cart_contents[ $key ]['data'] instanceof WC_Product ) {
+					// If the cart contents is missing the product object after filtering, add it back in.
+					$cart_contents[ $key ]['data'] = $product;
+				}
```
naman03malhotra: >I'm not sure what other logging would be useful here, but I'm open to ideas.

I mean, if we can add log a warning something like,

`Seems like a hook on "woocommerce_get_cart_item_from_session" is changing the type of data to something other than `WC_Product`, we have added the product back to the cart content....`

Something on these lines, to make folks aware.

This is something from top of my head, maybe I am being extra cautious.

naman03malhotra (approved):
>So a 3PD could have a legitimate reason to replace the contents of ['data'] with their own object, but it would still have to extend the basic product object in order to have all the necessary methods, and in that case, it would not get overwritten by our defensive check. If ['data'] contains some value that is not like a product, or is missing completely, I am therefore assuming it's not intentional.

Makes sense, I agree that adding the $product back to `data` is better experience that seeing a fatal since, WC_Product type is expected anyway.

plugins/woocommerce/includes/class-wc-cart-session.php
coreymckrill: Fair enough, added a `doing_it_wrong` in 761f4f53a0. That way it will go to the error log, but only if they have WP_DEBUG defined, so it shouldn't get noisy in production environments. Wdyt?

naman03malhotra: Thanks for adding it, seems like an apt middle ground. If we get complaints, we can iterate on this, but what its worth, it is much better than a fatal. Thanks for your work on this.

PR #55295, opened 2025-02-10
iAPI migration: Prevent quantity change notice showing on mini cart when adding items with the new iAPI sync event

luisherranz:

This pull request prevents a snackbar (which indicates if the quantity of a product has changed when the minicart block is opened automatically) from showing when clicking on the "Add to cart" button.

This problem, which was solved in PR #55023, has reappeared when moving the synchronization between the Store of the Interactivity API and the Store of `@wordpress/data` to a single event in #54979.

**Always use the Interactivity API to add products to the cart.**

1. Go to the options of the minicart block and activate the option "Open drawer when adding"
2. Add this PHP snippet somewhere in your site, and update the product IDs to match the IDs of some of your products.
   ```php
   add_action('woocommerce_before_calculate_totals', 'sync_cart_quantities', 20, 1);

   function sync_cart_quantities(WC_Cart $cart) {
	$product_id = 26;
	$other_product_id = 28;
	
	$quantity_of_first = 0;
	
	foreach ( $cart->get_cart_contents() as $cart_item_key => $cart_item) {
		if ($cart_item['product_id'] == $product_id) {
			$quantity_of_first = $cart_item['quantity'];
		}
	}

	foreach ( $cart->get_cart_contents() as $cart_item_key => $cart_item) {
		if ($cart_item['product_id'] == $other_product_id) {
			$cart->set_quantity($cart_item['key'], $quantity_of_first * 2, false);
		}
	}
	return $cart;
   }
   ```
4. In the frontend, add an unrelated product to the cart so that the quantity is two or more. 
5. The mini cart block will open, and the snack bar saying that each quantity has changed should not be shown.
6. Now add the product with `$product_id` to the cart, and then the `$other_product_id` to the cart. No snack bar should be shown.
7. Now add the `$product_id` to the cart again. A snack bar saying "The quantity of (the product with `$other_product_id`) was changed to 4." should appear.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


luisherranz:
After understanding the problem a little better thanks to @opr, I have worked on a new solution. I have also updated the instructions to test the pull request.

luisherranz:
Oh, I've also added some extra todos for some things that we have to change before finalizing the migration.

plugins/woocommerce-blocks/assets/js/base/stores/cart-items.ts
```diff
@@ -50,152 +56,158 @@ function generateError( json: StoreAPIError ) {
...
+			// Optimistically updates the number of items in the cart.
+			if ( itemIndex !== -1 ) {
+				state.cart.items[ itemIndex ].quantity = quantity;
+				key = state.cart.items[ itemIndex ].key || null;
+				if ( key ) quantityChanges.cartItemsPendingQuantity = [ key ];
+			} else {
+				state.cart.items.push( { id, quantity } );
+				itemIndex = state.cart.items.length - 1;
+				quantityChanges.productsPendingAdd = [ id ];
+			}
```
opr: Nice, I really like this approach 👌🏼

plugins/woocommerce-blocks/assets/js/data/cart/utils.ts
```diff
@@ -142,10 +142,10 @@ export const setIsCustomerDataDirty = debounce(
 );
 
 /**
- * Controls whether to ignore sync events.
+ * Sets whether it is syncing the stores or not.
```
opr: I think we could describe better what is happening here. I am lucky to have some context from pairing with Sam, but if I was coming to this fresh I would have no idea what this is for.

opr (changes requested):
Thanks @luisherranz, this is working great for me following your changes.

A couple of thoughts that aren't necessarily blockers, but I just wanted to hear what you think:

- How about removing all sync related logic from `receiveCart` and making a new action called `syncCart` where `isSyncingStores` would be implicitly true. I think the benefit here would be that reasoning with this code would be slightly easier. Right now there are several branches based on the value of `isSyncingStores`. The drawbacks is you have two functions that do basically the same thing. I don't feel super strongly about it. Not a blocker.
- Naming of `isSyncingStores` is possibly going to be confusing, consider `isSyncingFromInteractivityApi` or similar.

By the way another test you can add to the snippet, inside the 

```php
	if ($cart_item['product_id'] == $other_product_id) {
 		$cart->set_quantity($cart_item['key'], $quantity_of_first * 2, false);
 	}
```

Change it to

```php
	if ($cart_item['product_id'] == $other_product_id) {
 		$cart->set_quantity($cart_item['key'], $quantity_of_first * 2, false);
		
		// Remove the second item if the first item's quantity gets to 10. Expect to see a notice about item B removed from the cart.
		if ( $quantity_of_first === 10 ) {
				$cart->remove_cart_item( $cart_item['key'] );
		}
 	}
```

Then change item A to 10, and ensure you see a notice informing you that item B was removed from the cart.

I'm going to submit this as requested changes due to the docs/naming updates I mentioned, but in principle the approach is good, I'd be happy to approve once they're taken care of.

plugins/woocommerce-blocks/assets/js/data/cart/utils.ts
luisherranz: Done in https://github.com/woocommerce/woocommerce/pull/55295/commits/cc7647b214661c7c445588a6757d64628578536d.

luisherranz:
> Then change item A to 10, and ensure you see a notice informing you that item B was removed from the cart.

Ok, removing works fine too, thanks.

> - How about removing all sync related logic from `receiveCart` and making a new action called `syncCart` where `isSyncingStores` would be implicitly true. I think the benefit here would be that reasoning with this code would be slightly easier. Right now there are several branches based on the value of `isSyncingStores`. The drawbacks is you have two functions that do basically the same thing. I don't feel super strongly about it. Not a blocker.
> - Naming of `isSyncingStores` is possibly going to be confusing, consider `isSyncingFromInteractivityApi` or similar.

Sure, it makes sense. I made both changes in https://github.com/woocommerce/woocommerce/pull/55295/commits/cc7647b214661c7c445588a6757d64628578536d.

luisherranz:
> How about removing all sync related logic from receiveCart and making a new action called syncCart where isSyncingStores would be implicitly true. I think the benefit here would be that reasoning with this code would be slightly easier. Right now there are several branches based on the value of isSyncingStores. The drawbacks is you have two functions that do basically the same thing. I don't feel super strongly about it. Not a blocker.

My bad, I forgot about [this change](https://github.com/woocommerce/woocommerce/blob/b4c83764a6bbc0a923b66d77d0fc57973bb8e24d/plugins/woocommerce-blocks/assets/js/data/cart/resolvers.ts#L44) from [the previous pull request](https://github.com/woocommerce/woocommerce/pull/54979) where I prevent the initial fetch from synchronizing.

We can change the name of the action `syncCartWithIAPIStore` to something more generic and also use it on the `getCartData` resolver, or I can revert the logic and leave `receiveCart` as it was, whatever you say. I don't have a strong opinion on either option.

opr (approved):
This is working for me now following the changes, I tested with the snippets I provided. I don't see notices when adding the same item to the cart, but I do see them when expected in the snippet.

opr:
> We can change the name of the action syncCartWithIAPIStore to something more generic and also use it on the getCartData resolver, or I can revert the logic and leave receiveCart as it was, whatever you say. I don't have a strong opinion on either option.

I think the way it is now is OK actually, `recieveCart` remains focused to updating the data store with a cart object, and `syncCartWithIAPIStore` while similar, is also focused. Avoiding merging these methods is a good choice imo.

luisherranz:
Ok, I have gone back and forth a couple of times to see how the final code looked, and I think this is the cleanest solution. I have also tried to find better names for the variables and functions so that it is clearer what is happening.

@opr, tell me what you think.

plugins/woocommerce-blocks/assets/js/data/cart/index.ts
```diff
@@ -65,68 +66,39 @@ window.addEventListener( 'load', () => {
...
-			diffObjects( val1, val2, currentPath ); // Recurse for nested objects
-		}
-	}
-}
-
-let previousCart: object | null = null;
-let id = 0;
-
 // Emmits event to sync iAPI store.
+let previousCart: object | null = null;
```
opr, line 70: Don't know if this is due to your changes, but this event is dispatched on all changes in Cart/Checkout block too, is that expected/intended? An infinite loop is prevented by the `getTriggerStoreSyncEvent` flag which is great.

plugins/woocommerce-blocks/assets/js/data/cart/thunks.ts
```diff
@@ -143,13 +146,19 @@ export const applyExtensionCartUpdate =
...
+ * Fetch the cart but avoid triggering the event that syncs with the
+ * Interactivity API store to avoid infinite loops.
  *
+ * @param {QuantityChanges} quantityChanges The quantity changes data included in the sync event.
  * @throws Will throw an error if there is an API problem.
  */
-export const syncCartWithIAPIStore =
-	() =>
-	async ( { dispatch }: CartThunkArgs ) => {
+export const syncCartwithIAPIStore =
```
opr: nitpick on casing here
```suggestion
export const syncCartWithIAPIStore =
```

plugins/woocommerce-blocks/assets/js/data/cart/index.ts
```diff
@@ -65,68 +66,39 @@ window.addEventListener( 'load', () => {
...
+window.addEventListener(
+	'woocommerce-store-sync-required',
+	( event: Event ) => {
+		const customEvent = event as CustomEvent< {
+			type: string;
+			quantityChanges: QuantityChanges;
+		} >;
+		const { type, quantityChanges } = customEvent.detail;
+		if ( type === 'from_iAPI' ) {
+			wpDispatch( store ).syncCartwithIAPIStore( quantityChanges );
```
opr: Nitpick on casing here:
```suggestion
			wpDispatch( store ).syncCartWithIAPIStore( quantityChanges );
```

opr (approved):
Couple of nit picks, but going through the flow in my IDE seems reasonable, it's way easier to follow and reason with now, thanks @luisherranz

plugins/woocommerce-blocks/assets/js/data/cart/index.ts
luisherranz: It is, in some way, expected, as any change in the cart's state should trigger an update of the Interactivity API Store. Perhaps there is still no block that uses the Interactivity API that is designed to be displayed on the cart or checkout page, but that doesn't mean that they won't exist.

opr: Thought so :) nice future-thinking

samueljseay:
I have fixed up the e2e tests and I am happy with the approach here. Opting to leave the todos in place and we can look to address them later.

PR #55303, opened 2025-02-10
Fix for additional features not being registered in WP CLI

Konamiman:

This pull request fixes an issue in the features engine that has surfaced after https://github.com/woocommerce/woocommerce/pull/55118 was merged and was discovered while testing https://github.com/woocommerce/woocommerce/pull/54399. Here's what was happening:

1. In `class-wc-cli.php` the following line is executed: `FeaturesUtil::feature_is_enabled`. 
    * This causes `FeaturesController::get_feature_definitions` to run.
    * Since `$this->features` is empty, it gets initialized. Some feature definitions are hardcoded, but others are registered via the `woocommerce_register_feature_definitions` hook. The resulting list of features is cached.
 2. Then the `WooCommerce` class gets instantiated and as part of its construction, the appropriate classes subscribe to the `woocommerce_register_feature_definitions` hook.
    * However at that point it's too late for that: `FeaturesController` has already cached the feature definitions, and it won't fire the `woocommerce_register_feature_definitions` hook anymore.
    * Result: the features that depend on the hook don't get registered.

A symptom of this, for example, is the COGS functions not working anymore from CLI, e.g. invoking `get_cogs_value()` on a product throws a "doing it wrong" even if the feature is enabled. For https://github.com/woocommerce/woocommerce/pull/55118 this wasn't apparent because the blueprint feature is hardcoded in the `FeaturesController` class (it doesn't depend on the hook).

The fix consists of deferring the registration of additional features to the `init` hook, right before `before_woocommerce_init` is fired (because during the execution of `before_woocommerce_init` plugins are supposed to execute `declare_compatibility`, and by then all the feature definitions must be already available) and after all the relevant objects have been instantiated. Exception: if `init` has been fired already (this will happen when running unit tests), then the registration of additional features happens immediately, as it was the case before.

Also the class comment in `FeaturesController` has been extended to make clear that features declared via the hook won't be available before `before_woocommerce_init`.

Run the following in the command line:

```
wp eval 'var_dump(array_keys(Automattic\WooCommerce\Utilities\FeaturesUtil::get_features(true)));'
```

A list of features will be printed: with the fix this will include `custom_order_tables` and `cost_of_goods_sold`; without the fix these two will be missing.

Also verify that the COGS programmatic API works again (you need to enable the Cost of Goods Sold feature  in _WooCommerce - Settings - Advanced - Features_), for example:

```php
$p = wc_get_product(<existing product id>);
$p->set_cogs_value(1234);
$p->save();
echo wc_get_product(<existing product id>)->get_cogs_value(); //1234
```

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


albarin (approved):
Thanks @Konamiman, looks good, checked that `custom_order_tables` and `cost_of_goods_sold` are listed after the fix ✅

coreymckrill:
@Konamiman @albarin Something about this changeset made it so that WC Smooth Generator is now getting flagged as "not compatible" with HPOS, even though it declares support for it...

Konamiman:
> Something about this changeset made it so that WC Smooth Generator is now getting flagged as "not compatible" with HPOS, even though it declares support for it...

For the record, this is fixed in https://github.com/woocommerce/woocommerce/pull/55544

Issue #47792, opened 2024-05-24
[Enhancement]: Improving Variable Product API for Adding to Cart


wavvves:
### Describe the solution you'd like

When we try to add a variable product to the cart using the API, there's a bit of confusion around how we name attributes. The API expects the "user-friendly" attribute name (what we see on the `<label>`), but sometimes this doesn't match up with the internal naming system we use in the product markup. This disconnect can cause problems, especially when we throw special characters into the mix when building a request to the StoreAPI on the browser (especially if we have to parse HTML to build our request).

Proposed solution:

1. **Support for multiple naming formats**: Let's make the API more flexible by accepting attribute names in various formats, like unencoded, encoded, and encoded with dashes (URL safe). This way, developers can work with the naming conventions that suit their stores best. This is what we already do for attribute values.

![image](https://github.com/woocommerce/woocommerce/assets/17236129/4d378fff-710d-488c-8680-8cb4e9d8b36c)

So we would be supporting in this case, for `variation[0][attribute]`:

- `color-%f0%9f%96%8d%ef%b8%8f`
-  `attribute_color-%f0%9f%96%8d%ef%b8%8f`
- `Color %f0%9f%96%8d%ef%b8%8f`
- `Color 🖍️`

2. **Priority on encoded with dashes**: To keep things simple and error-free, we can prioritize the encoded with dashes format when the API processes attribute names. It's easy to retrieve from DOM parsing and safer to pass along. Above is the suggested matching priority.

3. **Maintain backward compatibility**: We'll ensure that any changes we make won't disrupt existing stores. It's essential to keep everything running smoothly for our users.

### Additional context

Related GH discussion https://github.com/woocommerce/woocommerce/discussions/45844

PR #55317, opened 2025-02-11
update: Store API `cart/add-item` attribute slug support

frosso:

Related to https://github.com/woocommerce/woocommerce/issues/47792

I'd like to add support to the Store API to add a product via the same attribute names that are present on the product page.
The product page will rendered multiple `select` elements with `name` attributes that are not supported by the Store API.
With these changes, there will be a 1:1 match between the attribute names used on the product page and the ones supported by the Store API.
This ensures also support for attribute names that have special characters and might get encoded.

- Create a (or edit an existing) variable product (e.g.: T-Shift; Attributes: "Color ✏️", "Size")
  - Ensure one of the attribute names contains a special character - e.g.: "Color ✏️")
  - Just for fun, ensure one of the attribute values also contains a special character (e.g.: "Blu~Èℹ️")
  - Ensure the attribute is "Used for variations"
  - Set the prices for the product variations, so that they're selectable
- Save the product
- Take note of the product ID
- Navigate to the product page
- Change the selected product attributes
- You can now make this Store API call to use the product attributes parsed from the `select` elements on product page:

```js
await wp.apiFetch({
    method: 'POST',
    path: '/wc/store/v1/cart/add-item',
    data: {
        variation: Array.from(document.querySelectorAll('.variations_form .variations select')).map(({ name, value }) => ( {
            attribute: name,
            value,
        })),
        id: parseInt(document.querySelector('.single_variation_wrap input[name="product_id"]').value, 10),
        quantity: 1,
    },
})
```

For this product selection:
Screenshot 2025-02-11 at 11 05 51 AM

This payload can now be sent to `POST /wc/store/v1/cart/add-item`:
```json
{
  "variation": [
    {
      "attribute": "attribute_size%f0%9f%98%86",
      "value": "XL?~Èℹ️"
    },
    {
      "attribute": "attribute_color",
      "value": "Blue"
    }
  ],
  "id": {product_id},
  "quantity": 1
}
```

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message update: Store API `POST /cart/add-item` to support attribute names from product page

</details>

<details>

#### Comment 

</details>


frosso:
ope - that's something I forgot to do before opening the issue. I'll add some unit tests.

plugins/woocommerce/src/StoreApi/docs/cart.md
```diff
@@ -426,7 +426,10 @@ When adding variations to the cart, the naming of the attribute is important.
 
 For global attributes, the attribute posted to the API should be the slug of the attribute. This should have a `pa_` prefix. For example, if you have an attribute named `Color`, the slug will be `pa_color`.
 
-For product specific attributes, the attribute posted to the API should be the name of the attribute. For example, if you have an attribute named `Size`, the name will be `Size`. This is case-sensitive.
+For product specific attributes, the attribute posted to the API can be one of the following:
+
+- the name of the attribute. For example, if you have an attribute named `Size`, the name will be `Size`. This is case-sensitive.
+- the slug of the attribute. For example, if you have an attribute named `Authograph ✏️`, the name will be `attribute_autograph-%e2%9c%8f%ef%b8%8f`. This is case-sensitive. You can get this slug from the related `select` on the product page.
```
FangedParakeet:
```suggestion
- the slug of the attribute. For example, if you have an attribute named `Autograph ✏️`, the name will be `attribute_autograph-%e2%9c%8f%ef%b8%8f`. This is case-sensitive. You can get this slug from the related `select` on the product page.
```
🔍

FangedParakeet (approved):
Code logic makes sense to me and the testing instructions worked as expected. There is just one tiny typo in the cart README, but I'm going to approve this anyways. Please just fix the typo and then you have my blessings to proceed. 🙏

plugins/woocommerce/src/StoreApi/docs/cart.md
frosso: Whoopsies, thank you! 🦅 👁️

plugins/woocommerce/src/StoreApi/Utilities/CartController.php
```diff
@@ -1378,6 +1378,21 @@ protected function sanitize_variation_data( $variation_data, $variable_product_a
 			$lowercase_attribute_label = strtolower( $attribute_label );
```
opr: I suppose for a small perf gain, this could be moved below `if` added in this PR. Makes sense to define things closer to where they're used in my opinion.

frosso: @opr that's fair - I think I was trying to modify the least number of lines :D But I agree with you! Changed in c17b4a2ae4

opr (approved):
Thanks @frosso the changes look good, and thanks a lot for writing tests and updating docs 🤩

One question: Is there any value to support to passing the "unencoded" slug?

Like if I pass

```json
{
  "variation": [
    {
      "attribute": "attribute_pa_sport-⚾️-棒球",
      "value": "Baseball ⚾️"
    },
    {
      "attribute": "attribute_length",
      "value": "37"
    },
    {
      "attribute": "attribute_wood",
      "value": "Oak"
    }
  ],
  "id": 1789,
  "quantity": 1
}
```

should this work, or are we only willing to support adding via the escaped slug? You mention getting the slug from the select in the docs, but that is different to the slug we display and allow copying in the `Edit Attribute` page.

image

Just curious what you think, I don't think that should be a blocker to this work, but if you do look into that re-request a review.

Thanks!

frosso:
> One question: Is there any value to support to passing the "unencoded" slug?

@opr that's a good question! After reading your comment, I started exploring that approach and even began implementing it. As I did, I also considered whether we should support unencoded attribute _values_ as well, to maintain consistency
But then the changes started being a bit more complex.

So I took a step back. With the current approach in this PR, we're already supporting the same attribute names and values that appear on the product page, as well as the values that would be passed when adding a product to the cart via the "add to cart" form.

I feel like that's a sufficient stopping point - with these changes, there's 1:1 match (and more!) between the "legacy" add-to-cart form (something that wasn't present before).

Given that, I'd be inclined to leave it as-is if that works for you :) 

Thanks again for your time!

opr:
@frosso yeah that's fine with me, I think if we get reports about this not working as expected (developers using the unencoded slug), we could point to the docs that show the encoded one should be used. Works for me and happy for you to merge when the CI is ✅

PR #55318, opened 2025-02-11
[e2e tests] Enable fullyParallel mode

adimoldovan:

Enabled `fullyParallel` mode for better balance of shards execution. This makes tests in a spec file run in parallel (which we're not using), enabling Playwright to split tests in shards instead of spec files. The resulting shards will have a more even number of tests, with better balance in terms of execution time.

This is recommended in the [Playwright docs](https://playwright.dev/docs/test-sharding#balancing-shards). 

I chose to enable this at project level rather than global level, for better control. API tests for instance failed with `fullyParallel` enabled.
I did not touch the other environments yet, we'll re-evaluate this down the line after monitoring the results.

Check CI, both Core e2e and Blocks e2e shards. All tests should pass.


Stojdza (approved):
I really like this approach! 🚀

PR #55251, opened 2025-02-07
[dev] Bump Playwright to 1.50.1

adimoldovan:
### Changes proposed in this Pull Request:

Bump Playwright to 1.50.1 and some Allure dependencies to their latest.

### How to test the changes in this Pull Request:

CI should be green.

PR #55319, opened 2025-02-11
[e2e tests] Fix allure-playwright output path

adimoldovan:

[55251](https://github.com/woocommerce/woocommerce/pull/55251) bumped allure-playwright to the latest version and this introduced a breaking change that was missed. The output directory field was renamed from `outputFolder` to `resultsDir` and the results were not generated in the expected path but in the default one.

Other than the path fix I removed the `ALLURE_RESULTS_DIR` env variable which is never used.

CI green.
Run any test. Check that `test-results/allure-result` folder is generated and it contains valid results.


Stojdza (approved):
`test-results/allure-result` generated as expected 👍

PR #55330, opened 2025-02-11
Migrate Product Collection from WC to WP iAPI

kmanijak:

Continue Product Collection migration from WC iAPI to WP iAPI.

Changes include:
- Replace `wc-navigation-id` with `wp-router-region`
- Adjust the frontend logic to it
- Remove loading bar animation as it's coming natively from iAPI router.

Testing covered:
- client-side navigation
    - one PC block on a page
    - multiple PC blocks on a page
- disabling client-side navigation when 3rd party block is added
- prefetching on hover
- JS events sending (PC rendering and viewing product)
- animations (loading bar, scroll to the first row of products)

> [!IMPORTANT]  
> It's OK CI fails on this PR with the following problems:
> - Lint error related to Cart -> Cart is being handled separately
> - E2E tests failing related to Cart -> Cart is being fixed spearately
> - E2E test failing related to Mini Cart: `Shopper → Notices -> Shopper can add item to cart, and will not see a notice in the mini cart` - It's being fixed in separate PR: https://github.com/woocommerce/woocommerce/pull/55295
>
> This PR aims another feature branch where we collect effort migrating iAPI and not `trunk`!

1. Smoke test Product Collection in terms of iAPI usage:
- Client-side navigation
    - one PC block on a page
    - multiple PC blocks on a page
- disabling client-side navigation when 3rd party block is added
- prefetching on hover
- JS events sending (PC rendering and viewing product)
- animations (loading bar, scroll to the first row of products)

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce-blocks/assets/js/blocks/product-collection/frontend.tsx
```diff
@@ -81,53 +93,32 @@ function scrollToFirstProductIfNotVisible( wcNavigationId?: string ) {
...
-				ref?.closest( '[data-wp-router-region]' ) as HTMLDivElement
-			 )?.dataset.wcNavigationDisabled;
-
-			if ( isDisabled ) {
+
+			if ( ! ref ) {
+				return;
+			}
+
+			const ctx = getContext< ProductCollectionStoreContext >();
```
tjcafferkey: Nitpick: We could move this into the `if()` condition so we don't execute unnecessarily

plugins/woocommerce-blocks/assets/js/blocks/product-collection/frontend.tsx
```diff
@@ -81,53 +93,32 @@ function scrollToFirstProductIfNotVisible( wcNavigationId?: string ) {
...
+
+			const routerRegionId = getRouterRegionId( ref );
```
tjcafferkey: Nitpick: Same with this, we could move this into the if() condition so we don't execute unnecessarily

tjcafferkey (approved):
These changes look good, just left a few comments.

One UX issue I've spotted is pagination in general when using this block on a page:

1. Add Product Collection block and select a collection with around 5 or more products
2. Decrease "Products Per Page" option to 2, increase "Max Pages to Show" to more than 1
3. Save and go to frontend.
4. There is no pagination, I have to specifically add the pagination block (which I may not even know exists) to get this to work with the above options.

Are we already aware of this?

kmanijak:
> One UX issue I've spotted is pagination in general when using this block on a page:
>
> Add Product Collection block and select a collection with around 5 or more products
> Decrease "Products Per Page" option to 2, increase "Max Pages to Show" to more than 1
> Save and go to frontend.
> There is no pagination, I have to specifically add the pagination block (which I may not even know exists) to get this to work with the above options.
> Are we already aware of this?

@tjcafferkey, it depends on which collection you choose:
- if "create your own", the Pagination is included by default. The assumption is this would show a browsable catalog in some form (first block on a screenshot)
- if any other collection like "On Sale Products" or "New Arrivals" the default settings are: show 1 page only, no pagination, 5 products. The assumption is this is used for promo purposes or as a banner and is not browsable by default (second block on a screenshot)

image

Does it make sense?

tjcafferkey:
I suppose this is unrelated to this PR so feel free to merge but FWIW it makes sense, I can see the thinking there. My thinking is that it's easier to remove a block you don't want than it is to add a block you can't find (they may not know it exists/know its name)

plugins/woocommerce-blocks/assets/js/blocks/product-collection/frontend.tsx
kmanijak: https://github.com/woocommerce/woocommerce/pull/55330/commits/e3cd2611fe0157a48ca150821024bd8145b4778d

plugins/woocommerce-blocks/assets/js/blocks/product-collection/frontend.tsx
kmanijak: https://github.com/woocommerce/woocommerce/pull/55330/commits/e3cd2611fe0157a48ca150821024bd8145b4778d

kmanijak:
@luisherranz pointed out iAPI router handles accessibility messages natively so I removed this from Product Collection because the announcements were duplicated https://github.com/woocommerce/woocommerce/pull/55330/commits/4dcfe34d39e785b9b3507011cd0243adcf909277.

Also @tjcafferkey, Luis said he'll make a review of this PR on Monday so I'm not merging it yet. Do you mind taking it over to address Luis review if needed and merging otherwise?

tjcafferkey:
>Do you mind taking it over to address Luis review if needed and merging otherwise?

Sure thing, thats fine with me!

plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/Renderer.php
```diff
@@ -249,7 +249,7 @@ private function enable_client_side_navigation( $block_content ) {
 		// Add `data-wp-navigation-id to the product collection block.
```
luisherranz:
```suggestion
		// Add `data-wp-router-region` to the product collection block.
```

plugins/woocommerce-blocks/assets/js/blocks/product-collection/frontend.tsx
```diff
@@ -50,14 +59,14 @@ const forcePageReload = ( href: string ) => {
  * Ensures the visibility of the first product in the collection.
  * Scrolls the page to the first product if it's not in the viewport.
  *
- * @param {string} wcNavigationId Unique ID for each Product Collection block on page/post.
+ * @param {string} wpRouterRegionId Unique ID for each Product Collection block on page/post.
  */
-function scrollToFirstProductIfNotVisible( wcNavigationId?: string ) {
-	if ( ! wcNavigationId ) {
+function scrollToFirstProductIfNotVisible( wpRouterRegionId?: string ) {
```
luisherranz: We do this in the Query block by simply moving the focus to the first element in the list, which is important for accessibility. When moving the focus, browsers also scroll if the element is not on the viewport.

https://github.com/WordPress/gutenberg/blob/7fbba99059313dbf00e43a5e0c2cf62885852f79/packages/block-library/src/query/view.js#L41-L43

Even if you want to keep the manual scroll, the change of focus for accessibility should be added. The problem is that the two scrolls can conflict.

plugins/woocommerce-blocks/assets/js/blocks/product-collection/frontend.tsx
```diff
@@ -17,13 +17,22 @@ export type ProductCollectionStoreContext = {
...
+
+const getRouterRegionId = ( ref: HTMLElement ) => {
+	const routerRegionElement = getRouterRegion( ref );
+	return routerRegionElement?.dataset?.wpRouterRegion;
+};
+
+const getNavDisabled = ( ref: HTMLElement ) => {
+	const routerRegionElement = getRouterRegion( ref );
+	return routerRegionElement?.dataset?.wpNavigationDisabled;
+};
```
luisherranz: Instead of using this method, we are now using the property `clientNavigationDisabled` of the `core/router` config. 

`wp_interactivity_config( 'core/router', array( 'clientNavigationDisabled' => true ) );`

https://github.com/WordPress/gutenberg/blob/7fbba99059313dbf00e43a5e0c2cf62885852f79/packages/block-library/src/query/index.php#L123

In the client, there is nothing to do. The router already looks at that property and, if it exists, does not use client-side navigation.

luisherranz (changes requested):
Looks good to me, although there are still a couple of things that can be improved.

plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/Renderer.php
tjcafferkey: Fixed in https://github.com/woocommerce/woocommerce/pull/55330/commits/dd24599bc375087db1e15e015d44c51b6caab5b9

plugins/woocommerce-blocks/assets/js/blocks/product-collection/frontend.tsx
tjcafferkey: Nice, this simplifies things quite a bit! https://github.com/woocommerce/woocommerce/pull/55330/commits/dd2193f06807ba0b4ac545dc32fd29e15ba8de73

plugins/woocommerce-blocks/assets/js/blocks/product-collection/frontend.tsx
tjcafferkey: I've implemented this fix in this commit https://github.com/woocommerce/woocommerce/pull/55330/commits/340ba477b36938ceee3e4877190f908abb1ee057

I'd appreciate it if you could double check this one to ensure I've understood correctly.

luisherranz:
Thanks for the changes, Tom. I'm taking a deeper look at how everything is handled in the Product Collection and I'd like to make some adjustments before this is merged. I'll push once I've finished.

luisherranz (approved):
That's it. Feel free to review my changes and merge.

tjcafferkey:
Reviewed and tested your changes @luisherranz . All looks good to me, thank you! Im going to go ahead and merge.

PR #55344, opened 2025-02-11
Marketplace: show two rows in Discover page

raicem:
It depends on the number of items provided in the API response. If it's eight, we'll show the second row.

Add the second row to the Discover page group if the number of provided items are equal to 8. 

Closes https://github.com/Automattic/woocommerce.com/issues/23145.

1. Connect your test store to local WooCommerce.com instance. Please see PR https://github.com/Automattic/woocommerce.com/pull/23220 for instructions.
2. Checkout [this branch](https://github.com/Automattic/woocommerce.com/tree/raicem/increase-featured-api-endpoint-capacity) from WCCOM. This will allow you to send more items to the in-app discover page.
3. In WCCOM, add more featured products to a **Discover our favorites** group. Add as high as 8. You can try 7, or 6 as well to check if it's breaking the layout.
    - For example, if you make it 7, you'll get a 2x3 items on laptop displays, but on larger displays will show a 1x4. Because we don't have enough items to fill the second row.
5. If you don't have any other groups locally, also add 4 items to **Payments** and **Marketing**. 
6. Give [this URL](https://woocommerce.test/wp-json/wccom-extensions/3.0/featured?no_cache=true) a visit and confirm your new items are appearing there. This updates the cache on the WCCOM side. 
7. Make sure you have the Discover page cache disabled in your test store. I recommend adding this hook:
```php
add_filter('transient_wc_addons_featured', function ($value, $transient) {
	return false;
}, 10, 2);
```
8. Build assets of your local WooCommerce installation. I used `pnpm --filter='@woocommerce/plugin-woocommerce' build`
9. Confirm you can get both two row groups and single row groups in the Discover page. 
10. Change display sizes and confirm you don't get a row that's not fully filled in.
11. Checkout an older tag in the WooCommerce repo, for example 9.6.2 tag, and rebuild assets. This is to confirm bumping up the capacity in the API doesn't break older stores. 
   - so far, I haven't been able to rebuild 9.6.2, because pnpm version errors. Kyle already [explored](https://github.com/Automattic/woocommerce.com/pull/23220#issuecomment-2653579244) this, and confirmed it's working but I'll give it another go. 

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 
Show more product cards on Marketplace Discover page
</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/admin/client/marketplace/components/product-list-content/product-list-content.tsx
```diff
@@ -28,8 +28,14 @@ export default function ProductListContent( props: {
 } ): JSX.Element {
 	const wccomHelperSettings = getAdminSetting( 'wccomHelper', {} );
 
+	const rowsClassName =
+		props.products.length >= 8
```
kdevnel: I doubt this is worth worrying about, but what if we want to show 3 rows? Is it better to futureproof this to some extent? AI suggestion:

```tsx
const ITEMS_PER_ROW = 4;
const rowCount = Math.ceil( props.products.length / ITEMS_PER_ROW );
const rowsClassName = `woocommerce-marketplace__product-list-content--${
	rowCount === 1 ? 'single-row' : `${ rowCount }-rows`
}`;
```

kdevnel (commented):
I decided to do an early review even though this is a draft still. Hope that's okay! Overall happy with this approach and it tests okay with my WCCOM branch. Left a code comment (not a dealbreaker) and something else.

I notice on mobile 8 is very long! I wonder if we'd actually want two different experiences on mobile rather than having a long list. Again, maybe something to just consider and feedback for future work.

Not approving just yet until this is ready for a full review.

plugins/woocommerce/client/admin/client/marketplace/components/product-list-content/product-list-content.tsx
```diff
@@ -50,18 +46,60 @@ export default function ProductListContent( props: {
...
 	useEffect( () => {
 		updateColumns();
+
 		// Update columns on screen resize to adjust for responsive layout
 		window.addEventListener( 'resize', updateColumns );
 
 		return () => window.removeEventListener( 'resize', updateColumns );
 	}, [] );
 
+	/**
```
kdevnel, line 56: Thank you for these super helpful comments!

plugins/woocommerce/client/admin/client/marketplace/components/product-list-content/product-list-content.scss
kdevnel, line 1: Beautiful removal of lots of code ❤️

kdevnel (approved):
This tests and works perfectly. I personally agree, I think mobile should be limited to 4 per group at all times.

I initially thought this was touching the same code as my group headers PR, but it looks like they don't interact at all! That saves some merging/rebasing.

raicem:
Thank you for the review @kdevnel! I have updated the row count for smaller displays. I was also was able to test on on 9.6.2, using `npx pnpm@9.1.3 install` and `npx pnpm@9.1.3 build`. Confirming it's working as expected.

Issue #54867, opened 2025-01-27
[Settings] Enable Saving


psealock:
Create the ability to Save form data when merchants click "Save".

The proof of concept handled this by gathering all input values from the form and submitting the data to `/wp-json/wc-admin/settings` as form data.

https://github.com/woocommerce/woocommerce/pull/48340/files#diff-08711774f4a65ea5c2516f4bee385d36dc0ec3366f8a7e5e2c31d7cf0e7fe4f9R36

### Requirements
* Collect all data within the DataForm. Keep in mind this includes any inputs with ids added via slotFill or rendered custom HTML created by an extension.
* DataForms have a concept of [internal data](https://github.com/woocommerce/woocommerce/blob/6cb8c4d3fdea55b3379b427f4d9bae296ba15fb0/packages/js/settings-editor/src/hooks/use-settings-form.ts#L60). Decide if this is useful. What is the source of truth?
* After clicking Save, other sections and tabs should be aware of the changes. @chihsuan had the idea to reconstruct the data created by `WC_Settings_Page::add_settings_page_data` on success and repopulate `window.wcSettings.admin.settingsData`. 
* Keep in mind this will be only for Legacy screens. Modern screens may also use this system or create an entirely new one. What would you recommend?

PR #55352, opened 2025-02-12
[Settings] Prepare client side data for saving

psealock:

Related to https://github.com/woocommerce/woocommerce/issues/54867

This PR creates functions to prepare settings data before sending it to the server. Similarly to legacy screens, an `onSubmit` prop on the form calls a function to gather form data. Instead of sending a POST request, we will eventually create an AJAX request to persist data.

I decided to avoid using `DataForm` internals for keeping track of form data because extensions can very well add elements to the form outside of `DataForm` with something like this:

```php
public function render_custom_view() {
	?>
	<input type="text" value="Hello world" id="my-extension-input" name="my-extension-input"/>
	<?php
}
```

#### NOTE
Checkboxes, radios, and toggles aren't giving the right values at times because either the id/name isn't attached to the input correctly, or the actual HTML isn't updated (only internal state updates). I plan to address this in a follow up using `hidden` inputs.

2. Add/Enable Snippets and the following Snippet

```php


add_action( 'woocommerce_settings_catalog_options', function() {
	?>
		<input type="text" value="Hello world" name="my_extension_input" id="my_extension_input"/>
	<?php
} );
```
3. Visit legacy WooCommerce > Settings > Products and see the added input in Shop pages section.
4. In the Network tab, preserve logs and clear all requests. Then make a change and click "Save"
5. See the first network POST request and note the payload by taking a screenshot to compare later
6. Turn on the Settings feature flag and go back to WooCommerce > Settings > Products 
7. See the added input and click Save. See the logged out values in the console and compare to your earlier screenshot. They should be the same with the exception of `woocommerce_shop_page_id` which has not yet been implemented.
8. Poke around a few other pages to see results logged.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


packages/js/settings-editor/src/legacy/form.tsx
```diff
@@ -0,0 +1,109 @@
...
+				return;
+			}
+
+			const value = ( input as HTMLInputElement ).value;
+			// Need to type data as Record<string, string> to allow string indexing
+			( formData as Record< string, string > )[
+				( input as HTMLInputElement ).name ||
+					( input as HTMLInputElement ).id
+			] = value;
+		} );
```
rjchow:
```suggestion
	const formRef = useRef< HTMLFormElement >( null );

	const getFormData = () => {
		if ( ! formRef.current ) {
			return {};
		}
		const formElements = formRef.current.querySelectorAll<
			HTMLInputElement | HTMLSelectElement | HTMLTextAreaElement
		>( 'input, select, textarea' );

		const formData: Record< string, string > = {};

		formElements.forEach( ( input ) => {
			// Avoid generic Gutenberg input ids. This will require upstream fixes.
			if ( input.id.startsWith( 'inspector-' ) ) {
				return;
			}

			formData[ input.name || input.id ] = input.value;
		} );
```

rjchow: The type overrides looked a bit iffy to me and I had a good idea of how to streamline it - but I gave cursor a chance to prove itself and it did what I hoped to!

Here's the prompt I used:
```
streamline the types with less type overrides?
```

With these changes there's more type safety as we aren't overriding types, and it's also much more readable

rjchow (approved):
Tests well as described, made a comment on some TS tidying
I noticed something which isn't within the scope of this PR - In legacy settings when "Enable product reviews" is unchecked the checkboxes below are hidden, but not in new settings

packages/js/settings-editor/src/legacy/form.tsx
psealock: Good callout, thanks @rjchow

psealock:
>I noticed something which isn't within the scope of this PR - In legacy settings when "Enable product reviews" is unchecked the checkboxes below are hidden, but not in new settings

Yeah thats a jQuery script controlling that behaviour. I'll address it in a follow up

packages/js/settings-editor/src/legacy/form.tsx
```diff
@@ -0,0 +1,104 @@
...
+			query.section = activeSection;
+		}
+
+		const formData = getFormData();
+		formData.save = 'Save changes';
+		formData._wpnonce = settingsData._wpnonce;
+		formData._w_http_referer = '/wp-admin/' + getNewPath( query );
+
+		// eslint-disable-next-line no-console
+		console.log(
```
adrianduffell, line 69: Do we want a console.log here when its merged?

psealock: Yes, I think so. The follow up PRs depend on it https://github.com/woocommerce/woocommerce/pull/55466

But its only temporary, I'll remove it when creating the API https://github.com/woocommerce/woocommerce/pull/55467

adrianduffell (approved):
Code looks good and it's testing well. Nice! 🚀

PR #55358, opened 2025-02-12
[API tests] Unskip `products-crud` on WPCOM and Pressable.

rodelgc:

Part of #54662.

The `describe( 'Product review tests: CRUD' )` suite passed in WPCOM and Pressable immediately after removing the skip tags, so no further changes were done in here.

In `test( 'can batch update product shipping classes' )` I suffixed the shipping classes with a random string to ensure uniqueness on every run. The reason why it fails on WPCOM and Pressable is because it tries to create duplicates of already existing shipping classes, possibly those that were left over by previous failed runs.

1. Make sure CI checks pass.
2. Run on WPCOM and Pressable, make sure it passes.
```bash
export E2E_ENV_KEY="..."
pnpm test:e2e:wpcom products-crud.test.js
pnpm test:e2e:pressable products-crud.test.js 
```

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 
Unskip products-crud.test.js on WPCOM and Pressable.

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/tests/e2e-pw/tests/api-tests/products/products-crud.test.js
+++ b/plugins/woocommerce/tests/e2e-pw/tests/api-tests/products/products-crud.test.js
@@ -1,7 +1,6 @@
 const {
 	test: baseTest,
 	expect,
-	tags,
 } = require( '../../../fixtures/api-tests-fixtures' );
 const { BASE_URL } = process.env;
 const { admin } = require( '../../../test-data/data' );
--- a/plugins/woocommerce/tests/e2e-pw/tests/api-tests/products/products-crud.test.js
+++ b/plugins/woocommerce/tests/e2e-pw/tests/api-tests/products/products-crud.test.js
@@ -959,91 +959,93 @@ test.describe( 'Products API tests: CRUD', () => {
 			);
 		} );
 
-		test(
-			'can batch update product shipping classes',
-			{ tag: [ tags.SKIP_ON_PRESSABLE, tags.SKIP_ON_WPCOM ] },
-			async ( { request } ) => {
-				// Batch create product shipping classes.
-				const response = await request.post(
-					`wp-json/wc/v3/products/shipping_classes/batch`,
-					{
-						data: {
-							create: [
-								{
-									name: 'Small Items',
-								},
-								{
-									name: 'Large Items',
-								},
-							],
-						},
-					}
-				);
-				const responseJSON = await response.json();
-				expect( response.status() ).toEqual( 200 );
-				expect( responseJSON.create[ 0 ].name ).toEqual(
-					'Small Items'
-				);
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/tests/e2e-pw/tests/api-tests/products/products-crud.test.js
```diff
@@ -1,11 +1,10 @@
 const {
 	test: baseTest,
 	expect,
-	tags,
 } = require( '../../../fixtures/api-tests-fixtures' );
-const { BASE_URL } = process.env;
+const { IS_WPCOM, IS_PRESSABLE } = process.env;
 const { admin } = require( '../../../test-data/data' );
-const shouldSkip = BASE_URL !== undefined;
+const shouldSkip = IS_WPCOM || IS_PRESSABLE;
```
adimoldovan: Is this still valid? Is see some steps are skipped based on this, but the comment refers to CI runs. If it's still valid for wpcom and pressable, maybe it makes sense to also update the comments.

rodelgc: Thanks Adi for noticing this, as I was also curious myself whether we still need to skip those steps. It seems that `shouldSkip` was introduced in #39858 to prevent false positives when running the daily tests against the daily smoke test site back then, which could either be the Bluehost site or the Pressable one.

I tried unskipping those steps skipped by `shouldSkip` and they all passed repeatedly in my experiment both in WPCOM and Pressable. So I'm not sure if we need to skip those steps anymore.

But with that said, maybe a safer approach is to deal with the removal of the `shouldSkip`'ed steps in a separate PR? I'm leaning towards unskipping those steps since they seem stable enough on WPCOM and Pressable, but I'm not sure if we'll encounter any gotcha's when we do that. 

For the meantime, I'm pushing 8aacbf2fd2011f897c33143c8e0b98b3e2bba5c8 to revert the `shouldSkip` condition to the original one, while at the same time reflecting on the comments that the decision to skip those steps was based on the possibility that they would fail due to the external host's caching. Let me know if this approach works for you.

I'll submit a different PR for removing `shouldSkip`.

rodelgc: Drafted #55622 to remove `shouldSkip`. I'll submit that for review after testing it several more times against WPCOM and Pressable, and after tinkering with their cache settings to see if they'll have an effect on those skipped steps.

adimoldovan (approved):
> Drafted https://github.com/woocommerce/woocommerce/pull/55622 to remove shouldSkip

Sounds good to me. Let's merge this one then 🚀.

Issue #55165, opened 2025-02-06
WooCommerce 9.6.1: New branding colours in woocommerce.css causing front-end display (styling) issues


syllie:
### Prerequisites

- [x] I have carried out troubleshooting steps and I believe I have found a bug.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

With new branding on your end the default colours for woocommerce.css are also updated.

If this would have been great if (finally!) the variables would be utilized as `var(--wc-<colourname>)` in the compiled stylesheet, this would have been a step forward, and we finally would able to override the variables and customize our themes with ease.

But no, someone thought it a great idea to introduce new variable colours where they used to be neutral, but the hex values are still hardcoded in the stylesheet, so now we can go and update our theme stylesheets to undo this mess. To be honest, I was not opting for lilac notification backgrounds, or purple text for placeholders in payment gateways  as an 'out of the box' experience.

![Image](https://github.com/user-attachments/assets/dcced31c-ea0f-496b-864d-ec3b581278fb)

![Image](https://github.com/user-attachments/assets/553779f8-b0b1-476c-b8af-073e0f23a263)

### Expected behavior

Return neutral values for 
--wc-secondary
--wc-secondary-text

OR
Finally do the right thing and return the variables in the stylesheet so we can override the stylesheet properly.

### Actual behavior

![Image](https://github.com/user-attachments/assets/fbe0118d-d5c5-4362-b406-1e23da1fe734)

![Image](https://github.com/user-attachments/assets/d1137551-1cc3-4694-a3bf-e4660fb31c6d)

![Image](https://github.com/user-attachments/assets/d5ddc8db-c2df-451e-aff7-6d010769c433)

### Steps to reproduce

Install a basic classic theme and enable WooCommerce.

### WordPress Environment

Latest version of Wordpress, latest version of WooCommerce. This is only a cosmetic issue.

### Isolating the problem

- [x] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [x] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/products/storefront/).
- [x] I can reproduce this bug consistently using the steps above.

leecollings-sb:
Is this likely to be fixed and reverted in a further update, or should I continue to search for other elements and create CSS overrides?

Obviously, I don't want to have to go through a load of work that will end up not needing to be done.

Really frustrated with this though!

[...]

PR #55366, opened 2025-02-12
Return secondary color CSS variable to its original color

Aljullu:

Closes https://github.com/woocommerce/woocommerce/issues/55165.

This PR returns the secondary color variable to its original value before the WooCommerce rebranding.

1. Install a theme that uses WooCommerce colors in visible parts of the UI, for example, Twenty Twenty Four.
2. Go to the Cart page when using the Cart shortcode.
3. Check the color of the Coupon code input field, it should be `#cfc8d8`.

Before | After
--- | ---
imatge | imatge

4. Optionally, check the CSS variables loaded in the page and verify the secondary color is `#e9e6ed`.

Before | After
--- | ---
imatge | imatge

5. With the theme [GeneratePress](https://wordpress.org/themes/generatepress/), go to the Cart page with the Cart shortcode and verify the buttons don't use a shade of purple.

Before | After
--- | ---
imatge | imatge


plugins/woocommerce/client/legacy/css/_variables.scss
```diff
@@ -11,7 +11,9 @@ $blue:          	#2ea2cc !default;
 $primary:           #720eec !default;                                    // Primary color for buttons (alt)
 $primarytext:       desaturate(lighten($primary, 50%), 18%) !default;    // Text on primary color bg
 
-$secondary:         desaturate(lighten($primary, 40%), 21%) !default;    // Secondary buttons
+// We use the old WooCommerce color as the base of the secondary color on purpose.
+// See: https://github.com/woocommerce/woocommerce/issues/55165.
+$secondary:         desaturate(lighten(#7F54B3, 40%), 21%) !default;    // Secondary buttons
```
naman03malhotra, line 16: Will reverting to the original color affect the branding side of things? Asking because there must be a reason were updated it in the first place.

Aljullu: Looking at it now, I think the secondary color should have been excluded from the rebranding. As https://github.com/woocommerce/woocommerce/issues/55165 illustrates, people assumed it was a gray completely disconnected from the WooCommerce purple. The secondary color was mostly used in the frontend, so any changes to it ended up impacting the frontend of existing stores.

In the PR, I left the calculation `desaturate(lighten(#7F54B3, 40%), 21%)`, but I think we could perfectly replace it with the computed value (`#e9e6ed`) to make it clearer that this color is disconnected from the WooCommerce purple.

Konamiman (approved):
Tested with GeneratePress, the change looks goos.

plugins/woocommerce/client/legacy/css/_variables.scss
naman03malhotra: Thanks for the context, it will be helpful to leave a comment so that we don't touch it in the future :)

Aljullu: Ok, I will do that in `trunk` if that's fine, so we don't delay the release of this fix. :slightly_smiling_face:

PR #55368, opened 2025-02-12
[e2e tests] Move and refactor email and settings specs

adimoldovan:

Part of pdV5qK-Ie-p2, address some of the follow-up todo comments:

I went deeper than initially planned into this one, but I'm happy with the results of the email specs refactoring. They are now less flaky, much faster, more maintainable and consistent in what and how they check.

What changed:
- `order-email-receiving.spec.js` was deleted. It was checking stuff that was already covered in order-emails spec.
- moved `order-emails.spec.js` to the emails folder
- moved `settings-email.spec.js` to the emails folder
- renamed `account-emails-receiving.spec.js` into `account-emails.spec.js`
- simplified the structure by removing unnecessary describe blocks
- renamed tests
- added a new email utils with 2 functions to check that an email exists and check the content of an email. With a fixed search, and sorting by timestamp DESC, this makes sure the right email is found. Coupled with unique receiver address this should fix issues with emails not being found when a lot of emails already exist.
- refactored `order-emails.spec.js`
	- use fixtures
	- removed the before each step that deleted all email logs via the UI
	- use a loop to run the same test with different data - more consistent, no code duplication, easier to see what email scenario is checked, easier to add new ones
	- create new random unique billing email addresses
	- use the newly created email util to check the expected email
- refactored `account-emails.spec.js`:
	- use fixtures
	- removed the before each step that deleted all email logs via the UI
	- create new random unique users - each test uses a different random user
	- user are created via the api - faster, less flaky
	- use the newly created email util to check the expected email

Unrelated changes:
- moved `settings-shipping.spec.js` to the shipping folder

CI should be green

To run the affected tests locally:

```
pnpm test:e2e email order settings
```


Stojdza (approved):
A lot of good stuff here 👏 

Tested locally with wp-env and against Pressable ✅

PR #55369, opened 2025-02-12
[e2e tests] Fix IS_MULTISITE always being set

adimoldovan:

https://github.com/woocommerce/woocommerce/pull/55162 introduced the global setup step that checks the `environment.wp_multisite` value and sets the environment variable `IS_MULTISITE`. However, the value of `IS_MULTISITE` is not checked when evaluating the variable in  tests like:
test.skip( process.env.IS_MULTISITE, '');

Only the variable being set is checked, 
and because of this all tests that should be skipped for multisite are always being skipped, even for non-multisite setups.

The fix is to unset IS_MULTISITE if `environment.wp_multisite` is false.

Run tests that are skipped in multisite. Make sure they run.

```
pnpm test:e2e account-email-receiveing onboarding-wizard users-create users-manage 
```

### DIFF
--- a/plugins/woocommerce/changelog/e2e-fix-is-multisite-always-true
+++ b/plugins/woocommerce/changelog/e2e-fix-is-multisite-always-true
@@ -1,0 +1,4 @@
+Significance: patch
+Type: dev
+
+E2E tests: fixed IS_MULTISITE always resolved to true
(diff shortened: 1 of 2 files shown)


Stojdza (approved):
There was a typo in the command `pnpm test:e2e account-email-receiveing...` but with `pnpm test:e2e account-email-receiving...` everything ran as expected 👍

Issue #48511, opened 2024-06-15
Shop page does not show associated template in right area settings


paaljoachim:
### Prerequisites

- [X] I have carried out troubleshooting steps and I believe I have found a bug.
- [X] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

Going to the shop page or as I called it in Norwegian butikken.
I see the following in the right sidebar page settings.

![Screenshot 2024-06-15 at 20 48 37](https://github.com/woocommerce/woocommerce/assets/5323259/2a276ea9-fdcb-4521-9ab4-56e1a846ea32)

Here is what the Cart page sidebar settings look like.
![Screenshot 2024-06-15 at 20 49 34](https://github.com/woocommerce/woocommerce/assets/5323259/40408a67-6615-4b39-8145-3c40679aaee6)

The Shop page lacks Template information in the settings sidebar.

### Expected behavior

I expected to see the associated Template in the sidebar Settings panel

### Actual behavior

Template is not seen in the settings panel.

### Steps to reproduce

Open up the Shop page and notice the missing Template information.

### WordPress Environment

WP 6.5.4
Theme: Twenty Twenty Four

### Isolating the problem

- [ ] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [X] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/products/storefront/).
- [X] I can reproduce this bug consistently using the steps above.

imanish003:
Hey @paaljoachim, Thanks for reporting this issue. I can confirm the issue on my end as well. I have added this issue to our backlog. 

Out of curiosity, why do you want to see the template information in the settings sidebar? Don't you use the Site Editor to edit the templates, including the Product Catalog (Shop page) template? I'd like to understand the purpose behind the need for template information in the settings sidebar.

paaljoachim:
Hi @imanish003 

Consistency is the key word here.
Creating a connection/association between a post/page <-> Templates in the site editor.

When Template information is listed in the post/page sidebar it shows which template is associated with the post/page. One can also go to edit the template as well by clicking the template and selecting edit. It gives direct access to editing the template.

So it is important to add the information as it shows the associated template and one does not start to wonder or have doubt about which template is associated with the given page/post.

paaljoachim:
I will also link in this issue I just made in regards to a consistent experience going through WooCommerce pages.
[...]

PR #55407, opened 2025-02-12
BlockTemplateController: Remove unused code to hide C&C template selector

tjcafferkey:

There was code to hide the template selector from cart and checkout Pages to encourage users to use the Site Editor instead.

This PR removes this code because:

* Its not being used anyway
* It is consistent with all other pages
* We have an issue to remove similar logic for the Shop page (https://github.com/woocommerce/woocommerce/issues/48511)
* Users can click "Edit template" from this selector now to get better access to the template and help them understand the relationship.

Closes # .

1. Load the cart and checkout pages in WP Admin
2. Ensure the template selector is still visible

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/remove-unused-cart-checkout-template-selector-logic
+++ b/plugins/woocommerce/changelog/remove-unused-cart-checkout-template-selector-logic
@@ -1,0 +1,4 @@
+Significance: patch
+Type: tweak
+
+Remove code that was not being used
--- a/plugins/woocommerce/src/Blocks/BlockTemplatesController.php
+++ b/plugins/woocommerce/src/Blocks/BlockTemplatesController.php
@@ -174,7 +174,6 @@ class BlockTemplatesController {
 		return $settings;
 	}
 
-
 	/**
 	 * Prevents shortcodes in templates having their HTML content broken by wpautop.
 	 *
--- a/plugins/woocommerce/src/Blocks/BlockTemplatesController.php
+++ b/plugins/woocommerce/src/Blocks/BlockTemplatesController.php
@@ -29,7 +29,6 @@ class BlockTemplatesController {
 		add_filter( 'taxonomy_template_hierarchy', array( $this, 'add_fallback_template_to_hierarchy' ), 10, 1 );
 		add_filter( 'block_type_metadata_settings', array( $this, 'add_plugin_templates_parts_support' ), 10, 2 );
 		add_filter( 'block_type_metadata_settings', array( $this, 'prevent_shortcodes_html_breakage' ), 10, 2 );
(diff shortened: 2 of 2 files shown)


opr (commented):
The code to hide the template changer is broken because the CSS targets the incorrect selector. 

This isn't caused by your PR, but changing the template doesn't seem to work, it always defaults to the Cart/Checkout template, even if you set and customize a new one.

I think this needs a product decision on how to proceed more than a technical one, so @pierorocca what do you think?

If we allow the template of the Cart and Checkout _pages_ to be edited, we'd have to make it respect the template the merchant chooses, potentially allowing them to remove the Cart block from their Cart page, not sure what the implications of this are in terms of how badly it breaks the store (e.g. does it make it totally inaccessible, fatal errors etc?) but it would certainly prevent orders being placed.

The alternative would be to fix the broken code and ensure the template picker is hidden when editing the Cart/Checkout pages, and look at the above as part of a separate effort. This is my recommendation.

> [!NOTE] 
> After more testing, it's only the editor that doesn't show the right template, so it's even more of a confusing experience for the merchant. We should either remove the selector (again) or make the editor show the correct selected tempalate.

tjcafferkey:
Hmm, I see the predicament. I was actually unaware it wasn't possible to change the template. That said, I have another suggestion.

We currently have an issue open to re-enable this template selector for the Shop page https://github.com/woocommerce/woocommerce/issues/48511

One of the main reasons is because it retains consistency and helps people understand the relationship between Pages and Templates, not only that but there's a shortcut (see below) to dive into Edit Template from here too.

What do you think about enabling the template selector (_for Cart, Checkout and Shop pages_) but hiding the options "Change template" and "Create a new template" options from the dropdown (see below).

We could then refactor the code responsible so its consistent because at the moment between Shop and C&C we are hiding this in two different ways. For comparison see this PR (https://github.com/woocommerce/woocommerce/pull/55409) to look at how it was hidden for the Shop page.

Screenshot 2025-02-13 at 08 09 47

opr:
> We currently have an issue open to re-enable this template selector for the Shop page #48511

Yeah I can see it making some sense for the Cart/Checkout if it actually works for the merchant, I think if we ever ship additional templates for Cart/Checkout then we'd have to do it.


> One of the main reasons is because it retains consistency and helps people understand the relationship between Pages and Templates, not only that but there's a shortcut (see below) to dive into Edit Template from here too.
> 
> What do you think about enabling the template selector (_for Cart, Checkout and Shop pages_) but hiding the options "Change template" and "Create a new template" options from the dropdown (see below).

I like this idea, it gives a hint about what's controlling the page, but doesn't let the merchant break their site 👌🏼

PR #55418, opened 2025-02-12
Prevent data stores being registered more than once due to circular dependency

opr:

Reverts a couple of changes we made to use `StoreDescriptor` when accessing data stores where importing that store descriptor would lead to a "circular dependency" (not a true circular dependency, as there's no script with a dependency on a script that depends on it, this issue is caused by File A loading File B which loads File A.)

1. Ensure you can place an order with a couple of different payment methods, there should be no console errors and the order should place fine.
2. Check the console for any messages like `Store "wc/store/validation" is already registered`. You should not see any.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce-blocks/assets/js/data/cart/utils.ts
+++ b/plugins/woocommerce-blocks/assets/js/data/cart/utils.ts
@@ -10,18 +10,22 @@ import {
 	Cart,
 	CartResponse,
 } from '@woocommerce/types';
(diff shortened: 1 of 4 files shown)


mikejolley (approved):
I don't understand the ins and outs of this but it did resolve the console errors so looks 🚢 ready 👍🏻

PR #55419, opened 2025-02-12
Do not call `get_visible` method when all lists are already hidden.

### LINKED ISSUE #55350: Ensure that `wcSettings.admin.visibleTaskListIds` in TaskLists.php loads only on the WooCommerce → Home page.
`wcSettings.admin.visibleTaskListIds` is [loaded](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Admin/Features/OnboardingTasks/TaskLists.php#L468) unconditionally on all WC admin pages.

We can check the current page I.D and load it only on WooCommerce -> Home.

The following code should work:

```
$page = PageController::get_instance()->get_current_page();
if ( $page && $page['id'] === 'woocommerce-home' ) {
	$settings['visibleTaskListIds'] = array_keys( self::get_visible() );
}
```

moon0326:

Closes #55350  

~~This PR removes use of `wcSettings.admin.visibleTaskListIds`.~~

~~I'm not exactly sure why we decided to populate `visibleTaskListIds` value in PHP when we already have [taskLists](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/admin/client/homescreen/layout.js#L203). My best guess is that we did not want to wait for the [getTaskLists](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/admin/client/homescreen/layout.js#L198) call to complete. However, I didn’t notice much of a difference when using taskLists instead of `wcSettings.admin.visibleTaskLists`~~

I initially removed `wcSettings.admin.visibleTaskListIds`, but https://github.com/woocommerce/woocommerce/pull/55285 uses it to optimize the tasks REST API call. I conducted a basic benchmark test on https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Admin/Features/OnboardingTasks/TaskLists.php#L468 and it did not take more than 1ms.  I believe we can keep it and use its value to optimize other calls. 

I have also made a change in [Do no check get_visible if all lists are hidden already](https://github.com/woocommerce/woocommerce/pull/55419/commits/9495bf9ed57c17713cbe76b29c495ea4f8699db4) to set the `visibleTaskListIds` to an empty array if all lists are hidden already.

1. Create a new JN site using this branch.
2. Complete the onboarding and navigate to WooCommerce -> Home
3. Verify that everything works as expected.
4. Interact with the task list and confirm that all functionalities work as expected.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Do not call `get_visible` function is all lists are already hidden.

</details>

<details>

#### Comment 

</details>


adrianduffell (approved):
Nice! The change looks good and tested well 🚀 

Agree this is the only instance of visibleTaskListIds being used in core.

plugins/woocommerce/src/Admin/Features/OnboardingTasks/TaskLists.php
```diff
@@ -465,8 +465,17 @@ public static function menu_task_count() {
...
 	}
+
+	/**
+	 * Check if all task lists are hidden.
+	 *
+	 * @return bool
+	 */
+	public static function all_hidden() {
+		$hidden_lists = get_option( TaskList::HIDDEN_OPTION, array() );
+		return count( $hidden_lists ) === count( self::$lists );
```
rjchow:
```suggestion
		return count( $hidden_lists ) === count( self::get_lists() );
```

rjchow (commented):
If the TaskList::HIDDEN_OPTION is not autoloaded I think this might result in a performance regression - my gut feel is that iterating through the lists in heap would be faster than get_option.

moon0326:
> If the TaskList::HIDDEN_OPTION is not autoloaded I think this might result in a performance regression - my gut feel is that iterating through the lists in heap would be faster than get_option.

It is autoloaded. Even if it's not autoloaded, I think this is the way we should go with. 

TaskLists has TaskList and TaskList has Task objects. We have X number of Task::can_view() calls and we don't have 100% control over them. 

[Shipping](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Admin/Features/OnboardingTasks/Tasks/Shipping.php#L86) task alone can take more than a single get_option 🥲 
[WooCommercePyaments](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Admin/Features/OnboardingTasks/Tasks/WooCommercePayments.php#L125) can be slow as well

adrianduffell:
> It is autoloaded

Interesting, I think this is a bad option to autoload since it will then be queried with every request (including front-end), but it is not something needed beyond the onboarding phase of the store, so is a bit wasteful.

I think it's probably better to place the hidden state in user meta. That way it's only loaded for admin users and could be queried performantly in `task_list_preloaded_settings` function. I think it also makes sense -- hiding a task list is a user preference. When there are multiple admins on the site, it should be possible to allow one user to hide it and another still see it.

rjchow (approved):
> TaskLists has TaskList and TaskList has Task objects. We have X number of Task::can_view() calls and we don't have 100% control over them.

Ah yeah ok good point, I was thinking it was just the `count( get_lists() )` call.

I don't have anything else to add, but what @adrianduffell mentioned about it being in user meta is interesting

moon0326:
> Interesting, I think this is a bad option to autoload since it will then be queried with every request (including front-end), but it is not something needed beyond the onboarding phase of the store, so is a bit wasteful.

I agree that we should not set an option to autoload if it's not globally used. However, all autoloaded options are queried in a single mysql query and cached. See https://github.com/WordPress/wordpress-develop/blob/6.7/src/wp-includes/option.php#L619. If it's an option used on pages like Woo Home, where speed is a priority, it might be worth considering.

> I think it's probably better to place the hidden state in user meta. That way it's only loaded for admin users and could be queried performantly in task_list_preloaded_settings function. I think it also makes sense -- hiding a task list is a user preference. When there are multiple admins on the site, it should be possible to allow one user to hide it and another still see it.

This is a good idea! I think we should look into it further and work on it in a separate PR. In my opinion, this will require a bit more testing. We also need to consider that this change might disrupt existing stores with multiple users, so we should apply it only to new stores.

adrianduffell:
> This is a good idea! I think we should look into it further and work on it in a separate PR.

Thanks @moon0326 , I'm happy with this plan. Let's merge! 🚀

PR #55442, opened 2025-02-13
Fix a 403 forbidden errors for direct access to woocommerce_uploads

### LINKED ISSUE #35166: Server errors when downloadable product images are accessed by media library for logged in admin users
### Prerequisites

### Describe the bug

We are experiencing issues for some of our clients where logged in admin users editing products generate server errors along the lines of `client denied by server configuration: .../wp-content/uploads/woocommerce_uploads/2021/11/downloadable-product-file.jpg, referer:.../wp-admin/post.php?post=1611&action=edit`

This caused serious problems for us as we have fail2ban running on our server, and these errors cause our client admin users to be locked out of their ecommerce sites.

We tracked the issue down to the media upload modal & media library, where WP attempts to load thumbnail images for downloadable products, which are stored in `woocommerce_uploads`. The `woocommerce_uploads` directory has an auto generated `.htaccess` with deny from all, so attempts to directly access these images are rejected by the server.

Prior to Woocommerce 5.7.0 thumbnails for downloadable products were not displayed in the media library, the default WP image icon was displayed.

Subsequent to https://github.com/woocommerce/woocommerce/pull/30385/commits connected to https://github.com/woocommerce/woocommerce/issues/29988 WP will attempt to serve these thumbnails to logged in admin users, however the .htaccess prevents this. Thumbnails for these items are not displayed, and server errors are generated.

We're currently hotfixing this by removing the problematic filters and re-adding them in a reverted version as described here https://github.com/woocommerce/woocommerce/issues/29988#issuecomment-1283172867

This issue is closely related to  https://github.com/woocommerce/woocommerce/issues/29988 but that issue has been closed.

### Expected behavior

Admin users should be able to open add media dialogs on the edit product page  or access the media section of the WP dashboard without generating server errors. Ideally thumbnail preview of downloadable product items would be visible in media dialogs if those are listed, however the display of a generic picture icon would be preferable to an error situation.

### Actual behavior

When admin users open add media / edit thumbnail icons on the edit product page, or access the media section of the WP dashboard, server logs will show `client denied by server configuration` errors. Since access to the `woocommerce_uploads` folder is restricted by `.htaccess`, thumbnails for downloadable item images are not shown, the user just sees empty boxes where those should be.

### Steps to reproduce

1. Add at least one downloadable item to a product.
2. On the edit product page, attempt to change the product thumbnail or otherwise access the media library.
3. Observer that where downloadable items are included in the media list, no thumbnail is shown. (may need to filter so that downloadable items are in the list, e.g. by filtering by media type)
4. Check server logs and observe `client denied by server configuration` errors. 
5. Optionally, if fail2ban is running, get locked out of the entire site.

### WordPress Environment

`
### WordPress Environment ###

WC Version: 7.0.0
REST API Version: ✔ 7.0.0
WC Blocks Version: ✔ 8.5.1
Action Scheduler Version: ✔ 3.4.0
Log Directory Writable: ✔
WP Version: 6.0.3
WP Multisite: –
WP Memory Limit: 256 MB
WP Debug Mode: –
WP Cron: –
Language: en_US
External object cache: –

### Server Environment ###

Server Info: nginx/1.20.2
PHP Version: 7.4.32
PHP Post Max Size: 100 MB
PHP Time Limit: 60
PHP Max Input Vars: 1000
cURL Version: 7.29.0
NSS/3.53.1

SUHOSIN Installed: –
MySQL Version: 5.5.5-10.5.17-MariaDB
Max Upload Size: 100 MB
Default Timezone is UTC: ✔
fsockopen/cURL: ✔
SoapClient: ✔
DOMDocument: ✔
GZip: ✔
Multibyte String: ✔
Remote Post: ✔
Remote Get: ✔

### Database ###

WC Database Version: 6.9.4
WC Database Prefix: wp_1046_
Total Database Size: 132.03MB
Database Data Size: 96.94MB
Database Index Size: 35.09MB
wp_1046_woocommerce_sessions: Data: 0.30MB + Index: 0.02MB + Engine MyISAM
wp_1046_woocommerce_api_keys: Data: 0.00MB + Index: 0.00MB + Engine MyISAM
wp_1046_woocommerce_attribute_taxonomies: Data: 0.00MB + Index: 0.01MB + Engine MyISAM
wp_1046_woocommerce_downloadable_product_permissions: Data: 0.02MB + Index: 0.03MB + Engine MyISAM
wp_1046_woocommerce_order_items: Data: 0.04MB + Index: 0.02MB + Engine MyISAM
wp_1046_woocommerce_order_itemmeta: Data: 0.22MB + Index: 0.17MB + Engine MyISAM
wp_1046_woocommerce_tax_rates: Data: 0.00MB + Index: 0.01MB + Engine MyISAM
wp_1046_woocommerce_tax_rate_locations: Data: 0.00MB + Index: 0.00MB + Engine MyISAM
wp_1046_woocommerce_shipping_zones: Data: 0.00MB + Index: 0.00MB + Engine MyISAM
wp_1046_woocommerce_shipping_zone_locations: Data: 0.00MB + Index: 0.01MB + Engine MyISAM
wp_1046_woocommerce_shipping_zone_methods: Data: 0.00MB + Index: 0.00MB + Engine MyISAM
wp_1046_woocommerce_payment_tokens: Data: 0.00MB + Index: 0.00MB + Engine MyISAM
wp_1046_woocommerce_payment_tokenmeta: Data: 0.00MB + Index: 0.01MB + Engine MyISAM
wp_1046_woocommerce_log: Data: 0.00MB + Index: 0.00MB + Engine MyISAM
wp_1046_actionscheduler_actions: Data: 0.11MB + Index: 0.13MB + Engine InnoDB
wp_1046_actionscheduler_claims: Data: 0.02MB + Index: 0.02MB + Engine InnoDB
wp_1046_actionscheduler_groups: Data: 0.02MB + Index: 0.02MB + Engine InnoDB
wp_1046_actionscheduler_logs: Data: 0.06MB + Index: 0.03MB + Engine InnoDB
wp_1046_commentmeta: Data: 0.00MB + Index: 0.00MB + Engine MyISAM
wp_1046_comments: Data: 0.14MB + Index: 0.07MB + Engine MyISAM
wp_1046_fsmpt_email_logs: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_give_commentmeta: Data: 0.14MB + Index: 0.16MB + Engine InnoDB
wp_1046_give_comments: Data: 1.52MB + Index: 0.00MB + Engine InnoDB
wp_1046_give_donationmeta: Data: 5.52MB + Index: 8.55MB + Engine InnoDB
wp_1046_give_donormeta: Data: 0.41MB + Index: 0.35MB + Engine MyISAM
wp_1046_give_donors: Data: 0.06MB + Index: 0.05MB + Engine MyISAM
wp_1046_give_formmeta: Data: 0.28MB + Index: 0.19MB + Engine InnoDB
wp_1046_give_log: Data: 3.52MB + Index: 0.66MB + Engine InnoDB
wp_1046_give_logmeta: Data: 0.16MB + Index: 0.16MB + Engine InnoDB
wp_1046_give_logs: Data: 2.52MB + Index: 0.00MB + Engine InnoDB
wp_1046_give_migrations: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_give_revenue: Data: 0.09MB + Index: 0.00MB + Engine InnoDB
wp_1046_give_sequential_ordering: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_give_sessions: Data: 0.02MB + Index: 0.02MB + Engine InnoDB
wp_1046_give_subscriptionmeta: Data: 0.02MB + Index: 0.03MB + Engine InnoDB
wp_1046_give_subscriptions: Data: 0.02MB + Index: 0.06MB + Engine InnoDB
wp_1046_links: Data: 0.00MB + Index: 0.00MB + Engine MyISAM
wp_1046_ms_snippets: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_oopspam_frm_ham_entries: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_oopspam_frm_spam_entries: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_options: Data: 8.25MB + Index: 0.17MB + Engine MyISAM
wp_1046_oses_attachments: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_oses_clicks: Data: 0.02MB + Index: 0.02MB + Engine InnoDB
wp_1046_oses_emails: Data: 2.50MB + Index: 0.02MB + Engine InnoDB
wp_1046_oses_email_attachments: Data: 0.02MB + Index: 0.02MB + Engine InnoDB
wp_1046_oses_failures: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_oses_jobs: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_postmeta: Data: 21.77MB + Index: 4.11MB + Engine MyISAM
wp_1046_posts: Data: 9.74MB + Index: 0.58MB + Engine MyISAM
wp_1046_redirection_404: Data: 0.60MB + Index: 0.11MB + Engine MyISAM
wp_1046_redirection_groups: Data: 0.00MB + Index: 0.00MB + Engine MyISAM
wp_1046_redirection_items: Data: 0.00MB + Index: 0.01MB + Engine MyISAM
wp_1046_redirection_logs: Data: 0.02MB + Index: 0.00MB + Engine MyISAM
wp_1046_snippets: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_termmeta: Data: 0.00MB + Index: 0.01MB + Engine MyISAM
wp_1046_terms: Data: 0.00MB + Index: 0.01MB + Engine MyISAM
wp_1046_term_relationships: Data: 0.06MB + Index: 0.13MB + Engine MyISAM
wp_1046_term_taxonomy: Data: 0.00MB + Index: 0.00MB + Engine MyISAM
wp_1046_usermeta: Data: 0.16MB + Index: 0.07MB + Engine MyISAM
wp_1046_users: Data: 0.01MB + Index: 0.01MB + Engine MyISAM
wp_1046_wcpdf_invoice_number: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_wc_admin_notes: Data: 0.08MB + Index: 0.00MB + Engine InnoDB
wp_1046_wc_admin_note_actions: Data: 0.05MB + Index: 0.02MB + Engine InnoDB
wp_1046_wc_category_lookup: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_wc_customer_lookup: Data: 0.02MB + Index: 0.03MB + Engine InnoDB
wp_1046_wc_download_log: Data: 0.02MB + Index: 0.03MB + Engine InnoDB
wp_1046_wc_order_coupon_lookup: Data: 0.02MB + Index: 0.03MB + Engine InnoDB
wp_1046_wc_order_product_lookup: Data: 0.02MB + Index: 0.06MB + Engine InnoDB
wp_1046_wc_order_stats: Data: 0.02MB + Index: 0.05MB + Engine InnoDB
wp_1046_wc_order_tax_lookup: Data: 0.02MB + Index: 0.03MB + Engine InnoDB
wp_1046_wc_product_attributes_lookup: Data: 0.02MB + Index: 0.02MB + Engine InnoDB
wp_1046_wc_product_download_directories: Data: 0.02MB + Index: 0.02MB + Engine InnoDB
wp_1046_wc_product_meta_lookup: Data: 0.02MB + Index: 0.09MB + Engine InnoDB
wp_1046_wc_rate_limits: Data: 0.02MB + Index: 0.02MB + Engine InnoDB
wp_1046_wc_reserved_stock: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_wc_tax_rate_classes: Data: 0.02MB + Index: 0.02MB + Engine InnoDB
wp_1046_wc_webhooks: Data: 0.02MB + Index: 0.02MB + Engine InnoDB
wp_1046_wfblockediplog: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_wfblocks7: Data: 0.02MB + Index: 0.05MB + Engine InnoDB
wp_1046_wfconfig: Data: 1.44MB + Index: 0.00MB + Engine InnoDB
wp_1046_wfcrawlers: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_wffilechanges: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_wffilemods: Data: 12.55MB + Index: 0.00MB + Engine InnoDB
wp_1046_wfhits: Data: 1.52MB + Index: 0.09MB + Engine InnoDB
wp_1046_wfhoover: Data: 0.02MB + Index: 0.02MB + Engine InnoDB
wp_1046_wfissues: Data: 0.02MB + Index: 0.06MB + Engine InnoDB
wp_1046_wfknownfilelist: Data: 5.52MB + Index: 0.00MB + Engine InnoDB
wp_1046_wflivetraffichuman: Data: 0.02MB + Index: 0.02MB + Engine InnoDB
wp_1046_wflocs: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_wflogins: Data: 0.06MB + Index: 0.03MB + Engine InnoDB
wp_1046_wfls_2fa_secrets: Data: 0.02MB + Index: 0.02MB + Engine InnoDB
wp_1046_wfls_settings: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_wfnotifications: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_wfpendingissues: Data: 0.02MB + Index: 0.06MB + Engine InnoDB
wp_1046_wfreversecache: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_wfsnipcache: Data: 0.02MB + Index: 0.05MB + Engine InnoDB
wp_1046_wfstatus: Data: 0.14MB + Index: 0.11MB + Engine InnoDB
wp_1046_wftrafficrates: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_wpfm_backup: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_wpforms_entries: Data: 0.34MB + Index: 0.02MB + Engine InnoDB
wp_1046_wpforms_entry_fields: Data: 0.19MB + Index: 0.14MB + Engine InnoDB
wp_1046_wpforms_entry_meta: Data: 0.02MB + Index: 0.06MB + Engine InnoDB
wp_1046_wpforms_form_locker_email_verification: Data: 0.02MB + Index: 0.03MB + Engine InnoDB
wp_1046_wpforms_tasks_meta: Data: 0.02MB + Index: 0.00MB + Engine InnoDB
wp_1046_wsal_metadata: Data: 9.34MB + Index: 14.81MB + Engine InnoDB
wp_1046_wsal_occurrences: Data: 4.44MB + Index: 1.89MB + Engine InnoDB
wp_1046_yoast_indexable: Data: 1.52MB + Index: 0.72MB + Engine InnoDB
wp_1046_yoast_indexable_hierarchy: Data: 0.09MB + Index: 0.14MB + Engine InnoDB
wp_1046_yoast_migrations: Data: 0.02MB + Index: 0.02MB + Engine InnoDB
wp_1046_yoast_primary_term: Data: 0.02MB + Index: 0.03MB + Engine InnoDB
wp_1046_yoast_seo_links: Data: 0.30MB + Index: 0.23MB + Engine InnoDB
wp_1046_yoast_seo_meta: Data: 0.08MB + Index: 0.00MB + Engine InnoDB

### Post Type Counts ###

acf-field: 45
acf-field-group: 7
attachment: 424
custom_css: 1
event: 260
flamingo_contact: 62
flamingo_inbound: 2311
give_forms: 11
give_log: 334
give_payment: 1177
give_pdf_template: 2
nav_menu_item: 45
oembed_cache: 22
page: 87
post: 42
product: 67
product_variation: 18
revision: 1854
shop_order: 327
shop_order_refund: 8
woocustomemails: 3
wpcf7_contact_form: 1
wpforms: 4

### Security ###

Secure connection (HTTPS): ✔
Hide errors from visitors: ✔

### Active Plugins (45) ###

Git Updater PRO: by Andy Fragen – 2.0.1
WP Activity Log for Yoast SEO: by WP White Security – 1.3.0
Advanced Custom Fields PRO: by WP Engine – 6.0.2
CAPTCHA 4WP: by WP White Security – 7.1.1
Akismet Anti-Spam: by Automattic – 5.0.1
Classic Editor: by WordPress Contributors – 1.6.2
Contact Form 7: by Takayuki Miyoshi – 5.6.3
Disable Comments: by WPDeveloper – 2.4.2
Flamingo: by Takayuki Miyoshi – 2.3
Git Updater: by Andy Fragen – 11.1.6
Give - Fee Recovery: by GiveWP – 1.9.8
Give - Form Field Manager: by GiveWP – 2.0.3
Give - MailChimp: by GiveWP – 1.5.0
Give - Manual Donations: by GiveWP – 1.6.0
Give - Recurring Donations: by GiveWP – 2.0.0
Give - Stripe Gateway: by GiveWP – 2.5.0
Give - Tributes: by GiveWP – 1.5.9
Give - Donation Plugin: by GiveWP – 2.22.2
Square Candy MailTrap SMTP: by Square Candy Design – 1.0.0
OOPSpam Anti-Spam: by OOPSpam – 1.1.27
Per-Form Confirmation Pages for GiveWP: by GiveWP – 1.1.1
PHPEnkoder: by Michael Greenberg – 1.15.1
Redirection: by John Godley – 5.3.4
Hide SEO Bloat: by SO WP – 3.14.8
Square Candy ACF Events: by Peter Wise – 1.7.0
Square Candy Better YouTube: by Peter Wise – 0.1
User Role Editor: by Vladimir Garagulya – 4.63.1
Woo Custom Emails Per Product: by Alex Mustin – 2.2.9
WooCommerce Stripe Gateway: by WooCommerce – 6.8.0
WooCommerce Name Your Price: by Kathy Darling – 3.3.9
PDF Invoices & Packing Slips for WooCommerce: by WP Overnight – 3.2.1
WooCommerce Product Fees: by Caleb Burks – 1.3.0
WooCommerce: by Automattic – 7.0.0
Wordfence Security: by Wordfence – 7.7.1
Yoast SEO: by Team Yoast – 19.7.1
WP Activity Log for WooCommerce: by WP White Security – 1.4.4
WPForce Logout: by Sanjeev Aryal – 1.5.0
WP Rollback: by GiveWP.com – 1.7.2
WP Activity Log for WPForms: by WP White Security – 1.2.1
WP Activity Log: by WP White Security – 4.4.2.1
WPAli: Woocommerce Order Builder: by ALI KHALLAD – 1.1.5
WPForms Form Locker: by WPForms – 2.2.0
WPForms Mailchimp: by WPForms – 2.2.0
WPForms: by WPForms – 1.7.7.2
WPS Hide Login: by WPServeur
NicolasKulka
wpformation – 1.9.6

### Inactive Plugins (14) ###

Advanced Custom Fields PRO: by Delicious Brains – 5.12.3
Custom Post Type UI: by WebDevStudios – 1.13.1
Disable Bloat for WordPress & WooCommerce: by Disable Bloat – 3.0.3
Favicon by RealFaviconGenerator: by Philippe Bernard – 1.3.25
FluentSMTP: by FluentSMTP & WPManageNinja Team – 2.2.0
Gallery Slideshow: by Jethin – 1.4.1
Malcure WP Malware Scanner & Firewall: by Malcure – 11.0
MP3-jPlayer: by Simon Ward – 2.7.3
Simple 301 Redirects: by WPDeveloper – 2.0.7
Square Candy jPlayer skin: by Peter Wise – 1.0
Woo Store Vacation: by Mahdi Yazdani – 1.4.6
WpAli: Woocommerce Order Builder: by ALI KHALLAD – 1.1.1
WP Mail SMTP: by WPForms – 3.5.2
WP Offload SES Lite: by Delicious Brains – 1.4.6

### Dropin Plugins (1) ###

maintenance.php: maintenance.php

### Settings ###

API Enabled: ✔
Force SSL: –
Currency: USD ($)
Currency Position: left
Thousand Separator: ,
Decimal Separator: .
Number of Decimals: 2
grouped (grouped)
simple (simple)
variable (variable)

featured (featured)
outofstock (outofstock)
rated-1 (rated-1)
rated-2 (rated-2)
rated-3 (rated-3)
rated-4 (rated-4)
rated-5 (rated-5)

Connected to WooCommerce.com: ✔
Enforce Approved Product Download Directories: –

### WC Pages ###

Shop base: #8 - /shop/
Cart: #9 - /cart/
Checkout: #10 - /checkout/
My account: #11 - /my-account/

### Theme ###

Name: storefront-child-theme
Version: 1.3.3
Author URL: http://squarecandydesign.com
Child Theme: ✔
Parent Theme Name: Storefront
Parent Theme Version: 4.1.3
Parent Theme Author URL: https://woocommerce.com/
WooCommerce Support: ✔

### Templates ###

Overrides: storefront-child-theme/woocommerce/content-single-product.php

### Admin ###

analytics
coupons
customer-effort-score-tracks
experimental-products-task
experimental-import-products-task
experimental-product-tour
shipping-smart-defaults
shipping-setting-tour
homescreen
marketing
multichannel-marketing
mobile-app-banner
navigation
onboarding
onboarding-tasks
remote-inbox-notifications
remote-free-extensions
payment-gateway-suggestions
shipping-label-banner
subscriptions
store-alerts
transient-notices
woo-mobile-welcome
wc-pay-promotion
wc-pay-welcome-page

Disabled Features: minified-js
new-product-management-experience
settings

Daily Cron: ✔ Next scheduled: 2022-10-19 21:24:38 -04:00
Options: ✔
Notes: 98
Onboarding: completed

### Action Scheduler ###

Complete: 202
Oldest: 2022-09-29 13:41:25 -0400
Newest: 2022-10-18 20:55:48 -0400

Pending: 9
Oldest: 2022-10-19 00:50:04 -0400
Newest: 2022-10-25 13:30:05 -0400

### Name Your Price ###

Version: 
Template overrides: –

Generated at: 2022-10-18 20:55:59 -04:00
`

### Isolating the problem

- [X] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/storefront/).

dwainm:
* Add a new function that checks if htaccess file exist and whether its contents allow or block direct url access to woocommerce_uploads directory.
* It also includes tests.

Closes #35166.

**Before checking out the branch replicate the issue:**
1. Checkout trunk
1. Be logged in as admin.
1. Make sure that `wp-content/uploads/woocommerce_uploads` contains a `.htaccess` file with the contents `deny from all`.
1. Edit one of your downloadable products or create a new one.
1. Under general product settings, add a file and make sure you upload a new image. This image will be added to the media library, but be stored in the woocommerce_uploads directory:
Screenshot 2025-02-13 at 14 06 51

1. After adding it, save the product and reload the page. 
1. Then under general product settings click on choose file, which should open the media library. You should now see a blank image for the product you uploaded. Attempting to access the image.
1. Edit the product thumbnail and add the image you uploaded as the new product thumbnail. The product image should also no longer load due to 403 forbidden error.

**Observe the fix after applying this branch**
1. Make sure you have the default WC product image set and that the id set is a valid media library image.  To set it go to WC setting > Product > General and look for replacement image.
1. Checkout this branch
1.  Go back to the product you just edited.
1. You should see the product thumbnail replaced by the placeholder image.
1. Under general product setting, clicking on choose file should bring up the media library, where you will see the image replaced by the placeholder image.

**Observe the fix for folks who do not have .htaccess restrictions** 
1. Delet the transient added in the code. The easiest way is to use a plugin like  [Transient Manager](https://wordpress.org/plugins/transients-manager/) and look for a transient with name matching `wc_htaccess_direct_access_....`. 
1. Rename the .htaccess in the woocommerce_uploads directory to tmp.
1. Relod the product page. 
1. The product thumbnail should be the image you uploaded with no error.
1. The downloadable product's choose file media library should also show the image you uploaded.   

### Changelog entry

<details>

<summary>Fix display errors for images added as product downloads.</summary>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 
Avoid display errors for images added as product downloads by showing the placeholder image instead for apache server configuration where access to woocommerce_uploads directly is not allowed.
</details>

<details>

</details>


barryhughes:
(Not a full review by any means 🙌🏼  this just caught my attention because I was subscribed to the original issue...) but might this be problematic on deployments where the server doesn't understand or support .htaccess files (Nginx) but where an .htaccess file is in place because of an earlier migration from a different server, or similar?

dwainm:
@barryhughes this doesn't change the .htaccess behaviour. It only shows the actual image (if we think it's all clear) or the the placeholder. If you wouldn't mind giving it another look to see if I've missed something?

barryhughes:
For sure, I realize it doesn't change anything about how `.htaccess` is or isn't interpreted (also, 'problematic' might not have been the right word to use). 

I meant more that the existence of an .htaccess file containing `Deny from all` doesn't necessarily mean access is denied (and its absence doesn't mean access is allowed, either, even if we *are* running under Apache 2.2).

barryhughes:
...I must be using the wrong search terms, because I can't seem to surface it, but I think we have a somewhat related issue that enumerates some of the possible challenges with this sort of test (such as, iirc, Apache doesn't necessarily always look at `.htaccess` files; Apache 2.4 has deprecated this syntax and introduced a different way to express this sort of rule; we may not be running under Apache or other server that supports `.htaccess` at all).

dwainm:
I'll take a look at the Apache docs. I think WC is outdated in that it actually creates this file, so if we need to alter the syntax, I'll incorporate it as well. At the moment, this PR provides a safer alternative that doesn't break a site (or cause errors that block people from accessing wp-admin). The worst-case scenario for the PR Is that you'll see the default WC image.

barryhughes:
> I think WC is outdated in that it actually creates this file

:100: ... exactly, which I'm pretty sure is the essence of the other issue I was struggling to find. Anyway, agree!

> The worst-case scenario for the PR Is that you'll see the default WC image.

Well, in some cases, wouldn't it still lead to the access blocking problem described in the linked issue? 

The current approach means we only think access is blocked when A) an Apache 2.2-style rule is in place and B) that rule lives within an `.htaccess` rule, but access *could* be blocked at a different level (say, `apache2.conf`), or in `.htaccess` but using Apache 2.4 syntax, or indeed some other method unrelated to Apache. 

In those cases, we'd still try and load the asset and that would potentially trigger fail2ban?

*(Edited this a couple of times to try and clarify my thoughts.)*

barryhughes:
On balance, across the userbase, this is still likely to represent a net improvement ... so defintely don't let my thoughts be a blocker there.

jorgeatorres (changes requested):
Hey @dwainm!

Total fan of the thought that went into this and all the tests. It does improve things for sites running under Apache and using our default .htaccess in `woocommerce_uploads/`, but I'm wondering if we could get into too much trouble by trying to be "too smart" here with detecting rules and stuff.

As @barryhughes pointed out, even if we focus on Apache alone, access to `woocommerce_uploads/` can still be prevented via other configuration files or even using a `.htaccess` inside that folder that has other rules beyond `deny from all`.
In fact, given `.htaccess` supports conditionals and other more complex syntax, I'd say that the presente of a `deny from all` in that file doesn't necessarily mean access has been blocked. Similarly, its absence doesn't imply it is allowed. Maybe it's restricted by IP or some other criteria.

If our plan is to let admins (or those with the `manage_woocommerce` capability) see thumbnails for images inside `woocommerce_uploads/` maybe we should instead serve them ourselves using a custom endpoint. That is, we could rewrite the `src` like we're doing right now but maybe the URL is instead `admin-ajax.php?action=our-endpoint&path=...&nonce=...` and we read the file and return it inside `our-endpoint`, obviously checking permissions, nonce, etc. As long as we limit this to those editing the product on the backend, this might be a good compromise even if it's slower than letting the web server take care of serving the file.

This in itself might be overkill (and I might be missing something that wouldn't let us do this), but given we don't have a way to accurately tell whether access is restricted or not by just reading a .htaccess file, in my opinion we should let things fail (as right now) or try to solve the issue for everyone (even if that means bypassing the .htaccess for admins).

What are your thoughts? Also curious about @barryhughes' if you're around.

barryhughes:
> Maybe we should instead serve them ourselves using a custom endpoint.

I like this idea. 

Digression, and really just a musing: can the logic from the download handler/force download system be re-used here, or extracted so both this endpoint *and* the download handler can use it?

One other option is testing if access is allowed via something like a `wp_remote_get()` call ... but that doesn't feel like it would be quite as reliable, and assets might appear accessible to us when, to the outside world, they actually are not (and would have performance implications that would be hard to fully address even with caching).

dwainm:
@barryhughes  and @jorgeatorres, I love the direction. Thank you for spending some time on this. I agree that depending on the content of `.htaccess` is not the best way forward. 

I like the idea of the src being dynamic. Let me revisit this with all that you've suggested.

dwainm:
@jorgeatorres  and @barryhughes , before cleaning it up and adding tests, etc, I wanted to see if you think the approach is ok. I tried an admin-only access approach, opting not to use existing download functionality to avoid complications and possible weaknesses (if that could be exploited somehow).

plugins/woocommerce/includes/class-wc-download-handler.php
```diff
@@ -26,6 +26,9 @@ public static function init() {
 		if ( isset( $_GET['download_file'], $_GET['order'] ) && ( isset( $_GET['email'] ) || isset( $_GET['uid'] ) ) ) { // WPCS: input var ok, CSRF ok.
 			add_action( 'init', array( __CLASS__, 'download_product' ) );
 		}
+
+		add_action( 'init', array( __CLASS__, 'serve_admin_image_src' ), 5 );
```
barryhughes: So I don't feel like the Download Handler is the right place for this, because the specific type of downloads this class is concerned with are *product downloads* (in the sense of digital assets a customer can purchase). Mixing in this logic—which is concerned with product thumbnails, as rendered in the admin environment—doesn't feel *quite* right, to me.

Hopefully my earlier note about reusing or sharing logic from this class didn't confuse things 😬 ... but in any case I would tend to place this somewhere else. It's possible we don't have a suitable existing home, but in that case we could perhaps start with a new class somewhere in the internal namespace (gives us more freedom to refactor later), or else we could define a true REST API endpoint.

plugins/woocommerce/includes/class-wc-download-handler.php
```diff
@@ -776,6 +779,102 @@ public static function track_download( $download, $user_id = null, $user_ip_addr
...
+
+		// Only allow image mime types
+		$mime_type = get_post_mime_type( $attachment_id );
+		$allowed_mime_types = [
+				'image/jpeg',
+				'image/jpg',
+				'image/png',
+				'image/gif',
+				'image/webp',
+				'image/svg+xml'
```
barryhughes: It's possible my knowledge is lagging behind recent changes, but I didn't think SVG was allowed by default, as it can pose certain security risks.

I understand support can be enabled, via custom code or plugins that solve this specific need, but I'm a little hesitant about hardcoding support for something that may not be supported in many cases 🤔 

- Do we need to verify the mime type ourselves? Possibly we do, but I figured it's worth asking the question.
- If we do, should we use `wp_get_image_mime()`?
- Or, should we perhaps keep this and perform an intersection of types that *we* allow, with types that *WordPress* allows?

```php
# We could combine our list of acceptable mimes with that of WordPress,
# and only allow through items that are on both lists.
$mimes_we_allow  = [ 'image/jpeg', /* ... */ ];
$mimes_wp_allows = get_allowed_mime_types();
$allowed_mimes   = array_intersect( $mimes_we_allow, $mimes_wp_allows );
```

plugins/woocommerce/includes/class-wc-download-handler.php
```diff
@@ -776,6 +779,102 @@ public static function track_download( $download, $user_id = null, $user_ip_addr
...
+		if ( ! wp_verify_nonce( $_GET['nonce'], 'admin-image-src-' . $attachment_id )
+				|| ! current_user_can( 'manage_woocommerce' )
+		   ) {
+			self::download_error( __( 'Invalid access token', 'woocommerce' ), '', 403 );
+		}
+
+		// Get file path
+		$file_path = get_attached_file( $attachment_id );
+		/* throw new Exception($file_path, 1); */
+		if ( ! $file_path || ! file_exists( $file_path ) ) {
```
barryhughes: Could consider swapping out `file_exists` with `is_readable`, which also checks if the file exists and further confirms it can be read.

```suggestion
		if ( ! $file_path || ! is_readable( $file_path ) ) {
```

barryhughes (commented):
Thanks for these changes!

I left a few comments, and I think the biggest question for me is: where should this logic live? 

I should also note that I didn't actually test this and decided not to get too fine-grained, just in case that question leads to further refactoring.

jorgeatorres (commented):
Hey @dwainm!

I like where this is heading 💯. Thanks for making these changes!
I agree with what @barryhughes has said. I think ideally these new code would live in a new internal class, which can both serve the file and alter the `img` tags on the admin as necessary.

I'd also suggest taking a look at how the code in `src/` is structured as it'd be best to use classes that can be instantiated (vs static classes). This is definitely not a blocker, but if you can make it work that way, it'd be useful for testing purposes.

Finally, I wonder if beyond the mime types restrictions Barry has mentioned, we should also add a restriction on the file size. Like, we possibly don't want to serve huge files this way, even if only on the admin. For those cases, I think we should return the placeholder (as usual).

What are your thoughts?

plugins/woocommerce/includes/class-wc-download-handler.php
dwainm: Thank you, I like the suggestion!

plugins/woocommerce/includes/class-wc-download-handler.php
dwainm: Thanks for that idea, I have not considered the security risk. I'll remove it.

It seems like we'll only need what WP allows, as it is a lot! https://developer.wordpress.org/reference/functions/wp_get_mime_types/ .  I'll alter the logic thoulg.  

If it's not an image type, we probably want to show the placeholder the WP provides.

dwainm: > Do we need to verify the mime type ourselves? Possibly we do, but I figured it's worth asking the question.

I think we do since we're actually generating a preview image. I'll look into what we can show for other types. Since we're thinking about moving this outside of the downloads handler it might make sense to handle all the mime/display logic as well (I know there's probably a WP function that gives us this for free, so I'll see what I can find).

plugins/woocommerce/includes/class-wc-download-handler.php
dwainm: Good point Barry, I'll see what I can find. My initial thought was the download handler seems fitting, but I understand what you mean by the purpose of this being unrelated to user downloads and more focused on admin download previews/viewing.

dwainm:
Thank you @jorgeatorres 

> I'd also suggest taking a look at how the code in src/ is structured as it'd be best to use classes that can be instantiated (vs static classes). This is definitely not a blocker, but if you can make it work that way, it'd be useful for testing purposes.

I'll do this, thanks for the hint.

> Finally, I wonder if beyond the mime types restrictions Barry has mentioned, we should also add a restriction on the file size. Like, we possibly don't want to serve huge files this way, even if only on the admin. For those cases, I think we should return the placeholder (as usual).

Great idea, I haven't considered that full for downloads could potentially be huge. I think that we should limit the types we allow for these items to be at max large which is (1024x1024).  Do you agree?

jorgeatorres:
> to be at max large which is (1024x1024). Do you agree?

I was just thinking we look at the actual file size before trying to serve. Maybe we don't try to serve files if they are > 20 MB or something like that, unless (of course) we already have a thumbnail of a reasonable size available.

dwainm:
> unless we already have a thumbnail of a reasonable size available.

I checked the uploads folder, and for images it seems like wp media library creates reasonable sized images based on predefined sizes. So we can assume that the thumbnail generated would be 300x300 and that large would be 1024x1024. 
See screenshot below:
Screenshot 2025-02-26 at 12 50 55

jorgeatorres:
Yeah, so as long as we have the thumbnails, that's fine. I was just thinking that if we don't have the thumbnail available we might still want to serve the original file, but we should just do that when the file size itself is reasonable.

plugins/woocommerce/includes/class-wc-download-handler.php
dwainm: Barry, while adding wp mime types, I realise that we're specifically adding the endpoint for image files [`serve_admin_image_src`]. We don't want to handle any other files. In that case I'm not pro adding the wp mime list as it includes way too many things.

dwainm:
@jorgeatorres I still have some tweaking to do, but wanted to confirm if you agree with the direction of plugins/woocommerce/src/Internal/ProductDownloads/AdminPreview.php

jorgeatorres:
> @jorgeatorres I still have some tweaking to do, but wanted to confirm if you agree with the direction of plugins/woocommerce/src/Internal/ProductDownloads/AdminPreview.php

It makes sense, but given (at least right now) we don't plan to make this available to anyone outside the admin, we might as well store it straight inside`Internal/Admin/ProductDownloadsPreview.php` or similar. What do you think?

In any case, this is the Internal namespace so we have some flexibility to move things around as needed, so I wouldn't worry too much about it.

dwainm:
That makes sense, I'll update it. Besides that, do you think this is good to be polished and set for final review?

jorgeatorres:
> That makes sense, I'll update it. Besides that, do you think this is good to be polished and set for final review?

There are some linting issues that I'm not sure why CI didn't pick up (maybe because this PR is still a draft?). If you could take a look at those, I think we're good to move from draft state. I do have some feedback, but I'd prefer to review the final thing (once you make the adjustments) if that's ok with you.

Again, thanks for all the thought and hard work that's going into this @dwainm!

dwainm:
@jorgeatorres , I still have the idea of a dedicated endpoint at the back of my mind, but wanted to check in to see if you're aligned with the direction and if you think that's something we'll definitely need?  Also any other feedback on the implementation would be appreciated.

dwainm:
Linting hasn't run on the open PR? I tried running `./vendor/bin/phpcs` locally, but it runs slow and requires heaps of memory.

plugins/woocommerce/includes/wc-product-functions.php
```diff
@@ -607,64 +608,72 @@ function wc_scheduled_sales() {
...
-	/*
-	 * If the user does not have the right capabilities,
-	 * filter out the image source and replace with placeholder
-	 * image.
-	 */
-	if ( isset( $response['url'] ) && strstr( $response['url'], 'woocommerce_uploads/' ) ) {
-		$response['full']['url'] = wc_placeholder_img_src();
+    // If not a WooCommerce manager, return placeholder for all image sizse.
+    if ( ! current_user_can( 'manage_woocommerce' ) ) {
+		$response['full']['url'] = wc_placeholder_img_src( $size );
```
barryhughes: At this point, I don't think `$size` has been defined.

plugins/woocommerce/src/Internal/Admin/ProductDownloadsPreview.php
```diff
@@ -0,0 +1,240 @@
...
+	 */
+	private function clean_buffers() {
+		if ( ob_get_level() ) {
+			$levels = ob_get_level();
+			for ( $i = 0; $i < $levels; $i++ ) {
+				@ob_end_clean(); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
+			}
+		} else {
+			@ob_end_clean(); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
+		}
```
barryhughes: Do we need the `else` clause? If there are no buffers (if `ob_get_level()` returns `(int) 0`, which I think is the only condition that will lead us into this path), there is nothing to clean.

plugins/woocommerce/tests/php/src/Internal/Admin/ProductDownloadsPreviewTest.php
```diff
@@ -0,0 +1,237 @@
...
+		$request->set_param( 'token', $token );
+		$request->set_param( 'product_id', $this->product_id );
+		$request->set_param( 'attachment_id', $this->attachment_id );
+		
+		$response = $this->preview->get_preview_permissions_check( $request );
+		
+		$this->assertTrue( $response );
+		
+		// Verify token was deleted after use
+		$this->assertFalse( get_transient( 'wc_preview_token_' . $token ) );
```
barryhughes: Feel free to disregard ... this is a bit of a quibble. We're testing an implementation detail here (ie, the transient used to store the one-time token has been deleted). 

Could demonstrating that a re-run of the exact same request won't succeed twice be better, or a worthwhile supplement to this test? In my mind, that describes the outcome a little more clearly.

plugins/woocommerce/changelog/55442-fix-35166-error-downloadable-product-image-viewing
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Avoid display errors for images added as product downloads by showing the placeholder image instead for apache server configuration where access to woocommerce_uploads directly is not allowed.
```
barryhughes, line 4: Optional thing/quibble: the underlying problem is not exclusive to Apache.

```suggestion
Improve handling of placeholder images for product downloads in cases where server restrictions prevent direct access to the relevant files.
```

barryhughes (commented):
Looks good! Left a couple of small notes, mostly optional/minor things. Tests well.

plugins/woocommerce/src/Internal/Admin/ProductDownloadsPreview.php
```diff
@@ -0,0 +1,240 @@
...
+			$stored['product_id'] != $request->get_param( 'product_id' ) ) {
+			return new WP_Error(
+				'woocommerce_rest_token_mismatch',
+				__( 'Token does not match requested resource.', 'woocommerce' ),
+				array( 'status' => 403 )
+			);
+		}
+		
+		// Delete the transient to prevent reuse.
+		delete_transient( $transient_key );
```
barryhughes: In general, it seems like every time we generate a one-time token and store it in a transient, there will be a matching request in which this line executes and the transient is removed.

Yet ... is it possible we could enter a situation where transients start accruing and not being cleaned up? What if, for some unrelated reason (or because of a future lazy-loading enhancement to WordPress), multiple thumbnail images URLs (and therefore tokens) are generated, but the images themselves are not actually fetched.

For cases where the transients are backed by the options table, that could lead to a big build-up of cruft (the transients will expire, but won't automatically be removed).

plugins/woocommerce/includes/wc-product-functions.php
dwainm: Good catch! Updated in a59e9daf2f

plugins/woocommerce/changelog/55442-fix-35166-error-downloadable-product-image-viewing
dwainm: 100% I agree, we serve all files from WooCommerce uploads in this way irrespective of restrictions and limit access to them for everyone besides admin.

plugins/woocommerce/src/Internal/Admin/ProductDownloadsPreview.php
dwainm: Good point, I had another idea to use the cache API, what do you think of the implementation 4d02c98a6f?

dwainm: https://github.com/woocommerce/woocommerce/pull/55442/commits/4d02c98a6f930003bb9e3f68133e8de2bd596928

plugins/woocommerce/src/Internal/Admin/ProductDownloadsPreview.php
dwainm: No need for all the complications here, our function is the only one on that endpoint and so I've opted for something much simpler in: fe79a6c3c3af5a873be8857084169d4dc3b49cc1

plugins/woocommerce/tests/php/src/Internal/Admin/ProductDownloadsPreviewTest.php
dwainm: I agree with you, no need to test that the cache/transient was stored. I'll simplify the testing in a sec.

dwainm: Removed in f75bf08010

dwainm: I've also added the new tests to make sure we're not able to get the same URL more than once in 62a4aae87d .

plugins/woocommerce/src/Internal/Admin/ProductDownloadsPreview.php
barryhughes: I'm not sure it's working *quite* as intended (try testing as a 'full' admin user, refreshing the media library a couple of times—disable your browser's cache, or clear each time—and with a persistent cache like Memcached or Redis enabled).

* When a persistent cache is ***not*** available, anything we store via `wp_cache_*` is lost at the end of each request. In this scenario, the [signatures won't be verified](https://github.com/woocommerce/woocommerce/blob/314acc39de849a4d0d6ddaede838eed221a615f8/plugins/woocommerce/src/Internal/Admin/ProductDownloadsPreview.php#L104-L112) (that is, it will always appear as if the signature has not been used, because the list of used signatures will always be empty at this point).
* When a persistent cache ***is*** available, we have a slightly different problem: we deliver the image, then store the used signature in the cache. However, for any given image [the signature will always be the same](https://github.com/woocommerce/woocommerce/blob/314acc39de849a4d0d6ddaede838eed221a615f8/plugins/woocommerce/src/Internal/Admin/ProductDownloadsPreview.php#L115-L116) (because the attachment ID, product ID, key and salt are constants). We never clear the list of used signatures, though, and so on subsequent refreshes of the media library [we refuse to supply the image](https://github.com/woocommerce/woocommerce/blob/314acc39de849a4d0d6ddaede838eed221a615f8/plugins/woocommerce/src/Internal/Admin/ProductDownloadsPreview.php#L107-L111).

Instead of implementing a signing process, what if we just verify that the current user has the necessary permissions to view the asset?

dwainm: I'll take a look. The issue was with the current user not being available on the rest endpoint. There's a way to handle this, but we're not calling the endpoint ourselves and so can't pass the noncne header as per the documentation: https://developer.wordpress.org/rest-api/using-the-rest-api/authentication/

barryhughes: It can also be passed as a query param (as an alternative to setting a header), so possibly we could update the `add_query_arg()` call near the end of `ProductDownloadsPreview::get_admin_image_src_url()`, something like:

```php
$url = add_query_arg(
	array(
		'size'      => $size,
		'signature' => $signature,
		'_wpnonce'  => wp_create_nonce( 'wp_rest' ),
	),
	$url
);
```

Untested, also not sure how best to document this behavior (in the sense that it is a different authentication model compared to most other `/wc/v3/*` endpoints), but it's a possible avenue.

dwainm: 🤦 Ashamedly I didn't notice the GET and POST query sentence in that post. Thanks for the great suggestion. I've simplified it quite a bit here: 0791115c8db81b4ae18dabeb75a38677281c746c

plugins/woocommerce/tests/php/src/Internal/Admin/ProductDownloadsPreviewTest.php
dwainm: And then I removed it again as I've opted to use a simpler mechanism for access control, so no need to worry about access counts. Admin can access it as many times as they want.

plugins/woocommerce/src/Internal/Admin/ProductDownloadsPreview.php
```diff
@@ -0,0 +1,182 @@
...
+
+		nocache_headers();
+		header( 'Content-Type: ' . $mime_type );
+		header( 'Content-Length: ' . filesize( $file_path ) );
+		header( 'Content-Disposition: inline; filename="' . basename( $file_path ) . '"' );
+
+		// We need to use readfile here for binary file output.
+		// phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_operations_readfile
+		readfile( $file_path );
+		exit;
```
barryhughes: Open-ended "philosophical" question ... we're operating inside the framework of the WP REST API here and, normally, a REST API controller does not terminate execution in this way, but instead returns a `WP_REST_Response` (or else a `WP_Error`, etc).

I understand why we aren't doing that here (performance), and that's probably fine, but I also wonder if this could cause problems—not necessarily breakages, but perhaps confusion—since it means the usual dispatch hooks in `WP_REST_Server` that developers would expect to fire and to have access to will not actually be triggered 🤔 

Again, I don't have any concrete ideas as to what problems this might cause (perhaps/hopefully none), but wanted to pause and consider it.

plugins/woocommerce/src/Internal/Admin/ProductDownloadsPreview.php
```diff
@@ -0,0 +1,182 @@
...
+		if ( $resized && isset( $resized['path'] ) ) {
+			$uploads_dir       = wp_upload_dir();
+			$resized_file_path = $uploads_dir['basedir'] . '/' . $resized['path'];
+			if ( is_readable( $resized_file_path ) ) {
+				$file_path = $resized_file_path;
+			}
+		}
+
+		// Clean any existing output.
+		ob_clean();
```
barryhughes: This is probably fine 95% of the time, but will only clean a single level of output bufferring. I think there was a more complete version in an earlier iteration of the PR: could it be worth revisiting that?

plugins/woocommerce/src/Internal/Admin/ProductDownloadsPreview.php
```diff
@@ -0,0 +1,182 @@
...
+	/**
+	 * Get secure URL for admin image
+	 *
+	 * @since 9.9.0
+	 * @param int    $product_id    Product ID.
+	 * @param int    $attachment_id Attachment ID.
+	 * @param string $size          Image size.
+	 * @return string Secure admin image URL.
+	 */
+	public function get_admin_image_src_url( $product_id, $attachment_id, $size ) {
```
barryhughes: There are lots of spots like this where type-hints could be handy. Or else, we could explicitly cast to `int` (or otherwise sanitize) before placing the values in the `$url` string (just as a security-in-depth measure).

barryhughes (commented):
Left a few other notes, but initial testing looks good :+1:

---

Screenshots from testing—first and third images are thumbs for downloadable files—showing the placeholders (for a non-admin user) and REST API-served thumbs (for a merchant):

<div align="center"></div>

<div align="center"></div>

plugins/woocommerce/src/Internal/Admin/ProductDownloadsPreview.php
dwainm: Good point, I've updated it here: 6ef4e7be4f0c94921c7665d47f265ba7862ac091

dwainm: I've simplified it a bit from the previous solution.

plugins/woocommerce/src/Internal/Admin/ProductDownloadsPreview.php
dwainm: I've updatet he code to add type hints and a few explicit casts here: c424ff1170c260bed911d5f49572028b21536ded

plugins/woocommerce/src/Internal/Admin/ProductDownloadsPreview.php
dwainm: I thought about it for a bit and appreciate you raising it.  There was a more rest friendly way to do it. Please see 4d74324017 .

jorgeatorres:
@dwainm: I have it on my list to review this tomorrow, but one thing I wanted to ask (and also curious about @barryhughes' thoughts) is if instead of a new REST endpoint we should just let this be [an authenticated AJAX action](https://developer.wordpress.org/reference/hooks/wp_ajax_action/). Behavior is pretty much the same, but going with REST makes it look like an official API or something, and maybe adds a little bit of overhead for no clear benefit.

barryhughes:
> If instead of a new REST endpoint we should just let this be [an authenticated AJAX action](https://developer.wordpress.org/reference/hooks/wp_ajax_action/). Behavior is pretty much the same, but going with REST makes it look like an official API or something.

Right ... well, I'm not sure what's best, but was wrestling with a similar thought. 

A further alternative (besides moving to an authenticated ajax action) could be moving it out of the `/wc/<version>` space and over to something like `/wc/asset/*`. I'm somewhat basing that on the approach used for transient files, which use `/wc/file/*` ... and I think that was the result of a similar set of considerations. 

(In case of any doubt, the above reply is just some light brainstorming, I'm not advocating strongly for `/wc/file/*` ... an ajax endpoint or similar may well be better 🙃 )

dwainm:
Hey @jorgeatorres  and @barryhughes . I like your thinking. So many things I didn't even consider!  I agree that we shouldn't put this in the versioned namespace as if it were something public facing. I'll take a look at wc/assests and wc/files and consider what makes most sense for the implementation.

dwainm:
Hey @jorgeatorres I'm probably not going to have the capacity to complete this, so fell free to close it if it needs further changes.

Issue #54912, opened 2025-01-28
Fix an issue where interactivity store notices in TT4 display too wide.


samueljseay:
From the comment here: https://github.com/woocommerce/woocommerce/pull/54552#issuecomment-2618405987

With the new implementation of store notices we're working on in https://github.com/woocommerce/woocommerce/issues/54259 there is a styling issue in TT4 where the injected notice is too wide. We need to address that before completing the migration.

Screenshot from the comment for convenience:
<img width="811" alt="Image" src="https://github.com/user-attachments/assets/b2f57779-5774-4206-85f5-768c89ca3bcb" />

samueljseay:
We don't need to worry about this now since we're going to display a notice within prod collection

Issue #54907, opened 2025-01-28
Fix issues with using block hooks API to inject woocommerce/store-notices block


samueljseay:
Before we merge the final implementation of interactivity powered store notices, there are some issues to test and resolve.

@gigitux outlined them here: https://github.com/woocommerce/woocommerce/pull/54552#issuecomment-2618405987

We will test and try to address these problems before we merge the work being done in https://github.com/woocommerce/woocommerce/issues/54259

samueljseay:
We don't need to do this now since we're not using block hooks any more for this.

PR #55452, opened 2025-02-13
Store Notices: Cache whether the store notices block has already been hooked

tjcafferkey:

Added a static boolean flag to prevent recursive rendering of store notices blocks in full-site editing mode. When Gutenberg rendered pages in full-site editing mode, store notices are added multiple times due to both template and post content conditions evaluating to true simultaneously.

There are outstanding issues with this functionality, as can be seen mentioned [here](https://github.com/woocommerce/woocommerce/pull/54552#issuecomment-2618655397). But follow up issues have been created to address those separately https://github.com/woocommerce/woocommerce/issues/54907 and https://github.com/woocommerce/woocommerce/issues/54912

1. Pull latest Gutenberg `trunk`, build and active plugin.
2. Load any Page and ensure that the Store Notices block has been added (happens via the Block Hooks API) and the rest of the page in the Editor loads correctly.
3. Check you don't see a "Block cannot be rendered inside itself" error.
4. Disable the Gutenberg plugin and retest

> [!IMPORTANT]  
> It's OK CI fails on this PR with the following problems:
> - Lint error related to Cart -> Cart is being handled separately
> - E2E tests failing related to Cart -> Cart is being fixed spearately
> - E2E test failing related to Mini Cart: `Shopper → Notices -> Shopper can add item to cart, and will not see a notice in the mini cart` - It's being fixed in separate PR: https://github.com/woocommerce/woocommerce/pull/55295
>
> This PR targets another feature branch and not `trunk`!

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


gigitux:
Thanks for working on this! I followed the same reproduction steps that I followed with #54552. These are my findings:


# With Gutenberg

### 🔴  Post - TT5 - Regression


https://github.com/user-attachments/assets/8cbae208-c714-4d05-b274-5399d1c0cac2

The notice isn't rendered anymore.


### 🔴  Page - TT5 - Regression


https://github.com/user-attachments/assets/67fdb267-9aa4-498c-be9f-8bbaf07c1c62.

The notice isn't rendered anymore

### 🟢 Product Catalog - TT5

image

Everything works as expected

# Without Gutenberg

### 🔴  Post - TT5 - Regression


https://github.com/user-attachments/assets/8cbae208-c714-4d05-b274-5399d1c0cac2

The notice isn't rendered anymore.


### 🔴  Page - TT5 - Regression


https://github.com/user-attachments/assets/67fdb267-9aa4-498c-be9f-8bbaf07c1c62.

The notice isn't rendered anymore.

### 🟢 Product Catalog - TT5

image

Everything works as expected.

tjcafferkey:
Thanks @gigitux. I have updated the condition checking if hooked blocks in post_content is supported as I believe that was causing one of the primary problems here, but that's not the only issue.

I'm just going to pull out a private message from @samueljseay to provide some context for this next bit

> i was also trying to avoid it showing on any template where product collection is not present so if it shows in post content in the editor and there is no product collection its a regression.

This suggests we want to be hooking the Store Notices block everywhere that the Product Collection block is. This works fine with default content (e.g. Product Catalog template) but when a user creates a new template/post/page and inserts the Product Collection block and then saves the entity, the algorithm for `ignoredHookedBlocks` is run upon saving the content to respect the users design choices and prevent any existing hooked blocks being added to anchor blocks. **As a result, in this scenario the Store Notices block will be prevented from being inserted.**

With that in mind I'm thinking that with the behavior we're trying to achieve, that using the Block Hooks API might not be the appropriate solution. I'm going to pause work on this PR until @samueljseay is back from AFK.

samueljseay:
As discussed with @tjcafferkey we agreed we don't need to use block hooks now. I have this draft PR up for a different approach: https://github.com/woocommerce/woocommerce/pull/55717 which I will close this in favor of.

Issue #55046, opened 2025-01-31
[Performance]: Remove unused/unnecessary wcSettings


moon0326:
### Prerequisites

- [x] I have carried out troubleshooting steps and I believe I have identified a performance issue.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Performance Details

We should check if we can improve TTFB by cleaning up the wcSettings values. These values are loaded on the server side and might slow down TTFB for some pages.

Examples:

1. `wcSettings.countries` and `wcSettigns.countryStates` are not used at all on WooCommerce -> Home and other WC admin pages.
2. `wcSettings.isLocale` and `wcSettings.admin.isLocale` are identical. We can remove one of them. 
3. We can possibly replace `wcSettings.admin.marketing.installedExtensions` with `wcSettings.admin.plugins.installedPlugins`. We also need to check if the value is used at all. `/wp-admin/admin.php?page=wc-admin&path=%2Fmarketing` already makes a request to `wp-json/wc-admin/plugins/installed?_locale=user` on page load.
4. `wcSettings.admin.onboarding` can probably be removed. We make Ajax calls on the new core profiler.

These are just a few examples. We should go over all of them and clean them up and compare FTTB. 

### Affected Sites

all

adrianduffell:
Is it correct we don't advise extensions to use these objects directly -- instead they should use data stores? We probably should create a dev advisory just in case, and guide devs to use the correct approach.

joshuatf:
Assigning this high priority as it affects all wc-admin routes and connected pages.  On a larger site, collectively the `woocommerce_admin_shared_settings` filters to 591ms.

joshuatf:
That's a good question, @adrianduffell.  I think our messaging in the docs is a little conflicted.  We do mention the use of `getSetting`, which I believe is the preferred method:

```
import { getSetting } from '@woocommerce/settings';
const notifyLowStockAmount = getSetting( 'notifyLowStockAmount', 2 );
```

But in [some docs](https://github.com/woocommerce/woocommerce/blob/trunk/docs/reporting/extending-woocommerce-admin-reports.md), we suggest directly using the `wcSettings` object.

adrianduffell:
> But in [some docs](https://github.com/woocommerce/woocommerce/blob/trunk/docs/reporting/extending-woocommerce-admin-reports.md), we suggest directly using the wcSettings object.

Ah, that's too bad 😅 

[...]

PR #55464, opened 2025-02-13
Remove wcSettings.admin.onboarding.currencySymbols

### LINKED ISSUE #55346: Clean up component setting values set by OnboardingSetupWizard.php
Location https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Internal/Admin/Onboarding/OnboardingSetupWizard.php#L44 and https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Internal/Admin/Onboarding/OnboardingSetupWizard.php#L46

OnboardingSetupWizard.php adds the following values to `wcSettings.admin.onboarding`, but most of them are either unused or can be replaced with a REST API call.

**onboarding.currencySymbols**: used in the [core profiler](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/admin/client/core-profiler/index.tsx#L454) and [activity panel](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/admin/client/homescreen/activity-panel/orders/index.js#L244). See if we can load them conditionally in the frontend.

**onboarding.wccom_connected**: Not used. Confirm the usage and remove it.
**onboarding.euCountries**: Used in [getCurrencyRegion](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/admin/client/dashboard/utils.ts#L32), but it seems that the function is not used anywhere. Confirm the usage and remove the value.

**onboarding.localeInfo**: Used in the core profiler. This is whopping 20K+ lines. We can probably load it via REST API. 
**onboarding.profile**: Not used, but double-check. 
**onboarding.pageCount**: Not used.
**onboarding.postCount**: Not used.
**onboarding.isBlockTheme**: Not used.

moon0326:

Partially closes #55346 

This PR removes `wcSettings.admin.onboarding.currencySymbols`

- Added `getCurrencies` function in the countries data store. 
- Removed `wccom_connected` -- this is not used anywhere.
- Removed the value in PHP
- Updated JS to use `getCurrencies` to get the currency data. 
- Updated `/wc/v3/data/currencies` endpoint to accept `incldue_extra` to return currency_position, thousand_separator, decimal_separator, and number_of_decimals fields.

1. Create a new JN site using this branch with Smooth Generator.
2. Navigate to `Tools -> Smooth Generator`. Create 10 products and orders.
3. Navigate to WooCommerce -> Home
4. Hide the main tasklist.
5. You should see "Orders" and "Stocks" 
6. Click on "Orders" and verify that the currency symbols are displayed correctly.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Remove use of `wcSettings.admin.onboarding.currencySymbols`

</details>

<details>

#### Comment 

</details>


packages/js/data/src/countries/resolvers.ts
```diff
@@ -58,6 +60,24 @@ export function* getCountries() {
...
+	try {
+		const url = 'wc/v3/data/currencies';
+		const results = yield apiFetch( {
+			path: url,
+			method: 'GET',
+		} );
+
+		return getCurrencySymbolsSuccess(
+			results.map(
+				( currency ) => `${ currency.code }: ${ currency.symbol }`
```
chihsuan: I set the store currency to NTD (New Taiwan Dollar), but the currency symbol is not displayed correctly. I believe this is because we store the currency symbols as an array, not an object.

I think we should convert it to match the same structure as the `wcSettings.admin.onboarding.currencySymbols`.

chihsuan: https://github.com/user-attachments/assets/915c1855-97cf-40af-a6b0-270e84b59bd4

plugins/woocommerce/client/admin/client/homescreen/activity-panel/orders/index.js
```diff
@@ -281,6 +261,9 @@ function OrdersPanel( { unreadOrdersCount, orderStatuses } ) {
 			return { isRequesting: false };
 		}
 
+		/* eslint-disable @wordpress/no-unused-vars-before-return */
+		const { getCurrencySymbols } = select( COUNTRIES_STORE_NAME );
```
chihsuan: I think you could move this line a little bit later after the first return statement so we don't need to use `/* eslint-disable @wordpress/no-unused-vars-before-return */`.  🙂

packages/js/data/src/countries/resolvers.ts
moon0326: Good catch! Fixed it in [Correct currencySymbol format](https://github.com/woocommerce/woocommerce/pull/55464/commits/fabfeb631e654185c585d3f6ae794ac839353700)

plugins/woocommerce/client/admin/client/homescreen/activity-panel/orders/index.js
moon0326: Thank you! Fixed it in [Fix lint issue](https://github.com/woocommerce/woocommerce/pull/55464/commits/28c49adeefaa63d3f1d34ba31279a06289737cc7)

chihsuan (dismissed):
Thanks, LGTM and tested well! 🚢 

Screenshot 2025-02-17 at 10 57 34

plugins/woocommerce/src/Internal/Admin/Onboarding/OnboardingSetupWizard.php
```diff
@@ -223,14 +223,9 @@ public function component_settings( $settings ) {
 			return $settings;
 		}
 
-		include_once WC_ABSPATH . 'includes/admin/helper/class-wc-helper-options.php';
-		$wccom_auth                 = \WC_Helper_Options::get( 'auth' );
```
rjchow, line 227: Might have missed this removal in the PR description / testing instructions

plugins/woocommerce/client/admin/client/homescreen/activity-panel/orders/index.js
```diff
@@ -17,7 +17,11 @@ import {
 } from '@woocommerce/components';
 import { getNewPath } from '@woocommerce/navigation';
 import { getAdminLink } from '@woocommerce/settings';
-import { ordersStore, ITEMS_STORE_NAME } from '@woocommerce/data';
+import {
+	ORDERS_STORE_NAME,
```
rjchow: I both  ORDERS_STORE_NAME and COUNTRIES_STORE_NAME should have been migrated to storeDescriptor format so this PR is accidentally reverting that

rjchow (commented):
Some preliminary feedback first, sorry bout the delay

plugins/woocommerce/src/Internal/Admin/Onboarding/OnboardingSetupWizard.php
moon0326: @rjchow Updated!

plugins/woocommerce/client/admin/client/homescreen/activity-panel/orders/index.js
moon0326: @rjchow Thanks! 

I don't think we have countriesStore yet. (https://github.com/woocommerce/woocommerce/blob/trunk/packages/js/data/src/index.ts#L50)

I was tempted to update it in this PR, but I figured we should update it in a separate PR.

Updated ordersStore in [Use ordersStore](https://github.com/woocommerce/woocommerce/pull/55464/commits/b3b88433f1a0bb96745631031577ad83131804af)

plugins/woocommerce/client/admin/client/homescreen/activity-panel/orders/index.js
```diff
@@ -313,9 +295,35 @@ function OrdersPanel( { unreadOrdersCount, orderStatuses } ) {
...
+
+		if ( ! symbol ) {
+			// This should never happen, but if it does, we'll just show the currency code.
+			return `${ orderCurrencyCode }${ total }`;
+		}
+
+		// If the order currency is different from the store currency, we show the currency code and amount in the order currency.
+		return CurrencyFactory( {
+			...storeCurrency,
+			symbol: decodeEntities( symbol ),
```
rjchow: I understand that this is just the existing code, but I think it would be a good improvement to include the full currency settings here instead of only the symbol - for example if the order currency is set to USD and the store config is set to JPY, the currency symbol appears on the wrong side:

image

plugins/woocommerce/src/Internal/Admin/Onboarding/OnboardingSetupWizard.php
```diff
@@ -223,14 +223,9 @@ public function component_settings( $settings ) {
...
-		$profile['wccom_connected'] = empty( $wccom_auth['access_token'] ) ? false : true;
-
-		$settings['onboarding']['currencySymbols'] = get_woocommerce_currency_symbols();
```
rjchow, line 230: Probably need to rebase on the other branch because this removal breaks the store onboarding so I have to manually set the currency settings to test this PR

rjchow (changes requested):
Left a suggestion for an improvement that's easier to make now than coming back to it later. Otherwise it tests as described, aside from the core profiler country setting not working due to some stale code that removes the locale info (from the other PR?) but not including the update to not rely on it.

plugins/woocommerce/client/admin/client/homescreen/activity-panel/orders/index.js
moon0326: @rjchow That's a good find. Fix it in [Support include_extra. When include_extra is set, the endpoint re…](https://github.com/woocommerce/woocommerce/pull/55464/commits/332baa8e4585561a955d0a554701102cba65edce) and [Rename getCurrencySymbols to getCurrencies](https://github.com/woocommerce/woocommerce/pull/55464/commits/247d34fbe895298c33f969d808fb25346832a575)

plugins/woocommerce/src/Internal/Admin/Onboarding/OnboardingSetupWizard.php
moon0326: @rjchow Do you mean that this PR doesn't work unless the value is set manually? I tested it on a JN site and didn't encounter any errors. 🤔

rjchow: As in setting the country in core profiler doesn’t set the currency etc, on a JN live branch build - so I tested it by setting them via the wc settings page

adrianduffell:
@moon0326 , could we place the following item into its own PR? It seems like a separate thing needing dedicated testing for it.

_Removed wccom_connected -- this is not used anywhere._

Generally, I also think we need to put property removal through a deprecation process first to help any 3PD usage be migrated. I mention a process in the [parent issue](https://github.com/woocommerce/woocommerce/issues/55046#issuecomment-2641741369).

plugins/woocommerce/src/Internal/Admin/Onboarding/OnboardingSetupWizard.php
moon0326: @rjchow I see what you mean now. I'll rebase.

moon0326:
> Generally, I also think we need to put property removal through a deprecation process first to help any 3PD usage be migrated. I mention a process in the [parent issue](https://github.com/woocommerce/woocommerce/issues/55046#issuecomment-2641741369).

That's a good idea, but it means we can't remove these values yet. Do you want to keep them for now and add a deprecation warning?

I've searched for the usage of these values on https://wpdirectory.net/, and only WooCommerce is using them.

packages/js/data/src/countries/resolvers.ts
```diff
@@ -58,6 +60,49 @@ export function* getCountries() {
...
+			method: 'GET',
+		} );
+
+		return getCurrenciesSuccess(
+			results.reduce( ( acc, currency ) => {
+				acc[ currency.code ] = {
+					code: currency.code,
+					symbol: currency.symbol,
+					symbolPosition: currency.currency_position,
+					decimalSeparator: currency.number_of_decimals,
```
rjchow:
```suggestion
					decimalSeparator: currency.decimal_separator,
```

image

rjchow (changes requested):
Core profiler bit is working now, but something seems not quite right with the new added code for the full currency formatting

moon0326:
Converting this PR to draft for now. We're going to display a deprecation warning before removing these properties. See https://github.com/woocommerce/woocommerce/pull/55706 f

PR #55467, opened 2025-02-14
[Settings] Add Legacy save API

### LINKED ISSUE #54867: [Settings] Enable Saving

### LINKED ISSUE #55530: [Settings] Tax page should not be visible if tax is disabled
### Prerequisites

### Describe the bug

Currently, the tax page is visible when the tax feature is disabled. When visiting the tax page, the blank content is shown.

We can use the `woocommerce_settings_tabs_array` filter to check whether the tax page should be visible.

### Expected behavior

We should hide the tax page from navigation if the tax is disabled, like the classic settings page.

### Actual behavior

 the tax page is visible when the tax feature is disabled. 

### Steps to reproduce

2. Enable `settings` feature flag
3. Turn on the Settings feature flag and go to WooCommerce > Settings 

### WordPress Environment

n/a

### Isolating the problem

psealock:

Closes https://github.com/woocommerce/woocommerce/issues/54867
Closes https://github.com/woocommerce/woocommerce/issues/55530

Enable the Save button on Settings.

* Create new Settings endpoint for saving.
* Use DataProvider context for keeping track of data.
* Add `<Notices />` to display core notices.

### Key Decisions made

1. Avoid a data store on the client in favor of using the [Context api](https://react.dev/reference/react/useContext) to create a Data Provider. I choose using the context because the legacy settings are powered by a large complex singular piece of data. Instead of manipulating pieces here and there, the entire object is returned from the api and replaced wholesale.
2. Create a new Admin API endpoint specific to the new Settings. Other possibilities include using the WooCommerce API settings endpoint or the existing Options endpoint. I decided on this path because it was easiest to call `WC_Admin_Settings::save();` which is the method employed by the traditional POST request we have on current screens.
3. Generally keep implementation as simple as possible without making design decisions. This will all come later. 

### Possible Next Steps

Here are some things we can do in follow ups and I'm are open to suggestions.

1. Use this new api to populate data instead of populating the global `window` on page construction.
2. Add unit test to for the new api endpoint.

1. WooCommerce > Settings. Poke around and make sure Settings are working and saving properly.
2. Add/Enable Gutenberg 19.8 or earlier.
3. Turn on the `settings` feature flag.
4. Manipulate a few settings and ensure a Notice is displayed.
5. Refresh the page and make sure settings were truly saved.
6. WooCommerce > Settings > General > Enable tax rates and calculations. Toggle this on/off and save. You should see the Tax tab in the sidebar update accordingly.  
7. Click on Taxes when it appears after saving. The "Tax options" section should be populated with settings.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


packages/js/settings-editor/src/index.tsx
```diff
@@ -28,9 +30,54 @@ dispatch( editSiteStore ).updateSettings( {
...
+		>
+			{ children }
+		</SettingsDataContext.Provider>
+	);
+};
+
+const Notices = () => {
+	// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+	// @ts-ignore
+	const notices = useSelect( ( select: any ) => {
```
psealock: I'm not sure why, but a type is required here.

plugins/woocommerce/src/Admin/API/Settings.php
```diff
@@ -0,0 +1,120 @@
...
+	public function save_settings( $request ) {
+        global $current_section, $current_tab;
+
+        $params = $request->get_params();
+
+        try {
+            // Get current tab/section and set global variables.
+            $current_tab     = empty( $params['tab'] ) ? 'general' : sanitize_title( wp_unslash( $params['tab'] ) ); // WPCS: input var okay, CSRF ok.
+            $current_section = empty( $params['section'] ) ? '' : sanitize_title( wp_unslash( $params['section'] ) ); // WPCS: input var okay, CSRF ok.
+
```
adrianduffell: It would be nice to add the tab and section to the schema in `save_items_schema`, and I think it can handle default values as well. Something like.

```
'tab' => array(
    'type'        => 'string',
    'description' => __( 'Description here', 'woocommerce' ),
     'default'     => 'general',
 ),
 ```

Comment (psealock) [reply plugins/woocommerce/src/Admin/API/Settings.php]: Nice, done in https://github.com/woocommerce/woocommerce/pull/55467/commits/30408928f79ddac61a62a0d79b7e9c2c32c1c5d3

**Comment by chihsuan** [comment]
@psealock The CI is failing because of linting/tests errors. Could you take a look? I'll review the PR soon.

File: packages/js/settings-editor/src/index.tsx [hunk 3]
```diff
@@ -28,9 +31,33 @@ dispatch( editSiteStore ).updateSettings( {
 	__experimentalDashboardLink: getAdminLink( 'admin.php?page=wc-admin' ),
 } );
 
+const Notices = () => {
+	const notices: { id: string; content: string }[] = useSelect(
```
chihsuan, line 35: It seems that `@wordpress/notices` notice action types don't match the types in the `@wordpress/components` package. I'm okay with converting the types like this. 👍 

Another option I considered is to make the types match the `SnackbarList` types.

```ts
	const notices = useSelect( ( select ) => {
		const { getNotices } = select( noticesStore );

		return getNotices().map( ( notice ) => ( {
			...notice,
			actions: notice.actions?.map( ( action ) => ( {
				label: action.label,
				onClick:
					action.onClick as unknown as React.MouseEventHandler< HTMLButtonElement >,
				url: action.url ?? '',
			} ) ),
		} ) );
	}, [] );
```

plugins/woocommerce/includes/admin/settings/class-wc-settings-page.php
```diff
@@ -188,6 +188,20 @@ public function add_settings_page( $pages ) {
...
+			 * Filters the settings tabs array.
+			 *
+			 * @since 2.5.0
+			 *
+			 * @param array $available_pages The available pages.
+			 */
+			$available_pages = apply_filters( 'woocommerce_settings_tabs_array', array() );
+
+			// If any page has removed itself from the tabs array, avoid adding this page to the settings editor.
+			if ( ! in_array( $this->id, array_keys( $available_pages ), true ) ) {
```
chihsuan: Nice! 👍  I think this would fix this issue https://github.com/woocommerce/woocommerce/issues/55530.

plugins/woocommerce/includes/admin/settings/class-wc-settings-page.php
```diff
@@ -188,6 +188,20 @@ public function add_settings_page( $pages ) {
...
 		 */
 		public function add_settings_page_data( $pages ) {
+			/**
+			 * Filters the settings tabs array.
+			 *
+			 * @since 2.5.0
+			 *
+			 * @param array $available_pages The available pages.
+			 */
+			$available_pages = apply_filters( 'woocommerce_settings_tabs_array', array() );
```
chihsuan: I am concerned (not too sure) about potential performance issues. It appears that `apply_filters` is invoked for each page. If this proves problematic, we could move this check to `get_page_data` to ensure it is only called once.

plugins/woocommerce/src/Admin/Features/Settings/Init.php
```diff
@@ -165,18 +165,33 @@ public static function add_component_settings( $settings ) {
 		$setting_pages = \WC_Admin_Settings::get_settings_pages();
 		$pages         = array();
 		foreach ( $setting_pages as $setting_page ) {
-			$scripts_before_adding_settings = $wp_scripts->queue;
-			$pages                          = $setting_page->add_settings_page_data( $pages );
-
+			$scripts_before_adding_settings                         = $wp_scripts->queue;
```
chihsuan: This changes the result of `settingsScripts`. The `$scripts_before_adding_settings` is now the same as next line where we compare the difference to know what scripts are needed for the current page.

packages/js/settings-editor/src/legacy/form.tsx
```diff
@@ -50,30 +60,60 @@ export const Form = ( {
...
+				method: 'POST',
+				body: payload,
+				credentials: 'same-origin', // Include cookies for nonce validation
+				headers: {
+					'X-WP-Nonce': settingsData._wpnonce,
+				},
+			}
 		);
+
+		const responseData = await response.json();
```
chihsuan: This might throw an error. I think we need to add a `try...catch`.

```ts
    try {
        // ... existing code ...
    } catch (error) {
        createNotice(
            'error',
            __('Failed to save settings. Please try again.', 'woocommerce')
        );
        console.error('Settings save error:', error);
    } finally {
        setIsBusy(false);
    }
```

packages/js/settings-editor/src/legacy/form.tsx
```diff
@@ -50,30 +60,60 @@ export const Form = ( {
...
+		updatedData._wpnonce = settingsData._wpnonce;
+		updatedData._w_http_referer = '/wp-admin/' + getNewPath( query );
+
+		const payload = new FormData();
+		for ( const [ key, value ] of Object.entries( updatedData ) ) {
+			payload.append( key, value );
+		}
+
+		const response = await fetch(
+			`/?rest_route=/wc-admin/legacy-settings&${ new URLSearchParams(
```
chihsuan: This might cause an issue if the store is in a subdirectory. For example, if the URL is `https://example.com/subdirectory/`, the fetch call to `/?rest_route=/wc-admin/legacy-settings` will return an invalid response. I wonder if we can use `apiFetch` to make the request. I think it will handle this automatically.

chihsuan (commented):
I've reviewed the code first and left some comments/suggestions. I'll test it later.

plugins/woocommerce/includes/admin/settings/class-wc-settings-page.php
psealock: Good call. I moved in in [move woocommerce_settings_tabs_array to get_page_data](https://github.com/woocommerce/woocommerce/pull/55467/commits/474de6457c639ebe6a18b14a755ec66837bdbdd2). Now the hook is called once.

plugins/woocommerce/includes/admin/settings/class-wc-settings-page.php
psealock: Yes it does!

packages/js/settings-editor/src/index.tsx
psealock: Ok, thanks for bringing this up. I will leave this as it is because its likely temporary. I added the notices to let merchants know the action succeeded/failed but all this depends on what Design determines about the saving mechanism, so I think we'll remove this eventually.

plugins/woocommerce/src/Admin/Features/Settings/Init.php
psealock: Oh yikes, didn't see that.

Fixed in [keep script diff logic intact](https://github.com/woocommerce/woocommerce/pull/55467/commits/eea96217fb55be67a823f018e19678cc94681b23)

packages/js/settings-editor/src/legacy/form.tsx
psealock: Fixed in [add try/catch to response handling](https://github.com/woocommerce/woocommerce/pull/55467/commits/8925a710987d1551ec6a3e18d5cd330198326f2e)

packages/js/settings-editor/src/legacy/form.tsx
psealock: Nice, fixed in [use apiFetch for api call](https://github.com/woocommerce/woocommerce/pull/55467/commits/812f9643b150d61db1d1b91259156b640c637a60)

chihsuan (commented):
Thanks for addressing the feedback. @psealock 👍  I've tested the changes. I can see the Tax tab update when I toggle the Tax option. However, I found that the settings are reset when I save the settings on other tabs. 🤔 

https://github.com/user-attachments/assets/f87f9e18-ca5d-4e47-a505-d1a771ab8ef2

psealock:
>However, I found that the settings are reset when I save the settings on other tabs. 🤔

Oh whoops! in a refactor I forgot to add the query parameters to the request, so any page other than the default main page was getting settings reset to default values. Fixed in [add query args](https://github.com/woocommerce/woocommerce/pull/55467/commits/f573ec141a09e31dc999c48b03cedeab0d81d437).

Thanks for going beyond the test instructions @chihsuan, can you check again?

chihsuan (approved):
Thanks, I tested it again and confirmed that settings persist when I save on other tabs. 👍

Just found another issue, I see that payment page renders duplicate methods sections when saving settings on other tabs. I'm unsure if it's easy to fix. I'm okay if you want to address it in a separate PR.


https://github.com/user-attachments/assets/e2ac7f29-60cf-4486-9538-49f6e364a61a

psealock:
Just noting here Settings in broken until https://github.com/woocommerce/woocommerce/pull/55848 is merged

adrianduffell (approved):
It's testing nice! Thanks for the improvements to the REST API schema 🚀

psealock:
>Just found another issue, I see that payment page renders duplicate methods sections when saving settings on other tabs. I'm unsure if it's easy to fix. I'm okay if you want to address it in a separate PR.

Thanks @chihsuan, I'll need to dive in a bit deeper on this on in a follow up

Issue #46771, opened 2024-04-19
Editing a post result in console error: Store "core/interface" is already registered


mattsherman:
### Discussed in https://github.com/woocommerce/woocommerce/discussions/45949

<div type='discussions-op-text'>

<sup>Originally posted by **igorschoester** March 26, 2024</sup>
When registering a block server side via `\Automattic\WooCommerce\Admin\Features\ProductBlockEditor\register_block_type_from_metadata`
And client side via `registerProductEditorBlockType` from `@woocommerce/product-editor`.

The block is also active when editing a post in the block editor. Now I could add some logic to prevent the registration there and be done. However, it highlighted a more interesting thing.
When I edit a post in the block editor, the import of `registerProductEditorBlockType` from `@woocommerce/product-editor` alone results in a console error: `Store "core/interface" is already registered.`.

I think this is due to the main entry of `@woocommerce/product-editor` is running code already: https://github.com/woocommerce/woocommerce/blob/d279466eb3274c30baf758e4451aa17a79d3f5c2/packages/js/product-editor/src/index.ts#L44
Though it could very well be that script bundling along the `@wordpress/data` package or something.
I must admit I'm not sure how it gets to the `core/interface` registration as I didn't investigate beyond the import being the problem.

As a side note: we are marking `@woocommerce/product-editor` as an external in our Webpack configuration, i.e. we read from `window.wc.productEditor`. Just preferring the import for the auto-generated dependencies via he `DependencyExtractionWebpackPlugin`.</div>

--

It doesn't seem to be related to running code in the main entry of @woocommerce/product-editor, because if I comment those out, the error still persists.

I think it has something to do with our importing of one or both of @wordpress/edit-site (in core WC) and @wordpress/edit-post (in @woocommerce/product-editor).

There appears to be a Gutenberg issue that seems related to this: https://github.com/WordPress/gutenberg/issues/54252

If I comment out the enqueuing of @woocommerce/product-editor ([here](https://github.com/woocommerce/woocommerce/blob/5fe34385a1b163c1294c1455824ab68c2e47f977/plugins/woocommerce/src/Internal/Admin/WCAdminAssets.php#L281)), then the error does not occur when editing a post. But, of course, then wc-admin pages (Home, Product Editor, etc.) do not load then! 😄 Perhaps one solution is to only enqueue @woocommerce/product-editor when on a wc-admin page.

galbaras:
[...]

PR #55496, opened 2025-02-14
Remove unused wp/editor enqueue

louwie17:

This is another fix for #46771 as the `wp-editor` package triggers the register call for the `core/interface` package. The thing is that when the product editor feature flag is not enabled we do not need this dependency at all.

We actually already had similar logic in place for the customize your store page: https://github.com/woocommerce/woocommerce/blob/c29b2e4c77c2797fc51141dc646bacc9508f0887/plugins/woocommerce/src/Internal/Admin/WCAdminAssets.php#L319-L326

Removing this dependency actually changes the enqueued scripts from 125 to 86.

Also addresses #46771 

1. With the new product editor **disabled** ( **WooCommerce > Settings > Advanced > Features** )
2. Go to **WooCommerce > Home**
3. Open your console, it should **not** show this error: `Store "core/interface" is already registered`
4. Now enable the new product editor using **WooCommerce > Settings > Advanced > Features**
2. Go to **WooCommerce > Home**
3. Open your console, it will show the error: `Store "core/interface" is already registered` ( this PR ( https://github.com/woocommerce/woocommerce/pull/55489 ) fixes the error in this use case )
4. Go to **Products > Add new product**
5. It should render the new product editor correctly

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/src/Internal/Admin/WCAdminAssets.php
+++ b/plugins/woocommerce/src/Internal/Admin/WCAdminAssets.php
@@ -9,7 +9,7 @@ use _WP_Dependency;
 use Automattic\WooCommerce\Admin\Features\Features;
 use Automattic\WooCommerce\Admin\PageController;
 use Automattic\WooCommerce\Internal\Admin\Loader;
-
+use Automattic\WooCommerce\Utilities\FeaturesUtil;
 /**
  * WCAdminAssets Class.
  */
@@ -322,6 +322,12 @@ class WCAdminAssets {
 					str_starts_with( wc_clean( wp_unslash( $_GET['path'] ) ), '/customize-store' )
 				);
 				if ( ! $is_customize_store_page && WC_ADMIN_APP === $script ) {
+					$script_assets['dependencies'] = array_diff( $script_assets['dependencies'], array( 'wp-editor' ) );
+				}
+
+				// Remove wp-editor dependency if the product editor feature is disabled as we don't need it.
+				$is_product_data_view_page = \Automattic\WooCommerce\Admin\Features\ProductDataViews\Init::is_product_data_view_page();
+				if ( 'wc-product-editor' === $script && ! ( FeaturesUtil::feature_is_enabled( 'product_block_editor' ) || $is_product_data_view_page ) ) {
 					$script_assets['dependencies'] = array_diff( $script_assets['dependencies'], array( 'wp-editor' ) );
 				}
 
--- a/plugins/woocommerce/changelog/fix-46771_remove_unused_wp_editor_enqueue
+++ b/plugins/woocommerce/changelog/fix-46771_remove_unused_wp_editor_enqueue
@@ -1,0 +1,4 @@
+Significance: minor
+Type: performance
+
+Remove the unneeded wp-editor script dependency when the product editor feature is not enabled.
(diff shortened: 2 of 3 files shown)


plugins/woocommerce/src/Internal/Admin/WCAdminAssets.php
```diff
@@ -325,6 +325,12 @@ public function register_scripts() {
 					$script_assets['dependencies'] = array_diff( $script_assets['dependencies'], array( 'wp-editor' ) );
 				}
 
+				// Remove wp-editor dependency if the product editor feature is disabled as we don't need it.
+				$is_product_data_view_page = \Automattic\WooCommerce\Admin\Features\ProductDataViews\Init::is_product_data_view_page();
+				if ( 'wc-product-editor' === $script && ! ( FeaturesUtil::feature_is_enabled( 'product_block_editor' ) || $is_product_data_view_page ) ) {
+					$script_assets['dependencies'] = array_diff( $script_assets['dependencies'], array( 'wp-editor' ) );
+				}
```
louwie17, line 332: Similar logic as above, used for the customize your store package.

Aljullu (commented):
Thanks for looking into this, @louwie17! Confirming it removes the error from WooCommerce > Home! I can still reproduce it in `/wp-admin/post-new.php`, I guess there might be different causes for this error?

One more question: I miss a lot of context related to the project, but I'm not sure to understand why are we loading `wc-product-editor` when the feature flag is disabled and in screens unrelated to the Product Editor. Is that script needed in those cases? The more scripts we can avoid loading on every screen, the better, but I'm not sure if it's possible.

louwie17:
> Thanks for looking into this, @louwie17! Confirming it removes the error from WooCommerce > Home! I can still reproduce it in `/wp-admin/post-new.php`, I guess there might be different causes for this error?

Hmm do you also have the WooCommerce Beta Tester plugin installed and activated? If so, try disabling it!
It depends on the product editor package and enqueues on every admin page 🤦  let me create a separate PR to address that.

> One more question: I miss a lot of context related to the project, but I'm not sure to understand why are we loading `wc-product-editor` when the feature flag is disabled and in screens unrelated to the Product Editor. Is that script needed in those cases? The more scripts we can avoid loading on every screen, the better, but I'm not sure if it's possible.

Yes that is also an option, my only slight hesitation of doing that now is that plugins may depend on the `window.wc.product-editor` package to be available even if the feature flag isn't enabled. And explicitly removing the dependency would also cause it not to be available there.
I did do a check and removing the `wc-product-editor` dependency doesn't actually save us that much, it only ends up not enqueuing it-self.
I guess the big problem with `wp/editor` was that it depended on a lot of things that also got enqueued.

Aljullu (approved):
> Hmm do you also have the WooCommerce Beta Tester plugin installed and activated? If so, try disabling it!

True, that fixed it, thanks!

> Yes that is also an option, my only slight hesitation of doing that now is that plugins may depend on the `window.wc.product-editor` package to be available even if the feature flag isn't enabled. And explicitly removing the dependency would also cause it not to be available there.
> I did do a check and removing the `wc-product-editor` dependency doesn't actually save us that much, it only ends up not enqueuing it-self.
> I guess the big problem with `wp/editor` was that it depended on a lot of things that also got enqueued.

Makes sense. IMO, if it hasn't been promoted as a stable API, we should be fine removing it. But no need to do it in this PR.

LGTM, thanks for working on this, @louwie17!

PR #55458, opened 2025-02-13
Add Cursor AI Assistant Rules Configuration

nerrad:
### Changes proposed in this Pull Request:

This PR introduces Cursor AI assistant rules to enhance development workflow and maintain consistency for its usage across the WooCommerce codebase. 

Why Add Cursor Rules?
- **Consistent Development Practices**: By codifying our development guidelines into Cursor rules, we ensure that AI-assisted development follows our established best practices.
- **Improved Collaboration**: New contributors using Cursor will automatically receive guidance for various nuances of developing in the WooCommerce monorepo which in turn will improve the AI assisted development has more accurate initial output.
- **Scalable Documentation**: These rules serve as living documentation that's immediately accessible to developers in their IDE, making it easier to maintain and follow our conventions.

Initial Implementation:
- Added configuration for git-related practices, including:
  - Branch naming conventions for different types of changes (feature, fix, refactor, etc.)
  - Commit message formatting guidelines
  - Best practices for atomic commits
- Added `.cursorignore` to `.gitignore` to maintain clean git status

Future Potential:
This initial setup creates a foundation for adding more rules that could cover:
- Coding standards and style guides
- Testing requirements
- Documentation conventions
- Common architectural patterns
- Security best practices

The rules are stored in `.cursor/rules/` and can be iteratively expanded as we identify more areas where automated guidance would be valuable.

### How to test the changes in this Pull Request:

This has no user impact and is best tested in future development flows (however I did use this to create the branch and the commits for this PR. I plan on iterating on this after the initial merge but rules can also be iterated on and improved over time as we learn what is most effective for the AI assitance.

### Changelog entry

-   [ ] Automatically create a changelog entry from the details below.

-   [x] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Details</summary>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

-   [ ] Fix - Fixes an existing bug
-   [ ] Add - Adds functionality
-   [ ] Update - Update existing functionality
-   [ ] Dev - Development related task
-   [ ] Tweak - A minor adjustment to the codebase
-   [ ] Performance - Address performance issues
-   [ ] Enhancement - Improvement to existing functionality

#### Message 

</details>

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 
This is a developer workflow related update that has no relevance to consumers of the repository unless they are using Cursor. I don't think this needs to be included in a changelog.

</details>

PR #55500, opened 2025-02-14
Add/wordpress rules

alexflorisca:

Following on from @nerrad 's PR https://github.com/woocommerce/woocommerce/pull/55458, I added a rules file for WordPress development in general. From here: https://cursor.directory/wordpress-woocommerce. This is open to changes and suggestions, but I've been using this for a few days and it's made my interaction with Cursor's AI much better.

1. Start coding and see if there any any improvement in your interaction with Cursor's AI

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

This is a developer workflow improvement and does not need to be included in the changelog.

</details>

### DIFF
--- a/.cursor/rules/wordpress.mdc
+++ b/.cursor/rules/wordpress.mdc
@@ -3,6 +3,8 @@ description: WordPress and WooCommerce dev
 globs: *.php
 ---
 
+<!-- Original rules from here: https://cursor.directory/wordpress-woocommerce -->
+
 You are an expert in WordPress, WooCommerce, PHP, and related web development technologies.
 
 Key Principles
--- a/.cursor/rules/wordpress.mdc
+++ b/.cursor/rules/wordpress.mdc
@@ -1,0 +1,90 @@
+---
+description: WordPress and WooCommerce dev
+globs: *.php
+---
+
+You are an expert in WordPress, WooCommerce, PHP, and related web development technologies.
+
+Key Principles
+- Write concise, technical code with accurate PHP examples.
+- Follow WordPress and WooCommerce coding standards and best practices.
+- Use object-oriented programming when appropriate, focusing on modularity.
+- Prefer iteration and modularization over duplication.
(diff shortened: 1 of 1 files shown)


nerrad:
I think we'll get better results if we get more specific with the individual rules. So things like coding standards can be globbed to specific files (eg. `*.php` or `*.js`). Keeping everything in one general file is harder to maintain and risks duplication as we expand the rules over time. More atomic rules will help with evaluating the impact of changes as well.

alexflorisca:
Yeah agreed, I've set that rule up for `.php` files locally. ~~I'm not sure how or if its possible to save that to the file or if it's just an editor setting?~~

It adds a header to the top of the file
```
---
description: WordPress and WooCommerce dev
globs: *.php
---
```

The intention is to keep adding as necessary: front-end files, test files, etc.

nerrad:
The rules you've copied are influenced by when everything was included in a `.cursorrules` file. This was before Cursor introduced more atomic rules via `.cursor/rules`. So, I still think you're packing too much into this single rule file and it'd work better split out into more specific rules. 

For example, looking at the rules there are a few things that stand out (this is not exhaustive, but just to give you an idea of how I'm thinking about this).

- General rules file: applied across the project and includes things like it is a WordPress project etc.
- Linting specific rules per lang (so can be globbed) - and it can also reference the actual standards the monorepo follows.
- Versioning rules
- General test rules (in additiona to language specific test rules).
- ....

Etc. Essentially, I'm asking you to look through what you've copy and pasted and think more critically about how the rules can be implemented. In the meantime, you can still use this locally in development. FWIW, I've found Cursor has been more accurate for me when I tailor things the way I described above.

alexflorisca:
Ok thanks for your feedback Darren, I'll close this PR for now and revisit this

PR #55510, opened 2025-02-15
Performance: Simpler query for orders months filter dropdown

### LINKED ISSUE #54579: [Performance]: Months filter cache query
### Prerequisites

### Performance Details

When visiting the admin orders pages, the months query is sometimes run, taking almost a full second for the query on a test site with ~1 million orders.

Request URL: `/wp-admin/admin.php?page=wc-orders`
Average query time: ~1s

Image

### Affected Sites

This affects sites with larger numbers of orders (~1 million on my test site) and where the months filter is still enabled (`woocommerce_' . $this->order_type . '_list_table_disable_months_filter`).

### Suspected causes

The use of `DISTINCT` and `ORDER BY` probably slow down this query, but both seem needed for this query to work.

We might consider running [the query](https://github.com/woocommerce/woocommerce/blob/7f513fe2e50a31bfb0687e55dce9ae8c857976cc/plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php#L798-L848) in the background instead of on page load or dynamically generating the months in PHP by oldest order instead of getting them from the database.

coreymckrill:

This introduces an alternative method for generating the list of year-months in the Months filter dropdown. The new method uses a much simpler query that just takes the oldest order and builds the list of options by adding every month from then until the current one. The tradeoff is that if a store has some large gaps of time between some orders, the dropdown may show months where no orders were actually created.

Fixes #54579

### To do

- [x] Maybe update or add unit tests
- [ ] ~Add documentation~

1. Ensure your test site has HPOS activated, as this change only relates to the list table for HPOS orders.
1. Install [Smooth Generator](https://github.com/woocommerce/wc-smooth-generator/releases/latest) on your test site for easy generation of orders with variable creation dates.
2. Starting with a fresh site with no orders, generate 1 order for the current date: `wp wc generate orders 1`
3. Go to the Orders list table and check the Months filter dropdown. It should just have a single entry for the current month.
4. Now generate an order from 10 years in the past: `wp wc generate orders 1 --date-start=2015-01-01 --date-end=2015-02-01`
7. Check the Months filter dropdown. There should now be an option for each month between 2015 and now.
8. Trash both of the test orders generated so far. Then generate a new one _in the future_ ✨ `wp wc generate orders 1 --date-start=2027-01-01 --date-end=2027-02-01`
9. Check the Months dropdown again. It shouldn't be broken! Instead, it should show every month between now and 2027.


jorgeatorres:
> The tradeoff is that if a store has some large gaps of time between some orders, the dropdown may show months where no orders were actually created.

Is this a trade off we're comfortable with? This would deviate from how the same dropdown works in other places in the admin (Posts, Pages, etc.). I understand a store probably has more orders than posts/pages, but still, listing every month since the first order is not too different from a hardcoded "month" selector.

Not saying I'm against, but it's probably worth discussing. Or am I overthinking this?

coreymckrill:
It's a fair question. I shared some other thoughts about it in [the issue](https://github.com/woocommerce/woocommerce/issues/54579#issuecomment-2657721942).

One way to compromise here, since this change is about optimizing for stores with a large number of orders, might be to add a filter hook here, with the default being the expensive query way, and the filter allowing you to toggle to the more performant way.

Thoughts, @joshuatf ?

coreymckrill:
@jorgeatorres Ok I've refactored this a bit and updated the description and testing instructions. Can you take another look?

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
```diff
@@ -823,34 +822,116 @@ protected function get_and_maybe_update_months_filter_cache(): array {
...
+	protected function get_simplified_months_filter_options(): array {
+		global $wpdb;
+
 		$orders_table = esc_sql( OrdersTableDataStore::get_orders_table_name() );
-		$utc_offset   = wc_timezone_offset();
+		$trash_status = esc_sql( OrderStatus::TRASH );
 
-		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
-		$order_dates = $wpdb->get_results(
+		$first_year_month = $wpdb->get_row(
```
joshuatf: Nice!  This seems like a reasonable approach.

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
```diff
@@ -823,34 +822,116 @@ protected function get_and_maybe_update_months_filter_cache(): array {
...
-					SELECT DISTINCT YEAR( t.date_created_local ) AS year,
-									MONTH( t.date_created_local ) AS month
-					FROM ( SELECT DATE_ADD( date_created_gmt, INTERVAL $utc_offset SECOND ) AS date_created_local FROM $orders_table WHERE type = %s AND status != 'trash' ) t
-					ORDER BY year DESC, month DESC
+					SELECT YEAR( t.date_created_gmt ) AS year,
+					       MONTH( t.date_created_gmt ) AS month
+					FROM $orders_table t
+					WHERE type = %s
+					AND status != '$trash_status'
+					ORDER BY year ASC, month ASC
```
joshuatf, line 826: Just curious if we've tested impact of `ORDER BY` on this query.  I'm not sure that we have alternatives (other than using ID and hoping 🤞  things are chronological), but this clause tends to be one that tends to come up a lot on larger sites.

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
```diff
@@ -823,34 +822,116 @@ protected function get_and_maybe_update_months_filter_cache(): array {
...
+						FROM ( SELECT DATE_ADD( date_created_gmt, INTERVAL $utc_offset SECOND ) AS date_created_local FROM $orders_table WHERE type = %s AND status != '$trash_status' ) t
+						ORDER BY year DESC, month DESC
+					",
+					$this->order_type
+				)
+				// phpcs:enable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+			);
+		}
+
+		update_option( $cache_option_date_name, time(), false );
```
joshuatf: I know this was copied over from the old approach, but curious why this uses option setters over `wp_cache` or even transients with expirations.

joshuatf (commented):
Sorry for the late review on this one.  I really like where this is headed!  Left a couple comments in the code.

>One way to compromise here, since this change is about optimizing for stores with a large number of orders, might be to add a filter hook here, with the default being the expensive query way, and the filter allowing you to toggle to the more performant way.
>Thoughts, @joshuatf ?

I think we need some way of enabling this by default in larger sites.  My personal opinion is that having more month options to select from is not that much of a nuisance for stores that might not be doing sales every month, but hopefully those cases are in the minority anyway.

If anything, I'd push towards making the simplified the default behavior, but if we feel the need to guard it, we could base the default behavior on store size (number of orders).

>Is this a trade off we're comfortable with? This would deviate from how the same dropdown works in other places in the admin (Posts, Pages, etc.). I understand a store probably has more orders than posts/pages, but still, listing every month since the first order is not too different from a hardcoded "month" selector.

This is a good point, though orders are a very different post type from posts and pages.  They are posts which (ideally) are created in much higher numbers, on a recurring basis, and not by merchants/site-owners.  While I might not appreciate the ability to have to guess months in a filter for posts I've created, it's still valuable information to filter by orders in April of 2024 and find out that no orders exist.

The current dropdown does provide this information as well without the need to navigate to the page, but the question is whether or not this information is beneficial to have on every page load and outweighs performance of loading orders on all other pages/filters.

I would strongly be in favor of performance for most order workflows over recognizing from a dropdown that I did not have orders in a given month from the previous year on every page.

jorgeatorres:
> If anything, I'd push towards making the simplified the default behavior, but if we feel the need to guard it, we could base the default behavior on store size (number of orders).

Yes, if we're happy with the less accurate dropdown, I'd prefer we just implement the one approach. Our list table is already overly complex as it is 😬.

> I would strongly be in favor of performance for most order workflows over recognizing from a dropdown that I did not have orders in a given month from the previous year on every page.

Gotcha!

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
coreymckrill: Haven't tested it, but I don't think we can rely on the ID being chronological. If a store has ever imported orders from another site or a 3rd party system, they could be from any date but they would still increment the ID column as usual.

coreymckrill:
Ok, based on the conversation above, I've switched back to the original approach of introducing the method with the simplified query and deprecating the old method. @jorgeatorres can you give it another quick look?

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
```diff
@@ -802,61 +801,93 @@ private function months_filter() {
...
 				"
-					SELECT DISTINCT YEAR( t.date_created_local ) AS year,
-									MONTH( t.date_created_local ) AS month
-					FROM ( SELECT DATE_ADD( date_created_gmt, INTERVAL $utc_offset SECOND ) AS date_created_local FROM $orders_table WHERE type = %s AND status != 'trash' ) t
-					ORDER BY year DESC, month DESC
+					SELECT YEAR( t.date_created_gmt ) AS year,
+					       MONTH( t.date_created_gmt ) AS month
+					FROM $orders_table t
+					WHERE type = %s
+					AND status != '$trash_status'
```
jorgeatorres: Maybe we can use `%s` here and pass `OrderStatus::TRASH` as an arg instead of including it in the query itself?

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
```diff
@@ -802,61 +801,93 @@ private function months_filter() {
...
-		update_option( $cache_option_value_name, $order_dates );
+		$end     = new \DateTime( 'now', wp_timezone() );
+		$options = array();
+
+		if ( is_object( $first_year_month ) ) {
+			$start = new \DateTime(
+				sprintf(
+					'%s-%s-01',
+					$first_year_month->year,
+					$first_year_month->month
```
jorgeatorres: Note that `$first_year_month` is in UTC (because it comes from `date_created_gmt`) so we can't assume it's in the WP timezone. For example, if your WP timezone is UTC-5 and you have an order with `date_created_gmt = 2025-01-01 01:00:00`, the dropdown will list January 2025, but should in fact list December 2025 (which is the order date in WP's timezone).

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
```diff
@@ -802,61 +801,93 @@ private function months_filter() {
...
+				wp_timezone()
+			);
+
+			// If, somehow, the oldest order date is in the future, swap the start and end of the range.
+			if ( $start > $end ) {
+				$end   = $start;
+				$start = new \DateTime( 'now', wp_timezone() );
+			}
+
+			while (
```
jorgeatorres: I might be missing something, but couldn't we use `DatePeriod` instead of all this "manual" adding of intervals?

```php
foreach ( new \DatePeriod( $start, new \DateInterval( 'P1M' ), $end ) as $d ) {
	// ...
}
```

jorgeatorres (changes requested):
Hey @coreymckrill! Overall this looks good to me and tests well, except for a corner case which I've documented in the feedback. Other than that, I just left a few tiny suggestions. Let me know what you think!

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
coreymckrill: Updated in 6e66ebae0d

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
coreymckrill: Good catch! I suspect the old approach has this same defect. I updated this in 6e66ebae0d and added a unit test for it.

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
coreymckrill: I actually didn't know about `DatePeriod` 😳 I tried your suggestion, but it doesn't always work quite right because it excludes the end date from the list of recurrences. There's an optional flag `DatePeriod::INCLUDE_END_DATE` to adjust this, but it's not available until PHP 8.2.

jorgeatorres: Yeah, we would have to include the end date manually, but maybe that's still an improvement / or at least more clear? This is not a blocker, of course.

jorgeatorres (approved):
LGTM. Tests well (tried a few corner cases by changing the timezone on my test site as well as the order's creation date).
Thank you @coreymckrill!

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
coreymckrill: Just my personal preference, I guess, but I think it's actually clearer/less complex if we don't have to add extra logic in order to get the correct last item. I think I'll leave it for now.

PR #55523, opened 2025-02-16
Add filter to customize of 'Your order' in form-checkout template

ryross:

Added new filter `woocommerce_checkout_order_review_heading` to allow customizing the 'Your Order' text on the checkout form.

1. Use add_filter() to override the text like this: ```add_filter('woocommerce_checkout_order_review_heading', function ($text) {
    return 'Order Summary';
});```

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/templates/checkout/form-checkout.php
+++ b/plugins/woocommerce/templates/checkout/form-checkout.php
@@ -55,7 +55,6 @@ if ( ! $checkout->is_registration_enabled() && $checkout->is_registration_requir
 	/**
 	 * Filter to modify the output of the order review heading.
 	 *
-	 * @since x.x.x
 	 *
 	 * @param string $output The heading text.
 	 */
--- a/plugins/woocommerce/templates/checkout/form-checkout.php
+++ b/plugins/woocommerce/templates/checkout/form-checkout.php
@@ -52,6 +52,13 @@ if ( ! $checkout->is_registration_enabled() && $checkout->is_registration_requir
 	<?php do_action( 'woocommerce_checkout_before_order_review_heading' ); ?>
 
 	<?php
+	/**
+	 * Filter to modify the output of the order review heading.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param string $output The heading text.
+	 */
 	$order_review_heading = apply_filters( 'woocommerce_checkout_order_review_heading', __( 'Your order', 'woocommerce' ) );
 	if ( $order_review_heading ) :
 		?>
--- a/plugins/woocommerce/templates/checkout/form-checkout.php
+++ b/plugins/woocommerce/templates/checkout/form-checkout.php
@@ -50,9 +50,14 @@ if ( ! $checkout->is_registration_enabled() && $checkout->is_registration_requir
 	<?php endif; ?>
 	
 	<?php do_action( 'woocommerce_checkout_before_order_review_heading' ); ?>
-	
-	<h3 id="order_review_heading"><?php esc_html_e( 'Your order', 'woocommerce' ); ?></h3>
-	
+
+	<?php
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/templates/checkout/form-checkout.php
```diff
@@ -50,9 +50,21 @@
...
 	<?php do_action( 'woocommerce_checkout_before_order_review_heading' ); ?>
-	
-	<h3 id="order_review_heading"><?php esc_html_e( 'Your order', 'woocommerce' ); ?></h3>
-	
+
+	<?php
+	/**
+	 * Filter to modify the output of the order review heading.
+	 *
+	 * @since x.x.x
```
alexflorisca: I think just a comment describing the filter is ok here, no need for doc style comment.

alexflorisca (dismissed):
Thanks for the PR. I don't see a problem with this. I've just left a comment about your comment!

plugins/woocommerce/templates/checkout/form-checkout.php
alexflorisca: Removing the @\since tag will also make the linting happy.

ryross:
```suggestion
```

plugins/woocommerce/templates/checkout/form-checkout.php
```diff
@@ -50,9 +50,20 @@
...
 	
 	<?php do_action( 'woocommerce_checkout_before_order_review_heading' ); ?>
-	
-	<h3 id="order_review_heading"><?php esc_html_e( 'Your order', 'woocommerce' ); ?></h3>
-	
+
+	<?php
+	/**
+	 * Filter to modify the output of the order review heading.
+	 *
```
ryross:
```suggestion
```

ryross:
Just removed the since :) Thanks!

Konamiman (changes requested):
Hi @ryross, thanks for your contribution. We generally prefer not to add new hooks (or any kind of "business logic") to templates, so could you please do the following changes:

1. Copy this line:

```php
$order_review_heading = apply_filters( 'woocommerce_checkout_order_review_heading', __( 'Your order', 'woocommerce' ) );
```

...to right before the point where the template is rendered, which is [here](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/shortcodes/class-wc-shortcode-checkout.php#L373).

2. Now replace the line in the template to the following, so that a fallback value is in place in case the variable is missing:

```php
$order_review_heading ??= __( 'Your order', 'woocommerce' );
```

3. Add a docblock to the hook (yes, that's actually required, sorry for the trouble). Use 9.9.0 for the `@since` tag (but don't worry too much about it, I'll adjust it myself if needed). It would be good mentioning that the value of `order_review_heading` can be an empty string to prevent it from being rendered but it should **not** be null.

And I think that would be it. Please comment back if something is unclear or if there's any problem with my suggestion. Thank you very much.

plugins/woocommerce/templates/checkout/form-checkout.php
Konamiman: @alexflorisca Sorry to disagree, but we actually _do_ require a proper standard docblock for any new hook added, including a `@since` tag and as many `@param` tags as appropriate. 🙂 Anyway I'm requesting additional changes in a separate comment.

alexflorisca: Ok that's my bad. I come front a more front-end world so that's good learning for me, thank you!

nerrad:
No response from the author, so I'm going to close.

PR #55544, opened 2025-02-17
Fix for feature compatibility declarations happening too early

Konamiman:

#55303 made changes to the features controller class in order to make the features registered via the `woocommerce_register_feature_definitions` action visible when running WP CLI commands. This was achieved by moving the firing of that action from the `get_feature_definitions` method to the `init` hook.

However, while that fixed the issue concerning WP CLI, it also broke the compatibility declaration mechanism: plugins declare compatibility with WooCommerce features, but that happens inside `before_woocommerce_init`, which is fire before `init`; thus extensions were trying to declare compatibility with extensions not yet registered, which had no effect.

This pull request fixes this by making sure that additional features are registered before plugins declare compatibility.

1. Follow the testing instructions of #55303, everything should still work.
2. Enable HPOS (in WooCommerce - Settings - Advanced - Features).
3. Install a HPOS compatible plugin (for example [Smooth Generator](https://github.com/woocommerce/wc-smooth-generator/releases)), activate it, and verify that it's *not* flagged as incompatible with HPOS in the plugins list.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


nathanss (approved):
I've manually tested the PR and saw that the issue is fixed for all of SomewhereWarm plug-ins

(didn't review the code)

NeosinneR:
@Konamiman I've quickly checked it with MailPoet, and there might still be an issue, but a different one. 

1. Install [MailPoet](https://wordpress.org/plugins/mailpoet/) - it's declared compatibility with `HPOS`, but not the `Email improvements` feature.
2. Go to `/wp-admin/plugins.php?plugin_status=incompatible_with_feature` and test with following states:
    - ✅ Enabled `HPOS`, disabled `Email improvements` - **no message shown**.
    - ❌ Enabled `HPOS`, enabled `Email improvements` - **no incompatibility is shown**.
3. Install this MailPoet build with removed declaration of HPOS compatibility ([mailpoet-removed-compatibility.zip](https://github.com/user-attachments/files/18826055/mailpoet-removed-compatibility.zip)).
    - ✅  Enabled `HPOS`, disabled `Email improvements` - **incompatibility with HPOS shown**.
    - ✅  Enabled `HPOS`, enabled `Email improvements` - **incompatibility with both shown**.
    - ❌ Disabled `HPOS`, enabled `Email improvements` - **no incompatibility is shown**.

It seems like the HPOS is somehow driving the check for other features so when the HPOS feature passes, the other checks are ignored.

albarin (approved):
Looking good and confirmed HPOS compatible plugin don't show as incompatible ✅

Issue #37861, opened 2023-04-19
$product->get_attributes() should return an array but returns null


maximejobin:
### Prerequisites

- [X] I have carried out troubleshooting steps and I believe I have found a bug.
- [X] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

This bug creates a FATAL ERROR.

Long story short, we have an attribute that is added dynamically. This is how we discovered this "contract" issue.

The following line expects the `$attributes` variable to be an array ([source](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/wc-conditional-functions.php#L327-L328)):

    $attributes           = $product->get_attributes();
    return ( in_array( $name, array_keys( $attributes ), true ) && in_array( $value, $variation_attributes[ $attributes[ $name ]['name'] ], true ) );

The proof is that the next line calls `array_keys` on it.

But the `get_attributes()` function calls `get_prop()` and returns whatever it gets ([source](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/abstracts/abstract-wc-product.php#L530)):

	/**
	 * Returns product attributes.
	 *
	 * @param  string $context What the value is for. Valid values are view and edit.
	 * @return array
	 */
	public function get_attributes( $context = 'view' ) {
		return $this->get_prop( 'attributes', $context );
	}

And as we can see in `get_prop`, it can return `null` ([source](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/abstracts/abstract-wc-data.php#L834)):

	/**
	 * Gets a prop for a getter method.
	 *
	 * Gets the value from either current pending changes, or the data itself.
	 * Context controls what happens to the value before it's returned.
	 *
	 * @since  3.0.0
	 * @param  string $prop Name of prop to get.
	 * @param  string $context What the value is for. Valid values are view and edit.
	 * @return mixed
	 */
	protected function get_prop( $prop, $context = 'view' ) {
		$value = null;

		if ( array_key_exists( $prop, $this->data ) ) {
			$value = array_key_exists( $prop, $this->changes ) ? $this->changes[ $prop ] : $this->data[ $prop ];

			if ( 'view' === $context ) {
				$value = apply_filters( $this->get_hook_prefix() . $prop, $value, $this );
			}
		}

		return $value;
	}

### Expected behavior

The `get_attributes` function should cast `null` to an empty array to respect its contract.

[...]

PR #55549, opened 2025-02-17
Check `attributes` variable is not null to avoid error in `in_array`

### LINKED ISSUE #52776: Prevent TypeError in meta_is_product_attribute() when variable attributes are not available
### Prerequisites

### Describe the bug

By default, when WooCommerce Subscriptions, the admin can't delete a subscription product or variation if it's attached to a subscription order. While it doesn't allow deleting the variation, one can just uncheck the "used for variation" in the attributes area without deleting anything.   This breaks the subscription renewal. 

### Expected behavior

The user shouldn't be allowed to uncheck the user for the varion option.

### Actual behavior

The customer payment page throws an error

### Steps to reproduce

* buy a var subs
* create a pending renewal order
8 go to the var subs product, while it doesn't allow deleting the variation, one can just uncheck the "used for variation" in the attributes area without deleting anything
* Go to pending order, and try accessing the customer payment page.

### WordPress Environment

Not needed. Can be produced in any environment with WooCommerce Subscriptions

### Isolating the problem

albarin:

This PR ensures the `get_attributes` function always returns an array (as stated on the function doc). Until now it was possible for it to return `null`, which was causing fatal errors.

Closes #52776 #37861

## Test #52776

1. Create a new variable product and get its ID.
2. In `trunk`, add the following code and use the product ID from the prev. step:
```PHP
add_filter(
	'woocommerce_product_get_attributes',
	'__return_null'
);

add_action(
	'admin_init',
	function () {
		meta_is_product_attribute( 'a_name', 1, [PRODUCT_ID] );
		die;
	}
);
```
3. Go to any admin page and check that you get an error.
4. Switch to this PR branch, refresh, you should not get an error anymore.

## Test `wc_get_formatted_variation`

1. Make sure you have a variable product with variations.
2. Add the following code and use the **variation id** from one of the product variations from step 1:
```php
add_filter( 'woocommerce_product_variation_get_attributes', function (){
	return null;
});

add_action( 'admin_init', function() {
	$product = wc_get_product([PRODUCT_VARIATION_ID]);

	var_dump(wc_get_formatted_variation($product, false));
	die;
} );
```
3. Go to any admin page and check that you get an empty string from the `wc_get_formatted_variation`.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Make sure the `get_attributes` function always returns an array.

</details>

<details>

#### Comment 

</details>


jorgeatorres (changes requested):
Hi @albarin!

This does fix the situation when the value is `null`, but precisely because the value is filterable, there are other cases this doesn't address, which seems right now would be a good opportunity to fix.

For example, by using `__return_empty_string` instead of `__return_null` in your example, we get a fatal again, so even if the fix was constrained to this particular function, I believe it needs to be more robust.

That said, I think it'd be a good opportunity to review how we use `$product->get_attributes()` across the codebase: if there are places where we specifically check for `null` or non-array and do something different to what we would've done with an empty array, then it'd be good to check why. If we don't, then I think it'd be best to make sure the return of `WC_Product::get_attributes()` is always an array (even an empty one on invalid values).

What are your thoughts?

albarin:
> That said, I think it'd be a good opportunity to review how we use $product->get_attributes() across the codebase: if there are places where we specifically check for null or non-array and do something different to what we would've done with an empty array, then it'd be good to check why. If we don't, then I think it'd be best to make sure the return of WC_Product::get_attributes() is always an array (even an empty one on invalid values).

@jorgeatorres Agree! I've been checking all the usages of `->get_attributes()` in the codebase.
Knowing it can be `null`, there are safe usages (like checking with `isset` before accessing) and unsafe usages (like using in a `foreach`, or an `array_*` function, or accessing it directly without any check). 
It seems like the general expectation is to always get an array as the `get_attributes` since it's what the documentation is saying:
```php
/**
 * Returns product attributes.
 *
 * @param  string $context What the value is for. Valid values are view and edit.
 * @return array
 */
public function get_attributes( $context = 'view' ) {
	return $this->get_prop( 'attributes', $context );
}
```

Should we then do something like this 👇 wdyt?
```php
/**
 * Returns product attributes.
 *
 * @param  string $context What the value is for. Valid values are view and edit.
 * @return array
 */
public function get_attributes( $context = 'view' ) {
    $attributes = $this->get_prop( 'attributes', $context );
    if ( is_null($attributes) ) {
        return array();
    }
    
    return $attributes;
}
```

jorgeatorres:
> there are safe usages (like checking with isset before accessing)

In any of those "safe usages" where we check that it is not null, is the behavior different from, say, `$attributes` being an empty array? Just curious if code in those places is just trying to be extra safe or it's actually trying to detect that this hasn't been set for some reason.


> Should we then do something like this 👇 wdyt?

I'd say we need to check that it is an array, not just not NULL. That is, if we decide to honor the documented method contract.

albarin:
@jorgeatorres 
> In any of those "safe usages" where we check that it is not null, is the behavior different from, say, `$attributes` being an empty array? Just curious if code in those places is just trying to be extra safe or it's actually trying to detect that this hasn't been set for some reason.

There are different situations:
1. Checking that specific attribute is set in the array (see [1](https://github.com/woocommerce/woocommerce/blob/e2b2e02ea8f836d6df413f707823c011315a43ec/plugins/woocommerce/includes/class-wc-product-variation.php#L144), [2](https://github.com/woocommerce/woocommerce/blob/cf33aa5177f3453ae09c3c94c97e5c3b120e483b/plugins/woocommerce/includes/abstracts/abstract-wc-product.php#L2130-L2133)).
2. Checking if it is not null, `is_array` or not empty (see [1](https://github.com/woocommerce/woocommerce/blob/5e69275f3f7b5088fd2c27395319da935519d074/plugins/woocommerce/includes/wc-attribute-functions.php#L204), [2](https://github.com/woocommerce/woocommerce/blob/284d195cd1973d1a87cefbce2b398bb9eb80a7f8/plugins/woocommerce/includes/wc-product-functions.php#L431), [3](https://github.com/woocommerce/woocommerce/blob/87cf80bafba0a4aad53b99d4fea6d5e0d63c3da9/plugins/woocommerce/includes/data-stores/class-wc-product-data-store-cpt.php#L987-L990), [4](https://github.com/woocommerce/woocommerce/blob/c9ee77c785f5e319e8b57bc11a765f892b7eb6bf/plugins/woocommerce/includes/data-stores/class-wc-product-variable-data-store-cpt.php#L200)).
3. Casting to `array` (see [here](https://github.com/woocommerce/woocommerce/blob/c9ee77c785f5e319e8b57bc11a765f892b7eb6bf/plugins/woocommerce/includes/data-stores/class-wc-product-variation-data-store-cpt.php#L298)).

> I'd say we need to check that it is an array, not just not NULL. That is, if we decide to honor the documented method contract.

Are you proposing something like 👇?
```php
public function get_attributes( $context = 'view' ) {
    $attributes = $this->get_prop( 'attributes', $context );
    if ( ! is_array($attributes) ) {
       return array();
    }
    
    return $attributes;
}
```

jorgeatorres:
> There are different situations: [...]

Ok, so for example the one where `is_array()` is involved seems to start doing some work when the property is an array (even an empty one). Didn't look to closely at whether it makes any difference (not array vs empty array), but those are the kinds of things we might need to change along with `WC_Product::get_attributes()`.
>

> Are you proposing something like 👇?

Yes. But I'd like to understand whether this would break anything, that's why I was suggesting we look for places in the code where in response to a non-array attributes value, the code does something special.

jorgeatorres:
By the way, thanks for the mini research @albarin! 💯

albarin:
> Yes. But I'd like to understand whether this would break anything, that's why I was suggesting we look for places in the code where in response to a non-array attributes value, the code does something special.

@jorgeatorres Okay, I'll do a more in depth research then 😄

albarin:
@jorgeatorres Okay, after some more research: 
I've found a bunch of occurrences of [this type](https://github.com/woocommerce/woocommerce/blob/cf33aa5177f3453ae09c3c94c97e5c3b120e483b/plugins/woocommerce/includes/abstracts/abstract-wc-product.php#L2130-L2133), checking if isset

albarin:
@jorgeatorres Okay, after some more research, I think we can group usage by type:

| d | Type | Example | Conclusion |
|--------|--------|--------|--------|
| 1 | `isset` | https://github.com/woocommerce/woocommerce/blob/cf33aa5177f3453ae09c3c94c97e5c3b120e483b/plugins/woocommerce/includes/abstracts/abstract-wc-product.php#L2130-L2133 | Not affected by returning `null` or `array()` |
| 2 | `if` |https://github.com/woocommerce/woocommerce/blob/87cf80bafba0a4aad53b99d4fea6d5e0d63c3da9/plugins/woocommerce/includes/data-stores/class-wc-product-data-store-cpt.php#L987-L990 | Not affected by returning `null` or `array()` |
| 3 | Used foreach loops and array functions (`array_keys`, `array_values`, etc.) | https://github.com/woocommerce/woocommerce/blob/cf33aa5177f3453ae09c3c94c97e5c3b120e483b/plugins/woocommerce/includes/abstracts/abstract-wc-product.php#L1828 https://github.com/woocommerce/woocommerce/blob/1c7a27b7fe51209e38cfec7b972c9414c93ab951/plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-products-v2-controller.php#L488 | Currently if returning `null` would result in a `warning` or `fatal` |

Then I found these 2 cases where `is_array` and `is_iterable` are used:
https://github.com/woocommerce/woocommerce/blob/9ef801f1c3786a8d394edb99b202d4947986f520/plugins/woocommerce/src/StoreApi/Utilities/ProductItemTrait.php#L43
Won't be affected, when returning `null`, it would return an empty array. Same as when returning `array()`, it will pass the `is_iterable` condition but the end result is the same because it will not get to the `foreach` loop.

https://github.com/woocommerce/woocommerce/blob/87c8b886125a025aa3facccb19c0520d8de0c505/plugins/woocommerce/includes/wc-product-functions.php#L498
I think this one could be affected. Right now, when returning `null` it will return an empty string, but if we return `array()` it will get inside the `if` condition and we could potentially return an HTML list without any elements inside. I think we could update the condition to check also whether the `$variation_attributes` is empty.

jorgeatorres:
This is awesome @albarin! 💯

> I think this one could be affected. Right now, when returning null it will return an empty string, but if we return array() it will get inside the if condition and we could potentially return an HTML list without any elements inside. I think we could update the condition to check also whether the $variation_attributes is empty.

Agreed. That's why I called that out "a few comments ago", because it seemed both odd and possibly affected by the changes. If we can update that in this PR too, that would be best. wdyt?

albarin:
@dwainm @jorgeatorres I've pushed new changes and updated the PR description. I think this can be reviewed now.

But I've been thinking if with this we may potentially be breaking custom code added via `woocommerce_product_variation_get_attributes` 🤔

jorgeatorres:
> But I've been thinking if with this we may potentially be breaking custom code added via `woocommerce_product_variation_get_attributes` 🤔

Could you expand on this a bit? You mean code that hooks onto the filter and returns something that _is not_ an array and that we would end up rewriting as an empty array? I'm probably missing something, but not sure what would be the possible use of that.

albarin:
> You mean code that hooks onto the filter and returns something that _is not_ an array and that we would end up rewriting as an empty array?

@jorgeatorres I mean exactly that ☝️ 
I know it would be a weird use of the hook, but you never know how it's being used out there 😅 I just wanted to discuss the concern and agree I don't see the usefulness of that.

jorgeatorres (approved):
Thanks for all the work that went into this @albarin!

Code looks good to me and provided test scenarios pass.

I think the concern about plugins using filters tied to `get_prop()` to override the value type of a property is valid, but if plugins are hooking at such low level not just to override a property's value but also its type, then it'd be very hard to have a coherent API because our functions might end up operating on... pretty much anything.

Unless we have reason to think this might be a problem (maybe we can do a quick search in marketplace extensions?) I'm good with enforcing the "expected" type for the `get_attributes()` function.

Would like to see what @dwainm thinks, once he has a chance to review.

dwainm (approved):
Thanks for all your effort @albarin . I enjoyed following along, and I'm sorry for missing out on the last comments after the PR was re-opened.

I've tested it again and it works. The PR code and the research findings make sense.

I think that we're doing well in enforcing the API's that we've already [agreed to](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/abstracts/abstract-wc-product.php#L547) .

I can't imagine anyone using the low-level APIs to specifically alter get_attributes to return anything other than the array type, so unless there's an easy way to figure out which plugins do this, I doubt the effort will be worth it. To add more weight to this idea, just looking at the research findings again. Many of those items actually require an array, specifically the [for each loop](https://github.com/woocommerce/woocommerce/blob/cf33aa5177f3453ae09c3c94c97e5c3b120e483b/plugins/woocommerce/includes/abstracts/abstract-wc-product.php#L1828) and so I personally think we should enforce the array type as you did in this PR.

PR #11455, opened 2016-07-17
Improved Structured Data

opportus:
## Switch to JSON-LD

There are many  benefits of using JSON-LD over Microdata.
JSON-LD makes working around/with the structured data much more easy, clean and flexible for other plugins, users or contributors.

**Pros**
- Modifying the default shema.org is now possible with a filter hook, avoiding heavy changes in functions/templates
- Infinitely more easy to read than Microdata
- The shema.org stuff is centralized in a single place instead of being scattered all over the project
- The structured data gets finally independant from templates
- Google seems to favour JSON-LD over every other formats, for good reasons

~~**Cons**~~
- ~~The switch itself, but I made it for you...~~

Note BTW that we've already [integrated it to Storefront](https://github.com/woothemes/storefront/commit/f05eea71176479a344ddf025b8e245ca383325cd) a while ago.
## Integration

I've choosed what seems the most appropriate to me - a dedicated class which should get more lines of code over time...
If for some reasons, you don't like this approach, let me know and I'll integrate it the way you want.
## Implementation

The "item type specific" methods generating the structured data are fired at specific action hooks, such as `woocommerce_single_product_summary` for single product, or `woocommerce_review_meta` for reviews - this avoid remaking a loop... From this point, the way to implement the rest of my changes becomes restricted.
I've also added a structured data generator in product categories... I don't know if there is a good reason to not have implemented it before but [Google says](https://developers.google.com/search/docs/guides/intro-structured-data#multiple-entities-on-the-same-page) that we should do it...

PR #55561, opened 2025-02-17
[WIP] Add priceType in all cases in structured data

Aljullu:

Follow-up of https://github.com/woocommerce/woocommerce/pull/55061.

See also https://github.com/woocommerce/woocommerce/issues/55043.

According to https://github.com/woocommerce/woocommerce/pull/55061#issuecomment-2653627414, after recent changes, the listed price is no longer showing up in Yandex for products on sale (only the sale price appears). While I wasn't able to reproduce, this is a PR that _might_ fix it.

The changes in this PR go against the [Google recommendations](https://developers.google.com/search/docs/appearance/structured-data/merchant-listing#sale-pricing-example), but apparently the [Google Rich Snippets tester tool](https://search.google.com/test/rich-results) doesn't complain about it. So if we find out that this PR indeed fixes the issue in Yandex, I think it would be worth fixing it.

1.

### DIFF
--- a/plugins/woocommerce/changelog/fix-55043-list-price-structured-data
+++ b/plugins/woocommerce/changelog/fix-55043-list-price-structured-data
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Add priceType in all cases in structured data
--- a/plugins/woocommerce/includes/class-wc-structured-data.php
+++ b/plugins/woocommerce/includes/class-wc-structured-data.php
@@ -243,6 +243,7 @@ class WC_Structured_Data {
 						'priceSpecification' => array(
 							array(
 								'@type'                 => 'UnitPriceSpecification',
+								'priceType'             => 'https://schema.org/ListPrice',
 								'price'                 => wc_format_decimal( $lowest, wc_get_price_decimals() ),
 								'priceCurrency'         => $currency,
 								'valueAddedTaxIncluded' => wc_prices_include_tax(),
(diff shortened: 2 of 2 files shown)


ivantipov:
Hi @Aljullu,

Sorry I haven't been clear – I had previously tried it with and without the priceType for all prices and it didn't work. After some research, it doesn't appear Yandex supports json-ld at all for products and instead uses the older itemprop format in the HTML (although changing the order of prices did fix the original issue). Was this format ever used in WooCommerce/when was it removed in favour of json-ld?

In any case, the client is happy to just keep the sale price – perhaps your best course of action is to just follow the Google standards.

https://yandex.ru/support/metrica/publishers/schema-org/json-ld.html?ysclid=m7ad0oxwrq938660784&lang=en

https://yandex.ru/support/webmaster/supported-schemas/goods-prices.html?ysclid=m7acsbddyk359374745&lang=en

Aljullu:
Thanks for the extra details, @ivantipov!

> I had previously tried it with and without the priceType for all prices and it didn't work.

Good to know. I will go ahead and close this PR, then.

> After some research, it doesn't appear Yandex supports json-ld at all for products and instead uses the older itemprop format in the HTML (although changing the order of prices did fix the original issue). Was this format ever used in WooCommerce/when was it removed in favour of json-ld?

Apparently the switch from microdata to JSON-LD was introduced in WooCommerce 3.0 (https://github.com/woocommerce/woocommerce/pull/11455), which is almost 9 years old. So it's likely unrelated to these most recent changes.

> In any case, the client is happy to just keep the sale price – perhaps your best course of action is to just follow the Google standards.

Yes, makes sense. Given that we couldn't reproduce and we didn't receive any other reports, I will go ahead and close this for now.

If you ever encounter a new issue, don't hesitate to raise it, though. And thanks again for your help resolving this!

PR #55577, opened 2025-02-17
Create `@woocommerce/blocks-checkout-events` package to allow event subscription

### LINKED ISSUE #44213: [Enhancement]: Make CheckoutEventsContext available to custom components

Make the [CheckoutEventsContext](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce-blocks/assets/js/base/context/providers/cart-checkout/checkout-events/index.tsx) available to custom components that extend the checkout block.

There don't appear to be any great alternatives as the checkout events are stored in a local state that's not available via store data. The checkout events are available to internal developers or to payment method integrations only.

(Please read the Additional context section so that the following code example makes sense).

The following could work, but it doesn't seem as clean of a solution.

```javascript
const MyCustomComponent = () => {
    const [openModal, setOpenModal] = useState(false);
    const {
            hasError
        } = useSelect(select => {
            const store = select(CHECKOUT_STORE_KEY);
            return {
                hasError: store.hasError()
            }
        });
    
    useEffect(()=>{
        if(hasError){
            setOpenModal(true);
        }
    }, [hasError]);
    
    if(openModal){
        return <VerificationModal/>;
    }
    return null;
}
```

### Additional context

The [Extending the WooCommerce Checkout Block Tutorial](https://developer.woo.com/2023/08/07/extending-the-woocommerce-checkout-block-to-add-custom-shipping-options/) provides an example of how a 3rd party developer could add a custom component to the checkout block using the `registerCheckoutBlock` helper function. While this is a helpful example, the custom component is limited in what it can do since it only has the `checkoutExtensionData` and `extensions` arguments available.

There will likely be scenarios in which a custom component needs access to checkout events like `onCheckoutValidationBeforeProcessing`, `onCheckoutSuccess`, `onSubmit` etc. 

Here is one such example:

A developer has created a plugin that requests identity verification based on certain cart conditions. Using action `woocommerce_store_api_cart_errors`, an error is added if verification is required. The developer wants to add a custom component that adds a callback to the `onCheckoutFail` event. Based on the existence of the error,  a confirmation modal with identity fields is rendered. Once the confirmation button is clicked, the promise is fulfilled and the payment should be re-attempted.

opr:
### Reasoning for changes

This PR aims to unblock third party developers who want to subscribe to various events during the checkout process such as `onCheckoutValidation`, `onCheckoutFail` and `onCheckoutSuccess`. The current API only allows payment methods to subscribe to these events. Recent requests to make `useCheckoutEventsContext` available to third parties prompted us to look into this, and I feel that the approach taken in this PR is better than refactoring the existing system.

##### Module boundaries and storing observers

The main reason is that trying to expose an API that works is impossible due to the way the observers are stored and how module caching creates new instances of things like context, so the internal components that import from `assets/js/base/context` wouldn't have access to the same things added via a helper function exposed in `@woocommerce/blocks-checkout`. We can't just import from `@woocommerce/blocks-checkout` internally because of circular dependencies, which we need to get around with relative imports.

Trying to make the existing system work was a large part of this effort, but unfortunately I kept running into this issue.

##### Reasoning with existing code

So if we can't make the existing system work, why not just refactor it to not rely on context? Truth be told, this is probably doable, but I found it very hard to reason with the existing code, and always have. The event emitter system has been a big source of confusion for me since I started working on WC Blocks, and the approach taken here, while using the same design pattern, aims to be clearer and easier to follow the flow of the program.

#### Is this an existing pattern anywhere?

As explained in https://github.com/woocommerce/woocommerce/pull/55577/files#r1958857616 this pattern (with some differences) exists in Gutenberg, specifically in the `@wordpress/data` `subscribe` function. Please see that comment for more info.

#### Why create a v2 of the event emitter?

Because refactoring all events to use the new system in one PR is a lot. The scope of this PR is already extremely broad and trying to keep its surface area as small as possible is important to help reviewers.

Specifically, the changes I am proposing in this PR are:

- Create a new script and alias called `@woocommerce/blocks-checkout-events`
  - This will be the location that extensions get the event registration functions from.
  - Unfortunately, the script also grants access to the `checkoutEventsEmitter` and `createEmitter` functions, in theory extensions could emit checkout events which is not ideal, but is unavoidable due to the module caching problem preventing us from moving the context-based approach over.
- Create a new `EventEmitter` in `assets/js/events`.
	- This will handle event subscriptions and notifying the listeners when an event they are subscribed to is emitted.
- Create a `checkoutEventsEmitter`
	- This is intended to be one of several event emitters that handle one area of concern. This one is specifically handling checkout events. We will have another one for handling payment events. If we identify the need for emitting other events, e.g. cart events, a new emitter can be created using the `EventEmitter` above.
- Move all the existing observer type guards to the `@woocommerce/types` module.
	- Since these are used in multiple places it makes sense to locate them "centrally".
- Update the `wc/store/checkout` data store to use the new `checkoutEventsEmitter` for emitting the `onCheckoutValidation`, `onCheckoutSuccess`, and `onCheckoutFail` events.
- Update `usePaymentMethodInterface` to pass these new event registration functions to payment methods. Previously they were derived from `useCheckoutEventsContext` but now they are derived from`@woocommerce/blocks-checkout-events` ⚠️  This is probably the change with highest stakes, so should be tested extensively. 

Closes: #44213

1. Install and activate [Stripe](https://wordpress.org/plugins/woocommerce-gateway-stripe/) 
2. Add an item to your cart and go to the Checkout block.
3. Try checking out with one of [Stripe's failing test cards](https://docs.stripe.com/testing#declined-payments)
4. Ensure the checkout fails and you see an error.
5. Try checking out with one of the valid cards. Ensure it is successful
6. Check out with COD and BACS, make sure they work too.
7. Try any other payment methods you have access to, e.g. Square, Mollie, Klarna.

##### `onCheckoutValidation` fails

1. Go to the Checkout block and enter the following into your JS console:
```js
wc.blocksCheckoutEvents.checkoutEvents.onCheckoutValidation( () => {
    return { type: 'failure', errorMessage: 'validation failed' };
} );
```

2. Try to check out. Ensure it fails with the error message.
3. Without reloading, enter it again, but change the error message property to something else.
4. Try to check out again, ensure both errors appear.

##### `onCheckoutValidation` success

1. Go to the Checkout block and enter the following into your JS console:
```js
wc.blocksCheckoutEvents.checkoutEvents.onCheckoutValidation( () => {
    return { type: 'success' };
} );
wc.blocksCheckoutEvents.checkoutEvents.onCheckoutValidation( () => {
    return true;
} );
wc.blocksCheckoutEvents.checkoutEvents.onCheckoutValidation( () => {
   console.log( 'doing nothing' );
} );
```

2. Try to check out, ensure checkout is successful.

##### `onCheckoutValidation` throws

1. Go to the Checkout block and enter the following into your JS console:
```js
wc.blocksCheckoutEvents.checkoutEvents.onCheckoutValidation( () => {
    throw new Error( 'error, but should not block execution' );
    return { type: 'success' };
} );
```

2. Try to check out, ensure checkout is successful. This works because we only want to log errors, not block checkout because of them.

#### `onCheckoutSuccess` success

1. Go to the Checkout block and enter the following into your JS console:
```js
wc.blocksCheckoutEvents.checkoutEvents.onCheckoutSuccess( () => {
    return { type: 'success' };
} );
```

2. Try to check out, ensure checkout is successful. 

#### `onCheckoutFail`

1. Go to the Checkout block and enter the following into your JS console:
```js
wc.blocksCheckoutEvents.checkoutEvents.onCheckoutFail( () => {
    return { type: 'success' };
} );
```

2. Try to check out using one of [Stripe's failing test cards](https://docs.stripe.com/testing#declined-payments).
3. Ensure checkout fails but only the Stripe error shows.
4. Reload the page and instead enter the following into your JS console:

```js
wc.blocksCheckoutEvents.checkoutEvents.onCheckoutFail( () => {
    return { type: 'error', message: 'custom error', messageContext: 'wc/checkout/payments' };
} );
```

5. Try to check out using one of [Stripe's failing test cards](https://docs.stripe.com/testing#declined-payments).
6. Ensure you see "custom error" in the payments error context.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


docs/cart-and-checkout-blocks/checkout-payment-methods/checkout-flow-and-events.md
```diff
@@ -161,14 +161,15 @@ There are a bunch of utility methods that can be used related to events. These a
...
+  noticeContexts,
+  responseTypes,
+  shouldRetry,
 } from '@woocommerce/base-context';
-};
+import {
+  isSuccessResponse,
+  isErrorResponse,
+  isFailResponse,
+} from '@woocommerce/types';
```
opr, line 172: I moved all these into the type-guards dir

plugins/woocommerce-blocks/assets/js/base/context/hooks/payment-methods/use-payment-method-interface.ts
```diff
@@ -40,11 +41,11 @@ export const usePaymentMethodInterface = (): PaymentMethodInterface => {
 		onCheckoutAfterProcessingWithSuccess,
 		onCheckoutAfterProcessingWithError,
 		onSubmit,
-		onCheckoutSuccess,
-		onCheckoutFail,
-		onCheckoutValidation,
 	} = useCheckoutEventsContext();
 
+	const { onCheckoutValidation, onCheckoutSuccess, onCheckoutFail } =
```
opr, line 46: These values are passed to payment methods. Previously they were derived from `useCheckoutEventsContext` but now they are derived from `checkoutEvents` which is an event emitter created with the "checkout" events. See `plugins/woocommerce-blocks/assets/js/events/checkout-events.ts`

plugins/woocommerce/src/Blocks/AssetsController.php
```diff
@@ -72,6 +72,7 @@ public function register_assets() {
 		$this->api->register_script( 'wc-cart-checkout-vendors', $this->api->get_block_asset_build_path( 'wc-cart-checkout-vendors-frontend' ), array(), true );
 		$this->api->register_script( 'wc-cart-checkout-base', $this->api->get_block_asset_build_path( 'wc-cart-checkout-base-frontend' ), array(), true );
 		$this->api->register_script( 'wc-blocks-checkout', 'assets/client/blocks/blocks-checkout.js' );
+		$this->api->register_script( 'wc-blocks-checkout-events', 'assets/client/blocks/blocks-checkout-events.js' );
```
opr, line 75: Wondering about naming here, and whether it's necessary to export it as its own script, we could maybe just put this in `blocks-checkout`, does it make sense to have it as a standalone script?

plugins/woocommerce-blocks/tests/js/jest.config.json
```diff
@@ -16,7 +16,8 @@
 		"@woocommerce/block-settings": "assets/js/settings/blocks",
 		"@woocommerce/editor-components(.*)$": "assets/js/editor-components/$1",
 		"@woocommerce/blocks-registry": "assets/js/blocks-registry",
-		"@woocommerce/blocks-checkout": "packages/checkout",
+		"@woocommerce/blocks-checkout$": "packages/checkout",
```
opr, line 19: `$` needed to prevent `blocks-checkout-events` matching the shorter pattern.

plugins/woocommerce-blocks/assets/js/types/type-guards/observers.ts
opr, line 1: I think it makes sense to have these type guards in the `@woocommerce/types`.

plugins/woocommerce-blocks/assets/js/events/event-emitter.ts
```diff
@@ -0,0 +1,121 @@
...
+export interface EventListenerWithPriority {
+	listener: EventListener;
+	priority: number;
+}
+/**
+ * Create an event emitter.
+ *
+ * @return The event emitter.
+ */
+export function createEmitter(): EventEmitter {
```
opr, line 48: The inspiration for this was taken from https://github.com/WordPress/gutenberg/blob/trunk/packages/data/src/utils/emitter.ts#L14 which drive's `@wordpress/data`'s `subscribe` however we currently don't need pause/resume functionality and we need to add a way to emit events and fail fast on error (`emitWithAbort`)

The way this is designed to be used is to create a new event emitter for each distinct set of events you want to handle, e.g. this PR adds a `checkoutEventsEmitter`, we could also have a `paymentEventsEmitter` and a `cartEventsEmitter` eventually.

The Gutenberg implementation also stores the listeners in a `Set` which is not something we can do as we need our listeners to have priority. Therefore, we need to sort them. Given a `Set` is unordered, this sort would need to happen on every emit, using an array instead allows us to sort once on insert, improving performance.

plugins/woocommerce-blocks/assets/js/events/event-emitter.ts
```diff
@@ -0,0 +1,121 @@
...
+			};
+		},
+
+		emit: async ( eventName: string, data: unknown ) => {
+			return await notifyListeners( eventName, data );
+		},
+
+		emitWithAbort: async ( eventName: string, data: unknown ) => {
+			return await notifyListenersWithAbort( eventName, data );
+		},
```
opr: We have this pattern in the current implementation, but I'm wondering if we could skip `emitWithAbort` and instead use a flag on `emit` to signal it should stop processing the first time an error/fail response is received?

Probably an optimization for later, but something to think about at least.

plugins/woocommerce-blocks/assets/js/data/checkout/types.ts
```diff
@@ -1,19 +1,16 @@
 /**
  * External dependencies
  */
-import type { Notice } from '@wordpress/notices/';
-import { DataRegistry } from '@wordpress/data';
-import { FieldValidationStatus } from '@woocommerce/types';
+import type { WPNotice } from '@wordpress/notices/build-types/store/selectors';
```
opr, line 4: This `WPNotice` change is a TS change unrelated to rest of the PR. Wee could revert if we want to reduce noise of this PR, but it was an easy fix while I was in this file.

plugins/woocommerce-blocks/assets/js/data/checkout/thunks.ts
```diff
@@ -99,7 +125,7 @@ export const __internalEmitValidateEvent: emitValidateEventType = ( {
  * to the observer responses
  */
 export const __internalEmitAfterProcessingEvents: emitAfterProcessingEventsType =
-	( { observers, notices } ) => {
+	( { notices } ) => {
```
opr, line 127: observers are all stored in the event emitter now, so we just emit the event and calling them is handled by that component. It seems wrong to pass observers around like this.

plugins/woocommerce-blocks/assets/js/data/checkout/thunks.ts
opr, line 1: This file contains possibly the most important runtime changes, it switches the `__internalEmitValidateEvent` and `__internalEmitAfterProcessingEvents` to use the new event emitter, rather than the old one that uses context.

It also aims to simplify the code slightly and reducing nesting a little. More info inline.

plugins/woocommerce-blocks/assets/js/data/checkout/thunks.ts
```diff
@@ -62,34 +69,53 @@ export const __internalProcessCheckoutResponse = (
...
-	setValidationErrors, // TODO: Fix this type after we move to validation store
+	setValidationErrors,
 } ) => {
 	return ( { dispatch, registry }: CheckoutThunkArgs ) => {
 		const { createErrorNotice } = registry.dispatch( noticesStore );
 		removeNoticesByStatus( 'error' );
-		emitEvent( observers, EVENTS.CHECKOUT_VALIDATION, {} ).then(
-			( response ) => {
-				if ( response !== true ) {
-					if ( Array.isArray( response ) ) {
```
opr, line 78: The previous name `response` was misleading, because this could be multiple observer response**s**.

The isArray check is no longer necessary as `emit` on the new event emitter always returns an array, even if no observer has run. If observers did run, but didn't return anything, `responses` will still be an empty array. Check the `emit` and `notifyListeners` function in `plugins/woocommerce-blocks/assets/js/events/event-emitter.ts`

plugins/woocommerce-blocks/assets/js/data/checkout/thunks.ts
```diff
@@ -62,34 +69,53 @@ export const __internalProcessCheckoutResponse = (
...
-				} else {
+		checkoutEventsEmitter
+			.emit( CHECKOUT_EVENTS.CHECKOUT_VALIDATION )
+			.then( ( responses ) => {
+				// If responses length is 0, then no observer returned a response that wasn't `true` or void, therefore,
+				// we can assume all observers passed and continue to processing. We also need to check if all responses
+				// are of type `success`, so we can skip adding any errors too.
+				if (
+					responses.length === 0 ||
+					responses.every( isSuccessResponse )
```
opr, line 85: In theory, observers can return `{ type: 'success' }` which is the same as returning void or true, but if all responses are of type success we can still continue to processing.

We can also reduce nesting here by returning early if all observers were successful.

plugins/woocommerce-blocks/assets/js/data/checkout/thunks.ts
```diff
@@ -62,34 +69,53 @@ export const __internalProcessCheckoutResponse = (
...
+						errorMessage,
+						validationErrors,
+						context = 'wc/checkout',
+					} ) => {
+						if (
+							// TODO: for a more consistent experience across observing events we should normalize the
+							// return values. For example this one expects `errorMessage` whereas `onCheckoutFail`
+							// expects `message`. It would be good to ensure all observer responses are the same shape
+							// otherwise this leads to confusion when typing internally, and confusion for the consumer.
+							typeof errorMessage === 'string' &&
```
opr, line 103: These checks are needed to prevent TS errors, but there's also a larger TS/DevEx problem here of mismatched observer response names. As seen in the comment, some expect `message` some expect `errorMessage` - we need to make these consistent.

plugins/woocommerce-blocks/assets/js/events/checkout-events.ts
```diff
@@ -0,0 +1,24 @@
...
+export const CHECKOUT_EVENTS = {
+	CHECKOUT_SUCCESS: 'checkout_success',
+	CHECKOUT_FAIL: 'checkout_fail',
+	CHECKOUT_VALIDATION: 'checkout_validation',
+};
+
+export const checkoutEventsEmitter = createEmitter();
+
+export const checkoutEvents = {
+	onCheckoutValidation: checkoutEventsEmitter.createSubscribeFunction(
```
opr, line 24: The idea I have here is to create individual emitters (`checkout`, `payment` to begin with) and for us to provide these "helper" functions, like we have on the `useCheckoutEventsContext` that allow third parties to subscribe to an event just by passing their callback to the function. The exact same pattern we have now, except events are created and stored in their own emitter objects, rather than all on the same context.

Potential improvement here is providing a map of helper function names and event names to `createEmitter` and having the helper functions created automatically.

plugins/woocommerce-blocks/assets/js/data/checkout/utils.ts
```diff
@@ -131,8 +137,11 @@ export const runCheckoutSuccessObservers = ( {
 	dispatch: ActionCreatorsOf< ConfigOf< typeof checkoutStore > >;
 	createErrorNotice: typeof originalCreateErrorNotice;
 } ) => {
-	let successResponse = null as null | Record< string, unknown >;
-	let errorResponse = null as null | Record< string, unknown >;
+	let successResponse = null as null | ObserverSuccessResponse;
```
opr, line 138: Small TS update as success and error responses should be one of the observer types, which now extends `Record< string, unknown >`.

mikejolley (commented):
- I looked over the code and nothing stood out aside from the deprecated events not moving over to the new package (discussed in slack)
- Dev testing steps via console all worked as described. I tested some fails on success event emitters also
- Stripe testing steps worked as described
- Tested core payment methods which are working

May want to get someone else to look over code for final approval but working for me 👍🏻  (I will approve monday after your changed if no-one adds a ✅ before then)

nielslange (approved):
Thanks for working on this, @opr. It took me a while to review this PR, but I finally managed doing so. I've successfully tested all steps and ran additional tests using [PayPal Payments for WooCommerce](https://woocommerce.com/products/woocommerce-paypal-payments/) and all default payments. Code-wise, I could not spot anything unusual. 

Is there a follow-up issue planned to document how third-party developers can subscribe to checkout events without referring back to this PR?

PR #55580, opened 2025-02-18
Fix coming soon determined too early in lifecycle

ilyasfoo:

Closes https://github.com/woocommerce/team-ghidorah/issues/380

The coming soon optimization in [this PR](https://github.com/woocommerce/woocommerce/pull/54906/files#diff-4036d5a5c66edc139f05d996eec952ce716b944160128e7cfb149b426343c42aR40) was deciding coming soon option too early. Since option calls are cached in memory, this raises the following issues:

1. It impedes the ability for 3pd and hosting providers to override the option value.
1. Feature class is not loaded by this time, making it limiting to build logic around it even if we hook into the option filter earlier.

As an example, the nightly core build fails in WordPress.com sites since the [override hook](https://github.com/Automattic/wc-calypso-bridge/blob/fd00cfcc99b670fe1be53ae530017741c2d2468a/includes/class-wc-calypso-bridge-coming-soon.php#L40) is not set up when `woocommerce_coming_soon` option is queried, which gets cached in memory.

ChatGPT confirmed all hooks newly wrapped into `plugins_loaded` are called after:

image
 
#### Regression test

1. Enable feature flag `coming-soon-newsletter-template`
3. Go to `Appearance > Editor > Templates > Page: Coming soon`
4. Select different templates under `Design` panel
5. Ensure each loaded correct UI, indicating styles and fonts are loaded as expected ([example with TT25 theme](https://github.com/user-attachments/assets/f6374dee-35fc-4444-b100-3aa2b75629f3))
6. Select any design and save
7. In an incognito window, open up the frontend shop page
8. Observe the page looks correct ([example](https://github.com/user-attachments/assets/b85b7730-5d2f-4928-ac06-35f1f77fc1c3))
9. Change to a non-FSE template, such as Storefront
10. Refresh the incognito window and ensure the coming soon page looks correct

#### Optional: Testing in WordPress.com

1. In a WordPress.com business or ecommerce site, install the build from this PR ([woocommerce.zip](https://github.com/user-attachments/files/18841679/woocommerce.zip))
2. Set the site to coming soon
3. Go to `/_cli` and run `option delete woocommerce_coming_soon` to ensure the option is deleted if it exist so that we'll be able to observe the coming soon value coming from the override hook
4. In an incognito window, open the `/shop`
5. Observe it shows coming soon page

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/fix-coming-soon-optimization-lifecycle-hook
+++ b/plugins/woocommerce/changelog/fix-coming-soon-optimization-lifecycle-hook
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix coming soon hooks to be after plugins_loaded hook
--- a/plugins/woocommerce/src/Internal/ComingSoon/ComingSoonRequestHandler.php
+++ b/plugins/woocommerce/src/Internal/ComingSoon/ComingSoonRequestHandler.php
@@ -37,17 +37,20 @@ class ComingSoonRequestHandler {
 	final public function init( ComingSoonHelper $coming_soon_helper ) {
 		$this->coming_soon_helper = $coming_soon_helper;
 		// Hook into plugins_loaded to ensure features are initialized to determine coming soon status.
-		add_action( 'plugins_loaded', function() {
-			// Skip if the site is live.
-			if ( $this->coming_soon_helper->is_site_live() ) {
-				return;
-			}
+		add_action(
+			'plugins_loaded',
+			function () {
+				// Skip if the site is live.
+				if ( $this->coming_soon_helper->is_site_live() ) {
+					return;
+				}
 
-			add_filter( 'template_include', array( $this, 'handle_template_include' ) );
-			add_filter( 'wp_theme_json_data_theme', array( $this, 'experimental_filter_theme_json_theme' ) );
-			add_action( 'wp_enqueue_scripts', array( $this, 'enqueue_styles' ) );
-			add_action( 'after_setup_theme', array( $this, 'possibly_init_block_templates' ), 999 );
-		} );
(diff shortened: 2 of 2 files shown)


rjchow (approved):
For me this is a case of unknown unknowns, but I tested it with JN and Hostinger, and reasoned through what I do know with the assistance of O3-mini and it seems sound.

psealock (commented):
On a JN site, I changed up the template and visit the site in incognito. All is as expected:

image

Then I switch to Storefront and refresh the incognito window. The fonts change (and disappear) but nothing else. Can you reproduce?

image

ilyasfoo:
Good catch, @psealock! I'm able to reproduce this with latest trunk, I think it's not related to this PR. It seems the font color is just too dark - I suspect it's the template inheriting styles from the theme:

image

psealock:
>I'm able to reproduce this with latest trunk

Gotcha, I was expecting the coming soon template to revert to the generic one when going to non FSE theme. So what I'm seeing is expected then?

chihsuan (approved):
Thanks for the fix! The changes look good to me. 👍

ilyasfoo:
> Gotcha, I was expecting the coming soon template to revert to the generic one when going to non FSE theme. So what I'm seeing is expected then?

@psealock Yes, that's correct. It should show the last saved pattern regardless of current theme, or the default pattern if coming soon template is not modified

psealock (approved):
Thanks @ilyasfoo. In this case, this is testing well and code looks good 🚢

PR #55314, opened 2025-02-11
[e2e tests] Restructure project - move coupons, editor and basic specs

adimoldovan:
### Changes proposed in this Pull Request:

Part of pdV5qK-Ie-p2
More context: pcr5Cj-70A-p2

Move multiple specs into newly created folders coupons, customer, editor and basic.
Renamed `wordpress-post.spec.js` to `post-comments.spec.js`.
Removed now empty 🎉 `merchant` and `shopper` folders

### How to test the changes in this Pull Request:

CI should be green.
To run the affected tests locally:

```
pnpm test:e2e coupons customer editor basic
```

PR #55581, opened 2025-02-18
[e2e tests] Fix product delete spec failing when there are not other products left

adimoldovan:

Test started failing in the daily runs: https://github.com/woocommerce/woocommerce/actions/runs/13382699728/job/37377928679#step:9:207

Fix the test 'can permanently delete a product from trash list' for cases when no other products are left after deletion. If no other products exist the page is redirected to Products admin task and the successful deletion notification is not displayed.

With #55314 the order of the tests shifted and the product-delete spec ended up in another shard in the daily runs, where only 5 shards are used. Because of this, no other products that are created by other tests existed, so the product delete test removed the last one.

Also made smaller tweaks from the "since we're here" category, to bring consistency:
- use the data utils to generate the test product
- converted require to import
- removed the describe block

Ensure there are no products in the site, then run the product-delete spec

```
pnpm test:e2e product-delete.spec
```

Ran the daily jobs that failed and it passed: https://github.com/woocommerce/woocommerce/actions/runs/13385462986


Stojdza (approved):
Worked as expected with no products on the site 👍

PR #55585, opened 2025-02-18
[e2e tests] Fix payment setup tasks and move spec into onboarding

adimoldovan:

The payment setup spec started failing in the [daily runs](https://github.com/woocommerce/woocommerce/actions/runs/13382699728/job/37381213737#step:9:411) since the test order changed. The cause was that a previous spec was hiding the tasks list and the payment tasks spec is using a direct url to the task which was not loading the content.

Fixed by adding a step in the fixture to ensure the tasks list is not hidden.

Check CI.

To run the test:

```
pnpm test:e2e payment-setup-task
```

The tests passed in the daily jobs where it they were failing before: https://github.com/woocommerce/woocommerce/actions/runs/13387110197/job/37386419126#step:9:110 (the other 2 failing tests are unrelated and will be addressed in different PRs)


Stojdza (approved):
Lokks good and it works locally as expected 👍

PR #55586, opened 2025-02-18
Fatal error fix: check if `orderby` is an array before using it as a key

### LINKED ISSUE #55269: [Fatal] PHP Fatal error: Uncaught TypeError: Illegal offset type in | wc-template-functions.php:1598
### Prerequisites

### Describe the bug

Issue witnessed in `9.6.0`

[link to code](https://github.com/woocommerce/woocommerce/blob/6f5032a0277a0eb006a442407d9d09c0c370b8be/plugins/woocommerce/includes/wc-template-functions.php#L1598)

```
PHP Fatal error:  Uncaught TypeError: Illegal offset type in /wordpress/plugins/woocommerce/9.6.0/includes/wc-template-functions.php:1598
Stack trace:
#0 /wordpress/core/6.7.1/wp-includes/class-wp-hook.php(324): woocommerce_catalog_ordering('')
#1 /wordpress/core/6.7.1/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters(NULL, Array)
#2 /wordpress/core/6.7.1/wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#3 /wordpress/plugins/woocommerce/9.6.0/src/Blocks/BlockTypes/ClassicTemplate.php(296): do_action('woocommerce_bef...')
#4 /wordpress/plugins/woocommerce/9.6.0/src/Blocks/BlockTypes/ClassicTemplate.php(146): Automattic\WooCommerce\Blocks\BlockTypes\ClassicTemplate->render_archive_product()
#5 /wordpress/plugins/woocommerce/9.6.0/src/Blocks/BlockTypes/AbstractBlock.php(98): Automattic\WooCommerce\Blocks\BlockTypes\ClassicTemplate->render(Array, '', Object(WP_Block))
#6 /wordpress/core/6.7.1/wp-includes/class-wp-block.php(537): Automattic\WooCommerce\Blocks\BlockTypes\AbstractBlock->render_callback(Array, '', Object(WP_Block))
#7 /wordpress/core/6.7.1/wp-includes/class-wp-block.php(517): WP_Block->render()
#8 /wordpress/core/6.7.1/wp-includes/blocks.php(2190): WP_Block->render()
#9 /wordpress/core/6.7.1/wp-includes/blocks.php(2242): render_block(Array)
#10 /wordpress/core/6.7.1/wp-includes/block-template.php(263): do_blocks('<!-- wp:templat...')
#11 /wordpress/core/6.7.1/wp-includes/template-canvas.php(12): get_the_block_template_html()
#12 /wordpress/core/6.7.1/wp-includes/template-loader.php(106): include('/wordpress/core...')
#13 /wordpress/core/6.7.1/wp-blog-header.php(19): require_once('/wordpress/core...')
#14 /wordpress/core/6.7.1/index.php(17): require('/wordpress/core...')
#15 {main}
  thrown in /wordpress/plugins/woocommerce/9.6.0/includes/wc-template-functions.php on line 1598
``` 

### Expected behavior

No fatals expected

### Actual behavior

Fatal errors were witnessed

### Steps to reproduce

We don't have steps to reproduce, but observed 48 instances of this problem on Fri Jan 31, 2025 across 2 sites. Hopefully, someone can dig in and apply a defensive check if appropriate. - To be updated after investigation.

### WordPress Environment

N/A

### Isolating the problem

albarin:

Before using the `$orderby` variable as a key of the `$catalog_orderby_options` we need to check it's not an array (since arrays cannot be used as array keys, see [the docs](https://www.php.net/manual/en/function.array-key-exists.php)). Otherwise, we'll get a fatal error.

Closes #55269 

1. On `trunk`, go to your shop catalog and select an order from the dropdown (you should be redirected to something like `/shop/?orderby=price-desc`).
2. Change the URL to make `orderby` an array: `/shop/?orderby[]=price-desc`.
3. Check that you see a fatal error similar to this one:
```
Fatal error: Uncaught TypeError: Cannot access offset of type array on array in /plugins/woocommerce/plugins/woocommerce/includes/wc-template-functions.php:1642 
```
4. Checkout to this PR branch.
5. Reload the array `orderby` URL and check the fatal error is no longer there.
6. Try to change the URL to add two or more `orderby` criteria:  `/shop/?orderby[]=price-desc&orderby[]=date` and check products are sorted by the first one.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Check if `orderby` is an array before using it as a key to avoid a fatal error.
</details>

<details>

#### Comment 

</details>


plugins/woocommerce/includes/wc-template-functions.php
```diff
@@ -1635,6 +1635,10 @@ function woocommerce_catalog_ordering( $attributes = null ) {
 			unset( $catalog_orderby_options['rating'] );
 		}
 
+		if ( is_array( $orderby ) ) {
+			$orderby = current( array_keys( $catalog_orderby_options ) );
```
coreymckrill: 🤔 It's too bad we can't see in the error message what was in the array that was passed here that caused the fatal error. My guess is that it was probably multiple catalog order options, e.g. `array( 'rating', 'menu_order' )`, because it's valid to specify multiple orderby's this way in `WP_Query`.

What if, instead of falling back on the first item in the enum list for catalog orderby options, we picked the first item in the array that was passed? If the first item isn't actually a valid orderby, it would still get fixed in the next conditional check. Better yet, we could do an array intersect first. Something like:

```php
if ( is_array( $orderby ) ) {
	$orderby = current( array_intersect( $orderby, $catalog_orderby_options ) );
}
```

Wdyt?

albarin: Great idea, I like it! 👌 
I noticed I was getting also a warning `Warning: Array to string conversion` from `WC_Query`, so I fixed that too and added a new test. I had to remove the `(string)` conversion which I think is unnecessary because `wc_clean` and `wp_unslash` already work well with arrays (according to docs).
Would you mind having another look? 🙏

plugins/woocommerce/includes/class-wc-query.php
```diff
@@ -587,7 +587,10 @@ public function get_catalog_ordering_args( $orderby = '', $order = '' ) {
 		// Get ordering from query string unless defined.
 		if ( ! $orderby ) {
 			// phpcs:ignore WordPress.Security.NonceVerification.Recommended
-			$orderby_value = isset( $_GET['orderby'] ) ? wc_clean( (string) wp_unslash( $_GET['orderby'] ) ) : wc_clean( get_query_var( 'orderby' ) );
+			$orderby_value = isset( $_GET['orderby'] ) ? wc_clean( wp_unslash( $_GET['orderby'] ) ) : wc_clean( get_query_var( 'orderby' ) );
+			if ( is_array( $orderby_value ) ) {
+				$orderby_value = $orderby_value[0];
+			}
```
coreymckrill: 🤔 I'm wondering if this is necessary, or if it might even cause a problem, given that [further down](https://github.com/woocommerce/woocommerce/blob/8309364c4f0ac56dda0f42047d91895960c5f9f7/plugins/woocommerce/includes/class-wc-query.php#L601) it expects that `$orderby_value` _could_ be an array, and the second item in the array might actually be an `order` value (`asc` or `desc`).

I don't actually know if this would cause a problem, just thinking of ways this could come back to bite us later...

coreymckrill (commented):
👍 Works great in testing.

* Multiple orderby array items
* Multiple items, with the first one an invalid value (uses the second one)
* Mutliple items, with both being invalid (falls back to the default sorting)

I do have one more concern to discuss (below) before merging.

plugins/woocommerce/includes/class-wc-query.php
albarin: Good point 🤔 
I think it is necessary but only when we are dealing with orderby coming from `$_GET`, otherwise cases like the one in the `test_get_catalog_ordering_args_when_GET_is_array` will fail.
I've changed it here https://github.com/woocommerce/woocommerce/pull/55586/commits/6ad5d2f1b31b5309d7339a13942f5571103de54e to only getting the first element of the array on that case, wdyt?

coreymckrill (approved):

plugins/woocommerce/includes/class-wc-query.php
coreymckrill: 👍 Sounds good to me

PR #55588, opened 2025-02-18
[e2e tests] Fix failing add custom product attributes test

adimoldovan:

The test `can add custom product attributes` started failing in [the daily runs](https://github.com/woocommerce/woocommerce/actions/runs/13382699728/job/37381213737#step:9:367) since the order of the tests changed yesterday. 
The update product action included a check for a response that includes `woocommerce_task_list_reminder_bar_hidden`, which is not happening, probably because the bar list settings are updated by other tests. This setting is not required for this test, (I assume) it was only added to check that all network requests completed.
Relaxing the partial url to be checked will make the test pass.

I also refactored the spec for consistency with the current guidelines:
- removed serial mode - it was unnecessary anyway, there is a single test
- remove the describe block
- used a fixture that creates a unique product using the data util
- extracted duplication into a function - also fixed the nesting steps and the conditional in test lint warnings
- removed dependency of variableProducts utils which we need to deprecate
- renamed functions and variables to respect naming conventions

Check CI, it should be green.
Ran the daily jobs that failed before and the test passed: https://github.com/woocommerce/woocommerce/actions/runs/13391529087/job/37400171068#step:9:220 (ignore other failing tests, they are unrelated and are addressed in different PRs)

To run test test:
```
pnpm test:e2e create-product-attributes
```


Stojdza (approved):
The changes look good and the CI is passing 👍

PR #55597, opened 2025-02-18
Add simplified IAM footer card on top

raja651:

Closes https://github.com/Automattic/woocommerce.com/issues/23091 .

1. Check out this branch.
2. Navigate to /plugins/woocommerce and run `pnpm --filter='@woocommerce/plugin-woocommerce' build`.
3. Visit the [Marketplace](http://localhost:8888/wp-admin/admin.php?page=wc-admin&path=%2Fextensions).
4. Confirm that the new features banner appears above the header and cycles through each feature.
5. Verify that it matches the 2sXEw33oLyTZ7Mu3808sed-fi-4314_154966 for both desktop and mobile views. (Note: We are not implementing arrows for the carousel as part of the MVP, so please ignore their absence in the design.)
6. Scroll the page and confirm that the banner remains sticky along with the header.
7. Close the banner, refresh the page, and ensure that the banner does not reappear.
8. Delete the local storage key wc_featuredBannerDismissed, refresh the page, and confirm that the banner is displayed again.
9. Run localStorage.setItem( 'debug', 'wc-admin:*' ); to output track events in the console
10. Close the banner and confirm the tracks event is fired to register the event
11. confirm Keyboard focus or mouse hover pauses the carousel on the current slide
12. Enable voice over(Command + f5) on your browser and check the accessibility of the carousel
13. Use control + option + left or right arrows to go through the carousel slides and confirm correct labels are read out for the carousel container and slides and close button.

Screenshot 2025-02-19 at 4 18 22 PM

Figma deisgns,( note we are not implementing the arrows for this iteration)
Screenshot 2025-02-25 at 12 59 52 PM

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Add new features dismissible banner for new users

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/admin/client/marketplace/components/banner/banner.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+	useEffect( () => {
+		const dismissed = localStorage.getItem( 'wc_featuredBannerDismissed' );
+		setIsDismissed( dismissed === 'true' );
+	}, [] );
+
+	const handleDismiss = () => {
+		localStorage.setItem( 'wc_featuredBannerDismissed', 'true' );
+		setIsDismissed( true );
+	};
```
KokkieH: Should we record a Tracks event if they dismiss the banner?

plugins/woocommerce/client/admin/client/marketplace/components/footer/footer.tsx
```diff
@@ -12,23 +12,23 @@ import './footer.scss';
 import IconWithText from '../icon-with-text/icon-with-text';
 import { MARKETPLACE_HOST } from '../constants';
 
-const refundPolicyTitle = createInterpolateElement(
+export const refundPolicyTitle = createInterpolateElement(
 	__( '30-day <a>money-back guarantee</a>', 'woocommerce' ),
 	{
 		// eslint-disable-next-line jsx-a11y/anchor-has-content
 		a: <a href={ MARKETPLACE_HOST + '/refund-policy/' } />,
```
KokkieH, line 19: Could we add Tracks to these links as well, differentiating between clicks in the footer and clicks in the top banner? I think it should work to re-use the new `<TrackedLink>` component I'd added for this.

Also fine if it's done in a separate PR, but we do need tracking if we want to measure the success of the banner.

KokkieH (commented):
Visually it looks and works as described.

The A11Y isn't great, though:

- There's no indication to screen readers that this is a dynamic element that changes it's content.
- If focus is on a slide with a link, and the slide advances, focus is thrown back to the skip-links at the very top of the DOM.

I'm not sure of the best way to handle this, though. Maybe if someone lands on the banner via keyboard nav, we pause it, and support pressing arrow keys to cycle it, along with appropriate announcements and focus handling? 

Also left some comments regarding recording clicks, but that could be done in a separate task.

Dan-Q:
Accessibility thoughts, [courtesy of the WAI](https://www.w3.org/WAI/ARIA/apg/patterns/carousel/):
- focussing a link in the banner should pause the cycle
    - personally, I'd prefer if it _also_ paused on mouseover/hover; this also helps with a11y for people who e.g. use their cursor to follow the words they're reading
- the banner (carousel container) should have `aria-roledescription="carousel"` and something like `aria-label="Features of the marketplace"` (something that say what the component _contains_, do not include the word "carousel")
- it should also get `aria-atomic="false"` and `aria-live="polite"`; this tells a11y tech that the thing will change (by itself), and that they _can_ announce it (when it's convenient), but that they _shouldn't_ announce other changes at the same time
- the individual slides should get `aria-roledescription="slide"`
- ideally, for best semantic support, the carousel container should be a `<ul>` and each item in it a `<li>`, rather than naked divs (this is a _list_ of features, right?)

Ideally, it should be possible to pause the rotation, too. Perhaps a ⏯️ button next to the close button?

Once all of those are done, it might be worth a re-test with a screenreader to see how usable it is.

plugins/woocommerce/client/admin/client/marketplace/components/footer/footer.tsx
raja651: I tried to use <TrackedLink> have couple of questions?
 1. Is there a reason why the targetUrl is called via event instead of addding it directly to hred
 ```
 <Link
						onClick={ () => {
							if ( onClickCallback ) {
								onClickCallback();
							} else {
								recordEvent( eventName );
							}
							window.location.href = getAdminLink( targetUrl );
							return false;
						} }
						href=""
						type="wc-admin"
					/>
```

2. getAdminLink( targetUrl ) shouldn't  this be done in the compoent that is calling it instead of doing it here otherwise external urls will not work.

KokkieH: If the link type is wc-admin, there's a special link handler function as part of the `Link` component that prevents default actions.

> getAdminLink( targetUrl ) shouldn't this be done in the compoent that is calling it instead of doing it here otherwise external urls will not work.

As all the links I were adding were to push people to the in-app marketplace I never even considered this. So yeah, this won't work for external links in its current form.

raja651: > If the link type is wc-admin, there's a special link handler function as part of the Link component that prevents default actions.

If i command click links to open it in a new tab it opens the current url again instead of triggering the callback so i'm wondering how i can use this for external links. I will try a few things and come back

plugins/woocommerce/client/admin/client/marketplace/components/banner/banner.tsx
raja651: yeah make sense did it in [d388d63](https://github.com/woocommerce/woocommerce/pull/55597/commits/d388d63754ff16c795b99a7dc69abdc556aa3508)

plugins/woocommerce/client/admin/client/marketplace/components/footer/footer.tsx
raja651: I discussed with Herman on this, we are planning to do this in a separate PR as the changes to `<TrackedLink>`  need more testing.

raja651:
@KokkieH @Dan-Q 

> I'm not sure of the best way to handle this, though. Maybe if someone lands on the banner via keyboard nav, we pause it, and support pressing arrow keys to cycle it, along with appropriate announcements and focus handling?

We had a design to use arrows on the carousel, but that is not implemented for MVP. So i went with pausing the slide rotation on hover and keyboard focus.

Some of the  Accessibility Improvements made on this commit [da99094](https://github.com/woocommerce/woocommerce/pull/55597/commits/da9909498439cdbc7e4e792c07abc2976ee23094) :
♿ Added ARIA roles/labels for carousel structure
📜 Semantic list structure ul and li
📌 Slide position tracking (aria-posinset/setsize)
🔔 Live content announcements (aria-live)
⏸️ Pause on hover/focus & keyboard navigation
📝 Text alternatives for screen readers

PS: Deepseek helped me create this summary of fixes 🥇

raja651:
I pushed a [fix](https://github.com/woocommerce/woocommerce/pull/55597/commits/118ea2e76643632f5928f82feb644a61eb04a64a) for vendor name overflow on product cards 
Screenshot 2025-02-20 at 7 24 24 PM

plugins/woocommerce/client/admin/client/marketplace/components/footer/footer.tsx
KokkieH: Changes are being made in #55740

KokkieH (approved):
Thanks for the changes.

Added https://github.com/woocommerce/woocommerce/pull/55597/commits/bb53a8360d408980d0689ce79feb98ac313d1a8a with the changes from our pairing session to change the carousel state with arrow buttons. Confirmed focus management and screen reader announcements work properly.

PR #55600, opened 2025-02-18
Product Collection: Move FilterableControl on top of order by controls

Aljullu:

Follow-up of https://github.com/woocommerce/woocommerce/pull/54166.

In #54166 we moved the _Order by_ controls of the Product Collection block up, but they were put before the `<FilterableControl />`, that made it so the order of the controls was inconsistent between when editing a post/page and when editing the template.

1. Create a post or page and add the Product Collection block. Select the _Create your own_ option.
2. Verify the _Query type_ appears before the _Order by_ control.

Before | After
--- | ---
imatge | imatge


gigitux (approved):
LGTM!

I would update the testing instructions to specify that the tester should click on “Create Your Own.” In some configuration like "Featured Products", `Query Type` is not rendered.

Aljullu:
> I would update the testing instructions to specify that the tester should click on “Create Your Own.” In some configuration like "Featured Products", `Query Type` is not rendered.

Ups, good catch! I updated the testing steps. :slightly_smiling_face: Thanks, @gigitux!

PR #55605, opened 2025-02-18
"Grow your business" marketplace promo

andfinally:

Closes 23109-gh-Automattic/woocommerce.com.

1. Switch to this branch.
2. Make sure `Enable tracking` is enabled in `wp-admin/admin.php?page=wc-settings&tab=advanced&section=woocommerce_com`.
3. Enter `localStorage.setItem( 'debug', 'wc-admin:*' );` in your JS console to enable logging of tracks events.
4. Go to `wp-admin/admin.php?page=wc-admin&task=marketing`.
5. Scroll to the bottom and observe the promo. Check that it looks OK on different viewports.
6. You should see a tracks event `wcadmin_task_marketing_marketplace_promo_shown` in your debug console.
7. Clicking on the button should take you to `wp-admin/admin.php?page=wc-admin&tab=extensions&path=%2Fextensions&category=marketing-extensions`.
8. You should see a tracks event `wcadmin_task_marketing_marketplace_promo_clicked` in the console.
9. Disable tracking in `WooCommerce > Settings > Advanced > WooCommerce.com`.
10. Refresh the Marketing task page.
11. You should not see the promo.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### Screenshots

image


plugins/woocommerce/client/admin/client/task-lists/fills/Marketing/TaskPromo.tsx
```diff
@@ -0,0 +1,70 @@
...
+						as="h2"
+						className="woocommerce-task-card__title"
+					>
+						{ title }
+					</Text>
+				</CardHeader>
+			) }
+			<CardBody>
+				<div className="woocommerce-plugin-list__plugin-logo">
+					<img src={ iconSrc as string } alt={ iconAlt as string } />
```
raicem: I believe we have an opportunity to be more concise with our types here. 

We accept `iconSrc` and `iconAlt` as optional props. Because they are optional, we need to assure TypeScript that this value is not not or undefined with `as string`. 

How about we pass the `iconSrc` to `TaskPromo`? 
In `client/admin/client/task-lists/fills/Marketing/index.tsx`
```tsx
<TaskPromo
   // other props
   iconSrc={ `${ WC_ASSET_URL }images/woo-app-icon.svg` }
   iconAlt={ __( 'Woo icon', 'woocommerce' ) }
/> 
```
That way, we're sure it's a string so we don't need to use `as string` to assure types. 
```suggestion
					<img src={ iconSrc } alt={ iconAlt } />
```

plugins/woocommerce/client/admin/client/task-lists/fills/Marketing/index.tsx
```diff
@@ -164,6 +165,20 @@ const Marketing: React.FC< MarketingProps > = ( { onComplete } ) => {
 		actionTask( 'marketing' );
 	};
 
+	useEffect( () => {
+		if ( window.wcTracks.isEnabled ) {
+			recordEvent( 'task_marketing_marketplace_promo_shown', {
+				task: 'marketing',
+			} );
+		}
+	}, [] );
```
raicem: Just an idea: maybe instead of checking if `window.wcTracks.isEnabled` is true, we can move the `recordEvent` call into the `TaskPromo` component. So when it gets rendered, we fire the Tracks event.

raicem (approved):
Works as expected! I have left two non-blocking comments.

plugins/woocommerce/client/admin/client/task-lists/fills/Marketing/TaskPromo.tsx
andfinally: TS doesn't like it when you try to use an optional prop with a default value in an attribute – it complains `Type string | undefined is not assignable to type string | undefined`. As a node, such a prop is fine, for example `name`, which we define in exactly the same way as `iconAlt`. 

None of the alternative type solutions suggested by Perplexity or Copilot work for me, and I want to keep these props as optional, so in https://github.com/woocommerce/woocommerce/pull/55605/commits/ddda7fe121d136b6dd2417fc2ff0e9533f8faa2c I've introduced what feels like an otherwise unnecessary condition to silence the error instead of this declaration.

plugins/woocommerce/client/admin/client/task-lists/fills/Marketing/index.tsx
andfinally: That is neater, thanks! Moved in https://github.com/woocommerce/woocommerce/pull/55605/commits/728646cd99122481f129d8646c082d922950fe6a.

PR #55615, opened 2025-02-19
Improve slow queries for the log list table when using the database logger

### LINKED ISSUE #54538: [Performance]:  Slow log query selector for DB stored logs
### Prerequisites

### Performance Details

When visiting the WC status logs page and a high number of logs exist, the query to show the visible 20 logs is very slow.

Request URL: `/wp-admin/admin.php?page=wc-status&tab=logs`
Average query time: ~3s

Image

### Affected Sites

This affects sites with larger log counts (~2 million in my test site) and where DB logging is used over file logging.  See `WC_LOG_HANDLER` for setting storage type.

### Suspected causes

The `ORDER BY` is probably the thing slowing this query down the most.  We might consider removing it and relying on the inherent insertion order unless the timestamp is explicitly selected for ordering in the UI.

```sql
SELECT log_id, timestamp, level, message, source, context FROM wp_woocommerce_log ORDER BY timestamp DESC, log_id DESC LIMIT 20 OFFSET 0
```

### Other context

Related to https://github.com/woocommerce/woocommerce/issues/54377

### LINKED ISSUE #54540: [Performance]:  Total count query for status log is slow
### Prerequisites

### Performance Details

When visiting the WC status logs page and a high number of logs exist, the query to get the total rows (needed for pagination) is slow.

Request URL: `/wp-admin/admin.php?page=wc-status&tab=logs`
Average query time: ~0.15s

Image

### Affected Sites

This affects sites with larger log counts (~2 million in my test site) and where DB logging is used over file logging.  See `WC_LOG_HANDLER` for setting storage type.

### Suspected causes

Counting all rows on higher row count databases is going to come with some performance issues.  We might consider bypassing this with either a cache or grabbing the system partition information from SQL which should have an indexed total.

### Other context

Related to https://github.com/woocommerce/woocommerce/issues/54377

### LINKED ISSUE #54541: [Performance]:  Query for distinct log sources is slow
### Prerequisites

### Performance Details

When visiting the WC status logs page and a high number of logs exist, the query to get the sources for logs used in dropdowns can be slow.

Request URL: `/wp-admin/admin.php?page=wc-status&tab=logs`
Average query time: ~1.8s

Image

### Affected Sites

This affects sites with larger log counts (~2 million in my test site) and where DB logging is used over file logging.  See `WC_LOG_HANDLER` for setting storage type.

### Suspected causes

The use of `DISTINCT` and `ORDER BY` slow down this query, but `DISTINCT` is also needed in this case to get unique values.  We might consider caching these values when higher row counts exist and DB logging is enabled, since they probably won't change that often.  Cache busting might need to be implemented though since this is a tool used for debugging and up-to-date records are needed.

```sql
SELECT DISTINCT source
FROM wp_woocommerce_log
WHERE source != ''
ORDER BY source ASC
```

### Other context

Related to https://github.com/woocommerce/woocommerce/issues/54377

coreymckrill:

This adds various query optimizations and caching interventions to mitigate performance issues reported on the database log viewer screen. Using a log table with ~1M test log entries, these changes reduced the total page generation time (as measured by the Query Monitor plugin) from 0.55s to 0.14s on my local environment. That's a ~75% reduction!

Fixes #54538
Fixes #54540
Fixes #54541

1. Install and activate the [Query Monitor](https://wordpress.org/plugins/query-monitor/) plugin so you can observe queries and performance on WP Admin screens.
1. Configure WC to use database logging instead of the default filesystem logging:
    * Go to WooCommerce > Status > Logs > Settings.
    * Under Log Storage choose Database and Save changes.
2. Download and unzip [this database dump file](https://github.com/user-attachments/files/18875272/wordpress_2025-02-19.sql.zip), and then import it into your database. I use [Sequel Ace](https://sequel-ace.com/) for things like this, but you could also do it with the [wp-cli command](https://developer.wordpress.org/cli/commands/db/import/). This should add ~1M test log entries to the `wp_woocommerce_log` table.
3. Go to WooCommerce > Status > Logs. You should see a list table of log entries, and it should tell you how many total entries there are. Look at the Query Monitor output in the admin bar at the top of the screen and note the first number, which is the total page generation time.
4. One of the changes in this PR is to the `ORDER BY` clause of the query that gets the log entries for the list table. It should now default to ordering by the `log_id`, which is the primary key for the log table. Click on the "Timestamp" column header of the log table so that the logs are ordered by timestamp instead. Query Monitor should show that this is significantly slower. Click the "Logs" tab link to return the list table to the default sorting state.
5. Add this snippet to your test site. It will add a new log entry every time you refresh the screen while looking at the site. (Don't forget to remove this snippet when you're done testing!)
    ```php
	add_action(
		'init',
		function () {
			wc_get_logger()->debug( 'test', array( 'source' => 'asdf1' ) );
		}
	);
    ```
6. While on the log viewer screen, refresh a couple of times. Note that the total number of log entries _doesn't_ change because that value gets cached for 10 minutes.
7. Rather than waiting 10 minutes for the cache to expire, you can run `wp transient delete --all`. Then refresh the screen. The total number should now update, and Query Monitor may tell you that the page generation time is a bit slower, because it had to run the extra query to refresh the total count number.
8. Update the snippet added earlier so that it has a different, unique `source` value, then refresh the log viewer screen again. Query Monitor should show a slower time again because the sources query had to run to refresh the cached value.
9. Check out the trunk branch and compare the numbers from Query Monitor to those that you saw with this PR branch.

### DIFF
--- a/plugins/woocommerce/changelog/fix-54538-slow-log-queries
+++ b/plugins/woocommerce/changelog/fix-54538-slow-log-queries
@@ -1,0 +1,4 @@
+Significance: minor
+Type: update
+
+Improve the performance of the log viewer when using the database storage option and a high number of log entries are present
--- a/plugins/woocommerce/includes/admin/class-wc-admin-log-table-list.php
+++ b/plugins/woocommerce/includes/admin/class-wc-admin-log-table-list.php
@@ -25,6 +25,13 @@ class WC_Admin_Log_Table_List extends WP_List_Table {
 	public const PER_PAGE_USER_OPTION_KEY = 'woocommerce_status_log_items_per_page';
 
 	/**
+	 * The key for the option that stores the list of unique sources that exist in the log table.
+	 *
+	 * @const string
+	 */
+	public const SOURCE_CACHE_OPTION_KEY = 'woocommerce_status_log_db_sources';
+
+	/**
 	 * Initialize the log table list.
 	 */
 	public function __construct() {
@@ -269,14 +276,7 @@ class WC_Admin_Log_Table_List extends WP_List_Table {
 	 * @global wpdb $wpdb
 	 */
 	protected function source_dropdown() {
-		global $wpdb;
-
-		$sources = $wpdb->get_col(
-			"SELECT DISTINCT source
-			FROM {$wpdb->prefix}woocommerce_log
-			WHERE source != ''
-			ORDER BY source ASC"
-		);
+		$sources = $this->get_sources();
 
 		if ( ! empty( $sources ) ) {
 			$selected_source = isset( $_REQUEST['source'] ) ? $_REQUEST['source'] : '';
@@ -297,6 +297,37 @@ class WC_Admin_Log_Table_List extends WP_List_Table {
 				</select>
 			<?php
 		}
+	}
+
+	/**
+	 * Get the list of unique sources in the log table.
+	 *
+	 * The query in this method can be slow when there are a high number of log entries. The list of sources also
+	 * most likely doesn't change that often. So this indefinitely caches the list into the WP options table. The
+	 * cache will get cleared by the log handler if a new source is being added. See WC_Log_Handler_DB::handle().
+	 *
+	 * @return array
+	 */
+	protected function get_sources() {
+		global $wpdb;
+
+		$sources = get_option( self::SOURCE_CACHE_OPTION_KEY, null );
+		if ( is_array( $sources ) ) {
+			return $sources;
+		}
+
+		$sql     = "
+			SELECT DISTINCT source
+			FROM {$wpdb->prefix}woocommerce_log
+			WHERE source != ''
+			ORDER BY source ASC
+		";
+		$sources = $wpdb->get_col( $sql );
+
+		// Autoload this option so that the log handler doesn't have to run another query when checking the source list.
+		update_option( self::SOURCE_CACHE_OPTION_KEY, $sources, true );
+
+		return $sources;
 	}
 
 	/**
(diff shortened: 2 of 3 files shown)


jorgeatorres (commented):
Hey @coreymckrill!

Love the enhancements here 💯. Code looks good to me (so no feedback there) but a couple of observations:

- Do you think it's worth clearing the count cache when entries are deleted? It's a bit weird that when actively deleting entries you'll see incorrect numbers. This also affects pagination (as you start getting "empty" pages at the end).
- When the cache is in effect, would it make sense to indicate that the counts are approximate somehow? Maybe rounding and using `~` or something like that? Not saying that's the best choice, but in my mind a number with lots of digits seems to indicate accuracy, which is not really the case here (even if it's just for 10 mins).

What are your thoughts?

coreymckrill:
@jorgeatorres 

> Do you think it's worth clearing the count cache when entries are deleted?

Yeah, that seems reasonable. I just added it in c04397e8ec. Note that the cache only invalidates when _deleting_ log entries though. I didn't want to add it when adding entries because that would mean two database writes for every added log entry.

> When the cache is in effect, would it make sense to indicate that the counts are approximate somehow?

I looked into this, and unfortunately it's not simple because of some limitations in `WP_List_Table` (the parent class of the log list table class). It looks like I'd have to override the `pagination` function entirely in order to be able to change the output of the total item count and its label, and given how much is going on in [that function](https://github.com/WordPress/WordPress/blob/365645c608586b6e0871187d12eafeeb076c633d/wp-admin/includes/class-wp-list-table.php#L1016-L1176), I think it'd be best to avoid that.

jorgeatorres (approved):
LGTM. Thank you @coreymckrill! This is great 💯.

PR #55620, opened 2025-02-19
[Remote Logging] Improve error filtering and enable non-critical log events

chihsuan:

This PR makes several improvements to WooCommerce's remote logging functionality:

1. Removes the restriction that only logs critical events, allowing logs of all severity levels to be sent to the remote logging server. This enables us to log useful info like:

```php
wc_get_logger()->info(
  'Test info log', 
  array( 'remote-logging' => true ) 
);
```

2. Improves third-party error detection logic to reduce the number of irrelevant errors logged. Currently, there are many irrelevant errors logged due to the lack of backtrace, for example: `Allowed memory size of 268435456 bytes exhausted`, which are not useful for debugging WC issues.

3. Updates the beta tester's test event to use 'info' level instead of 'critical' for testing purposes

**Test Log PHP Event**

1. Create a fresh site
2. Install and activate this [woocommerce-beta-tester.zip](https://github.com/user-attachments/files/18861006/woocommerce-beta-tester.zip).
3. Go to `Tools -> WCA Test Helper -> Remote Logging`
4. Enable remote logging if it's not already enabled
5. Click on `Log PHP Event` button
6. See PCYsg-11XN-p2 for instructions on viewing the event logs
7. Confirm that event is logged

**Test Third-Party Error Detection**

1. Go to `Tools -> WCA Test Helper -> Remote Logging`
2. Enable remote logging if it's not already enabled
3. Modify the theme's `functions.php` to add the following code:

TT4: `wp-content/themes/twentytwentyfour/functions.php`

```php
function simulate_memory_exhaustion() {
	$array = array();
	while ( true ) {
		$array[] = str_repeat( 'a', 1000000 );
	}
}
simulate_memory_exhaustion();
```

4. Go to any wc admin page on the site
5. See fatal error occurred
6. Filter logs by host name to confirm that the error is not logged.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/src/Internal/Logging/RemoteLogger.php
+++ b/plugins/woocommerce/src/Internal/Logging/RemoteLogger.php
@@ -183,7 +183,7 @@ class RemoteLogger extends \WC_Log_Handler {
 		}
 
 		// Record fatal error stats.
-		if ( WC_Log_Levels::get_level_severity( $level ) === WC_Log_Levels::get_level_severity( WC_Log_Levels::CRITICAL ) ) {
+		if ( WC_Log_Levels::get_level_severity( $level ) >= WC_Log_Levels::get_level_severity( WC_Log_Levels::CRITICAL ) ) {
 			try {
 				$mc_stats = wc_get_container()->get( McStats::class );
 				$mc_stats->add( 'error', 'critical-errors' );
--- a/plugins/woocommerce/src/Internal/Logging/RemoteLogger.php
+++ b/plugins/woocommerce/src/Internal/Logging/RemoteLogger.php
@@ -323,16 +323,17 @@ class RemoteLogger extends \WC_Log_Handler {
 			return false;
 		}
 
+		$wc_plugin_dir = StringUtil::normalize_local_path_slashes( WC_ABSPATH );
+
+		// Check if the error message contains the WooCommerce plugin directory.
+		if ( str_contains( $message, $wc_plugin_dir ) ) {
+			return false;
+		}
+
 		// Without a backtrace, it's impossible to ascertain if the error is third-party. To avoid logging numerous irrelevant errors, we'll consider it a third-party error and ignore it.
 		if ( isset( $context['backtrace'] ) && is_array( $context['backtrace'] ) ) {
-			$wc_plugin_dir   = StringUtil::normalize_local_path_slashes( WC_ABSPATH );
 			$wp_includes_dir = StringUtil::normalize_local_path_slashes( ABSPATH . WPINC );
 			$wp_admin_dir    = StringUtil::normalize_local_path_slashes( ABSPATH . 'wp-admin' );
-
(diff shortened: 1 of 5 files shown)


moon0326 (approved):
Nice! LGTM and tested well 👍 🚀

plugins/woocommerce/src/Internal/Logging/RemoteLogger.php
```diff
@@ -178,22 +178,19 @@ protected function should_handle( $level, $message, $context ) {
...
 
-		try {
-			// Record fatal error stats.
-			$mc_stats = wc_get_container()->get( McStats::class );
-			$mc_stats->add( 'error', 'critical-errors' );
-			$mc_stats->do_server_side_stats();
-		} catch ( \Throwable $e ) {
-			error_log( 'Warning: Failed to record fatal error stats: ' . $e->getMessage() ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
+		// Record fatal error stats.
+		if ( WC_Log_Levels::get_level_severity( $level ) === WC_Log_Levels::get_level_severity( WC_Log_Levels::CRITICAL ) ) {
```
ilyasfoo: Should this be critical and above? [Severity list](https://woocommerce.github.io/code-reference/files/woocommerce-includes-class-wc-log-levels.html#source-view.44)

ilyasfoo (approved):
Tests well, but I have a few comments. Pre-approving

> Currently, there are many irrelevant errors logged due to the lack of backtrace, for example: Allowed memory size of 268435456 bytes exhausted, which are not useful for debugging WC issues.

While in most cases it's true, specifically for memory related errors, some are caused by code-related issues such as unterminated loops, faulty recursive calls, and inefficient large data handling. IMO if there's a good way to tell the difference, it might still be worth collecting exhausted memory error.

plugins/woocommerce/src/Internal/Logging/RemoteLogger.php
chihsuan: Good catch!  Fixed in 7bcf7e79ae13e44e06c8bb8ea7601ad6d4370e92.

chihsuan:
> While in most cases it's true, specifically for memory related errors, some are caused by code-related issues such as unterminated loops, faulty recursive calls, and inefficient large data handling. IMO if there's a good way to tell the difference, it might still be worth collecting exhausted memory error.

Good point! @ilyasfoo I moved the error message check outside of backtrace check in ac309265123b86da9275fb0c188b405c808438b6.  The error message includes the path where the error occurs so we won't miss if the error is directly related to WC. If the error message doesn't include WC, I think we can safely ignore it since we're not able to tell if it's a WC issue or not.

PR #55623, opened 2025-02-19
Add a COGS value column to the product meta lookup table

Konamiman:

This PR is a follow-up to https://github.com/woocommerce/woocommerce/pull/54399. It adds a tool to the WooCommerce admin tools page that allows to create a `cogs_total_value` column in the `wp_wc_product_meta_lookup` table. The column is initially populated from existing cost values when it's created, and then it gets updated every time that a product is saved.

image

image

The column is nullable and it follows the same semantics of the `_cogs_total_value` product meta item. This means that zero cost values will be stored as `null` in the column, except for product variations, where `null` and zero have actually different meanings.

If the column exists then the "Cost" column in the products list in admin will be sortable:

image

When the Cost of Goods Sold feature is disabled the tool to create or remove the database column is "asymmetric": if the column doesn't exist then the tool to create it won't appear in the tools page, but if the column exists then the tool that allows to remove it from the table will appear.

**Important!** Note that once you run a tool from the WooCommerce tools page the page reloads with `&action=<tool name>` added to the URL. To avoid confussion you should remove that part from the URL before reloading the page.

1. Enable the Cost of Goods Sold feature and add costs for a few products, including variable products and variations; for the later, have variations with an explicit non-zero cost, with an explicit zero cost, and with no cost value. You can use the testing instructions of https://github.com/woocommerce/woocommerce/pull/54399 as guidance.
2. Go to the tools page (WooCommerce - Status - Tools) and run the "Create COGS columns in the product meta lookup table" tool.
3. Verify that the products meta lookup table has the new column and that it has the proper value for regular products (simple or variable) and for variations. For regular products the column should hold `null` when the cost is zero; for variations, an explicit cost of zero should be stored as such, and a not defined cost should be stored as `null`. Here's how you can check these values from the command line:

```
wp db query "select cogs_total_value from wp_wc_product_meta_lookup where product_id=<product or variation id>"
```

4. Modify the cost of a couple of products and verify that the value in the meta table has been updated (using the query from the previous step). Again, try with regular products and variations; and try with undefined, zero and non-zero values.
5. Go to the products list in admin and verify that the "Cost" column is now sortable and it works as expected. Remember that for now, for variable products what this column shows is simply the default cost as defined in the "General" tab.
6. Go back to the tools page and run the "Remove COGS columns from the product meta lookup table" tool. Verify that the `cogs_total_value` column has disappeared, you can do so by running the following and verifying that the column isn't part of the result:

```
wp db query "select * from wp_wc_product_meta_lookup limit 1"
```

6. Verify that the "Cost" column in the products list still shows correct values, but it isn't sortable anymore.
7. Reload the tools page an run the tool to create the column again.
8. Disable the Cost of Goods Sold feature from the WooCommerce - Settings - Advanced - Features page.
9. Reload the tools page, verify that the tool to remove the column is there, and run it.
10. Reload the tools page and verify that no tools related to the Cost of Goods Sold feature are appearing now.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/pr-55623
+++ b/plugins/woocommerce/changelog/pr-55623
@@ -1,0 +1,4 @@
+Significance: minor
+Type: add
+
+Add a COGS value column to the product meta lookup table
--- a/plugins/woocommerce/includes/admin/list-tables/class-wc-admin-list-table-products.php
+++ b/plugins/woocommerce/includes/admin/list-tables/class-wc-admin-list-table-products.php
@@ -34,6 +34,20 @@ class WC_Admin_List_Table_Products extends WC_Admin_List_Table {
 	protected $list_table_type = 'product';
 
 	/**
+	 * Caches the value of the "COGS is enabled" flag.
+	 *
+	 * @var bool
+	 */
+	private bool $cogs_is_enabled;
+
+	/**
+	 * Flag indicating if the COGS value column in the product meta lookup table can be used.
+	 *
+	 * @var bool
+	 */
+	private bool $use_cogs_lookup_column;
+
+	/**
 	 * Constructor.
 	 */
 	public function __construct() {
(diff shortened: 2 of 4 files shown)


albarin (approved):
Nice job!
Everything tests great, nothing much to add! ✅
Very good testing instructions!

PR #55633, opened 2025-02-19
[dev] Cleanup Blocks' Jest tests setup

WunderBart:
## What?

Clean up the existing Blocks' Jest tests setup before implementing the headless editor for the integration tests (see #55651).

## Suggested Changes

Organize Jest's scripts, config, and mocks. 💅 

## Testing Instructions

All tests should pass.


plugins/woocommerce-blocks/tests/js/config/testing-library.js
```diff
@@ -1,12 +1,14 @@
 /**
```
louwie17, line 1: Any specific reason for moving this to a `JS` file instead of keeping it as a `TS` file?

louwie17 (commented):
This mostly looks good from my perspective, just curious about one change of moving a TS file to JS?

plugins/woocommerce-blocks/tests/js/config/testing-library.js
WunderBart: This is only because we don't seem to have the TS configured well for the tests context. That file was the only *.ts one in the tests folder, and for it to not throw errors it had to be explicitly listed in the `tsconfig` [here](https://github.com/woocommerce/woocommerce/pull/55633/files#diff-df2e1022c74b5764079d86c44052fd5d89779e57ebf407fa2c67730eea52ba33L12) (renamed to testing-library.js in this PR). Let's make sure the TS config is properly set up for the entire folder first. I'll create an issue for it. Does that make sense?

louwie17: Gotcha, yeah that makes sense! I only wanted to clarify, as usually we promote converting things to TS versus using JS.
But I am fine with this in that case.

louwie17 (approved):
LGTM 🚀  Thanks for the clarification, this looks good 👍

PR #55642, opened 2025-02-19
Small edit to the checkout code snippet

alexflorisca:
Small change to an example code snippet. Changed it to verify that the field contains only letters. If we were to validate if the field was required, we would do it slightly differently, We would simply add the `required` property and set it to `true` when registering the field.


samueljseay (approved):
Nice! good spotting to improve this.

alexflorisca:
Thanks Sam. I've just update the docs manifest. @shaybanshee feel free to merge this if you're with the changes!

shaybanshee:
Hey @alexflorisca - can you update the manifest here? The merge is currently blocked - thank you!

shaybanshee (approved):

PR #55651, opened 2025-02-19
[dev] Introduce Integration Test Editor in Blocks' JS tests

WunderBart:
## What?

Supersedes https://github.com/woocommerce/woocommerce/pull/53255

This PR introduces the Integration Test Editor to the existing Blocks' Jest environment.

## Overview:
- **Follows the Gutenberg Approach:** Integration tests are placed within their respective parent block folders and run as part of the unit tests suite, similar to [how it's organized in Gutenberg](https://github.com/WordPress/gutenberg/blob/41deebd821899010453a60cab6594e0791809fdf/docs/contributors/code/testing-overview.md#integration-testing-for-block-ui). This avoids the need for a separate integration test suite, leveraging shared configurations for simplicity and consistency.
- **Example Test Included:** An [example integration test](https://github.com/woocommerce/woocommerce/blob/c5f5a4c7199947058ca58e423c442fcfe07927b3/plugins/woocommerce-blocks/assets/js/blocks/active-filters/test/block.ts) is provided to demonstrate how tests can now be written using this setup. This serves as a guideline for writing future integration tests for other blocks.
- **Flexible Structure for Wider Integration Tests:** While block-specific tests are kept within their respective block folders, the structure allows for wider integration tests to be added in the [tests/integration](https://github.com/woocommerce/woocommerce/tree/3809a3d8dc874ee2f5c8802d915166f26cd4f3d1/plugins/woocommerce-blocks/tests/integration) folder. This provides the flexibility to write more comprehensive tests without being restricted to block-specific contexts.
- **Optimized for CI:** Using a single suite for both unit and integration tests optimizes CI performance by reducing overhead and complexity, resulting in faster and more efficient test runs.

This implementation enhances test coverage while maintaining consistency with Gutenberg's approach, leading to a streamlined and efficient testing workflow.

## Testing Instructions

- All tests should pass.
- Make sure the new Active Filters block (integration) test passes.

---

cc: @glendaviesnz


plugins/woocommerce-blocks/assets/js/blocks/active-filters/test/block.ts
```diff
@@ -0,0 +1,53 @@
...
+import '@testing-library/jest-dom';
+import { fireEvent, screen, within } from '@testing-library/react';
+
+/**
+ * Internal dependencies
+ */
+import {
+	initializeEditor,
+	selectBlock,
+} from '../../../../../tests/integration/helpers/integration-test-editor';
```
gigitux, line 14: NIT: Should we create an import alias?

plugins/woocommerce-blocks/assets/js/blocks/active-filters/test/block.ts
```diff
@@ -0,0 +1,53 @@
...
+import '../';
```
gigitux, line 15: Is it necessary?

plugins/woocommerce-blocks/tests/integration/helpers/lock-unlock.js
gigitux, line 1: Should we add a comment that explains why this file is necessary?

gigitux (approved):
Awesome work! ❤️

I left few comments, but they are not blockers!

plugins/woocommerce-blocks/assets/js/blocks/active-filters/test/block.ts
WunderBart: I didn't create an alias as there seems to be a wider issue with path aliasing in Blocks. I see multiple levels that aren't syncing well: 
- [Jest](https://github.com/woocommerce/woocommerce/blob/306d3f48a239bb4faa8c555b38b0eb057624d842/plugins/woocommerce-blocks/tests/js/jest.config.json#L10-L38)
- [TS](https://github.com/woocommerce/woocommerce/blob/f617c09681748acf385decd8ed80ee6c3ecc6095/plugins/woocommerce-blocks/tsconfig.base.json#L26-L74)
- [ESLint](https://github.com/woocommerce/woocommerce/blob/88fa6f50bbdce4bacc62e79243ec61dc57c7d940/plugins/woocommerce-blocks/.eslintrc.js#L124-L157)
- [Webpack](https://github.com/woocommerce/woocommerce/blob/88fa6f50bbdce4bacc62e79243ec61dc57c7d940/plugins/woocommerce-blocks/bin/webpack-helpers.js#L41-L107)

An example sync issue can be where TS is satisfied, but ESLint or Jest still throw an error because they cannot resolve. 

I've also found some unused aliases, like `@woocommerce/blocks-test-utils` and more files that could use aliases, e.g. [here](https://github.com/woocommerce/woocommerce/blob/86112436ebafa32ea26048ac3fea15a0a4bcf04c/plugins/woocommerce-blocks/assets/js/blocks/order-confirmation/create-account/test/frontend-block.test.tsx#L12).

I think it would be great to create a separate issue to perform an audit on those configs and try coming up with more sustainable solution that would work across the aforementioned levels. I think we could get some inspiration from Gutenberg, where they have [a dedicated resolver](https://github.com/WordPress/gutenberg/blob/6f913027894852bd4dc45750b7b6a92bc7242e70/tools/eslint/import-resolver.js) that, e.g., resolves to the given root by default (`woocommerce-blocks/` in our case). Does that make sense?

cc: @ObliviousHarmony

plugins/woocommerce-blocks/assets/js/blocks/active-filters/test/block.ts
WunderBart: It imports and registers the block we're testing, so it's necessary to have it insertable within the headless editor.

plugins/woocommerce-blocks/tests/integration/helpers/lock-unlock.js
WunderBart: Yeah, good idea!

WunderBart: I was actually able to remove that file and restore the original import from the `@wordpress/block-library`. 🙌

plugins/woocommerce-blocks/assets/js/blocks/active-filters/test/block.ts
gigitux: Sounds great! @WunderBart, can you take care to open the issue? 🙇

gigitux:
Thanks for updating the PR! Let's :shipit: it!

plugins/woocommerce-blocks/assets/js/blocks/active-filters/test/block.ts
WunderBart: https://github.com/woocommerce/woocommerce/issues/55929

PR #55673, opened 2025-02-20
Fix apostrophes not correctly rendered in Checkout block saved addresses

### LINKED ISSUE #53622: Checkout block: apostrophes are not escaped correctly in saved addresses
> [!CAUTION]
> This issue is currently blocked, see https://github.com/woocommerce/woocommerce/issues/53622#issuecomment-2594599665

### Prerequisites

### Describe the bug

Apostrophe characters are not escaped correctly when printing the saved addresses:

### Expected behavior

In the screenshot above, the street should be `Carrer de l'Hospital` instead of `Carrer de l&#8217;Hospital`.

### Actual behavior

Apostrophes are printed as `&#8217;`.

### Steps to reproduce

1. Make a purchase, adding a street name and city name containing an apostrophe, ie: `Carrer de l'Hospital` for the street and `L'Hospitalet de Llobregat` for the city name. Make sure to use the `'` character, instead of `’`.
2. Proceed with the checkout and go back to the Shop page, add another product to cart and go to the Checkout page.
3. Notice the saved address displays the address with `&#8217;` instead of `'`.

### WordPress Environment

WP 6.7 and WooCommerce 9.5 and `trunk`.

### Isolating the problem

Aljullu:

Closes https://github.com/woocommerce/woocommerce/issues/53622.

This PR makes it so we decode the address fields before printing them to the screen.

The contents returned by [`getCustomerData()`](https://github.com/woocommerce/woocommerce/blob/4e425ab9145eb8a01456aea9d401307114d17a45/plugins/woocommerce-blocks/assets/js/data/cart/selectors.ts#L29) might or might not be encoded. Currently, when they are first hydrated from the backend values printed on the markup, they are encoded, but they are decoded once the user interacts with the form. I think in the long term it would be good to unify that and make sure the store always contains data either encoded or decoded. This PR doesn't fix that and simply decodes it just in case.

1. Make a purchase, adding apostrophes to as many fields as you can. Make sure to use the `'` character, instead of `’`.

2. Proceed with the checkout and go back to the Shop page, add another product to cart and go to the Checkout page again.
3. Verify the saved address displays the address with `'` instead of `&#8217;`.

Before | After
--- | ---
Captura de pantalla de 2025-02-20 10-23-57 | Captura de pantalla de 2025-02-20 10-13-42

### DIFF
--- a/plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/form.tsx
+++ b/plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/form.tsx
@@ -24,6 +24,7 @@ import { AddressFormValues, ContactFormValues } from '@woocommerce/settings';
 import { objectHasProp } from '@woocommerce/types';
 import { useCheckoutAddress } from '@woocommerce/base-context';
 import fastDeepEqual from 'fast-deep-equal/es6';
+import { decodeEntities } from '@wordpress/html-entities';
 
 /**
  * Internal dependencies
@@ -356,7 +357,9 @@ const Form = < T extends AddressFormValues | ContactFormValues >( {
 						type={ field.type }
 						ariaDescribedBy={ ariaDescribedBy }
 						value={
-							( values[ field.key as keyof T ] as string ) || ''
+							decodeEntities(
+								values[ field.key as keyof T ] as string
+							) || ''
 						}
 						onChange={ ( newValue: string ) =>
 							onChange( {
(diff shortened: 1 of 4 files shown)


tjcafferkey (approved):
This works as expected and makes sense @Aljullu so I am approving from my side. I do agree with your more long-term solution of standardizing whether the content is stored decoded/encoded to make it more consistent.

Even though I believe this PR is relatively low risk I'd wait for either @wavvves or @nielslange to review also given they work more closely in this area.

Thanks!

nielslange (approved):
Thanks for addressing this issue, @Aljullu. I've successfully tested your fix. 🙌

PR #55675, opened 2025-02-20
Add WC_VERSION input to the on-demand job

### LINKED ISSUE #55694: [e2e] Specify version of WC to use while running on-demand jobs
After merging this PR https://github.com/woocommerce/woocommerce/pull/55063, we have the latest WC version installed during test execution on the Pressable website. It would be helpful to have an option to specify which version of WC we want to run tests against when we're running [on-demand](https://github.com/woocommerce/woocommerce/actions/workflows/tests-on-demand.yml) jobs against external envs. The abovementioned PR introduced `WC_VERSION` env var and we can use it for `on-demand` jobs.

Stojdza:

Add an option to [`on-demand` job](https://github.com/woocommerce/woocommerce/actions/workflows/tests-on-demand.yml) to accept `WC_VERSION` as a parameter. With this change, we can specify which WC version will be installed on the external environments while running tests. 

Closes #55694 

ℹ️ **NOTE:** I'll revert this commit [7aaa20f](https://github.com/woocommerce/woocommerce/pull/55675/commits/7aaa20fc8334b1e3cc11ea718c73bdabd3489a87) as soon as we finish testing. I left it for testing, so that we can run only `basic.spec.js` and not taking too much of CI time.  

1. Go to https://github.com/woocommerce/woocommerce/actions/workflows/tests-on-demand.yml
2. Click on Run workflow button
3. Choose this PR's branch - `e2e/on-demand-add-wc-version`
4. One new input field should show in the dropdown -> "If you want to test against a specific WC version, provide it here. Available versions: https://github.com/woocommerce/woocommerce/releases"
5. In the input field enter one of the [previous releases](https://github.com/woocommerce/woocommerce/releases), eg. `9.7.0-rc.1`
6. Run the workflow

Expected results:
- Check the job for Pressable website. Specified WC version should be installed ([example](https://github.com/woocommerce/woocommerce/actions/runs/13437369015/job/37543034864#step:9:136));
- Go to Pressable website, verify that the specified WC version is installed and activated.

Additional case - passing WC version that doesn't exist, eg. `9.9.x.y.z`:
- The job should fail ([example](https://github.com/woocommerce/woocommerce/actions/runs/13437705199/job/37544017536#step:9:128))
- Does this failure make sense? Should we handle error in a different way? 

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Add an option to `on-demand` job to accept WC_VERSION

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ on:
         default: 'default'
         type: string
       wc-version:
-        description: 'Specific WooCommerce version to test against.'
+        description: 'Specific WooCommerce version to install by the test jobs supporting it'
         required: false
         type: string
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,10 @@ on:
         required: true
         default: 'default'
         type: string
+      wc-version:
+        description: 'Specific WooCommerce version to test against.'
+        required: false
+        type: string
 
 concurrency:
   # Cancel concurrent jobs but not for push event. For push use the run_id to have a unique group.
@@ -228,6 +232,7 @@ jobs:
           CODEVITALS_PROJECT_TOKEN: ${{ secrets.CODEVITALS_PROJECT_TOKEN }} # required by Metrics tests
           LAST_FAILED_RUN: ${{ vars.LAST_FAILED_RUN }}
           GITHUB_BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          WC_VERSION: ${{ inputs.wc-version }}
         run: |
           lastRunFile="${{ steps.download-last-run-info.outputs.download-path }}/last-run__${{ strategy.job-index }}/.last-run.json"
           lastRunFileDest="$ARTIFACTS_PATH/.last-run.json"
--- a/.github/workflows/tests-on-demand.yml
+++ b/.github/workflows/tests-on-demand.yml
@@ -21,7 +21,7 @@ on:
         required: false
       wc-version:
         type: string
-        description: 'If you want to test against a specific WC version, provide it here. Available versions: https://github.com/woocommerce/woocommerce/releases'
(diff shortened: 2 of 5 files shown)


.github/workflows/tests-on-demand.yml
```diff
@@ -19,6 +19,10 @@ on:
         type: string
         description: 'Custom event name: In case the `Event name` choice is `custom`, this field is required.'
         required: false
+      wc-version:
+        type: string
+        description: 'If you want to test against a specific WC version, provide it here. Available versions: https://github.com/woocommerce/woocommerce/releases'
```
adimoldovan: I would add the warning that only some jobs are configured to work with this and most (wp-env) will simply ignore the version.

.github/workflows/ci.yml
```diff
@@ -23,6 +23,10 @@ on:
         required: true
         default: 'default'
         type: string
+      wc-version:
+        description: 'Specific WooCommerce version to test against.'
```
adimoldovan: Not all test jobs are using this, it can be confusing

```suggestion
        description: 'Specific WooCommerce version to install by the test jobs supporting it'
```

adimoldovan (approved):
Tested with this job: https://github.com/woocommerce/woocommerce/actions/runs/13451376465/job/37586305600 that ran the `custom` event `core-e2e-pressable`. All good.

I left 2 comments on the wording. My fear is that this new parameter can be confusing because it is not used by most of the jobs.

> Additional case - passing WC version that doesn't exist, eg. 9.9.x.y.z:
> Does this failure make sense? Should we handle error in a different way?

I think the job should have failed at the install step and not proceed with the next ones. But this is not related to this PR and should be handled by another.

I'm approving. Nice work @Stojdza.

.github/workflows/tests-on-demand.yml
Stojdza: Updated description here [bc608bc](https://github.com/woocommerce/woocommerce/pull/55675/commits/bc608bcadb872c5c25274b38bfe8f4bebc86a612)

PR #55693, opened 2025-02-20
Add Filter to Override Cart Data Hydration, Allowing Cart Page Caching

jbouganim:

This PR introduces a new filter `woocommerce_blocks_cart_hydrate_api_request` to allow developers to override the default behavior of cart data hydration. This filter provides an option to disable hydration, enabling the cart page to be cached and to force the cart data to be fetched asynchronously. This can improve user experience by allowing the page to load instantly while cart data is retrieved in the background.

1. Apply the changes from this PR to your WooCommerce installation.
2. Add a filter in your theme or plugin to override the default hydration behavior:
```
   add_filter( 'woocommerce_blocks_cart_hydrate_api_request', '__return_false' );
```
3. Visit a product page and add an item to the cart.
4. Visit the cart page and observe that the page loads without pre-populated cart data. (Search the document source for product/cart data)
5. Verify that the cart data is fetched asynchronously via an API request.
6. Remove the filter and confirm that the cart data is hydrated as expected.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message
Introduced a new filter `woocommerce_blocks_cart_hydrate_api_request` to allow developers to disable cart data hydration and enable asynchronous API requests for cart data.

</details>

<details>

#### Comment 
This change provides developers with more flexibility in managing cart data loading, potentially improving performance and user experience by allowing the cart page to be cached.

</details>


opr:
Hey @jbouganim thanks for opening this PR, I just tested it out and it runs OK, I see the separate request to cart when the page has loaded.

I want to clarify, what is the use-case for this?

> This PR introduces a new filter `woocommerce_blocks_cart_hydrate_api_request` to allow developers to override the default behavior of cart data hydration. This filter provides an option to disable hydration, enabling the cart page to be cached and to force the cart data to be fetched asynchronously. This can improve user experience by allowing the page to load instantly while cart data is retrieved in the background.

I understand that this allows the cart page to be loaded without any hydrated data being present on the page, but what is the problem you're solving here? Is it that the cart/checkout pages can't be cached in their current form?

If so, then the same logic would need to be applied to the Checkout page.

opr (commented):
Hey @jbouganim thanks for opening this PR, I just tested it out and it runs OK, I see the separate request to cart when the page has loaded.

I want to clarify, what is the use-case for this?

> This PR introduces a new filter `woocommerce_blocks_cart_hydrate_api_request` to allow developers to override the default behavior of cart data hydration. This filter provides an option to disable hydration, enabling the cart page to be cached and to force the cart data to be fetched asynchronously. This can improve user experience by allowing the page to load instantly while cart data is retrieved in the background.

I understand that this allows the cart page to be loaded without any hydrated data being present on the page, but what is the problem you're solving here? Is it that the cart/checkout pages can't be cached in their current form?

If so, then the same logic would need to be applied to the Checkout page.

jbouganim:
Hi @opr 

Yes by preventing the hydration, we can have the cart page itself cached by the CDN or otherwise, which loads much faster, and having the cart data fetched asynchronously. 
Almost all of our clients prefer this method, versus the time tacked on to the initial cart page. 

It's true this can be applied to the checkout page as well, but there were some issues during testing so looking into this more when we have time.

opr:
@jbouganim thanks for clarifying. Personally, I think this idea is OK. If it allows these pages to be cached and the perceived load time is improved for visitors, then I like it.

> It's true this can be applied to the checkout page as well, but there were some issues during testing so looking into this more when we have time.

What issues did you face? I am hesitant about shipping a solution for one block and not the other.

jbouganim:
Hi @opr,

I understand that. The issue was the checkout block did not loading properly. I can look into it more when I have the bandwidth to debug why, just figured this update was a quick fix that many users would be able to appreciate in the meantime.

opr:
> just figured this update was a quick fix that many users would be able to appreciate in the meantime

Absolutely, I understand where you're coming from and do appreciate the time taken to open the PR. If you want to, you could set the PR back to draft while you work on the checkout too.

I think it's important to validate the approach works on both Cart and Checkout before we ship. Imagine if we shipped this and we find out it can't be done for checkout for one reason or another, we'd be stuck supporting two separate solutions for this.

jbouganim:
@opr 
Tested on checkout page but checkout block does not load up if it's not hydrated so there is some dependency there. Will set this to draft until we have some more time to investigate.

jbouganim:
@opr fyi... closing this, opened another PR https://github.com/woocommerce/woocommerce/pull/64289 which covers the cart/checkout and edge cache support for IAPI components like mini-cart/product buttons

PR #55695, opened 2025-02-20
[e2e tests] Refactor checkout specs

adimoldovan:

Part of pdV5qK-Ie-p2
Address follow-up todo comments - refactored checkout related specs. Simplified and consolidated tests. Ensured unique test data and proper cleanup.
Removed many tests that where checking stuff already covered in blocks e2e tests or unit tests.
Moved the coupons related specs in the coupons folder.
Kept a single `checkout.spec.js`, testing main checkout flows for both classic and blocks checkout. 

The new checkout.spec.js:
- uses parametrized tests, to run the same flows for both classic and blocks checkout pages. The blocks checkout page is only created if it doesn't already exist, making the test faster

Removed:
- cart-checkout-block-calculate-tax.spec.js - covered in blocks e2e tests, plus a simple coverage is also included in the new checkout spec.
- cart-checkout-calculate-tax.spec.js - covered in php unit tests, plus a simple coverage is also included in the new checkout spec.
- checkout-block.spec.js - merged with the new checkout.spec.js
- checkout-create-account.spec.js - merged with the new checkout.spec.js
- checkout-login.spec.js - merged with the new checkout.spec.js
- create-checkout-block.spec.js - covered in blocks e2e tests

Created a new spec `order-grace-period.spec.js` to cover the grace period feature instead of having this in the checkout flows.

CI green.

To run the affected tests locally:
```
pnpm test:e2e checkout/
```


Stojdza (approved):
Ran locally several times, works as expected 👍

plugins/woocommerce/tests/e2e-pw/tests/checkout/checkout.spec.js
```diff
@@ -1,672 +1,508 @@
...
-			{ tag: [ tags.COULD_BE_LOWER_LEVEL_TEST ] },
-			async ( { page } ) => {
-				await addAProductToCart( page, productId );
-
-				await page.goto( 'checkout/' );
+		// Check id COD payment is enabled and enable it if it is not
+		const codResponse = await api.get( 'payment_gateways/cod' );
+		const codEnabled = codResponse.enabled;
 
-				// first try submitting the form with no fields complete
```
ralucaStan, line 277: Will these scenarios be added back? Curious why this was considered as COULD_BE_LOWER_LEVEL_TEST.

PR #55707, opened 2025-02-20
Support wcSettings deprecation

moon0326:

Related comment: https://github.com/woocommerce/woocommerce/issues/55046#issuecomment-2641741369

This PR adds a feature that allows a deprecation warning to be added to a specific property in wcSettings.

Note: Currently, it's not possible to test this code on a JN site as all JN sites are built as a production ready build. You can either test it locally, or ask @moon0326 for a pre-built zip.

1. Checkout this branch locally and build `pnpm --filter='@woocommerce/plugin-woocommerce' watch:build`
2. Complete the core profiler and navigate to WooCommerce -> Home
3. Open your console.
4. Try accessing `window.wcSettings` and confirm you can still retrieve values.
5. Try accessing `window.wcSettings.admin.onboarding` and confirm you get a warning message. 

Extra: Add a few different properties in https://github.com/woocommerce/woocommerce/compare/update/add-deprecation-warning-for-wcsettings-admin?expand=1#diff-4f0f4ed32f4dbc5f71c04c0602c3e1625371339e2c58eea41b08400ab880bb6fR88 and test.

**Testing getAdminSetting**
1. Checkout this branch locally.
2. Find a client code where `getAdminSetting` is used.
3. Add the property used in `getAdminSetting` in https://github.com/woocommerce/woocommerce/pull/55707/files#diff-4f0f4ed32f4dbc5f71c04c0602c3e1625371339e2c58eea41b08400ab880bb6fR87
4. Navigate to the page where the code is used.
5. Confirm the warning message is logged in the console.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Added a feature that allows a deprecation warning to be added to a specific property in wcSettings.

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Internal/Admin/WCAdminSharedSettings.php
```diff
@@ -63,9 +63,42 @@ public function on_woocommerce_blocks_loaded() {
...
+	 *
+	 * Nested properties can be defined using dot notation.
+	 *
+	 * @return array
+	 */
+	protected function get_deprecated_properties() {
+		return array(
+			'admin.onboarding' => null,
+		);
+	}
```
chihsuan: Given that these deprecations are primarily for frontend JavaScript code, I think we could define these directly in JavaScript so we don't need PHP-JS data transfer, which reduces server response size (though minimal) but also simplifies the logic.

plugins/woocommerce/client/admin/client/utils/admin-settings.js
```diff
@@ -38,6 +38,17 @@ export function getAdminSetting(
 	fallback = false,
 	filter = ( val ) => val
 ) {
+	if ( window.deprecatedWcSettings ) {
+		const key = `admin.${ name }`;
+		if ( key in window.deprecatedWcSettings ) {
+			const warningMessage = window.deprecatedWcSettings[ key ]
+				? window.deprecatedWcSettings[ key ]
+				: `Deprecated: ${ name } has been deprecated. This value will be removed soon.`;
```
chihsuan: I think it would be great if we could mention the alternative in the deprecated message and maybe be more specific about which version it will be removed. What do you think?

plugins/woocommerce/src/Internal/Admin/WCAdminSharedSettings.php
moon0326: @chihsuan Good call! I'll make the change.

plugins/woocommerce/client/admin/client/utils/admin-settings.js
moon0326: @chihsuan So this is the default message. We can work on a customized message with an alternative when we work on a PR to add more properties other than admin.onboarding

> maybe be more specific about which version it will be removed

Hey @adrianduffell ✋ How long should we keep the deprecation messages? Do you think two release cycles would be long enough?

adrianduffell: @moon0326 , yeah let's do the removal in the 10.0 release.

moon0326: @chihsuan  @adrianduffell Sounds good! Updated messages with alternative and removal version in https://github.com/woocommerce/woocommerce/pull/55707/files#diff-b5ccd801e97e593229d243c9c5721c56d282ac1e2867df738e62d4ea105bfcbfR25

plugins/woocommerce/src/Internal/Admin/WCAdminSharedSettings.php
moon0326: Hey @chihsuan ✋ 

I've updated the code to define the property list in JS.

Could you review the code one more time please 🙏 ? I've moved things around for reusuability.

1. Added a generic proxy function in https://github.com/woocommerce/woocommerce/pull/55707/files#diff-916fa93bf98ff7dcb904fbda23451f2ab92b723a8e412beff66889abdd09fc99
2. I'm now using it in both [admin-settings.js](https://github.com/woocommerce/woocommerce/pull/55707/files#diff-b5ccd801e97e593229d243c9c5721c56d282ac1e2867df738e62d4ea105bfcbfR75) and [wcsetting-deprecation](https://github.com/woocommerce/woocommerce/pull/55707/files#diff-1e593140447e9917b857820396a32766b99f8286798ed6c0067f3e7ecc161997)

Updated tests accordingly as well.

Thanks!

rjchow (approved):
Tests well, only one suggestion would be to consider turning off the deprecation notice in prod build

Looks like some jest mocking of getAdminSettings is also causing JS test failure

moon0326:
> Tests well, only one suggestion would be to consider turning off the deprecation notice in prod build
> 
> Looks like some jest mocking of getAdminSettings is also causing JS test failure

Good call! Updated it in [Only warn in development](https://github.com/woocommerce/woocommerce/pull/55707/commits/8d81af2f4c54f4a7d76422080ea6b0719a7669f9)

chihsuan (approved):
Thanks, tested well! @moon0326 

As the deprecation message is only displayed in the development environment, it would be helpful to update the test instructions to guide QA on how to test this.

moon0326:
> As the deprecation message is only displayed in the development environment, it would be helpful to update the test instructions to guide QA on how to test this.

Thank you!

> As the deprecation message is only displayed in the development environment, it would be helpful to update the test instructions to guide QA on how to test this.

Good catch. Since JS files are generated during the build process, I don't think it's possible to test them on a JN site at the moment. I've left a note for QA.

PR #55717, opened 2025-02-21
Move iapi powered notices into product collection

samueljseay:

For the product button block, pre interactivity API, it would manually inject a React root/component to the page to show a store notice if there was an error with add to cart.

We attempted a block hooks based approach to ensuring the store notices block is present instead and also adding ability to add notices client-side, but block hooks has limitations that will not work for us in this case.

Instead of trying to solve this in the store notices block since this is just a workaround for a small problem, we have opted to have product button add notices to a notices container rendered at the top of product collection instead. At some point we need to have a broader discussion about reworking/redesigning/rearchitecting the concept of store notices that could be triggered from any block in the block tree, but since now is not that time, we believe this workaround will suffice as part of the goal of migrating to using core `@wordpress/interactivity` instead of the version that was incubated within WooCommerce. 

To ensure that the error notice still shows you can follow the steps from the original PR: https://github.com/woocommerce/woocommerce/pull/54552

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Display client-side error notices generated from product button block within product collection block

</details>

<details>

#### Comment 

</details>


plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/button/frontend.tsx
```diff
@@ -111,11 +112,43 @@ const { state } = store< Store >(
 					'../../../../base/stores/cart-items'
 				) ) as WooStore;
 
-				// Question: should this action throw so we can capture the error here?
```
samueljseay, line 121: Answer: yes I think so. Catching here to show error in the UI is a responsibility of the frontend so I moved this out of the store.

luisherranz (approved):
Okay, this looks good to me.

Let's keep an eye on who should trigger the store notice call when we keep iterating on the design of the `woocommerce` store!

Issue #53564, opened 2024-12-09
[Performance]: switch to as_has_scheduled_action() instead of as_next_scheduled_action() wherever possible


prettyboymp:
### Describe the solution you'd like

Most checks for existing scheduled action use `as_next_scheduled_action()` instead of `as_has_scheduled_action()` resulting in a much slower query than needed to verify if an action is already scheduled.

`as_has_scheduled_action()` is just like `as_next_scheduled_action()` except that it does not apply an order by clause to specifically fetch the next action.  The addition of the order by clause often requires a filesort during query execution for sites with a large number of scheduled actions, e.g. those running WooCommerce Subscriptions with a large number of active subscriptions.  The end result is that the check for an existing action takes seconds rather than milliseconds. 

We should go through all places that currently use `as_next_scheduled_action()` and determine if they actually need the next action or are just checking whether an action is scheduled and update accordingly.

### Additional context

`as_has_scheduled_action()` was added in Action-Scheduler 3.3.0.  We will likely either need a check that the function exists or a wrapper for it.  `WC_Action_Queue` should serve as a wrapper, however, adding we'll need to need a way to polyfill custom implementations of `WC_Queue_Interface` if we're adding a new method.

prettyboymp:
Some notes related to this task:

* There is no guarantee of what version of action scheduler is being loaded because older extensions may load action scheduler.  
* `as_has_scheduled_action()` may not be available in the loaded version.
* `as_has_scheduled_action()` as a check to create unique actions has a race condition concern, see https://github.com/woocommerce/action-scheduler/issues/828
* https://github.com/woocommerce/action-scheduler/pull/831 works through handling the scheduling of unique actions, but that too may not be available in the loaded version.  So we need to properly adapt in core for our usage.
 * The fixes in 831 above still need tweaks because some of the fallback checks still use an orderby clause.
 * The core handling selects from the DUAL table which isn't always available.  That needs a work around as well.

prettyboymp:
AutomateWoo also adds a large number of queries attempting to ensure that actions are scheduled.  Currently seeing between 10 - 20 queries per admin page load through ActionScheduler just from Automatwoo.

prettyboymp:
Currently blocked until we decide what path to take with improving the logic around scheduling of recurring actions in #55746 and whether we choose to migrate to a single hook that should be used to make sure recurring actions are scheduled.

prettyboymp:
[...]

PR #55746, opened 2025-02-21
Add background scheduler to refresh expensive queries

### LINKED ISSUE #54819: [Performance]: Create reusable interface for background indexing
### Prerequisites

### Performance Details

A number of plugins do make use of caches and transients to store expensive queries, however, those caches inevitably expire and often need to be re-run when admin routes are visited again, resulting in long load times as every plugin attempts to hydrate its various indexes.

A better option might be to handle this indexing in the background on non-admin requests so that these don't end up blocking page loads.  Action Scheduler may be a good candidate but more investigation is required.

Request URL: All admin routes
Average time: Varies from 0.2s ~ 8s

### Affected Sites

Most sites and plugins, especially those with higher post or order count tables.

### Suspected causes

Expensive queries are the root cause, but initial attempts have been made to optimize these queries with caching, leading to bulk indexing from multiple plugins after not visiting a site for a period of time.

### Other context

_No response_

joshuatf:

This PR adds a background scheduling service, which essentially acts as a thin wrapper around action scheduler, handling set up of scheduling in the background.

* Background actions can be scheduled to run at regular intervals
* The background cache service inserts new hooks or cleans up old ones hourly
* Removing an action from the background service automatically dequeues the action scheduler pending action

This PR also hooks up the background scheduler service to order counts and forcibly refreshes them to ensure cache is more readily available and help keep numbers accurate in case the "dynamic cache" in https://github.com/woocommerce/woocommerce/pull/54034 results in small discrepancies.

```php
use Automattic\WooCommerce\Admin\Schedulers\BackgroundAction;
use Automattic\WooCommerce\Admin\Schedulers\BackgroundScheduler;

$background_scheduler = new BackgroundScheduler();
$action = new BackgroundAction(
	array(
		'hook' => 'test_action',
		'interval_in_seconds' => 21600,
		'callback' => function() {
			return wp_cache_set( 'test_cache_key', $data );
		},
	)
);
$background_scheduler->register_action( $action );
```

Closes #54819  .

1. Navigate to Tools -> Scheduled Actions and search for a pending action with the ID `woocommerce_background_scheduler` (`/wp-admin/tools.php?page=action-scheduler&status=pending&s=woocommerce_background_scheduler&action=-1&paged=1&action2=-1`)
2. Click "Run" by hovering over the action hook name.
3. Look for a pending scheduled action of `woocommerce_order_count_shop-order`
4. Make sure the action is pending and will be run in approximately 12 hours.
5. Run the action.
6. Find the completed action and make sure that no errors have occurred.
7. Navigate to WooCommerce -> Orders.
8. Make sure that order counts are correct.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Caching/BackgroundCache.php
```diff
@@ -0,0 +1,91 @@
...
+	 * Any actions that have been removed will be unscheduled.
+	 *
+	 * @param string $id Action ID.
+	 * @return void
+	 */
+	public function run_callback( $id ) {
+		$action = $this->scheduled_actions[ $id ] ?? null;
+
+		if ( ! $action ) {
+			as_enqueue_async_action( self::UNSCHEDULE_HOOK_NAME, array( $id ) );
```
joshuatf: This has to be done in an async event as AS will not find the action we are currently processing and will have to cancel the next scheduled one.

plugins/woocommerce/src/Caches/OrderCountCacheService.php
```diff
@@ -29,6 +32,35 @@ final public function init() {
...
+		foreach ( $order_types as $order_type ) {
+			$background_cache_action = new CacheAction(
+				array(
+					'callback'            => function() use ( $order_type ) {
+						error_log( 'Removing order count cache for ' . $order_type );
+						$this->order_count_cache->remove( $order_type );
+						OrderUtil::get_count_for_type( $order_type );
+					},
+					'force_refresh'       => true,
+					'interval_in_seconds' => HOUR_IN_SECONDS * 12,
```
joshuatf, line 55: At the time of writing, the expiration for order counts cache is 24 hours.  This forcibly refreshes the cache well before we get to that point so we (hopefully) never see an expired cache and as a failsafe to keep numbers from getting out of sync in case we run into any issues in https://github.com/woocommerce/woocommerce/pull/54034.

plugins/woocommerce/src/Caching/BackgroundCache.php
```diff
@@ -0,0 +1,91 @@
...
+
+	/**
+	 * Run a background cache callback if the item is no longer cached or force refresh is true.
+	 * Any actions that have been removed will be unscheduled.
+	 *
+	 * @param string $id Action ID.
+	 * @return void
+	 */
+	public function run_callback( $id ) {
+		$action = $this->scheduled_actions[ $id ] ?? null;
```
prettyboymp: What do you think about adding a hook here to allow the scheduled background cache action to be registered with the `->scheduled_actions` prior to this point?

This would allow a late registration of it when it's needed and help avoid every action that needs cache refreshing from having to register on every request.

plugins/woocommerce/src/Caching/BackgroundCache.php
```diff
@@ -0,0 +1,91 @@
...
+	 *
+	 * @param CacheAction
+	 * @return void
+	 */
+	public function schedule_action( $background_cache_action ) {
+		if ( ! is_a( $background_cache_action, CacheAction::class ) ) {
+			throw new \InvalidArgumentException( 'Background cache action must be an instance of CacheAction' );
+		}
+
+		if ( function_exists( 'as_schedule_recurring_action' ) && false === as_next_scheduled_action( self::BACKGROUND_HOOK_NAME, array( $background_cache_action->get_id() ) ) ) {
```
prettyboymp: The calling of `as_next_scheduled_action()` can get pretty slow when the scheduled actions table gets filled up.   This is especially common in setups using HPOS syncing, Subscriptions, or webhooks.  Can we use `as_has_scheduled_action()` if it exists, it performs better for this check since it doesn't require sorting.  It's still not great, but at least an improvement if we just need to check if an action is already scheduled.

The use of `$args` with action scheduler also doesn't scale well when checking for an existing action as the indexes on the table just isn't able to support it.

prettyboymp (commented):
Thinking out loud here...

While reviewing this I was wondering if this would have more flexibility if it didn't worry about the `$cache_key` or `$force_refresh` and to have the callback handle all of that.  My thoughts were to avoid developers creating yet another cache key as a wrapper around more complex data that may live in multiple keys that are updated at the same time.  I realized though, that at that point, this wrapper really wouldn't simplify anything.

So I questioned why we need the wrapper and it led me back to thinking that the underlying issue goes back to ActionScheduler.  There is a lot of performance overhead in working with it which is why we hesitate to just register to a bunch of new actions to deal with keeping cache primed.

It would be great if we could come up with a way to have integrations only attempt to schedule the actions at specific points rather than on every request.  What do you think about having a scheduled action in core that is specifically used to schedule other actions?  

Just a rough idea:
E.g. a top level `woocommerce_background_caching_schedule` action that ran once every hour which only ran a filter for a cache clearing schedule/callback.  The BackgroundCache would take care of making sure all of the callbacks returned from the filter were scheduled.  Then, the action to run the callback would use a filter (the same or more callback specific) prior to attempting to run the callback so we avoid the need to load any extra info out of context of the BackgroundCache actions.

https://github.com/woocommerce/woocommerce/issues/53564 and https://github.com/woocommerce/action-scheduler/issues/1240 are somewhat related around the attempts to reduce the number and slowness of queries running on every page due to extensions making sure actions are scheduled.

Sorry, I know this bounced around a bit.

cc @barryhughes

plugins/woocommerce/src/Caching/BackgroundCache.php
joshuatf: We can move to `as_has_scheduled_action` and remove the use of `$args` in favor of action names that include the ID.

plugins/woocommerce/src/Caching/BackgroundCache.php
joshuatf: I see what you mean, though I'm not sure that registration adds that much overhead (even if we had 100s of registrations per page load) compared to "registration" via a hook.

I think in this case, I'd favor the more declarative API that allows us to validate early and does not add overhead.  But let me know if you think differently on how much overhead is being added with registration.

joshuatf:
>this would have more flexibility if it didn't worry about the $cache_key or $force_refresh and to have the callback handle all of that

@prettyboymp I had similar thoughts as I was writing this, though I was hesitant to make such a thin wrapper unless it offered some benefits.  I think in this case, it still might as dealing with scheduling, unscheduling, and callbacks alone can act as small hurdles for what should be a relatively simple task.

>It would be great if we could come up with a way to have integrations only attempt to schedule the actions at specific points rather than on every request. What do you think about having a scheduled action in core that is specifically used to schedule other actions?

I'm not opposed to this approach.  I think if we remove the conveniences around caching, then this would more aptly be named `BackgroundScheduler` or something along those lines.

joshuatf:
@prettyboymp I updated this class per our discussion to be a more generic background scheduler.  It does make this a rather thin wrapper around Action Scheduler though might be worth it if the following benefits are high enough:

1. It significantly lowers the barrier to scheduling an action by removing the need to add additional hooks, checking scheduled actions, and cleaning up unused scheduled actions.
2. The performance of checking scheduled actions during an admin page load is reduced and all scheduling (aside from the main action) occurs in the background.

Let me know what you think of this implementation.

plugins/woocommerce/changelog/fix-53996
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: performance
+
+Cache order counts for use in the admin
```
kalessil: One changelog entry is probably enough =)

plugins/woocommerce/src/Utilities/OrderUtil.php
```diff
@@ -205,10 +206,10 @@ public static function get_table_for_order_meta() {
 	public static function get_count_for_type( $order_type ) {
 		global $wpdb;
 
-		$cache_key        = \WC_Cache_Helper::get_cache_prefix( 'orders' ) . 'order-count-' . $order_type;
-		$count_per_status = wp_cache_get( $cache_key, 'counts' );
+		$order_count_cache = new OrderCountCache();
```
kalessil: I guess we can use `wc_get_container()->get( ... )` in this case. Also https://github.com/woocommerce/woocommerce/pull/56215/files#diff-481d2d3f71e191dd29104f0c5d1d07050f1a74ef49c72cafc9bd79de291476e8R109 will fix code navigation/resolution for the method in case it was reason for instantiate the object directly.

Also this way we have more flexibility injecting a substitute objects in our tests when needed.

kalessil: And there are more places in this PR where we could use `wc_get_container()->get( ... )`.

plugins/woocommerce/src/Internal/DependencyManagement/ServiceProviders/OrdersDataStoreServiceProvider.php
```diff
@@ -81,6 +84,7 @@ public function register() {
 		);
 		$this->share( OrderCache::class );
 		$this->share( OrderCacheController::class )->addArgument( OrderCache::class );
+		$this->share( OrderCountCacheService::class )->addArgument();
```
kalessil: Do we need `->addArgument();` here?

plugins/woocommerce/src/Internal/DependencyManagement/ServiceProviders/BackgroundSchedulerServiceProvider.php
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Service provider for the background cache mechanism.
+ */
+class BackgroundSchedulerServiceProvider extends AbstractServiceProvider {
+
+	/**
+	 * The classes/interfaces that are serviced by this service provider.
+	 *
+	 * @var array
```
kalessil, line 16: Nit: `array<int,class-string>`

plugins/woocommerce/src/Admin/Schedulers/BackgroundAction.php
```diff
@@ -0,0 +1,120 @@
+<?php
+
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\Admin\Schedulers;
+
+/**
+ * An action that can be scheduled to run in the background.
+ */
+class BackgroundAction {
```
kalessil, line 10: Looks like DTO to me, we could potentially make it final.

plugins/woocommerce/src/Caches/OrderCountCacheService.php
```diff
@@ -0,0 +1,147 @@
...
+	 * @param WC_Order $order The order.
+	 */
+	public function update_on_order_trashed( $order_id, $order ) {
+		if (
+			! $this->order_count_cache->is_cached( $order->get_type(), $this->get_prefixed_status( $order->get_status() ) ) ||
+			! $this->order_count_cache->is_cached( $order->get_type(), OrderStatus::TRASH ) ) {
+			return;
+		}
+
+		$this->order_count_cache->decrement( $order->get_type(), $this->get_prefixed_status( $order->get_status() ) );
```
kalessil: Nit:

```php
    $is_incremented = false !== $this->order_count_cache->increment( .... );
    if ( $is_incremented ) {
        $this->order_count_cache->decrement( .... );
    }
```

plugins/woocommerce/src/Caches/OrderCountCacheService.php
```diff
@@ -0,0 +1,147 @@
...
+	 */
+	public function update_on_order_status_changed( $order_id, $previous_status, $next_status, $order ) {
+		if (
+			! $this->order_count_cache->is_cached( $order->get_type(), $this->get_prefixed_status( $next_status ) ) ||
+			! $this->order_count_cache->is_cached( $order->get_type(), $this->get_prefixed_status( $previous_status ) )
+		) {
+			return;
+		}
+
+		$this->order_count_cache->decrement( $order->get_type(), $this->get_prefixed_status( $previous_status ) );
```
kalessil: Nit:

```php
    $is_incremented = false !== $this->order_count_cache->increment( .... );
    if ( $is_incremented ) {
        $this->order_count_cache->decrement( .... );
    }
```

plugins/woocommerce/src/Caches/OrderCountCacheService.php
```diff
@@ -0,0 +1,147 @@
...
+	private $order_count_cache;
+
+	/**
+	 * Class initialization, invoked by the DI container.
+	 *
+	 * @internal
+	 */
+	final public function init() {
+		$this->order_count_cache = new OrderCountCache();
+		add_action( 'woocommerce_new_order', array( $this, 'update_on_new_order' ), 10, 2 );
```
kalessil: The hooks list looks good!

plugins/woocommerce/src/Caches/OrderCountCache.php
```diff
@@ -0,0 +1,189 @@
...
+		$cache_key = $this->get_cache_key( $order_type, $order_status );
+		return wp_cache_set( $cache_key, $value, '', $this->expiration );
+	}
+
+	/**
+	 * Get the cache value for a given order type and status.
+	 *
+	 * @param string $order_type The type of order.
+	 * @param string[] $order_statuses The statuses of the order.
+	 * @return int[] The cache value.
```
kalessil: Nit: `@return array<string,int>`

plugins/woocommerce/src/Caches/OrderCountCache.php
```diff
@@ -0,0 +1,189 @@
...
+		return wp_cache_incr( $cache_key, $offset );
+	}
+
+	/**
+	 * Decrement the cache value for a given order status.
+	 *
+	 * @param string $order_type The type of order.
+	 * @param string $order_status The status of the order.
+	 * @param int $offset The amount to decrement by.
+	 * @return int The new value of the cache.
```
kalessil: Here and other wrappers for `wp_cache_*`: those functions might return false.

kalessil (commented):
Leaving some comments for code itself and will do manual/exploratory testing tomorrow.
PR is well shaped and I think we'll need any CFP.

joshuatf:
Hey @kalessil, sorry but I think the base branch automatically got bumped to `trunk` and most of the items you reviewed were from https://github.com/woocommerce/woocommerce/pull/54034.  This PR is only around the background scheduler.  I've pulled the latest changes so they should be reflected in the GH diff.

The question at this point is which of the following approaches do we take?

1. An opinionated scheduler API that helps register background jobs (current approach in this PR).  This adds some convenience in for register callbacks and unscheduling actions that are no longer registered.
2. A simple async hook (e.g., `woocommerce_background_scheduler_action`) that runs every hour and allow consumers to hook and set up scheduling for their own jobs.

I was in favor of approach 1 initially because I was under the impression that the nuisance to setting up and removing scheduled jobs is a barrier to caching for many plugins.  However, this creates a larger surface area in terms of API.  The latter approach adds more work for consumers to set up, though minimizes surface area.

Personally, I'm not strongly leaning towards either approach ATM and am fine to go with approach 2 if we think that the nuisance to setting up scheduled jobs is not that large.

joshuatf:
@kalessil @prettyboymp I put together the alternative approach in https://github.com/woocommerce/woocommerce/pull/56380 so that we can more easily compare.

kalessil:
@joshuatf, I like the approach in this PR more than the alternative:

- it's more modular, and as you noticed, proper UTing is more manageable compared to hooks
- also IIUC, we are better off setting the scheduler solution as an internal API for the core

One question, though: do you think we should mark the new components as `@internal` or allow external devs to use the new APIs?

joshuatf:
> One question, though: do you think we should mark the new components as @internal or allow external devs to use the new APIs?

I'm in favor of allowing external use to lower the barrier to refreshing cache for other extensions.  It seems to be a common issue amongst extensions and if we can create an easy way to handle this, we can hopefully reduce these queries on regular page loads across the board.

kalessil:
> I'm in favor of allowing external use to lower the barrier to refreshing cache for other extensions. It seems to be a common issue amongst extensions and if we can create an easy way to handle this, we can hopefully reduce these queries on regular page loads across the board.

If we agree to move forward with this implementation, we'll need to provide some documentation on the intended usage of this API. That'll make maintenance a bit easier for us.

joshuatf:
Closing in favor of https://github.com/woocommerce/woocommerce/pull/56380 which may be moved upstream to AS.

PR #55757, opened 2025-02-21
[dev] CI: Update pr-project-label to label PRs based on changelog type

### LINKED ISSUE #55713: [Enhancement]: Label PRs with a type based on changelog entry

To better help us gather metrics automatically on types of PRs included in each release we should label the PRs with their type.
We can use the type chosen for the corresponding changelog entry to determine the type.

_No response_

### Additional context

_No response_

tammullen:

Closes https://github.com/woocommerce/woocommerce/issues/55713 .

Adds a `type:` label automatically based on the changelog type (if present).

Based on https://github.com/woocommerce/woocommerce/actions/runs/13591013123/job/37997009429 the new job does make the workflow execution time increase from ~4s to ~48s. Open to suggestions if there is a more efficient way to tackle this.

1. Create a PR targeting this branch
2. Add a changelog entry to the branch, make it of type Update
3. Check that label `type: enhancement` is added to the PR

Repeat for the following Changelog types

Changelog Type |PR Label
-- | --
`Fix` | `type: fix`
`Add` | `type: feature`
`Update` | `type: enhancement`
`Tweak` | `type: enhancement`
`Enhancement` | `type: enhancement`
`Performance` | `type: performance`
`Dev` | `type: infrastructure`
no changelog | no label will be applied

Tested using https://github.com/woocommerce/woocommerce/pull/55934

Example of run https://github.com/woocommerce/woocommerce/actions/runs/13591013123/job/37997009429 

Also performed an LLM analysis.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment  Not included in release zip

</details>


.github/workflows/pr-project-label.yml
```diff
@@ -22,3 +22,86 @@ jobs:
...
+            pull-requests: write
+        needs: label_project
+        steps:
+          - name: Checkout code
+            uses: actions/checkout@v4
+            with:
+              ref: refs/pull/${{ github.event.pull_request.number }}/merge  # Ensures only merged changes are used
+              fetch-depth: 0
+
+          - name: Validate Trusted Execution
```
adimoldovan: Not sure why this step is needed. 
If it is, it could run as the first step, before checkout.

.github/workflows/pr-project-label.yml
```diff
@@ -22,3 +22,86 @@ jobs:
...
+        permissions:
+            contents: read
+            pull-requests: write
+        needs: label_project
+        steps:
+          - name: Checkout code
+            uses: actions/checkout@v4
+            with:
+              ref: refs/pull/${{ github.event.pull_request.number }}/merge  # Ensures only merged changes are used
+              fetch-depth: 0
```
adimoldovan: I don't think we need to fetch all history here, `fetch-depth: 1` should be enough. It should also improve the execution time.

.github/workflows/pr-project-label.yml
```diff
@@ -22,3 +22,86 @@ jobs:
...
+                  echo "Processing file: $FULL_PATH"
+    
+                  # Extract the type from file content
+                  TYPE=$(grep -iE '^Type: (fix|add|update|tweak|enhancement|performance|dev)' "$FULL_PATH" | awk -F': ' '{print tolower($2)}' | head -n 1)
+                  echo "Type: $TYPE"
+
+                  # Validate extracted type
+                  if [[ ! "${LABEL_MAP[$TYPE]}" ]]; then
+                    echo "Invalid changelog type found in $FULL_PATH: '$TYPE'. Skipping."
+                    continue
```
adimoldovan: Should it not fail if the type was not found? It could signal a change in the changelog types that we want to include in here.

.github/workflows/pr-project-label.yml
```diff
@@ -22,3 +22,86 @@ jobs:
...
+              CHANGED_FILES=$(gh pr view "$PR_NUMBER" --json files -q '.files[].path' | grep '^plugins/woocommerce/changelog/' || true)
+    
+              echo "Changed files detected:"
+              echo "$CHANGED_FILES"
+    
+              # Loop through only changed changelog files
+              for FILE in $CHANGED_FILES; do
+                FULL_PATH="$GITHUB_WORKSPACE/$FILE"
+    
+                if [[ -f "$FULL_PATH" ]]; then
```
adimoldovan: This condition seems redundant to me. If the file was found by the gh pr view command the path should be valid.

.github/workflows/pr-project-label.yml
```diff
@@ -22,3 +22,86 @@ jobs:
...
+    
+                  if [[ -n "$LABEL" && ! " ${APPLIED_LABELS[*]} " =~ " $LABEL " ]]; then
+                    APPLIED_LABELS+=("$LABEL")
+                  fi
+                else
+                  echo "Skipping non-existent file: $FULL_PATH (Check if checkout fetched all changes)"
+                fi
+              done
+    
+              echo "APPLIED_LABELS: ${APPLIED_LABELS[*]}"
```
adimoldovan: Naming nitpick: the labels are not yet applied at this point, so maybe LABELS_TO_BE_APPLIED is better?

.github/workflows/pr-project-label.yml
adimoldovan: After some testing, I found that it can happen for the file to not exists, but none of these cases apply on this workflow's trigger. Examples:
- run from any branch with a PR number of a PR that is not yet merged
- run from trunk for a PR that from a previous release that got its changelog removed

adimoldovan (commented):
Nice work on this @tammullen! I like that is has good error handling and detailed logging to better understand what's going on.
I would love it though for the code to be a standalone script. It would be easier to test, because you could run it locally with some env variables. This is how I tested it, by adding the code in a script called `label-pr.sh`:
```GITHUB_WORKSPACE=<my-local-path> PR_NUMBER=55857 label-pr.sh```

> the new job does make the workflow execution time increase from ~4s to ~48s. Open to suggestions if there is a more efficient way to tackle this.

See `sparse-checkout`: https://github.com/actions/checkout#fetch-only-the-root-files-and-github-and-src-folder. If I understood their docs correctly you should be able to only checkout a part of the repo, in our case `plugins/woocommerce/changelog` should be enough.  

Left some comments on code.

adimoldovan (approved):

PR #55787, opened 2025-02-24
Merge Back In Stock Notifications to WC core, pt 1: feature code

peterfabian:

This PR is the first WIP to initiate the merge of the [Back In Stock Notifications plugin](https://woocommerce.com/products/back-in-stock-notifications/) (BIS) to WC core.

The initial plan is to:
- Merge the code & keep it disabled for WC 9.9.
- Enable it for 5% of merchants in WC 10.0
- Enable it for all merchants in WC 10.1

#### What Does Enable Even Mean 
In order to prevent changes in the Single Product page design, the BIS, even when enabled, will have sign-ups **turned off** for out of stock products (`Allow sign-ups` unchecked). The code will be loaded, any pending notifications will be processed, settings will be accessible, etc.

Except for:
- stores that had BIS active and the setting was set to active
- new stores (approximated as stores with 0 products).

#### Data From The Past Reused

All the settings and data from the past (like customers subscribed to out of stock products) should be used correctly by the merged version. 

If merchant used BIS in the past, and disabled it at some point, then with the update to the new version of WC that includes BIS, these notifications will get processed (see decision log for discussion).

The only exception to this is the "Allow sign-ups" mentioned above. For all stores, except the ones that have BIS active & this setting set to true at the moment WC+BIS is getting installed, this setting will be set to false.

#### New Settings Location

The **BIS settings have changed their location** and will now be found under WC > Settings > Products > Customer Stock Notifications instead of WC > Settings > Stock Notifications. 

This section will be displayed once `Enable back in stock notifications for customers` under WC > Settings > Advanced > Features is checked. 

This checkbox governs the loading of the package (and thus also of all the merged BIS code) and thus acts as a kill-switch for the merged BIS code. It's linked to option `wc_feature_woocommerce_back_in_stock_notifications_enabled` that is utilized in `Packages.php` to `prep()`-are and `init()`-ialize the merged code.

This slightly differs from Brands, where the option to disable the Brands code wasn't available in the Settings UI.

#### DB initialization

BIS creates 3 db tables:
- wp_woocommerce_bis_activity
- wp_woocommerce_bis_notifications
- wp_woocommerce_bis_notificationsmeta

There are 2 options for how db tables for BIS can be created:
- during the install procedure
- during the activation of merged BIS code

This is fairly similar to how HPOS initializes its db tables.

##### DB Tables Created During Install
Same as for HPOS, BIS can add its db table schema to `WC_Install::get_schema`, if it's enabled. This way, the db tables can be created along with WC installation.

##### DB Tables Created On Activation  
Similarly to how HPOS tables are created based on the UI switch, here, it's linked to the `wc_feature_woocommerce_back_in_stock_notifications_enabled` option. When merged BIS gets activated using add/update/delete_option WordPress functions, the check for db tables is executed. If they're missing, the script tries to create them.

##### What If DB Tables Are Still Missing?
If the db tables are missing, there will be some errors in the logs, but it shouldn't end up with a fatal error.
As is the case for other WC tables, running WC > Status > Tools > `Verify base database tables` attempts to create the BIS DB tables again.

#### Activation

The activation is quite delicate and works as follows:
- the feature wp option should override the rollout settings (which is governed by BackInStockNotifications::is_enabled()).
- the feature wp option is initially without any value; in this state, the check if BIS feature is enabled falls back to the rollout setting.
- The WC > Settings > Advanced > Features UI checkbox has a default value that is set based on the rollout setting (but this is overridden by the feature wp option once it's set).
- the feature wp option is set (and later updated) when the store manager checks/unchecks the checkbox in the settings and clicks Save settings.

This way, the standalone BIS plugin can stay active for a customer with the merged BIS rollout set to false. Once the rollout is switched to yes, the standalone BIS plugin should be disabled automatically and replaced by the merged BIS instance.

There is one more thing that happens during activation: `wc_bis_daily` cron task is scheduled. 

#### Changes In The BIS Plugin

I tried to keep the changes in the BIS plugin to the minimum and merge the files as-is. I also tried to not cause any disruption to the public API of the plugin (i.e. I didn't remove any public methods, etc).

List of notable changes/things worth checking:
- I added **deprecations** to all publicly facing methods that don't make sense anymore. I might have missed some, though, as I haven't read all the code across the whole plugin. Also changed their return values to neutral values.
- I **removed some constants** that don't make sense anymore (`WC_BIS_VERSION`, `WC_BIS_SUPPORT_URL`, `WC_BIS_ABSPATH`), but that's somewhat questionable, as those are part of public interface, after all. Might need to reconsider.
- I updated the text domain to `woocommerce`
- I removed **selectSww** dependency, as that was added to solve a 1-pixel design problem. Third version of the select2 library within WC core would be too much.
- I **dropped** all the **WP Admin notices**. We might want a welcome notice, but it will probably be better to implement it using WC infrastructure.
  - this includes **loopback** tests and **Action Scheduler check** to help troubleshoot the plugin issues. These two functionalities would be better solved in the context of WC core itself.
- I **dropped** the **db update code** that copied a bunch of infrastructure from WC core, including the background processing libraries. There was also just one db update method that served to update to BIS v 1.1.0 (which was released in July 2021).
- I kept `WC_BIS()` function to make things backward compatible, but had to be more careful during initialization to not cause fatal errors (e.g. when someone attempts to activate the standalone BIS plugin while WC+BIS is active).
- I updated the time when the initialization of BIS code runs from `plugins_loaded` priority 9 to priority 10 (where it's initialized by `Packages.php`), see below for rationale.
- I updated `WC_BIS()->log()` calls to `wc_get_logger`.

##### BIS init moved from `plugins_loaded` priority 9 to priority 10

`Packages.php` has the `on_init` method hooked to `plugins_loaded` at priority 10, while the standalone BIS plugin is initialized at priority 9. It would probably be possible for me to unhook the existing action in `BackInStockNotifications::prepare` and rehook it to priority 9, then after initializing BIS rehook it back at priority 10, but it feels pretty hack-y while my guess is that it doesn't really matter too much if BIS is initialized at priority 9 or 10, so I've decided to change the BIS initialization priority instead of hacking `Packages.php`.

### PR Structure

The code has been split into 2 parts:
- the feature code that has been mostly copied from BIS plugin is part of this branch
- unit tests and e2e tests (rewritten to Playwright) have been moved to a new branch: https://github.com/woocommerce/woocommerce/pull/55836 to work around the limitation of 20k lines in one diff in GH.

### TODO:

- [x] Sort out the feature announcement
- [ ] Update since tags
- [x] Add tracking
- [ ] The plugin list looks slightly odd if there isn't any update to the BIS plugin (there's an extra border/shadow/sth)

##### Build disabled zip package

1. Check out this branch of WC.
2. Open `src/Internal/BackInStockNotifications.php`, set `BackInStockNotifications::is_enabled` to return false (or it's already set to return false).
3. Build the zip file, e.g. by running `./bin/build-zip.sh` from within `plugins/woocommerce`.
4. Rename the plugin to sth like `woocommerce-9.9.0-BIS-disabled.zip`.

##### Build enabled zip package with bumped version

1. Check out this branch of WC.
2. Open `src/Internal/BackInStockNotifications.php`, set `BackInStockNotifications::is_enabled` to return true.
3. Update WC versions in woocommerce.php (plugin comment at the top) and includes/class-woocommerce.php (first property of the class) to 9.9.1. (to simulate update later)
3. Build the zip file, e.g. by running `./bin/build-zip.sh` from within `plugins/woocommerce`.
4. Rename the plugin to sth like `woocommerce-9.9.1-BIS-enabled.zip`.

##### Test Rollout Disabled to Rollout Enabled Store with BIS plugin

This test confirms that a store with standalone BIS keeps the plugin if store is assigned to the inactive cohort at first, but it switches to merged BIS once the merged code gets enabled.

1. Start with a clean environment.
2. Install the current version of WC (latest stable).
6. Install & activate BIS notifications standalone plugin.
7. Go to WC > Settings > Stock Notifications and check that the "Allow sign-ups" is checked/turned on.
8. Install the disabled zip file `woocommerce-9.9.0-BIS-disabled.zip`.
9. There should be no error visible on the front end or in the log.
10. Standalone BIS plugin should remain active.
11. The Allow sign-ups should still be checked/turned on.
12. Install the enabled zip file `woocommerce-9.9.1-BIS-enabled.zip`.
13. A notice should appear saying that BIS plugin was disabled.
14. BIS plugin should be disabled.
15. Merged BIS should be active (WC > Stock Notifications should still appear in the menu).
16. Try to activate BIS plugin.
17. You should see wp die message saying that the plugin can't be enabled.
> This plugin cannot be activated because its functionality is now included in WooCommerce core.
> Return to the Plugins page

##### Test UI option in WC > Settings > Advnaced > Features overrides Rollout Disabled 

1. Start with a clean environment.
2. Install the current version of WC (latest stable).
3. Install the disabled zip file `woocommerce-9.9.0-BIS-disabled.zip`.
4. Go to WC > Settings > Advanced > Features.
5. Check the "Enable back in stock notifications for customers" checkbox.
6. Save settings.
7. Verify that BIS functionality becomes available despite the rollout being disabled.
8. Check that WC > Settings > Products > Customer Stock Notifications is accessible.
9. Verify that BIS settings and functionality work as expected.

##### Test UI option in WC > Settings > Advnaced > Features overrides Rollout Enabled

1. Start with a clean environment.
2. Install the current version of WC (latest stable).
3. Install the enabled zip file `woocommerce-9.9.1-BIS-enabled.zip`.
4. Go to WC > Settings > Advanced > Features.
5. Uncheck the "Enable back in stock notifications for customers" checkbox.
6. Save settings.
7. Verify that BIS functionality becomes unavailable despite the rollout being enabled.
8. Check that WC > Settings > Products > Customer Stock Notifications is no longer accessible.
9. Verify that BIS settings and functionality are disabled.

##### Test Rollout Enabled State without BIS plugin---new site

This test case checks that new stores have everything enabled by default when they upgrade 

1. Start with a clean environment.
2. Install the current version of WC (latest stable).
8. Install the enabled zip file `woocommerce-9.9.1-BIS-enabled.zip`. 
9. There should be no error visible on the front end or in the log.
10. Merged BIS should be active (e.g. you should see WC > Stock Notifications menu item).
11. The Allow sign-ups should still be checked/turned on.

##### Test Rollout Enabled State without BIS plugin---existing site

This test case checks that single product page won't be changed for out of stock products for existing stores with products.

1. Start with a clean environment.
2. Install the current version of WC (latest stable).
3. Create at least one product/import the sample product CSV file.
8. Install the enabled zip file `woocommerce-9.9.1-BIS-enabled.zip`. 
9. There should be no error visible on the front end or in the log.
10. Merged BIS should be active (e.g. you should see WC > Stock Notifications menu item).
11. The Allow sign-ups should still be unchecked/turned off. 

##### Test Rollout Enabled State without BIS plugin: try to force activate BIS plugin via WP-CLI

1. Start with a clean environment.
2. Install the current version of WC (latest stable).
3. Install the enabled zip file `woocommerce-9.9.1-BIS-enabled.zip`.
4. Try to activate the standalone BIS plugin using WP-CLI:
   ```
   wp plugin activate woocommerce-back-in-stock-notifications
   ```

5. Verify that the command fails with an appropriate error message indicating the plugin cannot be activated.
6. Check that the standalone BIS plugin remains deactivated.
7. Verify that the merged BIS functionality remains active.

##### Test Rollout Enabled State with BIS plugin: verify data and settings remain from the BIS plugin (sign up for notifications and send some out)

1. Start with a clean environment.
2. Install the current version of WC (latest stable).
3. Install & activate BIS notifications standalone plugin.
4. Create a test product and set it as out of stock.
5. Sign up for back-in-stock notifications using both guest and logged-in user accounts.
6. Configure BIS settings (email templates, thresholds, etc.).
7. Install the enabled zip file `woocommerce-9.9.1-BIS-enabled.zip`.
8. Verify that:
   - All existing notification subscriptions are preserved
   - All BIS settings are maintained
   - The notification subscriptions are intact
9. Set the product back in stock and verify that notifications are sent correctly
10. Check that the notification history is maintained

##### Test Rollout Enabled State with BIS plugin: it's possible to roll back to WC version without BIS and use standalone BIS.

1. Start with a clean environment.
2. Install the current version of WC (latest stable).
3. Install & activate BIS notifications standalone plugin.
4. Create test data (products, subscriptions, settings).
5. Install the enabled zip file `woocommerce-9.9.1-BIS-enabled.zip`.
6. Verify the migration worked correctly.
7. Roll back to WC version without BIS (e.g., 9.8.x).
8. Reactivate the standalone BIS plugin.
9. Verify that:
   - All data and settings are preserved
   - The standalone plugin works as expected
   - No data loss occurred during the rollback

##### Test Rollout Enabled State with BIS plugin: check that BIS settings have moved from WC > Settings > Stock Notifications to WC > Settings > Products > Customer stock notifications

1. Start with a clean environment.
2. Install the current version of WC (latest stable).
3. Install & activate BIS notifications standalone plugin.
4. Configure various BIS settings.
5. Check WC > Settings > Stock Notifications to confirm it's there. Change a couple settings and remember it.
6. Check WC > Stock Notifications > Activity log and remember the output.
5. Install the enabled zip file `woocommerce-9.9.1-BIS-enabled.zip`.
6. Verify that:
   - The old settings location (WC > Settings > Stock Notifications) is no longer accessible
   - All settings are available in the new location (WC > Settings > Products > Customer Stock Notifications)
   - All previously configured settings are preserved in the new location
7. Verify that:
   - The old Activity tab location is no longer accessible
   - All activity data is visible in the new location (WC > Status > Customer Stock Notifications)
   - Historical activity data is preserved
7. Test saving new settings in the new location
8. Verify that the settings are properly applied

##### Check The Included BIS Works As Expected

1. Set some products out of stock.
2. Sign up to receive their back in stock notifications as logged in user and guest.
3. Set some of the products to back in stock (don't forget it needs to cross the threshold set in the BIS settings).
4. Check that the notification emails have been sent correctly (e.g. via MailPit or some other means).
5. Check that `wc_bis_daily` task is scheduled via cron.
6. Check that the REST API command works to deactivate all notifications for a product:

```
curl -X POST --location "https://example.com/wp-json/wc/v3/back-in-stock/notifications/deactivate" \
    -H "Accept: application/json" \
    -H "Content-Type: application/json" \
    -d '{
          "productId": 67
        }' \
    --basic --user ck_115:cs_204
```
7. Ensure the unit tests included in the PR pass.
8. Ensure the e2e tests included in the PR pass.

--

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [x] Major

#### Type

#### Message 
Added Back In Stock Notifications functionality to WC core.
</details>

<details>

#### Comment 

</details>


peterfabian:
Some jobs failed because the diff it too large, not sure what to do about it.

> Error: Command failed: gh pr diff 55787 --name-only
could not find pull request diff: HTTP 406: Sorry, the diff exceeded the maximum number of lines (20000) (https://api.github.com/repos/woocommerce/woocommerce/pulls/55787)
PullRequest.diff too_large

peterfabian:
I tried to decrease the number of lines below 20k, but apparently it's not only about changed lines (+/-), but also includes context lines, so this needs further cutting. 

Edit: moved the tests to a separate PR: https://github.com/woocommerce/woocommerce/pull/55836, now it runs the CI checks.

plugins/woocommerce/src/Internal/BackInStockNotifications.php
```diff
@@ -0,0 +1,353 @@
...
+		// Include BIS files.
+		include_once WC_ABSPATH . '/includes/bis/class-wc-back-in-stock.php';
+
+		$wc_bis = wc_get_container()->get( LegacyProxy::class )->call_function( 'WC_BIS' );
+		$wc_bis->initialize_plugin();
+
+	}
+
+	/**
+	 * Deactivate signups for BIS to prevent changing the single product screen.
```
coreymckrill, line 156: I suspect this description will not age well. What does "changing the single product screen" mean in this case?

coreymckrill (commented):
> Testing Disabled State

I followed the instructions, and built the zip of WC with the `is_enabled` method returning false (it was already set to do this when I checked out the branch). However, when installing the new version of WC, it appears that BIS was automatically disabled, and shows the message `This plugin can no longer be activated because its functionality is now included in WooCommerce. It is recommended to delete it.` But the BIS functionality in WC is also turned off, since `is_enabled` returns false.

A couple of other initial thoughts:

* Why is the bulk of the BIS code getting added to `includes` instead of `src`? Is it because it's not namespaced? Ideally we wouldn't be adding new code in `includes`, but I suppose it might be difficult to adapt everything to namespaces if the autoloader for the `src` folder requires their usage.

plugins/woocommerce/src/Internal/BackInStockNotifications.php
peterfabian: If the sign-ups are allowed, there's a change in the single product screen: a new element for buyers to get notified once the product is back in stock.

Screenshot 2025-02-18 at 10 20 14

We wanted to avoid changing this screen for merchants just by updating Woo core, as it might be customized.

peterfabian:
> However, when installing the new version of WC, it appears that BIS was automatically disabled, and shows the message This plugin can no longer be activated because its functionality is now included in WooCommerce. It is recommended to delete it. But the BIS functionality in WC is also turned off, since is_enabled returns false.

Good catch. I think this happens because we fairly recently decided to switch the feature flag option `wc_feature_woocommerce_back_in_stock_notifications_enabled` to a setting (under the WC > Settings > Products > Inventory) and that means it gets a default value `yes`, which makes `Automattic\WooCommerce\Packages` class [think it's enabled](https://github.com/woocommerce/woocommerce/blob/mic/bis-merge-take-2/plugins/woocommerce/src/Packages.php#L135-L138) irrespective of the `is_enabled()` method. I need to think about how to handle this case correctly. Changing the Packages isn't easy as Brands code is already using it, too. Will need to think about the best way to handle this.

> Why is the bulk of the BIS code getting added to includes instead of src? Is it because it's not namespaced? Ideally we wouldn't be adding new code in includes, but I suppose it might be difficult to adapt everything to namespaces if the autoloader for the src folder requires their usage.

I've reused the same approach that was used to merge Brands plugin. Previous discussions on this topic can be found in the slack channel: p1721257169476979-slack-C07DH5W04L8 and comments below, but the tl;dr is that the code fits best into the `includes` folder and its subdirectories, as it doesn't follow PSR-4 and doesn't have namespaces. Some more discussion also wrt to autoloading over here: p1725283129091499-slack-C07DH5W04L8. Hope that makes sense.

plugins/woocommerce/src/Internal/BackInStockNotifications.php
coreymckrill: Ah, gotcha. I think it would be beneficial to make this doc block more explicit so that in a couple of years we can still understand what this was talking about.

plugins/woocommerce/includes/bis/class-wc-bis-sync.php
```diff
@@ -0,0 +1,274 @@
...
+	 * @return void
+	 */
+	public function sync() {
+
+		if ( ! $this->sync_needed ) {
+			return;
+		}
+
+		if ( wc_bis_debug_enabled() ) {
+			WC_BIS()->log( 'Product IDs: ' . print_r( $this->queue, true ), 'info', 'wc_bis_sync_logs' );
```
coreymckrill: It would be nice to just use `wc_get_logger()` directly here, and also clean this up to use [logging best practices](https://developer.woocommerce.com/docs/logging-in-woocommerce/#3-logging-best-practices) (i.e. don't output the array into the log message).

plugins/woocommerce/includes/bis/class-wc-bis-sync-tasks.php
```diff
@@ -0,0 +1,379 @@
...
+		if ( $throttled_notifications_count > 0 && class_exists( 'WC_BIS_Admin_Notices' ) ) {
+			self::handle_spam_notices( $product_ids, $throttled_notifications_count );
+		}
+
+		if ( 0 === $notifications_count ) {
+			return;
+		}
+
+		if ( wc_bis_debug_enabled() ) {
+			WC_BIS()->log( sprintf( 'Sending %d notifications for products %s ', $notifications_count, implode( ',', $product_ids ) ), 'info', 'wc_bis_sync_logs' );
```
coreymckrill: It would be nice to just use `wc_get_logger()` directly here, and also clean this up to use [logging best practices](https://developer.woocommerce.com/docs/logging-in-woocommerce/#3-logging-best-practices).

plugins/woocommerce/includes/admin/class-wc-bis-admin.php
```diff
@@ -0,0 +1,513 @@
...
+	 */
+	public static function admin_resources() {
+
+		$suffix = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min';
+		$version   = Constants::get_constant( 'WC_VERSION' );
+
+		wp_register_style( 'wc-bis-admin', WC()->plugin_url() . '/assets/css/bis-admin.css', array( 'sw-admin-css-select' ), $version );
+		wp_style_add_data( 'wc-bis-admin', 'rtl', 'replace' );
+
+		wp_register_script( 'wc-bis-writepanel', WC()->plugin_url() . '/assets/js/admin/wc-bis-admin' . $suffix . '.js', array( 'jquery', 'jquery-ui-datepicker', 'wp-util', 'sw-admin-select-init', 'wc-backbone-modal' ), $version );
```
coreymckrill: I'm seeing some errors in Query Monitor related to missing dependencies:

image

image

plugins/woocommerce/src/Internal/BackInStockNotifications.php
peterfabian: Thanks, that's good feedback. After the design review, it seems like this might be changing, so we'll see where we end up with this.

plugins/woocommerce/includes/admin/class-wc-bis-admin.php
peterfabian: Well spotted! I fixed this yesterday but forgot to push. Fixed in [067fdda](https://github.com/woocommerce/woocommerce/pull/55787/commits/067fdda85649fc099a9cd50e6b6392842c6a8537).

plugins/woocommerce/includes/bis/class-wc-bis-sync.php
peterfabian: Makes sense, thanks. Updated in [6508003](https://github.com/woocommerce/woocommerce/pull/55787/commits/6508003c3d0f1fc10032af4249040b53cfeeb0e0).

plugins/woocommerce/includes/bis/class-wc-bis-sync-tasks.php
peterfabian: You're right, updated in [6508003](https://github.com/woocommerce/woocommerce/pull/55787/commits/6508003c3d0f1fc10032af4249040b53cfeeb0e0).

jimjasson:
@peterfabian 

I started testing your PR and here's what I found so far: 

(1) Existing site with Back In Stock Notifications enabled 

I tested this scenario extensively and I confirmed that:
- The Back In Stock Notifications plugin remained active, while core's module was enabled. 
- When I enabled the core's module via the UI, Back In Stock Notifications was automatically disabled and a notice showed up. 
- When I tried to enable Back In Stock Notifications via the Plugins page, it wasn't possible. 
- When I tried to enable Back In Stock Notifications via CLI: `wp plugin activate woocommerce-back-in-stock-notifications` and `wp plugin activate woocommerce-back-in-stock-notifications --skip-plugins=woocommerce`, the plugin remained deactivated. 
- All the settings I had in Back In Stock Notifications were transferred to the Inventory tab and their values were honored. 
- When I changed the settings, disabled the core module and re-enabled the plugin, all changed settings were transferred back to the plugin settings.

The only thing I found a bit confusing as an existing user is that the settings changed location without any notice -- perhaps the notice that says that the Back In Stock Notifications has been disabled, should also mention the settings changing location?

(2) New sites

In a new site, I installed this WooCommerce branch and enabled Back In Stock Notifications via the UI, but the DB options weren't created.

peterfabian:
> (2) New sites

> In a new site, I installed this WooCommerce branch and enabled Back In Stock Notifications via the UI, but the DB options weren't created.

Thank a lot for testing, @jimjasson !

I believe I fixed this problem in [946eb26](https://github.com/woocommerce/woocommerce/pull/55787/commits/946eb2633ea2c274cb126858c92c3b3a19498055). 

The problem was that the hooks to set up the infrastructure were not hooked if the package wasn't active, so I had to add some way to run some code from the package even if it's disabled. Updated `Packages.php` to run this during the `prepare_packages`.

jimjasson:
Thanks @peterfabian! I confirmed that after the fix, the DB tables and the cron job are added to the site correctly.

plugins/woocommerce/includes/admin/class-wc-bis-admin.php
```diff
@@ -0,0 +1,513 @@
...
+				wp_enqueue_script( 'wc-bis-dashboard' );
+				wp_enqueue_script( 'flot' );
+				wp_enqueue_script( 'flot-resize' );
+				wp_enqueue_script( 'flot-time' );
+			}
+
+			wp_localize_script( 'wc-bis-writepanel', 'wc_bis_admin_params', $params );
+		}
+
+		wp_enqueue_style( 'wc-bis-admin' );
```
coreymckrill: Does this stylesheet need to get enqueued on every admin page?

plugins/woocommerce/includes/admin/class-wc-bis-admin.php
```diff
@@ -0,0 +1,513 @@
...
+			'i18n_dashboard_sent_chart_tooltip'         => __( '%notifications% sent on %date%', 'woocommerce' ),
+		);
+
+		wp_register_script( 'wc-bis-dashboard', WC()->plugin_url() . '/assets/js/admin/wc-bis-admin-dashboard' . $suffix . '.js', array( 'jquery', 'wc-bis-writepanel' ), $version );
+
+		/*
+		 * Enqueue specific styles & scripts.
+		 */
+		if ( WC_BIS()->is_current_screen( array( 'woocommerce_page_wc-settings' ) ) ) {
+			wp_enqueue_script( 'wc-bis-writepanel' );
```
coreymckrill, line 197: Does this script need to get enqueued on every WC Settings page?

jimjasson:
Thank you for the latest changes @peterfabian! 

I re-tested the upgrade + activation flow and confirmed that:
- When I have a new site with WooCommerce with the Back In Stock Notifications disabled and I upgrade WooCommerce to a version with Back In Stock Notifications enabled, then both the DB tables and the Cron job are created. Moreover, the "Allow sign-ups" option is enabled. 
- When I have an existing site with WooCommerce with the Back In Stock Notifications disabled and I upgrade WooCommerce to a version with Back In Stock Notifications enabled, then both the DB tables and the Cron job are created. Moreover, the "Allow sign-ups" option is disabled.

In both scenarios, the site didn't have the Back In Stock Notifications plugin.

jimjasson:
Looking at the code changes @peterfabian, I have the following concerns (from minor to more important);

(1) The `@since` and `@version` numbers in all files are still pointing to the Back In Stock Notifications plugin versions. These should be updated to either `x.x.x` or to the WooCommerce version you are targeting. 

(2) The Back In Stock Notifications file names don't follow core's naming conventions. Compare `class-wc-admin-brands` with `class-wc-bis-admin-ajax` -- I think it should become `class-wc-admin-bis-ajax`. Similar naming changes were made when merging Brands into core. 

(3) The `prepare_always` and `is_really_enabled` functions 

After our sync discussion and re-reading the code, I understand that with the Back In Stock Notifications package there is the need to run some code even when the package is disabled. For example, you want to register the **WooCommerce > Settings > Advanced > Features > Back In Stock Notifications** option, without loading the rest of the Back In Stock Notifications code. 

Therefore, `prepare`, which currently runs only for enabled packages isn't enough: https://github.com/woocommerce/woocommerce/blob/2b51d8248c746091057d1cefe5d7df6d8ab2ddc4/plugins/woocommerce/src/Packages.php#L172 and you would like a function similar to `prepare` that runs for all merged packages. 

Given that `prepare` was first added to the Brands module, my proposal here is to make it run for all merged packages (only Brands and Back In Stock Notifications are merged packages for now), even when they are not enabled. Then, inside each `prepare` distinguish which code will run for when the package is enabled and which code will run always regardless of the status of the package. 

As for `is_really_enabled`, I understand that the need for such a function stems for the fact that `is_enabled` essentially check if there are any restrictions tied to our gradual rollout plan. Looking back at Brands I think `is_enabled` was a poor naming choice and this should change in both packages so that:
- `is_enabled` returns `Packages::is_package_enabled( self::$PACKAGE_NAME );` and;
- a new function should be created to check if the package should be enabled/disabled based on the rollout plan. 

pinging @xristos3490 for his thoughts on this as well.

peterfabian:
Thanks for the comments, @jimjasson !

> (1) The `@since` and `@version` numbers in all files are still pointing to the Back In Stock Notifications plugin versions. These should be updated to either x.x.x or to the WooCommerce version you are targeting.

Agreed, it's still in the TODO list in the PR description:

> TODO:
> - Sort out the feature announcement
> - Update since tags <---- HERE
> - Add tracking
> - The plugin list looks slightly odd if there isn't any update to the BIS plugin (there's an extra border/shadow/sth)

> (2) The Back In Stock Notifications file names don't follow core's naming conventions. Compare class-wc-admin-brands with class-wc-bis-admin-ajax -- I think it should become class-wc-admin-bis-ajax. Similar naming changes were made when merging Brands into core.

I'm a bit split here. Were you not afraid of breaking backward compatibility? Also, to me, it feels more important to retain the convention where file name maps to class name (and vice versa) than to rename wc-bis-admin... to wc-admin-bis...

> (3) The prepare_always and is_really_enabled functions

Let's discuss with others, too. I agree there are multiple options on how to handle these abstractions.

I agree with changing the semantics of `is_enabled` and adding another method to handle the rollout enabled/disabled states.

As for `prepare_always`, I don't have a strong opinion.

peterfabian:
Updated the testing instructions to (hopefully) cover activation scenarios more clearly.

coreymckrill:
Looking at this from the following lens:

> * loading and activation (mostly the files in src subdirectory, especially BackInStockNotifications.php)
> * performance wise, this won't cause problems/slow down their efforts to improve core's performance
> * creating the database makes sense
> * coding standards for WC core have been upheld

**Loading and activation**

The BackInStockNotifications.php file looks fine to me. One question I have is, thinking past the initial rollout, what will need to change here when this is no longer a "feature" that can be enabled/disabled from Settings > Advanced > Features? I don't actually see anything in here that looks like it would be problematic for that transition, but it seems worth thinking about.

**Performance**

This looks pretty well self-contained. I don't see any extra queries running on screens where they shouldn't be. I left a couple of notes about scripts and stylesheets that probably don't need to be loaded everywhere.

**Database**

As above, thinking about the longer-term past the initial rollout, would it make sense to consolidate the database schema into `WC_Install` so that future us can see at a glance all the extra tables that WC generates, rather than having to go hunting for them in multiple files?

**Coding standards**

The linter seems happy. Other than the things already discussed above (`includes` vs `src`, file names), I don't have any concerns here.

peterfabian:
Thanks for your input, @coreymckrill !

> > Database
>
> As above, thinking about the longer-term past the initial rollout, would it make sense to consolidate the database schema into WC_Install so that future us can see at a glance all the extra tables that WC generates, rather than having to go hunting for them in multiple files?

I can definitely do that. I tried to follow the example of HPOS and product lookup tables and only added the tables to the schema if the feature is enabled, but perhaps we want to approach this differently? Given that this will be happening more in the future, should we create some kind of guideline about how we want to approach this? (e.g. also discussed in p7bje6-6UG-p2).

> I left a couple of notes about scripts and stylesheets that probably don't need to be loaded everywhere.

Thanks, I fixed those in [bbd60fb](https://github.com/woocommerce/woocommerce/pull/55787/commits/bbd60fb3d2ed436bd2f61b9e5531ff957044eeaa).

peterfabian:
> As for is_really_enabled, I understand that the need for such a function stems for the fact that is_enabled essentially check if there are any restrictions tied to our gradual rollout plan. Looking back at Brands I think is_enabled was a poor naming choice and this should change in both packages so that:
>
> - is_enabled returns Packages::is_package_enabled( self::$PACKAGE_NAME ); and;
> - a new function should be created to check if the package should be enabled/disabled based on the rollout plan.

Implemented in [87dacbb](https://github.com/woocommerce/woocommerce/pull/55787/commits/87dacbb42587becf75bfb0c15b36dea9bfe2ac12).

I think we should double check if Brands behaves as expected now, too, to be on the safe side.

plugins/woocommerce/includes/bis/compatibility/core/class-wc-bis-core-compatibility.php
```diff
@@ -0,0 +1,191 @@
...
+
+	/**
+	 * Returns true if site is using block theme.
+	 *
+	 * @since  1.6.0
+	 *
+	 * @return boolean
+	 */
+	public static function wc_current_theme_is_fse_theme() {
+		return function_exists( 'wc_current_theme_is_fse_theme' ) ? wc_current_theme_is_fse_theme() : false;
```
xristos3490, line 190: I think this should also be deprecated. After all, it's another Core global function. It's also used [in the Templates controller](https://github.com/woocommerce/woocommerce/pull/55787/files#diff-c32c0e4eb9ef79a55cf6a9c277ab869b5434723090569076431c21bc6fb5511fR74) when we load the frontend assets.

plugins/woocommerce/includes/admin/settings/class-wc-bis-admin-settings.php
```diff
@@ -0,0 +1,348 @@
...
+		 * @internal For exclusive usage of WooCommerce core, backwards compatibility not guaranteed.
+		 */
+		public function add_customer_stock_notifications_settings( array $settings, string $section_id ): array {
+			if ( 'bis_settings' === $section_id ) {
+				$title_item = array(
+					'title' => __( 'Customer stock notifications', 'woocommerce' ),
+					'type'  => 'title',
+				);
+
+				if ( ! BackInStockNotifications::is_enabled() ) {
```
xristos3490: I don't know how we can end up here. Considering this controller won’t function if the feature is disabled, right?

xristos3490 (commented):
Fantastic work so far @peterfabian! 👏  Before looking into the Packages controller, I found a few issues that need addressing. 

The first issue involves the following actions:
- Install the BIS plugin.
- Use this branch.
- With the BIS plugin activated, try to enable the feature from the "Advanced > Features" checkbox.
- This results in an error because the proxy still accesses the plugin version of `WC_BIS_Install`, where the schema method is private. Haven't looked for solutions, but the most obvious one I guess would be to use a reflection here.

The error:
```
[11-Mar-2025 16:56:01 UTC] PHP Fatal error:  Uncaught TypeError: call_user_func_array(): Argument #1 ($callback) must be a valid callback, cannot access private method WC_BIS_Install::get_schema() in /Users/chris/Repos/woocommerce/plugins/woocommerce/src/Proxies/LegacyProxy.php:98
Stack trace:
#0 /Users/chris/Repos/woocommerce/plugins/woocommerce/src/Proxies/LegacyProxy.php(98): call_user_func_array('WC_BIS_Install:...', Array)
#1 /Users/chris/Repos/woocommerce/plugins/woocommerce/src/Internal/BackInStockNotifications.php(327): Automattic\WooCommerce\Proxies\LegacyProxy->call_static('WC_BIS_Install', 'get_schema')
#2 /Users/chris/Repos/woocommerce/plugins/woocommerce/src/Internal/BackInStockNotifications.php(374): Automattic\WooCommerce\Internal\BackInStockNotifications::bis_tables_exist()
#3 /Users/chris/Repos/woocommerce/plugins/woocommerce/src/Internal/BackInStockNotifications.php(451): Automattic\WooCommerce\Internal\BackInStockNotifications::maybe_create_database_tables()
#4 /Users/chris/Repos/woocommerce/plugins/woocommerce/src/Internal/BackInStockNotifications.php(393): Automattic\WooCommerce\Internal\BackInStockNotifications::maybe_update_bis_infrastructure(NULL, 'yes', 'wc_feature_wooc...')
#5 /Users/chris/Sites/woo/wp-includes/class-wp-hook.php(324): Automattic\WooCommerce\Internal\BackInStockNotifications::handle_add_option('wc_feature_wooc...', 'yes')
#6 /Users/chris/Sites/woo/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters('', Array)
#7 /Users/chris/Sites/woo/wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#8 /Users/chris/Sites/woo/wp-includes/option.php(1168): do_action('add_option_wc_f...', 'wc_feature_wooc...', 'yes')
#9 /Users/chris/Sites/woo/wp-includes/option.php(921): add_option('wc_feature_wooc...', 'yes', '', 'on')
#10 /Users/chris/Repos/woocommerce/plugins/woocommerce/includes/admin/class-wc-admin-settings.php(993): update_option('wc_feature_wooc...', 'yes', 'yes')
#11 /Users/chris/Repos/woocommerce/plugins/woocommerce/includes/admin/settings/class-wc-settings-page.php(532): WC_Admin_Settings::save_fields(Array)
#12 /Users/chris/Repos/woocommerce/plugins/woocommerce/includes/admin/settings/class-wc-settings-advanced.php(542): WC_Settings_Page->save_settings_for_current_section()
#13 /Users/chris/Sites/woo/wp-includes/class-wp-hook.php(324): WC_Settings_Advanced->save('')
#14 /Users/chris/Sites/woo/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters('', Array)
#15 /Users/chris/Sites/woo/wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#16 /Users/chris/Repos/woocommerce/plugins/woocommerce/includes/admin/class-wc-admin-settings.php(89): do_action('woocommerce_set...')
#17 /Users/chris/Repos/woocommerce/plugins/woocommerce/includes/admin/class-wc-admin-menus.php(186): WC_Admin_Settings::save()
#18 /Users/chris/Sites/woo/wp-includes/class-wp-hook.php(324): WC_Admin_Menus->save_settings('')
#19 /Users/chris/Sites/woo/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters(NULL, Array)
#20 /Users/chris/Sites/woo/wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#21 /Users/chris/Sites/woo/wp-settings.php(726): do_action('wp_loaded')
#22 /Users/chris/Sites/woo/wp-config.php(102): require_once('/Users/chris/Si...')
#23 /Users/chris/Sites/woo/wp-load.php(50): require_once('/Users/chris/Si...')
#24 /Users/chris/Sites/woo/wp-admin/admin.php(34): require_once('/Users/chris/Si...')
#25 /Users/chris/.composer/vendor/laravel/valet/server.php(110): require('/Users/chris/Si...')
#26 {main}
  thrown in /Users/chris/Repos/woocommerce/plugins/woocommerce/src/Proxies/LegacyProxy.php on line 98

```

The second one is related to the admin menu. See the double highlighted items under "WooCommerce":
Screenshot 2025-03-11 at 7 00 06 PM

This is caused by a few tweaks we made to modify the highlighted item in this jQuery script: https://github.com/woocommerce/woocommerce-back-in-stock-notifications/blob/536b6d72edac3197f3ebd9ec14581caa6e624d5e/resources/js/admin/wc-bis-admin.js#L10-L16 To solve this properly I think we need to:
- Remove the script above.
- Fix the "Notifications" (second tab) submenu to highlight the correct parent item with some fixing in some globals like the `global $parent_item`.

plugins/woocommerce/includes/bis/class-wc-bis-templates.php
```diff
@@ -0,0 +1,136 @@
...
+	 */
+	public function frontend_scripts() {
+
+		$suffix  = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min';
+		$version = Constants::get_constant( 'WC_VERSION' );
+
+		// Styles
+		wp_register_style( 'wc-bis-css', WC()->plugin_url() . '/assets/css/frontend/bis.css', false, $version, 'all' );
+		wp_style_add_data( 'wc-bis-css', 'rtl', 'replace' );
+		wp_enqueue_style( 'wc-bis-css' );
```
xristos3490, line 84: Something else to consider is that we load a single CSS for each frontend view. If a block theme is active, then we also load another one. Honestly, regarding this merge, I think it would be better to: 
- Repurpose the `dequeue_product_scripts` function to also handle styles, 
- Merge the block theme with the main CSS (possibly by prefixing rules with the `woocommerce-uses-block-theme` class).

WDYT? @peterfabian

peterfabian:
> The first issue involves the following actions:
> 
> - Install the BIS plugin.
> - Use this branch.
> - With the BIS plugin activated, try to enable the feature from the "Advanced > Features" checkbox.
> - This results in an error because the proxy still accesses the plugin version of WC_BIS_Install, where the schema method is private. Haven't looked for solutions, but the most obvious one I guess would be to use a reflection here.

Thanks for noticing this. I did some additional tweaks to the activation, namely:
- moved the schema definition to BackInStockNotifications class
- when activating the merged BIS via UI settings, I deactivate the standalone BIS in the same request in `WC_Settings_Advanced::save`, as otherwise I'd need 2 redirects to make the merged BIS fully active (clicking Save settings just sends a POST request, but doesn't refresh the page, so in the first reload the BIS plugin would have been deactivated (at this point merged BIS still can't be loaded), and another reload is needed to activate merged BIS)
- I had to tweak the admin notice about deactivation to survive the redirect, so I'm using transients.

Please see [fba6deb](https://github.com/woocommerce/woocommerce/pull/55787/commits/fba6deba1bacc98917377d3d0c86529f29e6678d) and [39870fb](https://github.com/woocommerce/woocommerce/pull/55787/commits/39870fbd2ebf2c7290fb4d0ccb088b5c1efc461d).

peterfabian:
> The second one is related to the admin menu. See the double highlighted items under "WooCommerce"

Well spotted! To reduce the number of changes in this PR vs standalone BIS, I opted to just [change the class for Activity page](https://github.com/woocommerce/woocommerce/pull/55787/commits/03dc5348566879351048cde85e536c0a641b6488) instead of removing the jQuery part and reimplementing it on the backend (e.g. as [here](https://github.com/Automattic/wc-calypso-bridge/blob/0a99e8f6660f6e485f0ad81282d10b7132c1660f/includes/class-wc-calypso-bridge-ecommerce-admin-menu.php#L273-L334)) as you suggested. Please let me know if you disagree.

plugins/woocommerce/includes/bis/compatibility/core/class-wc-bis-core-compatibility.php
peterfabian: Good point, updated in [bb8ca00](https://github.com/woocommerce/woocommerce/pull/55787/commits/bb8ca005f7e36d67819e78926bebd24607423996).

plugins/woocommerce/includes/admin/settings/class-wc-bis-admin-settings.php
peterfabian: You're right, that's an atavism from when the settings were handled differently. Removed in [3d439dd](https://github.com/woocommerce/woocommerce/pull/55787/commits/3d439ddc32efc74722c6f17fe87d1c2311c6467b).

plugins/woocommerce/includes/bis/class-wc-bis-templates.php
peterfabian: Sounds like a good suggestion. Something like this perhaps? [458101e](https://github.com/woocommerce/woocommerce/pull/55787/commits/458101e0f5baa15d681bbbe15390e010d0629db6)

peterfabian:
> The linter seems happy.

Heh, seems like it was showing up as ✅ only because it had a fatal during the run. Will fix the complaints now that it actually runs.

plugins/woocommerce/client/legacy/css/bis.scss
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Bundles integration styles.
+ */
+.bundle_form #wc_bis_already_registered,
+.bundle_form #wc_bis_product_form {
+  margin-top: 20px;
+}
+
+// Single product fix.
+div.product .wc_bis_product_form {
```
xristos3490: To avoid any visual regressions I would suggest prefixing here with block theme's frontend wrapper. So:
```suggestion
.woocommerce-uses-block-theme div.product .wc_bis_product_form {
```

plugins/woocommerce/includes/bis/class-wc-bis-ajax.php
```diff
@@ -0,0 +1,28 @@
...
+	exit;
+}
+
+/**
+ * Front-end AJAX filters.
+ *
+ * @class    WC_BIS_Ajax
+ * @version  1.0.0
+ */
+class WC_BIS_Ajax {
```
xristos3490: I believe this is noop and can be removed.

plugins/woocommerce/src/Packages.php
```diff
@@ -170,6 +177,13 @@ public static function is_package_enabled( $package ) {
 	 * Especially useful when running actions early in the 'plugins_loaded' timeline.
 	 */
 	public static function prepare_packages() {
+		// Prepare all merged packages for initialization even if they are disabled.
+		foreach ( self::$merged_packages as $package_name => $package_class ) {
+			if ( method_exists( $package_class, 'prepare_always' ) ) {
+				call_user_func( array( $package_class, 'prepare_always' ) );
```
xristos3490: While this can be very opinionated stuff, it seems to me that a more appropriate name would be `setup` here.

plugins/woocommerce/includes/class-woocommerce.php
```diff
@@ -1188,6 +1190,24 @@ public function mailer() {
 		return WC_Emails::instance();
 	}
 
+	/**
+	 *
+	 *
+	 * @return WC_Back_In_Stock|WC_BIS_Noop
+	 */
+	public function bis_notifications() {
```
xristos3490: Where is this used? 🤔

plugins/woocommerce/src/Internal/BackInStockNotifications.php
```diff
@@ -0,0 +1,516 @@
...
+			// When WC_BIS() is present before loading BIS from core, it will fatal during init(), 
+			// so init() needs to be skipped. This should only be triggered once after 
+			// a plugin update during the request when BIS plugin is deactivated.
+			self::$bis_plugin_is_active = true;
+		}		
+
+		// Check for CLI activation via WP-CLI
+		if ( defined('WP_CLI') && WP_CLI ) {
+			global $argv;
+			if ( is_array( $argv ) && in_array( 'plugin', $argv ) && in_array( 'activate', $argv ) && in_array( $bis_plugin_name, $argv ) ) {
```
xristos3490: If I want to activate bis via CLI, I'm gonna probably write this:
```
wp plugin activate woocommerce-back-in-stock-notifications
```
Which doesn't include the PHP filename thus it falls to: "Cannot redeclare WC_BIS() etc".

xristos3490 (commented):
Shared some more findings/questions while I am wrapping up manual testing.

plugins/woocommerce/includes/bis/class-wc-bis-templates.php
xristos3490: Looking good! Left a comment here https://github.com/woocommerce/woocommerce/pull/55787/commits/458101e0f5baa15d681bbbe15390e010d0629db6#r1995311931

plugins/woocommerce/includes/class-woocommerce.php
peterfabian: Nowhere at the moment. I was thinking whether to include this so that devs would get the BIS using the `WC()-> bis_notifications()` method instead of `WC_BIS()`. Similar to `queue`, `shipping`, `mailer`. What do you see as preferable?

xristos3490: Interesting! My two cents are that the `WC_BIS()` as a standalone namespace is preferable for now. I feel we should only integrate well-abstracted utilities/traits into that global function or have a strong reason for doing so?

plugins/woocommerce/client/legacy/css/bis.scss
xristos3490: Same for the second rule of the old `blocktheme.css`

dakota:
I mostly spent my time testing various activation/rollout mechanisms and for the most part it works really well. There was one issue with using the CLI where it failed. @xristos3490 pointed out the reason here: https://github.com/woocommerce/woocommerce/pull/55787#discussion_r1995458826

One small note, the method to change to enable/disable is `is_enabled_for_rollout`, not `is_enabled`

- ✅ Test Rollout Disabled to Rollout Enabled Store with BIS plugin
- ✅ Test UI option in WC > Settings > Advnaced > Features overrides Rollout Disabled
- ✅ Test UI option in WC > Settings > Advnaced > Features overrides Rollout Enabled
- ✅ Test Rollout Enabled State without BIS plugin---new site
- ✅ Test Rollout Enabled State without BIS plugin---existing site
- ⛔ Test Rollout Enabled State without BIS plugin: try to force activate BIS plugin via WP-CLI
   When running `wp plugin activate woocommerce-back-in-stock-notifications` I get 
   ```
    PHP Fatal error:  Cannot redeclare WC_BIS() (previously declared in /Users/waltherlalk/Studio/bis-test/wp-content/plugins/woocommerce/includes/bis/class-wc-back-in-stock.php:371) in /Users/waltherlalk/Studio/bis-test/wp-content/plugins/woocommerce-back-in-stock-notifications/woocommerce-back-in-stock-notifications.php on line 486
   ```

- ✅ Test Rollout Enabled State with BIS plugin: verify data and settings remain from the BIS plugin (sign up for notifications and send some out)
- ✅ Test Rollout Enabled State with BIS plugin: it's possible to roll back to WC version without BIS and use standalone BIS.
- ✅ Test Rollout Enabled State with BIS plugin: check that BIS settings have moved from WC > Settings > Stock Notifications to WC > Settings > Products > Customer stock notifications
- ✅ Check The Included BIS Works As Expected

Overall, well done, looking good!

plugins/woocommerce/src/Internal/BackInStockNotifications.php
peterfabian: Well spotted, I haven't tested this so far, so thanks for noticing!

plugins/woocommerce/client/legacy/css/bis.scss
peterfabian: Makes sense, done in 5a6a6000402616fff3064b854340ea561462dfa2.

plugins/woocommerce/includes/bis/class-wc-bis-ajax.php
peterfabian: Correct, removed.

plugins/woocommerce/src/Packages.php
peterfabian: I like the suggestion, renamed in [acbcded](https://github.com/woocommerce/woocommerce/pull/55787/commits/acbcdedcf5e17054fdb201a08e0bb5f1db6221be).

plugins/woocommerce/includes/class-woocommerce.php
peterfabian: You're probably right that it's unnecessary. Removed in [69ebea3](https://github.com/woocommerce/woocommerce/pull/55787/commits/69ebea3edcfdc59b467098db7144b94df04e141e).

plugins/woocommerce/src/Internal/BackInStockNotifications.php
peterfabian: Updated.

peterfabian:
As it was decided to reimplement this feature in a different way (integrate more deeply with core and not include everything in the legacy `includes` folder), this PR is no longer relevant.

PR #55044, opened 2025-01-31
Code Snippets Overhaul: Free Shipping Customizations

jessepearson:
### Submission Review Guidelines:

-   I have followed the [WooCommerce Contributing Guidelines](https://github.com/woocommerce/woocommerce/blob/trunk/.github/CONTRIBUTING.md) and the [WordPress Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/).
-   I have checked to ensure there aren't other open [Pull Requests](https://github.com/woocommerce/woocommerce/pulls) for the same update/change.
-   I have reviewed my code for [security best practices](https://developer.wordpress.org/apis/security/).
-   Following the above guidelines will result in quick merges and clear and detailed feedback when appropriate.

### Changes proposed in this Pull Request:

Closes #55040 .

* Updates to the _Free Shipping Customization_ doc
* Added a prefix of `fsc_` to each of the functions being added.
* Refactored functions to make them easier to read and modify.
* Added notes into the functions to explain what they do.
* Updated descriptive text of the functions to try to be more clear.
* Removed the function/snippet that excluded states, as that can be handled with shipping zones.
* Moved the `is_available` snippet to the bottom, as I believe that will not be used as much as the others.
* Provided example usage of `is_available`

### How to test the changes in this Pull Request:

These changes do not affect WooCommerce directly, but they do provide additional functionality through snippets.

1. Be sure you have a product that is shippable set up.
2. Be sure you have a shipping zone set up.
3. The shipping zone should have Free Shipping set up that gets triggered at a certain amount, like $30.
4. The shipping zone should have another method set up, like Flat Rate for a fee.
5. You will also need to have Local Pickup (WooCommerce > Settings > Shipping > Local Pickup) set up for use.
6. Apply each snippet to see if it functions as intended. Using the Code Snippets plugin is likely the easiest way to add/edit them.

### Changelog entry

-   [ ] Automatically create a changelog entry from the details below.

-   [x] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Details</summary>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

-   [ ] Fix - Fixes an existing bug
-   [ ] Add - Adds functionality
-   [ ] Update - Update existing functionality
-   [ ] Dev - Development related task
-   [ ] Tweak - A minor adjustment to the codebase
-   [ ] Performance - Address performance issues
-   [ ] Enhancement - Improvement to existing functionality

[...]

PR #55796, opened 2025-02-24
Implement hide shipping rates when free is available option

### LINKED ISSUE #52872: [Enhancement]: define both minimum and maximum order amount for shipping

A common scenario in most ecommerce stores is to offer free shipping above a certain minimum order total.
For example, standard shipping costs $5, but it is free for orders above $40.
In order to implement this super common scenario, I would set a free shipping with the minimum order amount of $40 and a paid shipping option.
Unfortunately, for orders above $40 the shopper qualifies for free shipping, but the paid shipping is shown as well, which makes no sense because it is the same shipping option with different price.

My solution is to add a maximum order amount above which the shipping is hidden.
So paid shipping would be from $0 to $39.99 and free shipping would be from $40.

This is how I added the textfield with a php snippet.
Image

And this is the code.

```

/**
 * Add the maximum order amount field to shipping methods
 */
function add_maximum_amount_to_shipping_methods() {
    // Get all shipping methods
    $shipping_methods = WC()->shipping()->get_shipping_methods();
    
    // Loop through each shipping method
    foreach ($shipping_methods as $shipping_method) {
        // Add filter for each shipping method
        add_filter('woocommerce_shipping_instance_form_fields_' . $shipping_method->id, 
            'add_shipping_maximum_order_amount');
    }
}
add_action('woocommerce_init', 'add_maximum_amount_to_shipping_methods');
 
/**
 * Add maximum order amount field to shipping method settings
 */
function add_shipping_maximum_order_amount($settings) {
    // Add our maximum order amount field with detailed explanation
    $settings['max_amount'] = array(
        'title'             => __('Maximum Order Amount', 'woocommerce'),
        'type'              => 'number',
        'description'       => __('When a customer\'s cart total exceeds this amount, this shipping method will be hidden from the checkout options. Set to 0 to have no maximum limit.<br><br><strong>Example:</strong> If you set this to 1000, this shipping method will only be available for orders under 1000. This is useful when you want to restrict certain shipping methods for large orders.<br>', 'woocommerce'),
        'default'           => '0',
        'desc_tip'          => false,
        'custom_attributes' => array(
            'min'  => '0',
            'step' => '0.01'
        )
    );
    
    return $settings;
}
 
/**
 * Hide shipping methods if cart total exceeds maximum amount
 */
function hide_shipping_when_cart_exceeds_maximum($rates, $package) {
    if (empty($rates)) {
        return $rates;
    }
 
    $cart_total = WC()->cart->get_subtotal();
    
    foreach ($rates as $rate_id => $rate) {
        // Get the instance ID from the rate ID
        $instance_id = substr($rate->id, strrpos($rate->id, ':') + 1);
        $method_id = substr($rate->id, 0, strrpos($rate->id, ':'));
        
        // Get the settings for this specific shipping method instance
        $settings = get_option('woocommerce_' . $method_id . '_' . $instance_id . '_settings');
        
        if (isset($settings['max_amount']) && 
            !empty($settings['max_amount']) && 
            $settings['max_amount'] > 0 && 
            $cart_total > (float)$settings['max_amount']) {
            unset($rates[$rate_id]);
        }
    }
    
    return $rates;
}
add_filter('woocommerce_package_rates', 'hide_shipping_when_cart_exceeds_maximum', 10, 2);
 
/**
 * Add maximum amount to shipping method title in admin
 */
function add_maximum_amount_to_admin_shipping_title($title, $method) {
    if (!is_admin()) {
        return $title;
    }
 
    $settings = get_option('woocommerce_' . $method->id . '_' . $method->instance_id . '_settings');
    
    if (isset($settings['max_amount']) && !empty($settings['max_amount']) && $settings['max_amount'] > 0) {
        $title .= ' ' . sprintf(
            __('(Maximum order: %s)', 'woocommerce'),
            wc_price($settings['max_amount'])
        );
    }
    
    return $title;
}
add_filter('woocommerce_shipping_instance_name', 'add_maximum_amount_to_admin_shipping_title', 10, 2);
```

Another possible option is to add a drop-down menu in each shipping option that allows to hide the selected options if the edited one is shown.

### Additional context

_No response_

### LINKED ISSUE #54370: [Enhancement]: Allow both a minimum and a maximum threshold for shipping methods

On my store, customers qualify for free shipping if their cart total is above $60, otherwise they pay $5 for shipping.
I can set the minimum threshold, but if they reach the $60 there are two shipping methods, the paid and the free one, which makes no sense, since those are the same shipping method.
A fix could be done by allowing merchants to set both a minimum threshold and a maximum threshold.

For example: free shipping for orders above $60.
Shipping method 1: flat rate, minimum threshold 0, maximum threshold $59.99, shipping price: $5.
Shipping method 2: free shipping, minimum threshold $60, maximum threshold ∞, shipping price: $0.

_No response_

### Additional context

Request by @tomxygen on https://woocommerce.com/feature-request/fix-the-free-shipping-for-order-above-x/

ismaeldcom:

Closes #54370
Closes #52872 

- Add `Hide shipping rates when free is available` option in **WooCommerce → Settings → Shipping → Shipping settings**
- Implement the logic in `WC_Shipping->calculate_shipping_for_package`

After reviewing the two issues and related conversations, I've tried to find a good compromise between what's required and what we can offer in a simple way without overlapping with existing plugins.

I've decided on a new option in the shipping settings to hide any shipping rates when free shipping is available. This covers the two most used cases without any complex configuration, such as the maximum order amount mentioned in the questions, and leaves more advanced use cases to existing plugins.

| Before | After |
|--------|--------|
| Screenshot 2025-02-24 at 12 59 02|Screenshot 2025-03-03 at 11 48 29| 
|Screenshot 2025-02-24 at 12 57 30|Screenshot 2025-02-24 at 12 57 51|

1. Checkout the PR branch.
2. Ensure the store has at least a flat rate, local pickup, and free shipping configured and enabled.
3. Add a product that meets the condition for all the shipping rates to the cart.
4. Go to the cart
5. All the shipping methods should be available
6. Go to **WooCommerce → Settings → Shipping → Shipping settings** and enable `Hide shipping rates when free is available`.
7. Go to the cart
8. Only free shipping and local pickup should be available.

### Post-release
- [ ] Update WooCommerce documentation

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Implement hide shipping rates when free is available option

</details>

<details>

#### Comment 

</details>


ismaeldcom:
Hey @kallehauge, I'd appreciate an extra pair of eyes here, knowing that this is a new feature and not just a fix 🙂

jessepearson:
I like the work that's been done here, but I believe this is going to turn into a problem if we only disable the Flat Rate shipping method and not other methods. The snippet and plugin mentioned in #54370 hide all other methods when Free Shipping is available, do you think that would be a possibility?

The snippet mentioned has better versions here: https://github.com/woocommerce/woocommerce/pull/55044

To elaborate on the above, if we only disable Flat Rate, then support will get requests about not being able to disable other methods and merchants may find this setting not useful to them. If we allow them to disable all methods, then it's like they added the snippet by only checking a checkbox.

I think the downside would be if a merchant offers a shipping method that is faster than the free shipping, however, it would be their choice to check the box to enable this or not.

CC @juliaamosova @pierorocca

kloon:
I tend to agree with @jessepearson here. I'd suggest changing this to an option to hide all other shipping methods instead.

ismaeldcom:
Thanks for the comments!

I thought about this for a long time and decided to only hide flat rate shipping to cover the case where they also offer local pickup, which is the only other available type in Core.

Maybe I'm missing something here, but are there any other types of shipping that could be visible? 🤔 

Anyway, if you think that it will be better to just hide any other shipping method, I'll update it 🙂

pierorocca:
It's definitely complicated. Mass hiding doesn't make sense because a shopper may have a preferred carrier or preferred service level. Even if there's a free shipping option, it might not be fast enough service or with a carrier they trust or prefer. There are also forwarding services e.g. forward to FedEx retail location that aren't local pickup that we wouldn't want to hide.

I think the question to answer is, is Flat Rate always the method that replaces free shipping? Why can't a weight based or size based shipping also be replaced? I may be wrong, the safe bet here is conditional logic and AFAIK the WooCommerce Conditional Shipping and Payments plugin is best suited to handling this complexity. 

I really like the new mix/max thresholds but I'm not convinced the hide flat rate setting won't create more problems than it solves. @laurajohnsonwoo could you lend your expertise please?

jessepearson:
> It's definitely complicated. Mass hiding doesn't make sense because a shopper may have a preferred carrier or preferred service level. Even if there's a free shipping option, it might not be fast enough service or with a carrier they trust or prefer. There are also forwarding services e.g. forward to FedEx retail location that aren't local pickup that we wouldn't want to hide.
> 
> I think the question to answer is, is Flat Rate always the method that replaces free shipping? Why can't a weight based or size based shipping also be replaced? I may be wrong, the safe bet here is conditional logic and AFAIK the WooCommerce Conditional Shipping and Payments plugin is best suited to handling this complexity.

I had these thoughts, as well. We've provided snippets (linked above) to be able to hide all other shipping methods if free is available. This is a fairly common request from store owners, enough to where we have created documentation for the snippet.

I also thought about, what if we allowed the merchant to select what other methods get hidden when free is available, but that could turn into a nightmare trying to support in the future.

If we provide a simple checkbox to allow merchants to hide all other methods if Free Shipping is available, this just allows us to remove the snippet from our docs and direct them to a setting instead. It can reduce certain support requests where store owners are asking for help with the snippet, and just making that option easier.

For any store owners that want something more advanced, like hiding basic rates if Free is available, but also keep express options, they can use a plugin to do so.

pierorocca:
> I also thought about, what if we allowed the merchant to select what other methods get hidden when free is available, but that could turn into a nightmare trying to support in the future.

This is best served by the conditional shipping and payment plugin IMO since only the merchant would have knowledge of all of the shipping options they've enabled.

I think what you're advocating for here @jessepearson is this one use case where the shipping options are simplistic enough, disabling all other shipping methods, excluding local pickup, meets a certain segment's demand?  For all other cases, the above named plugin is recommended?

kloon:
What about adding a setting to all shipping methods where you can toggle a "Hide me when the following methods are shown" and then a have multi-select where you can select any of the available shipping methods?

ismaeldcom:
I think we should aim to keep it simple, hiding every other shipping method or the max threshold can be a good core solution. Anything beyond that needs to take into account custom cases, and that fits better with code snippets or plugins.

ismaeldcom:
I just updated the code to hide any shipping rate (except local pickup) when a free one is available.

plugins/woocommerce/includes/class-wc-shipping.php
```diff
@@ -358,6 +358,27 @@ public function calculate_shipping_for_package( $package = array(), $package_key
...
+				$free_shipping = array();
+				$local_pickup  = array();
+
+				foreach ( $package['rates'] as $rate ) {
+					if ( 'free_shipping' === $rate->method_id ) {
+						$free_shipping[] = $rate;
+						continue;
+					}
+
+					if ( 'local_pickup' === $rate->method_id ) {
```
jessepearson: I believe that `local_pickup` is the legacy option that is only available if a store was using it, the new version is named `pickup_location`. I'm wondering if we should include both for backwards compatibility.

This works:

```
					if ( in_array( $rate->method_id, array( 'local_pickup', 'pickup_location' ) ) ) {
						$local_pickup[] = $rate;
					}
```

The way I tested with the legacy method was to find [the `is_legacy_local_pickup_active` method](https://github.com/woocommerce/woocommerce/blob/8f39c7d6827622e7434e4820712ba39beb866dc2/plugins/woocommerce/src/Blocks/Shipping/ShippingController.php#L551) and set it to return `true`. I was then able to add the method to a shipping zone.

jessepearson (changes requested):

plugins/woocommerce/includes/class-wc-shipping.php
ismaeldcom: Thanks for mentioning this! My local env isn't working well because I only see the legacy way of adding it, the new one shows an empty page. But after setting up a new env site, I've noticed what you're referring to.

Addressed in 99e31c3.

jessepearson (approved):
Thank you for the updates and working on this. I tested with both the legacy and block methods for local pickup and they show along with Free Shipping once it's enabled.

Shipping zones:
Screenshot 2025-03-04 at 9 25 40 AM

New setting box unchecked:
Screenshot 2025-03-04 at 9 25 59 AM

Cart with all methods showing:
Screenshot 2025-03-04 at 9 26 16 AM

New settings box checked:
Screenshot 2025-03-04 at 9 25 50 AM

Cart only showing free and local options:
Screenshot 2025-03-04 at 9 25 20 AM

ResponsiveFlow:
> think we should aim to keep it simple, hiding every other shipping method or the max threshold can be a good core solution.

I agree. However, what do other platforms do?

ismaeldcom:
>I agree. However, what do other platforms do?

Each platform handles this differently. Our goal is to address the most common customer support requests with a simple core solution while reducing the need for custom snippets in our documentation. For more advanced setups, existing plugins remain the best option.

We're always looking to improve WooCommerce, so if this solution doesn’t fully meet your needs, feel free to submit a new enhancement request 🙂

plugins/woocommerce/includes/class-wc-shipping.php
```diff
@@ -358,6 +358,27 @@ public function calculate_shipping_for_package( $package = array(), $package_key
...
+				$free_shipping = array();
+				$local_pickup  = array();
+
+				foreach ( $package['rates'] as $rate ) {
+					if ( 'free_shipping' === $rate->method_id ) {
+						$free_shipping[] = $rate;
+						continue;
+					}
+
+					if ( in_array( $rate->method_id, array( 'local_pickup', 'pickup_location' ), true ) ) {
```
mikejolley, line 372: There are other locations in the codebase where the list of pickup methods is filterable. `woocommerce_local_pickup_methods` Should that be used here? @ismaeldcom

senadir: Also not just that, any custom shipping method that [declares itself as `pickup` compatible](https://github.com/woocommerce/woocommerce/blob/40e8ca877af1c7f3ef371102a07793a478330072/plugins/woocommerce/src/Blocks/Shipping/PickupLocation.php#L43) will be filtred out here, should ideally check the method `supports` for `'local-pickup'`, this should pickup `pickup_location` by default and any third party one. At least, we use `supports` for UI purposes as well.

ismaeldcom: Thanks for the additional context!

I think using the `woocommerce_local_pickup_methods` filter here makes sense. Since we're filtering the rates, checking for the shipping method `supports` will add an extra overhead that's already handled in the previous block of code. And it will not cover the legacy one.

WDYT?

senadir: I don't think there's much overhead as the method is already loaded in memory at that point, otherwise, you wouldn't get the shipping rates. I think it can be combined with `array_includes( $method->supports, 'local-pickup' ) || 'local_pickup' === $method->id`

ismaeldcom: Follow-up in #56357

ismaeldcom: Resolved in #56357. Thanks for your support @senadir

ant-topps:
Some feedback or comments on this.  While nice in concept, and hey, why wouldn't every store roll this out??? Free shipping can be costly for a business.  To this end, we've specifically asked our developer to keep other (faster) paid options selectable. Obviously, Local Pickup is preferable to Free Shipping as well.

pierorocca:
You make a good point that just because there's a free option doesn't mean there aren't other options available with a faster service or a more reliable shipper. I don't think hiding based only on the existence of free is the right option here. It should be up to the shipping provider to define which methods are available IMO.

ismaeldcom:
That's a good point! With this solution, we aimed to simplify the most common cases we've encountered in our support requests by providing a solution that doesn't require custom code snippets or external plugins. There are many use cases that necessitate a more customizable solution, which we can evaluate in the future.

pierorocca:
What's the possibility that it creates new support tickets for shipping options that are no longer appearing and impacting merchant cost? I.e. do the changes only hide the paid option that is 100% equivalent to the free option?


Free - 2 days
$4.99 - 2 days
$7.99 -  Overnight


Would overnight still appear?

senadir:
@pierorocca I don't think we can always hit 100% of use cases in our features, for plenty of stores, free shipping is enough, and the important part is that **this is an opt-in option**, so merchants can choose to enable this if they want. This is a popular option, [a plugin that enables that](https://wordpress.org/plugins/wc-hide-shipping-methods/) has 30k install, plenty of others use a snippet, and we even have [a support page for that.](https://woocommerce.com/document/hide-other-shipping-methods-when-free-shipping-is-available/)
If a single shipping method wants to show or hide things, it can be done at that level. Merchants who want detailed control to the smallest bit can use [Conditional Shipping and Payment ](https://woocommerce.com/products/conditional-shipping-and-payments/) which allows you to hide/show on very specific conditons.

pierorocca:
Thanks for that clarification @senadir. The opt-in option satisfies my due diligence.

PR #55804, opened 2025-02-24
Refactor `show_shipping` usage by merging blocks and legacy cart functionality

### LINKED ISSUE #54894: Review and remove use of `cart->show_shipping` in Store API
While reviewing https://github.com/woocommerce/woocommerce/pull/53466 we found that [`show_shipping`](https://github.com/woocommerce/woocommerce/blob/a9600503d8b85ca7b6596b3439996446c4a6241d/plugins/woocommerce/includes/class-wc-cart.php#L1586) is used to determine whether the Cart has a full address.

This method checks if shipping is enabled and there are items in the cart.

If so, it checks if the `Hide shipping costs until an address is entered` option is enabled, and if it is, checks the country, state, and postcode of the address. These checks also have an associated filters `woocommerce_shipping_calculator_enable_state` etc. to enable/disable the check _and_ disable the field on the shipping calculator.

Since the Cart/Checkout block use locale to determine which fields to show/require these filters will have no effect, but may affect the outcome of the `show_shipping` check.  By looking at its use, it seems this method should be used for UI showing/hiding aspects of the Cart/Checkout templates not working out whether rates should be shown/hidden.

In https://github.com/woocommerce/woocommerce/pull/53466 a new method was added to `ShippingController` called [`has_full_shipping_address`](https://github.com/woocommerce/woocommerce/pull/53466/files#diff-50b1c15c8d28eb9d250957175d13e1169a942f41f322cebc1b1130b7824f0a9dR421) which is a more thorough check of the address field and it also uses schema, so in line with what the calculator and address forms are using.

The goal of this ticket is:

- Investigate what `show_shipping` is being used for and identify if it's still needed following https://github.com/woocommerce/woocommerce/pull/53466. Specifically, check if `$has_calculated_shipping` in the Cart Schema is something we need to keep around. As it is used in a store API response we would need to release a new version of the API without this, or just figure out a way to deprecate this gracefully without requiring a new version, e.g. by always returning true.
- If the logic to check a full address is still needed as the point show_shipping was used before #53466, see if `show_shipping` can be replaced with `has_full_shipping_address`

mikejolley:

This PR consolidates shipping address handling between blocks and the legacy cart shortcode. `needs_shipping()` and `show_shipping()` are refactored to use the `WC_Customer->has_full_shipping_address()` class method to see if a full shipping address has been provided.

`WC_Cart_Totals` no longer checks `show_shipping`, instead opting to just grab whatever has been calculated in the cart.

Finally, the `WC_Cart` class tracks if calculations have ran or not, so we no longer need to derive this from the Store API.

Closes #54894

Existing tests were updated to match new logic e.g. to ensure shipping methods exist before tests run. The ShippingController tests were updated to use the method in `WC_Customer` (which was moved and updated).

One thing that might be an issue, tests previously passed without a defined `city` because the `show_shipping` method only checked for country, state, postcode. This was however inconsistent with the shipping form which also requested `city`. With the changes, `city` is now required too. Do you see an issue with this @opr? We could make city optional in the `has_full_shipping_address` method if so. 

1. Setup shipping and shipping zones, and enable "Hide shipping costs until an address is entered"
2. In an incognito session go through the full flow, testing shipping calculator and rate selection. 
3. Note that shipping methods and totals are hidden until the calculator is used. 
4. Test this on both legacy shortcode cart and block based cart.
5. Turn off "Hide shipping costs until an address is entered". Repeat the above tests, but notice shipping is shown without a full address. You can check this again with incognito session, and a "Default customer location" under general settings set to store address.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/includes/class-wc-checkout.php
+++ b/plugins/woocommerce/includes/class-wc-checkout.php
@@ -957,7 +957,7 @@ class WC_Checkout {
 			} elseif ( ! in_array( $shipping_country, array_keys( WC()->countries->get_shipping_countries() ), true ) ) {
 				if ( WC()->countries->country_exists( $shipping_country ) ) {
 					/* translators: %s: shipping location (prefix e.g. 'to' + ISO 3166-1 alpha-2 country code) */
-					$errors->add( 'shipping', sprintf( __( 'Unfortunately <strong>we do not ship %s</strong>. Please enter an alternative shipping address.', 'woocommerce' ), WC()->countries->shipping_to_prefix() . ' ' . $shipping_country ) );
+					$errors->add( 'shipping', sprintf( __( 'Unfortunately <strong>we do not ship %s</strong>. Please enter an alternative shipping address.', 'woocommerce' ), WC()->countries->shipping_to_prefix( $shipping_country ) . ' ' . $shipping_country ) );
 				}
 			} else {
 				$chosen_shipping_methods = WC()->session->get( 'chosen_shipping_methods' );
--- a/plugins/woocommerce/includes/class-wc-customer.php
+++ b/plugins/woocommerce/includes/class-wc-customer.php
@@ -8,7 +8,7 @@
 
 defined( 'ABSPATH' ) || exit;
 
-require_once dirname( __FILE__ ) . '/legacy/class-wc-legacy-customer.php';
+require_once __DIR__ . '/legacy/class-wc-legacy-customer.php';
 
 /**
  * Customer class.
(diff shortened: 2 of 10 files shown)


plugins/woocommerce/includes/class-wc-customer.php
```diff
@@ -268,10 +268,88 @@ public function has_shipping_address() {
...
 			}
 		}
-
 		return false;
 	}
 
+	/**
+	 * Checks whether the address is "full" in the sense that it contains all required fields to calculate shipping rates.
+	 *
+	 * @since 9.8.0
```
opr, line 277: ~should be 9.9.0 I think we missed feature freeze for 9.8~ I'm wrong, ignore!

plugins/woocommerce/includes/class-wc-checkout.php
```diff
@@ -957,7 +957,7 @@ protected function validate_checkout( &$data, &$errors ) {
 			} elseif ( ! in_array( $shipping_country, array_keys( WC()->countries->get_shipping_countries() ), true ) ) {
 				if ( WC()->countries->country_exists( $shipping_country ) ) {
 					/* translators: %s: shipping location (prefix e.g. 'to' + ISO 3166-1 alpha-2 country code) */
-					$errors->add( 'shipping', sprintf( __( 'Unfortunately <strong>we do not ship %s</strong>. Please enter an alternative shipping address.', 'woocommerce' ), WC()->countries->shipping_to_prefix() . ' ' . $shipping_country ) );
+					$errors->add( 'shipping', sprintf( __( 'Unfortunately <strong>we do not ship %s</strong>. Please enter an alternative shipping address.', 'woocommerce' ), WC()->countries->shipping_to_prefix( $shipping_country ) . ' ' . $shipping_country ) );
```
opr, line 960: Good find 👌🏼

opr (commented):
> Existing tests were updated to match new logic e.g. to ensure shipping methods exist before tests run. The ShippingController tests were updated to use the method in WC_Customer (which was moved and updated).
> 
> One thing that might be an issue, tests previously passed without a defined city because the show_shipping method only checked for country, state, postcode. This was however inconsistent with the shipping form which also requested city. With the changes, city is now required too. Do you see an issue with this @opr? We could make city optional in the has_full_shipping_address method if so.

The only issue I see is that it is a change in behaviour from previous releases, so if people have E2E tests that skip city, they will now break. Given city isn't used for any shipping zone calculations in core, we can probably make it optional. If carriers _need_ city, then I expect them to use the hook to make it required.

Should we make the values of `woocommerce_shipping_calculator_enable_city` etc. remove the fields from the block-based shipping calculator too? Seems like they don't affect it right now.

alexflorisca (changes requested):
Thanks Mike, code looks good, couldn't spot any issue there. I'm with Thomas that we should keep the function backwards compatible with the `show_shipping` option by making the city optional. 

 The `show_shipping()` is also still used [here](https://github.com/woocommerce/woocommerce/blob/refactor/show-shipping-usage-54894/plugins/woocommerce/templates/cart/cart-totals.php#L41)] - and maybe other places. Can we check any other uses of it and change these to use the new `has_full_shipping_address()` function? Some docs might also need to be updated here.

I also noticed that only the shortcode version of the shipping. calculator respects the `woocommerce_shipping_calculator_enable_city` filter. Might be worth doing that as a separate PR though, just thought it relevant to mention here . Otherwise I'll create an issue.

Thanks!

mikejolley:
> The show_shipping() is also still used [here](https://github.com/woocommerce/woocommerce/blob/refactor/show-shipping-usage-54894/plugins/woocommerce/templates/cart/cart-totals.php#L41)] - and maybe other places. Can we check any other uses of it and change these to use the new has_full_shipping_address() function? Some docs might also need to be updated here.

@alexflorisca no, its still valid to use it for display purposes. But it shouldn't be used for shared calculation classes like WC_Cart_Totals which was fixed in this PR.

> I'm with Thomas that we should keep the function backwards compatible with the show_shipping option by making the city optional.

To my understanding this was a bug? `show_shipping` was deeply coupled to the shortcode shipping calculator, but shipping calculator did show a city field. Both should align. I don't think this is a backwards compatibility issue. As @opr said, it may impact someones own tests but I don't see it as something that will affect customers who need to fill the city field in anyway. 

~~I would be in favour of making city optional/hidden by default however. It doesn't seem necessary to have it enabled seeing as zones don't use city.~~

Its not used for zones but is used for taxes, see https://github.com/woocommerce/woocommerce/pull/18076

How do you want to proceed @opr @alexflorisca ?

opr:
> How do you want to proceed @opr @alexflorisca ?

Is there a way to check if any taxes rely on specific city? That seems like we're getting a little too deep into checking stuff for this field so might not be worth doing.

I think we can continue as we were planning to in this PR and make it required then, but probably going to need a dev advisory noting that.

mikejolley:
> Is there a way to check if any taxes rely on specific city? That seems like we're getting a little too deep into checking stuff for this field so might not be worth doing.

@opr Possible but given its required in core right now I don't think we should.

R/E the tests, the only changes were to the tests specific to `show_shipping` method, so whilst there is a slight change in behaviour in that (checking city is set), actual forms such as shipping calculator, checkout, already require city to be set (its a required field). For that reason I'm not sure a dev advisory is needed because any e2e should be setting data via those forms.

opr (approved):

alexflorisca (approved):
Ok, just caught up on this thread, I'm convinced

PR #55806, opened 2025-02-24
Improve the parsing logic on the redirection controller

bor0:

We've seen a bunch of PHP warnings on WCCOM. This PR improves the parsing logic on the redirection controller.

1. Go to http://localhost:8080/wp-admin/admin.php?page=wc-admin&path=home
2. You should not see any notices in the error log

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/src/Admin/Features/ProductBlockEditor/RedirectionController.php
+++ b/plugins/woocommerce/src/Admin/Features/ProductBlockEditor/RedirectionController.php
@@ -155,6 +155,13 @@ class RedirectionController {
 		$path        = esc_url_raw( wp_unslash( $_GET['path'] ) );
 		$path_pieces = explode( '/', wp_parse_url( $path, PHP_URL_PATH ) );
 
+		if ( 2 !== count( $path_pieces ) ) {
+			return array(
+				'page'       => null,
+				'product_id' => null,
+			);
+		}
+
 		return array(
 			'page'       => $path_pieces[1],
 			'product_id' => 'product' === $path_pieces[1] ? absint( $path_pieces[2] ) : null,
--- a/plugins/woocommerce/src/Admin/Features/ProductBlockEditor/RedirectionController.php
+++ b/plugins/woocommerce/src/Admin/Features/ProductBlockEditor/RedirectionController.php
@@ -155,16 +155,9 @@ class RedirectionController {
 		$path        = esc_url_raw( wp_unslash( $_GET['path'] ) );
 		$path_pieces = explode( '/', wp_parse_url( $path, PHP_URL_PATH ) );
 
-		if ( 3 !== count( $path_pieces ) ) {
-			return array(
-				'page'       => null,
-				'product_id' => null,
-			);
-		}
-
 		return array(
-			'page'       => $path_pieces[1],
-			'product_id' => 'product' === $path_pieces[1] ? absint( $path_pieces[2] ) : null,
+			'page'       => $path_pieces[1] ?? '',
+			'product_id' => 'product' === ( $path_pieces[1] ?? '' ) ? absint( $path_pieces[2] ?? 0 ) : null,
 		);
 	}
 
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/src/Admin/Features/ProductBlockEditor/RedirectionController.php
```diff
@@ -155,6 +155,13 @@ protected function get_parsed_route(): array {
 		$path        = esc_url_raw( wp_unslash( $_GET['path'] ) );
 		$path_pieces = explode( '/', wp_parse_url( $path, PHP_URL_PATH ) );
 
+		if ( 2 !== count( $path_pieces ) ) {
```
mcliwanow: @bor0 , I did only code review here but in next lines I see
```php
'product_id' => 'product' === $path_pieces[1] ? absint( $path_pieces[2] ) : null,
```

which suggests that `count( $path_pieces )` could be more than 2

bor0: Ah, right! It should be 3. I think it expects the path to be a valid URL, but maybe Kyle can confirm.

```
wp> explode( '/', wp_parse_url( $path, PHP_URL_PATH ) );
=> array(1) {
  [0]=>
  string(4) "home"
}
wp> $path = 'http://localhost:8080/wp-admin/admin.php?page=wc-admin&path=home'
=> string(64) "http://localhost:8080/wp-admin/admin.php?page=wc-admin&path=home"
wp> explode( '/', wp_parse_url( $path, PHP_URL_PATH ) );
=> array(3) {
  [0]=>
  string(0) ""
  [1]=>
  string(8) "wp-admin"
  [2]=>
  string(9) "admin.php"
}
```

bor0: Fixed in 0fac6af5de, thanks!

kdevnel (approved):
This looks good and tests as described. My concern is that this is a hard-coded value. If  `$path_pieces` ever changes, you have to make sure to change it here. Shouldn't be a problem, but I feel like a code comment on those lines would be a good thing here.

plugins/woocommerce/src/Admin/Features/ProductBlockEditor/RedirectionController.php
```diff
@@ -155,6 +155,13 @@ protected function get_parsed_route(): array {
 		$path        = esc_url_raw( wp_unslash( $_GET['path'] ) );
 		$path_pieces = explode( '/', wp_parse_url( $path, PHP_URL_PATH ) );
 
+		if ( 3 !== count( $path_pieces ) ) {
+			return array(
+				'page'       => null,
```
mcliwanow: I see how this will make the error gone, but are we sure about this change? As it's not 1 to 1 replacement of the previous code., so it might result in unintended consequences. Unless it was also buggy.

Previous code would always return `'page'       => $path_pieces[1],` if the count of `$path_pieces` is at least 2. New code won't get there if count of `$path_pieces` is 2. 

`product_id` will work as expected

bor0: > As it's not 1 to 1 replacement

This is correct. If we want to keep it closer to the previous behavior, a safer change would be to use `$path_pieces[1] ?? ''` or similar approach.

bor0: Would appreciate another 👀 after 3be805b

mcliwanow: This change looks like it should work. I'll be able to retest it later today, if needed.

PR #55814, opened 2025-02-24
[e2e] Add support to install nightly builds

Stojdza:

Adding an option to install the latest nightly WC build during [on-demand](https://github.com/woocommerce/woocommerce/actions/workflows/tests-on-demand.yml) test execution on external envs (Pressable, WPCOM).

1. CI green. This change should not affect CI jobs and tests against wp-env
2. Example run of on-demand job can be found [here](https://github.com/woocommerce/woocommerce/actions/runs/13508973108/job/37745024407#step:9:127)
3. Go to [on-demand](https://github.com/woocommerce/woocommerce/actions/workflows/tests-on-demand.yml) and click on Run workflow, select this branch and in the last field type `nightly`
   - Take a look at the beginning of test execution. Is `9.8.0-dev` installed?
   - If yes, feel free to cancel the workflow. 

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Add support to install nightly builds in e2e tests. 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/tests/e2e-pw/fixtures/install-wc.setup.js
```diff
@@ -148,8 +152,14 @@ setup( 'Install WC using WC Beta Tester', async ( { wcbtApi, wpApi } ) => {
 		wpApi
 	);
 
-	if ( finalActivatedWcVersion === resolvedVersion ) {
-		console.log( 'Installing with WC Beta Tester is finished.' );
+	if (
+		wcVersion === 'nightly'
+			? finalActivatedWcVersion.endsWith( '-dev' )
```
adimoldovan, line 157: I don't think `endsWith( '-dev' )` is enough to validate the new version was successfully installed. If the previous version was also a nightly and the new nightly version install fails this will return true. Although, I'm not sure we can find the difference between 2 nightly versions at this point, so I guess it\s better than nothing?

Stojdza: I agree it's not an ideal solution, but I couldn't find a way to determine what is the nightly version that will be installed, and compare that to the currently installed version. So I ended with this check, as it's better than nothing. 

After some thinking, I figured out that I could get the version from [here](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/woocommerce.php), but 1) it will add even more complexity to the script and 2) it's not ideal case as well, because, _I think_, it can happen that versions [here](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/woocommerce.php) and in the generated nightly build are not the same. What do you think?

adimoldovan: I cannot think of a good solution, the nightly will have the same version. Let's merge as it is, we can come back and improve it if we can find a better solution.

adimoldovan (approved):

Issue #54992, opened 2025-01-30
[Performance]: Unrelated/unused scripts are loaded on admin.php?page=wc-admin* pages


moon0326:
### Prerequisites

- [x] I have carried out troubleshooting steps and I believe I have identified a performance issue.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Performance Details

On /wp-admin/admin.php?page=wc-admin&path=* pages, unused/unrelated scripts loaded. 

**Example pages:**

/admin.php?page=wc-admin
admin.php?page=wc-admin&path=%2Fcustomers
admin.php?page=wc-admin&path=%2Fmarketing

**Loaded scripts:**

block-serialization-default-parser.min.js
blocks.min.js
block-editor.min.js
block/templates/index.js
block-library.min.js
product-editor/index.js
csv-export/index.js

### Steps to Reproduce

1. Install the latest WC
2. Open browser inspector and click on `Network`
3. Search for one of the scripts listed.
4. Navigate to WooCommerce and other pages that starts with /wp-admin/admin.php?page=wc-admin

### Affected Sites

All

### Suspected causes

These scripts are probably being loaded unconditionally. We should load them only when necessary.

adrianduffell:
This might actually make things perceptively worse 🤔 At the moment, WCA instantly navigates to pages in its routes. If we remove the pre-loaded scripts, would it create a delay to load the scripts on demand when navigating to the dependent pages?

chihsuan:
> This might actually make things perceptively worse 🤔 At the moment, WCA instantly navigates to pages in its routes. If we remove the pre-loaded scripts, would it create a delay to load the scripts on demand when navigating to the dependent pages?

I think that's a valid concern. We should ensure optimization is done in a way that doesn't negatively impact the user experience.

louwie17:
@chihsuan I just pushed [up a PR](https://github.com/woocommerce/woocommerce/pull/55496) that removes the dependency for `wp-editor`, which I think is related or overlaps with this issue. I did a similar fix as what you guys had done for the customize your store package: https://github.com/woocommerce/woocommerce/pull/55496

chihsuan:
@louwie17 Thanks for the update! That's great to hear about removing the wp-editor dependency!

chihsuan:
> product-editor/index.js

[...]

PR #55822, opened 2025-02-25
Enhance WooCommerce Admin Performance by Splitting Scripts

chihsuan:

Related to https://github.com/woocommerce/woocommerce/issues/54992#issuecomment-2677531830

This PR refactors the WooCommerce Admin client-side architecture to improve performance and maintainability by splitting the scripts into two separate entry points:

1. WC App - For WooCommerce Admin pages
2. WC Embed - For embedded WooCommerce components in classic admin pages

Additionally, the PR includes:

- Use TypeScript to refactor the split client-side code
- Add more unit tests for related changes
- Extraction of shared logic into separate utility files to avoid duplication and loading unnecessary code
- Fix admin-layout slot fills component types as errors raised when using the components in the header.

The primary motivation for this change is to enhance performance for classic WooCommerce admin pages (orders, products, etc.). Currently, all WooCommerce admin scripts are loaded on every admin page, but embedded pages only require certain components. By splitting the scripts into separate entry points, we can significantly reduce the JavaScript payload for classic admin pages, resulting in faster page loads.

```
Before:
└── client/admin/
    └── client/
        ├── index.js                 # Single entry point for all admin pages
        ├── header/
        │   └── index.js             # Header components for all contexts
        ├── layout/
        │   └── index.js             # Layout components for all contexts
        └── stylesheets/
            └── _index.scss          # All styles bundled together

After:
└── client/admin/
    ├── webpack.config.js            # Updated with multiple entry points
    └── client/
        ├── index.tsx                # Entry point for WC Admin pages
        ├── embed.tsx                # New entry point for embedded pages
        ├── shared.tsx               # Shared utilities between both entry points
        ├── header/
        │   ├── index.tsx            # Header for WC Admin pages
        │   ├── embed.tsx            # Lightweight header for embedded pages
        │   └── shared.tsx           # Shared header components
        ├── layout/
        │   ├── index.tsx            # Layout for WC Admin pages
        │   ├── embed.tsx            # Lightweight layout for embedded pages
        │   └── shared.tsx           # Shared layout components
        └── stylesheets/
            ├── _index.scss          # Styles for WC Admin pages
            └── _embed.scss          # Minimal styles for embedded pages
```

#### Script size

Comparison of Script Sizes and Resources Loaded:

| Page   | Before: Entry Script Size | Before: Resources Loaded | After: Entry Script Size | After: Resources Loaded |
|--------|---------------------------|--------------------------|--------------------------|-------------------------|
| Home   | 211 kB                    | 8.7 MB                   | 74.1 kB                  | 8.6 MB                  |
| Order  | 211 kB                    | 9.1 MB                   | 135 kB                   | 8.6 MB                  |

Before Screenshots:
- Home: Screenshot 2025-02-25 at 16 36 21
- Order: Screenshot 2025-02-25 at 16 37 01

After Screenshots:
- Home: Screenshot 2025-02-25 at 17 17 20
- Order: Screenshot 2025-02-25 at 17 17 44

1. Go to the WooCommerce admin pages such as Home, Analytics, Marketing, etc.
2. Confirm that the pages load correctly, especially header, activity panel, and homescreen.
3. Go to `WooCommerce > Home > Add product task
4. Create a new product
5. Confirm `✏How easy was it to add a product?` notice is shown
6. Confirm that the Task list reminder bar is shown on the Product screen. 
7. Go to the embedded pages such as Orders, Products, etc.
8. Confirm that the pages load correctly, especially the header and activity panel.
9. Go to settings > site visibility and confirm that it renders correctly.
10. Smoke-test WC Admin pages to confirm that the changes do not introduce any regressions.

I've also verified the embedded layout renders the same DOM structure as before.

Screenshot 2025-02-25 at 17 16 48

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/admin/client/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+	// Render the App.
+	createRoot( appRoot ).render(
+		<ErrorBoundary>
+			<HydratedPageLayout />
+		</ErrorBoundary>
+	);
+
+	// Render the Customer Effort Score Tracks.
+	renderCustomerEffortScoreTracks( appRoot );
+}
```
moon0326, line 64: I think [reactify-classic-payments-settings](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/admin/client/index.js#L96) check has been removed.

I wasn't able to load WooCommerce -> Settings -> Payments with the feature flag enabled.

plugins/woocommerce/client/admin/client/shared.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { CustomerEffortScoreTracksContainer } from '@woocommerce/customer-effort-score';
+import { createRoot } from '@wordpress/element';
+import debugFactory from 'debug';
+
+/**
+ * Internal dependencies
+ */
+import { initRemoteLogging } from './lib/init-remote-logging';
+// Initialize remote logging early to log any errors that occur during initialization.
+initRemoteLogging();
```
moon0326: Since this file is meant to be shared by other components, do you think we should call this method conditionally? Or should we let the upper-level component decide when to call it? Importing a single function from this file will trigger initRemoteLogging, which might be unintentional. However, calling it multiple times wouldn’t be too bad since there is already a [check](https://github.com/woocommerce/woocommerce/blob/trunk/packages/js/remote-logging/src/remote-logger.ts#L453) in place.

moon0326 (commented):
This is amazing! Nice work 💯

Overall, it looks good and everything works as expected without any errors.

I've left two minor comments.

plugins/woocommerce/client/admin/client/index.tsx
moon0326: Please disregard this comment. It looks like this is implemented in https://github.com/woocommerce/woocommerce/pull/55823

plugins/woocommerce/client/admin/client/shared.tsx
chihsuan: Good point! I'll move this to upper-level component.

chihsuan: Updated in 1dfaccb337a3d9d9614f4803e409bfa47615e992

plugins/woocommerce/src/Internal/Admin/WCAdminAssets.php
```diff
@@ -284,7 +284,7 @@ public function register_scripts() {
 		);
 
 		$scripts_map = array(
-			WC_ADMIN_APP    => 'app',
+			WC_ADMIN_APP    => PageController::is_embed_page() ? 'embed' : 'app',
```
psealock, line 287: Have you done this so that scripts declaring `WC_ADMIN_APP` as a dependency won't trigger loadin of the whole wc-admin, just the embed script?

I'm wondering if there are any scripts out there depending on `WC_ADMIN_APP` and expect the entire app to be present. My guess is no, considering the `wc.*` packages are available on embed pages, but thought maybe worth mentioning.

psealock (approved):
This is looking really good @chihsuan. The code looks good and I've done a solid amount of poking around. Nothing jumps out, not errors found. Just a question about `WC_ADMIN_APP` script name.

plugins/woocommerce/src/Internal/Admin/WCAdminAssets.php
chihsuan: > Have you done this so that scripts declaring WC_ADMIN_APP as a dependency won't trigger loadin of the whole wc-admin, just the embed script?

Good question! Yes, the refactoring was done with the intention that scripts declaring `WC_ADMIN_APP` as a dependency will only trigger the loading of the necessary embed script rather than the entire `wc-admin`.

> I'm wondering if there are any scripts out there depending on WC_ADMIN_APP and expect the entire app to be present. My guess is no, considering the wc.* packages are available on embed pages, but thought maybe worth mentioning.

I believe there are no scripts that require the entire app to be present on embed pages to function properly. Additionally, this PR extracts the necessary code into the embed script, ensuring compatibility with existing scripts. The other codes are unused, and methods are not exposed globally like `wc.*, so third-party scripts should not be affected.

psealock (approved):
Sounds reasonable to me 🚢

PR #55824, opened 2025-02-25
[e2e tests] Fix strict mode violation in payment task check

adimoldovan:

Fixes the test 'Can visit the payment setup task from from the task list', which started [failing in daily runs](https://github.com/woocommerce/woocommerce/actions/runs/13513059854/job/37762224268). The cause was a strict mode violation. If the current task is `Get paid` there are two visible buttons with the same text.

This is not failing in trunk, where using 6 shards as opposed to daily runs where there are 5 shards only. The difference in tests order makes for a different environment state a thte time of this test execution.

Run shard 3/5

```
pnpm test:e2e --shard=3/5
```
Tested in CI here: https://github.com/woocommerce/woocommerce/actions/runs/13518943547/job/37773683956

### DIFF
--- a/plugins/woocommerce/changelog/e2e-fix-payment-task-get-paid-strict
+++ b/plugins/woocommerce/changelog/e2e-fix-payment-task-get-paid-strict
@@ -1,0 +1,4 @@
+Significance: patch
+Type: dev
+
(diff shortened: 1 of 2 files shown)


Stojdza (approved):
Looks good and works as expected :shipit:

PR #52067, opened 2024-10-16
Add the Cost of Goods Sold related code and REST APIs for the orders and order items

Konamiman:
### Changes proposed in this Pull Request:

Following https://github.com/woocommerce/woocommerce/pull/51675 this pull request adds programmatic and REST API support for the orders and order items related part of the Cost of Goods Sold feature.

#### Changes in code API

* The following methods are added to `WC_Order_Item`:
  * `has_cogs(): bool` - A boolean that by default returns false. This method must be overridden in derived classes representing line items that manage a COGS value.
  * `calculate_cogs_value(): bool` - If the calculation succeeds it sets the resulting value with `set_cogs_value` and returns true.
  * `calculate_cogs_value_core(): ?float` - It simply throws an exception. This method must be overridden whenever `has_cogs` returns true.
  * `get_cogs_value(): float` - Returns the value that was set with `set_cogs_value`.
  * `set_cogs_value(float): void` - Marked as `@internal`, intended for use only by `calculate_cogs_value` and when loading the object from the database.

* The following methods are added to `WC_Order_Item_Product` (overridding the methods in `WC_Order_Item`):
  * `has_cogs(): bool` - Returns true.
  * `calculate_cogs_value_core(): ?float` - Calculates the COGS value by multiplying the product value by the quantity. If the product no longer exists it returns null and then `calculate_cogs_value` won't change the current value.

* The following methods are added to `WC_Abstract_Order`:
  * `has_cogs(): bool` - Same meaning as in `WC_Order_Item`.
  * `calculate_cogs_total_value(): float` - Invokes `calculate_cogs_total_value_core` and sets the result with `set_cogs_total_value`.
  * `calculate_cogs_total_value_core(): float` - Calculates the value by looping through all the order items for which `has_cogs` returns true, invoking their `calculate_cogs_value` methods, and adding up the value returned by their `get_cogs_value` methods.
  * `get_cogs_total_value(): float` - Returns the value that was set with `set_cogs_value`.
  * `set_cogs_total_value(float): void` - Marked as `@internal`, intended for use only by `calculate_cogs_total_value` and when loading the object from the database.
 
* The following methods are added to `WC_Order`:
  * `has_cogs(): bool` - Overridden to always return true.

The order's `calculate_cogs_total_value` method is invoked in two places: inside `calculate_totals` in the same object, and in `WC_Checkout::create_order`.

This design allows to add new line item types and new order types in the future, with or without COGS management.

#### Storage

The COGS values are stored as follows:

[...]

PR #55840, opened 2025-02-25
Add Cost of Goods Sold information in the order editor in admin

Konamiman:

This pull request adds Cost of Goods Sold information in the order details page in admin:

image

The backend changes here are minimal since all the code API to deal with cost values in orders was already introduced in https://github.com/woocommerce/woocommerce/pull/52067.

Note that the existing "Cost" column is renamed to "Price" to avoid confussion. This change is permanent, it applies even when the Cost of Goods Sold feature is disabled.

Additional changes:

* Two new hooks are introduced: `woocommerce_order_item_no_cogs_html` and `woocommerce_order_item_cogs_html`.
* Given the above, to reduce confussion and ambiguity the existing `woocommerce_empty_cogs_html` and `woocommerce_get_cogs_html` hooks are renamed to `woocommerce_product_empty_cogs_html` and `woocommerce_product_get_cogs_html` respectively.
* A bug in the order items data store (introduced in https://github.com/woocommerce/woocommerce/pull/52067) is fixed: when the cost metadata for a line item had to be deleted, all the existing metadata entries for all the existing order items were actually deleted.

1. Enable the Cost of Goods Sold feature if you haven't done that yet (WooCommerce - Settings - Advanced - Features, check the box for "Cost of Goods Sold" and save).
2. Create a few products and assign costs to them (you can use the testing instructions of https://github.com/woocommerce/woocommerce/pull/54399). Include simple and variable products, and variations with and without an explicit cost set.
3. Create an order using the shop frontend, include the products you created or modified above. Add more than one unit of at least one of the products with cost.
4. Open the order details in admin and verify that the values displayed in the "Cost" column are correct for each order item. In particular, items corresponding to products with no cost should display zero; for variations with no explicit cost set, the default cost as set in the parent product should display; and for items with more than one unit of the product, the cost should be the unit cost times the quantity.
5. Verify that the "Cost Total" amount in the order summary is indeed the sum of all the item costs.
6. Edit the order by adding new items, deleting items, and modifying quantities. Verify that all costs are updated accordignly. (The exception is the case of adding new items: in this case you need to hit "Recalculate" for the cost of the new lines to show, but you need to do that anyway in order to update the order totals).
7. Verify that hitting "Recalculate" correctly updates the costs, even if you haven't changed anything in the order.
8. Disable the Cost of Goods Sold feature, reload the order page and verify that no cost information is shown whatsoever (but the column previously titled "Cost" still displays "Price" instead).
9. Repeat all with an order created directly from the admin backend.

There's an additional feature that needs a small code change to be tested. The `get_cogs_value_html` method of the `WC_Order_Item` class (and derived classes) will return a dash if the `has_cogs` method of the class returns false (as opposed to rendering a zero if `has_cogs` returns true but a particular instance has no cost defined), but we are using this method only for product items, for which `has_cogs` always returns true. So to test this behavior you can do the following temporary change to the `WC_Order_Item_Product` class: redefine the `has_cogs` method as follows:

```php
public function has_cogs(): bool {
	return $this->get_prop( 'cogs_value' ) !== 0.0;
}
```

Then reload the order details page and verify that the lines corresponding to products with a cost of zero display a dash, instead of zero, in the cost column.

With the previous code modification still in place, you can add the following snippet to test the new hooks:

```php
add_filter('woocommerce_order_item_no_cogs_html', fn($html, $item) => '<p>No cost for ' . $item->get_id() . '</p>', 2, 10);
add_filter('woocommerce_order_item_cogs_html', fn($html, $value, $item) => '<p>Cost for ' . $item->get_id() . ': ' . $value . '</p>', 3, 10);
```

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/includes/admin/meta-boxes/views/html-order-item.php
+++ b/plugins/woocommerce/includes/admin/meta-boxes/views/html-order-item.php
@@ -50,7 +50,7 @@ $row_class    = apply_filters( 'woocommerce_admin_html_order_item_class', ! empt
 	<?php do_action( 'woocommerce_admin_order_item_values', $product, $item, absint( $item_id ) ); ?>
 
 	<?php if ( wc_get_container()->get( CostOfGoodsSoldController::class )->feature_is_enabled() ) : ?>
-		<td class="item_cost" width="1%" data-sort-value="<?php echo esc_attr( $item->get_cogs_value() ); ?>">
+		<td class="item_cost_of_goods" width="1%" data-sort-value="<?php echo esc_attr( $item->get_cogs_value() ); ?>">
 			<div class="view">
 				<?php
 				echo $item->get_cogs_value_html();
@@ -58,7 +58,7 @@ $row_class    = apply_filters( 'woocommerce_admin_html_order_item_class', ! empt
 			</div>
 		</td>
 	<?php endif; ?>
-	<td class="item_price" width="1%" data-sort-value="<?php echo esc_attr( $order->get_item_subtotal( $item, false, true ) ); ?>">
+	<td class="item_cost" width="1%" data-sort-value="<?php echo esc_attr( $order->get_item_subtotal( $item, false, true ) ); ?>">
 		<div class="view">
 			<?php
 			echo wc_price( $order->get_item_subtotal( $item, false, true ), array( 'currency' => $order->get_currency() ) ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
--- a/plugins/woocommerce/includes/admin/meta-boxes/views/html-order-item.php
+++ b/plugins/woocommerce/includes/admin/meta-boxes/views/html-order-item.php
@@ -53,7 +53,7 @@ $row_class    = apply_filters( 'woocommerce_admin_html_order_item_class', ! empt
 		<td class="item_cost_of_goods" width="1%" data-sort-value="<?php echo esc_attr( $item->get_cogs_value() ); ?>">
 			<div class="view">
 				<?php
-				echo $item->get_cogs_value_html();
+				echo wp_kses_post( $item->get_cogs_value_html() );
 				?>
 			</div>
 		</td>
--- a/plugins/woocommerce/includes/admin/meta-boxes/views/html-order-item.php
+++ b/plugins/woocommerce/includes/admin/meta-boxes/views/html-order-item.php
@@ -9,6 +9,8 @@
 
 defined( 'ABSPATH' ) || exit;
 
+use Automattic\WooCommerce\Internal\CostOfGoodsSold\CostOfGoodsSoldController;
+
 $product      = $item->get_product();
 $product_link = $product ? admin_url( 'post.php?post=' . $item->get_product_id() . '&action=edit' ) : '';
 $thumbnail    = $product ? apply_filters( 'woocommerce_admin_order_item_thumbnail', $product->get_image( 'thumbnail', array( 'title' => '' ), false ), $item_id, $item ) : '';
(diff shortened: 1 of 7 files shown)


poligilad-auto:
Thanks for this @Konamiman!
I don't know if this is possible as part of this PR, but @lauroraa (our UX copywriter) would like to push to consistently apply Sentence case to all copy, currently the admin is very mixed. 

So, could we change "Items Subtotal", "Order Total" and "Cost Total" > "Items subtotal", "Order total" and "Cost total".

I see there is also ["Transaction Fee"](https://prnt.sc/revn2jWk0ssl), but I guess that's added by WooPayments?

Konamiman:
> So, could we change "Items Subtotal", "Order Total" and "Cost Total" > "Items subtotal", "Order total" and "Cost total".

It would be possible but for consistency and better housekeeping that's what I would do:

1. In this PR use "Cost Total" which is consistent with the current casing of the other existing labels of the page.
2. Then, after the PR is merged, open a new one where the casing for all the labels is changed.

plugins/woocommerce/includes/admin/meta-boxes/views/html-order-item.php
```diff
@@ -47,7 +49,16 @@
 
 	<?php do_action( 'woocommerce_admin_order_item_values', $product, $item, absint( $item_id ) ); ?>
 
-	<td class="item_cost" width="1%" data-sort-value="<?php echo esc_attr( $order->get_item_subtotal( $item, false, true ) ); ?>">
+	<?php if ( wc_get_container()->get( CostOfGoodsSoldController::class )->feature_is_enabled() ) : ?>
+		<td class="item_cost" width="1%" data-sort-value="<?php echo esc_attr( $item->get_cogs_value() ); ?>">
+			<div class="view">
+				<?php
+				echo $item->get_cogs_value_html();
```
jorgeatorres: Maybe we should pass this through `wp_kses_post()` or similar just in case. Given the value is filterable, it might contain unsafe HTML.

plugins/woocommerce/includes/admin/meta-boxes/views/html-order-item.php
```diff
@@ -47,7 +49,16 @@
 
 	<?php do_action( 'woocommerce_admin_order_item_values', $product, $item, absint( $item_id ) ); ?>
 
-	<td class="item_cost" width="1%" data-sort-value="<?php echo esc_attr( $order->get_item_subtotal( $item, false, true ) ); ?>">
+	<?php if ( wc_get_container()->get( CostOfGoodsSoldController::class )->feature_is_enabled() ) : ?>
+		<td class="item_cost" width="1%" data-sort-value="<?php echo esc_attr( $item->get_cogs_value() ); ?>">
```
jorgeatorres: I'm not too sure we should hijack the `item_cost` CSS class as there might be 3rd party scripts or even just styles targeting what now would be the "Price" column as it has always used the other class. I know it's a bit confusing, but I'd suggest we use a new CSS class for the COGS column and despite the new "Price" name leave `item_cost` as the CSS class for the usual price column.

What do you think?

plugins/woocommerce/includes/admin/meta-boxes/views/html-order-items.php
```diff
@@ -41,7 +43,10 @@
 			<tr>
 				<th class="item sortable" colspan="2" data-sort="string-ins"><?php esc_html_e( 'Item', 'woocommerce' ); ?></th>
 				<?php do_action( 'woocommerce_admin_order_item_headers', $order ); ?>
-				<th class="item_cost sortable" data-sort="float"><?php esc_html_e( 'Cost', 'woocommerce' ); ?></th>
+				<?php if ( $cogs_is_enabled ) : ?>
+					<th class="item_cost sortable" data-sort="float"><?php esc_html_e( 'Cost', 'woocommerce' ); ?></th>
+				<?php endif; ?>
+				<th class="item_price sortable" data-sort="float"><?php esc_html_e( 'Price', 'woocommerce' ); ?></th>
```
jorgeatorres: Same as above.

jorgeatorres (changes requested):
Hey @Konamiman!

This tests well. Thank you! I left some feedback but it's more about "backwards compatibility". Nothing seems broken.

plugins/woocommerce/includes/admin/meta-boxes/views/html-order-item.php
Konamiman: Ok, that makes sense since other parts of the same page are enclosed in `wp_kses_post` already. Done.

plugins/woocommerce/includes/admin/meta-boxes/views/html-order-item.php
Konamiman: Done, I've reverted the existing class to the original `item_cost` name, and renamed the new class to `item_cost_of_goods`.

jorgeatorres (approved):
LGTM. Thank you @Konamiman!

I've tested that:
- The new column doesn't appear on the order edit screen unless COGS is enabled in WC > Settings > Advanced > Features.
- The new column updates correctly when changing quantities and adding or removing items on the order edit screen.
- The items table on the order edit screen can be sorted as usual by different columns, including this new one.
- CSS styles look correct:
   Screenshot 2025-03-07 at 14 47 09

PR #55843, opened 2025-02-25
Related Products: avoids querying results on each page load

### LINKED ISSUE #30349: Loading product page always write something to DB
Writing to database every pageload is very bad idea (especially on frontend)

To replicate:
-) Install Query Monitor
-) Load product page
-) see list of 'UPDATE' queries from QM  console

Please read Wordpress code quality & best pratice here:
=> https://docs.wpvip.com/technical-references/code-review/vip-warnings/#h-front-end-database-writes
=> https://docs.wpvip.com/technical-references/code-quality-and-best-practices/term-queries-should-consider-include_children-false/

The problem is in wc_get_related_products() function here:
https://github.com/woocommerce/woocommerce/blob/b19500728b4b292562afb65eb3a0c0f50d5859de/includes/wc-product-functions.php#L894

If the related products count is under limit it will keeps doing this expensive taxonomy queries every single pageload. and set_transient is the one which will do 'UPDATE' query to db to update the transient_timeout

Proposed solution:
-) If related products count is under limit we just need to lower the timeout, from the default 'DAY_IN_SECONDS' to 'MINUTE_IN_SECONDS' or 'HOUR_IN_SECONDS'

cpapazoglou:

This PR improves the performance of related products handling in WooCommerce by:

1. Optimizing related products caching to query only once per day instead of on every page load
2. Moving related product transient deletion to an async process to prevent blocking operations

Closes https://github.com/woocommerce/woocommerce/issues/30349

1. Install Query Monitor
2. Edit the single product templated to show related products by category
3. Create three products in the same category ( master product, 2x related products)
4. Load master product page 
5. The related products should be shown
6. See list of 'UPDATE' queries from QM console
7. Reloading the page should not do consequent `UPDATEs` ( don't query related products on each page load )
8. Edit one related product and remove it from the category
9. Reload master product page
10. The related product should not be shown ( transient cleared )

|Before | After|
|-------|------|
|CleanShot 2025-02-25 at 20 06 09|CleanShot 2025-02-25 at 20 05 41|

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 
Query related products once per day when loading product details page

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/includes/wc-product-functions.php
```diff
@@ -1834,3 +1835,34 @@ function wc_product_attach_featured_image( $attachment_id, $product = null, $sav
...
+
+/**
+ * Delete related product transients when a product is saved.
+ * This is necessary because changing one product might affect many related products.
+ *
+ * @since 9.8.1
+ * @param int $product_id Product ID.
+ */
+function wc_delete_related_product_transients( $product_id ) {
+	global $wpdb;
```
cpapazoglou: I was looking at returning early only for new products when there isn't a category and tag set but by default all products get the `uncategorised` category 🤦 .

Also, I would like to bail early if there is no change in categories or tags when updating a product but couldn't find a way to identify changes.

plugins/woocommerce/includes/wc-product-functions.php
```diff
@@ -1834,3 +1835,34 @@ function wc_product_attach_featured_image( $attachment_id, $product = null, $sav
 	}
 }
 add_action( 'add_attachment', 'wc_product_attach_featured_image' );
+
+/**
+ * Delete related product transients when a product is saved.
```
cpapazoglou: Reasoning behind this nuclear approach https://github.com/woocommerce/woocommerce/issues/30349#issuecomment-2683037010

plugins/woocommerce/includes/wc-product-functions.php
```diff
@@ -1834,3 +1835,30 @@ function wc_product_attach_featured_image( $attachment_id, $product = null, $sav
...
+		)
+	);
+
+	// Also delete the actual transient entries
+	$wpdb->query( 
+		$wpdb->prepare( 
+			"DELETE FROM {$wpdb->options} WHERE option_name LIKE %s",
+			$wpdb->esc_like( '_transient_wc_related_' ) . '%'
+		)
+	);
```
Aljullu: Not a big deal, but wouldn't it be better to do this in one query instead of two? Or is there a reason to prefer doing two queries?

```suggestion
	// Delete both timeout and actual transient entries in one query
	$wpdb->query( 
	    $wpdb->prepare( 
	        "DELETE FROM {$wpdb->options} WHERE option_name LIKE %s OR option_name LIKE %s",
	        $wpdb->esc_like( '_transient_timeout_wc_related_' ) . '%',
	        $wpdb->esc_like( '_transient_wc_related_' ) . '%'
	    )
	);
```

plugins/woocommerce/includes/wc-product-functions.php
Aljullu: Those were my first thoughts when reviewing this PR, but I understand we don't have access to the previous categories/tags and because of that we have no way to know which products need to be updated, right?

But if that's the case, do we need to pass `$product_id` into this function? :thinking:

Aljullu (commented):
Thanks for working on this, @cpapazoglou! This is not my area of expertise, but I tried to do a review anyway.

One concern I have is that we are no longer calling [`delete_transient()`](https://developer.wordpress.org/reference/functions/delete_transient/) to delete those transients, so there are some actions that won't be triggered (ie: [`delete_transient_wc_related_ID`](https://developer.wordpress.org/reference/hooks/delete_transient_transient/)). I guess we could trigger the actions manually, but we would need to know which products are being updated.

I took a quick look and didn't find any extensions listening to those actions, but wondering about your thoughts on this. :thinking:

plugins/woocommerce/includes/wc-product-functions.php
cpapazoglou: Good idea, will do!

plugins/woocommerce/includes/wc-product-functions.php
cpapazoglou: Good catch! I had a call with Tom and decided to delete the transients only for the affected products ( old related products and new related products ) so function signature will change anyway. Thanks for the review!

cpapazoglou:
- Changed the function to clears only affected related products
- Moved the logic to the scheduler to avoid performance issues from calling `wc_get_related_products( $post_id, 1000 );`

Noticed that `[wc_delete_product_transients](https://github.com/woocommerce/woocommerce/blob/91d6172e4bac8159492972b477abde2bed28a62a/plugins/woocommerce/includes/wc-product-functions.php#L127)` is called twice when saving a product. I will investigate why.

CleanShot 2025-02-26 at 17 18 32

cpapazoglou:
Alright the duplicate save product call comes from `class-wc-meta-box-product-data.php:437` and `class-wc-meta-box-product-images.php:101` which seem unrelated to these changes and should be investigated in a separate issue (https://github.com/woocommerce/woocommerce/issues/55882)

Stack Traces
```
2025-02-26T15:45:22+00:00 Debug Product save() called. Stack trace: 
WC_Product::save in /woocommerce/plugins/woocommerce/includes/admin/meta-boxes/class-wc-meta-box-product-data.php on line 437
WC_Meta_Box_Product_Data::save in /wp-includes/class-wp-hook.php on line 324
WP_Hook::apply_filters in /wp-includes/class-wp-hook.php on line 348
WP_Hook::do_action in /wp-includes/plugin.php on line 517
do_action in /woocommerce/plugins/woocommerce/includes/admin/class-wc-admin-meta-boxes.php on line 273
WC_Admin_Meta_Boxes::save_meta_boxes in /wp-includes/class-wp-hook.php on line 326
WP_Hook::apply_filters in /wp-includes/class-wp-hook.php on line 348
WP_Hook::do_action in /wp-includes/plugin.php on line 517
do_action in /wp-includes/post.php on line 5060
wp_insert_post in /wp-includes/post.php on line 5162
wp_update_post in /wp-admin/includes/post.php on line 453
edit_post in /wp-admin/post.php on line 227

2025-02-26T15:45:22+00:00 Debug Product save() called. Stack trace: 
WC_Product::save in /woocommerce/plugins/woocommerce/includes/admin/meta-boxes/	
WC_Meta_Box_Product_Images::save in /wp-includes/class-wp-hook.php on line 324
WP_Hook::apply_filters in /wp-includes/class-wp-hook.php on line 348
WP_Hook::do_action in /wp-includes/plugin.php on line 517
do_action in /woocommerce/plugins/woocommerce/includes/admin/class-wc-admin-meta-boxes.php on line 273
WC_Admin_Meta_Boxes::save_meta_boxes in /wp-includes/class-wp-hook.php on line 326
WP_Hook::apply_filters in /wp-includes/class-wp-hook.php on line 348
WP_Hook::do_action in /wp-includes/plugin.php on line 517
do_action in /wp-includes/post.php on line 5060
wp_insert_post in /wp-includes/post.php on line 5162
wp_update_post in /wp-admin/includes/post.php on line 453
edit_post in /wp-admin/post.php on line 227
```

plugins/woocommerce/includes/wc-product-functions.php
```diff
@@ -158,6 +173,67 @@ function wc_delete_product_transients( $post_id = 0 ) {
...
+	// Create placeholders for the IN clause.
+	$placeholders = implode( ',', array_fill( 0, count( $transient_names ), '%s' ) );
+
+	// Delete specific timeout and transient entries in a single query.
+	$wpdb->query(
+		$wpdb->prepare(
+			"DELETE FROM {$wpdb->options} WHERE option_name IN ($placeholders)",
+			$transient_names
+		)
+	);
```
tjcafferkey: Not all transients are stored in the database. We can implement a check for this like the below (haven't verified it though)

```suggestion
	// Delete specific timeout and transient entries.
	if ( wp_using_ext_object_cache() ) {
		// When using external object cache, we need to delete each transient individually
		foreach ( $related_product_ids as $id ) {
			delete_transient( 'wc_related_' . $id );
		}
	} else {
		// For database storage, we can use a single query for better performance
		$wpdb->query(
			$wpdb->prepare(
				"DELETE FROM {$wpdb->options} WHERE option_name IN ($placeholders)",
				$transient_names
			)
		);
	}
```

plugins/woocommerce/includes/wc-product-functions.php
```diff
@@ -158,6 +173,67 @@ function wc_delete_product_transients( $post_id = 0 ) {
 	do_action( 'woocommerce_delete_product_transients', $post_id );
 }
 
+/**
+ * Asynchronously delete related product transients when a product is changed.
+ * This is necessary because changing one product might affect many related products.
+ *
+ * @since 9.8.1
```
tjcafferkey: This should be `9.8.0` since feature freeze for the release is on Moday, beyond that it would be `9.9.0`

plugins/woocommerce/includes/wc-product-functions.php
```diff
@@ -142,14 +143,28 @@ function wc_delete_product_transients( $post_id = 0 ) {
...
 			delete_transient( $transient . $post_id );
 		}
+
+		$is_cli = Constants::is_true( 'WP_CLI' );
+		if ( $is_cli ) {
+			wc_delete_related_product_transients( $post_id );
+		} else {
+			// Schedule the async deletion of related product transients.
+			// This should run async cause it also fetches the
+			// related products of the current product.
```
tjcafferkey: I think expanding on this comment would be beneficial. Something to the effect of "_Scheduled async because we can't be sure how many related product transients we're going to query_"

plugins/woocommerce/includes/wc-product-functions.php
```diff
@@ -1043,8 +1117,9 @@ function wc_get_related_products( $product_id, $limit = 5, $exclude_ids = array(
 	$transient     = get_transient( $transient_name );
 	$related_posts = $transient && is_array( $transient ) && isset( $transient[ $query_args ] ) ? $transient[ $query_args ] : false;
 
-	// We want to query related posts if they are not cached, or we don't have enough.
-	if ( false === $related_posts || count( $related_posts ) < $limit ) {
+	// Query related posts if they are not cached
+	// Should happen only once per day due to transient expiriation set to 1 DAY_IN_SECONDS, to avoid performance issues
```
tjcafferkey: Lint 
```suggestion
	// Should happen only once per day due to transient expiriation set to 1 DAY_IN_SECONDS, to avoid performance issues.
```

plugins/woocommerce/includes/wc-product-functions.php
```diff
@@ -456,12 +532,10 @@ function wc_get_formatted_variation( $variation, $flat = false, $include_names =
 				} else {
 					$variation_list[] = '<dt>' . wc_attribute_label( $name, $product ) . ':</dt><dd>' . rawurldecode( $value ) . '</dd>';
 				}
-			} else {
-				if ( $flat ) {
+			} elseif ( $flat ) {
 					$variation_list[] = rawurldecode( $value );
```
tjcafferkey:
```suggestion
				$variation_list[] = rawurldecode( $value );
```

plugins/woocommerce/includes/wc-product-functions.php
```diff
@@ -158,6 +173,67 @@ function wc_delete_product_transients( $post_id = 0 ) {
...
+		$transient_names[] = '_transient_wc_related_' . $id;
+	}
+
+	// Create placeholders for the IN clause.
+	$placeholders = implode( ',', array_fill( 0, count( $transient_names ), '%s' ) );
+
+	// Delete specific timeout and transient entries in a single query.
+	$wpdb->query(
+		$wpdb->prepare(
+			"DELETE FROM {$wpdb->options} WHERE option_name IN ($placeholders)",
```
tjcafferkey: I think this should be `({$placeholders})` for interoperability. But also I don't think the WP standards like using variables for this. You might have to inline it.

plugins/woocommerce/includes/wc-product-functions.php
```diff
@@ -158,6 +173,67 @@ function wc_delete_product_transients( $post_id = 0 ) {
...
+
+	if ( ! is_numeric( $post_id ) ) {
+		return;
+	}
+
+	$transient_name       = 'wc_related_' . $post_id;
+	$old_transient        = get_transient( $transient_name );
+	$old_related_products = array();
+
+	if ( is_array( $old_transient ) ) {
```
tjcafferkey: Lets also check it's not an empty array

```suggestion
	if ( is_array( $old_transient ) && ! empty( $old_transient ) ) {
```

tjcafferkey (changes requested):
Thanks for this @cpapazoglou. Overall I like the approach, and understand why we're scheduling an async action. I've just proposed some changes let me know what you think 😄

tjcafferkey:
It would also be great if we could include a description of the change in this PR 😃

cpapazoglou:
> Thanks for this @cpapazoglou. Overall I like the approach, and understand why we're scheduling an async action. I've just proposed some changes let me know what you think 😄

Thanks Tom, I have addressed your feedback in https://github.com/woocommerce/woocommerce/pull/55843/commits/a2fb4180827329f15c2e0de599e5b8d35f592e6f

tjcafferkey:
@cpapazoglou when I adding/removing a related product from the category and refresh the master product, the related product does not disappear/reappear as expected. I even gave it a few minutes thinking the action had not run but to not avail.

But, when I went to a different product and then _returned_ to my master product the related product was present/absent as expected.

Is there something triggering this only when I am visiting a new product page, rather than refreshing my current (master) product?

cpapazoglou:
> @cpapazoglou when I adding/removing a related product from the category and refresh the master product, the related product does not disappear/reappear as expected. I even gave it a few minutes thinking the action had not run but to not avail.
> 
> But, when I went to a different product and then _returned_ to my master product the related product was present/absent as expected.
> 
> Is there something triggering this only when I am visiting a new product page, rather than refreshing my current (master) product?

I believe this is a scheduling / timing thing. Can you monitor http://localhost:10003/wp-admin/admin.php?page=wc-status&tab=action-scheduler&orderby=schedule&order=desc whether the action has run?

Sometimes it's instant and other times can take up to 60secs in my testing. Here is a demo

https://github.com/user-attachments/assets/8eaa4dba-48ae-4d26-804a-a5061d059299

tjcafferkey (approved):
Thanks @cpapazoglou this works as expected and fixes the original issue. There are still CI checks to pass but I am pre-approving. Once they have all passed we can go ahead and merge 🙌🏻

PR #55853, opened 2025-02-26
Fix shipping calculator address validation

### LINKED ISSUE #52567: Cart - confusing behaviour on "check delivery address" when address is not changed
With the following settings:

Image

And shipping zones such as:

Image

When a user visits the cart and has not provided an address yet, we default the store country as partially filling the address. Which is ok, but it leads to a weird behaviour where clicking "Check delivery options" closes the address panel but does not show a delivery option which is useless. As far as UX is concerned the form is valid but no delivery option is shown.

I think the expected behaviour in this case should either be:

* Invalidate the address missing fields
* ~Show the "New Zealand" shipping option~

> [!NOTE]
> @ralucaStan mentioned that during the estimation session, the team agreed to focus on **Invalidate the address missing fields** as expected outcome.

This is because of logic in the code that says to close the panel if the address has not changed.

Screen recording:

https://github.com/user-attachments/assets/1716f447-b4a6-40fc-a266-e8f6b95aaebf

nielslange:

Closes #52567.

When the settings `Enable the shipping calculator on the cart page` and `Hide shipping costs until an address is entered` are enabled, and the shopper’s address is incomplete, clicking `Check delivery options` closes the shipping calculator without an error, but no shipping options are shown. The missing address fields do not show validation errors, leading to a poor user experience.

This PR ensures that when a shopper enters an incomplete address, the shipping calculator displays inline validation messages for the missing fields and prevents form submission. This ensures that shoppers understand which filed are missing to proceed.

#### Before all test cases:

1. Go to `/wp-admin/admin.php?page=wc-settings&tab=shipping&section=options`.
2. Enable `Enable the shipping calculator on the cart page`.
3. Enable `Hide shipping costs until an address is entered`.
4. Go to `/wp-admin/admin.php?page=wc-settings&tab=shipping`.
5. Set up shipping zones and methods as follows

Screenshot 2025-02-26 at 13 44 14

6. Go to `/wp-admin/admin.php?page=wc-settings&tab=general`.
7. Select `Netherlands` as the store address country.
8. Select `Shop country/region` as the default customer location.

#### Incomplete address without state field:

1. Open a new incognito window.
2. Add a product to the cart.
3. Go to the cart page.
4. Click on `Enter address to check delivery options` to open the shipping calculator.
5. Verify that the fields `Country/Region`, `Postal code` and `City` are visible.
6. Verify that only the field `Country/Region` is filled out.
7. Click on `Check delivery options`.
8. Verify that the inline validation messages for `Postal code` and `City` are visible.
9. Verify that the form cannot be submitted.
10. Provide a postal code, e.g. `1012 AB`.
11. Click on `Check delivery options`.
12. Verify that the inline validation message for `City` is visible.
13. Verify that the form cannot be submitted.
14. Remove the post code.
15. Provide a city, e.g. `Amsterdam`.
16. Click on `Check delivery options`.
17. Verify that the inline validation message for `Postal code` is visible.
18. Verify that the form cannot be submitted.
19. Provide a postal code, e.g. `1012 AB`.
20. Provide a city, e.g. `Amsterdam`.
21. Click on `Check delivery options`.
22. Verify that the form can be submitted.
23. Verify that the shipping method `Free shipping` is visible.

#### Incomplete address with optional state field:

1. Open a new incognito window.
2. Add a product to the cart.
3. Go to the cart page.
4. Click on `Enter address to check delivery options` to open the shipping calculator.
5. Select `Germany` as `Country/Region`.
6. Verify that the fields `Country/Region`, `Postal code`, `City` and `State/County (optional)` are visible.
7. Click on `Check delivery options`.
8. Verify that the inline validation messages for `Postal code` and `City` are visible.
9. Verify that the inline validation message for `State/County (optional)` is not visible, as this field is not required.
10. Verify that the form cannot be submitted.
11. Provide a postal code, e.g. `69102`.
12. Provide a city, e.g. `Heidelberg`.
13. Click on `Check delivery options`.
14. Verify that the form can be submitted.
15. Verify that the shipping method `Flat rate` is visible.

#### Incomplete address with required state field:

1. Open a new incognito window.
2. Add a product to the cart.
3. Go to the cart page.
4. Click on `Enter address to check delivery options` to open the shipping calculator.
5. Select `United States (US)` as `Country/Region`.
6. Verify that the fields `Country/Region`, `City`, `State`, and `ZIP Code` are visible.
7. Click on `Check delivery options`.
8. Verify that the inline validation messages for `City`, `State`, and `ZIP Code` are visible.
9. Provide a city, e.g. `San Francisco`.
10. Select a state, e.g. `California`.
11. Provide a zip code, e.g. `94110`.
12. Click on `Check delivery options`.
13. Verify that the form can be submitted.
14. Verify that the message `No delivery options available for ...` is visible.
15. Open the shipping calculator again.
16. Click on `Check delivery options`.
17. Verify that the form can still be submitted.
18. Open the shipping calculator again.
19. Change the city, e.g. `New York`.
20. Change the state, e.g. `New York`
21. Change the postal code, e.g. `10012`.
22. Click on `Check delivery options`. 
23. Verify that the form can still be submitted.

#### Filtered address

1. Install and activate the [Code Snippets](https://wordpress.org/plugins/code-snippets/) plugin.
2. Add the following code snippet (see [Disabling fields for a single country](https://developer.woocommerce.com/docs/cart-and-checkout-removing-checkout-fields/#1-disabling-fields-for-a-single-country)):

```php
add_filter('woocommerce_get_country_locale', function( $locale ) {

	$locale['DZ']['postcode']['required'] = false;
	$locale['DZ']['postcode']['hidden'] = true;

	$locale['DZ']['city']['required'] = false;
	$locale['DZ']['city']['hidden'] = true;

	 return $locale;
});
```

3. Open a new incognito window.
4. Add a product to the cart.
5. Go to the cart page.
6. Click on `Enter address to check delivery options` to open the shipping calculator.
7. Select `Algeria` as `Country/Region`.
8. Verify that the fields `Country/Region` and `State/County` are visible.
9. Click on `Check delivery options`.
10. Verify that the inline validation message for `State/County` is visible.
11. Select a state/county, e.g. `Adrar`.
12. Click on `Check delivery options`.
13. Verify that the form can be submitted.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

Prevent shipping calculator submission with an incomplete address and display inline validation errors.

</details>

<details>

#### Comment 

</details>


plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/address.tsx
```diff
@@ -34,16 +37,32 @@ const ShippingCalculatorAddress = ( {
 		( select ) => {
 			return {
 				hasValidationErrors:
-					select( validationStore ).hasValidationErrors,
+					select( validationStore ).hasValidationErrors(),
 				isCustomerDataUpdating:
 					select( CART_STORE_KEY ).isCustomerDataUpdating(),
 			};
 		}
```
samueljseay: This is a good change! While we're here can we pass `[]` to the useSelect as deps, to avoid redundant re-renders. To get the most from `useSelect` we must always pass deps, unfortunately it has not been done consistently in the past.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/address.tsx
```diff
@@ -34,16 +37,32 @@ const ShippingCalculatorAddress = ( {
...
+		'shipping',
+		address.country
+	);
+
+	const hasRequiredFields = () => {
+		const requiredFields = formFields
+			.filter( ( field ) => field.required && ! field.hidden )
+			.map( ( field ) => field.key );
+		return requiredFields.every(
+			( field ) => address[ field ] && address[ field ].trim() !== ''
```
samueljseay: I'm curious why we have a validation system via `hasValidationErrors` but it doesn't already handle validating required fields? Is there some discussion somewhere about this?

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/address.tsx
```diff
@@ -34,16 +37,32 @@ const ShippingCalculatorAddress = ( {
...
+		);
+	};
```
samueljseay: It's bad practice to create a closure over some value from a hook in case the callback could run with a stale value before the next re-render.

This could be converted to useCallback to ensure the callback always has the latest value:

```suggestion
	const hasRequiredFields = useEffect(() => {
		const requiredFields = formFields
			.filter( ( field ) => field.required && ! field.hidden )
			.map( ( field ) => field.key );
		return requiredFields.every(
			( field ) => address[ field ] && address[ field ].trim() !== ''
		);
	}, [ formFields ] );
```

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/address.tsx
```diff
@@ -62,25 +81,29 @@ const ShippingCalculatorAddress = ( {
 				variant="outlined"
 				onClick={ ( e ) => {
```
samueljseay: This callback should also use useCallback, not your fault, but we need to leave the code better than we found it. If we want to prevent redundant re-renders this is one easy way.

It's bad practice in most cases to pass an inline function to a component because every render a new function will be created, React will see that as a change and re-render the `Button`.

To simplify how many useCallbacks you have you have 2 options:

1. You could put the logic of `hasRequiredFields` into this callback and remove it altogether.
2. You could declare `hasRequiredFields` outside of the component because it doesn't need to be inline there, and I do see people put too many functions needlessly inside a component. Have it take `formFields` as an argument instead. There is really no need to create a closure around `formFields`.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/address.tsx
```diff
@@ -34,16 +37,32 @@ const ShippingCalculatorAddress = ( {
...
+		const requiredFields = formFields
+			.filter( ( field ) => field.required && ! field.hidden )
+			.map( ( field ) => field.key );
+		return requiredFields.every(
+			( field ) => address[ field ] && address[ field ].trim() !== ''
+		);
+	};
 
 	const validateSubmit = () => {
 		showAllValidationErrors();
```
samueljseay: This function is quite useless, let's just put this logic inside the callback passed to `Button` that we will migrate to `useCallback`

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/address.tsx
```diff
@@ -62,25 +81,29 @@ const ShippingCalculatorAddress = ( {
...
-							if ( typeof address[ key ] !== 'undefined' ) {
-								addressToSubmit[ key ] = address[ key ];
-							}
-						} );
-						return onUpdate( addressToSubmit );
+						if ( ! addressChanged ) {
+							return onCancel();
+						}
+
+						const addressToSubmit = addressFields.reduce(
```
samueljseay: I like the move to use reduce here, but this one will be tougher to type and there are a few typing problems here. 

Here's an alternative that's better typed:

```ts
const addressToSubmit: Partial< ShippingAddress > =
							Object.fromEntries(
								addressFields
									.filter(
										( key ) => address[ key ] !== undefined
									)
									.map( ( key ) => [ key, address[ key ] ] )
);
```

or you could use `for .. in`

Some tips though for using reduce

* Don't cast to `ShippingAddress` it's not true,  the return type here is gonna be a partial of `ShippingAddress` since you're filtering out keys/values that are not present.

* Use generics with `reduce` rather than casting the `accumulator`. e.g. `reduce<Partial<ShippingAddress>>`

Note that the update function won't like being passed a `Partial<ShippingAddress>` so it seems like it's type is wrong. We can try update it, but if it causes a lot of cascading type problems, we could cast as we pass it, add a comment why we're casting and raise an issue to fix the technical debt.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/address.tsx
```diff
@@ -34,16 +37,32 @@ const ShippingCalculatorAddress = ( {
 		( select ) => {
 			return {
 				hasValidationErrors:
-					select( validationStore ).hasValidationErrors,
+					select( validationStore ).hasValidationErrors(),
 				isCustomerDataUpdating:
 					select( CART_STORE_KEY ).isCustomerDataUpdating(),
```
samueljseay: how about using the store here like in the other call to get rid of the typing issue?

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/address.tsx
samueljseay: I've noted that we need to improve the types of `useFormFields` will maybe raise an issue for that: p1740669343260899-slack-C8X6Q7XQU 

I think though that this code could still be simplified, haven't tested but maybe:


```
const hasRequiredFields = (
  formFields: FormField[],
  address: Address
) =>
  formFields
    .filter((field) => field.required && !field.hidden)
    .every((field) => address[field.key]?.trim() !== "");
```

also doesn't need to be declared inline in the component, how about declaring outside the component and passing `formFields` like I suggested in another comment.

samueljseay (commented):
Haven't tested yet, but just looked at code. I will submit the review so far so you can have a look at my suggestions, and will test when I get a chance tomorrow.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/address.tsx
nielslange: Thanks for pointing this out. With https://github.com/woocommerce/woocommerce/pull/55853/commits/37af800851e0b4673ced601cc3dbfc4fadb35b87, I've wrapped this call in a `useCallback()` and improved the readability.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/address.tsx
nielslange: With c0b12704446a485d4cd4a15be92a7b09ea2dac23, I've extracted the inline function and created handleClick() instead.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/address.tsx
nielslange: With c0b12704446a485d4cd4a15be92a7b09ea2dac23, I've wrapped `validateSubmit()` with `useCallback()`. Thanks for suggesting these performance improvements.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/address.tsx
nielslange: I see that you've already addressed this with bdbae6595ed71e13ab9c88d69f032c8cd9f5e629. Thanks for improving this.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/address.tsx
nielslange: With 8f1bcb27304a19199621ea4df3062e5cfa8f93cd, I've picked up your first suggestion as I find that more readable.

> Note that the update function won't like being passed a Partial<ShippingAddress> so it seems like it's type is wrong. We can try update it, but if it causes a lot of cascading type problems, we could cast as we pass it, add a comment why we're casting and raise an issue to fix the technical debt.

I see that you also addressed this with bdbae6595ed71e13ab9c88d69f032c8cd9f5e629. Thanks a lot for doing so. 🙌

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/address.tsx
nielslange: I assume that due to bdbae6595ed71e13ab9c88d69f032c8cd9f5e629 this feedback can be marked as resolved, correct?

nielslange: Good question, @samueljseay. I'm not aware of any discussions regarding this. Maybe @opr knows if this has been discussed somewhere.

nielslange:
Thanks for your review, @samueljseay. I've addressed all of your suggestions. For https://github.com/woocommerce/woocommerce/pull/55853#discussion_r1973608914 and https://github.com/woocommerce/woocommerce/pull/55853#discussion_r1973905428 I left a comment. You could review the PR again now.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/test/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+	last_name: '',
+	company: '',
+	address_1: '',
+	address_2: '',
+	city: '',
+	state: '',
+	postcode: '',
+	country: '',
+	phone: '',
+} as ShippingAddress;
```
samueljseay: Don't need to cast do this:

`const address: ShippingAddress = {`

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/test/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+		} );
+
+		// Submit form with filled fields
+		fireEvent.click(
+			screen.getByRole( 'button', { name: 'Check delivery options' } )
+		);
+
+		// Verify validation errors are cleared
+		errorMessages.forEach( ( message ) => {
+			expect( screen.queryByText( message ) ).not.toBeInTheDocument();
```
samueljseay, line 78: Nice job adding some testing library tests here. 👌🏻

samueljseay (approved):
Thanks @nielslange I have made one small nitpick suggestion but I'm pre-approving.

I've tested all the steps you outlined successfully and happy with improvements you've made to the code.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/test/index.tsx
nielslange: Thanks, Sam.

plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/shipping-calculator/test/index.tsx
nielslange: Great catch. I've addressed that and merged the PR.

nielslange:
> Thanks @nielslange I have made one small nitpick suggestion but I'm pre-approving.
> 
> I've tested all the steps you outlined successfully and happy with improvements you've made to the code.

Thanks for reviewing and approving this PR, @samueljseay. And thanks for your great TS suggestions. 🙌

PR #55863, opened 2025-02-26
[dev][ci-jobs] Clean up the test env config options - remove unused cascade key

### LINKED ISSUE #47272: Add tests in ci-jobs covering dependency changes
With https://github.com/woocommerce/woocommerce/pull/47271 the ci-jobs utility returned test jobs for a parent project if a dependency was changed.
We should cover this with some unit tests.

adimoldovan:

The `cascade` key is an option in the CI jobs config we're not using. Instead, jobs are triggered when a dependency was changed, independent of it having cascading configured or not. Introduced with #47271 and then improved by #49330.
This PR removes the implementation of cascading keys, to remove potential confusion on how this works.

Also closes https://github.com/woocommerce/woocommerce/issues/47272

Check monorepo utils tests. They should pass.
Run the util locally with different pr numbers. Check the jobs output makes sense. Example

```
pnpm utils ci-jobs -p 55077 -e pull_request
```


tools/monorepo-utils/src/ci-jobs/index.ts
```diff
@@ -77,7 +88,7 @@ const program = new Command( 'ci-jobs' )
 		for ( const job of jobs.test ) {
 			const optional = job.optional ? ' (optional)' : '';
 			job.name = `${ job.name } - ${ job.projectName } [${ job.testType }]${ optional }`;
-			Logger.notice( `-  ${ job.name }` );
+			// Logger.notice( `-  ${ job.name }` );
```
kalessil, line 91: Should we uncomment this line?

kalessil (approved):
LGTM!

tools/monorepo-utils/src/ci-jobs/index.ts
adimoldovan: Yes, the jobs are being listed down the line again, this was only duplicating the output. I should have removed it instead.

PR #55872, opened 2025-02-26
Optimize product query for product creation task on page load

### LINKED ISSUE #54443: Unneeded slow count query used for Products Onboarding Task
## Issue

The query generated by `Automattic\WooCommerce\Admin\Features\OnboardingTasks\Tasks\Products::count_user_products` runs a slow full count via `SQL_CALC_FOUND_ROWS` and returns the default number of posts (normally 10) even though the data returned from this function is only ever used to check if at least one product exists.

Example Query:

```
SELECT SQL_CALC_FOUND_ROWS wp_posts.ID
FROM wp_posts LEFT JOIN wp_postmeta ON ( wp_posts.ID = wp_postmeta.post_id AND wp_postmeta.meta_key = '_headstart_post' ) LEFT JOIN wp_postmeta AS mt1 ON ( wp_posts.ID = mt1.post_id )
WHERE 1=1 AND (
wp_postmeta.post_id IS NULL
OR
mt1.meta_key = '_edit_last'
) AND wp_posts.post_type = 'product' AND ((wp_posts.post_status = 'publish'))
GROUP BY wp_posts.ID
ORDER BY wp_posts.post_date DESC
LIMIT 0, 10
```

The results are cached, however, they are also deleted any time a product is created/modified or any post type is moved to/from trash.

joshuatf:

The product creation task was making calls on each page load to check if a published, edited product exists.  This PR improves performance by doing a few things:

* Limits the need to re-query products once the task has initially been completed.  We use the existing option which tracks task completion.
* Optimizes the SQL by removing the use of `SQL_CALC_FOUND_ROWS` and adding `LIMIT 0,1` since we only need to know if a single record exists.
* Updates the transient on product update, creation, and untrash so we don't need to re-query products on the next page load.

Closes #54443 .

For the following tests, it can be helpful to download a transient manager such as [this](https://wordpress.org/plugins/transients-manager/) to help manage transients.

#### Test Query

For all tests, enable `WP_DEBUG` and `WP_DEBUG_LOG` in your wp-config.php file and add the below PHP snippet.
```PHP
add_filter( 'posts_request', function( $sql ) {
	error_log( $sql );
	return $sql;
}, 10, 1 );
```

1. Start with a fresh site with no products (you may need to delete the `woocommerce_product_task_has_product_transient` transient if you've previously visited a page)
4. Navigate to any WooCommerce page
5. In your debug log, find the query that includes `_headstart_post`.  It should look like this:
```SQL
SELECT   wp_posts.ID
FROM wp_posts
LEFT JOIN wp_postmeta ON ( wp_posts.ID = wp_postmeta.post_id AND wp_postmeta.meta_key = '_headstart_post' )  LEFT JOIN wp_postmeta AS mt1 ON ( wp_posts.ID = mt1.post_id )
WHERE 1=1  AND ( 
  wp_postmeta.post_id IS NULL 
  OR 
  mt1.meta_key = '_edit_last'
) AND wp_posts.post_type = 'product'
AND ((wp_posts.post_status = 'publish'))
GROUP BY wp_posts.ID
ORDER BY wp_posts.post_date DESC
LIMIT 0, 1
```
4. Note that `SQL_CALC_FOUND_ROWS` is not used in the query and the result is limited to 1.

#### Testing product updates cache

1. After visting any WooCommerce page, confirm that `woocommerce_product_task_has_product_transient` is set with a value of `no`.
2. Add a product
3. Check that the value of the transient has changed to `yes`
4. Update the transient value back to `no` and delete the `woocommerce_task_list_tracked_completed_tasks` option.
6. Update the existing product with any changes
7. Check that the value of the transient has changed back to `yes`
8. Trash the product
9. Update the transient value back to `no` and delete the `woocommerce_task_list_tracked_completed_tasks` option.
10. Untrash the product
11. Check that the value of the transient has changed back to `yes`
12. Check that the above query with `_headstart_post` has not been run again since the initial time.

#### Testing task completion 
1. Make sure at least one published product is added to your site.
7. Delete the transient `woocommerce_product_task_has_product_transient`
8. Make sure the above query with `_headstart_post` is not run again.
9. Optionally, delete all published products and check that the task is still complete

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/src/Admin/Features/OnboardingTasks/Tasks/Products.php
+++ b/plugins/woocommerce/src/Admin/Features/OnboardingTasks/Tasks/Products.php
@@ -11,7 +11,7 @@ use Automattic\WooCommerce\Internal\Admin\Onboarding\OnboardingProfile;
  * Products Task
  */
 class Products extends Task {
-	const PRODUCT_COUNT_TRANSIENT_NAME = 'woocommerce_product_task_product_count_transient';
+	const HAS_PRODUCT_TRANSIENT = 'woocommerce_product_task_has_product_transient';
 
 	/**
 	 * Constructor
@@ -182,7 +182,7 @@ class Products extends Task {
 	 * @return void
 	 */
 	public static function delete_product_count_cache() {
-		delete_transient( self::PRODUCT_COUNT_TRANSIENT_NAME );
+		delete_transient( self::HAS_PRODUCT_TRANSIENT );
 	}
 
 	/**
@@ -191,28 +191,18 @@ class Products extends Task {
 	 * @return bool
 	 */
 	public static function has_products() {
-		$product_counts = get_transient( self::PRODUCT_COUNT_TRANSIENT_NAME );
-		if ( false !== $product_counts && is_numeric( $product_counts ) ) {
-			return (int) $product_counts > 0;
+		$product_exists = get_transient( self::HAS_PRODUCT_TRANSIENT );
+		if ( $product_exists ) {
+			return 'yes' === $product_exists;
 		}
 
-		$product_counts = self::count_user_products();
-		set_transient( self::PRODUCT_COUNT_TRANSIENT_NAME, $product_counts );
-		return $product_counts > 0;
-	}
-
-	/**
-	 * Count the number of user created products.
-	 * Generated products have the _headstart_post meta key.
-	 *
-	 * @return int The number of user created products.
-	 */
-	private static function count_user_products() {
 		$args = array(
-			'post_type'   => 'product',
-			'post_status' => ProductStatus::PUBLISH,
-			'fields'      => 'ids',
-			'meta_query'  => array( // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
+			'post_type'      => 'product',
+			'post_status'    => ProductStatus::PUBLISH,
+			'posts_per_page' => 1,
+			'no_found_rows'  => true,
+			'fields'         => 'ids',
+			'meta_query'     => array( // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
 				'relation' => 'OR',
 				array(
 					'key'     => '_headstart_post',
@@ -227,7 +217,9 @@ class Products extends Task {
 
 		$products_query = new \WP_Query( $args );
 
-		return $products_query->found_posts;
+		$value = $products_query->post_count > 0 ? 'yes' : 'no';
+		set_transient( self::HAS_PRODUCT_TRANSIENT, $value );
+		return $value;
 	}
 
 	/**
--- a/plugins/woocommerce/src/Admin/Features/OnboardingTasks/Tasks/Products.php
+++ b/plugins/woocommerce/src/Admin/Features/OnboardingTasks/Tasks/Products.php
@@ -82,6 +82,10 @@ class Products extends Task {
 	 * @return bool
 	 */
 	public function is_complete() {
+		if ( $this->has_previously_completed() ) {
+			return true;
+		}
+
 		return self::has_products();
 	}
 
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/src/Admin/Features/OnboardingTasks/Tasks/Products.php
```diff
@@ -173,42 +176,60 @@ public function possibly_add_load_sample_return_notice_script( $hook ) {
...
+	 * @param int $product_id Product ID.
+	 * @param WC_Product $product Product object.
 	 */
-	public static function has_products() {
-		$product_counts = get_transient( self::PRODUCT_COUNT_TRANSIENT_NAME );
-		if ( false !== $product_counts && is_numeric( $product_counts ) ) {
-			return (int) $product_counts > 0;
+	public function maybe_set_has_product_transient( $product_id, $product ) {
+		if ( ! $this->has_previously_completed() && $this->is_valid_product( $product ) ) {
+			set_transient( self::HAS_PRODUCT_TRANSIENT, 'yes' );
```
prettyboymp, line 199: Should we go ahead and mark the task as previously completed here as well?  Or is there a reason to wait until WC Admin is visited?

prettyboymp (approved):
Great work.  Everything tested well and looks good.

One thing I noticed, which isn't specific to this PR or a blocker for it, is that the `Task::possibly_track_completion()` only runs to mark a task complete during `::get_json()` when loading for WC Admin.  So the dashboard widget WC_Admin_Dashboard_Setup will still run checks after an item is complete until a user visits WC Admin.

plugins/woocommerce/src/Admin/Features/OnboardingTasks/Tasks/Products.php
joshuatf: We can go ahead and call `possibly_track_completion()` here, though we'll keep to keep the setting of the transient to avoid circular logic.

The completion is probably due for a small refactor if we want to start actively marking tasks complete instead of passively/lazily as we currently do.

joshuatf:
>One thing I noticed, which isn't specific to this PR or a blocker for it, is that the Task::possibly_track_completion() only runs to mark a task complete during ::get_json() when loading for WC Admin. So the dashboard widget WC_Admin_Dashboard_Setup will still run checks after an item is complete until a user visits WC Admin.

Currently all tasks are checked for completion when they are fetched as opposed to when actual completion occurs.  These lazy checks are a bit less prone to errors in most cases and require less work on the IC implementing the task.

The downside is the performance cost of needing to calculate completion at fetch, which is mainly problematic if we encounter an expensive query like the one in this PR.

I'm not opposed to refactoring this a bit in another PR so that we can more actively mark tasks complete, though per our Slack discussion, I want to double check that it's okay to not revert task completion once a task has been complete from a UX perspective.

plugins/woocommerce/src/Admin/Features/OnboardingTasks/Tasks/Products.php
```diff
@@ -223,7 +244,9 @@ private static function count_user_products() {
 
 		$products_query = new \WP_Query( $args );
 
-		return $products_query->found_posts;
+		$value = $products_query->post_count > 0 ? 'yes' : 'no';
```
laurendavissmith, line 248: I could be misreading the code but I think if you change this to only store `$products_query->post_count > 0` instead of the string. Then up on L222 you can check for `$product_exists !== false` which could save a DB query in the event get_transient is returning false due to being set to false and not simply missing.

joshuatf: I believe that might actually be less efficient and never allow the negative case to use the transient.  For example:

```php
$product_exists = get_transient( self::HAS_PRODUCT_TRANSIENT );
if ( $product_exists !== false ) {
    // We won't ever come in here if the transient was set to false, so if the query has previously run, the transient can't be used and this forces us to re-query below.
    return $product_exists;
}
```

Let me know if I misunderstood your suggestion!

laurendavissmith: Yeah you're right, the storing of strings for future eval was a bit of a code smell for me and I thought I could save a db query in the process but yeah we'd end up not hitting it in some cases making it worse 👍

szepeviktor:
@Chouby Here is the surprise!

PR #55881, opened 2025-02-26
[Experimental] Add product review block

louwie17:

This PR is for demo purposes and will be split up into multiple PRs, items done:
**New blocks**
- Created blockified product reviews block in replacement of the `core/comments` block ( this is mostly just a template container )
- Review rating block ( the stars for a specific review ) this does not exist in WP
- Review form
     - New block as the form is extendable and relies on a couple different functions when rendering the front-end that allows the rating to be added for example.
- Reviews title
     - Couldn't use the comments title as "Comment" was hardcode everywhere, also it was relying on the `comments` endpoint which doesn't work for reviews.

**Variation blocks**
- Review template
     - Needed to overwrite the `Edit` function in order to use the `reviews` data store when rendering within a product query loop for example. Front-end works fine as is. 
- Review author
     - Needed to also overwrite the `Edit` function as placeholder is [hardcoded to `Comment`](https://github.com/WordPress/gutenberg/blob/c56672504bada434aee1ec331366cd2a10ab2b38/packages/block-library/src/comment-author-name/edit.js#L97). Given that logic relies on a `useSelect` and also rendered InspectorControls, it did not make sense to conditionally render our version.

**Variation blocks ( that still need to be created )**
- Review date
- Review content
- Review Pagination

https://github.com/user-attachments/assets/1fc2ebce-fc20-4ce3-bea9-7830571752d7

1.
2.
3.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


imanish003:
Great work @louwie17! I've completed some functional testing and identified a few items for discussion:

**Error Handling for Empty Reviews**

I noticed an error occurs when attempting to submit an empty comment or review. While this behavior appears to also exist in the legacy review tab, I believe we should address this issue in our new blockified product reviews implementation. Are you able to reproduce this error? Am I missing something?

https://github.com/user-attachments/assets/bf555748-cc31-45d6-a558-25209a03fa37

**Review Form State After Submission**

I suggest keeping the review accordion open after a review is submitted, consistent with the legacy reviews tab behavior. Currently, when users submit a review, the page scrolls to the reviews section and maintains it in an open state. Could we implement this same user experience in the new block? Let me know if I am missing something.

**Visual Consistency**

I noticed several UI differences between the new block and the legacy design. To maintain backward compatibility, I suggest aligning the new block's interface more closely with the existing one. However, if these UI changes were intentional, I'd appreciate understanding the reasoning behind moving away from the legacy design.

image

louwie17:
Thanks for looking at this and testing it @imanish003!

> **Error Handling for Empty Reviews**
> 
> I noticed an error occurs when attempting to submit an empty comment or review. While this behavior appears to also exist in the legacy review tab, I believe we should address this issue in our new blockified product reviews implementation. Are you able to reproduce this error? Am I missing something?

I agree that we should fix this, I see this on my local as well.

> **Review Form State After Submission**
> 
> I suggest keeping the review accordion open after a review is submitted, consistent with the legacy reviews tab behavior. Currently, when users submit a review, the page scrolls to the reviews section and maintains it in an open state. Could we implement this same user experience in the new block? Let me know if I am missing something.

Maybe? But should be able to, this is actually related to this conversation as well ( https://github.com/woocommerce/woocommerce/discussions/55224#discussioncomment-12117728 ) around being able to open a accordion by URL, and something I had previously discussed with @gigitux, that we did want to support.

> **Visual Consistency**
> 
> I noticed several UI differences between the new block and the legacy design. To maintain backward compatibility, I suggest aligning the new block's interface more closely with the existing one. However, if these UI changes were intentional, I'd appreciate understanding the reasoning behind moving away from the legacy design.

Yeah, one thing I didn't address in this PR was styling. Having said that, I don't think we need to keep the styling identical to the classic version. It will entirely depend on what the design looks like for the new block based theme.
Given this theme will also be using an accordion instead of tabs, the design will be quite a bit different anyhow, at the moment it mostly matches the design of the `comments` within WordPress.
If we wanted to, we could add some default styling attributes to the default template that aligns a bit more to the current version. I do think the current version looks kinda outdated though.

imanish003:
> Maybe? But should be able to, this is actually related to this conversation as well ( https://github.com/woocommerce/woocommerce/discussions/55224#discussioncomment-12117728 ) around being able to open a accordion by URL, and something I had previously discussed with @gigitux, that we did want to support.

I like the idea of opening an accordion via specific URL. 👍🏻

> Visual Consistency

I understand your perspective, but I have some concerns about deviating from the legacy design. Since we plan to replace the existing product review section with this new block in the template, maintaining visual consistency is important. When users update WooCommerce, they'll immediately notice these design changes, which could potentially break the look and feel of their existing stores. I believe making significant visual changes would constitute a breaking change. The core issue lies with the template itself, and we should avoid introducing breaking changes at the template level. Am I missing something?

gigitux:
> When users update WooCommerce, they'll immediately notice these design changes, which could potentially break the look and feel of their existing stores. I believe making significant visual changes would constitute a breaking change. The core issue lies with the template itself, and we should avoid introducing breaking changes at the template level. Am I missing something?

Just a quick note to clarify that we don’t plan to add this block to the Woo default template. We’ll consider replacing the current default block once a migration strategy is in place (no ETA).

For this reason, I think that this is a good opportunity to introduce significant visual changes if needed.

imanish003:
> once a migration strategy is in place (no ETA).

Just to confirm, by migration strategy, do you mean versioning for templates?

gigitux:
> > once a migration strategy is in place (no ETA).
> 
> Just to confirm, by migration strategy, do you mean versioning for templates?

Sorry for not being clear! Yes!

imanish003:
@gigitux Thank you for confirming. It would be really great to have versioning for templates ASAP. 💯

PR #55888, opened 2025-02-26
Version Checks: removes check for WP 6.1

### LINKED ISSUE #42105: Remove checks against WP 6.1 version
In several places of the codebase, we check whether the running WP version is higher than 6.1. We can already remove these checks as both [WC Blocks](https://github.com/woocommerce/woocommerce-blocks/blob/b5b6d6c6210691dfc91fb4ac81816068cd790b39/woocommerce-gutenberg-products-block.php#L12) and [WC core](https://https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/woocommerce.php#L11) require WP 6.3 or higher.

These checks can be found here:

https://github.com/search?q=repo%3Awoocommerce%2Fwoocommerce-blocks%20isWpVersion(%20%276.1%27&type=code

cpapazoglou:

Removes checks for WP 6.1 as [WC core](https://github.com/woocommerce/woocommerce/blob/ffff59f12979d9debea531db0cd99efbce4de94e/plugins/woocommerce/woocommerce.php#L11) requires WP 6.6 or higher.

Closes https://github.com/woocommerce/woocommerce/issues/42105

Automated testing should be sufficient

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 
Removes check for WP 6.1
</details>

<details>

#### Comment 

</details>


plugins/woocommerce-blocks/assets/js/blocks/classic-template/archive-product.ts
```diff
@@ -66,9 +65,7 @@ const getBlockifiedTemplateWithTermDescription = (
 ) => getBlockifiedTemplate( inheritedAttributes, true );
 
 const isConversionPossible = () => {
-	// Blockification is possible for the WP version 6.1 and above,
-	// which are the versions the Products block supports.
-	return isWpVersion( '6.1', '>=' );
+	return true;
```
kmanijak, line 68: With this change every `isConversionPossible` in templates return `true` now. What about removing the `isConversionPossible` completely and the related logic. I'm happy to contribute with that if you don't want to extend the scope of this PR.

plugins/woocommerce-blocks/assets/js/blocks/product-collection/utils.tsx
```diff
@@ -138,35 +137,33 @@ export function getDefaultValueOfFilterable() {
...
 
-				return blockSettings;
+			// Below condition is to support WP >=6.4 where Pagination specifies the parent.
+			// Can be removed when minimum WP version is set to 6.5 and higher.
+			if ( blockSettings?.parent ) {
+				return {
+					...blockSettings,
+					parent: [ ...blockSettings.parent, blockJson.name ],
+				};
 			}
```
kmanijak, line 162: Based on the comment, seems like this could be removed as well. However, can't recall how to test it without digging deeper so to avoid extending this PR scope it's fine to keep it as-is.

kmanijak (commented):
Thanks for addressing this issue! I left a comment about simplifying the logic in `classic-template` directory. Let me know what you think about it, I'm happy to help if needed! 🙌

kmanijak:
BTW I spotted there are also instances of `isWpVersion( '6.5', '>=' )` in the product editor. If the current minimum version is 6.6 we could potentially remove those as well, WDYT?

plugins/woocommerce-blocks/assets/js/blocks/classic-template/archive-product.ts
cpapazoglou: There are two instances with `false`

https://github.com/woocommerce/woocommerce/blob/4352324a699f6306f2274e483f9f1fb92f57b0c6/plugins/woocommerce-blocks/assets/js/blocks/classic-shortcode/index.tsx#L48

https://github.com/woocommerce/woocommerce/blob/4352324a699f6306f2274e483f9f1fb92f57b0c6/plugins/woocommerce-blocks/assets/js/blocks/classic-template/index.tsx#L53

I am afraid I don't have enough context here. I see it's used in two cases

https://github.com/woocommerce/woocommerce/blob/4352324a699f6306f2274e483f9f1fb92f57b0c6/plugins/woocommerce-blocks/assets/js/blocks/classic-shortcode/index.tsx#L189

https://github.com/woocommerce/woocommerce/blob/4352324a699f6306f2274e483f9f1fb92f57b0c6/plugins/woocommerce-blocks/assets/js/blocks/classic-template/index.tsx#L239

kmanijak: I see what you mean! Most likely it can be removed but it would require more testing and verification so I created a separate issue which I'm happy to tackle during cooldown: https://github.com/woocommerce/woocommerce/issues/55957!

cpapazoglou: Thanks for creating a follow up issue https://github.com/woocommerce/woocommerce/issues/55957

plugins/woocommerce-blocks/assets/js/blocks/product-collection/utils.tsx
cpapazoglou: I prefer not extending the scope of the current PR 😅

kmanijak (approved):
LGTM! Thanks for addressing and responding to all the comments! 🙌

PR #55902, opened 2025-02-27
Handle use of symbols in createDeprecatedObjectProxy

psealock:

This is a follow up to https://github.com/woocommerce/woocommerce/pull/55848 where array spreading was causing the proxy to error. That fixed worked, but something has changed on trunk causing a similar error, which I'm still unsure of.

I used Cursor's Composer tab extensively to arrive at a solution. This is an explanation of what was happening:

> Ah! Now I think I understand what's happening. Let's look at the key parts:
```js
const scripts = Array.from(
    new Set([
        ...(SETTINGS_SCRIPTS._default || []),
        ...(SETTINGS_SCRIPTS[activePage] || []),
    ])
);
```
> When you're spreading arrays (...) to create the Set, JavaScript uses the iterator Symbol (Symbol.iterator) internally. This is why we're hitting Symbol properties - not because of the URLs themselves, but because we're using array spreading on the arrays we get from the proxied object.
So the sequence is:
1. We try to spread `SETTINGS_SCRIPTS._default`
2. To spread an array, JavaScript needs to access its Symbol.iterator property
3. Our proxy tries to convert this Symbol to a string for the deprecation path
4. Boom! Error.
>We needed to handle the Symbol.iterator property access that happens during array spreading, even though we're just working with arrays of URL strings.

### Solution

Only processes deprecation warnings for string and number properties. I'm operating under the assumptions `wcSettings` will only contain those property types.

1. Add/Enable Gutenberg 19.8 or earlier.
2. Turn on `settings` feature flag.
3. WooCommerce > Settings should not fatally error.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


adrianduffell (approved):
It's testing well, nice! Possibly things get weird when the deprecated property has a `null` or `undefined` value. However I would prioritize merging this to fix settings rather than digging into the weeds. That case could be handled in a follow-up if needed 🚀

psealock:
>Possibly things get weird when the deprecated property has a null or undefined value.

Do you mean things may get weird if the `prop` value is null or undefined, ie `proxiedObj[ null ] = 'some defined value'`?

Yeah, that would be weird. In case you meant the other way, ie `proxiedObj[ 'some defined value' ] = null`, I added tests for that in [Add null and undefined tests](https://github.com/woocommerce/woocommerce/pull/55902/commits/8f1387a26464c28688c978b790e85dd95fc91a45).

>However I would prioritize merging this to fix settings rather than digging into the weeds.

👍🏽

moon0326 (approved):
LGTM and tested well 👍 🚀 

> Only processes deprecation warnings for string and number properties. I'm operating under the assumptions wcSettings will only contain those property types.

This sounds good to me!

PR #55905, opened 2025-02-27
Enable json_search_products to bring back products in any status if r…

normana10:

I _love_ using `<select class="wc-product-search"...`s in my admin-related plugins

One issue is that it seems the product search can't bring back "scheduled" products

I found this `false` here and by turning it to `true` I get back my scheduled products

So I added a GET param that callers can set if they'd like to search across products in _all_ statuses

1. Create a "scheduled" product
2. Find a `<select class="wc-product-search"` product search dropdown (such as the one in the "coupons" section)
3. Search for the product created in step 1, the "scheduled" product should **not** appear as an option
4. Modify it to include the `all_statuses=true` param
5. Search for the product created in step 1, the "scheduled" product should appear as an option

### Changelog entry

-   [x ] Automatically create a changelog entry from the details below.

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

Expanded Ajax method `json_search_products` to enable searching products in all statuses (such as "scheduled")

</details>

<details>

Expanded Ajax method `json_search_products` to enable searching products in all statuses (such as "scheduled")

</details>


Konamiman:
Hi @normana10, thanks for your contribution. The value passed to `search_products` should be a boolean, could you please change it to:

```php
filter_var( $_GET['all_statuses'] ?? false, FILTER_VALIDATE_BOOLEAN )
```

This makes sure that the value is intepreted correctly, so `true, `yes`, `on` and `1` are interpreted as true and everything else as false.

> Likely doesn't require a changelog as it's a minor addition to existing functionality?

We require a changelog for all pull requests, even for the smallest ones, so could you please add one?

Konamiman (changes requested):
Hi @normana10, thanks for your contribution. The value passed to `search_products` should be a boolean, could you please change it to:

```php
filter_var( $_GET['all_statuses'] ?? false, FILTER_VALIDATE_BOOLEAN )
```

This makes sure that the value is intepreted correctly, so `true`, `yes`, `on` and `1` are interpreted as true and everything else as false.

> Likely doesn't require a changelog as it's a minor addition to existing functionality?

We require a changelog for all pull requests, even for the smallest ones, so could you please add one?

normana10:
@Konamiman 

Makes sense to me

Change made and pushed

And changelog added

(`filter_var` is pretty neat, I was totally unaware of it!)

Konamiman:
Hi @normana10, I'm struggling a bit with the testing instructions. I might be confusing something (and I'm actually proficient with the backend than with the frontend), but I see that the product selector you mention is actually performing the product search by using an ajax request via JS; however your instructions seem to suggest that a simple change in markup is enough. Could you please provide more details regarding how you are introducing that query string argument?

nerrad:
The rationale and use case doesn't make a strong enough case for this change and the author didn't respond to a question.

PR #55916, opened 2025-02-27
[Performance] Rework unregistered statuses population to avoid heavy DB query.

### LINKED ISSUE #54464: Slow Admin\API\Reports\Order\DataStore::get_all_statuses() query being run when not needed

kalessil:

Closes #54464

SQL query is already optimized: distinct select over an indexed field is the fastest solution for this query type.

In this PR, we increase the transient retention for caching all order statuses in analytics to 12 months and add logic to keep it up-to-date with analytics syncing processes. Also we add some deprecations as per this discussion: p1740650640023039-slack-C0410KV3YLW.

The other context remains unchanged - the setting will be added on all admin pages as previously, but due to longer transient retention, the heavy DB query effectively goes away.

- Create multiple orders in different statuses (including custom one)
- Let analytics to sync the order data (should appear in `wp_wc_order_stats` DB table)
- Navigate through some admin pages and verify `SELECT DISTINCT status FROM wp_wc_order_stats` query is not exeuted
- Navigate to Analytics settings page from different entry points (Admin dashboard, Woocommerce hole) and ensure the custom statutes are listed on the analytics settings page.


plugins/woocommerce/src/Internal/Admin/Loader.php
```diff
@@ -408,9 +407,16 @@ public static function add_component_settings( $settings ) {
 		// WooCommerce Branding is an example of this - so pass through the translation of
 		// 'WooCommerce' to wcSettings.
 		$settings['woocommerceTranslation'] = __( 'WooCommerce', 'woocommerce' );
-		// We may have synced orders with a now-unregistered status.
-		// E.g An extension that added statuses is now inactive or removed.
-		$settings['unregisteredOrderStatuses'] = self::get_unregistered_order_statuses();
+
+		if ( PageController::is_admin_page() && Features::is_enabled( 'analytics' ) ) {
```
kalessil: Finding out in p1740650640023039-slack-C0410KV3YLW where the settings exposed in this class are getting used (for testing and testing instructions).

kalessil: Concluded in favour of deprecations in the loader class

plugins/woocommerce/src/Internal/Admin/Settings.php
```diff
@@ -207,10 +212,14 @@ public function add_component_settings( $settings ) {
 		// WooCommerce Branding is an example of this - so pass through the translation of
 		// 'WooCommerce' to wcSettings.
 		$settings['woocommerceTranslation'] = __( 'WooCommerce', 'woocommerce' );
-		// We may have synced orders with a now-unregistered status.
-		// E.g An extension that added statuses is now inactive or removed.
-		if ( PageController::is_admin_page() ) {
-			$settings['unregisteredOrderStatuses'] = $this->get_unregistered_order_statuses();
+
+		if ( PageController::is_admin_page() && Features::is_enabled( 'analytics' ) ) {
+			$is_analytics_page = str_starts_with( wc_clean( wp_unslash( $_GET['path'] ?? '' ) ), '/analytics/' );
```
joshuatf: Unfortunately, these settings need to be hydrated on all wc-admin page loads and not just those with `analytics` in the path.

You can see this breaking if we start on `/wp-admin/admin.php?page=wc-admin&path=%2Fanalytics%2Fsettings` you should see unregistered statuses:
Screenshot 2025-02-28 at 10 41 56 AM

However, if we start on WooCommerce -> Home (refresh the page there) and navigate to Analytics -> Settings, we'll see those statuses are missing:

Screenshot 2025-02-28 at 10 45 00 AM

joshuatf (commented):
Noted an issue with this in the comments.  I think the fix for this is either (short-term) more aggressive caching or (ideally) async settings requests so we don't need to hydrate all page data.

This is a long time issue in wc-admin and at some point we'll need to have a better system in place to get page-specific settings without hydrating them on all pages.

plugins/woocommerce/src/Internal/Admin/Settings.php
kalessil: Re-evaluated how to play nicely with hydration and re-iterating with long-living transient (1 week -> 1 month), which gets actualized on order status change.

plugins/woocommerce/src/Admin/API/Reports/Orders/DataStore.php
```diff
@@ -32,6 +31,13 @@ public function __construct() {
 		parent::__construct();
 	}
 
+	/**
+	 * Set up all the hooks for maintaining data consistency (transients and co).
+	 */
+	public static function init() {
+		add_action( 'woocommerce_analytics_update_order_stats', array( __CLASS__, 'actualize_all_statuses_transient_if_necessary' ) );
```
joshuatf: Nice!  This is a creative way to keep the cache up to date.

plugins/woocommerce/src/Admin/API/Reports/Orders/DataStore.php
```diff
@@ -612,9 +618,7 @@ protected function get_order_attributions_by_order_ids( $order_ids ) {
 	public static function get_all_statuses() {
 		global $wpdb;
 
-		$cache_key = 'orders-all-statuses';
-		$statuses  = Cache::get( $cache_key );
-
+		$statuses = get_transient( 'woocommerce_analytics_orders_statuses_all' );
```
joshuatf: Any particular reason we're shifting over to manual implementation of the cache over the `Cache` class for reports?

plugins/woocommerce/src/Admin/API/Reports/Orders/DataStore.php
```diff
@@ -623,12 +627,36 @@ public static function get_all_statuses() {
...
+	 * Ensure the order status will present in `get_all_statuses` call result.
+	 *
+	 * @param int $order_id Order ID.
+	 */
+	public static function actualize_all_statuses_transient_if_necessary( $order_id ) {
+		$order = wc_get_order( $order_id );
+		if ( $order ) {
+			$status = self::normalize_order_status( $order->get_status() );
+			if ( 'shop_order_refund' === $order->get_type() ) {
+				$parent_order = wc_get_order( $order->get_parent_id() );
```
joshuatf: Do we need to get the parent here?  I wonder if can skip this section since any updates to the parent will also be processed separately.

plugins/woocommerce/src/Internal/Admin/Loader.php
```diff
@@ -445,20 +444,19 @@ public static function add_component_settings( $settings ) {
...
+	 * @deprecated migrate to \Automattic\WooCommerce\Internal\Admin\Settings instead.
 	 */
 	public static function get_order_statuses( $statuses ) {
-		$formatted_statuses = array();
-		foreach ( $statuses as $key => $value ) {
-			$formatted_key                        = preg_replace( '/^wc-/', '', $key );
-			$formatted_statuses[ $formatted_key ] = $value;
-		}
-		return $formatted_statuses;
+		return Settings::get_order_statuses( $statuses );
```
joshuatf, line 454: Nice clean-up 👏

plugins/woocommerce/src/Internal/Admin/Loader.php
```diff
@@ -445,20 +444,19 @@ public static function add_component_settings( $settings ) {
 	 *
 	 * @param array $statuses Order statuses.
 	 * @return array formatted statuses.
+	 *
+	 * @deprecated migrate to \Automattic\WooCommerce\Internal\Admin\Settings instead.
```
joshuatf, line 449: Since we're deprecating this should we also add a `wc_deprecated_function` in this function and change the use of `self::get_order_statuses()` in this class?

plugins/woocommerce/src/Internal/Admin/Loader.php
```diff
@@ -475,61 +473,32 @@ public static function get_unregistered_order_statuses() {
 	 *
 	 * @param array $groups Array of setting groups.
 	 * @return array
+	 *
+	 * @deprecated migrate to \Automattic\WooCommerce\Internal\Admin\Settings instead.
```
joshuatf, line 482: Same question as above, re: deprecation.

plugins/woocommerce/src/Internal/Admin/Loader.php
```diff
@@ -475,61 +473,32 @@ public static function get_unregistered_order_statuses() {
...
+		return Settings::get_instance()->add_settings( $settings );
 	}
 
 	/**
 	 * Gets custom settings used for WC Admin.
 	 *
 	 * @param array $settings Array of settings to merge into.
 	 * @return array
+	 *
+	 * @deprecated migrate to \Automattic\WooCommerce\Internal\Admin\Settings instead.
```
joshuatf, line 510: Same comment as above, re: deprecation and removing this classes calls.

plugins/woocommerce/src/Admin/API/Reports/Orders/DataStore.php
```diff
@@ -623,12 +627,36 @@ public static function get_all_statuses() {
...
 
 		return $statuses;
 	}
 
+	/**
+	 * Ensure the order status will present in `get_all_statuses` call result.
+	 *
+	 * @param int $order_id Order ID.
+	 */
+	public static function actualize_all_statuses_transient_if_necessary( $order_id ) {
```
joshuatf: Nitpicky, but maybe the following to keep similar semantics across the repo:

```suggestion
	public static function maybe_update_order_statuses_transient( $order_id ) {
```

joshuatf (commented):
Left a few minor nitpicks and a question in the code, but this is testing perfectly for me and no blockers to the functionality!

The 1 year expiration is pretty long/aggressive, but given this narrow use case around analytics and that store owners can always flush transients if they want to refresh this, I think it's okay.

* ✅ Able to start on any of the SPA pages and navigate to Settings without regressions
* ✅ Added a custom status, noted that it was added in the transient via logging
* ✅ Removed the custom status from WooCommerce, confirmed that the custom status continued to show in analytics settings
* ✅ Confirmed that the DB call was not made again during this process via error logging

Custom statuses show after adding and navigating between SPA pages
Screenshot 2025-03-06 at 7 46 17 PM

Custom statuses continue to show after the status is unregistered
Screenshot 2025-03-06 at 7 47 04 PM

plugins/woocommerce/src/Admin/API/Reports/Orders/DataStore.php
kalessil: Thanks!

plugins/woocommerce/src/Admin/API/Reports/Orders/DataStore.php
kalessil: Yes, it's not allowing us to set a custom retention period, and I want to avoid modifying any public APIs in this PR.

plugins/woocommerce/src/Admin/API/Reports/Orders/DataStore.php
kalessil: I replicated the original logic from here: https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Admin/API/Reports/Orders/Stats/DataStore.php#L510. Didn't dig into the original reasoning though.

plugins/woocommerce/src/Internal/Admin/Loader.php
kalessil: I guess we can - trying it out.

plugins/woocommerce/src/Admin/API/Reports/Orders/DataStore.php
kalessil: Addressed in fee108fe4ce27baf12ebb39997a297865d196944

plugins/woocommerce/src/Internal/Admin/Loader.php
kalessil: > and change the use of self::get_order_statuses() in this class

Addressed in f713df1070e1357988ced54c7a186b1d71d05beb

kalessil: > Since we're deprecating this should we also add a wc_deprecated_function in this function

Addressed in 52d4c175df66d99f239337f8721b44766819ac77, 255d23ea43cc545d7c35a441269d75dead8f5639

plugins/woocommerce/src/Admin/API/Reports/Orders/DataStore.php
joshuatf: That makes sense.  Thanks for the explanation!

plugins/woocommerce/src/Admin/API/Reports/Orders/DataStore.php
joshuatf: I believe that's just needed for the lookup tables to store a status around the refund.  We can probably forgo that logic here since the original order would have already run through here.

plugins/woocommerce/src/Internal/Admin/Loader.php
```diff
@@ -445,26 +444,29 @@ public static function add_component_settings( $settings ) {
...
 	}
 
 	/**
 	 * Get all order statuses present in analytics tables that aren't registered.
 	 *
 	 * @return array Unregistered order statuses.
+	 *
+	 * @deprecated migrate to \Automattic\WooCommerce\Internal\Admin\Settings instead.
 	 */
 	public static function get_unregistered_order_statuses() {
```
joshuatf, line 464: There's one more call above to `self::get_unregistered_order_statuses()` that we can update to `Settings::get_unregistered_order_statuses()`.

plugins/woocommerce/src/Internal/Admin/Loader.php
```diff
@@ -445,26 +444,29 @@ public static function add_component_settings( $settings ) {
...
+		wc_deprecated_function( __CLASS__ . '::' . __FUNCTION__, '9.9.0', 'Direct `\Automattic\WooCommerce\Internal\Admin\Settings` use' );
+
 		$registered_statuses   = wc_get_order_statuses();
 		$all_synced_statuses   = OrdersDataStore::get_all_statuses();
```
joshuatf, line 468: I'm getting the following error when trying to call this function:

```
Fatal error: Uncaught Error: Class 'Automattic\WooCommerce\Internal\Admin\OrdersDataStore' not found
in /Users/joshua/Projects/woocommerce/plugins/woocommerce/src/Internal/Admin/Loader.php on line 467
```

We'll need to add `use Automattic\WooCommerce\Admin\API\Reports\Orders\DataStore as OrdersDataStore;` to this file.

plugins/woocommerce/src/Internal/Admin/Loader.php
```diff
@@ -445,26 +444,29 @@ public static function add_component_settings( $settings ) {
...
 
 	/**
 	 * Get all order statuses present in analytics tables that aren't registered.
 	 *
 	 * @return array Unregistered order statuses.
+	 *
+	 * @deprecated migrate to \Automattic\WooCommerce\Internal\Admin\Settings instead.
 	 */
 	public static function get_unregistered_order_statuses() {
+		wc_deprecated_function( __CLASS__ . '::' . __FUNCTION__, '9.9.0', 'Direct `\Automattic\WooCommerce\Internal\Admin\Settings` use' );
```
joshuatf: Since `get_unregistered_order_statuses` is private in `Settings`, the replacement should probably be `null|undefined`.

```suggestion
		wc_deprecated_function( __CLASS__ . '::' . __FUNCTION__, '9.9.0' );
```

joshuatf (approved):
✅ Tested again and no regressions since the last review
✅ Checked that deprecation warnings were shown for respective functions
⚠️ One broken function that existed prior to this PR and a couple of things we can remove

Thanks for the work on this, @kalessil!  Testing well and pre-approving.  Left a couple of comments on things we can update prior to merge, but pre-approving since they are all trivial changes.  Feel free to ping me if you feel you need another review, but I think we're good to go after those.

plugins/woocommerce/src/Admin/API/Reports/Orders/DataStore.php
kalessil: Ok, I'll simplify - based on the built context, we should be good.

plugins/woocommerce/src/Internal/Admin/Loader.php
kalessil: On it! Will also lookup if we can add a UT to smoke test this class.

plugins/woocommerce/src/Internal/Admin/Loader.php
kalessil: Good one, thanks!

plugins/woocommerce/src/Internal/Admin/Loader.php
kalessil: `\Automattic\WooCommerce\Internal\Admin\Settings::get_unregistered_order_statuses` is a private non-static one, unfortunately.

plugins/woocommerce/src/Internal/Admin/Loader.php
kalessil: Covered in 9ecbcc45a2290df3ad8580577a91e7e8c80a5ddd

plugins/woocommerce/src/Admin/API/Reports/Orders/DataStore.php
kalessil: Addressed in da98c43c6bf56b430b8fa24e899c03e5f619ca88

plugins/woocommerce/src/Internal/Admin/Loader.php
kalessil: Addressed in 90d9a76f09c5fa845dfa421faa7cbc9f582261c5

PR #55923, opened 2025-02-27
Minor fixes for email improvements feature

### LINKED ISSUE #55873: Minor issues in email improvements feature
1. https://github.com/woocommerce/woocommerce/pull/55747#discussion_r1967404976
2. ~https://github.com/woocommerce/woocommerce/pull/55747#discussion_r1967431238~ Moved to https://github.com/woocommerce/woocommerce/issues/55956
3. ~https://github.com/woocommerce/woocommerce/pull/55747#discussion_r1967435402~ Moved to https://github.com/woocommerce/woocommerce/issues/55956
4. https://github.com/woocommerce/woocommerce/pull/55690#discussion_r1967189777
5. Horizontal scrollbar in WooCommerce Subscriptions emails
    Image
6. Different errors when sending test email preview (they should be the same as in email preview)
    - In WooCommerce Bookings, which are not yet updated for email preview version
        Image
    - In any email that fails to render. Sending a test email will surface the actual error, instead of showing human-friendly one.
        Image
7. Missing email preview heading when the email preview is not floating
    Image

NeosinneR:

Out of the issues described in https://github.com/woocommerce/woocommerce/issues/55873:
1. ✅ Fixed
2. ❌ Can't replicate
3. ❌ Can't replicate
4. ✅ Fixed
5. ⚠️ This is an issue in the WC Subscriptions plugin. I'll report it there.
6. ✅ Fixed.
7. ✅ Fixed.

Closes #55873.

1. Follow description from https://github.com/woocommerce/woocommerce/issues/55873

### DIFF
--- a/packages/js/customer-effort-score/changelog/55873-email-improvements-fixes
+++ b/packages/js/customer-effort-score/changelog/55873-email-improvements-fixes
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix CES options being on two rows when scrollbar is visible
--- a/packages/js/customer-effort-score/src/style.scss
+++ b/packages/js/customer-effort-score/src/style.scss
@@ -3,27 +3,30 @@
 
 .woocommerce-customer-effort-score__selection {
 	margin: 1em 0 1.5em 0;
+	max-width: 100%;
+	width: 500px;
 
 	.woocommerce-customer-effort-score__radio-control {
 
 		.components-radio-control__group-wrapper,
 		.components-v-stack {
-			display: flex;
-			flex-direction: row;
-			flex-wrap: wrap;
+			display: grid;
+			grid-template-columns: repeat(auto-fit, minmax(90px, 1fr));
 			margin: 0 auto;
 			color: var(--wp-admin-theme-color);
-			gap: unset;
+			gap: 4px;
+			justify-items: center;
 		}
 	}
 
 
 
 	.components-radio-control__option {
+		column-gap: 0;
+		
 		&:not(:last-child) {
 			// Override package component style.
 			margin-bottom: 0;
-			margin-right: 4px;
 		}
 
 		input {
--- a/plugins/woocommerce/changelog/55873-email-improvements-fixes
+++ b/plugins/woocommerce/changelog/55873-email-improvements-fixes
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Prevent partial render when email preview rendering fails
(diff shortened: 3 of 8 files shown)


alex-mpoet (commented):
@NeosinneR Fixes 1, 4, 6 confirmed.

Could you please give more context about [this commit](https://github.com/woocommerce/woocommerce/pull/55923/commits/a8b0bffc62e7f99bc7edb41d6b2c539070698b23)? Is this also for fix number 6? Any example of what is this trying to fix? Will this suppress error messages so they won't be visible even with `WP_DEBUG` enabled? By the way, do you know what is the idiomatic WP way for a developer to get the real error message in [this case](https://github.com/woocommerce/woocommerce/pull/55923/files#diff-881f83e4deccd7fe511133ba99bcf26a2b782706bd638be7e8284d224e887869R285-R289), only modifying the source code?

2 & 3 are still reproducible in Firefox on Windows. On a slower host it's even more pronounced and even starts occurring in Chrome. I'll send you videos.

Update: notified you about finding the issue number 7 (missing email preview heading when the email preview is not floating).

NeosinneR:
@alex-mpoet 

> Could you please give more context about [this commit](https://github.com/woocommerce/woocommerce/pull/55923/commits/a8b0bffc62e7f99bc7edb41d6b2c539070698b23)? Is this also for fix number 6?

Yes, this is also for number 6, specifically the first screenshot for WooCommerce Bookings emails.

> Any example of what is this trying to fix?

When you look at the screenshot below, the WCB email starts rendering, and only then does an error occur. This is why I used the output buffer, so it prevents this specific case. It's easily replicable by doing any PHP error in the code (even a syntax error).

417172825-37d11148-6a15-4c59-a2e2-18b3dafc1f7e

> By the way, do you know what is the idiomatic WP way for a developer to get the real error message in [this case](https://github.com/woocommerce/woocommerce/pull/55923/files#diff-881f83e4deccd7fe511133ba99bcf26a2b782706bd638be7e8284d224e887869R285-R289), only modifying the source code?

Yeah, I think with this PR, you'd have to modify the source code to get the actual error - I've made this change mostly to improve the UX for merchants, where seeing an error might be more scary than a human-readable error. Do you think we should somehow allow getting an actual error?

alex-mpoet:
> When you look at the screenshot below, the WCB email starts rendering, and only then does an error occur. 

Thanks, what I looked for was in the background 🙂, now its clearer!

> Do you think we should somehow allow getting an actual error?

I'm not sure. MailPoet has a few places like this where you can only view the actual error after modifying the source code. Not the most handy way, and certainly resulted in a few support tickets where we just had to "uncover" the error for the user. I wonder if in WP or WooCommerce it's somehow solved, e.g. hiding the error when `WP_DEBUG` is disabled and showing when `WP_DEBUG` is enabled. I'm not aware if that is the case so I asked.

alex-mpoet (approved):
Approved!

PR #55931, opened 2025-02-27
Introduce _wc_delete_transients

tjcafferkey:

This utility function was inspired by this PR https://github.com/woocommerce/woocommerce/pull/55843 by @cpapazoglou whereby sometimes we need to delete a batch of transient data associated with a product.

Looping through and using `delete_transient` isn't always the most performant for batch deletion. This takes the original implementation of [delete_transient](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/option.php) and re-implements it for batch deletion.

### Performance Benefits

I benchmarked this function deleting 500 transients vs using `delete_transient` on each transient individually. The results show that `wc_delete_transients` is over 100% more performant.

1. This comes with unit tests. Review these tests and ensure they are passing as expected.
2. For batch deletion logic, enable Query Monitor and go to your sites homepage.
3. Go to Database Queries tab and filter by "DELETE"
4. Add the below code to your themes `function.php` file
5. Reload the page, and check that the transients are processed in batches.

```php
/**
 * Creates test transients and deletes them to test batch deletion logic.
 * 
 * This function creates 501 test transients with a specific prefix,
 * then uses wc_delete_transients() to delete them all at once.
 * 
 * @return void
 */
function test_transient_batch_deletion() {   
    // Create a unique prefix for our test transients
    $prefix = 'test_transient_batch_';
    
    // Create transients in smaller batches to avoid memory issues
    $batch_size = 50;
    $total_transients = 300;
    $transient_names = array();
    
    for ($batch = 0; $batch < ceil($total_transients / $batch_size); $batch++) {
        $batch_transients = array();
        $start = $batch * $batch_size + 1;
        $end = min(($batch + 1) * $batch_size, $total_transients);
        
        for ($i = $start; $i <= $end; $i++) {
            $transient_name = $prefix . $i;
            set_transient($transient_name, "Test value $i", HOUR_IN_SECONDS);
            $batch_transients[] = $transient_name;
        }
        
        // Add batch to the full list
        $transient_names = array_merge($transient_names, $batch_transients);
        
        // Free memory
        unset($batch_transients);
    }

	_wc_delete_transients($transient_names);
}
add_action('init', 'test_transient_batch_deletion');

```

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/includes/wc-core-functions.php
+++ b/plugins/woocommerce/includes/wc-core-functions.php
@@ -2751,8 +2751,8 @@ function wc_delete_transients( $transients ) {
 			// Use a single query for better performance.
 			$result = $wpdb->query(
 				$wpdb->prepare(
-					'DELETE FROM ' . $wpdb->options . ' WHERE option_name IN ( ' . implode( ', ', array_fill( 0, count( $transient_names ), '%s' ) ) . ' )',
-					$transient_names
+					'DELETE FROM ' . $wpdb->options . ' WHERE option_name IN ( ' . implode( ', ', array_fill( 0, count( $options_to_clear ), '%s' ) ) . ' )',
+					$options_to_clear
 				)
 			);
 
--- a/plugins/woocommerce/includes/wc-core-functions.php
+++ b/plugins/woocommerce/includes/wc-core-functions.php
@@ -2723,7 +2723,7 @@ function _wc_delete_transients( $transients ) {
 		}
 
 		// Limit the number of items in a single query to avoid exceeding database query parameter limits.
-		if ( count( $transient_names ) > 999 ) {
+		if ( count( $transient_names ) > 499 ) {
 			// Process in smaller chunks of 200 option names (100 transients) to reduce memory usage.
 			$chunks  = array_chunk( $transients, 100 );
 			$success = true;
--- a/plugins/woocommerce/includes/wc-core-functions.php
+++ b/plugins/woocommerce/includes/wc-core-functions.php
@@ -2686,3 +2686,78 @@ function wc_cache_get_multiple( $keys, $group = '', $force = false ) {
 	}
 	return $values;
 }
+
+/**
+ * Delete multiple transients in a single operation.
+ *
+ * This function efficiently deletes multiple transients at once, using a direct
+ * database query when possible for better performance.
+ *
+ * @since 9.9.0
+ * @param array $transients Array of transient names to delete (without the '_transient_' prefix).
+ * @return bool True on success, false on failure.
+ */
+function wc_delete_transients( $transients ) {
+	global $wpdb;
+
+	if ( empty( $transients ) || ! is_array( $transients ) ) {
+		return false;
+	}
+
+	// If using external object cache, delete each transient individually.
+	if ( wp_using_ext_object_cache() ) {
+		foreach ( $transients as $transient ) {
+			delete_transient( $transient );
+		}
+		return true;
+	} else {
+		// For database storage, create a list of transient option names.
+		$transient_names = array();
+		foreach ( $transients as $transient ) {
+			$transient_names[] = '_transient_timeout_' . $transient;
+			$transient_names[] = '_transient_' . $transient;
+		}
+
+		// Limit the number of items in a single query to avoid exceeding database query parameter limits.
+		if ( count( $transient_names ) > 999 ) {
+			// Process in chunks of 1000 option names (500 transients).
+			$chunks  = array_chunk( $transients, 500 );
+			$success = true;
+
+			foreach ( $chunks as $chunk ) {
+				$result = wc_delete_transients( $chunk );
+				if ( ! $result ) {
+					$success = false;
+				}
+			}
+
+			return $success;
+		}
+
+		try {
+			// Use a single query for better performance.
+			$result = $wpdb->query(
+				$wpdb->prepare(
+					'DELETE FROM ' . $wpdb->options . ' WHERE option_name IN ( ' . implode( ', ', array_fill( 0, count( $transient_names ), '%s' ) ) . ' )',
+					$transient_names
+				)
+			);
+
+			if ( false === $result ) {
+				wc_get_logger()->error(
+					sprintf( 'Database error when deleting transients: %s', $wpdb->last_error ),
+					array( 'source' => 'wc_delete_transients' )
+				);
+				return false;
+			}
+
+			return true;
+		} catch ( Exception $e ) {
+			wc_get_logger()->error(
+				sprintf( 'Exception when deleting transients: %s', $e->getMessage() ),
+				array( 'source' => 'wc_delete_transients' )
+			);
+			return false;
+		}
+	}
+}
--- a/plugins/woocommerce/includes/wc-core-functions.php
+++ b/plugins/woocommerce/includes/wc-core-functions.php
@@ -2796,7 +2796,7 @@ function _wc_delete_transients( $transients ) {
 		} catch ( Exception $e ) {
 			wc_get_logger()->error(
 				sprintf( 'Exception when deleting transients: %s', $e->getMessage() ),
-				array( 'source' => 'wc_delete_transients' )
+				array( 'source' => '_wc_delete_transients' )
 			);
 			return false;
 		}
--- a/plugins/woocommerce/includes/wc-core-functions.php
+++ b/plugins/woocommerce/includes/wc-core-functions.php
@@ -2723,8 +2723,8 @@ function _wc_delete_transients( $transients ) {
 		}
 
 		// Limit the number of items in a single query to avoid exceeding database query parameter limits.
-		if ( count( $transient_names ) > 499 ) {
-			// Process in smaller chunks of 200 option names (100 transients) to reduce memory usage.
+		if ( count( $transients) > 199 ) {
+			// Process in smaller chunks to reduce memory usage.
 			$chunks  = array_chunk( $transients, 100 );
 			$success = true;
 
--- a/plugins/woocommerce/includes/wc-core-functions.php
+++ b/plugins/woocommerce/includes/wc-core-functions.php
@@ -2690,14 +2690,18 @@ function wc_cache_get_multiple( $keys, $group = '', $force = false ) {
 /**
  * Delete multiple transients in a single operation.
  *
+ * IMPORTANT: This is a private function (internal use ONLY).
+ *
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/includes/wc-core-functions.php
```diff
@@ -2686,3 +2686,78 @@ function wc_cache_get_multiple( $keys, $group = '', $force = false ) {
...
 	return $values;
 }
+
+/**
+ * Delete multiple transients in a single operation.
+ *
+ * This function efficiently deletes multiple transients at once, using a direct
+ * database query when possible for better performance.
+ *
+ * @since 9.9.0
```
cpapazoglou: Should this be `9.8.0`?

plugins/woocommerce/tests/php/includes/wc-core-functions-test.php
```diff
@@ -224,4 +224,53 @@ function () {
 		$this->assertEquals( '', $result['state'] );
 		remove_all_filters( 'woocommerce_geolocate_ip' );
 	}
+
+	/**
+	 * Test wc_delete_transients() function.
+	 */
+	public function test_wc_delete_transients() {
```
tjcafferkey, line 231: I considered testing with 501 transients to verify batch deletion logic, but decided against it to maintain test performance. Is there any way we can test this without the overhead of creating hundreds of database records? 😕

plugins/woocommerce/tests/php/includes/wc-core-functions-test.php
```diff
@@ -224,4 +224,53 @@ function () {
...
+
+		// Verify transients exist before deletion.
+		$this->assertEquals( 'test_value_1', get_transient( $transient_name1 ) );
+		$this->assertEquals( 'test_value_2', get_transient( $transient_name2 ) );
+
+		// Delete the transients.
+		wc_delete_transients( array( $transient_name1, $transient_name2 ) );
+
+		// Force WordPress to refresh its memory cache.
+		wp_cache_flush();
```
cpapazoglou: Is this required? If so, I wonder if it should be part of `wc_delete_transients` itself.

plugins/woocommerce/tests/php/includes/wc-core-functions-test.php
```diff
@@ -224,4 +224,53 @@ function () {
...
+
+		// Verify transients are deleted.
+		$this->assertFalse( get_transient( $transient_name1 ) );
+		$this->assertFalse( get_transient( $transient_name2 ) );
+
+		// Test with a single transient name (string instead of array).
+		$transient_name3 = 'wc_test_transient_3';
+		set_transient( $transient_name3, 'test_value_3', HOUR_IN_SECONDS );
+		$this->assertEquals( 'test_value_3', get_transient( $transient_name3 ) );
+
```
cpapazoglou, line 254: Maybe also testing passing a single transient as string too before mapping it to an array?
```suggestion
$this->assertFalse( wc_delete_transients( $transient_name3 ) );
```

cpapazoglou (approved):
LGTM, I added some nitpicks.

plugins/woocommerce/includes/wc-core-functions.php
cpapazoglou: You beat me to it 💪

plugins/woocommerce/tests/php/includes/wc-core-functions-test.php
cpapazoglou: I think we can avoid the database records since the functionality to delete the records is already tested in other units. Perhaps, creating a dummy array with 501 transients and checking that `test_wc_delete_transients` is called twice?

plugins/woocommerce/tests/php/includes/wc-core-functions-test.php
tjcafferkey: Yeah good point, this forced me to dive a bit deeper into how we delete options and they also handle cache deletion which I've now added and mirrored in this functions functionality https://github.com/woocommerce/woocommerce/pull/55931/commits/12626b373f095286206e243952bff575521142fb

plugins/woocommerce/tests/php/includes/wc-core-functions-test.php
tjcafferkey: Good idea! Added in https://github.com/woocommerce/woocommerce/pull/55931/commits/1eb53749c418a925c9ef1ef0355f5a4b7eb3fd17

plugins/woocommerce/includes/wc-core-functions.php
```diff
@@ -2686,3 +2686,121 @@ function wc_cache_get_multiple( $keys, $group = '', $force = false ) {
...
+			// It also allows us to only delete the options we know exist.
+			$options_to_clear = array();
+			if ( ! wp_installing() ) {
+				$options_to_clear = $wpdb->get_col(
+					$wpdb->prepare(
+						'SELECT option_name FROM ' . $wpdb->options . ' WHERE option_name IN ( ' . implode( ', ', array_fill( 0, count( $transient_names ), '%s' ) ) . ' )',
+						$transient_names
+					)
+				);
+			}
```
tjcafferkey, line 2755: I wasn't sure about doing this DB query since we already _know_ the transient names, but this mirrors WP cores `delete_option` functionality and ensures we are only deleting the cache of options we _know_ exist.

plugins/woocommerce/tests/php/includes/wc-core-functions-test.php
tjcafferkey: Oh! Great idea. Let me work on that

plugins/woocommerce/includes/wc-core-functions.php
```diff
@@ -2686,3 +2686,121 @@ function wc_cache_get_multiple( $keys, $group = '', $force = false ) {
...
+
+			// Use a single query for better performance.
+			$result = $wpdb->query(
+				$wpdb->prepare(
+					'DELETE FROM ' . $wpdb->options . ' WHERE option_name IN ( ' . implode( ', ', array_fill( 0, count( $transient_names ), '%s' ) ) . ' )',
+					$transient_names
```
cpapazoglou: Should we use `$options_to_clear` instead of `$transient_names` since some transients might not exist as you indicated in https://github.com/woocommerce/woocommerce/pull/55931/files#r1975025661?
```suggestion
					'DELETE FROM ' . $wpdb->options . ' WHERE option_name IN ( ' . implode( ', ', array_fill( 0, count( $options_to_clear ), '%s' ) ) . ' )',
					$options_to_clear
```

tjcafferkey: Great spot. I agree!

plugins/woocommerce/includes/wc-core-functions.php
```diff
@@ -2686,3 +2686,119 @@ function wc_cache_get_multiple( $keys, $group = '', $force = false ) {
...
+		$transient_names = array();
+		foreach ( $transients as $transient ) {
+			$transient_names[] = '_transient_' . $transient;
+			$transient_names[] = '_transient_timeout_' . $transient;
+		}
+
+		// Limit the number of items in a single query to avoid exceeding database query parameter limits.
+		if ( count( $transient_names ) > 999 ) {
+			// Process in smaller chunks of 200 option names (100 transients) to reduce memory usage.
+			$chunks  = array_chunk( $transients, 100 );
```
cpapazoglou: It's weird allowing 999 at once but for bigger values we chunk them per 100. Should we check for `> 100`?

cpapazoglou (approved):
Tested it with
✅  `$total_transients = 300;` -> 1 database query
✅  `$total_transients = 501;` -> 6 database queries ( chunked per 100 )

Let's address https://github.com/woocommerce/woocommerce/pull/55931/files#r1975151771 and this should be ready to be merged.

plugins/woocommerce/includes/wc-core-functions.php
tjcafferkey: I think you're right in pointing out that we should reduce 999 as its quite a lot of records to process in one operation, and previously we were batching by 500 but that wasn't very efficient in terms of memory after testing.

What do you think about reducing 999 to 499?

tjcafferkey: Also to address your `> 100` question, I guess I was trying to balance memory vs efficiency. I'm open to changing the batch amount, or to threshold but I felt like batching 100 at a time was pretty safe above 499 transient names (or 250 transients).

tjcafferkey: https://github.com/woocommerce/woocommerce/pull/55931/commits/9d9820c77eb5ebc81048c4f48880ec53157da08b

plugins/woocommerce/includes/wc-core-functions.php
```diff
@@ -2686,3 +2686,119 @@ function wc_cache_get_multiple( $keys, $group = '', $force = false ) {
...
+		$transient_names = array();
+		foreach ( $transients as $transient ) {
+			$transient_names[] = '_transient_' . $transient;
+			$transient_names[] = '_transient_timeout_' . $transient;
+		}
+
+		// Limit the number of items in a single query to avoid exceeding database query parameter limits.
+		if ( count( $transient_names ) > 499 ) {
+			// Process in smaller chunks of 200 option names (100 transients) to reduce memory usage.
+			$chunks  = array_chunk( $transients, 100 );
```
cpapazoglou: What do you think of the following for making it simpler?
```suggestion
		if ( count( $transients) > 199 ) {
			// Process in smaller chunks to reduce memory usage.
			$chunks  = array_chunk( $transients, 100 );
```

cpapazoglou: Or even `>99`

plugins/woocommerce/includes/wc-core-functions.php
cpapazoglou: I followed up in https://github.com/woocommerce/woocommerce/pull/55931#discussion_r1975186095 🤦

plugins/woocommerce/includes/wc-core-functions.php
tjcafferkey: I've accepted your suggestion which I think is reasonable. This is an internal function so we can always update it should we need to in the future.

PR #55298, opened 2025-02-10
Add product_brand Term to Product Single HTML template.

rawdreeg:
### Submission Review Guidelines:

-   I have followed the [WooCommerce Contributing Guidelines](https://github.com/woocommerce/woocommerce/blob/trunk/.github/CONTRIBUTING.md) and the [WordPress Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/).
-   I have checked to ensure there aren't other open [Pull Requests](https://github.com/woocommerce/woocommerce/pulls) for the same update/change.
-   I have reviewed my code for [security best practices](https://developer.wordpress.org/apis/security/).
-   Following the above guidelines will result in quick merges and clear and detailed feedback when appropriate.

### Changes proposed in this Pull Request:

Add product_brand Term to Product Single HTML template.

### How to test the changes in this Pull Request:

1. Go to Products > Brands and create a brand. And assign it to a product.
2. Activate one FSE theme like TwentyTwentyThree.
3. Go to Appearance -> Editor. Click on the WP logo and go to templates. Edit the WooCommerce Single Product template
4. Confirm you can see the brands block as follow:

https://github.com/user-attachments/assets/a75de481-97cb-4b60-9f23-6f9efc919984

5.  View a product with a brand taxonomy assigned and confirm the brand taxonomy is displayed correctly.
6. Switch to a non-FSE theme, and confirm Brands display correctly on the page.

##### Test `woocommerce_product_brands_output` filter.

7. Switch to no-FSE theme.
8. Add this temporary filter to the brands on the page:

```
add_filter('woocommerce_product_brands_output', '__return_empty_string');
```
9. View a product with a brand taxonomy assigned and confirm that the brand doesn't show.

### Changelog entry

-   [x] Automatically create a changelog entry from the details below.

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Details</summary>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

-   [ ] Fix - Fixes an existing bug
-   [x] Add - Adds functionality
-   [ ] Update - Update existing functionality
-   [ ] Dev - Development related task
-   [ ] Tweak - A minor adjustment to the codebase
-   [ ] Performance - Address performance issues
-   [ ] Enhancement - Improvement to existing functionality

#### Message 
Add product_brand Term to Product Single HTML template.
</details>

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

PR #55936, opened 2025-02-27
Update: Use Block Hooks API to dynamically inject brand terms instead of hardcoding them in the single product template.

rawdreeg:

This PR addresses a backward compatibility issue with WooCommerce Brands in block themes introduced by #55298. Previously, the approach was to add a post-terms block for product brands directly in the single product template, which inadvertently broke backward compatibility for users with customized templates but still relied on the PHP-based brand output.

p1740661397355949-slack-C02FL3X7KR6

#### Testing notes
- Test with both classic and block themes
- Verify brands appear correctly in product meta section in block themes
- Confirm PHP-based output still works correctly in classic themes
- Test with customized templates to ensure backward compatibility is maintained
- This approach ensures that WooCommerce Brands works reliably across different theme types while following WordPress best practices for block integration.

#### Steps to test

- Go to Products > Brands and create a brand. And assign it to a product.
- Activate one FSE theme like TwentyTwentyFive.
- Go to Appearance -> Editor. Click on the WP logo and go to templates. Edit the WooCommerce Single Product template
- Confirm you can see the brands block as follow:
Screen Recording 2025-02-27 at 20 00 51

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Use Block Hooks API to dynamically inject brand terms instead of hardcoding them in the single product template.
</details>

<details>

#### Comment 

</details>


plugins/woocommerce/includes/class-wc-brands.php
```diff
@@ -396,9 +400,7 @@ public function brand_description() {
 	public function show_brand() {
 		global $post;
 
-		$has_block_template = apply_filters( 'woocommerce_has_block_template', false, 'taxonomy-product_brand' ); // phpcs:ignore WooCommerce.Commenting.CommentHooks.MissingHookComment
-
-		if ( is_singular( 'product' ) && ! $has_block_template ) {
+		if ( is_singular( 'product' ) && ! wc_current_theme_is_fse_theme() ) {
```
Aljullu: Heads-up that a theme might be a block theme (so `wc_current_theme_is_fse_theme()` returns `true`) but we might be rendering the _Classic Template_ block, which in the end is the same as the PHP template.

So I think it would be better t o remove the check from here:

```suggestion
		if ( is_singular( 'product' ) ) {
```

And instead add:

```PHP
remove_action( 'woocommerce_product_meta_end', array( $this, 'show_brand' ) );
```

When we add the hooked block.

Aljullu (commented):
Thanks for working on this, @rawdreeg! I took a look at the code and tested the PR and overall it's working well on my end.

I noticed the Brands taxonomy is added outside of the _Row_. Unfortunately, I think the block hooks API doesn't allow us to add it inside the _Row_. In any case, that's much better than what we had until now:

WC 9.7 | This branch
--- | ---
imatge | imatge

cc @tjcafferkey as he worked on the block hooks API in case he has an idea. :pray: 

---

One issue we will have is with stores that:

1. Have removed the PHP hook to hide the Brands that were displayed in WC 9.7.
2. Have edited the _Single Product_ template to add the _Brands_ taxonomy block:

imatge

In that case, when they update, they will end up with two blocks showing the brands:

imatge

I can't think of any way to avoid that, but wanted to highlight it just in case.

---

I also left one comment inline about accounting for the case that a block theme might be using the _Classic Template_ block. :point_down:

plugins/woocommerce/includes/class-wc-brands.php
rawdreeg: Nice catch. Implemented in edf3e136a648a89995de79e6671aa5d96245fb6a

rawdreeg:
Thanks for having a look @Aljullu.

> I noticed the Brands taxonomy is added outside of the Row. Unfortunately, I think the block hooks API doesn't allow us to add it inside the Row. In any case, that's much better than what we had until now:

Yes. This is the cleanest I could do it. 

> One issue we will have is with stores that:
> 
> Have removed the PHP hook to hide the Brands that were displayed in WC 9.7.
> Have edited the Single Product template to add the Brands taxonomy block:
> I can't think of any way to avoid that, but wanted to highlight it just in case.

Thanks for highlighting. There are many ways the user could be hiding the brand before this ( including CSS ), so it'd be hard to cover all cases.

plugins/woocommerce/includes/class-wc-brands.php
```diff
@@ -1093,6 +1096,64 @@ public function wc_brands_on_block_template_register( $template_id, $template_ar
...
+		// Only add the block as the last child of the product meta block
+		if ( 'woocommerce/product-meta' === $anchor_block_type &&
+			 'last_child' === $relative_position &&
+			 $context instanceof WP_Block_Template &&
+			 'single-product' === $context->slug ) {
+
+				remove_action( 'woocommerce_product_meta_end', array( $this, 'show_brand' ) );
+
+				// Check if the template already has a product brand block
+				if ( ! $this->template_already_has_brand_block( $context ) ) {
```
rawdreeg, line 1120: @Aljullu 

One issue we will have is with stores that:

> Have removed the PHP hook to hide the Brands that were displayed in WC 9.7.
> Have edited the Single Product template to add the Brands taxonomy block:
> I can't think of any way to avoid that, but wanted to highlight it just in case.

I added this bit to cover the case you described above. If a user already has the brand terms in the block, I leave it alone.

plugins/woocommerce/includes/class-wc-brands.php
```diff
@@ -1093,6 +1095,80 @@ public function wc_brands_on_block_template_register( $template_id, $template_ar
...
+	 * @param WP_Block_Template $template The template object.
+	 * 
+	 * @return boolean True if template contains a brand block.
+	 */
+	private function template_already_has_brand_block( $template ) {
+		if ( ! $template || empty( $template->content ) ) {
+			return false;
+		}
+
+		return strpos( $template->content, '"term":"product_brand"' ) !== false;
```
Aljullu: I think this check is too broad and might return false positives: any block with an attribute named `term` with value `product_brand` would return true.

Could we make it more specific? I'm thinking about something along these lines:

```suggestion
		return strpos( $template->content, '<!-- wp:post-terms {"term":"product_brand"' ) !== false;
```

Aljullu (commented):
Thanks for the updates, @rawdreeg! I added one additional comment.

plugins/woocommerce/includes/class-wc-brands.php
rawdreeg: Thanks @Aljullu. Nice catch! I committed the suggestion, restricting the check to post-term blocks.

rawdreeg:
> Thanks for the updates, @rawdreeg! I added one additional comment.

Thanks @Aljullu. Nice catch! I committed to the suggestion of restricting the check to post-term blocks.

plugins/woocommerce/includes/class-wc-brands.php
```diff
@@ -1093,6 +1095,80 @@ public function wc_brands_on_block_template_register( $template_id, $template_ar
...
+
+		if ( 'core/post-terms' === $hooked_block_type &&
+			 'last_child' === $relative_position &&
+			 'woocommerce/product-meta' === $parsed_anchor_block['blockName'] ) {
+
+			$parsed_hooked_block['attrs'] = array(
+				'term'	 => 'product_brand',
+				'prefix' => __( 'Brands: ', 'woocommerce' ),
+			);
+		}
```
Aljullu: Not sure what are the best practices with block hooks, but wondering if we should make sure `attrs` is empty before setting them to make sure we don't override an existing block? I'm not sure if that's even a problem, though, considering we use the `hooked_block_core/` hook. :thinking: 

cc @tjcafferkey in case you have a preference.

Aljullu (approved):
LGTM, thanks for working on this, @rawdreeg! :raised_hands: 

I see there are some tests failing, I have re-run them in case they are flaky ones. Besides that, I left one last question inline, but it's a question, not a change request, so feel free to ignore. :smile: 

I would appreciate a second review from @tjcafferkey given that he knows block hooks much better than anyone else.

plugins/woocommerce/includes/class-wc-brands.php
rawdreeg: Thanks @Aljullu. I think it doesn't hurt to add an extra check: [901ed79](https://github.com/woocommerce/woocommerce/pull/55936/commits/901ed798a49a37279dcdd85dbd4bed6299174c23)

tjcafferkey: >Not sure what are the best practices with block hooks, but wondering if we should make sure attrs is empty before setting them to make sure we don't override an existing block?

This will never be an existing block. This filter only gets run for blocks which the API is hooking into the template. Once the block is updated or moved its no longer a hooked block. Adding the check is fine though and shows intent either way.

tjcafferkey (approved):
I've got nothing to add here (which is good news!), straight forward use case and looks good to me! Thanks for pinging though @Aljullu and @rawdreeg  🙌🏻  always happy to review such cases.

PR #55940, opened 2025-02-27
Add/email editor sending transactional emails

### LINKED ISSUE #55674: Add woo transactional content into email(s) managed via block editor and ensure they are sent
We need to figure out the best way how to include the content of Woo transactional emails in the email content.
The ultimate solution would be a set of configurable blocks, but given the complexity of the ideal solution and time constraints, we need to figure out a temporary solution.

### Requirements

1. An admin can see the Woo email content with dummy data in the editor
3. When sending, the content is replaced with the real data

### LINKED ISSUE #55801: Use the email editor for a transactional email(s)
This ticket aims to orchestrate the email editor and the sending part to send any transactional email.  We don't want to perform any significant UI changes as part of this ticket.  Those will come in later stages. This is only to ensure the editor can be loaded, the email can be composed, and it is sent.

1. Pick any email type, for example, the reset password email
2. Make sure users can open the email and edit the header and footer; the content part of the email will be loaded from the old email notification.
3. Make sure all styles can be edited and the email template can be edited
4. Make sure the email stores properly and all email parts are saved, including styles and template
5. Once the users trigger the action (in this example, they trigger the reset password), ensure the email is rendered and all the important parts (reset password link) are sent.

costasovo:

Closes #55674, Closes #55801.

This PR adds the initial implementation of replacing Woo transactional emails with emails created in the email block editor.

I've provided more details in a GH discussion: https://github.com/woocommerce/woocommerce/discussions/55965

**Do not merge immediately. I would like to get feedback on the approach**

There will be follow-up tasks. Also mentioned in  https://github.com/woocommerce/woocommerce/discussions/55965

Screenshot 2025-02-28 at 9 25 55
Screenshot 2025-02-28 at 9 30 50

### Key Changes

- Added simple placeholder block for WooCommerce content
- Transactional emails that are associated with an email created in the block editor are replaced by the block based emails  during the sending
- The WooCommerce content placeholder is replaced by a content captured from the associated WC_Email during the 
sending
- Currently supported personalized tags are also processed and replaced in the sent email

1. Enable the `block_email_editor` experimental flag (in DB, set `woocommerce_feature_block_email_editor_enabled` option to yes).
2. Navigate to WP Admin > Woo Emails
3. Create or edit an email template
4. Use an id of WC_Email you want to replace as the email title/campaign name. This is editable in the component in the center of the header in the editor. The list of WC_Emails can be found [here](https://github.com/woocommerce/woocommerce/tree/trunk/plugins/woocommerce/includes/emails) and you can grab the id from the constructor ([example](https://github.com/woocommerce/woocommerce/blob/65be92caa43dcad3c61b2b1319a2116e748db35b/plugins/woocommerce/includes/emails/class-wc-email-customer-processing-order.php#L30)). This is a temporary way of setting the association.
5. Do the action that triggers the email. For example, for `customer_processing_order` you need to create a new order.
6. You should receive the email created in the block editor and it should contain the main content from the WC_Email. Note: you can use [WP Mail Logging plugin](https://wordpress.org/plugins/wp-mail-logging/) to capture the emails

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/add-email-editor-sending-transactional-emails
+++ b/plugins/woocommerce/changelog/add-email-editor-sending-transactional-emails
@@ -1,0 +1,4 @@
+Significance: minor
+Type: add
+
+Add the base functionality for replacing Woo transactional emails with emails created via the block email editor
--- a/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/constants.ts
+++ b/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/constants.ts
@@ -1,0 +1,1 @@
+export const NAME_SPACE = 'woocommerce/email-editor-integration';
--- a/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/global.d.ts
+++ b/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/global.d.ts
@@ -1,5 +1,5 @@
 interface Window {
-	MailPoetEmailEditor: {
+	WooCommerceEmailEditor: {
 		current_post_type: string;
 		current_post_id: string;
 		email_types: {
--- a/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/global.d.ts
+++ b/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/global.d.ts
@@ -1,0 +1,10 @@
+interface Window {
+	MailPoetEmailEditor: {
+		current_post_type: string;
+		current_post_id: string;
+		email_types: {
+			value: string;
+			label: string;
+		}[];
+	};
+}
--- a/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/index.ts
+++ b/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/index.ts
@@ -10,8 +10,8 @@ import { registerBlockType } from '@wordpress/blocks';
  * Internal dependencies
  */
 import { wooContentBlock } from './woo_content_placeholder_block';
-
-const NAME_SPACE = 'woocommerce/email-editor-integration';
+import { NAME_SPACE } from './constants';
+import { modifySidebar } from './sidebar_extension';
 
 addFilter(
 	'woocommerce_email_editor_send_button_label',
@@ -20,3 +20,4 @@ addFilter(
 );
 
 registerBlockType( 'woo/email-content', wooContentBlock );
+modifySidebar();
--- a/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/index.ts
+++ b/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/index.ts
@@ -4,6 +4,12 @@
  * External dependencies
  */
 import { addFilter } from '@wordpress/hooks';
+import { registerBlockType } from '@wordpress/blocks';
+
+/**
+ * Internal dependencies
+ */
+import { wooContentBlock } from './woo_content_placeholder_block';
 
 const NAME_SPACE = 'woocommerce/email-editor-integration';
 
@@ -12,3 +18,5 @@ addFilter(
 	NAME_SPACE,
 	() => 'Save WooCommerce email template' // This is a temporary label to confirm the integration works, it will be updated in the future.
 );
+
+registerBlockType( 'woo/email-content', wooContentBlock );
--- a/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/sidebar_extension.tsx
+++ b/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/sidebar_extension.tsx
@@ -1,0 +1,55 @@
+/**
+ * External dependencies
+ */
+import { addFilter } from '@wordpress/hooks';
+import { SelectControl } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import {
+	useEntityRecord,
+	store as coreDataStore,
+	Post,
+} from '@wordpress/core-data';
+import { useDispatch } from '@wordpress/data';
+import { EntityRecordResolution } from '@wordpress/core-data/build-types/hooks/use-entity-record';
+
+/**
+ * Internal dependencies
+ */
+import { NAME_SPACE } from './constants';
+
+const SidebarExtensionComponent = () => {
+	const { current_post_id, current_post_type, email_types } =
+		window.MailPoetEmailEditor;
+	const email = useEntityRecord(
+		'postType',
+		current_post_type,
+		current_post_id
+	) as EntityRecordResolution< Post >;
+	const { editEntityRecord } = useDispatch( coreDataStore );
+	const email_type_options = [ { value: '', label: '---' }, ...email_types ];
+
+	return (
+		<SelectControl
+			label={ __( 'Email type', 'woocommerce' ) }
+			options={ email_type_options }
+			value={ email.editedRecord.slug }
+			onChange={ ( value ) => {
+				editEntityRecord(
+					'postType',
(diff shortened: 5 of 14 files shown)


triple0t (commented):
I tested the PR according to the testing notes, and it works well. Well done.

Editor
Screenshot 2025-03-03 at 12 52 25 PM

Email content
Screenshot 2025-03-03 at 2 19 00 PM
Screenshot 2025-03-03 at 2 19 06 PM

I also confirmed that the Global styles settings are applied
Screenshot 2025-03-03 at 2 37 27 PM


I recognize this is an initial draft, so I have some feedback.
I don't have any complaints about the PR approach.

1. Can we register a [select](https://developer.wordpress.org/block-editor/reference-guides/components/select-control/) element (with Woo email templates ID) on the email settings sidebar and allow users to pick the template instead of using the email title/campaign name field to indicate the WooCommerce transaction email template?
2. We can simplify [this section](https://github.com/woocommerce/woocommerce/blob/4325d84486a58a3ab2930f057fccb335666a4538/plugins/woocommerce/src/Internal/EmailEditor/Integration.php#L72-L82) by moving the `block_email_renderer->register_hooks` method to an `init` function.
3. We need to create an issue for adding `Order` and `User` related personalization tags.
4. When the block email editor is enabled for a transactional email, I suggested we remove [these other string contents](https://github.com/user-attachments/assets/9f53d56f-f37b-421c-83dc-087e3ecf9b3f) from the email template and pre-populate them on the block email editor content. Note: We don't need to implement the suggestion in this PR.

costasovo:
@triple0t Thanks for the in-depth review and the suggestions.

> Can we register a [select](https://developer.wordpress.org/block-editor/reference-guides/components/select-control/) element (with Woo email templates ID) on the email settings sidebar and allow users to pick the template instead of using the email title/campaign name field to indicate the WooCommerce transaction email template?

I added it in [d34d74c](https://github.com/woocommerce/woocommerce/pull/55940/commits/d34d74cbfeb6b75919c61db40bbbd9cfc360a475). The value is saved in `wp_post` in the `slug` column. This select component is a temporary solution. I assume that in the end, we end up with something like a pre-generated set of emails and we won't allow users to switch. But it is handy in the meantime.

> We can simplify [this section](https://github.com/woocommerce/woocommerce/blob/4325d84486a58a3ab2930f057fccb335666a4538/plugins/woocommerce/src/Internal/EmailEditor/Integration.php#L72-L82) by moving the block_email_renderer->register_hooks method to an init function.

Done in [0125df8](https://github.com/woocommerce/woocommerce/pull/55940/commits/0125df85402c39024910b268fa114d9f5aeba74a)

> We need to create an issue for adding Order and User related personalization tags.

I created a draft for the issue https://github.com/orgs/woocommerce/projects/297/views/1?pane=issue&itemId=100275855

> When the block email editor is enabled for a transactional email, I suggested we remove [these other string contents](https://github.com/user-attachments/assets/9f53d56f-f37b-421c-83dc-087e3ecf9b3f) from the email template and pre-populate them on the block email editor content. Note: We don't need to implement the suggestion in this PR.

At this point I was looking for a general solution that would work also for the third party emails. I think that in the next steps we will focus on the core's transactional emails one by one. A possible approach for removing the unwanted parts of the content could be adding a condition to the current template and skipping it (something similar [to this](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/templates/emails/customer-processing-order.php#L42-L49)). I would do that in a follow-up task. 

I would like to get this PR merged. There are a few follow-up tasks that we could work on. But I think we should first merge the [PR with namespace refactor](https://github.com/woocommerce/woocommerce/pull/55938) to avoid conflicts.

triple0t:
>I added it in [d34d74c](https://github.com/woocommerce/woocommerce/pull/55940/commits/d34d74cbfeb6b75919c61db40bbbd9cfc360a475). The value is saved in wp_post in the slug column. This select component is a temporary solution. I assume that in the end, we end up with something like a pre-generated set of emails and we won't allow users to switch. But it is handy in the meantime.

You might need to update the [query](https://github.com/woocommerce/woocommerce/blob/d34d74cbfeb6b75919c61db40bbbd9cfc360a475/plugins/woocommerce/src/Internal/EmailEditor/BlockEmailRenderer.php#L91-L96) here. The Email editor template is not used when selecting the transaction email from the select component.

> I created a draft for the issue https://github.com/orgs/woocommerce/projects/297/views/1?pane=issue&itemId=100275855
> 
> At this point I was looking for a general solution that would work also for the third party emails. I think that in the next steps we will focus on the core's transactional emails one by one. A possible approach for removing the unwanted parts of the content could be adding a condition to the current template and skipping it (something similar [to this](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/templates/emails/customer-processing-order.php#L42-L49)). I would do that in a follow-up task.

Thank you.

> I would like to get this PR merged. There are a few follow-up tasks that we could work on. But I think we should first merge the https://github.com/woocommerce/woocommerce/pull/55938 to avoid conflicts.

Merged. Please rebase.

costasovo:
@triple0t Thanks for merging the PR and checking the changes.

> You might need to update the [query](https://github.com/woocommerce/woocommerce/blob/d34d74cbfeb6b75919c61db40bbbd9cfc360a475/plugins/woocommerce/src/Internal/EmailEditor/BlockEmailRenderer.php#L91-L96) here. The Email editor template is not used when selecting the transaction email from the select component.

Great catch! Thank you! It worked for me because I still had the title changed in the DB. Fixed in [e1ed8b0](https://github.com/woocommerce/woocommerce/pull/55940/commits/e1ed8b09e02c099dac070ada3ce53fc28a207431)

I've updated namespaces after rebase in [a2e4bd5](https://github.com/woocommerce/woocommerce/pull/55940/commits/a2e4bd549e33bf6fefeac4b655747b59d4b3cb8e)

I ran into a case when I defied block based email that is set as `plain` and it sent HTML code as plain/text. I've addressed that in [455c29f](https://github.com/woocommerce/woocommerce/pull/55940/commits/455c29f91e624ca668bb19e2774757d5ebc55f29)

triple0t (approved):
> I ran into a case when I defied block based email that is set as plain and it sent HTML code as plain/text. I've addressed that in [455c29f](https://github.com/woocommerce/woocommerce/pull/55940/commits/455c29f91e624ca668bb19e2774757d5ebc55f29)

Instead of hardcoding it [here](https://github.com/woocommerce/woocommerce/blob/455c29f91e624ca668bb19e2774757d5ebc55f29/plugins/woocommerce/includes/emails/class-wc-email.php#L721), I would prefer we respect user preference and send `plain/text` when requested and `HTML` when requested (we can default to `HTML`).

The Block editor renderer [returns](https://github.com/woocommerce/woocommerce/blob/455c29f91e624ca668bb19e2774757d5ebc55f29/plugins/woocommerce/src/Internal/EmailEditor/BlockEmailRenderer.php#L75) both HTML and TEXT. We can pass the user preference across and [return](https://github.com/woocommerce/woocommerce/blob/455c29f91e624ca668bb19e2774757d5ebc55f29/plugins/woocommerce/src/Internal/EmailEditor/BlockEmailRenderer.php#L76-L78) the requested email data.

I'm approving the PR. **Thank you, and well done**.
You can merge to trunk when you're ready.

costasovo:
> Instead of hardcoding it [here](https://github.com/woocommerce/woocommerce/blob/455c29f91e624ca668bb19e2774757d5ebc55f29/plugins/woocommerce/includes/emails/class-wc-email.php#L721), I would prefer we respect user preference and send plain/text when requested and HTML when requested (we can default to HTML).

@triple0t Great point! Thinking about it, I don't remember seeing a plain text option in the designs with the block editor. I as for now I assume that block based emails will always be sent as HTML. Also, this seems to be easy to change, so I'll leave it as is for now.

Thank you for the thorough review and valuable remarks! Much appreciated!

PR #55949, opened 2025-02-28
Remove @woocommerce/interactivity JS scripts and PHP code. Migrate remaining blocks.

samueljseay:

This removes `@woocommerce/interactivity` entirely from the monorepo.

**Also** it seems that several experimental blocks flew under our radar by accident. None of these need changes in the JavaScript apart from to change the import and build them as script modules. This includes:

* accordion group blocks
* add to cart with options blocks

1. See CI pass
2. Smoke test, check frontend blocks, editor side blocks, (mini cart, shop page, filters) check for console errors. 

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Remove `@woocommerce/interactivity` JS scripts and PHP code. Migrate remaining blocks

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/legacy/js/frontend/add-to-cart.js
```diff
@@ -15,7 +15,7 @@ jQuery( function( $ ) {
 		this.$liveRegion = this.createLiveRegion();
 
 		$( document.body )
-			.on( 'click', '.add_to_cart_button:not(.wc-interactive)', { addToCartHandler: this }, this.onAddToCart )
+			.on( 'click', '.add_to_cart_button:not(.wp-interactive)', { addToCartHandler: this }, this.onAddToCart )
```
samueljseay: Hmm it think I better leave this as it was.

samueljseay:
@louwie17 pulled you in to check accordion group if you don't mind? I have done some manual testing and it seems to be working ok.

e2e failures I don't think are related but still investigating.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartForm.php
```diff
@@ -162,6 +162,9 @@ protected function render( $attributes, $content, $block ) {
 
 			return '';
 		}
+
+		wp_enqueue_script_module( $this->get_full_block_name() );
```
Aljullu, line 166: This is required by all blocks whose frontend script is generated as a script module?

plugins/woocommerce-blocks/bin/webpack-entries.js
```diff
@@ -266,6 +266,12 @@ const frontendScriptModuleBlocksToSkip = [
 	'product-filter-rating',
 	'product-filter-active',
 	'product-filter-removable-chips',
+	'add-to-cart-form',
+	'add-to-cart-with-options',
+	'add-to-cart-with-options-quantity-selector',
+	'add-to-cart-with-options-variation-selector',
+	'add-to-cart-with-options-grouped-product-selector',
+	'add-to-cart-with-options-grouped-product-selector-item',
```
Aljullu: No need to do in this PR, but I wonder if we could add this info into the `blocks` const (ie, adding a property to blocks that use iAPI), so we don't need to keep this list and the one up in `webpack-config-interactivity-blocks-frontend.js` up to date.

Aljullu: Also, out of curiosity, why isn't the accordion block here? :confused:

Aljullu (approved):
I did some smoke testing of the _Add to Cart with Options_ block (blockified and non-blockified), changing quantities in the Stepper mode, adding to cart (only in the non-blockified one), etc. and it seemed to work well in all cases.

image

I'm approving from my side, please note I didn't test the accordion block. I also left some questions inline, they are not change requests, but mostly for me to understand how the iAPI works with our blocks. :sweat_smile:

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartForm.php
samueljseay: @Aljullu Yes that's right. I had considered abstracting this into something run just after the render call, but, In future though, the plan is to support `block.json`'s `viewScriptModule` so i didn't spend time on that. instead we enqueue the script module as late as possible.

plugins/woocommerce-blocks/bin/webpack-entries.js
samueljseay: > No need to do in this PR, but I wonder if we could add this info into the blocks const (ie, adding a property to blocks that use iAPI),

That's one way we could approach it, but I believe in future we could probably greatly simplify webpack configs and builds from how they're setup now, which is why I kept this separate. A lot of the complexity exists because of how they're managed by AssetsController but we will in future use `block.json` and I believe we'll also be able to move to a more vanilla webpack build that's closer to Gutenbergs.

samueljseay: > Also, out of curiosity, why isn't the accordion block here? 😕

Whoops look like I missed that, thanks for spotting.

louwie17 (approved):
> @louwie17 pulled you in to check accordion group if you don't mind? I have done some manual testing and it seems to be working ok.

Thanks I appreciate you updating this right away as well. This tested well on my end and changes look good, so I am approving for the Accordion block work in particular :)

PR #55968, opened 2025-02-28
[Block Checkout] Fix wrong shipping address being posted during checkout when using local pickup

mikejolley:

When placing an order via Local Pickup, the wrong address was being sent to the server. This could result in either the incorrect values being persisted to the order, or validation errors which are unresolvable.

To reproduce:

1. Make sure both Local Pickup and regular Shipping Methods are enabled and checkout in a new incognito window
2. Select "shipping" in the top most method selector
3. Enter a _partial_ shipping address. Use UK country and leave postcode blank.
4. Untoggle use shipping as billing address
5. Enter a _full_ US billing address.
6. Now toggle to "local pickup". When you do this, the shipping section is hidden and only the billing section remains.
7. Place order. See validation error due to missing postcode. 

This PR resolves the above issue and some TS warnings in the touched file.

1. Make sure both Local Pickup and regular Shipping Methods are enabled and checkout in a new incognito window
2. Select "shipping" in the top most method selector
3. Enter a _partial_ shipping address. Use UK country and leave postcode blank.
4. Untoggle use shipping as billing address
5. Enter a _full_ US billing address.
6. Now toggle to "local pickup". When you do this, the shipping section is hidden and only the billing section remains.
7. Place order successfully.
8. View the order in the admin area. Confirm the shipping and billing addresses are the same.

Just the above steps and smoke testing placing orders.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce-blocks/assets/js/events/event-emitter.ts
```diff
@@ -11,7 +11,7 @@ import {
 
 export type EventListener = (
 	data: unknown
-) => Promise< ObserverResponse | true > | ObserverResponse | true;
+) => Promise< ObserverResponse | boolean > | ObserverResponse | boolean;
```
alexflorisca: Is this because event listener callbacks can return `false` as well? I couldn't find an example of this though, is there anywhere you found this to be true?

alexflorisca (commented):
Sorry it took so long to get to this. 

I followed your testing instructions and everything was ok. Until I decided to change my payment method before I placed the order, then all the address data disappeared! This is a bug with the persisting fields, not your PR though, I've raised it here https://github.com/woocommerce/woocommerce/issues/56155. Partial addresses aren't persisted, so if you enter a partial address, then update another field, the PUT endpoint will return a new cart which overrides the data on the client. 

What is strange to me, and maybe relevant to this PR, is that even though the shipping address is partial and doesn't get persisted, when I check "use same address for billing" and enter a full billing address, this isn't persisted either. I would expect it to be. Either way the logic around that needs to be revisited, but we can have that conversation in the other issue.

For this PR, I think it's ok to go ahead. I smoke tested placing order, changing addresses a bunch of time with a mixture of full and partial, and the outcome is always the same.

Code looks good, just one non-blocking comment.

You'll need to solve merge conflicts as woo blocks has moved inside woocommerce as well.

plugins/woocommerce-blocks/assets/js/events/event-emitter.ts
mikejolley: There was an error case that would fallback to `false` without generating a WP_Error. I think it was in `CheckoutProcessor`. 

Edit: yeah, `checkValidation`. It returns false or true.

mikejolley:
Thanks @alexflorisca. Partial updates don't seem related to this change so agree they need to be looked into.

I cannot merge without approval however.

alexflorisca (approved):

PR #55978, opened 2025-02-28
Fix rest API endpoint ( payment_gateway and tax_class)

lbonomo:

1. Implemented a conditional in the CLI to prevent the `Warning: Undefined array key "X-WP-Total"` error on endpoints that do not implement the `count` output format. If the `count` format is requested on an endpoint where it is not implemented, display the message: `Error: Count format not implemented yet.`

2. Implemented a `count` output format for `Payment gateway` and `tax classes` endpoints.

#### Current error
```bash
$ wp wc [payment_gateway|tax_class] list --user=admin --format=count
Warning: Undefined array key "X-WP-Total" in /var/www/html/wp-content/plugins/woocommerce/includes/cli/class-wc-cli-rest-command.php on line 239
```

#### CLI
**Requirement**: To run the following test, you need to have data in your WooCommerce installation. If you don't have data, you can use this command to install the `Smooth Generator` plugin and generate `products` and `orders`.

```bash
wp plugin install https://github.com/woocommerce/wc-smooth-generator/releases/download/1.2.1/wc-smooth-generator.zip --activate
wp wc generate orders 10 
wp wc generate products 10 --type=simple
```
*Test*
```bash
wp wc order_note list $(wp wc shop_order list --user=1 --format=ids --per_page=1) --user=1 --format=count
Error: Count format not implemented yet.
```
Or

```bash
wp wc product_review list $(wp wc product list --user=1 --per_page=1 --format=ids) --user=1 --format=count
Error: Count format not implemented yet.
```

#### Endpoint
1. Run the following commands using WP-CLI. No error should be logged or printed while doing so.
   ``` 
   wp> wp wc payment_gateway list --user=admin --format=count
   wp> wp wc tax_class list --user=admin --format=count
   ```
   In particular, error `Warning: Undefined array key "X-WP-Total"` should not appear.
2. (Optional)
   1. Use the request software of your choice to make GET requests to:
      - `<yoursite>/wp-json/wc/v3/payment_gateways/`
      - `<yoursite>/wp-json/wc/v3/taxes/classes/`
      Make sure the requests work correctly and that, in all cases, headers `X-WP-Total` (reflecting the number of items returned) and `X-WP-TotalPages` are set.

Captura de pantalla de 2025-02-28 12-06-10
Captura de pantalla de 2025-02-28 12-09-22

### DIFF
--- a/plugins/woocommerce/changelog/fix-cli-and-endpoints
+++ b/plugins/woocommerce/changelog/fix-cli-and-endpoints
@@ -1,0 +1,6 @@
+Significance: minor
+Type: fix
+
+- Adding header validation to prevent this error:
+`PHP Warning:  Undefined array key "X-WP-Total" in /app/wordpress/wp-content/plugins/woocommerce/includes/cli/class-wc-cli-rest-command.php on line 239`
+- Adding `X-WP-Total` and `X-WP-TotalPages` header on `tax_class` and `payment_gateway` endpoints.
--- a/plugins/woocommerce/changelog/fix-cli-and-endpoints
+++ b/plugins/woocommerce/changelog/fix-cli-and-endpoints
@@ -1,6 +1,4 @@
 Significance: minor
 Type: fix
 
-- Adding header validation to prevent this error:
-`PHP Warning:  Undefined array key "X-WP-Total" in /app/wordpress/wp-content/plugins/woocommerce/includes/cli/class-wc-cli-rest-command.php on line 239`
-- Adding `X-WP-Total` and `X-WP-TotalPages` header on `tax_class` and `payment_gateway` endpoints.
+Add validation to prevent PHP warning in CLI when using the 'count' format.
--- a/plugins/woocommerce/changelog/55978-fix-endpoints
+++ b/plugins/woocommerce/changelog/55978-fix-endpoints
@@ -1,0 +1,4 @@
+Significance: minor
+Type: fix
+
+Add support to `count` output format on `tax_class` and `payment_gateway` endpoints.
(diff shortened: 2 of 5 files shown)


plugins/woocommerce/changelog/fix-cli-and-endpoints
```diff
@@ -0,0 +1,6 @@
+Significance: minor
+Type: fix
+
+- Adding header validation to prevent this error:
+`PHP Warning:  Undefined array key "X-WP-Total" in /app/wordpress/wp-content/plugins/woocommerce/includes/cli/class-wc-cli-rest-command.php on line 239`
+- Adding `X-WP-Total` and `X-WP-TotalPages` header on `tax_class` and `payment_gateway` endpoints.
```
jorgeatorres: Changelogs are compiled into the final readme.txt so ideally they would be short. I'm offering up a suggestion but feel free to adapt as you see fit.

```suggestion
Add validation to prevent PHP warning in CLI when using the 'count' format.
```

Given this PR also takes care of adding support for some endpoints, I'd suggest you add another changelog file for the other changes. You can use the same `pnpm --filter='@woocommerce/plugin-woocommerce' changelog add` tool to create a new one.

jorgeatorres (changes requested):
Hi @lbonomo,

Thanks for your contribution 🥇. Overall this looks good to me and tests well. I just left some tiny feedback related to the changelog file.

Also, could you please rework your testing instructions a bit so that we can better illustrate the changes. For example, this PR adds support for `count` to both the `payment_gateway` and `tax_class` commands, so those wouldn't result in an error _even_ if we were using the old code.

As such I'd suggest you keep those steps in the instructions, but as a way to prove that the new endpoints have support.
And then we can add a new step to test an endpoint that doesn't (for example, `wp wc order_note list <order_id>`) and ask to confirm that in that case the CLI error is output, but no PHP warning is displayed or logged.

Hope that makes sense, but let me know if that's not the case. Once again, thank you for your contribution!

jorgeatorres (approved):
LGTM. Followed the provided instructions and tested both via CLI and REST API.

Thank you for making the changes and for providing clearer instructions @lbonomo!

PR #55979, opened 2025-02-28
[Experimental] Introduce Add to Cart with Options Grouped Product Selector Item CTA block

### LINKED ISSUE #54800: Blockified Add to Cart with Options block: grouped product CTA should render a Quantity Selector, a checkbox or a button depending on the circumstances
When rendering the CTA in a product inside a Grouped Product, we might need to render a quantity selector, a checkbox or a button depending on several conditions. Ie, it should render a checkbox instead of an input field if the product is sold individually:

Image

_Beanie - Limited edition_ has been set to be sold individually, so it renders a checkbox instead of an input field.

For products with options, this should be a button.

Aljullu:

Closes #54800.

This PR updates the Grouped Product Selector block with a new inner block which takes care of displaying the correct CTA for the children products. That might be a numeric input, a checkbox or a button.

This approach has the issue that global styles used in the _Add to Cart Button_ block and _Quantity Selector_ blocks won't be applied to this new block. In the future, we could evaluate if it's worth adding the CSS classes from those blocks into this one. While I'm not a big fan of this approach, I understand this block is a bit unique because it can render three completely different UI elements depending on the product.

imatge

Note: this PR doesn't allow changing the quantity input between _Input_ and _Stepper_ styles. I will create a follow-up issue if this PR gets merged. But I wanted to keep the PR small.

1. Create a simple product, a simple product sold individually, and a simple product out of stock. (You can also test with other product types, but it's not required as they should all be covered)
2. Create a grouped product and add the three products you created in step 1 as children.
3. Create a post and add the _Single Product_ block, select the grouped product you created in step 2. In the editor, verify the correct CTA are displayed for the children: quantity selector for the simple product, checkbox for the simple product sold individually and a button for the product out of stock.
4. Publish the page and visit it in the frontend. Verify the correct CTAs are also displayed in the frontend.
5. Visit the product template in the frontend and verify the correct CTAs are display there too.


mdperez86 (approved):
Nice job @Aljullu!!! This is working as described in the testing instructions.

PR #55570, opened 2025-02-17
[Experimental] Hidden visibility parsing for Store API and additional fields

mikejolley:
### Submission Review Guidelines:

-   I have followed the [WooCommerce Contributing Guidelines](https://github.com/woocommerce/woocommerce/blob/trunk/.github/CONTRIBUTING.md) and the [WordPress Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/).
-   I have checked to ensure there aren't other open [Pull Requests](https://github.com/woocommerce/woocommerce/pulls) for the same update/change.
-   I have reviewed my code for [security best practices](https://developer.wordpress.org/apis/security/).
-   Following the above guidelines will result in quick merges and clear and detailed feedback when appropriate.

### Changes proposed in this Pull Request:

Adds support on the PHP side for `hidden` field rules for checkout fields. Updates and refactors how schema is evaluated for fields.

### How to test the changes in this Pull Request:

_This work is behind a feature flag._ Testers should do a full build to check there are no regressions.

```
pnpm --filter=@woocommerce/plugin-woocommerce build
```

Test with some additional fields with various rules. Example:

<details>

```php
add_action(
	'woocommerce_init',
	function () {
		woocommerce_register_additional_checkout_field(
			array(
				'id'       => 'namespace/vat-number',
				'label'    => 'VAT Number',
				'location' => 'address',
				'required' => true,
				'rules'    => array(
					'hidden'     => array(
						'customer' => array(
							'properties' => array(
								'address' => array(
									'properties' => array(
										'country' => array(
											'type' => 'string',
											'not'  => array(
												'enum' => array_merge( WC()->countries->get_european_union_countries( 'eu_vat' ), array( 'GB' ) ),
											),
										),
									),
								),
							),
						),
					),
					'validation' => array(
						'type'    => 'string',
						'pattern' => '^[A-Z]{2}[0-9A-Z]{2,12}$',
					),
				),
			)
		);
		woocommerce_register_additional_checkout_field(
			array(
				'id'       => 'namespace/reason-for-purchase',
				'label'    => 'Reason for Purchase',
				'location' => 'contact',
				'required' => true,
			)
		);
	}
);
```

</details>

Testing checkout both fields will be required. Order should place and fields should appear under My Account.

For development builds, the feature flag is enabled:

```
pnpm --filter=@woocommerce/plugin-woocommerce watch:build
```

1. With the code snippet above, go to the my account page. Edit an address. If its not already a GB address, add one and save.
[...]

PR #53553, opened 2024-12-09
Persist additional fields and payment method on checkout

alexflorisca:
### Changes proposed in this Pull Request:

This PR persists the additional fields tied to an order, as well as the order notes and payment method on checkout. This means all checkout fields are now persisted, and enables developers to do things like changing the cart totals or adding fees based on any checkout field.

Notable changes

- new `PUT` /checkout endpoint that persists the fields and updates totals. This returns a cart, which is updated client side
- push-changes, similar to the cart store, for the checkout store that sync with server.
- PUT requests get cancelled when clicking Place Order
- Note that saved payment methods are not persisted as part of this PR due to complications with tokens. We can address this in a future PR if necessary.

Closes #42079.

### How to test the changes in this Pull Request:

Using the [WooCommerce Testing Instructions Guide](https://github.com/woocommerce/woocommerce/wiki/Writing-high-quality-testing-instructions), include your detailed testing instructions:

**Pre-testing**
1. Install and activate the [Additional Checkout Fields Tester](https://github.com/woocommerce/additional-checkout-fields-tester) plugin
2. Install and activate the [Stripe Payment Gateway](https://wordpress.org/plugins/woocommerce-gateway-stripe/) plugin
3. Enable the Direct bank transfer, cheque, cash on delivery and stripe payment methods. 
4. Make sure you have a saved payment method available for your user. You can do this by checking out with stripe once, while making sure the checkbox `Save payment information to my account for future purchases.` is ticked.

**Test the Cart page**
1. Add some items to your cart & visit the cart page
2. Make sure there is not PUT request on page load
3. Click on an express payment method and close the modal. There should be no PUT request.

> [!NOTE]
>  The following test scenarios all happen on the Checkout page, unless otherwise mentioned.

**Test the additional fields on checkout**
1. Go to checkout
2. Make sure there is no PUT request on page load
3. Click an express payment method and close the modal. Make sure there are no PUT requests.
5. Fill in each additional field under contact information and additional information
6. Make sure there is a PUT request to the `wc/store/checkout` endpoint for each one
7. Make sure there is no PUT request to `wc/store/checkout`  endpoint when the email is filled in or the create account checkbox is ticked.
8. Refresh the page. The information you entered should still be there.

**Test the order notes**
1. Fill in the order notes and again check for a PUT request.
2. Refresh the page and make sure the data has persisted.
3. Untick the Order notes checkbox and refresh the page again.
4. The order notes checkbox should not be ticked. When ticking the checkbox again, the text field should be empty.

[...]

PR #55989, opened 2025-02-28
[Additional Fields] Fix persistence and partial updates of additional fields

### LINKED ISSUE #56155: [Additional Fields] Partial address overwritten when other fields are persisted
### Prerequisites

### Describe the bug

When a shopper visits the checkout block, enters a partial address, then changes another additional field, order notes or payment method, the address on the client is overwritten with a blank address from the server. This happens because the address (both shipping and billing) aren't persisted until a full address is entered.

Another scenario that demonstrates the same problem.
1. Enter a full shipping address
2. Untick "Use same address for billing
3. Enter a full billing address
4. Delete the postcode for the shipping address
5. Change the payment method
6. See that the billing address has been overwritten with the shipping address.

### Expected behaviour

I would expect the address on the client not to be overwritten. A possible solution would be to check if the address we get back from the server is empty, and only overwrite the address on the client if it's not empty.

### Actual behaviour

The address on the client is overwritten when it is only entered partially.

### Steps to reproduce

1. Add an item to your cart
2. Visit the Checkout page
3. Enter a partial address (e.g. UK address without the postcode)
4. Change payment method, or add something to the order notes.
5. See that the address details entered are gone.

### WordPress Environment

WP 6.7
WC 9.7

### Isolating the problem

alexflorisca:

This PR fixes 4 bugs with the additional fields on checkout and some flakey end to end tests

1. The `woocommerce_store_api_checkout_update_order_from_request` action is not being called on the \PUT endpoint of the checkout.

There was some duplication of the code in `update_order_from_request` function used by the checkout POST route. I've refactored the checkout /PUT route to use the same function and therefore call the `woocommerce_store_api_checkout_update_order_from_request` action.

Bug introduced in PR #53553.

2. Additional fields registered in the contact location were not being persisted.

Needed to get fields for the `contact` location as well as `order` when persisting.

Bug introduced in PR #55570.

3. Required checkboxes persist invalid values

With the work in #53553, when a field has an invalid value, it still gets persisted. With text fields this isn't an issue because the validation is done server side and if they are empty they do not get persisted. But with checkboxes, this isn't the case, and the invalid value was being persisted. This PR addresses that.

4. [Partial address overwritten when other fields are persisted](https://github.com/woocommerce/woocommerce/issues/56155)

Fixes #56155.

When a shopper visits the checkout block, enters a partial address, then changes another additional field, order notes or payment method, the address on the client is overwritten with a blank address from the server. This happens because the address (both shipping and billing) aren't persisted until a full address is entered.

The fix for this was to use `receiveCartContents` rather than `receiveCart` which updates the cart, but not the addresses. We also added calls to `__internalIncrementCalculating()` and `__internalDecrementCalculating` in order for `waitForCheckoutToFinishUpdating()` to return true if the PUT request is still happening. This is used in the checkout e2e tests.

1. Add the code from [this gist](https://gist.github.com/alexflorisca/051e0fe9b0bae22f5c4efdae9c505ba9) to your functions.php, woocommerce.php or add it using a code snippet plugin.
2. Install the [Additional Field Tester plugin](https://github.com/woocommerce/additional-checkout-fields-tester). 
3. Insert an additional required checkbox either with a code snippet plugin or by adding this code to `woocommerce.php` or `functions.php` files:

```
		array(
			'id'       => 'first-plugin-namespace/truck-size-ok',
			'label'    => 'Can a truck fit down your road?',
			'location' => 'address',
			'type'     => 'checkbox',
                         'required' => true,
		)
	);
```

6. Enable the Cash on delivery, cheque and bank transfer payment methods.
7. Add some products to your Cart and visit the Checkout. 
8. Change between the above 3 payment methods, and check that the totals update to add 2, 3 and 5 units of currency to the total.
9. Edit each of the additional fields and wait for a /PUT request to checkout endpoint
10. Refresh the page
11. Check that the value is persisted.
12. Click the required checkbox. There should be a PUT request to the checkout endpoint.
13. Untick the checkbox. There should be no request and you should see an inline error.

**Test that the address doesn't get overridden by the PUT request**
1. Open a new incognito window, add some items to cart and visit the checkout page
2. Fill in the address partially, miss out the postcode.
3. Fill in another additional field and check that a PUT request is successful.
4. The address you entered should still be there.
5. Enter the postcode and any other missing required fields.
6. Check for a request to the /batch endpoint
7. Modify one of the additional fields again.
8. Check that the address is not overwritten.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Fix contact additional fields not persisting, required checkboxes persisting when invalid and `woocommerce_store_api_checkout_update_order_from_request` not triggered on a PUT request to the `/wc/store/v1/checkout` endpoint.

</details>

<details>

#### Comment 

</details>


mikejolley (commented):
I like this approach @alexflorisca I followed your testing steps and tested with 2 required fields, one being a checkbox.

- When the checkbox is invalid, no push ✅ 
- When the checkbox was valid, push ✅ 
- When the checkbox is in an invalid state and I edit the other field, pushes the valid field ✅ 

So I think this looks good.

I do think we need to insert a unit/e2e (?) test to check this functionality—hopefully GPT can write something to test the above case? That way we can avoid another regression.

plugins/woocommerce-blocks/assets/js/data/checkout/utils.ts
```diff
@@ -230,3 +233,36 @@ export const getPaymentResultFromCheckoutResponse = (
...
+		return true;
+	}
+
+	const validationStore = select( VALIDATION_STORE_KEY );
+
+	// Check each additional field for validation errors
+	// The validation store prefixes fields with 'additional-field-'
+	for ( const fieldKey of Object.keys( additionalFields ) ) {
+		// We don't know which location the field is in, so we need to check both
+		let error = validationStore.getValidationError(
```
mikejolley: @senadir is there a better way to detect or know where a field is on client side? I feel like there will be.

mikejolley: @alexflorisca Nadir shared this which can be reused instead of checking for both https://github.com/woocommerce/woocommerce/blob/74c589d6efea4003652e6b1d3eb40291549c9001/plugins/woocommerce-blocks/assets/js/settings/blocks/constants.ts#L124-L155

alexflorisca:
Thanks Mike, just pushed some changes - a test and using the constants file you pointed out. Have another look when you can 👍

plugins/woocommerce-blocks/tests/e2e/tests/checkout/additional-fields.shopper.block_theme.spec.ts
```diff
@@ -908,5 +908,101 @@ test.describe( 'Shopper → Additional Checkout Fields', () => {
 				'This is a nice gift'
 			);
 		} );
+
+		test( 'Shopper can refresh the page, and not loose any entered data', async ( {
```
mikejolley:
```suggestion
		test( 'Shopper can refresh the page, and not lose any entered data', async ( {
```

plugins/woocommerce-blocks/tests/e2e/tests/checkout/additional-fields.shopper.block_theme.spec.ts
```diff
@@ -908,5 +908,101 @@ test.describe( 'Shopper → Additional Checkout Fields', () => {
...
+
+			// Check that contact and order fields are persisted
+			await checkoutPageObject.page.waitForRequest(
+				'**/wc/store/v1/checkout**'
+			);
+
+			// Check that address fields are persisted
+			await checkoutPageObject.page.waitForRequest(
+				'**/wc/store/v1/batch**'
+			);
```
mikejolley: Is it normal for both to be triggered?

mikejolley:
@alexflorisca new test if failing, something about the batch. Left a comment.

plugins/woocommerce-blocks/tests/e2e/tests/checkout/additional-fields.shopper.block_theme.spec.ts
alexflorisca: We're filling in the address as well, so yes. I thought it'd be good to check that address fields also get persisted.

plugins/woocommerce-blocks/assets/js/atomic/blocks/product-elements/stock-indicator/test/block.test.tsx
```diff
@@ -58,6 +58,10 @@ jest.mock( '@woocommerce/settings', () => ( {
 				simple: 'Simple product',
 			};
 		}
+		if ( param === 'globalPaymentMethods' ) {
```
alexflorisca, line 61: `assets/js/settings/blocks/constants.ts` is loaded as part of this test (via a long stacktrace) now that we are importing constants to use in the `validateAdditionalFields` function. This test mocks `getSetting`, but it doesn't mock some of the keys used in the constants file, so the test fails. Added these extra mocks here

mikejolley (approved):
Did another run through testing instructions and I see CI is passing now so this should be ok to merge.

I would however suggest bulking out the testing instructions to include some additional fields so that other people not familiar with additonal field registration can test this PR.

Thanks!

alexflorisca:
Thanks Mike, edited the instructions with some code examples, good point!

alexflorisca:
This commit https://github.com/woocommerce/woocommerce/pull/55989/commits/8c453683155848a1452b299c4adce891e531a3df is https://github.com/woocommerce/woocommerce/pull/56225 merged to this branch. It fixes https://github.com/woocommerce/woocommerce/issues/56155 and some flakey e2e tests

mikejolley (approved):
Overall I'm happy with the changes and I've gone through the testing steps a final time and it appears to be functional. I've tested that fields are no longer overwritten and watched console network logs to ensure thats working ok.

I think so long as CI passes (I pushed a small tweak) this should be good to ship. @senadir do you want to do a final check since this has grown in scope and we've collaborated on this?

plugins/woocommerce-blocks/assets/js/data/checkout/utils.ts
```diff
@@ -230,3 +237,37 @@ export const getPaymentResultFromCheckoutResponse = (
...
+	// Check each additional field for validation errors
+	// The validation store prefixes ads a prefix depending on the field location
+	for ( const fieldKey of Object.keys( additionalFields ) ) {
+		let prefix = '';
+		if ( CONTACT_FORM_KEYS.includes( fieldKey as keyof ContactForm ) ) {
+			prefix = 'contact_';
+		} else if (
+			ORDER_FORM_KEYS.includes( fieldKey as keyof AdditionalValues )
+		) {
+			prefix = 'additional-information_';
```
senadir: I don't think this exists anymore, they should be prefixed with `order_`

senadir:
```suggestion
			prefix = 'order_';
```

senadir (approved):
I left a small about using the wrong error prefix but otherwise all is good.

plugins/woocommerce-blocks/tests/e2e/tests/checkout/additional-fields.shopper.block_theme.spec.ts
```diff
@@ -908,5 +901,96 @@ test.describe( 'Shopper → Additional Checkout Fields', () => {
...
+			await checkoutPageObject.page
+				.getByLabel(
+					'Do you want to subscribe to our newsletter? (optional)'
+				)
+				.check();
+
+			// Check that contact and order fields are persisted
+			await checkoutPageObject.page.waitForResponse(
+				'**/wc/store/v1/checkout**'
+			);
```
WunderBart: For stability, `waitForResponse` promise should be started before the actions that trigger the request are performed, e.g.:

```js
const responsePromise = this.page.waitForResponse( /* ... * / );

await this.page.getByLabel( 'Get it cheaper?' ).check();

await responsePromise;
```

More info here: https://playwright.dev/docs/api/class-page#page-wait-for-response

plugins/woocommerce-blocks/tests/e2e/tests/checkout/additional-fields.shopper.block_theme.spec.ts
```diff
@@ -908,5 +901,96 @@ test.describe( 'Shopper → Additional Checkout Fields', () => {
...
+
+			// Check the required checkbox and refresh.
+			await checkoutPageObject.page
+				.getByLabel( 'Test required checkbox' )
+				.check();
+
+			// This should generate a PUT request to checkout.
+			await checkoutPageObject.page.waitForResponse(
+				'**/wc/store/v1/checkout**'
+			);
```
WunderBart: Same as above.

plugins/woocommerce-blocks/tests/e2e/tests/checkout/checkout.page.ts
```diff
@@ -68,6 +68,15 @@ export class CheckoutPage {
...
 		}
 	) {
+		await Promise.race( [
+			this.page
+				.getByRole( 'group', { name: 'Shipping address' } )
+				.waitFor( { state: 'visible' } ),
+			this.page
+				.getByRole( 'group', { name: 'Billing address' } )
+				.waitFor( { state: 'visible' } ),
+		] );
```
WunderBart: This will still throw a rejected promise in the trace, so let's use the `locator.or` here, e.g.:

```js
await this.page
	.getByRole( 'group', { name: 'Shipping address' } )
	.or( this.page.getByRole( 'group', { name: 'Billing address' } ) )
	.waitFor( { state: 'visible' } ),

```

More info here: https://playwright.dev/docs/api/class-locator#locator-or

alexflorisca: I'm reluctant to do this because of this

> Note that when both locators match something, the resulting locator will have multiple matches, potentially causing a [locator strictness](https://playwright.dev/docs/locators#strictness) violation.

Both billing and shipping addresses may be visible, so this will introduce more flakiness potentially. Although I do get the improved stack trace this would give.

alexflorisca:
@senadir and I noticed that additional fields were also not being validated on PUT today. This was happening because we missed the `validate_callback` of the PUT endpoint. Fixed in https://github.com/woocommerce/woocommerce/pull/55989/commits/95f80207f23b876db03eb61c0d89ab004d724fda. Nadir if you could give this a look over tomorrow and double check it works that would be great.

@WunderBart I addressed your comments about the `waitForResponse`, thanks for pointing that out. 

CI failures seem unrelated to this PR

plugins/woocommerce-blocks/tests/e2e/tests/checkout/checkout.page.ts
WunderBart: In such a case, I think we can append `.first()` to that locator so we won't get yelled at by Playwright:

```js
await this.page
	.getByRole( 'group', { name: 'Shipping address' } )
	.or( this.page.getByRole( 'group', { name: 'Billing address' } ) )
	.first()
	.waitFor( { state: 'visible' } ),
```

mikejolley: This got reopened yesterday, after this change I assume? https://github.com/woocommerce/woocommerce/issues/56037

WunderBart: > This go reopened yesterday after this change I assume https://github.com/woocommerce/woocommerce/issues/56037

sorry, not sure I follow

alexflorisca: Me either?

mikejolley: The flakey test was reopened due to the tests in this PR. The problem with the `#email` selector. It only happened once—I wasn't sure if this logic was being changed and causing that to resurface.

alexflorisca:
Failing tests have been fixed on trunk. p1741956168053479/1741293577.344589-slack-C03CPM3UXDJ

PR #56039, opened 2025-03-03
[Performance]: WooCommerce status widget on dashboard now loads async

### LINKED ISSUE #46931: woocommerce_dashboard_status is extremely slow for high volume stores
### Prerequisites

### Describe the bug

The status widget from the WC_Admin_Dashboard class runs several queries that all add up to an extremely slow load time for sites with a lot of data. 

Note, that the worst of the queries added to this widget come from WooCommerce Subscriptions, but WooCommerce core has a few that take several seconds to complete.  The overall load time for an uncached visit to the dashboard each day is nearly two minutes.

### Expected behavior

~~These queries should be cached to prevent them from running every time the dashboard is rendered.  Since this is just a report, it shouldn't need to be updated constantly with live data.  It should be good enough to have~~ The cached report should be regenerated as an async cron/action as opposed to being rendered during a user visit to the dashboard.

### Actual behavior

The initial priming of cache takes place during a user/admin visit to the dashboard causing long wait times.

### Steps to reproduce

1. On a site with a high number of orders, use a tool like debug bar to show the queries being run.
2. Load the wp-admin dashboard and review the queries.

### WordPress Environment

WPCOM

### Isolating the problem

naman03malhotra:

Closes #46931 
From the issue ^
>The overall load time for an uncached visit to the dashboard each day is nearly two minutes.

In high volume store, this PR should bring down the home page load times from minutes to milliseconds.

At present, the status widget loads synchronously, which is a blocker for the page to load due to a high number of slow queries in the status widget. This PR enables the status widget to load asynchronously, thus eliminating the long page wait times for sites with a high number of orders.

1. Create a site with a large number of orders (>1k) with status `completed`. You can use [smooth generator](https://github.com/woocommerce/wc-smooth-generator) for this. You can use `wp wc generate orders <nr of orders> --status=completed`
2. From `Status > Scheduled Actions`, run the pending actions to generate analytics data.
3. When on `trunk`, Visit `https://w.test/wp-admin/index.php`.
4. When using a tool like query monitor, you'll see a number of slow queries related to the status widget, and the complete page has to load for the widget to render.
5. Now switch to `fix/46931-async-widget`, reload the home page. You should be able to see that the above queries and slow queries related to the status widget no longer exist.
6. You'll observe a placeholder indicating the widget is loading, and after a few milliseconds (or seconds, depending on the processing time of widget queries), the widget gets loaded.
7. You'll also observe an Ajax request set from the browser post the past is loaded with action `woocommerce_load_status_widget`

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


albarin (approved):
- Checked in `trunk` I was seeing the slow query which disappeared when in the PR branch.
- Checked the widget now loads async on the dashboard.

loading

loaded

naman03malhotra:
@prettyboymp requested a review from you as well, to make sure this is the outcome we wanted.

plugins/woocommerce/includes/class-wc-ajax.php
```diff
@@ -3636,6 +3637,20 @@ public static function toggle_gateway_enabled() {
 		wp_die();
 	}
 
+	/**
+	 * AJAX handler for asynchronously loading the status widget content.
+	 */
+	public static function load_status_widget() {
+		check_ajax_referer( 'wc-status-widget', 'security' );
```
prettyboymp, line 3644: We should add the capability checks from `WC_Admin_Dashboard::should_display_widget()` here as well.

prettyboymp (changes requested):
Great simple fix to offload the report loading from the initial request.  This has been a huge pain point when first logging into the admin.

I have just one suggestion around security.

plugins/woocommerce/includes/class-wc-ajax.php
naman03malhotra: 💯 Added as part of https://github.com/woocommerce/woocommerce/pull/56039/commits/c9e0ba2a9ac482de6be0376ca0e51c3fd2e53fa7

naman03malhotra:
@albarin @prettyboymp can please take a final look 👀

albarin (approved):

prettyboymp (approved):
Looks great.  :shipit:

vbelolapotkov:
FYI folks (no action needed), that PR broke WooCommerce Subscriptions integration and potentially some other integrations, because now hooks for extending widget content are moved from dashboard loading request (we had hooks here) to new ajax request (no hooks here).

See WOOSUBS-697 for more context, if curious.

PR #56058, opened 2025-03-04
Update README.md instructions for PNPM

nerrad:
The default `pnpm install` changes the lock file too frequently and *most* users will not be installing new packages. So it's better to default to the new command so that it avoids unnecessary lock file changes.


samueljseay (approved):
Thanks for updating this! It's a good start,

Maybe later we can add some context about when and why they might need it without the flag.

kalessil (approved):

PR #56063, opened 2025-03-04
Product Gallery: introduce mobile layout by forcing the Thumbnails below the main image

### LINKED ISSUE #56030: Product Gallery: Force Thumbnails below large image on mobile
When on mobile, we want to force opinionated layout of Product Gallery.

If Thumbnails block is in use, force the Thumbnails to be rendered horizontally below the large image no matter where they were put in Editor.

Image

kmanijak:

This PR introduces forced mobile view for Product Gallery block:
- Product gallery groups are stacked vertically
- Thumbnails appear below the main gallery image and are arranged horizontally in a row

Closes https://github.com/woocommerce/woocommerce/issues/56030

1. **Default Layout Testing**
   - Go to Editor > Single Product template
   - Replace the Product Image Gallery block with the Product Gallery (beta) block.
   - Leave thumbnails in their default position (left side)
   - Save the template
   - View on mobile device
   - ✓ Verify large image appears first
   - ✓ Verify thumbnails appear below the large image in a horizontal row

Desktop | Mobile
-- | --
image | image

2. **Alternative Layout Testing**
   - Return to Editor
   - Move thumbnails to different positions (change Gallery Area layout settings to do that):
     a. Right side of large image
     b. Top of large image
     c. Bottom of large image
   - Save each variation
   - View each variation on mobile
   - ✓ Verify that regardless of editor position, mobile always shows:
     - Large image first
     - Thumbnails below in a horizontal row

Example with horizontal thumbnails above large image
Editor | Desktop | Mobile
-- | -- | --
image | image | image

3. **Number of thumbnails**
   - Return to Editor
   - Change the number of thumbnails
   - Save and view on mobile
   - ✓ Verify that number of thumbnails is preserved

Example (product with 6 images. "+X" button on the last thumbnail may differ depends on number on images):

4 | 6
-- | --
image | image

5. **Editor preview**
   - Return to Editor
   - Move thumbnails to different positions (change Gallery Area layout settings to do that):
     a. Right side of large image
     b. Top of large image
     c. Bottom of large image
   - Switch to mobile view in Editor
   - ✓ Verify the layout is forced in Editor as well:

image

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


tjcafferkey:
I think this selector you're targeting `.wp-block-group.is-layout-flex ` is too generic and is conflicting with the header on TT5. Only happens when that gallery block is present. Likely need to nest the selector

| trunk | branch |
|--------|--------|
| CleanShot 2025-03-04 at 15 20 29 | CleanShot 2025-03-04 at 15 24 33 |

plugins/woocommerce-blocks/assets/js/blocks/product-gallery/style.scss
```diff
@@ -326,3 +326,17 @@ body.wc-block-product-gallery-dialog-open {
 		}
 	}
 }
+
+// On mobile devices (<600px):
+// - Product gallery groups are stacked vertically
+// - Thumbnails appear below the main gallery image and are arranged horizontally in a row
+@include breakpoint("<600px") {
+	.wp-block-group.is-layout-flex {
```
roykho: This is too generic as Tom pointed out and also if you give it some specificity like adding `wp-block-woocommerce-product-gallery`, you won't have to use `!important`.

kmanijak:
> I think this selector you're targeting .wp-block-group.is-layout-flex  is too generic and is conflicting with the header on TT5. Only happens when that gallery block is present. Likely need to nest the selector

> This is too generic as Tom pointed out and also if you give it some specificity like adding wp-block-woocommerce-product-gallery, you won't have to use !important.

Good catch @tjcafferkey, @roykho! I removed so much code generated by Cursor and missed the obvious flaw. Fixed in https://github.com/woocommerce/woocommerce/pull/56063/commits/998ad52a99e3fcd5b73f21ddc20b96918a520ccf 🙌

tjcafferkey:
**Not related to this PR** but the add to cart form needs optimising a bit for smaller devices I feel. @Aljullu is this your realm? Sorry if I've pinged the wrong person.

CleanShot 2025-03-06 at 13 52 32

tjcafferkey:
In my environment it doesn't look like the editor side thumbnails are being forced to go under the main image like is happening on the frontend.

| Desktop                                      | Mobile                                      |
|--------|--------|
| CleanShot 2025-03-06 at 13 55 39 | CleanShot 2025-03-06 at 13 56 08 |
| CleanShot 2025-03-06 at 13 56 32 | CleanShot 2025-03-06 at 13 57 10 |

kmanijak:
> In my environment it doesn't look like the editor side thumbnails are being forced to go under the main image like is happening on the frontend.

Ah, Product Gallery wrapper missed classname when using `useBlockProps`  in editor: bd1c5d1e95ee6d552129518da82dbe119c7ef036

roykho:
I am not sure if it is related to this PR or if we can even fix this but I noticed a gap at the top (I think a top margin or padding applied). 

CleanShot 2025-03-06 at 08 39 29

Steps to reproduce:

- Add the Product Gallery (Beta)
- Highlight Gallery Area block
- Transform to Stack
- Then Transform to Row
- You will then see a top "margin/padding" in the editor.
- This is also shown in the frontend.

kmanijak:
> I am not sure if it is related to this PR or if we can even fix this but I noticed a gap at the top (I think a top margin or padding applied).

@roykho, not related to this PR but good catch. I explained that behaviour in another PR with the video: https://github.com/woocommerce/woocommerce/pull/54799#issuecomment-2627142410. The gap can be brought back to top alignment manually via block settings.

cpapazoglou (dismissed):
We need to be thoughtful of other blocks wrapping the thumbnails and the large image. Currently, it's possible to wrap them in `columns` which makes the changes to `.wp-block-group` obsolete.

CleanShot 2025-03-07 at 17 18 29

CleanShot 2025-03-07 at 17 19 05

My point is, if we want to be opinionated we also need to force the correct markup in the editor ( perhaps not allowing other blocks between thumbnails and large image? ).

Another idea:
- We can change the default Product Gallery blocks to be wrapped in columns which already have the option to stack in mobile - though we would need to alter the ordering and would make different orientations harder to implement.

CleanShot 2025-03-07 at 17 22 23

kmanijak:
> My point is, if we want to be opinionated we also need to force the correct markup in the editor ( perhaps not allowing other blocks between thumbnails and large image? ).

I see what you mean but also don't necessarily agree. I think we can be opinionated for default setup. Once some manipulation are made, merchants should rely on their configuration. Like you said, after inserting Columns, there's exactly an option to break them on mobile or not 🤔 But maybe I'm trying to simplify things?

But I think I should introduce a custom class to Gallery Area block and make the styling based on it, rather than core CSS classes.

> Another idea:
> We can change the default Product Gallery blocks to be wrapped in columns which already have the option to stack in mobile - though we would need to alter the ordering and would make different orientations harder to implement.

There was an idea to use the blocks like this but that prevents easy switch to horizontal layout. And also seems to be introducing pretty complex inner blocks structure unless we replace "Gallery Area" (Group variation) with Columns (or "raw" Group block).

cpapazoglou:
>  I think we can be opinionated for default setup. Once some manipulation are made, merchants should rely on their configuration. 

That's fair.

> But I think I should introduce a custom class to Gallery Area block and make the styling based on it, rather than core CSS classes.

Combined with the first point, what do you think of introducing a `bool(true)` "Default behavior" in the `Gallery Area` block that:

- locks the whole block - not allowing structure modifications
- applies the class required for forcing the thumbnails below main image
- we can use it for more "default" modifications in the future
- we explain in the editor panel which modifications these are ( currently only this one )

roykho:
@kmanijak 

> But I think I should introduce a custom class to Gallery Area block and make the styling based on it, rather than core CSS classes.

Yeah I think that might be a good idea.

@cpapazoglou 

> locks the whole block - not allowing structure modifications

So we should stay away from locking blocks as much as possible. This was one of our design decisions a little while back. It is expected that when users change what was there by default, they know what they're doing and can either revert, re-add the block or just not saving the page and refresh.

kmanijak:
Hey folks, I inserted custom classes and made the mobile styling based on them. Be aware that this require re-adding the block from the inserter in order for newly added classes to be applied, it won't work for existing blocks.

> So we should stay away from locking blocks as much as possible. This was one of our design decisions a little while back. It is expected that when users change what was there by default, they know what they're doing and can either revert, re-add the block or just not saving the page and refresh.

Roy made a great summary, that's the general rule we follow.

roykho:
Looking good so far. I just have a concern. When switching to row, the pager is put to the side of the large image. But I think I always would expect the pager to be centered at the bottom. wdyt?

CleanShot 2025-03-11 at 07 51 18

cpapazoglou:
> Looking good so far. I just have a concern. When switching to row, the pager is put to the side of the large image. But I think I always would expect the pager to be centered at the bottom. wdyt?
> 
> CleanShot 2025-03-11 at 07 51 18

The pager is now removed, merging with `trunk` won't show it at all anymore.

roykho (approved):
Great work!

✅ I've tested the thumbnails in all four locations above, below, left and right of the large image. In all cases, the editor (mobile) and frontend (mobile) shows the thumbnails below the large image:
CleanShot 2025-03-11 at 10 37 23


CleanShot 2025-03-11 at 10 36 35

kmanijak:
Thanks folks for syncing between yourself and updating the branch, much appreciated! 🙇

PR #56089, opened 2025-03-04
Added caching to Tax Onboarding Task.

### LINKED ISSUE #54908: [Performance]: Improve performance of OnboardingTask\Tax::is_complete() check
### Prerequisites

### Performance Details

The onboarding task for Tax setup attempts to check if any taxes have been added to determine if the task is complete.  However, it does so by loading and counting all the taxes that exist in the database.

https://github.com/woocommerce/woocommerce/blob/5acc35da54786fdf900bf81e7297aede64ee658f/plugins/woocommerce/src/Admin/Features/OnboardingTasks/Tasks/Tax.php#L119-L123

This is run on every admin page load.

### Affected Sites

This affects sites where there are a lot of tax locations, e.g. where a 3rd party API service isn't being used.  There are no filters on the query, so the speed of the query is solely based on the amount of data in the tax_rates table.

On a local site with ~30k tax locations loaded (3/4's of current US zip codes), it takes around 35ms to run the query (not including processing/object hydration).

### Suspected causes

_No response_

## Suggested Improvements

* Modify the check for existing taxes so that the query only needs to check for the existence of a tax rate rather than loading the entire dataset.
* Cache the result of the check for existing tax rates. Make sure cache is cleared on relevant actions.

prettyboymp:

This PR does the following:

1.  Adds caching around the check for existing tax rates to avoid the need to repeatedly query the tax rates.
2. Marks the Tax task as actioned once a tax rate exists as an indefinite completion marker.  This allows the task to resurface if the 3rd party implementation is uninstalled but no tax rates were ever created.

Closes #54908 .

1.  Setup a site from scratch, or clear any tax rates and the following option values: `woocommerce_task_list_tracked_completed_tasks`, `woocommerce_no_sales_tax`, `woocommerce_task_list_tracked_completed_actions`.
2. Use a query monitoring plugin to watch for queries.
3. Visit the WooCommerce Admin to view the current Onboarding Task List.
4. The tax task should show as incomplete.
Screenshot 2025-03-04 at 1 35 03 PM
5. Verify that the query `SELECT 1 {$wpdb->prefix}woocommerce_tax_rates limit 1` only run during the initial load (if not already cached from the Dashboard metabox).
6. Go to setup taxes manually and insert a new tax rate.
7. Revisit the WooCommerce Admin and verify that the task now shows as complete and that the query against the tax rates is no longer being run.  This should be the case even if cache is cleared.


plugins/woocommerce/src/Admin/Features/OnboardingTasks/Task.php
```diff
@@ -391,15 +391,19 @@ public function has_previously_completed() {
...
 	/**
-	 * Track task completion if task is viewable.
+	 * Track task completion if task is viewable and is complete.
+	 *
+	 * @param bool $verify_is_complete Whether to test if $this->complete() prior to tracking completion. Default true.
+	 *
+	 * @return void
 	 */
-	public function possibly_track_completion() {
+	public function possibly_track_completion( $verify_is_complete = true ) {
```
joshuatf: Instead of adding an additional argument here, could we refactor `is_complete()` to cache its value in a `is_complete` property inside the instance so we can access it without needing to recalculate?

plugins/woocommerce/src/Admin/Features/OnboardingTasks/Tasks/Tax.php
```diff
@@ -117,14 +118,45 @@ public function is_complete() {
 			$third_party_complete = apply_filters( 'woocommerce_admin_third_party_tax_setup_complete', false );
 
 			$this->is_complete_result = $is_wc_connect_taxes_enabled ||
-				count( TaxDataStore::get_taxes( array() ) ) > 0 ||
-				get_option( 'woocommerce_no_sales_tax' ) !== false ||
-				$third_party_complete;
+                   get_option( 'woocommerce_no_sales_tax' ) !== false ||
```
joshuatf: Just double checking, but seems like we're putting the less expensive checks first so we might not need to run the more expensive checks, right?

If so, 👍

plugins/woocommerce/src/Admin/Features/OnboardingTasks/Tasks/Tax.php
```diff
@@ -117,14 +118,45 @@ public function is_complete() {
...
 
+	/**
+	 * Determines if a tax rate exists in the database.  Result is indefinitely cached.
+	 *
+	 * @return bool
+	 */
+	private function has_existing_tax_rates() {
+		global $wpdb;
+
+		$has_existing_tax_rates = wp_cache_get( 'onboarding_task_tax_rates_exist' );
```
joshuatf: Should we prefix these with `woocommerce_`?

plugins/woocommerce/src/Admin/Features/OnboardingTasks/Tasks/Tax.php
```diff
@@ -28,6 +28,7 @@ class Tax extends Task {
 	public function __construct( $task_list ) {
 		parent::__construct( $task_list );
 		add_action( 'admin_enqueue_scripts', array( $this, 'possibly_add_return_notice_script' ) );
+		add_action( 'woocommerce_tax_rate_added', array( $this, 'track_actioned_on_tax_rate_added' ) );
```
joshuatf, line 31: Nice addition!

joshuatf (commented):
Nice optimization!  Left one comment on how we might be able to implement a cached `is_complete` across all tasks.

plugins/woocommerce/src/Admin/Features/OnboardingTasks/Task.php
prettyboymp: @joshuatf, I initially attempted to avoid adding the additional parameter, but since each `Task` implementation essentially overrides `::is_complete()`, there isn't a great way to do that, especially since some Tasks are implemented in other extensions, e.g. WooCommerce Payments.

This isn't an improvement specifically needed for the Tax Task, but for the way `Task::get_json()` is used because the current implementation has `::is_complete()` get called twice, which in some Tasks, is expensive.  So this was an attempt to optimize external task implementations a little as well.  I'm good leaving it out if that is preferred.

joshuatf: Ya, I see what you mean and get how this gives us some immediate performance gains in the short-term.  I think my preference would be to refactor `is_complete()` to cover these cases.

```php
abstract Task {
    /**
     * Cache completion for each page load.
     *
     * @var bool|null
     */
    private $is_complete = null;

    public function is_complete() {
        if ( null !== $this->is_complete ) {
            return $this->is_complete;
        }

        if ( $this->is_actioned() ) {
            return true;
        }

        if ( $this->has_previously_completed() ) {
            return true;
        }

        $this->is_complete = $this->calculate_completion(); // Naming TBD.
        return $this->is_complete;
    }
}

class ConcreteTask {
    /**
     * Calculate if a task is complete, bypassing any caching.
     *
     * @return bool
     */
    private function calculate_completion() {
        // Run expensive queries to check for completion.
        return $result;
    }
}
```

This accomplishes two things:

1. We get built-in caching so that other calls to `is_complete()` don't re-query during the same page load.  Note that some tasks have already [taken it upon themselves to do this](https://github.com/woocommerce/woocommerce/blob/97647674de2273eab3abf12f30ded84fe51c1bdd/plugins/woocommerce/src/Admin/Features/OnboardingTasks/Tasks/AdditionalPayments.php/#L14-L19).
2. We completely bypass the need to query at all if consumers opt-in to the new usage.  I opened a thread for discussion in the @woocommerce/ghidorah Slack channel to check current expectations around tasks and keeping them marked complete.

joshuatf: Obviously this means a couple external tasks, like those from WooCommerce Payments, might not get the immediate benefit, but the change is trivial for them to make and take advantage of the performance gains.

I'm also okay with this change in a follow-up if you prefer, though I'd like to avoid the additional logic in `possibly_track_completion` as I think it adds complexity.  If we go this route, maybe we could just apply the performance changes to the Tax task.

plugins/woocommerce/src/Admin/Features/OnboardingTasks/Tasks/Tax.php
```diff
@@ -117,14 +118,45 @@ public function is_complete() {
...
 			$this->is_complete_result = $is_wc_connect_taxes_enabled ||
-				count( TaxDataStore::get_taxes( array() ) ) > 0 ||
-				get_option( 'woocommerce_no_sales_tax' ) !== false ||
-				$third_party_complete;
+				false !== get_option( 'woocommerce_no_sales_tax' ) ||
+				$third_party_complete ||
+				$this->is_actioned();
+
+			if ( ! $this->is_complete_result && $this->has_existing_tax_rates() ) {
+				$this->mark_actioned();
```
joshuatf, line 126: We should also add `$this->is_complete_result = true;` in here otherwise this function still returns `false` the first time even though its actioned.  This results in the task being marked incomplete the first time and complete on subsequent reloads.

**First load**

Screenshot 2025-03-17 at 3 32 04 PM

**Reload**
Screenshot 2025-03-17 at 3 33 06 PM

plugins/woocommerce/src/Admin/Features/OnboardingTasks/Tasks/Tax.php
```diff
@@ -117,14 +118,45 @@ public function is_complete() {
...
+	 * Determines if a tax rate exists in the database.  Result is indefinitely cached.
+	 *
+	 * @return bool
+	 */
+	private function has_existing_tax_rates() {
+		global $wpdb;
+
+		$has_existing_tax_rates = wp_cache_get( 'woocommerce_onboarding_task_tax_rates_exist' );
+		if ( false === $has_existing_tax_rates ) {
+			$rate_exists            = (bool) $wpdb->get_var( "SELECT 1 {$wpdb->prefix}woocommerce_tax_rates limit 1" );
```
joshuatf: This query will always return `1`; we'll need a `FROM` in here:

```suggestion
			$rate_exists            = (bool) $wpdb->get_var( "SELECT 1 FROM {$wpdb->prefix}woocommerce_tax_rates limit 1" );
```

joshuatf (approved):
Testing well with a couple very minor modifications.

* ✅ Tax query is hit on first load only and not subsequent loads
* ✅ Caching prevents query from being re-run when persistent object cache is enabled
* ✅ Once task is actioned, the check for taxes no longer gets hit
* ⚠️ SQL query always returns `1`; fixed by addition of `FROM`
* ⚠️ Initially complete task shows incomplete on first load

Pre-approving this PR as tested with the 2 update suggestions left in the comments.  Feel free to ping me if you need another review, but I've tested with those changes and things are working well otherwise.

Screenshot 2025-03-17 at 3 48 49 PM

plugins/woocommerce/src/Admin/Features/OnboardingTasks/Tasks/Tax.php
```diff
@@ -117,14 +118,45 @@ public function is_complete() {
 			$third_party_complete = apply_filters( 'woocommerce_admin_third_party_tax_setup_complete', false );
 
 			$this->is_complete_result = $is_wc_connect_taxes_enabled ||
-				count( TaxDataStore::get_taxes( array() ) ) > 0 ||
-				get_option( 'woocommerce_no_sales_tax' ) !== false ||
-				$third_party_complete;
+				false !== get_option( 'woocommerce_no_sales_tax' ) ||
```
kalessil: Guess we can micro-optimize this a bit:

```php
			$this->is_complete_result = $is_wc_connect_taxes_enabled ||
				$third_party_complete ||
				false !== get_option( 'woocommerce_no_sales_tax' ) ||
				$this->is_actioned();
```

plugins/woocommerce/src/Admin/Features/OnboardingTasks/Tasks/Tax.php
kalessil: Shouldn't [this approach](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Internal/ProductAttributesLookup/LookupDataStore.php#L709) perform slightly faster?

`0 !== ( (int) $wpdb->get_var( "SELECT EXISTS (SELECT 1 FROM {$wpdb->prefix}woocommerce_tax_rates)" ) )` in our case and IIRC it's performance is not depending on having/not having `limit 1` in the inner query.

kalessil (commented):
Like the optimization added in this PR! Leaving a couple of comments to consider as well.

PR #56098, opened 2025-03-05
[dev] CI: fix CI-workflow failures on tagging caused by wp-env overrides.

kalessil:

Fixes failing CI-workflow on tagging event ([example](https://github.com/woocommerce/woocommerce/actions/runs/13655956509/job/38218706009)).

The root cause is that some of the jobs will override the wp-env config and use the core zip file and ignore the checked-out codebase, causing the test environment provisioning failures. We link the necessary script separately in wp-env configuration to fix this.

- CI-checks shall pass

### DIFF
--- a/plugins/woocommerce/.wp-env.json
+++ b/plugins/woocommerce/.wp-env.json
@@ -32,6 +32,7 @@
 			},
 			"mappings": {
 				"wp-cli.yml": "./tests/wp-cli.yml",
+				"test-env-setup.sh": "./tests/e2e-pw/bin/test-env-setup.sh",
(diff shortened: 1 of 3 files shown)


adimoldovan (approved):
Thanks for the fix @kalessil ! 

Tested locally and it worked.

```
# create the override config
CI=1 RELEASE_TAG=nightly ARTIFACT_NAME=woocommerce-trunk-nightly.zip 
WP_ENV_CONFIG_PATH=<path to the wp-env config> node .github/workflows/scripts/override-wp-env-plugins.js

# start the env
CI=1 pnpm filter="@woocommerce/plugin-woocommerce" env:test
```

PR #56102, opened 2025-03-05
Add `woocommerce` as a dependency in WooCommerce Beta Tester plugin

SantosGuillamot:

I suggest adding `woocommerce` as a required plugin for WooCommerce Beta Tester. As explained in [this article](https://make.wordpress.org/core/2024/03/05/introducing-plugin-dependencies-in-wordpress-6-5/), since WordPress 6.5 it is possible to define the dependencies in the plugin itself, which handles some workflows by default.

**Before**

https://github.com/user-attachments/assets/dfd13c36-591a-4a5e-9f86-36a79633130a

**After**

Screenshot 2025-03-05 at 09 17 09

### DIFF
--- a/plugins/woocommerce-beta-tester/.wp-env.json
+++ b/plugins/woocommerce-beta-tester/.wp-env.json
@@ -1,7 +1,4 @@
 {
 	"phpVersion": "7.4",
-	"plugins": [ 
-        ".",
-        "https://downloads.wordpress.org/plugin/woocommerce.zip"
-     ]
+	"plugins": [ "https://downloads.wordpress.org/plugin/woocommerce.zip", "." ]
 }
--- a/plugins/woocommerce-beta-tester/woocommerce-beta-tester.php
+++ b/plugins/woocommerce-beta-tester/woocommerce-beta-tester.php
@@ -12,6 +12,7 @@
  * WC requires at least: 9.4
  * WC tested up to: 9.5
  * Text Domain: woocommerce-beta-tester
+ * Requires Plugins: woocommerce
  *
  * @package WC_Beta_Tester
  */


plugins/woocommerce-beta-tester/.wp-env.json
```diff
@@ -1,7 +1,4 @@
 {
 	"phpVersion": "7.4",
-	"plugins": [ 
-        ".",
-        "https://downloads.wordpress.org/plugin/woocommerce.zip"
-     ]
+	"plugins": [ "https://downloads.wordpress.org/plugin/woocommerce.zip", "." ]
```
SantosGuillamot, line 3: The order of this array defines the order the plugins are installed. WooCommerce needs to be installed before WooCommerce Beta Tester.

plugins/woocommerce-beta-tester/woocommerce-beta-tester.php
```diff
@@ -12,6 +12,7 @@
  * WC requires at least: 9.4
  * WC tested up to: 9.5
  * Text Domain: woocommerce-beta-tester
+ * Requires Plugins: woocommerce
```
chihsuan, line 15: Thank you @SantosGuillamot for the PR! While it's a good suggestion, we should avoid making `woocommerce` a required plugin for WooCommerce Beta Tester. The beta tester includes features that are intended to function independently of `woocommerce`.

For instance, the beta tester allows us to install any desired WooCommerce version, including versions from the development branch (Live branch feature).

plugins/woocommerce-beta-tester/.wp-env.json
chihsuan: This looks good to me!

chihsuan (changes requested):

plugins/woocommerce-beta-tester/woocommerce-beta-tester.php
SantosGuillamot: Oh, so WooCommerce is not a dependency of WooCommerce Beta Tester? Can WooCommerce Beta Tester plugin work without WooCommerce installed?

If that's the case, I think that's not clear just by reading the plugin readme and this message users get when WooCommerce is deactivated is confusing:

Screenshot 2025-03-05 at 10 40 20

Just to clarify, adding it as a required plugin doesn't mean it needs to be the latest WooCommerce version. As long as you have any version of the plugin, including development branches, it should work as expected if I am not mistaken.

chihsuan: > Oh, so WooCommerce is not a dependency of WooCommerce Beta Tester? Can WooCommerce Beta Tester plugin work without WooCommerce installed?

Some features depend on WooCommerce, but it's not necessary to have WooCommerce installed to use the plugin. 

> If that's the case, I think that's not clear just by reading the plugin readme, and this message users get when WooCommerce is deactivated is confusing:

Agree with you, I think when the notice was introduced, the beta tester was a dependency of WooCommerce, but subsequent updates introduced new features that do not require WooCommerce.

We probably should tweak the notice to make it clearer that WooCommerce needs to be installed to **fully** utilize the beta tester's features.

> Just to clarify, adding it as a required plugin doesn't mean it needs to be the latest WooCommerce version. As long as you have any version of the plugin, including development branches, it should work as expected if I am not mistaken.

Yes, that's correct. In our scenario, we don't have WooCommerce installed, but we still want to use the WooCommerce Beta Tester plugin to install and test any desired version of WooCommerce.

SantosGuillamot:
Closing this because WooCommerce Beta Tester plugin doesn't always have WooCommerce as a dependency. See explanation [here](https://github.com/woocommerce/woocommerce/pull/56102#discussion_r1980991531).

PR #56103, opened 2025-03-05
Adjustments to the COGS edition in the product editor

Konamiman:

This pull implements makes some adjustments to the Cost of Goods Sold edition for products that was implemented in https://github.com/woocommerce/woocommerce/pull/54399:

* The cost field in the General tab is moved so that it's now in the prices section, instead of in a section of its own:

| Before | After |
| ------ | ----- |
| image       |  image     |

* When the cost edited in the General tab is zero now the field cost is initially empty with a "0" placeholder, instead of loading with an explicit 0 value:

| Before | After |
| ------ | ----- |
| image |  image |

* An additional help text is displayed under the cost field for variations when the field is empty, including a link that opens the General tab. This text appears or disappears dinamically as needed when the field is edited.

| Before | After |
| ------ | ----- |
| image | image |

* When the type of a product is changed between variable and any other type, the tooltip of the cost field in the General tab is updated accordingly (previously it only changed after the product was saved and the page was reloaded).

* The "Remove all costs" bulk action for variations in the bulk actions dropdown has been shifted to right below the "Pricing" section (it previously was at the very end of the dropdown):

| Before | After |
| ------ | ----- |
|        |       |
| image |  image |

Additionally, this pull request extends the `woocommerce_wp_text_input` function to allow displaying a tooltip text **and** a description text under the field, both at the same time:

* When the supplied `description` argument is an array of two strings (instead of one single string), the first element of the array will be used as the tooltip, and the second element will be used as the description text. The `desc_tip` argument is ignored in this case.
* A new argument is added: `description_hidden`, when it's present and it's true the description text span will be rendered with a `hidden` CSS class.

1. Repeat the testing instructions of https://github.com/woocommerce/woocommerce/pull/54399, verifying that everything continues working as expected and that the changes in element placements described are effective.
2. Set the value of the cost field to 0 in the General tab and save, after reload the field will be empty with a 0 placeholder.
   * Note: the backend isn't currently storing cost values of zero (a product with no cost is assumed to have a cost of zero), so the only options when a product has no cost are displaying 0 (previous behavior) or displaying empty (change implemented here), **both** when no cost has been defined for a product and when a cost of 0 is set and saved. Variations are an exception to this.
4. Open the "Variations" tab of a variable product and verify that the changes to the cost field work as expected:
   * Initially (after the product editor is loaded), a "You can specify a default value for all variations" helper text is present below the field for the variations for which the field is empty (and only for those).
   * Deleting the contents of the cost field makes the helper text appear, adding any value makes the text disappear. This happens immediately, there's no need to take the focus off the field.
   * The above works both when directly typing or erasing the contents of the field, as well as when cutting or pasting the contents.
   * Each variation is independent from each other (changing the contents of the field for one variation doesn't affect the help text of the other variations).
   * The "default value" link opens the General tab when clicked.
5. In the General tab change the product type between "Simple" and "Variable" and verify that the tooltip for the cost field changes (for variable products it has "This will be applied as the default value for variations" added at the end).

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/includes/admin/class-wc-admin-assets.php
```diff
@@ -448,6 +449,11 @@ public function admin_scripts() {
 					'i18n_attributes_used_for_variations_placeholder' => sprintf( esc_attr__( 'Enter options for customers to choose from, f.e. “Blue” or “Large”. Use “%s” to separate different options.', 'woocommerce' ), esc_attr( WC_DELIMITER ) )
 				);
 
+				$cogs_controller = wc_get_container()->get( CostOfGoodsSoldController::class );
+				if($cogs_controller->feature_is_enabled()) {
+					$params['cogs_value_tooltip_simple_products'] = esc_attr($cogs_controller->get_general_cost_edit_field_tooltip(false));
+					$params['cogs_value_tooltip_variable_products'] = esc_attr( $cogs_controller->get_general_cost_edit_field_tooltip(true));
+				}
```
naman03malhotra: This part of the code can use some liniting.

naman03malhotra (commented):
Tests well overall, but the couldn't somehow get the custom HTML part to be working.
 
I've tested this PR from the instructions given in https://github.com/woocommerce/woocommerce/pull/56103


✅  Try the quick editor ("Quick edit" link in the product rows in the products table). Verify that the cost shown in the "Cost" field is initially 0, and that you can change it and save, and the value gets persisted too.

✅  Try the bulk editor (mark the checkbox for a few products in the products table, select "Edit" in the "Bulk actions" dropdown, then click "Apply"). Verify that the cost is modified for the selected products only if the cost field is set to "Change to:", and that an empty value sets a cost of zero.

✅  Open the editor for a variation ("Variations" tab when editing a variable product). Verify that the "Cost" field shows a placeholder with the cost value set for the parent product in the "General" tab (or zero, if none has been set). Verify that you can set a value and save it, and that you can then delete the value and save it too (note that an empty value is not equivalent to zero in this case).

✅  Try the "Remove custom costs" bulk action for variations (edit a variable object, go to the "Variations" tab, click the "Bulk actions" dropdown). Try accepting and cancelling the confirmation prompt (it should remove the custom cost of all the variations in the first case, and no costs should be changed in the second case).

✅  Verify that the "Cost" column in the products table shows the proper values for each product (zero is indicated with a hyphen).

❓ Try the cost column HTML filters with the following snippet:

```
add_filter('woocommerce_empty_cogs_html', function($html, $product)  {
    return "<p>No cost for " . $product->get_id() . "</p>";
}, 2, 10);

add_filter('woocommerce_get_cogs_html', function($html, $value, $product)  {
    return
        0.0 === $value ?
        $html :
        '<p>Cost for ' . $product->get_id() . ': ' . $value . '</p>';
}, 3, 10);
```

✅  Finally, disable the Cost of Goods Sold feature and verify that none of the added UI elements that you just tested is visible anymore, but you can still edit the rest of the product fields (using regular editor, quick edit, bulk edit, variation editor and variation bulk action).



From this PR: https://github.com/woocommerce/woocommerce/pull/56103

✅  1. Set the value of the cost field to 0 in the General tab and save, after reload the field will be empty with a 0 placeholder.
 
✅  2. Open the "Variations" tab of a variable product and verify that the changes to the cost field work as expected:
   * Initially (after the product editor is loaded), a "You can specify a default value for all variations" helper text is present below the field for the variations for which the field is empty (and only for those).
   * Deleting the contents of the cost field makes the helper text appear, adding any value makes the text disappear. This happens immediately, there's no need to take the focus off the field.
   * The above works both when directly typing or erasing the contents of the field, as well as when cutting or pasting the contents.
   * Each variation is independent from each other (changing the contents of the field for one variation doesn't affect the help text of the other variations).
   * The "default value" link opens the General tab when clicked.

✅ 3. In the General tab change the product type between "Simple" and "Variable" and verify that the tooltip for the cost field changes (for variable products it has "This will be applied as the default value for variations" added at the end).

From testing instruction from this PR:

| Before | After |
| ------ | ----- |
| image       |  image     |

| Before | After |
| ------ | ----- |
| image |  image |

| Before | After |
| ------ | ----- |
| image | image |

| Before | After |
| ------ | ----- |
|        |       |
| image |  image |

plugins/woocommerce/includes/admin/class-wc-admin-assets.php
Konamiman: PHPCBF is ignoring these lines for some reason. Fixed manually.

Konamiman:
> couldn't somehow get the custom HTML part to be working.

These hooks were renamed in https://github.com/woocommerce/woocommerce/pull/55840, details are in the PR description.

naman03malhotra (approved):
The hook changes the view. Thanks for the tip!
image

naman03malhotra:
@Konamiman the Lint CI is failing, though I've approved it. But let's make sure it passes before we merge this PR.

PR #56110, opened 2025-03-05
Email Editor - Fix saving non public emails

### LINKED ISSUE #55862: Investigate the error caused by using `public = false` or `exclude_from_search`  when registering email editor post-types.
Thanks for sharing the details. Could you please create an issue for investigating the error so that we don't forget? 

**Context**: Using `public = false` or `exclude_from_search` for the email editor post-type leads to this error when saving the email. We need to take a closer look and fix it.
Screenshot 2025-02-25 at 11 32 40 AM

I think the error is related to this code https://github.com/woocommerce/woocommerce/blob/88f1c3c00c9bfc7c51f0b250623c4be761381d52/packages/php/email-editor/src/Engine/Templates/class-templates.php#L140-L167
If I'm not mistaken, we already added the templates registry so we could use it to list email templates instead of [fetching all templates and filtering by the prefix](https://github.com/woocommerce/woocommerce/blob/88f1c3c00c9bfc7c51f0b250623c4be761381d52/packages/php/email-editor/src/Engine/Templates/class-templates.php#L158-L166).

_Originally posted by @costasovo in https://github.com/woocommerce/woocommerce/pull/55754#discussion_r1971215468_

costasovo:

Closes #55862

This PR improves the code that adds email related tempaltes to list of allowed templates when saving an email. When a template is not in the list the REST API can't save the email - template association. Previous code was relying on the template prefix, but in case the extender used a different prefix it didn't worked and email couldn't be saved.

Note: We have this workaround in place to allow having emails set as non public (see [register_post_type](https://developer.wordpress.org/reference/functions/register_post_type/)). Alternatively we could drop this code and force all emails to be set as public and use other arguments to specify the behaviour. I chose to keep it for now.

1. See that the woo_email post_type is configured with `public` => `false`
2. Create a new email in the block editor and make some edits
3. Save the email
4. Observe that email was properly saved

Note: In case you want to replicate the issue. The issue was easily replicable by switching the woo_email [CPT to public](https://github.com/woocommerce/woocommerce/blob/be742f8fd2d496b176bdccbc6e8b808c73405e31/plugins/woocommerce/src/Internal/EmailEditor/Integration.php#L98). The [new template prefix we added in Woo](https://github.com/woocommerce/woocommerce/blob/be742f8fd2d496b176bdccbc6e8b808c73405e31/plugins/woocommerce/src/Internal/EmailEditor/EmailTemplates/TemplatesController.php#L25) was different from the prefix used [in the package](https://github.com/woocommerce/woocommerce/blob/be742f8fd2d496b176bdccbc6e8b808c73405e31/packages/php/email-editor/src/Engine/Templates/class-templates.php#L23). The package prefix used to be `mailpoet` and we changed that in https://github.com/woocommerce/woocommerce/pull/55938 to `woocommerce` so currently the issue is not replicable just by swtiching CPT to public. To replicate it you need to change one of the prefixes to emulate that and integrator uses a different prefix.

The issue was causing that it was not possible to save the email. I replicated the issue locally and after the change I was able to save emails as expected.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/packages/php/email-editor/changelog/improve-allowed-templates-check-on-save
+++ b/packages/php/email-editor/changelog/improve-allowed-templates-check-on-save
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+Use email templates registry when listing allowed templates for association with an email post
--- a/packages/php/email-editor/src/Engine/Templates/class-templates.php
+++ b/packages/php/email-editor/src/Engine/Templates/class-templates.php
@@ -155,11 +155,18 @@ class Templates {
 		if ( $post_type && ! in_array( $post_type, $this->post_types, true ) ) {
 			return $templates;
 		}
-		$block_templates = get_block_templates();
+		$block_templates       = get_block_templates();
+		$email_templates_slugs = array_map(
+			function ( Template $template ) {
+				return $template->get_slug();
+			},
+			$this->templates_registry->get_all()
+		);
 		foreach ( $block_templates as $block_template ) {
-			// Ideally we could check for supported post_types but there seems to be a bug and once a template has some edits and is stored in DB
-			// the core returns null for post_types.
-			if ( $block_template->plugin !== $this->template_prefix ) {
+			if ( ! in_array( $block_template->slug, $email_templates_slugs, true ) ) {
+				continue;
+			}
+			if ( isset( $templates[ $block_template->slug ] ) ) {
 				continue;
 			}
 			$templates[ $block_template->slug ] = $block_template;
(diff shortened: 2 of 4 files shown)


pavel-mailpoet (approved):
Testing on Jurassic ninja: https://d.pr/i/SL4WAU and the email was saved without issues: https://d.pr/i/xBECB6

PR #56117, opened 2025-03-05
iAPI: Switch the cart endpoint in the `woocommerce` store

luisherranz:

This PR changes the endpoint being used in the `woocommerce` store from `cart/items` to `update-item` and `add-item`.

This change is necessary because when updating the cart, we want the entire cart to be returned from the server in case there have been any changes, rather than just the item in question.

I haven't tested it extensively yet, but initially, it seems to fix the issue reported here: https://d.pr/v/E3v6NP

1. Add a couple of Product Collection blocks.
2. Make it so that one of the products is limited to being added only once in the cart.
3. Try adding that product to the cart more than once in each of the Product Collection blocks.
4. You should see a notice in the relevant Product Collection block (and not in the other).
5. Now add a different product to the cart.
6. Without refreshing the page, mark that product as Out of Stock.
7. Try adding it to the cart again.
8. You should see a notice in the relevant Product Collection block (and not in the other).

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/base/stores/cart-items.ts
```diff
@@ -29,26 +30,22 @@ export type Store = {
...
 
-function isSuccessfulResponse(
+function isApiErrorResponse(
 	res: Response,
-	json: CartItemsResponse | CartItemResponse
-): json is Item | Item[] {
-	return res.status.toString().startsWith( '2' );
+	json: unknown
+): json is ApiErrorResponse {
+	return ! res.status.toString().startsWith( '2' );
```
samueljseay: nitpick but should we just make this `return ! res.ok`?

plugins/woocommerce/client/blocks/assets/js/base/stores/cart-items.ts
```diff
@@ -149,17 +135,17 @@ export const { state, actions } = store< Store >(
 
 				try {
 					const res: Response = yield fetch(
-						`${ state.restUrl }wc/store/v1/cart/items`,
+						`${ state.restUrl }wc/store/v1/cart`,
```
samueljseay, line 147: I think we'll need to update the e2e tests to change which endpoint they're waiting to resolve and that's possibly why quite a few are failing right now.

samueljseay (changes requested):
I think the e2e tests will need some adjustment. 1 nitpick. All looks good otherwise.

plugins/woocommerce/client/blocks/assets/js/base/stores/cart-items.ts
luisherranz: Indeed, thank you.

plugins/woocommerce/client/blocks/assets/js/base/stores/cart-items.ts
luisherranz: Ok, I'll look into it. Thanks.

plugins/woocommerce/client/blocks/assets/js/base/stores/cart-items.ts
luisherranz: Fixed in https://github.com/woocommerce/woocommerce/pull/56117/commits/f157de8a195560ba91f17b13a214fd3f669c5b73.

luisherranz:
I ended up making more changes than expected on this PR, so I made a video to explain them.

<div>
    <a href="https://www.loom.com/share/565b71839b8740bd96ce511ea19b1d53">
      
    </a>
  </div>

https://www.loom.com/share/565b71839b8740bd96ce511ea19b1d53

I haven't looked at the e2e tests yet, but this should be ready for a preliminary review.

plugins/woocommerce/src/Blocks/AssetsController.php
```diff
@@ -115,6 +115,26 @@ public function register_assets() {
 			",
 			'before'
 		);
+
+		wp_register_script_module(
```
samueljseay, line 119: I don't like seeing this. If we could explore improving this now I'd favor it.

I'm still not sure how the GB implementation ends up generating modules with the module name prefix: https://github.com/WordPress/gutenberg/blob/trunk/tools/webpack/script-modules.js but I notice they don't modify the standard config much at all.

I messed around locally and found that adding aliases in to Webpack did generate an asset.php file with better looking entries. Haven't tested it though:

```js
resolve: {
		alias: {
			'@woocommerce/stores': path.resolve(
				__dirname,
				'../assets/js/base/stores'
			),
		},
	},
```

example in asset.php output:

```php
'stores/woocommerce/cart.js'                    => array(
		'dependencies' => array(
			'@wordpress/interactivity',
			array(
				'id'     => '@woocommerce/stores/store-notices',
				'import' => 'dynamic',
			),
		),
		'version'      => 'bd97956b3a6257f151ab',
		'type'         => 'module',
	),
```

luisherranz:
I'll look into the e2e tests.

samueljseay (approved):
tests well, uses the correct endpoints, shows errors correctly.

🚢

PR #56125, opened 2025-03-05
Update e2e `addToCart` and `goToCheckout` helper to reduce flakiness 

mikejolley:

Several of the flaky test reports have the same issue:

```
waiting for locator('#email') to be visible
```

This comes from the `goToCheckout` helper following an `addToCart` in the e2e test suite. From seeing the test fail videos its not clear why this is happening. It seems to be redirecting to the cart page after adding to cart, and getting stuck there.

To attempt to mitigate this we'll wait for the dom to full load (in case of redirect) before attempting to redirect to the checkout. I'm not certain if this will fix the flaky tests so this should be monitored and looked into deeper if it fails to fix the tests.

The following flaky tests are affected by this:

- https://github.com/woocommerce/woocommerce/issues/55797
- https://github.com/woocommerce/woocommerce/issues/55984
- https://github.com/woocommerce/woocommerce/issues/55994
- https://github.com/woocommerce/woocommerce/issues/56004
- https://github.com/woocommerce/woocommerce/issues/56007
- https://github.com/woocommerce/woocommerce/issues/56023

| Before | After |
| ------ | ----- |
|        |       |

N/A. This is a change to CI only.

Running e2e locally where the problem is not apparent.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


mikejolley:
After a few runs I've not seen the redirect issue in the https://github.com/woocommerce/woocommerce/actions/runs/13696696941/job/38302806793?pr=56125 test. There is still a separate flaky test there but its not related to this one and can be tackled separately.

plugins/woocommerce/client/blocks/tests/e2e/utils/frontend/frontend-utils.page.ts
```diff
@@ -41,27 +41,31 @@ export class FrontendUtils {
 				url.includes( 'batch' )
 			);
 		} );
+		// Wait for all XHR requests to finish. Some blocks do another request to the cart endpoint after the above.
+		// eslint-disable-next-line playwright/no-networkidle
+		await this.page.waitForLoadState( 'networkidle' );
```
WunderBart, line 46: Using `networkidle` is discouraged as it does not produce a stable outcome.

https://playwright.dev/docs/api/class-frame#frame-wait-for-load-state

plugins/woocommerce/client/blocks/tests/e2e/utils/frontend/frontend-utils.page.ts
```diff
@@ -41,27 +41,31 @@ export class FrontendUtils {
...
 	async goToCheckout() {
-		await this.page.goto( '/checkout' );
+		await this.page.goto( this.requestUtils.baseURL + '/checkout' );
+		// Wait for the email field to be visible
 		await this.page.locator( '#email' ).waitFor();
 	}
 
 	async goToCart() {
-		await this.page.goto( '/cart' );
+		await this.page.goto( this.requestUtils.baseURL + '/cart' );
```
WunderBart, line 56: The `baseUrl` is automatically added by Playwright, no need to add it explicitly.

WunderBart (changes requested):
Left a couple of comments!

plugins/woocommerce/client/blocks/tests/e2e/utils/frontend/frontend-utils.page.ts
mikejolley: I'm aware its discouraged hence the lint disabling. 

However I noticed when debugging tests there are several API requests raised in quick succession after adding to cart, possibly due to something in product blocks, which cause an additional request to the `/cart` route. It was my intention to wait for these to finish before continuing because this was the only workaround I had any success with. 

Without this in place the redirect to checkout fails (sometimes) after `addToCart` and you are sent to the cart page. I think it redirects too early.

WunderBart: I think this works mainly because it introduces a slight delay, which might be just enough time for the request to fully propagate. While it’s fine if it helps stabilize the test, it’s not the ideal solution—we should have a clear way to determine when it's safe to proceed.  

Is there a specific endpoint we can check to confirm that the cart has been updated and is ready for new items?

cc: @kmanijak for some insight on how the add to cart request from the product collection is propagated 👀 🙏

WunderBart (approved):
I'm unblocking this one if it fixes the flakiness. Let's just remove those explicit `baseUrl`s [here](https://github.com/woocommerce/woocommerce/pull/56125#discussion_r1983635517)! 🙏

plugins/woocommerce/client/blocks/tests/e2e/utils/frontend/frontend-utils.page.ts
mikejolley: @WunderBart The `cart-item` route check should have been enough since that is the request that actually adds something to cart. And you can see in the failed test videos it does add something because it says "1 in cart" in the block. But I couldn't find anything else at fault.

WunderBart: I've been debugging it and found a weird race condition that I described in #56177. Basically, the backend does not seem to catch up to the speed the automation is adding the items to the cart, even if we wait for the `cart/items` request to finish. With 2 items added one by one, we often end up with just one of them, but the weird part is that sometimes the second one is overwriting the first one 🤷

samueljseay: Could it be the cart syncing we do in product-button? I adjusted this e2e code recently: https://github.com/woocommerce/woocommerce/pull/56125/files/1a133d036c5e1311af139c83f961804265de3e4f#diff-08c85d077ed71853aa7e713841707635d8fe92e71e84b9ae6f1ad8c5d2c16a73R39

We're about to change the endpoints being called to add to cart so it might change again: https://github.com/woocommerce/woocommerce/pull/56117/files#diff-c5b02e6435a239a6d14da3763adaf8956e615ae3f670a52a5c23e0dc0e337aadR78

mikejolley:
Not merging this as is. Looking for alternatives/the cause.

PR #56140, opened 2025-03-05
[Experimental] Blockified Add to Cart with Options block: update the Variation Selector block to support iAPI

### LINKED ISSUE #55696: Blockified Add to Cart with Options block: update the Variation Selector block according to the designs
We need to update the _Variation Selector (Experimental)_ block according to the designs in pfrEX1-HI-p2. The block is used inside the _Add to Cart with Options_ block when editing a variable product.

mdperez86:

Closes #55696

| Before | After |
| ------ | ----- |
|        |       |

6. Then click the new Add to Cart with Options block and from the toolbar click on the product type selector and select Variable Product 
image

7. A list of product attributes should be shown.
8. By default the attribute options should be shown as Pills but it can be changed to a Dropdown
9. Save the changes.
10. Go to the store and select a variable product. Any change made in the editor should be reflected in the frontend.

https://github.com/user-attachments/assets/6a23816d-6344-46d0-be0c-134846bac42a

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorAttributeName.php
```diff
@@ -23,4 +25,41 @@ class AddToCartWithOptionsVariationSelectorAttributeName extends AbstractBlock {
...
+		$content = '';
+
+		if ( isset( $attribute_name ) ) {
+			$classes_and_styles = StyleAttributesUtils::get_classes_and_styles_by_attributes( $attributes, array(), array( 'extra_classes' ) );
+	
+			$wrapper_attributes = get_block_wrapper_attributes(
+				array(
+					'class' => esc_attr( $classes_and_styles['classes'] ),
+					'for'   => esc_attr( 'attribute_' . sanitize_title( $attribute_name ) ),
+					'id'    => esc_attr( 'attribute_' . sanitize_title( $attribute_name ) . '_label' ),
```
Aljullu: Wondering if we should prefix the ids with the product slug, so if there are several Single Products blocks with different products using the same attributes we don't end up with duplicate ids.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/style.scss
```diff
@@ -0,0 +1,48 @@
...
+	border-radius: $grid-unit-40;
+	padding: 0.25em 1em;
+	text-transform: uppercase;
+	font-size: var(--wp--preset--font-size--small);
+	cursor: pointer;
+	user-select: none;
+
+	&:hover {
+		background-color: color-mix(in srgb, var(--wp--preset--color--contrast) 10%, transparent);
+	}
```
Aljullu, line 21: Not a blocker, but it would be nice to avoid showing the `:hover` styles in the editor.

[Enregistrament de pantalla del 2025-03-11 13-09-10.webm](https://github.com/user-attachments/assets/db0c4e4e-0830-4848-9b88-307574db237d)

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelector.php
Aljullu, line 1: There are several methods from this class that can now be removed, no?

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorItemTemplate.php
```diff
@@ -23,4 +24,99 @@ class AddToCartWithOptionsVariationSelectorItemTemplate extends AbstractBlock {
...
+	 * @param array    $attributes Block attributes.
+	 * @param WP_Block $block The Block.
+	 * @return string Row HTML
+	 */
+	private function get_product_row( $product_attribute_name, $product_attribute_terms, $attributes, $block ): string {
+		global $attribute_name;
+		global $attribute_terms;
+
+		$attribute_name = $product_attribute_name;
+		$attribute_terms = $this->get_terms( $product_attribute_name, $product_attribute_terms );
```
Aljullu: What are the pros and cons of introducing these globals? I wonder if we could use block context instead.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorAttributeOptions.php
```diff
@@ -16,11 +18,203 @@ class AddToCartWithOptionsVariationSelectorAttributeOptions extends AbstractBloc
...
+	 *
+	 * @param array    $attributes Block attributes.
+	 * @param string   $content Block content.
+	 * @param WP_Block $block Block instance.
+	 * @return string Rendered block output.
+	 */
+	protected function render( $attributes, $content, $block ): string {
+		global $product;
+		global $attribute_name;
+		global $attribute_terms;
```
Aljullu: `global $attribute_terms;` doesn't seem to be used in this method, no?

Aljullu (commented):
This is looking great, @mdperez86! I did a round of testing and everything works as described. I added some minor inline comments, but besides the one about migrating globals to block context, none of the others is a blocker IMO.

I would like if we can get a second review from the iAPI-side, though.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelector.php
mdperez86: Done

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorAttributeOptions.php
mdperez86: done

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorItemTemplate.php
mdperez86: done

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-name/block.json
```diff
@@ -54,5 +54,6 @@
 				"padding": true
 			}
 		}
-	}
+	},
+	"usesContext": [ "attributeId", "attributeName", "attributeTerms" ]
```
Aljullu: I think it would be good to prefix the context with the plugin name:

```suggestion
	"usesContext": [ "woocommerce/attributeId", "woocommerce/attributeName", "woocommerce/attributeTerms" ]
```

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/block.json
```diff
@@ -20,5 +20,6 @@
 	"$schema": "https://schemas.wp.org/trunk/block.json",
 	"supports": {
 		"inserter": false
-	}
+	},
+	"usesContext": [ "attributeId", "attributeName", "attributeTerms" ]
```
Aljullu: Same here. :slightly_smiling_face:

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
```diff
@@ -1,9 +1,152 @@
...
+					context.selected = '';
+				} else {
+					context.selected = context.option.value;
+				}
+				context.focused = context.option.value;
+			},
+			handleKeyDown( event: KeyboardEvent ) {
+				const context = getContext< PillsContext >();
+
+				let flag = false;
```
Aljullu: What do you think about adding a more description name to this variable? Something like `keyWasProcessed` or `shouldPreventEvent`.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorItemTemplate.php
```diff
@@ -23,4 +25,133 @@ class AddToCartWithOptionsVariationSelectorItemTemplate extends AbstractBlock {
...
+			return '';
+		}
+
+		// Get an instance of the current Post Template block.
+		$block_instance = $block->parsed_block;
+
+		$new_block = new WP_Block(
+			$block_instance,
+			array(
+				'attributeId'    => uniqid(),
```
Aljullu: I think HTML element ids shouldn't start with a number, and `uniqid()` can return values like `67d1b217a8603`. So we might want to prefix a string before the id.

Aljullu (approved):
I did another round of testing, and it's looking good on my end. Works as described and the code looks good. As mentioned, I think we still need an approval from the iAPI-side.

Besides that, I left some inline minor comments and questions.

Aljullu:
Btw, no need to block this PR because it's an experimental feature, but we will need to add tests to make sure we are correctly rendering the product attributes in the form when displaying a variable product.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/block.json
mdperez86: done

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
mdperez86: done

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorItemTemplate.php
mdperez86: done

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
luisherranz, line 1: This file should be divided into two: one for the pills and another for the dropdown, and then conditionally enqueue only the one that is being used.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
```diff
@@ -1,9 +1,152 @@
...
+type PillsContext = Context & {
+	tabIndex?: number;
+	selected?: string;
+	focused?: string;
+};
+
+store(
+	'woocommerce/add-to-cart-with-options-variation-selector-attribute-options__pills',
+	{
+		state: {},
```
luisherranz: There is no need to add an empty state.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
```diff
@@ -1,9 +1,152 @@
...
+		actions: {
+			handleClick() {
```
luisherranz: The names of the actions are better when they are more descriptive, as those actions could be used in different situations. For example, here it could be `actions.toggleSelected`.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
```diff
@@ -1,9 +1,152 @@
...
+				let keyWasProcessed = false;
+
+				switch ( event.key ) {
+					case ' ':
+						if ( context.selected === context.option.value ) {
+							context.selected = '';
+						} else {
+							context.selected = context.option.value;
+						}
+						context.focused = context.option.value;
```
luisherranz: This looks like a copy-paste of the code found in `handleClick`. Rename `handleClick` and reuse it here (for example, `actions.toggleSelected()`).

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
```diff
@@ -1,9 +1,152 @@
...
+
+				if ( ! context.selected && ! context.focused ) {
+					if ( context.options[ 0 ] === context.option ) {
+						context.tabIndex = 0;
+					}
+					return;
+				}
+
+				context.option.isSelected =
+					context.selected === context.option.value;
```
luisherranz: This should be derived state instead:

```js
get isSelected() {
    const { selected, option } = getContext();
    return selected === option.value;
}
```

More about derived state and how to use it in [this guide](https://developer.wordpress.org/block-editor/reference-guides/interactivity-api/core-concepts/undestanding-global-state-local-context-and-derived-state/).

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorAttributeOptions.php
```diff
@@ -16,11 +18,211 @@ class AddToCartWithOptionsVariationSelectorAttributeOptions extends AbstractBloc
...
+	 * @return string The pills.
+	 */
+	protected function render_pills( $attributes, $content, $block ) {
+		$attribute_id    = $block->context['woocommerce/attributeId'];
+		$attribute_name  = $block->context['woocommerce/attributeName'];
+		$attribute_terms = $block->context['woocommerce/attributeTerms'];
+
+		return sprintf(
+			'<div %s>
+				<template data-wp-each--option="context.options">
```
luisherranz: Why is it necessary to use `data-wp-each` here instead of a normal PHP loop? In other words, is `context.options` an array that can change dynamically on the frontend, or is it a static array?

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
```diff
@@ -1,9 +1,152 @@
...
+);
+
+type DropdownContext = Context & {
+	isSelected: 'selected' | undefined;
+};
+
+store(
+	'woocommerce/add-to-cart-with-options-variation-selector-attribute-options__dropdown',
+	{
+		state: {},
```
luisherranz: Not necessary if it's empty.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorItemTemplate.php
```diff
@@ -23,4 +25,133 @@ class AddToCartWithOptionsVariationSelectorItemTemplate extends AbstractBlock {
...
+	 */
+	private function get_product_row( $product_attribute_name, $product_attribute_terms, $attributes, $block ): string {
+		$attribute_name  = $product_attribute_name;
+		$attribute_terms = $this->get_terms( $product_attribute_name, $product_attribute_terms );
+
+		if ( empty( $attribute_terms ) ) {
+			return '';
+		}
+
+		// Get an instance of the current Post Template block.
```
luisherranz: Is this comment correct? Maybe it should mention "Variation Selector Item Template" instead of "Post Template"?

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorItemTemplate.php
```diff
@@ -23,4 +25,133 @@ class AddToCartWithOptionsVariationSelectorItemTemplate extends AbstractBlock {
...
+			$block_instance,
+			array(
+				'woocommerce/attributeId'    => 'wc_product_attribute_' . uniqid(),
+				'woocommerce/attributeName'  => $attribute_name,
+				'woocommerce/attributeTerms' => $attribute_terms,
+			),
+		);
+
+		// Render the inner blocks of the Post Template block with `dynamic` set to `false` to prevent calling
+		// `render_callback` and ensure that no wrapper markup is included.
```
luisherranz: Same here.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
```diff
@@ -1,9 +1,152 @@
...
+						const { ref } = getElement();
+						ref?.focus();
+					}
+					return;
+				}
+
+				context.tabIndex = -1;
+			},
+		},
+	}
```
luisherranz: Let's make these stores private.

```suggestion
	}, 
    { lock: true }
```

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
```diff
@@ -1,9 +1,152 @@
...
+		},
+		callbacks: {
+			init() {
+				const context = getContext< DropdownContext >();
+				if ( context.selected === context.option.value ) {
+					context.isSelected = 'selected';
+				}
+			},
+		},
+	}
```
luisherranz: Same here.


```suggestion
	},
    { lock: true }
```

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
```diff
@@ -1,9 +1,152 @@
...
+			handleChange() {
+				const context = getContext< DropdownContext >();
+				context.selected = context.option.value;
+			},
+		},
+		callbacks: {
+			init() {
+				const context = getContext< DropdownContext >();
+				if ( context.selected === context.option.value ) {
+					context.isSelected = 'selected';
```
luisherranz: Why is this needed? Cannot be `isSelected` derived? Also, at what point, other than here, is its value changing?

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorAttributeOptions.php
```diff
@@ -16,11 +18,211 @@ class AddToCartWithOptionsVariationSelectorAttributeOptions extends AbstractBloc
...
+		);
+
+		$options = array_merge(
+			array( $default_option ),
+			$attribute_terms
+		);
+
+		return sprintf(
+			'<select %s>
+				<template data-wp-each--option="context.options">
```
luisherranz: Same question here about `context.options`.

luisherranz (changes requested):
In general everything is fine, although I've added some comments with improvements.

I have also tried to remove `context.focused` because it seems quite close to `context.selected`, which simplifies the code a bit. There is a case where, after unselecting a pill that is not the first one, when you navigate backwards ("shift+tab"), it goes to the first pill, but other than that, it behaves the same. Example: https://github.com/luisherranz/woocommerce/commit/9da9e5ade291fc2867c93bec06bab71762c4dfe5

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
mdperez86: Well, I do agreed with the "more descriptive naming" approach, but not with the meaning of the name. This action is in fact an event, and the only reason why it might not be perceived like so is because we are not using the event parameter of the function. What about the `handleKeyDown( event: KeyboardEvent )` for example?
I think that if we want to be more descriptive then we should not relay only on descriptive naming but also on its meaning.
I would rather prefer to maintain the event function (to be explicit) and extract a shared functionality to a separate action with your suggested naming similarly to what I did in this follow up [PR](https://github.com/woocommerce/woocommerce/pull/56378/files#diff-baaa04fb4f93cdad22fe635ee17fdc572b9878c182c6cab5abcd5f33e29f19dbR35).

Addressed in [9c13835](https://github.com/woocommerce/woocommerce/pull/56140/commits/9c13835d2b0472867fa104a45a8c1835b209f0f8)

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
mdperez86: Addressed in [9c13835](https://github.com/woocommerce/woocommerce/pull/56140/commits/9c13835d2b0472867fa104a45a8c1835b209f0f8)

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
mdperez86: Addressed in [1f2b15f](https://github.com/woocommerce/woocommerce/pull/56140/commits/1f2b15f6f110c6a443ee224450b1409c518b9543)

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorAttributeOptions.php
mdperez86: No, it is not necessary. But why is a PHP loop necessary in this case?

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorItemTemplate.php
mdperez86: Done -> [c1a8673](https://github.com/woocommerce/woocommerce/pull/56140/commits/c1a8673d37a80537f79e11fa8b1e3ed2b1c9e4ca)

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorItemTemplate.php
mdperez86: Done -> [c1a8673](https://github.com/woocommerce/woocommerce/pull/56140/commits/c1a8673d37a80537f79e11fa8b1e3ed2b1c9e4ca)

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
mdperez86: done -> [c1a8673](https://github.com/woocommerce/woocommerce/pull/56140/commits/c1a8673d37a80537f79e11fa8b1e3ed2b1c9e4ca)

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
mdperez86: Done -> [c1a8673](https://github.com/woocommerce/woocommerce/pull/56140/commits/c1a8673d37a80537f79e11fa8b1e3ed2b1c9e4ca)

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
mdperez86: Done -> [57e4785](https://github.com/woocommerce/woocommerce/pull/56140/commits/57e4785dd7301400cf0e8fecea611ca71692ec11)

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
mdperez86: Yep, I think this is going to be solved when we extract common components to their own folder/place. I think the approach followed by @samueljseay [here] (https://github.com/woocommerce/woocommerce/blob/d6a254acf9794aa4386edc7d38d3b3b404b613a8/plugins/woocommerce/src/Blocks/InteractivityComponents/Dropdown.php) is a good one to continue with this. But I don't think this should be done as part of this PR.

mdperez86:
> In general everything is fine, although I've added some comments with improvements.
> 
> I have also tried to remove `context.focused` because it seems quite close to `context.selected`, which simplifies the code quite a bit. There is a case where, after unselecting a pill that is not the first one, when you navigate backwards ("shift+tab"), it goes to the first pill, but other than that, it behaves the same. Example: [luisherranz@9da9e5a](https://github.com/luisherranz/woocommerce/commit/9da9e5ade291fc2867c93bec06bab71762c4dfe5)

Exactly, that was my initial implementation but, I found it a rare behavior, that's why I change it to allow this case.

Please take another look, I was able to made some adjustments based on your suggestions and the code know looks better.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorAttributeOptions.php
luisherranz: To generate the pills. Something like this:

```php
$pills = '';
foreach ( $attribute_terms as $attribute_term ) {
	$pills .= sprintf(
		'<div %s>%s</div>',
		$this->get_normalized_attributes(
			array(
				'role'                       => 'radio',
				'class'                      => 'wc-block-add-to-cart-with-options-variation-selector-attribute-options__pill',
				'data-wp-bind--tabindex'     => 'state.pillTabIndex',
				'data-wp-bind--aria-checked' => 'state.isPillSelected',
				'data-wp-watch'              => 'callbacks.watchSelected',
				'data-wp-on--click'          => 'actions.handleClick',
				'data-wp-on--keydown'        => 'actions.handleKeyDown',
			),
		),
		$attribute_term['label']
	)
}
```

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
luisherranz: If your logic is not coupled to an event, you don't need to have a separate action, and you shouldn't call `event.preventDefault()` unnecessarily.

By the way, for accessibility, shouldn't the pills be buttons instead of divs?

luisherranz:
> Exactly, that was my initial implementation but, I found it a rare behavior, that's why I change it to allow this case.

Is it a personal preference or an accessibility concern?

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
mdperez86: No, they are radio buttons using the [Roving TabIndex pattern](https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio/).

mdperez86: >If your logic is not coupled to an event, you don't need to have a separate action

This is a semantic problem an not related to where the logic should be placed. If we have `'data-wp-on--click'` then whatever the value we assign, at the end, it'll always be an event listener and not an action.
I know that, in this case and like many others, we can replace the event listener by a direct call to the action function just because the listener will have no other logic rather than calling the action. But I think this is a bad practice that lead us to bypass layers of responsibilities.
I don't have any problem to implement this in the way you suggested but I'm pretty sure I'll have to revert it back to an event listener once we start to collect analytics data.

>you shouldn't call event.preventDefault() unnecessarily.

That is true.

Addressed -> [2cc688d](https://github.com/woocommerce/woocommerce/pull/56140/commits/2cc688d951c9879c4e074055157bddfa4ac2a6ef)

mdperez86:
> > Exactly, that was my initial implementation but, I found it a rare behavior, that's why I change it to allow this case.
> 
> Is it a personal preference or an accessibility concern?

It was a personal perception that does not affect accessibility. In my opinion if we are focusing on a pill within a pills group that is different to the first one and we uncheck it, and after that we press Shift+Tab then we should move the focus to a different element outside of the group, not to the first element within the same group.

Since this is a brand new component with a specific behavior we can always continue the discussion with the UX team and create an issue to implement whatever is needed to make this working. We can address this when we move the component to its own place as mentioned in https://github.com/woocommerce/woocommerce/pull/56140#discussion_r1999453242

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
luisherranz: > No, they are radio buttons using the [Roving TabIndex pattern](https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio/).

Ok, nice. ~Then, you seem to be missing the `aria-checked` attribute?~ My bad, it's already there 🙂

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorAttributeOptions.php
mdperez86: Addressed in [58d0ecb](https://github.com/woocommerce/woocommerce/pull/56140/commits/58d0ecba986006b5e32b4cb23a6f0866597744e6)

luisherranz (approved):
Thanks for the changes, Maikel 🙂

helgatheviking:
Question... is there on the front end to know what variation ID has been selected? Third parties (it's me, I'm third parties) are going to need to listen for changes to the variation ID and do things when a specific variation is selected. The classic templates had a lot of custom events and already showed the variation price/stock changes on selection. (gonna test this out shortly)

update- looked like that question is being explored in https://github.com/woocommerce/woocommerce/issues/56870

PR #56142, opened 2025-03-05
First pass at a block.json for cart block. Closes #55985.

### LINKED ISSUE #55985: The Cart block does not have a block.json file
As the title says, there is no `block.json` file for the Cart block. Since all the other blocks do, this is an outlier, I am not sure if there's a historical reason for this, but it would be good to align with what we have.

This would be useful to allow third party developers to filter metadata, which is currently not possible on the cart block.

The settings from `index.js` should be moved into `block.json` and the block registered using that.

helgatheviking:
Closes #55985 

For some reason the cart block doesn't have a `block.json` this is an attempt to add one.


plugins/woocommerce/changelog/issues-55985-cart-block
senadir, line 1: should be renamed to just `plugins/woocommerce/changelog/issues-55985-cart-block`

plugins/woocommerce/changelog/56128-undefined-array-key-PaymentUtils
senadir, line 1: sounds like you need a rebase, this file shouldn't be here.

gigitux:
Sorry @senadir – I think I misclicked and removed you as a reviewer instead of myself 😅

senadir (dismissed):
One issue I'm seeing is that this doesn't load translation it seems for the block in the editor

Right now:
image

Expected result:
image


We have seen this issue before when moving to block.json, I think this will get fixed on the next release, as I don't see any special handling for Checkout block who has block.json


Please add some testing steps like "Add cart, ensure it works"...

senadir:
Hey @helgatheviking still feel like bringing this PR to the finish line?

helgatheviking:
@senadir going AFK until mid June at earliest. Happy to let somebody else take it over.

helgatheviking:
@senadir refreshed this

Issue #56090, opened 2025-03-04
[e2e] Some Product Filter and Product Collection e2e tests failing on WP-1


tammullen:
Since March 1st the following Block e2e tests have been failing when ran with WordPress L-1 version

product-filters/attribute-filter-frontend.block_theme.spec.ts
product-collection/extensibility-events.block_theme.spec.ts

Example results: https://a8c-woo-test-reports.s3.amazonaws.com/reports/daily-checks/20250304-trunk/blocks-e2e-report-wp-latest-1/index.html 

Possibly related to https://github.com/woocommerce/woocommerce/pull/55080?

gigitux:
@samueljseay @luisherranz I added the `priority:high` label since this issue needs to be fixed before the WooCommerce 9.8 release. I also assigned it to you, as it seems you’re already working on it!

luisherranz:
I've opened these two fixes related to WP 6.6: 

- https://github.com/woocommerce/woocommerce/pull/56131
- https://github.com/woocommerce/woocommerce/pull/56154

luisherranz:
- https://github.com/woocommerce/woocommerce/pull/56154

I've tested this PR locally with WP 6.6, and it fixes the `emits wc-blocks_product_list_rendered event on init and on page change` test.

luisherranz:
~However, I can't reproduce the `woocommerce/product-filter-attribute - Frontend > With default display style` errors locally when testing with WP 6.6 in `trunk`.~

I was able to reproduce now, and indeed this PR fixes the `woocommerce/product-filter-attribute - Frontend > With default display style` tests:

- https://github.com/woocommerce/woocommerce/pull/56131

PR #56154, opened 2025-03-06
iAPI: Workaround for bug in WP 6.6's iAPI during Product Collection navigation

luisherranz:

This PR introduces a workaround for a bug that is present in the version of the Interactivity API included in WordPress 6.6.

The bug in question is the evaluation of a getter outside the scope when merging the state properties that are serialized in the HTML during a client-side navigation. In question, in [this line](https://github.com/WordPress/gutenberg/blob/a74a70ed203bbcbb5f3d34335df9631178647acb/packages/interactivity/src/store.ts#L28). When `target[key]` contains a getter, it is evaluated outside the scope of a directive and it fails, causing an error and stopping the application.

The workaround intercepts and modifies the serialized state to remove the derived state value sent from the server (an empty object) to prevent [that line](https://github.com/WordPress/gutenberg/blob/a74a70ed203bbcbb5f3d34335df9631178647acb/packages/interactivity/src/store.ts#L28) from being reached for the following getter: `store( 'woocommerce/product-button' ).state.addToCartText`

In WordPress 6.7, it does not modify the current behavior. This is done by checking the ID of the script that contains the serialized state, which in 6.6 was `wp-interactivity-data` and in WordPress 6.7 is `wp-script-module-data-@wordpress/interactivity`.

1. Add a Product Collection block that has a Product Button in the template.
2. Go to the frontend and use client-side navigation to navigate.
3. Check that there are no errors in the console.

Now test with WP 6.6.2

1. Add this to your `.wp-env.override.json` file: `{ "core": "WordPress/WordPress#6.6.2" }`
2. Restart `wp-env`.
3. Repeat the first steps.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/frontend.tsx
```diff
@@ -59,7 +87,13 @@ const productCollectionStore = {
 					'@wordpress/interactivity-router'
 				);
 
-				yield actions.navigate( ref.href );
+				// Todo: Remove after support for WP .6.6 is dropped.
+				if ( document.getElementById( 'wp-interactivity-data' ) ) {
+					const html = yield fetchUrlAndReplaceState( ref.href );
```
DAreRodz, line 97: Maybe we could implement a minimal cache to avoid fetching the HTML again before calling `navigate`. 🤔

DAreRodz (commented):
Thanks @luisherranz! I think this fix is the best we can do without modifying the Interactivity API implementation.

I've left a comment below. 👇

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/frontend.tsx
DAreRodz: Another option could be calling `actions.prefetchOnHover` synchronously, i.e., without `yield`. That would fetch the HTML again, but it won't replace the cached one if it was already cached.

That would prevent waiting for the HTML again if it was already cached.

```suggestion
					actions.prefetchOnHover();
```

DAreRodz:
@luisherranz, I've just updated the WordPress version in the `.wp-env.override.json` snippet you shared (it pointed to `6.7.2`). After doing so, I saw errors in the console. 😬 

```
Uncaught (in promise) TypeError: Cannot destructure property 'animationStatus' of 'getContext(...)' as it is undefined.
    at get addToCartText (frontend.tsx:87:13)
    at Reflect.get (<anonymous>)
    at signals_core_module_y.x (interactivity.js?ver=6.6.2:61:1786)
    at signals_core_module_y.h (interactivity.js?ver=6.6.2:57:3969)
    at signals_core_module_y.get (interactivity.js?ver=6.6.2:57:4790)
    at Object.get (interactivity.js?ver=6.6.2:61:1826)
    at deepMerge (interactivity.js?ver=6.6.2:83:20)
    at store (interactivity.js?ver=6.6.2:307:5)
    at interactivity.js?ver=6.6.2:327:7
    at Array.forEach (<anonymous>)
```

luisherranz:
Oh, there was another prefetch. Yeah, I didn't catch that because I only had two pages, thanks 🙂

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/frontend.tsx
luisherranz: To be honest, I wouldn't optimize anything since this will only be necessary during one version of WooCommerce, since when the next one after that comes out, it will no longer need to support WordPress 6.6.

DAreRodz:
Another thing I've seen is that every time you hover over a link, the HTML is fetched. Maybe it's worth implementing a cache to avoid multiple requests to the same URL. 🤔

luisherranz:
Ok. Done in https://github.com/woocommerce/woocommerce/pull/56154/commits/6e92897626a7205eed9d266717610cea297b3007.

luisherranz:
I've tested this PR locally with WP 6.6, and it fixes the `emits wc-blocks_product_list_rendered event on init and on page change` test reported [here](https://github.com/woocommerce/woocommerce/issues/56090).

luisherranz:
@DAreRodz, @samueljseay This is ready for review.

samueljseay (approved):
The implementation looks straightforward to me. Todos are a fine trade-off. There are some aspects like the caching I don't know much about so I'll leave that to you folks.

I smoke tested various iAPI blocks and I didn't see the errors (or any additional errors) in WP6.6.

DAreRodz:
> There are some aspects like the caching I don't know much about so I'll leave that to you folks.

@samueljseay, that's just to avoid multiple fetch requests to the same resources. The iAPI has its own logic for this, but the workaround is fetching the HTML outside of the iAPI runtime and passing it to the `prefetch` and `navigate` calls directly.

That's why we added this minimal cache. 🙂

samueljseay:
@DAreRodz Are we ok to merge this or did you want to do another review?

DAreRodz:
@samueljseay we're OK, let's merge the PR. ☺️

luisherranz:
Thanks, folks! I've merged it now.

Let's keep an eye to remove it once WP 6.6 is not supported anymore.

PR #56156, opened 2025-03-06
[e2e tests] Refactor usage of rest api client

### LINKED ISSUE #55315: [e2e tests] Remove the api authentication by consumer token
E2E tests are not consistent in using API authentication. We see both basic with username and password and the use of consumer token. Generating (and removing) the token is done via the UI in a global setup and teardown, taking time and adding another point of failure.

We should drop the authentication by token and remove the setup and teardown and only use basic authentication everywhere.

To still cover the creation and removal of a token we should create a test to check that.

adimoldovan:

Closes #55315
Closes https://github.com/woocommerce/woocommerce-quality/issues/737

Standardises API client usage across all end-to-end tests, improving both execution speed and long-term maintainability.

We had different inconsistent instances and a lot of duplication:
- some tests creating a new instance with token auth
- some tests using the fixture (which used token auth)
- some tests creating a new instance with basic auth
- different other instances across utils
- some tests and utils are using Playwright's `request` - this was not in scope of this PR.

Main changes:
- refactored the api-client to expose a static instance. The instance is created if it doesn't already exists and uses basic authentication.
- consolidated `api`, `wcAdminApi` and `wpApi` fixtures into a basic `restApi` fixture.
- removed the consumer token setup and teardown, because the token is no longer used.
- added a new test `settings/consumer-token.spec.js` to validate the flow of token management.
- refactored the tests and utils that were using the api to use the new `restApi` fixture or the static instance directly.

E2E tests pass in CI

E2E tests passed in CI


plugins/woocommerce/tests/e2e-pw/utils/settings.js
```diff
@@ -15,7 +17,7 @@ function resolvePath( path ) {
  * @return {Promise<void>} A promise that resolves when the update is complete.
  */
 export async function updateValue( path, desiredValue ) {
```
lanej0, line 19: (not related to this PR) why are we doing things like this?  Replace one line of code with another line of code. Seems like needless abstraction.

lanej0 (approved):
This was a beast of a PR to review.  I don't know if we could have broken it up across smaller changesets though 🤷🏼   

The changes look good, and it's a great move towards greater consistency in the tests.  

CI all passed, I am getting failures locally when I run various tests -- but that seems more likely that it's a local setup thing as ALL the CI runs passed.

PR #56158, opened 2025-03-06
Marketplace: better update the update badge count in My Subscription tab

### LINKED ISSUE #46484: My subscriptions notification badge is not synced automatically in some cases in the WooCommerce Extensions page
### Prerequisites

### Describe the bug

The notification badge that appears next to `My subscription` in `WooCommerce > Extensions` isn't synced automatically, and some times it can cause confusion to the users.

You can spot this behaviour when you install a plugin in WooCommerce 8.8.0 (which contains the Woo Update Manager) and it raises a notification for whatever reason (for example, and old version that requires an update but you can't update it because you have already installed the extension in other sites).

In this case, when you uninstall the plugin that triggered that notification, you can see how the notification doesn't change. This is because it's not synced automatically, and you need to refresh the subscriptions in-app and then refresh the browser so that you can see the changes.

This behaviour can be confusing to our users, so it would be great if we could find the way to automatically sync the notification badge next to `My subscriptions` in order to improve the user experience.

### Expected behavior

The number inside the notification badge next to `My subscriptions` should be decreased by 1 compared to the one noted in the step 5 of the testing instructions after uninstalling AutomateWoo.

If the number was 1 prior to uninstalling the plugin, now the notification badge should disappear.

### Actual behavior

The notification badge next to `My subscriptions` looks exactly the same than prior to uninstalling AutomateWoo. You need to refresh your subscriptions in the `My subscriptions` page and then refresh the page so that you can see the notification badge updated accordingly.

### Steps to reproduce

1. Make sure you have 2 sites with a plugin from WooCommerce.com like AutomateWoo connected to your WooCommerce.com account (I think this is the maximum number of sites you can connect a paid plugin to WooCommerce.com with your account simultaneously). We need to make sure you have spent the number of simultaneous installations for a WooCommerce.com plugin.
2. Create a new site with WooCommerce 8.8.0.
3. Install an outdated version of AutomateWoo on this site too (which would exceed the number of plugins installations that you can connect to WooCommerce.com).
4. Connect this site to WooCommerce.com, which should install the Woo Update Manager plugin.
5. Go to `WooCommerce > Extensions > My Subscriptions` and make sure there is an entry for AutomateWoo in the `Installed on this store` section. Also make sure there is a red notification badge next to `My subscriptions`. This notification badge should have a number inside, note this number.
6. Go to `WooCommerce > Plugins > Installed Plugins` and uninstall AutomateWoo.
7. Go again to `WooCommerce > Extensions > My Subscriptions`.

### WordPress Environment

Any site with WooCommerce 8.8.0.

### Isolating the problem

raicem:

The badge shown in the My Subscriptions tab is the number of outdated WooCommerce.com products. 

image

Whenever a user deactivates or activates a Woo plugin we ping WooCommerce.com to establish a connection for the subscription. As a result, we flush cache to re-calculate the pending updates for WooCommerce.com products. 

However, the timing of this cache was not ideal. 

- Activation: if a subscription is maxed out or not found, we _don't_ flush the cache. So product becomes active in the local store, but we pretend it's not. We should start showing a badge if the product is out of date, but we don't.
- Deactivation: if subscription wasn't connected in the first place, we never flush the cache as a result of deactivation. Users expecting update badge to disappear, but it doesn't.

As a result, the `_woocommerce_helper_updates_count` transient is never flushed and  `WC_Helper_Updater::get_updates_count` returns stale data. 

Also, when we calculate the number of pending updates, we also count deactivated plugins as well. This PR also changes that behavior and they are not counted towards the badge.

Closes #46484.

1. Checkout this branch and get a new build `pnpm install && pnpm build`
2. We need to max out subscription connections first. From WooCommerce.com My Subscriptions page, share that paid subscription to two other other accounts. This will max out all connections.
    - To confirm, WCCOM My Subscription will show `Not available - shared with 2 others` message 
3. Start your local WooCommerce dev environment. Connect it to the same WooCommerce.com account. 
4. Download the outdated version of the paid plugin you have subscription to and upload it to your test site. 
    - Please ping me if you need assistance in finding the outdated versions. 
5. After activating the plugin, go to[ My Subscription tab ](http://localhost:8888/wp-admin/admin.php?page=wc-admin&term=afdfd&tab=my-subscriptions&path=%2Fextensions)and confirm you see a badge next to the tab.
   - if you don't have WooCommerce Update Manager installed, the badge starts from 1. So +1 for missing WUM and then +1 for oudated plugin. 
6. Hover the badge to read the explanation. 
    - Unfortunately, if the badge is there for WUM, the explanation is not great. 
7. Deactivate the plugin and go back to [the My Subscriptions tab](http://localhost:8888/wp-admin/admin.php?page=wc-admin&term=afdfd&tab=my-subscriptions&path=%2Fextensions) again. Confirm the badge is gone. 
8. Repeat the same steps with a paid theme as well.
9. Perform a search and hover over the badges next to the tabs such as Extensions and Themes. Confirm the explanation now includes the number as well.
10. Please test around this issue.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 
Better updates the pending updates count in the Marketplace My Subscriptions tab.
</details>

<details>

#### Comment 

</details>


plugins/woocommerce/includes/admin/helper/class-wc-helper-updater.php
```diff
@@ -690,6 +690,10 @@ public static function get_updates_count() {
 				continue;
 			}
 
+			if ( ! is_plugin_active( $plugin['_filename'] ) ) {
+				continue;
+			}
```
raicem, line 699: This is a change that I need to double check with the team. This will **not** count deactivated products in the update badge.

plugins/woocommerce/client/admin/client/marketplace/components/tabs/tabs.tsx
```diff
@@ -98,6 +98,28 @@ const renderTabs = (
...
+									: __(
+											'Number of matching items in this category',
+											'woocommerce'
+									  )
+							}
+							aria-label={
+								tabKey === 'my-subscriptions'
+									? __(
+											'Number of updates available for your subscriptions',
+											'woocommerce'
```
KokkieH: Voiceover is reading out this text without quoting the actual number, which is confusing.

Screenshot_2025_07_17__12_44

We should include the actual count here as well for this screen reader text to make sense.

KokkieH (changes requested):
It works as described for plugins and themes, but there's a change needed to the aria-label to make it less confusing.

plugins/woocommerce/client/admin/client/marketplace/components/tabs/tabs.tsx
raicem: Thanks for the suggestion, updated in: [`7908479` (#56158)](https://github.com/woocommerce/woocommerce/pull/56158/commits/79084799bd9b44b94368946fc9b6dbff1e8673a1)

KokkieH (approved):
Thanks for the change. Screen reader text is much clearer now.

PR #56174, opened 2025-03-06
Improve WooPayments onboarding experiment error handling

elazzabi:

Add a caching mechanism and error handling for the WooPayments onboarding experiment to prevent repeated failures and provide more robust experiment tracking.

(For Bug Fixes) This bug happens in specific cases. So far, only @elazzabi and @anu-rock found it during testing. We wanted to fix this once and for all following the PR https://github.com/woocommerce/woocommerce/pull/55981 (same checks, same logic)

If you are encountering this bug, the WooCommerce > Settings > Payments view will be broken:

Screenshot 2025-03-06 at 09 36 18

1. Get this PR
2. Refresh the Payments tab
3. You should see a list of payment methods now

Screenshot 2025-03-06 at 10 46 43

5. Check that a transient is created and the value is `error`

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 
Improve WooPayments onboarding experiment error handling using a transient to store the API response in case of an error.

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Internal/Admin/Settings/PaymentProviders/WooPayments.php
```diff
@@ -196,11 +196,25 @@ public function get_onboarding_url( WC_Payment_Gateway $payment_gateway, string
...
 		// If the experiment is active and the store is in the treatment group, we will do live onboarding.
 		// Otherwise, we will do test-account-first onboarding (control group).
-		if ( ! $live_onboarding && Experimental_Abtest::in_treatment( 'woocommerce_payment_settings_onboarding_2025_v1' ) ) {
-			$live_onboarding = true;
+		try {
+			// If we have a cache entry that indicates an error, disable the live onboarding.
+			if ( 'error' === $cached_result ) {
+				$live_onboarding = false;
+			} else {
+				$live_onboarding = ! $live_onboarding && Experimental_Abtest::in_treatment( 'woocommerce_payment_settings_onboarding_2025_v1' );
```
anu-rock: Won't this will disable live onboarding for merchants **who are eligible** for it but are not in the experiment's treatment group? You probably didn't intend that?

anu-rock (approved):
Code changes are looking good and tested well. Pre-approving with a logic-related comment.
___

#### Review checklist

| Parameter  | Result |
| Approach  | ✅ All good |
| Implementation  | ✅ Mostly good, left a comment |
| Automated tests  | ⚠️ Missing but optional |
| Naming  | ✅ All good |
| Typing  | ✅ All good |
| Changelog  | ✅ All good |
| Documentation  | Not applicable |

#### Scenarios tested

| On `trunk`, I visited the new Payment settings page  | ✅ The payment providers list was empty. | Screenshot 2025-03-07 at 5 46 40 PM |
| On this branch, I visited the new Payment settings page  | ✅ The payment providers list was displayed as expected. A transient named `wc_experiment_failure_woocommerce_payment_settings_onboarding_2025_v1` was created with "error" as its value. | Screenshot 2025-03-07 at 5 46 15 PM |

plugins/woocommerce/src/Internal/Admin/Settings/PaymentProviders/WooPayments.php
elazzabi: Hmm I'm not sure about that. I kept the original check:

```php
if ( ! $live_onboarding && Experimental_Abtest::in_treatment( 'woocommerce_payment_settings_onboarding_2025_v1' ) ) {
  $live_onboarding = true;
}
```

Just made it a little shorter
```php
$live_onboarding = ! $live_onboarding && Experimental_Abtest::in_treatment( 'woocommerce_payment_settings_onboarding_2025_v1' );
```

So I think it should match the earlier behavior? 🤔

anu-rock: The earlier check was setting `$live_onboarding` to true in a specific condition. In the new check, `$live_onboarding` may be set as true or **false**. I think that's a side effect when `$live_onboarding` is already true because a merchant meets the eligibility criteria in: https://github.com/woocommerce/woocommerce/blob/4e99a8bfd252ce5881fd6c72372a3e4583298a1d/plugins/woocommerce/src/Internal/Admin/Settings/PaymentProviders/WooPayments.php#L180-L197.

dmallory42: I spoke to @elazzabi about this on Slack and I think the only case in which the behaviour is different from the original behaviour is when an exception is thrown during the experiment fetching (essentially only when the wordpress API is unavailable. In that case, the `live_onboarding` gets set to false, which is equivalent to the merchant being in the control group as they see the test drive flow. 

I _think_ this is how the logic works, however I'd feel more comfortable getting a second pair of eyes from @vladolaru as he's more familiar with this particular logic than me.

vladolaru: @anu-rock is right on this one. If something fails, we should remember the failure and decide _nothing_ - whatever was decided before, goes. If our routing logic decided it should do live onboarding, so be it; if it decided to do test account onboarding, the same.

To make things clearer, I refactored a little bit and adjusted the code [here](https://github.com/woocommerce/woocommerce/pull/56174/commits/5b75de3160c8cb61a237309ce40257b0ceaaef3c). I also made sure that we don't read the transient unnecessarily (when live onboarding is true).

elazzabi: That makes sense! Thank you for working on it @vladolaru 🙇

PR #56186, opened 2025-03-06
Remove pending sync order count when no sync is in progress

### LINKED ISSUE #54860: [Performance]: Slow orders_pending_sync query when loading Advanced Features Settings Page
### Prerequisites

### Performance Details

The query run by the `Orders\DataSynchronizer::get_current_orders_pending_sync_count()` is a slow and complex query taking multiple seconds to run on sites with a large order volume.  The query is run every time the `wp-admin/admin.php?page=wc-settings&tab=advanced&section=features` page is requested, regardless of the state of the order storage settings.

https://github.com/woocommerce/woocommerce/blob/ccc6eb86c76bafdb642f09c5f694e801dee9e414/plugins/woocommerce/src/Internal/DataStores/Orders/DataSynchronizer.php#L472

### Affected Sites

Sites with > 500k orders will begin to see a degradation of this count query.

### Suspected causes

_No response_

## Suggested Solutions

* Only retrieve and display the full count of out of sync orders if the Sync Orders Job is actively running.
* Otherwise, just run an exists check for out of sync records and base the interface options on whether out of sync orders exist.

Image

joshuatf:

This PR adds the ability to check for the existence of any pending sync orders instead of forcing a count on all rows.  Because this query needs to run on almost every page load for this feature page, this significantly reduces the load time.  On a testing site with ~178k orders, the previous query typically clocked in around ~400ms and the new query takes approximately ~1.5ms.  This will have an even larger impact as the site scales.

The trade-off here is that we will not have access to the pending sync order count until the process has begun, but after some discussion on Slack, while I believe this information is nice to have, it does not ultimately affect the decision of whether or not to begin the sync.

Closes #54860 .

| Before | After |
| ------ | ----- |
|    Screenshot 2025-03-06 at 12 17 19 PM  |   Screenshot 2025-03-06 at 2 45 52 PM |

Note that by default, order count is no longer shown, but instead a message indicating that pending orders do exist.
Screenshot 2025-03-06 at 12 17 44 PM

When a sync has begun, you will still see the total count pending sync.
Screenshot 2025-03-06 at 12 14 46 PM

1. Navigate to `/wp-admin/admin.php?page=wc-settings&tab=advanced&section=features`
2. Make sure that the option to keep orders in sync is disabled
3. Create some orders
4. Navigate to `/wp-admin/admin.php?page=wc-settings&tab=advanced&section=features` again
5. Note that the text "There are currently orders out of sync." is shown next to the "Order data storage" description.
6. Click "Sync orders now"
7. Note that the number of pending sync orders is now shown
8. When complete, note that the message "There are currently orders out of sync" is no longer shown and you can switch between storage options.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/src/Internal/DataStores/Orders/CustomOrdersTableController.php
+++ b/plugins/woocommerce/src/Internal/DataStores/Orders/CustomOrdersTableController.php
@@ -460,7 +460,7 @@ class CustomOrdersTableController {
 			return 'no';
 		}
 
-		$sync_is_pending = 0 !== $this->data_synchronizer->get_current_orders_pending_sync_count();
+		$sync_is_pending = $this->data_synchronizer->has_orders_pending_sync();
 		if ( $sync_is_pending && ! $this->changing_data_source_with_sync_pending_is_allowed() ) {
 			throw new \Exception( "The authoritative table for orders storage can't be changed while there are orders out of sync" );
 		}
@@ -612,7 +612,7 @@ class CustomOrdersTableController {
 
 		$get_disabled = function () {
 			$compatibility_info = $this->features_controller->get_compatible_plugins_for_feature( 'custom_order_tables', true );
-			$sync_complete      = 0 === $this->data_synchronizer->get_current_orders_pending_sync_count();
+			$sync_complete      = ! $this->data_synchronizer->has_orders_pending_sync();
 			$disabled           = array();
 			// Changing something here? You might also want to look at `enable|disable` functions in Automattic\WooCommerce\Database\Migrations\CustomOrderTable\CLIRunner.
 			$incompatible_plugins = $this->plugin_util->get_items_considered_incompatible( 'custom_order_tables', $compatibility_info );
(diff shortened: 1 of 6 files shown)


plugins/woocommerce/src/Internal/DataStores/Orders/CustomOrdersTableController.php
```diff
@@ -612,7 +612,7 @@ private function get_hpos_setting_for_feature() {
 
 		$get_disabled = function () {
 			$compatibility_info = $this->features_controller->get_compatible_plugins_for_feature( 'custom_order_tables', true );
-			$sync_complete      = 0 === $this->data_synchronizer->get_current_orders_pending_sync_count();
+			$sync_complete      = ! $this->data_synchronizer->has_orders_pending_sync();
```
joshuatf, line 615: A number of these places are not using the first argument of `$use_cache` and result in re-querying the pending orders each time.

I suppose this is probably necessary to keep up-to-date information for switching between storage versions.

This is less of a problem with the new query that is significantly more performant, but if we wanted to go a step further, we could rely on cached values and bust this case on order create/update/delete.  However, I was a little hesitant to do this since this count is important to avoid switching storage options which could result in data corruption if orders are pending.

prettyboymp (approved):
I tested while having ~1.25M out of sync orders and the query went from taking 3.5s to 0.001s.  I expected a big improvement, but that is huge. Great work.  

Just need to update the failing test.

PR #56193, opened 2025-03-07
[dev] Remove unused @woocommerce/woocommerce-rest-api dev dependency

adimoldovan:

Following up on #56156, the `@woocommerce/woocommerce-rest-api` dependency is no longer in use.

CI needs to be green.

CI green.


plugins/woocommerce/package.json
```diff
@@ -703,7 +715,6 @@
 		"@woocommerce/api": "workspace:*",
 		"@woocommerce/e2e-utils-playwright": "workspace:*",
 		"@woocommerce/eslint-plugin": "workspace:*",
-		"@woocommerce/woocommerce-rest-api": "^1.0.1",
```
kalessil, line 706: Is it feasible to drop it from blocks package dependencies as well?

adimoldovan: Not sure, I only worked on removing it from core. 
Doing a quick search it looks like it's not used in blocks either. 
I'll try a remove it and see if breaks anything.

adimoldovan:
Closing in favour of https://github.com/woocommerce/woocommerce/pull/56203

plugins/woocommerce/package.json
kalessil: Thanks for trying!

adimoldovan: Opened a new PR, I messed something up in this one: https://github.com/woocommerce/woocommerce/pull/56203
CI passed, I think we're good.

PR #54911, opened 2025-01-28
[Monorepo] Merge Blocks Library into the WooCommerce Plugin

WunderBart:
## What?

Tracking issue: https://github.com/woocommerce/woocommerce/issues/54973
Announcement: pdV5qK-Re-p2
CFT: pdV5qK-Qf-p2

This pull request serves as the primary integration point for merging the `@woocommerce/block-library` client files into the `@woocommerce/woocommerce-plugin` workspace. It encompasses the following stages:

1. **Part 1: Relocate Files** (7e5fdc3)
   Relocates existing blocks client files (`plugins/woocommerce-blocks`) to their new locations within the WooCommerce plugin (`plugins/woocommerce/client/blocks`), preparing for subsequent updates and documentation enhancements.
2. **Part 2: Update Files** ([#54919](https://github.com/woocommerce/woocommerce/pull/54919) → adb9947)
   Updates the relocated files to ensure compatibility and functionality within the woocommerce plugin, addressing any necessary code adjustments and improvements.
3. **Part 3: Update Documentation** ([#54971](https://github.com/woocommerce/woocommerce/pull/54971) → e9b4d36)
   Revises and updates the documentation to reflect the new file structure and any changes introduced during the relocation and update processes.

By integrating these stages into a single PR, we aim to streamline the transition of the Blocks Client into the monorepo.

## Testing Instructions

As described in #54919.
The documentation part has already been reviewed in #54971.

PR #56195, opened 2025-03-07
[dev] Remove composer from Blocks client

WunderBart:
## What?

(Follow up to https://github.com/woocommerce/woocommerce/pull/54911)

Remove composer from Blocks client, which includes:
- changelogger,
- phpcs
- phpunit

### changelogger

Since #54911, Blocks became a child of the WooCommerce plugin, which means any changes to Blocks are also changes to the plugin. Adding the changelog via the Blocks Library is no longer valid and will produce the following error in CI:

```
PR touches file plugins/woocommerce/client/woocommerce/changelog/my-changelog-entry, marking plugins/woocommerce as touched.
---
Error: Project plugins/woocommerce is being changed, but no change file in plugins/woocommerce/plugins/woocommerce/changelog/ is touched!
```

For any current and future Blocks changes, we should only be using the plugin's changelogger via `pnpm --filter='@woocommerce/plugin-woocommerce' changelog add`.

### phpcs & phpunit

We no longer have php files inside the Blocks Library as it's just client files now.

## How to test

CI should pass.


kalessil (commented):
I guess we can drop both composer.json and composer.lock in this case - I'm not aware of any remaining use for those two files.

plugins/woocommerce/client/blocks/.wp-env.json
```diff
@@ -21,15 +21,5 @@
 				"wp-content/themes/twentytwentyfour": "https://downloads.wordpress.org/theme/twentytwentyfour.latest-stable.zip"
 			}
 		}
-	},
-	"config": {
```
kalessil, line 25: I'm not sure about this one, but if CI passes, we are good.

kalessil (approved):
Good one!

plugins/woocommerce/client/blocks/.distignore
```diff
@@ -24,14 +24,10 @@
 .wp-env.json
 babel.config.js
 blocks.ini
-composer.json
```
kalessil, line 27: .distignore can be obsolete as well as we don't build any zips for it separately.

PR #56199, opened 2025-03-07
Improve woo content styles in block email

### LINKED ISSUE #56020: Apply email styles to captured Woo content
In https://github.com/woocommerce/woocommerce/pull/55940, we introduced a simple block that is a placeholder for the email content generated by Woo when sending.

The content that is captured during the sending. It carries own inline styles. In this task we want to apply the styles from the email editor to that captured content.

Context is in the [discussion here](https://github.com/woocommerce/woocommerce/discussions/55965). The email can contain some styles like text colours which we want to replace. When applying styles in the email editor we want them to be reflected in the whole email. However, we need to ensure that the layout in the auto generated woo contents stays intact.

- Use CSS inliner
- Grab the CSS from Theme_Controller
- We may need to remove some styles from the content before inlining.
- We will need to do it also for the Woo content displayed in the preview and editor.

costasovo:

Closes #56020

Note: I used the color to highlight applied email editor styles.

| Before | After |
| ------ | ----- |
|     Screenshot 2025-03-07 at 11 16 06 |    Screenshot 2025-03-07 at 11 08 13   |

## Changes
1. **Woo Commerce Captured Content Styling Enhancement**
   - Added inlining of styles configured in the block editor
   - Removed a couple of styles (color, font-family) inlined from Woo. As for now, this looks sufficient. 

2. **Woo Content Capture Refactoring**
   - Moved mail content capture logic to `WooContentProcessor`
   - Improves separation of concerns and code organization and we get reusable serivice for capturing the content

1. Enable the block_email_editor experimental flag (in DB, set woocommerce_feature_block_email_editor_enabled option to yes).
2. Navigate to WP Admin > Woo Emails
3. Create or edit an email template
4. Configure the email to be associated with a WC_Email using the select for email type in the right sidebar. This is a temporary way of setting the association.
Screenshot 2025-03-07 at 12 45 08

5. Do the action that triggers the email. For example, for `customer_processing_order` you need to create a new order.
6. You should receive the email created in the block editor and it should contain the main content from the WC_Email. Note: you can use [WP Mail Logging plugin](https://wordpress.org/plugins/wp-mail-logging/) to capture the emails

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Internal/EmailEditor/BlockEmailRenderer.php
```diff
@@ -32,21 +32,31 @@ class BlockEmailRenderer {
...
 	/**
 	 * Initialize the renderer.
 	 *
+	 * @param WooContentProcessor $woo_content_processor Service for extracting WooCommerce content from WC_Email object.
 	 * @internal
 	 */
-	final public function init(): void {
+	final public function init( WooContentProcessor $woo_content_processor ): void {
+		$this->woo_content_processor = $woo_content_processor;
 		add_action( 'woocommerce_blocks_renderer_initialized', array( $this, 'register_block_renderers' ) );
```
triple0t: Missed this previously.
We need to rename this from `woocommerce_blocks_renderer_initialized` to `woocommerce_email_editor_blocks_renderer_initialized` to indicate it is an email editor core filter.
As this can be easily mistaken for a Woo core blocks filter

plugins/woocommerce/src/Internal/EmailEditor/BlockEmailRenderer.php
```diff
@@ -32,21 +32,31 @@ class BlockEmailRenderer {
...
 	 * Constructor.
 	 */
 	public function __construct() {
-		$editor_container   = Email_Editor_Container::container();
-		$this->renderer     = $editor_container->get( EmailRenderer::class );
-		$this->personalizer = $editor_container->get( Personalizer::class );
+		$editor_container       = Email_Editor_Container::container();
+		$this->renderer         = $editor_container->get( EmailRenderer::class );
+		$this->personalizer     = $editor_container->get( Personalizer::class );
+		$this->theme_controller = $editor_container->get( Theme_Controller::class );
```
triple0t: Left over?
Screenshot 2025-03-10 at 12 47 58 PM

Let's remove it if it is unused.

triple0t (commented):
Well done. The code looks good to me, but I have some feedback.

The capture email does not have Text color applied when set with Global styles.

| Editor | Render |
| ------ | ----- |
|  Global-styles-editor  |     Global-styles-render  |

What worked

Screenshot 2025-03-10 at 12 18 16 PM

* Font family (Block and Global Style) 
* Text colors (Block)
* Background color (Block and Global Style)

plugins/woocommerce/src/Internal/EmailEditor/BlockEmailRenderer.php
costasovo: > Missed this previously. We need to rename this from `woocommerce_blocks_renderer_initialized` to `woocommerce_email_editor_blocks_renderer_initialized` to indicate it is an email editor core filter. As this can be easily mistaken for a Woo core blocks filter

@triple0t Good catch! As this is part of a renderer that theoretically can be used without the editor. What about renaming it to `woocommerce_email_blocks_renderer_initialized`? I would skip the `editor` part. Would you be ok with such a name?
I also see `woocommerce_blocks_renderer_parsed_blocks` so I would rename this one to `woocommerce_email_blocks_renderer_parsed_blocks`. Ok?

triple0t: @costasovo that's okay with me.

plugins/woocommerce/src/Internal/EmailEditor/BlockEmailRenderer.php
costasovo: Addressed in [8eaa56d](https://github.com/woocommerce/woocommerce/pull/56199/commits/8eaa56d2f6869bab6727d3b76cce1ffc92283307).

plugins/woocommerce/src/Internal/EmailEditor/BlockEmailRenderer.php
costasovo: Addressed in [c1482c2](https://github.com/woocommerce/woocommerce/pull/56199/commits/c1482c2d7dc04e7099278525a9c69815af9c90ad)

costasovo:
@triple0t Thank you for the thorough review! I have addressed the remarks about an unused service and confusing filter prefixes.
I was not able to replicate the problem with the font color not being applied. I noticed on the screenshot that the color was not applied even on the paragraph block, which was not part of the change in this PR. So, I think it is possible that styles were not saved during the testing. Could you please try to replicate the issue and double-check that styles are saved after you change the text color?

Screenshot 2025-03-17 at 14 22 13

triple0t (approved):
LGTM. Well done.

Email Editor
Screenshot 2025-03-17 at 5 22 30 PM

Preview
Screenshot 2025-03-17 at 5 22 37 PM

Email Client
Screenshot 2025-03-17 at 5 40 23 PM

I tested 
* Font family (Block and Global Style)
* Text colors (Block and Global Style)
* Background color (Block and Global Style)


Noting here: 

> > The capture email does not have Text color applied when set with Global styles.

> I was not able to replicate the problem with the font color not being applied. I noticed on the screenshot that the color was not applied even on the paragraph block, which was not part of the change in this PR. So, I think it is possible that styles were not saved during the testing. Could you please try to replicate the issue and double-check that styles are saved after you change the text color?

When I tested with the email logging plugin, the Global style text color settings were not applied, but on testing with an email client, the styles were applied correctly.

PR #56202, opened 2025-03-07
Add Product Description Block: A New Block for Product Content

### LINKED ISSUE #55723: [Product Details Block] Create product description block
**Description**

As part of the blockifying process of product details we want need to create a block to display the description of a product. As one of the tabs ( now accordions ) will be a description accordion.

We should be able to use a variation of the existing post content block in core: https://wordpress.org/documentation/article/post-content-block/
As this already displays the correct Description when using it within a single product template.

**Note:**
Product description tab does expose the `woocommerce_product_description_heading` filter at the moment.
Given users can change the heading within the site editor now, our plan is to not support that, but as @gigitux suggested we may want to create documentation for developers to use "render_block" filter to update this.

**Acceptance criteria**

- [ ] Create a "Product Description" block that is a variation of the[ post content block](https://github.com/WordPress/gutenberg/tree/trunk/packages/block-library/src/post-content)
- [ ] Update the product details block default template to use the new "Product Description" block

imanish003:

Closes #55723

This PR introduces a new Product Description block that can be used in various contexts within WooCommerce:

- Single Product Template
- Single Product Block
- Product Collection Block

The block displays the product description content and includes the following key features:

1. **Context-Aware Display**: Automatically shows the appropriate product description based on context
2. **Editable Content**: Content is editable when used in Single Product block and user has edit permissions
3. **Read-only in Product Collection**: Description content is read-only when used within Product Collection block
4. **Placeholder Text**: Shows helpful placeholder text in the editor when specific product context is not available
5. **Recursion Prevention**: Built-in protection against recursive rendering of the block within itself

Here is a quick demo of Product Description block:

https://github.com/user-attachments/assets/f6e81668-e679-419b-a19a-1e4e3cd4106c

#### Scenario 1: Product Description block is used in Single Product template

1. Go to "Single Product" template
2. Add "Product Description" block
3. Verify that it's showing placeholder text as shown in screenshot below: "This block displays the product description. When viewing a product page, the description content will automatically appear here."
   image
4. Save and go to frontend of any product
5. Verify that it's showing product description as shown in screenshot below:
   image

#### Scenario 2: Product Description block is used in Product Collection block

1. Create a new post
2. Add "Product Collection" block and click on "create your own" button.
3. Inside "Product Template" inner block, add "Product Description" block.
4. Verify that you are able to see product description of each product on Editor side. Also, verify that description is read-only and cannot be edited.
5. Save and go to frontend
6. Verify that for each product, it's showing product description as shown in screenshot below:
   image

#### Scenario 3: Product Description block inside Single Product block

1. Create a new post or edit existing post
2. Add "Single Product" block and select a product
3. Inside "Single Product" block, add "Product Description" block
4. Verify that you are able to see _correct_ product description of selected product on Editor side. Also, verify that description is read-only and cannot be edited.
5. Save and go to frontend
6. Verify that it's showing product description.
Here is screenshot of Editor side and Frontend side:
image

7. Return to the Editor and select the Product Description block. Click the Edit button in the Toolbar (shown in the screenshot below).
   1. Verify that you can now edit the Product Description. After making changes, save the post and confirm that the updated description appears correctly in both the Editor and on the Frontend.
image

#### Scenario 4: Product Description block inside Product Details block

No testing is required for this scenario as it behaves identically to the "Single Product template". When a Product Description block is added inside a Product Details block within a Single Product Template, it automatically inherits the product context from the template and displays the correct product description. 

#### Scenario 5: Invalid context - Adding Product Description block to a new post

1. Create a new post
2. Add "Product Description" block
3. On Editor side, it should show following warning: "This block requires a product context to function. You can use it inside a Single Product template, Single Product block, Product Collection block, or within other product-related blocks."
4. Save and go to frontend
5. Verify that Frontend doesn't render anything i.e. it should be empty

#### Scenario 6: Product Description block is available in Product Details block

1. Go to Single Product template
2. Add "Blockified Product Details" block
3. Verify that Product Description block is available in Product Details block by default as shown in screenshot below:

image

#### Scenario 7: Inside "Query Loop" block

> This scenario is similar to Product Collection block.

1. Create a new post
2. Add "Query Loop" block
3. Set "Post type" to "Product" using Inspector control
4. Add "Product Description" block inside "Post Template" block
image

5. Verify that it's showing product description of each product.
6. Save and go to frontend
7. Verify that it's showing product description of each product.

### Changelog entry

<details>

#### Significance

- [ ] Patch
- [ ] Minor
- [x] Major

#### Type

- [x] Add - Adds functionality

#### Message 
Add Product Description Block: A New Block for Product Content

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/blocks/product-description/block.json
```diff
@@ -0,0 +1,69 @@
...
+	"$schema": "https://schemas.wp.org/trunk/block.json",
+	"apiVersion": 3,
+	"name": "woocommerce/product-description",
+	"version": "1.0.0",
+	"title": "Product Description",
+	"description": "Displays the description of the product.",
+	"category": "woocommerce",
+	"textdomain": "woocommerce",
+	"usesContext": [ "postId", "postType", "queryId" ],
+	"supports": {
```
louwie17, line 16: Did you copy these `supports` from the `Content` block?

plugins/woocommerce/client/blocks/assets/js/blocks/product-description/edit.tsx
```diff
@@ -0,0 +1,219 @@
...
+						'This block requires a product context to function. You can use it inside a Single Product template, Single Product block, Product Collection block, or within other product-related blocks.',
+						'woocommerce'
+					) }
+				</Warning>
+			</div>
+		);
+	}
+
+	const isDescendentOfProductCollection = Number.isFinite( queryId );
+	const isEditable = userCanEdit && ! isDescendentOfProductCollection;
```
louwie17: Any specific reason you decided to disable editing for its use within the product collection block, but keep it within the single product block?

plugins/woocommerce/src/Blocks/BlockTypesController.php
```diff
@@ -474,6 +474,7 @@ protected function get_block_types() {
 			'ReviewsByProduct',
 			'RelatedProducts',
 			'ProductDetails',
+			'ProductDescription',
```
louwie17: I see that we are conditionally registering it in the `index.tsx` file already, but the file would still get enqueued.
Would it make sense to not instantiate the class either, as we are doing for the product details block, see items under this condition: `if ( Features::is_enabled( 'experimental-blocks' ) ) {`

plugins/woocommerce/client/blocks/assets/js/blocks/product-description/edit.tsx
```diff
@@ -0,0 +1,219 @@
...
+		props;
+	const blockProps = useBlockProps( { className: layoutClassNames } );
+
+	const userCanEdit = useCanEditEntity( 'postType', postType, postId );
+	if ( userCanEdit === undefined ) {
+		return null;
+	}
+
+	// Show warning if the block is not used in a product context.
+	if ( postType && postType !== 'product' ) {
```
louwie17: I don't keeping this, but I think we should do the WP approach by not even allowing the user to add the block and showing a notice instead, or is this something we cannot configure on our end?
Screenshot 2025-03-10 at 3 55 03 PM

louwie17 (commented):
Nice work @imanish003, this tested well! I did leave a couple suggestions within the code and a couple questions.
Also are you able to add the description block to the blockified product details block right away, so replacing the `core/paragraph` usage [here](https://github.com/woocommerce/woocommerce/blob/99f04562c084808e02515a526664fdc0c77fa688/plugins/woocommerce/client/blocks/assets/js/blocks/product-details/edit.tsx#L23) and using `woocommerce/product-description`

> Editable Content: Content is editable when used in Single Product block and user has edit permissions

Nice, this was pretty cool! Also because our new editor supports blocks for the description so this automatically worked within the single product block and changes where reflected correctly.

imanish003:
> Also are you able to add the description block to the blockified product details block right away, so replacing the core/paragraph usage [here](https://github.com/woocommerce/woocommerce/blob/99f04562c084808e02515a526664fdc0c77fa688/plugins/woocommerce/client/blocks/assets/js/blocks/product-details/edit.tsx#L23) and using woocommerce/product-description

Good point, I will make the changes.

imanish003:
> Nice, this was pretty cool! Also because our new editor supports blocks for the description so this automatically worked within the single product block and changes where reflected correctly.

Ah nice, I didn't know that our new editor supports adding blocks to the description. So this is great. 🚀

plugins/woocommerce/client/blocks/assets/js/blocks/product-description/block.json
imanish003: Yes, I copied those from the content block. I don't see any good reason to remove them because our block is similar to the content block. However, let me know if you think any of these should be removed.

plugins/woocommerce/client/blocks/assets/js/blocks/product-description/edit.tsx
imanish003: Okay, good question. TBH, I don't have any specific reason. It was there for Query Loop for Post Content block. So, I assumed that there was a reason behind not making it editable for Query Loop block. So, I thought I should not make it editable for product collection block as well. But, I don't see any good reason for this, TBH. So, I think we should remove this condition. So, we can allow users to edit description in the product collection block. 

Out of curiosity, I looked at the Gutenberg code and looks like Bart made this change in https://github.com/WordPress/gutenberg/pull/48642. CC: @WunderBart was there any specific reason for adding this condition? i.e.
```tsx
const isDescendentOfQueryLoop = Number.isFinite( queryId );
```

imanish003: I removed the condition in [d3dc6e9](https://github.com/woocommerce/woocommerce/pull/56202/commits/d3dc6e93df1bbf39086b7aab3c45093db3af5eb2)

plugins/woocommerce/src/Blocks/BlockTypesController.php
imanish003: Ah, good catch. Sometimes I get confused between `ProductDetails` and `BlockifiedProductDetails`. My bad 😅 I will make the changes.

plugins/woocommerce/client/blocks/assets/js/blocks/product-description/block.json
louwie17: No, in that case we can leave them. I was just curious.

plugins/woocommerce/client/blocks/assets/js/blocks/product-description/edit.tsx
louwie17: Makes me wonder if there is a performance issue in something like a query loop as it would have to pull in quite a few descriptions, and descriptions may not be overly small 🤷‍♂️ 
Having said that, a user would have to explicitly add it within the query loop, so I did be ok with removing the condition as you did.

plugins/woocommerce/client/blocks/assets/js/blocks/product-description/edit.tsx
imanish003: > I don't keeping this, but I think we should do the WP approach by not even allowing the user to add the block and showing a notice instead, or is this something we cannot configure on our end?

AFAIK I don't believe we can configure this. [This code](https://github.com/WordPress/gutenberg/blob/e093fefd041eb6cc4a4e7f67b92ab54fd75c8858/packages/editor/src/components/provider/use-hide-blocks-from-inserter.js#L45-L70) is responsible for displaying this snake bar and is prefixed with `__unstable`, so I don't think it would be wise to use this experimental filter `blockEditor.__unstableCanInsertBlockType`. What do you think?

plugins/woocommerce/client/blocks/assets/js/blocks/product-description/edit.tsx
imanish003: Even if we don't make it editable, the product collection will still fetch the product descriptions if they are included within the loop. So, there maybe a performance issue if it is used in the product collection. 

As you mentioned, users must explicitly include it in the Product collection block. For now, we can remove this condition, and should any performance issues occur later, we can reassess it. Does this sound good to you?

plugins/woocommerce/client/blocks/assets/js/blocks/product-description/edit.tsx
louwie17: Gotcha, that makes sense, not using that is correct. It also sounds like they want to remove it in the future for a more declarative approach.
Speaking of that, would the use of `[ancestor` work](https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/nested-blocks-inner-blocks/#defining-an-ancestor-block-relationship)? or does that not work with a template?
Although does look like there is an open issue for this https://github.com/WordPress/gutenberg/issues/41062 which a lot of us Woo FSE folks got involved in :)

imanish003: @louwie17 

> Speaking of that, would the use of [ancestor work](https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/nested-blocks-inner-blocks/#defining-an-ancestor-block-relationship)? or does that not work with a template?

Yeah, you are right. We can't use ancestor property for templates. We can just provide the block names for this property.

After reading this, I recalled that we have a utility function [registerProductBlockType](https://github.com/woocommerce/woocommerce/blob/10992f3f02a28b8820e04295315e3251e94fa012/plugins/woocommerce/client/blocks/assets/js/atomic/utils/register-product-block-type.ts#L325-L413) that we can use. I have made few commits and used this function to register Product Description block. In short this function:
- Removes `ancestor` property on Single Product template and allow inserting blocks on it
- On other pages, it keeps `ancestor` property. This way, we will be able to use Product Description block on only inside allowed blocks i.e. Single Product block and Product Collection block.

Could you please test this and check if it works as expected? 🙂

plugins/woocommerce/client/blocks/assets/js/blocks/product-description/block.json
```diff
@@ -8,6 +8,10 @@
 	"category": "woocommerce",
 	"textdomain": "woocommerce",
 	"usesContext": [ "postId", "postType", "queryId" ],
+	"ancestor": [
+		"woocommerce/single-product",
+		"woocommerce/product-template"
```
louwie17: We should also add `core/post-template` to this list, as otherwise it prevents users from using this in a query loop based on products:
Screenshot 2025-03-12 at 4 14 39 PM

plugins/woocommerce/client/blocks/assets/js/blocks/product-description/edit.tsx
```diff
@@ -124,33 +124,14 @@ function EditableContent( { context = {} } ) {
...
-				</Warning>
-			</div>
-		);
-	}
-
-	const isDescendentOfProductCollection = Number.isFinite( queryId );
-	const isEditable = userCanEdit && ! isDescendentOfProductCollection;
-
-	return isEditable ? (
+	return userCanEdit ? (
```
louwie17, line 135: I think we should add the product collection limitation back in 🙈 sorry!
In testing it, I noticed that you can click on each block within a query loop for each individual product/post that is rendered, but it will select the same block type in the block list view.
This causes some interesting behaviour around blocks that have a block based descriptions and one's that don't, where the context keeps switching ( see video below ). 
This gives me the sense that the blocks shown within a query loop should ideally not be content aware for handling any settings. If we add this edit support, it would mean they are content aware and users see different things depending on which product they are looking at:

https://github.com/user-attachments/assets/0ee3322c-3f53-4814-97d8-332e2d8888da

plugins/woocommerce/client/blocks/assets/js/blocks/product-description/edit.tsx
```diff
@@ -0,0 +1,200 @@
...
+	const { postId: contextPostId, postType: contextPostType } = context;
+	const hasAlreadyRendered = useHasRecursion( contextPostId );
+
+	if ( contextPostId && contextPostType && hasAlreadyRendered ) {
+		return <RecursionError />;
+	}
+
+	return (
+		<RecursionProvider uniqueId={ contextPostId }>
+			{ contextPostId && contextPostType ? (
```
louwie17, line 225: I think we should update this condition to be product specific? and/or update the product details block to be limited by the single-product, product-template, and post-template parent ancestors.
Mostly because I can cause this funny bug where if I add the product details ( with the product description auto added ) to the top level of a page or post it will render the content of the page and post, causing this weird recursion:
Screenshot 2025-03-12 at 4 27 11 PM

plugins/woocommerce/client/blocks/assets/js/blocks/product-description/index.tsx
```diff
@@ -0,0 +1,24 @@
+/**
+ * External dependencies
+ */
+import { isExperimentalBlocksEnabled } from '@woocommerce/block-settings';
+import { registerProductBlockType } from '@woocommerce/atomic-utils';
```
louwie17, line 5: Nice find!

plugins/woocommerce/client/blocks/assets/js/blocks/product-details/edit.tsx
```diff
@@ -19,13 +19,7 @@ const createAccordionItem = (
...
 
 const descriptionAccordion = createAccordionItem( 'Description', [
-	[
-		'core/paragraph',
-		{
-			content:
-				'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed eget turpis eget nunc fermentum ultricies. Nullam nec sapien nec0',
-		},
-	],
+	[ 'woocommerce/product-description', {}, [] ],
```
louwie17, line 22: Thanks 🙇‍♂️

plugins/woocommerce/client/blocks/assets/js/blocks/product-description/edit.tsx
louwie17: Maybe the most reliable way is to make use of the product data store and use `select( productsStore ).getProduct( contextPostId )` and see if it returns a product? instead of checking the post type specifically?

plugins/woocommerce/client/blocks/assets/js/blocks/product-description/edit.tsx
louwie17: Hey nice, that is pretty cool!
It seemed to test well 👍 💯

louwie17 (changes requested):
Thanks for the updates @imanish003, and for finding that custom function to limit the addition of the block in certain scenarios that tested well.

I did noticed a couple more edge cases and did have a change of mind on the edit condition within the query loop, I think we should add it back in. I left a detailed comment with an example and why I think we should.

plugins/woocommerce/client/blocks/assets/js/blocks/product-description/block.json
imanish003: Sure, I can make this change. But the catch is that after adding this new property we will need some additional logic. So here is a new commit 380bd8e with all the changes. Let me know if that looks good to you.

plugins/woocommerce/client/blocks/assets/js/blocks/product-description/edit.tsx
imanish003: I am not sure if I understand you correctly because I still think the behavior is as expected. Basically, when there are no inner blocks, we have this classic inner block that allows us to edit the product description. However, if there are inner blocks like paragraphs, users can make changes to those paragraphs directly, and after the user saves the post, the product description will also be updated. So, the editing of the product description still works. I am uncertain if I am understanding you correctly. Isn't that what we want? Here is a quick demo video:

https://github.com/user-attachments/assets/4e45e379-671b-417e-bc5e-4e2997a480af

plugins/woocommerce/client/blocks/assets/js/blocks/product-description/edit.tsx
imanish003: > Mostly because I can cause this funny bug where if I add the product details ( with the product description auto added ) to the top level of a page or post it will render the content of the page and post, causing this weird recursion:

IMO Similar to Product Description block, we should restrict Product Details block to be available only in Product Collection block, Single Product block, or other product-aware blocks. In my testing, If I add Product Details block then Product Description block doesn't work at all, because it's in invalid context. Isn't that happening for you as well?

IMO It should be safe to assume that if Product Description block is inside Product Details block then it will have the product context. What do you think?

louwie17: > Isn't that happening for you as well?

I hadn't tested that yet, although likely similar, but would show the post comments instead of product comments on the front end, things in the site editor should work as I am specifically hitting the Woo reviews API for the reviews.

> IMO It should be safe to assume that if Product Description block is inside Product Details block then it will have the product context. What do you think?

Yes, I suppose so! We can stick with that, as I am not sure what other edge cases there could be aside from adding the HTML manually, but a user would have to explicitly do this.

plugins/woocommerce/client/blocks/assets/js/blocks/product-description/edit.tsx
louwie17: Aaah I didn't notice that it does explicitly list the changed products on save, that is pretty cool!

Having said that, I still wonder if there is a larger reason for having this condition ( I asked in Slack to see ), because the post content block is not the only one the condition is also in the:
- Post author
- Post date
- Post title
- Post excerpt
- Post Image

louwie17 (approved):
Thanks for this @imanish003, it tested well and thanks for the follow ups. I think this is good to merge.
I do foresee a couple follow ups:
- Update product details block to only be added within a product context
- Update the **Singe Product** block to allow product description as a direct child ( change [this array](https://github.com/woocommerce/woocommerce/blob/99f04562c084808e02515a526664fdc0c77fa688/plugins/woocommerce/client/blocks/assets/js/blocks/single-product/constants.tsx#L82) )
- Follow up on the conversation around disabling the edit mode within a query loop ( IMO I think we are better off disabling this for now and enabling it in the future, than potentially the other way around ).

imanish003:
> Update product details block to only be added within a product context

Okay, I will create a follow up PR for this. Here is the [draft PR](https://github.com/woocommerce/woocommerce/pull/56431) and I will mark it ready for review.

> Update the Singe Product block to allow product description as a direct child ( change [this array](https://github.com/woocommerce/woocommerce/blob/99f04562c084808e02515a526664fdc0c77fa688/plugins/woocommerce/client/blocks/assets/js/blocks/single-product/constants.tsx#L82) )

Interesting, do you know the specific use of this array? Because I am able to manually add product description block as inner block of a single product block. So how this array impact the product description block? 

> Follow up on the conversation around disabling the edit mode within a query loop ( IMO I think we are better off disabling this for now and enabling it in the future, than potentially the other way around ).

TBH, I don't have any strong opinion, so I am happy to make this change if you are in favor of it.

PR #56208, opened 2025-03-07
[e2e tests] Remove cart and checkout page conversion during global setup

### LINKED ISSUE #50758: [e2e tests] Remove the conversion of cart and checkout blocks to classic in global setup

adimoldovan:

Closes #50758

Removes the global setup step that converts the default cart and checkout pages to their shortcode versions.
Updated all affected tests that were expecting the classic frontend by default to work with the blocks version.

e2e tests pass.
CYS tests failing with WP6.8-beta2 is not related to this PR.


plugins/woocommerce/tests/e2e-pw/playwright.config.js
```diff
@@ -80,7 +80,7 @@ export const setupProjects = [
 
 export default defineConfig( {
 	timeout: 120 * 1000,
-	expect: { timeout: 20 * 1000 },
+	expect: { timeout: CI ? 20 * 1000 : 10 * 1000 },
```
Stojdza, line 83: Is this change intentional?

Stojdza (approved):
Ran several specs locally. LGTM 👍

plugins/woocommerce/tests/e2e-pw/playwright.config.js
adimoldovan: Yes, the env is much faster locally and should not need more than 10s. Even in CI 20s is too much I think, and I plan on tweaking that in the future. But that's for another time.
We'll need to override these for the external environments though.

adimoldovan:
Bypassing the CYS test failing with WP6.8 beta2. That is an issue not related to this PR and being addressed separately.

PR #56215, opened 2025-03-07
[Performance] Initial implementation of product reviews pending review transient.

### LINKED ISSUE #55015: [Performance]: Unapproved comments bubble for product reviews
### Prerequisites

### Performance Details

In the WP Admin menu under Products > Reviews a count bubble is included for the amount of unapproved reviews. This value is cached, but since the menu is included on every page the additional load can appear on any random page.

Image

Considering that comments are used for a lot of things, like product reviews, order notes (even with HPOS enabled), etc... the cache can also be cleared quite often. For example every time a new order comes in the comment count cache can be cleared even if it's not a new product review that comes in.

This is the stack trace from query monitor:

Image

The query to populate this count is the following:

```
SELECT COUNT(*) FROM wp_comments
JOIN wp_posts ON wp_posts.ID = wp_comments.comment_post_ID
WHERE ( comment_approved = '0' )
AND comment_type IN ('review', '', 'comment')
AND wp_posts.post_type IN ('product')
AND comment_type != 'order_note'
AND comment_type != 'webhook_delivery'
ORDER BY wp_comments.comment_date_gmt DESC
```

In the past we had to account for product reviews coming in without a `comment_type` set, see: https://github.com/woocommerce/woocommerce/blob/9.6.0/plugins/woocommerce/includes/wc-update-functions.php#L1899
However I don't believe this is necessary any longer. If we can assume that every new comment that comes in has the `comment_type`, set to `review` then the post table join doesn't seem necessary in this case. At most we can include a filter for all "review" comment types so other plugins can extend this query if they use a custom comment type. See also [suggestion here](https://github.com/woocommerce/woocommerce/issues/54286#issuecomment-2624468091).

The filter for comment types also seems a little conflicting, a comment_type can't be both IN a specific set and != to specific values, just the IN looks like it will be enough. Maybe it's different code snippets that are all adding to this query.

Also if we are just interested in the total count, we can drop the order by.

In addition to the suggestions to improve the query, I think it should also be possible to be more selective when clearing the cache. From my testing this seems to happen even when an order note comes in. I'd assume we can limit to clearing only when a product review comes in or is changed / deleted.

### Affected Sites

Sites with larger comment / posts tables. Since the comments table is shared it could be any site with a large amount of any of these:

- Posts with comments
- Products with reviews
- Orders with order notes (status changes, payment, shipping handling etc...)
- Frequent webhook delivery

### Suspected causes

As outlined above the suspected causes could be:
- too frequent clearing of the cached count
- join with posts table
- unnecessary order by for a count query

### Other context

Discovered when working through spike: https://github.com/woocommerce/automatewoo/issues/1890#issuecomment-2597884738

kalessil:

Closes #55015.

Introduce a caching solution to track the pending reviews counter.

Screenshot from 2025-03-12 15-12-09

- On a test site, create several product reviews
- Navigate to Products -> Reviews (expected no counter bubble)
- Unapprove/upprove some of them and **refresh the page** - counter should appear, and the query monitor should show no DB query
- Try additional review actions (exploratory) - the counter bubble should reflect only unapproved comments.

Note: FE bugs require a page refresh to verify the bubble counter is displayed correctly


prettyboymp (commented):
The changes look good.  I haven't tested, but it looks like it should cover the changes needed.

What do you think about switching from transients to wp_cache?  

It would allow the usage of `wp_cache_incr()`/`wp_cache_decr()`, which: 
* won't bump the expiry (for most object cache implementations) allowing any drift in accuracy to get reset.  The current implementation could allow this reset to get starved for any sites that receive at least one product review a day.
* wp_cache_incr/decr is less vulnerable to race conditions that would potentially cause any count drift to begin with.
* It would simplify the logic a little in `maybe_bump_products_reviews_pending_moderation_counter ` around incrementing when the transient has expired since, when the cache is expired, the increment just fails - so the cache just remains empty until the next time the admin is loaded.

plugins/woocommerce/src/Caches/CommentsCountCache.php
```diff
@@ -0,0 +1,103 @@
...
+	 * @param int    $increment The value increment to be applied.
+	 * @return false|int
+	 */
+	public function increment( string $name, int $increment = 1 ) {
+		if ( $this->use_cache_api ) {
+			return wp_cache_incr( $name, $increment, self::COMMENT_COUNT_CACHE_GROUP );
+		}
+
+		$transient_value = get_transient( $name );
+		if ( false !== $transient_value && false !== set_transient( $name, ++$transient_value, DAY_IN_SECONDS ) ) {
```
prettyboymp: There is a potential to starve the expiration since calling `set_transient()` here will reset the expiration for another day where the incr/decr functionality for all object cache services I'm aware of will keep the original expiry.

I don't know how likely or how big of a deal that is though.  A site getting enough comments to cause counting collisions and starvation likely needs to implement object caching pretty soon.

kalessil: Here we go: f50e9176b84e869b9829b62139461fff671d0f07 (tested manually, and the expiration timestamp is not getting changed when we update the values).

plugins/woocommerce/includes/class-wc-comments.php
```diff
@@ -225,6 +237,68 @@ public static function delete_comments_count_cache() {
 		delete_transient( 'wc_count_comments' );
 	}
 
+	/**
+	 * Fetches (and populates if needed) the counter.
+	 *
+	 * @return int
+	 */
+	public static function get_products_reviews_pending_moderation_counter() {
```
kalessil: Should we move those new methods to a new service class?

plugins/woocommerce/src/Container.php
```diff
@@ -106,9 +106,10 @@ public function __construct() {
 	 * Finds an entry of the container by its identifier and returns it.
 	 * See the comment about ContainerException in RuntimeContainer::get.
 	 *
-	 * @param string $id Identifier of the entry to look for.
+	 * @template T
+	 * @param class-string<T> $id Identifier of the entry to look for.
 	 *
-	 * @return mixed Resolved entry.
+	 * @return T Resolved entry.
```
kalessil, line 112: It fixes the broken return type resolution for this method; now, code navigation (and SCA) around its returned objects works.

plugins/woocommerce/includes/class-wc-comments.php
```diff
@@ -306,6 +317,67 @@ public static function delete_comments_count_cache() {
...
+	 *
+	 * @param int         $comment_id Comment ID.
+	 * @param \WP_Comment $comment    Comment object.
+	 * @return void
+	 */
+	public static function maybe_bump_products_reviews_pending_moderation_counter( $comment_id, $comment ): void {
+		$needs_bump = '0' === $comment->comment_approved;
+		if ( $needs_bump && in_array( $comment->comment_type, array( 'review', 'comment', '' ), true ) ) {
+			$is_product = 'product' === get_post_type( $comment->comment_post_ID );
+			if ( $is_product && false !== wp_cache_get( self::PRODUCT_REVIEWS_PENDING_COUNT_CACHE_KEY, self::COMMENT_COUNT_CACHE_GROUP ) ) {
```
prettyboymp: I think you can skip the `false !== wp_cache_get()` check here.  The incr/decr functions should just silently fail if the values don't exist.

prettyboymp (approved):
Everything looks good.  Just left one minor comment.

plugins/woocommerce/includes/class-wc-comments.php
kalessil: Sounds good: simplified in c3f882c3ad09a038849fbbd233fbcbd25654ab79!

PR #56233, opened 2025-03-07
Support `overrides` property for the free extensions feed

moon0326:

This PR introduces `overrides` property that allows overriding existing values in the spec.
Additionally, it adds `context_plugins` rule.

Let's say we have the following spec definition.

```
{
  "name": "PayPal",
  "description": "Test",
  "is_visible": true,
  "order" : 1,
  "key": "woocommerce-paypal-payments",
  "label": "Get paid with PayPal",
  "install_priority": 5,
  "requires_jpc": true
}
```

If we need to set a different value for order when the feed includes WooCommerce Payments, we can use the following overrides.

```
{
  "name": "PayPal",
  "description": "Test",
  "is_visible": true,
  "order" : 1,
  "overrides": [
    {
      "field": "order",
      "value": 3,
      "rules": [
        {
          "type": "context_plugins",
          "name": "name",
          "value": "WooPayments",
          "operation": "="
        }
      ]
    }
  ],
  "key": "woocommerce-paypal-payments",
  "label": "Get paid with PayPal",
  "install_priority": 5,
  "requires_jpc": true
}
```

#### context_plugins

This rule type receives a list of plugins (see https://github.com/woocommerce/woocommerce/pull/56233/files#diff-2f3016b95795e4ddd5bdda3de76f3b71604600347811b8e8c47bba6999ff9ba4R85) and runs a comparison against a property.

Definition:

```
 {
  "type": "context_plugins",
  "name": "name",
  "value": "WooPayments",
  "operation": "="
}
```

**Desktop**
1. Create a new JN site with this branch. When redirected to the core profiler, leave it. We'll come back to it.
2. Install and activate 
[free-extensions-test.zip](https://github.com/user-attachments/files/19195376/free-extensions-test.zip). This will override the free extensions feed with a test file (extensions.json).
3. Navigate back to the core profiler and proceed.
4. Choose a WooPayments supported country. 
5. You should see "Get paid with PayPal" in the 3rd slot.
6. Navigate back to the country selection and choose a country where WooPayments is not supported.
7. You should see "Get paid with PayPal" in the 1st slot.

If possible, test other rules by changing `extensions.json` in [free-extensions-test.zip](https://github.com/user-attachments/files/19195376/free-extensions-test.zip)

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

Support `order` property from the free extensions feed.

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/admin/client/core-profiler/index.tsx
```diff
@@ -535,9 +535,11 @@ const getPlugins = fromPromise( async () => {
 		onboardingStore
 	).getFreeExtensions();
 	return (
-		extensionsBundles.find(
-			( bundle ) => bundle.key === 'obw/core-profiler'
-		)?.plugins || []
+		extensionsBundles
+			.find( ( bundle ) => bundle.key === 'obw/core-profiler' )
+			?.plugins.sort(
+				( a, b ) => ( a.order ?? Infinity ) - ( b.order ?? Infinity )
```
ilyasfoo: Nit: Currently, we can't order `[0, 1, , , , 5]` since all unordered extensions will be placed after and it pushes all with values to the front and becomes `[0, 1, 5, , , ,]`.

I think having a fixed index mapping is better in terms of stability. For example, we can have ordering set for plugins with contractual obligation and don't have to explicitly set everything else.

I've this ChatGPT proposed solution for fixed index ordering if you're interested: https://chatgpt.com/share/67ce582f-e188-800a-bcc7-43a4141dccec. Let me know what you think!

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/EvaluateExtension.php
```diff
@@ -59,6 +59,12 @@ private static function evaluate( $extension ) {
 	 */
 	public static function evaluate_bundles( $specs, $allowed_bundles = array() ) {
 		$bundles = array();
+		$evaluate_order = new EvaluateOrder();
+		$context = array(
+			'vars' => array(
+				'platform' => wp_is_mobile() ? 'mobile' : 'desktop',
```
ilyasfoo: It seems [`wp_is_mobile`](https://developer.wordpress.org/reference/functions/wp_is_mobile/) isn't a reliable way detect screen sizes. For example, 13-inch iPad Pro would still be recognized as mobile, even with a wide horizontal screen. I think the best way for determining screen sizes is still JS/CSS since its dynamic in nature, perhaps we'll need to inject another prop? 😅

ilyasfoo (commented):
Thanks @moon0326! I gave a quick look and the only issue is with detecting mobile screens via PHP.

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/EvaluateExtension.php
moon0326: @ilyasfoo Good catch!

I think the best approach would be adding another prop as you suggested. It feels a bit overkill to add another prop since we only have a single use case at the moment. However, all other approaches have a disadvantage.

1. Fix `wp_is_mobile`: It's not responsive. If you use a desktop browser and resize your window, it'll always use the desktop sort order. 
2. Hard-code in CSS: Not a long term solution 🥲

plugins/woocommerce/client/admin/client/core-profiler/index.tsx
moon0326: @ilyasfoo Good catch. I think we can also solve this by setting the array index as `order` value.

I've made the change in [Use array index for the default sort order](https://github.com/woocommerce/woocommerce/pull/56233/commits/2e25e3594467f7e925df265380bd535c3f6946b7)

We have a case where the auto-filled order value is the same as the manually entered order value. I think we can handle that on our end. We'll probably enter an order for every item anyway.

What do you think?

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/EvaluateExtension.php
moon0326: @woocommerce/ghidorah What do you think we hard-code it for now (PayPal) and add another prop when we have at least one more case 🤔 ?

moon0326: @ilyasfoo I've added `order_mobile` in the latest commits. 

Please re-download [free-extensions-test.zip](https://github.com/user-attachments/files/19169988/free-extensions-test.zip) for testing. 

https://github.com/user-attachments/assets/03438dd7-00fc-4f1e-9846-d72f69ebcd9f

adrianduffell: > I've added order_mobile in the latest commits

Nice, I think adding the prop is a good solution. I could see a slight improvement to make would be to think about the order in terms of lists and grids instead of mobile vs desktop. I find mobile/desktop a bit ambiguous because it could still be possible for the grid view to be shown on a large mobile device. So perhaps this could be better named as `order_list` and it's up to the client-side to determine whether it is showing the recommendations as a list.

Going a step further, in the case of PayPal, the list order is arguably the correct order, and the override needs to happen on the grid view to shift it to the third position. So perhaps the props could look like:

| Prop   | Description                                          |
|------------|------------------------------------------------------|
| `order`    | The regular order                                   |
| `order_list` | An override when the layout is in a list format   |
| `order_grid` | An override when the layout is in a grid format   |

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/GetRuleProcessor.php
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Automattic\WooCommerce\Internal\Admin\RemoteFreeExtensions;
+
+use Automattic\WooCommerce\Admin\RemoteSpecs\RuleProcessors\GetRuleProcessor as OriginalGetRuleProcessor;
+use Automattic\WooCommerce\Admin\RemoteSpecs\RuleProcessors\RuleProcessorInterface;
+use Automattic\WooCommerce\Internal\Admin\RemoteFreeExtensions\Rules\ContextPluginsRuleProcessor;
+use Automattic\WooCommerce\Internal\Admin\RemoteFreeExtensions\Rules\ContextVarsRuleProcessor;
+
+/**
+ * A custom GetRuleProcessor class to support context_vars and context_plugins rule types.
```
adrianduffell: Suggestion: `context_vars` could be completely removed if there is no current use-case for it (assuming there will be a separate prop for mobile ordering).

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/GetRuleProcessor.php
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * GetRuleProcessor class.
+ */
+class GetRuleProcessor {
```
adrianduffell: I feel the name of this could be more descriptive. `Perhaps GetRuleProcessorForContext` 🤔

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/Rules/ContextPluginsRuleProcessor.php
```diff
@@ -0,0 +1,71 @@
...
+	/**
+	 * The list of plugins.
+	 *
+	 * @var array a list of plugins.
+	 */
+	private array $plugins;
+
+	/**
+	 * Constructor.
+	 * @param array $plugins a list of plugins.
```
adrianduffell: Is this actually a list of objects? I first interpreted it as strings, e.g: `[  'woocommerce', 'jetpack'  ]`

moon0326: @adrianduffell It is actually a full plugin object. This rule can access any plugin properties, not just name. I'll document it with a few more examples.

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/GetRuleProcessor.php
moon0326: @adrianduffell Good suggestion! 👍

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/GetRuleProcessor.php
moon0326: @adrianduffell Yup! Sounds good 👍

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/EvaluateExtension.php
moon0326: @adrianduffell Thank you for the suggestion 👍 

So, `order` is used for the 2x3 layout (with more rows coming soon), and `order_list` is used for the 1x8 layout. But where do we use `order_grid`? 🤔 Isn't the 2x3 layout already a grid?

moon0326: If both @adrianduffell @ilyasfoo agree, I would like to focus on the backend part in this PR so we can merge soon.

In this PR:

1. Add `order` only in this PR. 
2. Remove any frontend changes.

In a separate PR: 

1. Add any additional props. Adding a new props requires just one line. Example: https://github.com/woocommerce/woocommerce/pull/56233/files#diff-2f3016b95795e4ddd5bdda3de76f3b71604600347811b8e8c47bba6999ff9ba4R88
2. Update the frontend to use the added props.

adrianduffell: > I would like to focus on the backend part in this PR

Sounds ok to me!

adrianduffell: > So, order is used for the 2x3 layout (with more rows coming soon), and order_list is used for the 1x8 layout. But where do we use order_grid? 🤔 Isn't the 2x3 layout already a grid?

@moon0326 That's close, I would see `order` being the default/regular order used by both 2x3 and 1x8 layouts. Then, each layout would process the following overrides to determine the final order:

* 2x3 layout would process `order_grid`
* 1x8 layout would process `order_list`

In this approach, it might be more meaningful to name it `order_grid_override` and `order_list_override`.

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/EvaluateOrder.php
```diff
@@ -0,0 +1,62 @@
...
+	/**
+	 * Evaluates the order property.
+	 *
+	 * @param RuleEvaluator $rule_evaluator The rule evaluator.
+	 * @param object        $order The order to evaluate.
+	 *
+	 * @return mixed The evaluated order.
+	 */
+	protected function evaluate_order( RuleEvaluator $rule_evaluator, $order, $default_sort_order ) {
+		// Return 9999 if the order is not set. This is the highest possible order.
```
ilyasfoo: Should we remove this?
```suggestion
		// Return 9999 if the order is not set. This is the highest possible order.
```

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/EvaluateExtension.php
ilyasfoo: > In this approach, it might be more meaningful to name it order_grid_override and order_list_override.

@adrianduffell  Nice, I think your name suggestion is clear!

> I would like to focus on the backend part in this PR so we can merge soon.

@moon0326 Sounds good to me!

plugins/woocommerce/client/admin/client/core-profiler/index.tsx
ilyasfoo: That's smart! I think if the index does not include non-visible plugins, it should work well

> We'll probably enter an order for every item anyway.

I'd agree 👍

ilyasfoo (commented):
Nice work, @moon0326! This is testing well so far. I'm assuming you'll be making more changes according to [this](https://github.com/woocommerce/woocommerce/pull/56233#discussion_r1988283301), I'll come back to review it when you're done

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/GetRuleProcessor.php
moon0326: @adrianduffell Updated it in [Rename rule processor](https://github.com/woocommerce/woocommerce/pull/56233/commits/76fe21ca5ad6b00c9d26ee4a26dd51293d8e8b31)

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/Rules/ContextPluginsRuleProcessor.php
moon0326: @adrianduffell Added a comment in [Add details for plugin obj](https://github.com/woocommerce/woocommerce/pull/56233/commits/bc78768c1b57ad9ccccebe5cd2754840eedf93ac)

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/EvaluateOrder.php
moon0326: @ilyasfoo Good catch! This has been removed.

moon0326:
@ilyasfoo @adrianduffell 

I've pushed `overrides` changes.

1. The `overrides` property has been moved to the root level. It can override any values, not just the order. I've updated the PR description with an example.
2. The classes related to `overrides` have been moved to `Automattic\WooCommerce\Admin\RemoteSpecs` namespace as it can be used by other specs (may be in the future?)
3. Test has been updated. 
4. I've updated the test instructions.

ilyasfoo (approved):
Nice work, @moon0326! Tests well.

> The classes related to overrides have been moved to Automattic\WooCommerce\Admin\RemoteSpecs namespace as it can be used by other specs (may be in the future?)

I agree. Optionally, it'll be good to add the new rules to [readme](https://github.com/woocommerce/woocommerce/blob/cb1217db4c9697529d020c0e957c52153b7ccdbf/plugins/woocommerce/src/Admin/RemoteSpecs/RuleProcessors/README.md#L452).

#### Review checklist

| Parameter  | Result |
| Approach  | ✅ All good |
| Automated tests  | ✅ All good |
| Naming  | ✅ All good |
| Typing  | ✅ All good |
| Changelog  | ✅ All good |
| Documentation  | Commented |

#### Scenarios tested

| Selected non-WooPayments country  | ✅ Functionality worked correctly | image |
| Selected WooPayments country  | ✅ Functionality worked correctly | image |

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/Rules/ContextPluginsRuleProcessor.php
adrianduffell: Nice, the new documentation helps a lot!

adrianduffell:
Nice! Tested well per instructions on a JN site, thanks @moon0326 🚀 

Screen Shot 2025-03-12 at 10 38 22 pm

Will the override system be able to handle placing PayPal in third position when WooCommerce Shipping is hidden for non-US stores? This could be handled as a follow-up if not. This screenshot shows PayPal in second spot for Japan with the free-extenstions-test.zip rules:

Screen Shot 2025-03-12 at 10 40 47 pm

I also have a nit that can perhaps be addressed in a follow-up:

Should this rule be an IN operation? It seems that `value` is being compared to a list of plugin names  🤔 
```
"rules": [
        {
          "type": "context_plugins",
          "name": "name",
          "value": "WooPayments",
          "operation": "="
        }
      ]
 ```

adrianduffell (approved):

moon0326:
> Will the override system be able to handle placing PayPal in third position when WooCommerce Shipping is hidden for non-US stores? This could be handled as a follow-up if not. This screenshot shows PayPal in second spot for Japan with the free-extenstions-test.zip rules:

When WooCommerce Shipping is unavailable, PayPal still sets the order value to 3 in the feed (rest api).

We can use rules to adjust the order value for other plugins, but it would be quite complex.

In my opinion, the best and simplest way to fix this issue is to apply the sorting [logic](https://github.com/woocommerce/woocommerce/pull/56233#discussion_r1986542479) in the frontend, as described by @ilyasfoo


>  also have a nit that can perhaps be addressed in a follow-up:
> Should this rule be an IN operation? It seems that value is being compared to a list of plugin names 🤔

`context_plugins` receives a full list of plugin objects. See https://github.com/woocommerce/woocommerce/pull/56233/files#diff-2f3016b95795e4ddd5bdda3de76f3b71604600347811b8e8c47bba6999ff9ba4R88

It then loops [through](https://github.com/woocommerce/woocommerce/pull/56233/files#diff-1ada547f41b559f1812eeb526c3ae8b3aa781a79fed1484d0e1ee081da93d20eR56) the list and runs a comparison operation. `in` won't work in this case as it's not a list of plugin names (string[]).

Issue #55531, opened 2025-02-17
[Settings] Implement proper script injection and state management


chihsuan:
In the WooCommerce Settings pages with custom views, we've identified a systemic issue where scripts are being injected multiple times without proper state management. This leads to:

1. Multiple script injections of the same code
2. Accumulated state from previous renders
3. Duplicate event listeners
4. Uncleaned component instances (e.g., modals, tooltips, etc.)
5. Potential memory leaks

This issue is currently visible in the Shipping settings where modals stack, but the problem is more fundamental and affects all custom views that inject scripts.

### Current Behavior
- Scripts are injected each time a page is rendered 
- State from previous renders persists
- Event listeners accumulate
- UI components (modals, tooltips, etc.) aren't properly cleaned up
- No centralized management of script lifecycle and state

### Expected Behavior
- Scripts should be injected only once or have proper re-injection handling
- State should be properly initialized and cleaned up between view changes
- Event listeners should be properly managed and removed when not needed
- Components should clean up their resources when unmounted
- Clear separation between different views' states

I'm not sure if it really possible to fix this issue because scripts that injected by third party plugins are not under our control. But we can try to find a solution to improve the situation and write a developer note for the community.

psealock:
@chihsuan thanks yes this is a difficult problem. I'm wondering if we can determine if a screen has scripts and if so render it in an iframe.

The problem is when we switch between tabs/sections DOM elements are removed, along with any DOM manipulations done via javascript and any event listeners. We have to then re-run the scripts each time navigating back to a section to ensure behaviour is re-established. As you mentioned, state is reapplied and can accumulate, causing various problems.

If a navigation to a section with scripts exist, we can render contents in an iframe. The iframe can then be removed on subsequent navigation.

My concern would be making sure the iframe's `window` has all the required properties. I'm wondering if something like this may be possible:

```js
const iframe = document.createElement('iframe');
document.body.appendChild(iframe);

const iframeWindow = iframe.contentWindow;

// Copy properties manually
iframeWindow.jQuery = window.jQuery;
iframeWindow.Backbone = window.Backbone;
```
chatGPT suggests inheritance may work, but I haven't tested this.

```js
const iframe = document.createElement("iframe");
document.body.appendChild(iframe);

Object.setPrototypeOf(iframe.contentWindow, window);
```

Another thing to explore is if ajax requests still work, especially if nonces are involved.

[...]

PR #56246, opened 2025-03-10
[Settings] Update to Calypso Site Admin package

### LINKED ISSUE #53818: [Settings] Update Router usage to Calypso site admin package

psealock:

Closes https://github.com/woocommerce/woocommerce/issues/53818

This PR introduces shared a8c components created in https://github.com/Automattic/wp-calypso/pull/99754, starting with just the Router.

The aim of this PR is to swap out the previous router from `@wordpress/edit-site` to `@calyspso/site-admin`. I wanted to use the new router without making any major refactors in the way data and routing is handled. You can see the routes passed to `RouterProvider` is just an empty array (for now), `<RouterProvider routes={ [] } pathArg="page">`.

### Notable changes

* I removed directly calling `possiblyRenderSettingsSlots` and related slot registrations. It turns out the entirety of Settings slots have been condensed to a single external script [wp-admin-scripts/settings-embed](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/admin/client/wp-admin-scripts/settings-embed/index.tsx). This external script gets executed just like any other script, including ones any extension may use to render slotFill. So by making this script work, we in turn assure 3PD scripts also work.
* I greatly simplified the root of the application, `plugins/woocommerce/client/admin/client/settings/index.js`, which previously was responsible for handling scripts. In order to keep the router and all its routes within the Settings Editor package, I needed to move the append scripts logic to the settings package, where it makes more sense anyways.
* There was a lot of noise in the console from plugins being re-registered. I added a check before plugins register to avoid re-registering. Because of this, testing with the `settings` feature flag is required.
* I moved `settingsScripts` into the data context.

1. ~Add/Enable Gutenberg 19.8 or earlier.~ Since https://github.com/woocommerce/woocommerce/pull/56271 has been merged, there is no longer a Gutenberg dependency.
2. Before turning on the feature flag for `settings`, poke around the current pages. Ensure pages with slotFill render normally, such as Site Visibility, Email, or Payments.
3. Turn on the `settings` feature flag.
4. Review the same settings screens and see the slotFills rendered. Navigate around the app and ensure content refreshes appropriately.
5. Install the [Migration Tester](https://github.com/woocommerce/woocommerce-settings-migration-tester) by running `npm run zip` from project root and installing the zip.
6. Test the various tabs in Legacy Settings and Modern Settings - they should function normally. Note Legacy Settings > Javascript isn't highlighting properly. I'd like to address this in a follow up because the same issue can be seen on `trunk`.

Interesting to note the stacked modal issue reported in [[Settings] Implement proper script injection and state management](https://github.com/woocommerce/woocommerce/issues/55531) is no longer visible. I don't think that issue is solved, but something has changed.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Update new Settings to use @automattic/site-admin's router instead of Gutenberg

</details>

<details>

#### Comment 

</details>


packages/js/settings-editor/package.json
```diff
@@ -34,6 +34,7 @@
 		"access": "public"
 	},
 	"dependencies": {
+		"@automattic/site-admin": "^0.0.1",
```
kalessil, line 37: It looks like a new dependency, we should add it to `.syncpackrc` to ensure the monorepo will use same version of this dependency.

rjchow (approved):
An inspection of the code in this PR looks ok to me, and I've tested the functionality of this PR when testing https://github.com/woocommerce/woocommerce/pull/56271 so I'm omitting testing here.

packages/js/settings-editor/package.json
kalessil: Thanks!

plugins/woocommerce/src/Admin/Features/Settings/Init.php
```diff
@@ -41,6 +41,7 @@ public function __construct() {
 		add_filter( 'woocommerce_admin_shared_settings', array( __CLASS__, 'add_component_settings' ) );
 		add_action( 'admin_enqueue_scripts', array( $this, 'enqueue_settings_editor_scripts' ) );
 		add_action( 'admin_enqueue_scripts', array( $this, 'enqueue_settings_editor_styles' ) );
+		add_action( 'woocommerce_navigation_is_connected_page', array( $this, 'disable_embed_if_woocommerce_settings_page' ) );
```
chihsuan: I found that the `payment-method-promotions` style is not being loaded because we disabled the embed page. This caused some styles to be missing.

Trunk:
Screenshot 2025-03-25 at 15 33 13

This branch:

Screenshot 2025-03-25 at 15 32 08

chihsuan (commented):
Nice work! @psealock 👍 

Overall, this looks good. Just found an issue that styles are not being loaded because we disabled the embed page as commented in the PR.

I've updated the PR to fix unit test issues by aligning specific WordPress packages with their project counterparts for the test environment, as seen in commit 3b8085e320b76dcb32a3a52599cce9f4997188df. Additionally, I updated @woocommerce/blocks to ise `@wordpress/private-apis` version wp-6.7, as it is required by @wordpress/editor wp-6.7, which is incompatible with `@wordpress/private-apis` version wp-6.6.

plugins/woocommerce/src/Admin/Features/Settings/Init.php
psealock: Good find @chihsuan, thank you.

I think excluding the new Settings as an `embed` page was too heavy handed. Many styles, including some from your screenshot, are actually included in [styles loaded based on the page being an embed](https://github.com/woocommerce/woocommerce/blob/f56bdb35838ab5f13c7e5067f0fb5d1c32730ae7/plugins/woocommerce/src/Internal/Admin/WCAdminAssets.php#L249). Instead, I returned Settings as an embed page, created a `PageController::is_modern_settings_page` function and conditionally loaded the embed header.

Fixed in [Return Settings to an 'embed' page](https://github.com/woocommerce/woocommerce/pull/56246/commits/3086594919daf1737f71a1e27954b28cd55b492a)

As far as the `payment-method-promotions` styles, they are loaded in a `admin_enqueue_scripts` hook. These don't fire when Settings loops through pages to gather assets and configs. I made a separate [issue to track this](https://github.com/woocommerce/woocommerce/issues/56737).

chihsuan (approved):
Thanks @psealock, good work on migrating to the Calypso Site Admin package! 👍 

I confirmed that styles are loaded correctly. While I also found a few issues such as can't save settings but I was able to reproduce them in trunk as well.
__

#### Test environment

<details>
- WordPress: 6.7.2
- Theme: Twenty Twenty-Five Version: 1.1
- WooCommerce - Development version with update/settings-router-calypso-package branch
- PHP: 8.3
- Web Server: Nginx
- Browser: Brave
- OS: macOS Sequoia 15.3.1
</details>

#### Review checklist

| Parameter  | Result |
| Approach  | ✅ Good migration strategy with backward compatibility |
| Automated tests  | ✅ Tests updated appropriately |
| Naming  | ✅ All good |
| Typing  | ✅ All good |
| Changelog  | ✅ Appropriate changelog entry |
| Documentation  | ✅ All good |
| PR description  | ✅ Clear and detailed |
| Testing instructions  | ✅ Comprehensive |
| Manual testing  | ✅ |

#### Scenarios tested

| Settings pages with slotFill | ✅ Function normally | |
| Legacy Settings tabs | ✅ Function normally | |
| Modern Settings tabs | ✅ Function normally | |
| Migration Tester | ✅ Works as expected | |

Screenshot 2025-03-27 at 10 48 37

Screenshot 2025-03-27 at 10 48 54

Screenshot 2025-03-27 at 10 49 03

Screenshot 2025-03-27 at 10 49 21
Screenshot 2025-03-27 at 10 49 49

Issue #54930, opened 2025-01-28
Deprecate `wc_current_theme_is_fse_theme()` in favor of `wp_is_block_theme()`


Aljullu:
We introduced [`wc_current_theme_is_fse_theme()`](https://github.com/woocommerce/woocommerce/blob/3229fb312a88e6bfe4fca7edaeeb52e1726713a2/plugins/woocommerce/includes/wc-conditional-functions.php#L515) long time ago, when WordPress core didn't have `wp_is_block_theme()`. Today, however, we could perfectly use the function from WP core instead of `wc_current_theme_is_fse_theme()`.

This issue is about:

1. Replacing all internal usages of `wc_current_theme_is_fse_theme()` with `wp_is_block_theme()`.
2. [Deprecating](https://github.com/woocommerce/woocommerce/blob/3229fb312a88e6bfe4fca7edaeeb52e1726713a2/docs/contributing/deprecation-in-core.md#how-do-we-deprecate-functions) `wc_current_theme_is_fse_theme()` and simplifying its logic, there is no longer need to check whether `wp_is_block_theme()` exists, as it will always exist.

Akshaykomar890:
Hi @Aljullu , I'd like to work on this issue. Could you please assign it to me if it's still available?

Aljullu:
Thanks for the interest, @Akshaykomar890! You can go ahead and work on it and create a PR as nobody else is working on it right now. 🙂 

Let us know if you need any help!

PR #56254, opened 2025-03-10
Deprecate wc_current_theme_is_fse_theme() in favor of wp_is_block_the…

rishiballa:
…me()

#### Significance
-  Minor

#### Type
- Enhancement - Improvement to existing functionality - #54930

#### Message

<details>

This fix is about:

Replacing all internal usages of wc_current_theme_is_fse_theme() with wp_is_block_theme().
[Deprecating](https://github.com/woocommerce/woocommerce/blob/3229fb312a88e6bfe4fca7edaeeb52e1726713a2/docs/contributing/deprecation-in-core.md#how-do-we-deprecate-functions) wc_current_theme_is_fse_theme() and simplifying its logic, there is no longer need to check whether wp_is_block_theme() exists, as it will always exist.

</details>


rishiballa:
fixes #54930

plugins/woocommerce/client/blocks/docs/internal-developers/testing/releases/761.md
```diff
@@ -15,7 +15,8 @@ Zip file for testing: [woocommerce-gutenberg-products-block.zip](https://github.
 | ----------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------ |
 | ![Screenshot showing a PHP warning](https://user-images.githubusercontent.com/3616980/172639823-1b671ac0-58ea-4f4b-9b00-aa481d4acd72.png) | ![Screenshot showing no PHP warning](https://user-images.githubusercontent.com/3616980/172639911-7a1416fe-f49d-4c58-90b7-7e07637b1777.png) |
 
-### Replace instances of wp_is_block_theme() with wc_current_theme_is_fse_theme(). ([6590](https://github.com/woocommerce/woocommerce-blocks/pull/6590))
+### Replace instances of wp_is_block_theme() with wp_is_block_theme()
```
samueljseay, line 18: It looks like you've done a blanket find/replace, but we don't want these docs to be edited.

plugins/woocommerce/includes/blocks/class-wc-brands-block-template-utils-duplicated.php
```diff
@@ -361,7 +361,8 @@ public static function should_use_blockified_product_grid_templates() {
 		$use_blockified_templates = wc_string_to_bool( get_option( Options::WC_BLOCK_USE_BLOCKIFIED_PRODUCT_GRID_BLOCK_AS_TEMPLATE ) );
 
 		if ( false === $use_blockified_templates ) {
-			return function_exists( 'wc_current_theme_is_fse_theme' ) && wc_current_theme_is_fse_theme();
+			return function_exists( 'wc_current_theme_is_fse_theme' ) && wp_is_block_theme()
```
samueljseay, line 364: This will break as we remove `wc_current_theme_is_fse_theme`. You'll need to remove the `function_exists` check

plugins/woocommerce/includes/class-wc-frontend-scripts.php
```diff
@@ -89,7 +89,8 @@ public static function get_styles() {
 					'media'   => 'all',
 					'has_rtl' => true,
 				),
-				'woocommerce-blocktheme'  => wc_current_theme_is_fse_theme() ? array(
+				'woocommerce-blocktheme'  => wp_is_block_theme()
+ ? array(
```
samueljseay, line 93: This indentation looks funky, are you using WooCommerce's PHP formatting tools?

plugins/woocommerce/includes/wc-update-functions.php
```diff
@@ -2740,7 +2740,8 @@ function wc_update_910_add_launch_your_store_tour_option() {
  * Add woocommerce_hooked_blocks_version option for existing stores that are using a theme that supports the Block Hooks API
  */
 function wc_update_920_add_wc_hooked_blocks_version_option() {
-	if ( ! wc_current_theme_is_fse_theme() && ! current_theme_supports( 'block-template-parts' ) ) {
+	if ( ! wp_is_block_theme()
+ && ! current_theme_supports( 'block-template-parts' ) ) {
```
samueljseay, line 2744: I won't list every time the formatting is bad, but it looks like you need to fix code formatting across the PHP files :)

plugins/woocommerce/includes/wc-conditional-functions.php
```diff
@@ -517,7 +517,8 @@ function wc_is_file_valid_csv( $file, $check_path = true ) {
  * @since 6.0.0
  * @return bool
  */
-function wc_current_theme_is_fse_theme() {
+function wp_is_block_theme()
```
samueljseay, line 520: We don't want to replace this. We want to add a deprecation here. You can use `wc_deprecated_function`

samueljseay (changes requested):
Thanks for initiating this change :)

It appears you've used a find/replace in your editor and that's not quite enough.

1. It has broken formatting in all the PHP files. I personally use https://marketplace.visualstudio.com/items?itemName=obliviousharmony.vscode-php-codesniffer in conjunction with the Woo code standards to autoformat PHP files to match the standard.

2. You've accidentally replaced some occurrences of strings that are not correct, in release docs and also accidentally overwritten the original function

3. If we deprecate we can use `wc_deprecate_function` to inform users to stop using this function, that code should either be a part of this PR, or it could be a separate PR, but in the case we're not deprecating here we should update the PR description ie, "Replace calls to `wc_current_theme_is_fse_theme` with `wp_is_block_theme`

gigitux:
Thanks for working on this, @rishiballa, and for the review, @samueljseay!  

Just a heads-up: we're currently working on a fix affecting some of these files—see [#56331](https://github.com/woocommerce/woocommerce/pull/56331).  

@rishiballa, it would be helpful to rebase this PR on top of #56331 or wait until it's merged.

Aljullu:
I'm going to close this PR, as this was already fixed in https://github.com/woocommerce/woocommerce/pull/56747. Thanks for your help, @rishiballa!

PR #56256, opened 2025-03-10
Product Filters: unified iAPI store

### LINKED ISSUE #54574: [Experimental] Product filters - iAPI one unified store
Out of the optimistic updates we're doing https://github.com/woocommerce/woocommerce/pull/53021, one question rose from the conversation of the structural design is if we could/should use one unified iAPI store for the Product Filters block. Ref pdnLyh-7Ok-p2#comment-5937

This issue is to discover if that is possible and create a PoC and/or a complete working solution.

### LINKED ISSUE #56570: Product Filters - Attributes show more not working
When the attributes list is long, there is a feature in place to hide part of the list and have a "show more" link and when click will reveal the rest of the attributes. Currently clicking on the link gives an error `Uncaught TypeError: Cannot set properties of undefined (setting 'showAll')`

Image

## Steps to reproduce

NOTE: Edit the https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterCheckboxList.php#L54 and change 15 to 2. This makes testing easier and not have to add a bunch of attributes.

1. Go to Appearance -> Editor -> Templates -> WooCommerce -> Product Catalog template
2. Add the Product Filters block to the page
3. Save and go to the frontend
4. You should see the "Show more..." link
5. Click on it and witness the console error

dinhtungdu:

- I have checked to ensure there aren't other open [Pull Requests](https://github.com/woocommerce/woocommerce/pulls) for the same update/change.

This PR refactors the Product Filters API to improve state management and code organization by:

- Converting multiple stores into a unified store under `woocommerce/product-filters`
- Simplifying the filter data structure by adding attribute slug to the filter item type, allowing attribute, stock, and rating to share the same actions
- Processing the filter count at the filter options block instead of including it with the title.
- Refactoring the price input handler to use `withScope`.

Closes #54574 #56570

(For Bug Fixes) Bug introduced in PR # N/A

| Before | After |
| ------ | ----- |
|        |       |

1. Add Product Filter to the Product Catalog template
2. Test the following filter interactions:
   - Select and unselect attribute filters
   - Apply price filters
   - Apply rating filters
   - Use the clear all filters button
3. Verify that no regression is introduced:
   - Filters are applied correctly and update the product list
   - Active filters are displayed properly
   - Filter counts are shown correctly
   - Clear all functionality works as expected
   - URL parameters are updated correctly when filters change

The changes have been tested in a development environment with:

- Default WordPress theme
- WooCommerce core plugin
- Various filter combinations including attributes, price ranges, and ratings

### Changelog entry

<details>

#### Significance

- [ ] Patch
- [ ] Minor
- [ ] Major

#### Type

#### Message

</details>

<details>

#### Comment

</details>


roykho:
Roping in Sam and Luis also to get their expertise.

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/frontend.ts
```diff
@@ -205,11 +226,11 @@ const productFiltersStore = store( 'woocommerce/product-filters', {
 				return ( window.location.href = url.href );
 			}
 
-			const { actions } = yield import(
+			const { actions: routerActions } = yield import(
```
samueljseay: I'm not sure we can get better types from generators but this would give the types and not result in type error. Still not ideal:

```ts
const routerModule: typeof import('@wordpress/interactivity-router') =
				yield import( '@wordpress/interactivity-router' );

yield routerModule.actions.navigate( url.href );
```

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/frontend.ts
```diff
@@ -224,4 +245,7 @@ const productFiltersStore = store( 'woocommerce/product-filters', {
 	},
 } );
 
-export type ProductFiltersStore = typeof productFiltersStore;
+export type ProductFiltersStore = {
```
samueljseay: This type is not a good representation of the full store.

How about changing the destructuring to export the full type:

```ts
const productFiltersStore = store(...);

const { state, actions } = productFiltersStore;

export type ProductFiltersStore = typeof productFiltersStore;
```

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterAttribute.php
```diff
@@ -121,20 +121,18 @@ function ( $item ) use ( $product_attributes_map ) {
...
-					'type'      => 'attribute',
-					'value'     => $term,
-					'label'     => $attribute_label . ': ' . $term_object->name,
-					'attribute' => array(
-						'slug'      => $product_attribute,
-						'queryType' => 'or',
-					),
+					'type'               => 'attribute/' . $product_attribute,
+					'value'              => $term,
+					'activeLabel'        => $attribute_label . ': ' . $term_object->name,
```
samueljseay: Why you hatin' on interpolated strings? 😂

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterCheckboxList.php
```diff
@@ -86,25 +79,32 @@ class="wc-block-product-filter-checkbox-list__item"
...
 									type="checkbox"
-									aria-label="<?php echo esc_attr( $aria_label ); ?>"
-									data-wp-on--change--parent-action="<?php echo esc_attr( $parent . '::actions.toggleFilter' ); ?>"
+									aria-label="<?php echo esc_attr( $item['ariaLabel'] ); ?>"
+									data-wp-on--change="actions.toggleFilter"
 									value="<?php echo esc_attr( $item['value'] ); ?>"
-									data-wp-bind--checked="<?php echo esc_attr( $parent . '::state.isItemSelected' ); ?>"
-									data-filter-item="<?php echo esc_attr( wp_json_encode( $item, JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_AMP ) ); ?>"
+									data-wp-bind--checked="state.isFilterSelected"
+									data-wp-context="<?php echo esc_attr( wp_json_encode( array( 'item' => $item ), JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_AMP ) ); ?>"
```
samueljseay: nitpick, I wonder if we should write a wrapper/helper for `escape_attr` that passes all the JSON encoding attributes. this gets ridiculously noisy in HTML rendering. Doesn't have to be done in this PR though of course.

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterRating.php
```diff
@@ -197,12 +193,8 @@ private function render_rating_label( $rating, $count_label ) {
 		?>
 		<div class="wc-block-components-product-rating">
 			<div class="wc-block-components-product-rating__stars" role="img" aria-label="<?php echo esc_attr( $rating_label ); ?>">
-				<span style="width: <?php echo esc_attr( $width ); ?>%" aria-hidden="true">
-				</span>
+				<span style="width: <?php echo esc_attr( $width ); ?>%" aria-hidden="true"></span>
```
samueljseay, line 196: Should this specify units? ie px?

samueljseay (approved):
A few small things but I'm pre-approving because none are blocking. I haven't undertaken testing for this PR, I'm just doing a code review.

Overall: fantastic work. This greatly simplifies the filters code and I love to see it.

roykho (approved):
Great work!

✅ Smoke tested the filters
✅ Applied different combination of settings
✅ Checked clear filters work and also individual removable chips

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterRemovableChips.php
```diff
@@ -58,36 +52,35 @@ protected function render( $attributes, $content, $block ) {
...
+							<?php echo esc_html( $item['activeLabel'] ); ?>
 						</span>
 						<button
 							type="button"
 							class="wc-block-product-filter-removable-chips__remove"
 							aria-label="<?php echo esc_attr( $remove_label ); ?>"
-							data-wp-on--click="<?php echo esc_attr( $parent_block . '::actions.removeFilter' ); ?>"
-							data-filter-item="<?php echo esc_attr( wp_json_encode( $item, JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_AMP ) ); ?>"
+							data-wp-on--click="actions.removeActiveFilter"
+							data-wp-context="<?php echo esc_attr( wp_json_encode( array( 'item' => $item ), JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_AMP ) ); ?>"
```
luisherranz: There is a helper for this case:

```suggestion
							<?php echo wp_interactivity_data_wp_context( array( 'item' => $item ) ); ?>
```

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterPriceSlider.php
```diff
@@ -143,15 +138,13 @@ class="max"
...
 							data-wp-on--focus="actions.selectInputContent"
-							data-wp-on--input="actions.debounceSetPrice"
-							data-wp-on--change--set-price="woocommerce/product-filter-price::actions.setMaxPrice"
-							data-wp-on--change--navigate="woocommerce/product-filters::actions.navigate"
+							data-wp-on--input="actions.debounceSetMaxPrice"
 							aria-label="<?php esc_attr_e( 'Filter products by maximum price', 'woocommerce' ); ?>"
 						/>
 					<?php else : ?>
-					<span data-wp-text="woocommerce/product-filter-price::state.formattedMaxPrice"><?php echo wp_kses_post( $formatted_max_price ); ?></span>
+					<span data-wp-text="state.formattedMaxPrice"><?php echo wp_kses_post( $formatted_max_price ); ?></span>
```
luisherranz: The Interactivity API handles server-side rendering but you need to populate the state in the server.

```php
wp_interactivity_state( '...', array(
    'maxPrice'          => $max_price,
    'formattedMaxPrice' => wc_price( $max_price, array( 'decimals' => 0 ) ),
));
```

```suggestion
					<span data-wp-text="state.formattedMaxPrice"></span>
```

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/frontend.ts
```diff
@@ -127,52 +141,62 @@ const productFiltersStore = store( 'woocommerce/product-filters', {
...
 		) => {
 			const context = getContext< ProductFiltersContext >();
 			context.activeFilters = context.activeFilters.filter(
 				( item ) => ! callback( item )
 			);
 		},
-		removeActiveFiltersByType: ( type: ActiveFilter[ 'type' ] ) => {
-			productFiltersStore.actions.removeActiveFiltersBy(
-				( item ) => item.type === type
+		_selectFilter: () => {
```
luisherranz: You don't need to put _private actions_ in the store. You can transform them into normal functions. You can access context and derived state inside those functions without any problem because the scope is passed down from the action to the function automatically (only functions that are not called synchronously need to be wrapped with `withScope`).

```js
function selectFilter() {
  const context = getContext< ProductFiltersContext >();
	// ...
}
```

Then:

```js
  toggleFilter: () => {
			if ( state.isFilterSelected ) {
				unselectFilter();
			} else {
				selectFilter();
			}
			actions.navigate();
		},
```

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/inner-blocks/clear-button/frontend.ts
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Internal dependencies
+ */
+import { ProductFiltersContext } from '../../frontend';
+
+type ClearButtonContext = {
+	parent: string;
+};
+
+const { actions } = store( 'woocommerce/product-filters', {
```
luisherranz: You can import the main type of the store here, to type `actions.navigate`.


```suggestion
const { actions } = store< ProductFilter >( 'woocommerce/product-filters', {
```

Also, you can join it to its own inferred types, or other store part types, depending on what this store part should be aware of.

More information on how to work with TypeScript in the Interactivity API in [this guide](https://developer.wordpress.org/block-editor/reference-guides/interactivity-api/core-concepts/using-typescript/).

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/inner-blocks/clear-button/frontend.ts
```diff
@@ -0,0 +1,29 @@
...
+	actions: {
+		clearFilters: () => {
+			const context = getContext<
+				ProductFiltersContext & ClearButtonContext
+			>();
+
+			if ( context.parent === 'woocommerce/product-filter-active' ) {
```
luisherranz: Leftover? I can't see any reference to `parent` in the context anymore.

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterCheckboxList.php
luisherranz: There's [`wp_interactivity_data_wp_context()`](https://developer.wordpress.org/reference/functions/wp_interactivity_data_wp_context/) in Core.

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterActive.php
```diff
@@ -30,16 +33,20 @@ protected function render( $attributes, $content, $block ) {
...
 		$wrapper_attributes = array(
-			'data-wp-interactive'  => $this->get_full_block_name(),
 			'data-wp-key'          => wp_unique_prefixed_id( $this->get_full_block_name() ),
-			'data-wp-bind--hidden' => '!state.hasSelectedFilters',
-			/* translators:  {{label}} is the label of the active filter item. */
-			'data-wp-context'      => wp_json_encode( array( 'removeLabelTemplate' => __( 'Remove filter: {{label}}', 'woocommerce' ) ), JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_AMP ),
+			'data-wp-context'      => wp_json_encode(
+				array(
+					/* translators:  {{label}} is the label of the active filter item. */
+					'removeLabelTemplate' => __( 'Remove filter: {{label}}', 'woocommerce' ),
```
luisherranz: If this translation does not vary, it is not necessary to include it in each context. It can be included only once using [`wp_interactivity_config`](https://developer.wordpress.org/block-editor/reference-guides/interactivity-api/api-reference/#wp_interactivity_config).

luisherranz (commented):
I've only taken a quick look at the code, but in principle, the change to unify all the blocks in the same namespace seems to make sense.

I'll leave some comments here for improvements.

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterActive.php
dinhtungdu: `removeLabelTemplate` only belongs to Active Fitlers. `activeLabelTemplate` is different based on the filter type. So I think it's better in this case to use context.

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterCheckboxList.php
dinhtungdu: @luisherranz looks like there isn't a function that only returns data, not the whole attribute string. Do you think we should add one, to use for `get_block_wrapper_attributes`, for example?

dinhtungdu: I refactored some places to use `wp_interactivity_data_wp_context` in 9200fab932e1.

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/frontend.ts
dinhtungdu: What should I do if I need to access store `actions` or `state` inside the function, like the `_unselectFilter()` case? It will still work but Eslint will complain about 'actions' was used before it was defined.

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/frontend.ts
dinhtungdu: This is actually my first implementation, but Eslint complained about actions/state was used before it was defined. So I had to manual type the store to make it happy.

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/frontend.ts
dinhtungdu: Thanks for the tip! I applied it in 7ed4b86970c5.

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/frontend.ts
luisherranz: I deleted that rule for the frontend files since it's not necessary.

- https://github.com/woocommerce/woocommerce/pull/56117/files#diff-ac2e79e01fc0519e24640ff7b3f71c2bc010d022770449cd91ccb93d6730d265R322-R327

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/frontend.ts
luisherranz: This -> https://github.com/woocommerce/woocommerce/pull/56256#discussion_r2005550220 🙂

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterPriceSlider.php
dinhtungdu: Thanks, I applied this in 6b9fedd68041

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterRemovableChips.php
dinhtungdu: Addressed in 9200fab932e1

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterRating.php
dinhtungdu: The rating size can be vary, so we need to use relative unit here.

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterAttribute.php
dinhtungdu: Haha, I was quickly testing something and forgot edit. Addressed in 2de833762bf4.

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/frontend.ts
dinhtungdu: Great! I updated this PR to use type inference in d9424fc5061c

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/frontend.ts
dinhtungdu: Addressed in d9424fc5061c

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/inner-blocks/clear-button/frontend.ts
dinhtungdu: Addressed in d9424fc5061c

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/inner-blocks/clear-button/frontend.ts
dinhtungdu: Yeah, the whole file is left over. I deleted it in 1a33b904007a.

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/frontend.ts
```diff
@@ -127,52 +172,31 @@ const productFiltersStore = store( 'woocommerce/product-filters', {
...
+		toggleFilter: () => {
+			if ( state.isFilterSelected ) {
+				unselectFilter();
+			} else {
+				selectFilter();
+			}
+			actions.navigate();
 		},
-		*navigate() {
+		*navigate(): Generator {
```
luisherranz, line 195: Could you please make a PR in [this repo](https://github.com/woocommerce/gutenberg) to [export `NavigateOptions` and `PrefetchOptions`](https://github.com/woocommerce/gutenberg/blob/5cc4b71f45572433468892968fa3df27c7c93a8d/packages/interactivity-router/src/index.ts#L24-L36) and add a TODO commment here to delete the `Generator` type with the link of that PR?

Use the [`interactivity-api` branch](https://github.com/woocommerce/gutenberg/tree/interactivity-api) in that repository as the base for that PR (don't open the PR against `WordPress/gutenberg`).

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/inner-blocks/clear-button/frontend.ts
luisherranz: Thanks.

What are your thoughts about this way of working with types and sharing them among the different store parts?

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterActive.php
luisherranz: What I'm saying is that there's no need to use `context` to store a string that doesn't change depending on its position in the DOM, doesn't need to be reactive, and you simply need to serialize it from the server. For those cases, you can also use `wp_interactivity_config`. But up to you 🙂

luisherranz (approved):
LGTM code-wise. Thanks for reviewing my suggestions, Tung 🙂

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterActive.php
dinhtungdu: I see. `removeLabelTemplate` and `activeLabelTemplates` don't need to be reactive. I thought that'd make it (mentally) easier for the dev to get the label template only from context. But using context in terms of reactivity is a different school of thoughts and it makes a lot of sense too. It tells us the data is reactive without explicitly saying that.

dinhtungdu: Addressed in f6fb544

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/frontend.ts
dinhtungdu: I added a todo in eb7e5509afa5. And here is the PR https://github.com/woocommerce/gutenberg/pull/8.

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterActive.php
luisherranz: It's not a big deal, but that'd be the correct way to do it in my opinion 🙂

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/frontend.ts
luisherranz: Thanks!

PR #56258, opened 2025-03-10
Don't load order item from db on `ajax_order_item`

migueldamota:

This PR updates the `maybe_add_order_item` function to reload the order before generating HTML fragments to show the new order item with also new added meta data or other changes.

1. Use this PR for the plugin
2. Create a filter for `woocommerce_ajax_order_item`
```php
add_filter( 'woocommerce_ajax_order_item', 'ajax_order_item', 10, 4 );
```
3. Create a function for that filter and e.g. update a meta data for that item
```php
function ajax_order_item( $item, $item_id, $order, $product ) {		
    $item->update_meta_data( 'a_meta_key', 'a_meta_value' );
    $item->save_meta_data();
    return $item;
}
```
4. Go to an order and add a new order item
5. After adding, the new meta data should be visible instantly, as you can see in the screenshot
image

------

Before these changes the meta data would only be visible after reload or when a new item was added (Screenshot)
image

After new item added:
image

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message
Reload order when adding an item via `add_order_item` ajax request before generating HTML fragments

</details>

<details>

#### Comment 

</details>


albarin:
Hey @migueldamota! Thanks for your contribution! 🙌 
Could you please add more detailed testing steps? (like the specific `woocommerce_ajax_order_item` filter code, etc.).
It would help us a lot in the review.
Thank you!

migueldamota:
Hi @albarin, yeah sure! I'll put them in this comment:

1. Use this PR for the plugin
2. Create a filter for `woocommerce_ajax_order_item`
```php
add_filter( 'woocommerce_ajax_order_item', 'ajax_order_item', 10, 4 );
```
3. Create a function for that filter and e.g. update a meta data for that item
```php
function ajax_order_item( $item, $item_id, $order, $product ) {		
    $item->update_meta_data( 'a_meta_key', 'a_meta_value' );
    $item->save_meta_data();
    return $item;
}
```
4. Go to an order and add a new order item
5. After adding, the new meta data should be visible instantly, as you can see in the screenshot
image

------

Before these changes the meta data would only be visible after reload or when a new item was added (Screenshot)
image

After new item added:
image

albarin:
@migueldamota Thanks so much for the updated testing steps and sorry for the delay 🙏 

After reviewing, we have a suggestion: instead of not loading the item from the DB, we think it would be more consistent to reload the whole order again before rendering, right before the `include __DIR__ . '/admin/meta-boxes/views/html-order-items.php'`.
Do you think you could update the PR to follow that approach?
Thanks so much! 🙌

migueldamota:
Hi @albarin. No problem at all.

Do you mean calling `$order = wc_get_order( $order_id );`? If so, I could add that yes.

Quick question to my change, should I leave it or revert it? Because logically I would leave it because in my head it doesn't make sense to create a new item and then fetch it again from the database.
Also why exactly do you want to call `wc_get_order()` again? Isn't the current `$order` enough to be passed down? Besides the new created item

albarin:
> Do you mean calling `$order = wc_get_order( $order_id );`? If so, I could add that yes.

Yes, that's what I mean 👍 


> Quick question to my change, should I leave it or revert it? Because logically I would leave it because in my head it doesn't make sense to create a new item and then fetch it again from the database.
Also why exactly do you want to call wc_get_order() again? Isn't the current $order enough to be passed down? Besides the new created item.

I think revert it. You are making a good point. The reasoning is that, after the `woocommerce_ajax_order_item,` there are other actions and filters where things could change, and it seems better to refresh the whole order instead of tracking exactly what changed on each of them.

Thanks!

albarin:
Thanks, @migueldamota. Would you also mind updating the PR description and changelog message to reflect the new implementation changes? 🙏 Thank you!

migueldamota:
@albarin Yeah sure!

migueldamota:
@albarin Does this then has to be a minor bump or is a fix enough here? If a fix is enough, I would update the Significance

albarin:
@migueldamota I think a fix would be enough.

albarin (approved):

PR #56262, opened 2025-03-10
[e2e-tests] Make tests compatible with WP v6.8

WunderBart:
## What?

Related thread: p1741273072772969-slack-C02FL3X7KR6

Fixing the failing tests by providing necessary updates to align with WP v6.8 while keeping the v6.7 backwards compatibility.

## How to test

CI should pass.


WunderBart:
I've updated the tests to align with 6.8, but there are still failures that seem to be pointing to an actual regression:

1. The `Edit site` button in admin bar opens the homepage instead of the template: 

```
[chromium] › tests/e2e/tests/templates/cart-template.block_theme.spec.ts:51:6 › Test the cart template › Admin bar edit site link opens site editor 
[chromium] › tests/e2e/tests/templates/checkout-template.block_theme.spec.ts:55:6 › Test the checkout template › Admin bar edit site link opens site editor 
```

cc: @woocommerce/rubik, is this an expected behavior for the v6.8?

2. Inspector controls for the `woocommerce/product-filter-active`'s chips block does not have the "space between items" control:

```
[chromium] › tests/e2e/tests/product-filters/removable-chips-editor.block_theme.spec.ts:48:2 › woocommerce/product-filter-active › should display the correct inspector layout controls 
[chromium] › tests/e2e/tests/product-filters/removable-chips-editor.block_theme.spec.ts:76:2 › woocommerce/product-filter-active › should add correct layout CSS class when modifying layout settings 
```

cc: @roykho @dinhtungdu do you know if it's expected that setting to be unavailable in v6.8?

---

✅ _edit: Both issues have been fixed._

plugins/woocommerce/client/blocks/tests/e2e/tests/accordion/accordion.block_theme.spec.ts
```diff
@@ -36,13 +36,16 @@ test.describe( `${ blockData.slug } Block`, () => {
...
+
 			await expect(
-				editor.insertBlock( { name: blockData.slug } )
-			).rejects.toThrow(
-				new RegExp(
-					`Block type '${ blockData.slug }' is not registered.`
-				)
-			);
+				await editor.getBlockByName( blockData.slug )
+			).toBeHidden();
```
gigitux, line 48: I'm just curious - in WordPress 6.8 this exception isn't raised anymore?

plugins/woocommerce/client/blocks/tests/e2e/tests/checkout/checkout-block.merchant.block_theme.spec.ts
```diff
@@ -147,72 +147,72 @@ test.describe( 'Merchant → Checkout', () => {
...
-			'You must accept our Terms and Conditions and Privacy Policy to continue with your purchase.'
-		);
-		await expect( checkboxWithError ).toHaveAttribute(
-			'aria-invalid',
-			'true'
-		);
 
-		await frontendUtils.page
-			.getByLabel(
+	// This test is flaky. Skipping for now.
```
gigitux: Should we open an issue for this?

plugins/woocommerce/client/blocks/tests/e2e/tests/templates/constants.ts
```diff
@@ -126,7 +126,7 @@ export const CUSTOMIZABLE_WC_TEMPLATES: TemplateCustomizationTest[] = [
 
 			await page.goto( '/product/wordpress-pennant/' );
 		},
-		templateName: 'External Product Add to Cart with Options',
+		templateName: 'External Product Add To Cart With Options',
```
gigitux, line 129: Is there a particular reason for this change? Or just for consistency?

gigitux (approved):
Thanks for working on this! Awesome work! 🚀 

I left a few questions, but mostly for my curiosity.

Do you think that we should open a dedicated issue for keeping track of WP 6.7 compatibility to ensure that we can easily clean up them once that WP 6.9 is out?

plugins/woocommerce/client/blocks/tests/e2e/tests/accordion/accordion.block_theme.spec.ts
WunderBart: It's different - Instead of a rejected promise in v6.7, it now throws a handled error in the canvas stating that the block is not supported. Checking if the block simply isn't there works for both scenarios.

plugins/woocommerce/client/blocks/tests/e2e/tests/checkout/checkout-block.merchant.block_theme.spec.ts
WunderBart: No need, some of those should be addressed already by https://github.com/woocommerce/woocommerce/pull/55989. I will rebase and see if these skips are still needed.

plugins/woocommerce/client/blocks/tests/e2e/tests/templates/constants.ts
WunderBart: I updated the template search logic which resulted in case-sensitive matching. The original name is uppercased so I had to update it in the constant there.

WunderBart:
> Do you think that we should open a dedicated issue for keeping track of WP 6.7 compatibility to ensure that we can easily clean up them once that WP 6.9 is out?

@gigitux, I think we could just create a reminder-issue to revisit the compatibility patches when v6.9 is out. It should be easy to find them as I tagged them all with the `Keep WP v6.7 compatibility` comment.

WunderBart:
The [current failures](https://github.com/woocommerce/woocommerce/actions/runs/13900572157?pr=56262) are coming from flaky tests. Going to rebase and hope for the best!

plugins/woocommerce/client/blocks/tests/e2e/tests/templates/cart-template.block_theme.spec.ts
```diff
@@ -45,14 +44,6 @@ test.describe( 'Test the cart template', () => {
...
 		).toBeVisible();
 	} );
-
-	test( 'Admin bar edit site link opens site editor', async ( { admin } ) => {
-		await admin.page.goto( permalink );
-		await admin.page.locator( '#wp-admin-bar-site-editor a' ).click();
-		await expect(
-			admin.editor.canvas.getByLabel( 'Block: Title' )
-		).toBeVisible();
-	} );
```
WunderBart, line 55: Removed as per p1741950598023229/1741710131.611319-slack-C02FL3X7KR6.

plugins/woocommerce/client/blocks/tests/e2e/tests/templates/checkout-template.block_theme.spec.ts
```diff
@@ -51,26 +50,6 @@ test.describe( 'Test the checkout template', () => {
...
-
-		// Close welcome popup.
-		await admin.page.getByRole( 'button', { name: 'Get started' } ).click();
-
-		await expect(
-			editor.canvas.getByRole( 'button', {
-				name: 'Place Order',
-			} )
-		).toBeVisible();
-	} );
```
WunderBart, line 73: Removed as per p1741950598023229/1741710131.611319-slack-C02FL3X7KR6.

gigitux:
> > Do you think that we should open a dedicated issue for keeping track of WP 6.7 compatibility to ensure that we can easily clean up them once that WP 6.9 is out?
> 
> @gigitux, I think we could just create a reminder-issue to revisit the compatibility patches when v6.9 is out. It should be easy to find them as I tagged them all with the `Keep WP v6.7 compatibility` comment.

It makes sense for me 👍

plugins/woocommerce/client/blocks/tests/e2e/tests/checkout/checkout-block.merchant.block_theme.spec.ts
WunderBart: Follow up: p1742291538937259/1741293577.344589-slack-C03CPM3UXDJ

plugins/woocommerce/client/blocks/tests/e2e/utils/admin/index.ts
```diff
@@ -6,24 +6,26 @@ import { Admin as CoreAdmin } from '@wordpress/e2e-test-utils-playwright';
...
-			window.wp.data
-				.dispatch( window.wp.preferences.store )
-				.set( 'core/edit-widgets', 'welcomeGuide', false );
-			return (
-				window.wp.data
-					.select( window.wp.preferences.store )
-					.get( 'core/edit-widgets', 'welcomeGuide' ) === false
-			);
-		} );
+		await this.page
```
opr, line 9: Hi @WunderBart I see this test is [being marked as flaky again](https://github.com/woocommerce/woocommerce/issues/48186#issuecomment-2743126698)

Is there a reason you changed back to closing the dialog "manually"? I know that doing it this way isn't 100% in the spirit of E2E tests, but given that the welcome guide is not the system under tests, and in light of the flakiness I think using the data store to guarantee the welcome guide is closed is acceptable.

Please let me know your thoughts 🙇🏼

WunderBart: The previous implementation was extremely flaky when tested against v6.8-betaX as the Welcome Guide kept reopening, while the manual closing seemed to have fixed the issue in a backwards compatible way. We can try reverting this change, but we'll also need to ensure it's stable against 6.8 before merging. Does that make sense?

WunderBart: > Welcome Guide kept reopening

This might be an issue to investigate on its own, but let's first try running against the beta3 and see if it's still flaky! 🤞

opr: Yes it does make sense, thanks for the explanation. How can I test against beta3 locally? I am happy to take care of this but would need help with setup, thanks

WunderBart: Attaching the flaky failure trace for reference: [trace.zip](https://github.com/user-attachments/files/19446813/trace.zip)

☝️ FWIW, it looks like the Welcome Guide never opened so the utility failed waiting on it. 

> How can I test against beta3 locally?

You need to update [the wp-env config](https://github.com/woocommerce/woocommerce/blob/e1402912df15b70859143e8135cff77ff8d3dafb/plugins/woocommerce/client/blocks/.wp-env.json#L2) by replacing the `wordpress-latest` URL with https://wordpress.org/wordpress-6.8-beta3.zip. After that you'll need to run the following to update the environment:

```
pnpm --filter='@woocommerce/block-library' wp-env start --udpate
```

...or just reset the entire env with:

```
pnpm --filter='@woocommerce/block-library' env:restart
```

WunderBart: BTW, when the PR is opened, you should also be able to manually run the `daily run` workflow against your branch which runs the E2E tests against the current, previous and next WP versions. The workflow can be triggered from here: 

https://github.com/woocommerce/woocommerce/actions/workflows/tests-daily-run.yml

Screenshot 2025-03-25 at 12 23 50

WunderBart: Left a comment on a related issue in Gutenberg: https://github.com/WordPress/gutenberg/issues/69276#issuecomment-2751176703

opr: For now should we consider skipping or even just removing this test? The flakiness is a real blocker and this isn't a critical flow. What do you think @WunderBart ?

WunderBart: Yeah, I think we can mark the tests as `fixme` as we did here https://github.com/woocommerce/woocommerce/pull/56517. Let's also create an issue to remember to investigate that test. How does that sound?

PR #56268, opened 2025-03-10
tweak: sanitization on shipping and billing address is happening twice

Tofandel:

This is just a small tweak to avoid sanitizing twice the shipping and billing data (expand the diff and you will see it is already done just below)

1. Go to the checkout (block checkout)
2. Update your address
3. It will still update correctly

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 
tweak: sanitizate shipping and billing address only once

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/StoreApi/Routes/V1/CartUpdateCustomer.php
```diff
@@ -132,26 +132,19 @@ protected function get_route_post_response( \WP_REST_Request $request ) {
...
 		// Get data from request object and merge with customer object, then sanitize.
-		$billing  = $this->schema->billing_address_schema->sanitize_callback(
-			wp_parse_args(
-				$request['billing_address'] ?? [],
-				$this->get_customer_billing_address( $customer )
-			),
-			$request,
-			'billing_address'
-		);
-		$shipping = $this->schema->billing_address_schema->sanitize_callback(
```
Tofandel, line 143: Also this was the wrong schema to run against, this doesn't cause any issue in the core because the fields are the same except for the email field which is optional, but it might cause issues if extensions adds required fields only in billing but not in shipping

alexflorisca (dismissed):
Thank you for contributing to WooCommerce. The code changes look good. I tested that both shipping and billing addresses persist. I also tested an xss string in both addresses to make sure the sanitisation works properly. 👍

nerrad:
It looks like trunk already has a correction for the double sanitization. It must have got addressed since this was changed.

Issue #52520, opened 2024-11-03
[Settings] Use Calypso Site Admin sidebar components


psealock:
https://github.com/Automattic/wp-calypso/pull/99754 Provides a new `automattic/site-admin` package that contains a sidebar components for use in new site admins.

Remove usage of Sidebar components from `@wordpress/edit-site` and use the new package from Calypso instead.

This is closely related to https://github.com/woocommerce/woocommerce/issues/53818.

psealock:
https://github.com/woocommerce/woocommerce/pull/52655 grabs the Gutenberg components directly from NPM. Next, we can grab and distribute them from `@woocommerce/admin-layout` to Settings and Product Editors. But first, https://github.com/woocommerce/woocommerce/issues/52635 must be fixed so we can test.

I will mark this issue as blocked until https://github.com/woocommerce/woocommerce/issues/52635 is resolved.

psealock:
These elements are now being gathered directly from `@wordpress/edit-site` with no manipulation, so Product Editor can do the same as there is little need to pull them from a single re-exported place. 

The only reason why a shared version would be useful is if we run into privateApi issues with `unlock` when we release to production. We can reopen this issue at that time if need be.

psealock:
Reopening because Analytics will use the same layout

psealock:
Marking this as blocked until https://github.com/woocommerce/woocommerce/pull/53531 is resolved so we can update dependencies to fix failing tests

psealock:
I have updated this issue to reflect ongoing work in https://github.com/Automattic/wp-calypso/pull/99754 and removing the blocked status.

PR #56271, opened 2025-03-11
[Settings] Update sidebar components calypso package

psealock:

Closes https://github.com/woocommerce/woocommerce/issues/52520

This PR swaps out Gutenberg edit site components for the ones provided by `@automattic/site-admin`. This removes the dependency on Gutenberg 19.8.

Note that the SiteHub component has been removed temporarily because its not available just until https://github.com/Automattic/wp-calypso/pull/99784 is merged and released. I will handle this in a follow up.

1. Turn on the `settings` feature flag and **disable** Gutenberg if you are testing this locally.
2. Head to WooCommerce > Settings and poke around.
3. Make sure slotFill screens function appropriately.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


packages/js/settings-editor/src/components/sidebar/index.tsx
```diff
@@ -2,18 +2,15 @@
  * External dependencies
  */
 import { createElement } from '@wordpress/element';
-/* eslint-disable @woocommerce/dependency-group */
+import { addQueryArgs } from '@wordpress/url';
```
rjchow, line 5: Most of the stuff in `@wordpress/url` can be handled with [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) but if we're using it for consistency then fine by me

plugins/woocommerce/src/Admin/Features/Features.php
```diff
@@ -313,16 +313,16 @@ public static function maybe_load_beta_features_modal( $hook ) {
...
 	public static function load_scripts() {
-		if ( ! PageController::is_admin_or_embed_page() ) {
-			return;
-		}
-
-		$features         = self::get_features();
-		$enabled_features = array();
-		foreach ( $features as $key ) {
-			$enabled_features[ $key ] = self::is_enabled( $key );
+		$settings_page_feature_enabled = self::is_enabled( 'settings' ) && get_current_screen()->id === 'woocommerce_page_wc-settings';
```
rjchow: Nitpicky comment here, but I'd name this `$is_on_modern_settings_page`, I had a bit of a brain tangle when reading the below logic, which reads as "if is admin page OR settings feature enabled", implying that if settings feature is enabled then it would be loaded on all pages - which is not the case

plugins/woocommerce/src/Admin/Features/Settings/Init.php
```diff
@@ -53,6 +54,21 @@ public function is_settings_page() {
...
 	}
 
+	/**
+	 * Turn off the embed page if the current page is the WooCommerce settings page.
+	 *
+	 * @param bool $is_connected_page Whether the current page is a connected page.
+	 * @param string $current_page The current page.
+	 * @return bool
+	 */
+	public function disable_embed_page_for_settings( $is_connected_page, $current_page = null ) {
```
rjchow:
```suggestion
	public function disable_embed_if_woocommerce_settings_page( $is_connected_page, $current_page = null ) {
```

Naming nitpick again 😅

plugins/woocommerce/src/Admin/Features/Settings/Init.php
```diff
@@ -53,6 +54,20 @@ public function is_settings_page() {
 		return $screen && 'woocommerce_page_wc-settings' === $screen->id;
 	}
 
+	/**
+	 * Turn off the embed page if the current page is the WooCommerce settings page.
+	 *
+	 * @param bool $is_connected_page Whether the current page is a connected page.
+	 * @return bool
+	 */
+	public function disable_embed_page_for_settings( $is_connected_page ) {
```
rjchow:
```suggestion
	public function disable_embed_if_woocommerce_settings_page( $is_connected_page ) {
```
Naming nitpick again 😅

rjchow (commented):
I'm not sure if it's a result of this PR but I think it would be relevant to fix here - seems like there are two sets of superfluous scrollbars

image

The 'Settings saved successfully' notice also doesn't auto-dismiss nor can it be manually dismissed

plugins/woocommerce/src/Admin/Features/Features.php
psealock: "modern" in this context has been used for new pages that are not legacy, but brand new. How about the very verbose `$is_on_settings_page_with_feature_enabled`?

psealock:
>I'm not sure if it's a result of this PR but I think it would be relevant to fix here - seems like there are two sets of superfluous scrollbars

Thanks @rjchow, I'm not seeing that locally or on JN but I wonder if CSS isn't loading properly because have seen that happen from time to time. Maybe its best to keep an eye on this going forward.

>The 'Settings saved successfully' notice also doesn't auto-dismiss nor can it be manually dismissed

Noted, thank you. I will address in a follow up

rjchow (approved):
This LGTM! I tested according to the instructions and found nothing noteworthy. 

Thanks for addressing the feedback!

psealock:
CI errors are caused be the private API setup somehow changed. The tests are trying to unlock modules that haven't yet been locked. Since the same issue is seen on https://github.com/woocommerce/woocommerce/pull/56246, I will merge this into the base branch and attempt to address the issue there.

Issue #56002, opened 2025-03-03
Multiple duplicate queries to for shipping zone data on Add New Post and Add New Page


nicdwilson:
### Prerequisites

- [x] I have carried out troubleshooting steps and I believe I have found a bug.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

When opening the `Add New Post` and the `Add New Page` multiple queries for shipping data are triggering.  This was discovered in 9474675-zd-a8c on a new site (1 order, test data only) while investigating 503 errors occurring when opening the `Add New Post` page.

As the site was a new site, about to start development, there were no shipping zones or methods yet registered. This appeared to cause the 503 and intense resource usage when opening this page.

Adding a dummy zone and method resulted in the resource usage returning to normal, and the Add Post page working as expected.

### Expected behavior

Duplicate queries should not occur, and shipping setup might not be completed for some time. I'm not entirely sure of why these queries are running on this page (due to checkout block requirements?)

### Actual behavior

```
Automattic\WooCommerce\Blocks\Utils\CartCheckoutUtils::get_shipping_zones()
Automattic\WooCommerce\Blocks\Shipping\ShippingController->hydrate_client_settings
```

Run multiple queries.

![Image](https://github.com/user-attachments/assets/6731245d-d48b-4cf5-ae38-9cf49f5e1f31)

### Steps to reproduce

I was unable to reproduce this extreme resource usage on an established site, even after removing the shipping zones. I was able to view the duplicate queries though.

Noting that the original environment was running PHP 8.4 and Litespeed and I was unable to test on an equivalent environment for replication.

Once shipping zones were added, duplicate queries dropped and resource usage appeared to return to normal.

### WordPress Environment

<details><summary>System Status Report</summary>

```
### WordPress Environment ###

WordPress address (URL): [Redacted]
Site address (URL): [Redacted]
WC Version: 9.7.0
Legacy REST API Package Version: The Legacy REST API plugin is not installed on this site.
Action Scheduler Version: ✔ 3.9.0
Log Directory Writable: ✔
WP Version: 6.7.2
WP Multisite: –
WP Memory Limit: 512 MB
WP Debug Mode: –
WP Cron: –
Language: en_US
External object cache: –

### Server Environment ###

Server Info: LiteSpeed
PHP Version: 8.4.3
PHP Post Max Size: 256 MB
PHP Time Limit: 360
PHP Max Input Vars: 8000
[...]

PR #56297, opened 2025-03-11
Avoid loading settings for wc-shipping-method-pickup-location when not enqueued

### LINKED ISSUE #54913: [Performance]: Inefficient querying/loading of shipping data on the Every Admin Page
### Prerequisites

### Performance Details

The `Automattic\WooCommerce\Blocks\Shipping\ShippingController->hydrate_client_settings()` is extremely inefficient and likely doesn't perform well with a large set of shipping data.  On every admin page the contorller:

* Loads and renders all shipping zones to settings, regardless of the number of zones and regardless of the current page being shown in the admin.
* Runs 3 queries per shipping zone during the construction of each `WC_Shipping_Zone` construction.  E.g. 200 shipping zones results in 600 queries being run for the page.

### Affected Sites

This affects sites that have a large number of shipping zones and locations.  The priority of improvements here will depend on how common this scenario is.

### Suspected causes

_No response_

## Suggested Improvements

These are ranked in priority based on the findings of common use cases.

* Avoid loading the shipping zone settings on pages that don't need it.
* Add caching of shipping data to the datastore.
* Introduce the ability to prime the cache of a set of shipping zones at once, so each doesn't need to be loaded individually - similar to Order Data Store caching

prettyboymp:

Closes #54913, #56002.

(For Bug Fixes) Bug introduced in PR https://github.com/woocommerce/woocommerce-blocks/pull/7842.

This change prevents the localized settings for the `wc-shipping-method-pickup-location` script from being generated unless the script has actually been enqueued to avoid the overhead of generating the settings for any page except the Shipping -> Local Pickup settings page where it is actually used.

1.  Local Pickup will only load if `CartCheckoutUtils::is_checkout_block_default() === true`.  So first, make sure the site is setup to use the `woocommerdce/checkout` block for checkout. 
  1. Edit the page that is currently set as the checkout page within Woo.
  2. Confirm that the page contains the Checkout block.  This cannot be the legacy checkout shortcode.
2.  Add Shipping Zones vie WooCommerce > Settings > Shipping > Shipping Zones.
4. Visit Shipping Zones vie WooCommerce > Settings > Shipping > Local Pickup - verify that the settings render properly.
5. View the page source and verify that inline script that sets `hydratedScreenSettings` is set.
6. With Query Monitory enabled, visit another page within the admin.  You should no longer see queries agains the shipping zone tables outside of the Shipping Settings page.


plugins/woocommerce/src/Blocks/Shipping/ShippingController.php
```diff
@@ -232,6 +232,11 @@ public function register_settings() {
 	 * Hydrate client settings
 	 */
 	public function hydrate_client_settings() {
+		if ( ! wp_script_is( 'wc-shipping-method-pickup-location', 'enqueued' ) ) {
```
joshuatf: Nitpick, but given how many times we reuse this string `wc-shipping-method-pickup-location` it might be good to set it in a constant.

plugins/woocommerce/src/Blocks/Shipping/ShippingController.php
```diff
@@ -232,6 +232,11 @@ public function register_settings() {
 	 * Hydrate client settings
 	 */
 	public function hydrate_client_settings() {
```
joshuatf, line 240: I'm wondering why this whole function isn't inside of the `PickupLocation` class so that related logic is grouped together.

On that same note, `wp_localize_script` seems more appropriate than adding an inline global variable, but there may be a reason this needs to be done.

joshuatf (commented):
This tests well for me without any regressions!

* ✅ Pickup continues to work as expected
* ✅ Pickup settings are persisted and hydrated correctly on reload
* ✅ Other pages don't have the `hydratedScreenSettings` property and don't run the queries needed for hydrating data
* ❓ Left a question around how we enqueue they settings data, but if there's a reason this is set up this way, I'm happy to approve this PR as is.

Screenshot 2025-03-12 at 4 22 28 PM

plugins/woocommerce/src/Blocks/Shipping/ShippingController.php
prettyboymp: I converted this to a `private const`.

plugins/woocommerce/src/Blocks/Shipping/ShippingController.php
prettyboymp: > I'm wondering why this whole function isn't inside of the PickupLocation class so that related logic is grouped together.

@joshuatf, I started to move some of these methods into the PickupLocation Shipping Method class, but quickly realized that most of the ShippingController class seems to deal specifically with pickup locations.  There is definitely some refactoring that should happen here, but it was going to start being bigger than I wanted to throw into a planned performance improvement.

> On that same note, wp_localize_script seems more appropriate than adding an inline global variable, but there may be a reason this needs to be done.

Since this isn't specifically for localization, the "best practice" is to use `wp_add_inline_script()` - https://developer.wordpress.org/reference/functions/wp_localize_script/#more-information.  This was new to me as well.

joshuatf: >Since this isn't specifically for localization, the "best practice" is to use wp_add_inline_script() - https://developer.wordpress.org/reference/functions/wp_localize_script/#more-information. This was new to me as well.

🤯 I always thought `wp_localize_script` was encouraged to scope the variables.  TIL.

joshuatf (approved):
LGTM!  Re-tested and everything continues to work as expected with the Local Pickup page and the settings are not loaded on other pages.  🚢

PR #56304, opened 2025-03-11
Product Filters: Remove unneeded legacy filters pattern

### LINKED ISSUE #56301: Product Filters: Remove legacy filter patterns
With the new Product Filters releasing soon, there is no need to have a block pattern that inserts the legacy filters anymore. We should remove this.

roykho:

Closes #56301

This PR removes the unneeded legacy filter patterns as the new product filters block will be released soon. There will be no replacement needed for the new filters because the new filters is already in a group when inserted. Note that I also had to update the logic for registering the block patterns because we're caching the patterns, if the pattern file is removed, PHP will complain So I added a logic to check if the source path exists before trying to use it.

1. Go to the Appearance -> Editor -> Patterns -> WooCommerce
2. Ensure you no longer see a pattern for Product Filters in the list and searching for "filters" should have no results.
3. Ensure you don't see any PHP warnings about path not found or to that effect.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Blocks/BlockPatterns.php
```diff
@@ -114,7 +114,9 @@ public function register_block_patterns() {
 
 		$patterns = $this->get_block_patterns();
 		foreach ( $patterns as $pattern ) {
-			$this->pattern_registry->register_block_pattern( $pattern['source'], $pattern, $this->get_patterns_dictionary() );
+			if ( file_exists( $pattern['source'] ) ) {
```
Aljullu: I was exploring if we could get rid of this check. In theory, this shouldn't be needed because `get_pattern_cache()` already checks if the cache matches the current WC version; otherwise it invalidates it. However, that doesn't cover dev sites running `trunk`, so I don't think there is any way to remove it. (One alternative would be to disable the caching in dev sites, but I don't think it adds any value)

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/inspector-controls.block_theme.spec.ts
```diff
@@ -553,10 +553,12 @@ test.describe( 'Product Collection: Inspector Controls', () => {
...
 
-			await page.getByLabel( 'Toggle block inserter' ).click();
-			await page.getByRole( 'tab', { name: 'Patterns' } ).click();
+			const addBlockButton = editor.canvas
+				.locator( '[data-type="woocommerce/product-collection"]' )
+				.getByRole( 'button', { name: 'Add block' } );
+			await addBlockButton.click();
 			await page.getByPlaceholder( 'Search' ).fill( 'product filters' );
-			await page.getByLabel( 'Product Filters' ).click();
+			await page.getByText( 'Product Filters' ).click();
```
Aljullu: Not a big deal, but I think this can be replaced with Gutenberg utils, which I would expect to be a bit more performant and less brittle to break in future updates:

```TS
			const productCollectionBlock = await editor.getBlockByName(
				'woocommerce/product-collection'
			);
			const productCollectionClientId =
				( await productCollectionBlock
					.last()
					.getAttribute( 'data-block' ) ) ?? '';
			await editor.insertBlock(
				{ name: 'woocommerce/product-filters' },
				{ clientId: productCollectionClientId }
			);
```

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/inspector-controls.block_theme.spec.ts
```diff
@@ -602,10 +604,12 @@ test.describe( 'Product Collection: Inspector Controls', () => {
...
```
Aljullu: Same here. :slightly_smiling_face:

Aljullu (approved):
This is testing as described, and the code looks good.

imatge

I left one small suggestion for tests, but leaving up to you if you think is worth it. I also left an inline comment, but nothing to change there, I was just thinking out loud. :smile:

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/inspector-controls.block_theme.spec.ts
roykho: Good idea, thanks for the suggestion. I pushed up changes.

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/inspector-controls.block_theme.spec.ts
roykho: Good idea, thanks for the suggestion. I pushed up changes.

plugins/woocommerce/src/Blocks/BlockPatterns.php
roykho: @Aljullu thanks for pointing this out. I thought about it some more and decided to remove the `file_exists` check because that diverges from the cached data and could end up with bugs later on.

I will create follow up issue to see how we should tackle cached data and dev environments. I do know in the past (many years ago) we always used `WP_DEBUG` in our development but I think we might have better ways now thus this needs to be a separate issue.

Aljullu: Sounds good, thanks!

PR #56323, opened 2025-03-12
Product Filters: refactor the paginated filter results handling

dinhtungdu:

- I have checked to ensure there aren't other open [Pull Requests](https://github.com/woocommerce/woocommerce/pulls) for the same update/change.
- Following the above guidelines will result in quick merges and clear and detailed feedback when appropriate.

This PR refactors how  paginated filter result is handled in the Product Filters block. Instead removing the page argument from the URL on the client side, we send the canonical URL from the server and use it to process new filter URL.

Follow up https://github.com/woocommerce/woocommerce/pull/55080.

Not applicable as this is a behavioral/architectural change without UI modifications.

1. Create a product archive page (shop or category) with the Product Filters block
2. Add some products and apply filters to create pagination (e.g., enough products to span multiple pages)
3. Navigate to page 2 or 3 of the filtered results
4. Apply different filters and verify that:
   - no 404 error.
   - the URL updates correctly without retaining old pagination parameters. 
   - the product results are updated correctly.

#### Note

You can use the following snipet to limit the number of product per page for the Shop page:

```php
add_filter(
    'loop_shop_per_page', function () {
        return 3;
    }
);
```

The changes have been tested in the following environments:

- WordPress 6.7.2
- Block theme (Twenty Twenty-Five)
- Both pretty and plain permalink structures

The testing focused on ensuring proper URL handling and navigation for paginated filtered results.


plugins/woocommerce/src/Blocks/BlockTypes/ProductFilters.php
```diff
@@ -36,6 +36,12 @@ protected function enqueue_data( array $attributes = array() ) {
 		$this->asset_data_registry->add( 'isProductArchive', is_shop() || is_product_taxonomy() );
 		$this->asset_data_registry->add( 'isSiteEditor', 'site-editor.php' === $pagenow );
 		$this->asset_data_registry->add( 'isWidgetEditor', 'widgets.php' === $pagenow || 'customize.php' === $pagenow );
+
+		$canonical_url_no_pagination = get_pagenum_link( 1 );
```
roykho, line 40:
```suggestion
		$canonical_url_no_pagination = get_pagenum_link();
```

Nitpick: Page number here is not really needed as it already defaults to 1 and is optional.

roykho (approved):
Great work!

Just added a small nitpick suggestion.

✅ Tested with block theme and non block theme to ensure both navigation works
✅ Tested with pretty permalink and plain permalink

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilters.php
dinhtungdu: I see. But in this case, I think explicitly passing `1` give us a clearer explaination that we're getting the URL fro the first page, without have to look at the definition of the `get_pagenum_link` function.

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilters.php
```diff
@@ -36,6 +36,12 @@ protected function enqueue_data( array $attributes = array() ) {
...
+		if ( is_singular() ) {
+			$canonical_url_no_pagination = get_permalink();
+		}
+		$this->asset_data_registry->add( 'canonicalUrl', html_entity_decode( $canonical_url_no_pagination ) );
```
luisherranz, line 44: It'd be great to migrate this to `wp_interactivity_config` at some point.

PR #56327, opened 2025-03-12
Show wccom connection url mismatch notification

thilinah:

This PR shows a notification when the URL of the WooCommerce.com connection and the site is out of sync.

Screenshot 2025-03-12 at 17 57 38

#### How URLs can go out of sync

The URL of the WooCommerce.com connection and the site can go out of sync mainly due to two reasons.

1. Changing the site home URL -  e.g:  The site `testsite.com` is connected to WooComemrce.com and the later the HOME_URL of the site is updated to `www.testsite.com`.

2. Replicating a production DB to create a staging site or another new site - It's a common practise among developers to copy the production DB from an existing site to a staging site, for the purpose of testing or building a new store. Since the DB will contain the authentication details of the WooCommerce.com connection of the original site, the new site will also share the same connection.

#### What are the implications

- If the site URL has been changed, as point (1), it will only cause issues if the site purchase a host plan. Since the host plan will be linked to the connection for current URL `www.testsite.com` instead of `testsite.com`. 

- In case of point (2), the user deactivate a plugin from the staging site, the product connection on the original site will be disconnected.

#### The solution implemented on this PR

- Show a notification when site urls are out of sync.
- Show CTA to Reconnect the site.

The notification will be shown only when the original connection doesn't have any linked host plan orders. The reason is:
- The number of connected sites with host-plan orders small compared to number of connections without host-plan orders.
- We still don't have reliable way to migrate host-plan order on connection changes.

Closes # .

| Before | After |
| ------ | ----- |
|        |       |

1. Create the build from this PR and install it on customer store.
2. If the PR (https://github.com/Automattic/woocommerce.com/pull/23609) is merged, use WCCOM trunk to test. If not use the branch of the PR.
3. Under in-app subscriptions (`/wp-admin/admin.php?page=wc-admin&tab=my-subscriptions&path=%2Fextensions`), refresh the subscription list.
4. This will save the transient `_woocommerce_helper_connection_data`. To test the notification we can modify the url of the store and set `alert_url_mismatch` to `true`. 
5. Open a wp shell on the customer store ( if you are using wp-env use the command `pnpm -- wp-env run cli wp shell`) and use the following code to modify the transient.
```
$data = get_transient('_woocommerce_helper_connection_data');

$data['url'] = 'http://testsiteurl.test/';
$data['alert_url_mismatch'] = true;

set_transient( '_woocommerce_helper_connection_data', $data, 1 * HOUR_IN_SECONDS );
``` 
6. Refresh the in-app subscription list to see the URL mismatch notice.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

Show a notification when there is a mismatch in WooCommerce.com connection URL

</details>

<details>

#### Comment 

</details>


thilinah:
When reconnecting a site we should consider migrating connected products in WooCommerce.com.

plugins/woocommerce/client/admin/client/marketplace/utils/functions.tsx
```diff
@@ -491,19 +491,25 @@ const subscribeUrl = ( subscription: Subscription ): string => {
 
 // If you need to add support for a different page, make sure to
 // update WC_Helper::get_source_page() in the backend.
-const connectUrl = ( page = 'wc-admin' ): string => {
+const connectUrl = ( page = 'wc-admin', reconnect = false ): string => {
 	const wccomSettings = getAdminSetting( 'wccomHelper', {} );
 
-	if ( ! wccomSettings.connectURL ) {
+	if ( ! reconnect && ! wccomSettings.connectURL ) {
```
anaspk, line 497: Do we really need to change the conditions here? I think just `if ( ! wccomSettings.connectURL ) {` is still enough as the value of `reconnect` doesn't really matter (regardless of whether it is `true` or `false`, we do the same thing.)

anaspk (approved):
Thank you @thilinah , LGTM and works as described 👍 

I have noticed one weird thing tho. Not sure if it is related to this PR or not.

If I click the Reconnect button in the notice, but then abort the connection process midway and comeback to My subscriptions page, I see this connection success message in addition to the notice, even though no new connection was created.

Screenshot 2025-03-28 at 6 07 02 PM

plugins/woocommerce/client/admin/client/marketplace/utils/functions.tsx
thilinah: On the next if condition we check the variable `wccomSettings.reConnectURL` (not `wccomSettings.connectURL`)

Issue #55729, opened 2025-02-21
[Product Details Block] Product reviews block


louwie17:
**Description**

As part of the blockifying process of product details we need to create a block to display the product reviews. As one of the tabs ( now accordions ) will be a reviews accordion.

WordPress already has a Comments block, which we could potentially use or should at-least mimic: https://wordpress.org/documentation/article/comments-block/

Product Reviews can also be extended, and some extensions may add additional questions to the form, we should look into this.
When testing for this https://github.com/woocommerce/woocommerce/discussions/55224 I also kept track of some plugins that modify or replace the reviews section:

### List of plugins with notes

Name | Product Details Block
-- | --
Advanced Reviews By Aheadworks | TBD - Adds additional questions - advantage/disadvantage
Sparks for WooCommerce | Adds additional questions to the reviews block
WooCommerce Product Reviews Pro | Replaces review block
WooCommerce Tab Manager | Allows hiding/showing

### Hooks to support with compatibility layer

See this file: https://github.com/woocommerce/woocommerce/blob/8ee74ede0ebc1957089af08207ae2a1a8250610e/plugins/woocommerce/templates/single-product/review.php

- `woocommerce_review_before`
- `woocommerce_review_before_comment_meta`
- `woocommerce_review_meta`
- `woocommerce_review_before_comment_text` - Part of review content block
- `woocommerce_review_comment_text` - Part of review content block
- `Woocommerce_review_after_comment_text` - Part of review content block

**Acceptance criteria**

- [x] Research if we can make use of the [ comments block](https://github.com/WordPress/gutenberg/tree/trunk/packages/block-library/src/comments)
- [x] Update the product details block default template to use the new "Product Reviews" block
- [x] Assess whether this could be a variation of the Core Comments block or a standalone block.  
- [ ] Conduct extensibility investigation.
- [x] Main Product Reviews block
- [x] Product reviews title block [#56332](https://github.com/woocommerce/woocommerce/pull/56332)
- [x] Product review rating block [#56332](https://github.com/woocommerce/woocommerce/pull/56332)
- [ ] Product review form block ( in review )
- [ ] Product review template block ( a variation of the comments template )
- [ ] Review Author block ( a variation of the comment author block )
- [ ] This may require a custom block because of the verified owner label )
- [ ] Review Date block ( a variation of the comment date block )
- [ ] Review Content block ( a variation of the comment content block )
- [ ] Review pagination block ( a variation of the comments variations block )

Issue #51648, opened 2024-09-24
Migrate to `get_block_wrapper_attributes` function


gigitux:
Currently, we are using our own dedicated implementation to add wrapper attributes to the main wrapper of blocks. This approach has the following drawbacks:

- For the style, we are using [`StyleAttributeUtils`](https://github.com/woocommerce/woocommerce/blob/bca89cb14fa7f52c8d04680b73661d57deea5f07/plugins/woocommerce/src/Blocks/Utils/StyleAttributesUtils.php#L7). This class relies on custom logic to generate the right classes and styles. If those change upstream, we may not notice it and break the style experience.
- For the `wp-block-${block-name}` class, we add it manually for each block. Occasionally, we could forget to add it and [this cause issues with core features](https://github.com/woocommerce/woocommerce/issues/49739).

@louwie17 shared the [`get_block_wrapper_attributes` function ](https://developer.wordpress.org/reference/functions/get_block_wrapper_attributes/). This function returns all the attributes (including class names and styles) to attach to the main wrapper of blocks.

By using the core `get_block_wrapper_attributes` function, we can minimize the amount of custom code we need to maintain. Also, relying on the core implementation ensures that our wrapper attributes are always up-to-date with the latest upstream changes, reducing the risk of style and class-related issues.

ralucaStan:
@gigitux this PR does not have any reviewers? Is there anything blocking it?

bhavz-10:
Hello @gigitux 

I would like to work on this.

I checked out the Cart Link Block and tested it for a case with all the attributes being assigned. Currently, it uses. https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Blocks/BlockTypes/CartLink.php#L28-L45. If we just use the get `get_block_wrapper_attributes` without making use of `StyleAttributesUtils::get_classes_and_styles_by_attributes`, it gives the correct result since the block has support for the available attributes. 

Approach:

 - We would need to make a list of the blocks which need to migrate
   first and have less complications compared to the ones that are more
   complex. 
 - Then work on migrating them to use [get_block_wrapper_attributes function](https://developer.wordpress.org/reference/functions/get_block_wrapper_attributes/) standalone without the usage of [StyleAttributeUtils](https://github.com/woocommerce/woocommerce/blob/bca89cb14fa7f52c8d04680b73661d57deea5f07/plugins/woocommerce/src/Blocks/Utils/StyleAttributesUtils.php#L7)

[...]

PR #56332, opened 2025-03-12
Add product review title and rating blocks

louwie17:

This PR builds on top of the initial product reviews PR ( #56276 ) and adds two new blocks:
- **Product Review Rating** - A block to display the **star** rating of an individual review.
- **Product Reviews Title** - A block to replace the **Post Comments Title**, which updates the wording to use reviews, but also makes use of the WC Reviews REST endpoint within context aware blocks.

Part of https://github.com/woocommerce/woocommerce/issues/55729

Closes WOOPLUG-3849

Screenshot 2025-03-13 at 1 44 09 PM

https://github.com/user-attachments/assets/54af7715-51aa-4262-aaeb-fc781981b59a

**Note:** These tests will be focused towards the two new blocks, the reviews title and the review rating. Meaning that the review block still contains other blocks that pertain to comments specifically. These will be addressed in follow up PRs.

1. Load this branch and have the **WooCommerce Beta Tester** plugin activated
2. Go to **Tools > WCA Test Helper > Features** and enable the **experimental blocks** feature flag
3. Go to **Appearance > Editor > Templates** and edit the **Single Product** template
4. Add the **Blockified Product Details** block ( an accordion )
5. The third accordion ( Reviews ) should contain the **Blockified Product Reviews** block as its child
6. It should include the **Product Reviews Title** as the first nested block that uses the "product reviews" copy and allows you toggle the review count and product title ( this should also be reflected correctly on the front-end ).
7. The comment template will have a third column that contains the product rating ( the 5 stars )
8. Save the template
9. Visit a product with and without reviews, if reviews exist, it should correctly render the title with the correct number and show the correct star rating for each review, if no reviews exist, it shouldn't render any review and render "0 reviews for ..." as the title. ( We may want to create a follow up to have an explicit "No review" title block, as in the current block it displays "There are no reviews yet." and "Be the first to review “Black and White”" )

**Test areas with context**

Follow the tests below either within a page where you use either the **Single product** or **product collection** block.

1. Add the **Blockified Product Reviews** block as a child of one of the above blocks ( make sure you have a product with and without reviews displayed )
2. If the parent block has reviews it should show the correct number in the **Product Reviews Title** block. Ratings will unfortunately not be shown yet, because we still make use of the **Comments Template** which uses the comments endpoint.
3. If the parent block does not have reviews, it should show the correct number of zero reviews in the title.
4. Save the page and preview it, it should render the title correctly and also the reviews that include the star rating.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/packages/js/data/src/reviews/resolvers.ts
+++ b/packages/js/data/src/reviews/resolvers.ts
@@ -7,7 +7,7 @@ import { addQueryArgs } from '@wordpress/url';
  * Internal dependencies
  */
 import { NAMESPACE } from '../constants';
-import { setError, updateReviews } from './actions';
+import { setError, setReview, updateReviews } from './actions';
 import { fetchWithHeaders } from '../controls';
 import { ReviewObject, ReviewsQueryParams } from './types';
 
@@ -33,6 +33,23 @@ export function* getReviews( query: ReviewsQueryParams ) {
 		yield updateReviews( query, response.data, totalCount );
 	} catch ( error ) {
 		yield setError( JSON.stringify( query ), error );
+	}
+}
+
+export function* getReview( id: number ) {
+	try {
+		const url = addQueryArgs( `${ NAMESPACE }/products/reviews/${ id }` );
+		const response: {
+			headers: Map< string, string >;
+			data: ReviewObject;
+		} = yield fetchWithHeaders( {
+			path: url,
+			method: 'GET',
+		} );
+
+		yield setReview( id, response.data );
+	} catch ( error ) {
+		yield setError( JSON.stringify( id ), error );
 	}
 }
 
--- a/packages/js/data/src/reviews/resolvers.ts
+++ b/packages/js/data/src/reviews/resolvers.ts
@@ -38,7 +38,7 @@ export function* getReviews( query: ReviewsQueryParams ) {
 
 export function* getReview( id: number ) {
 	try {
-		const url = addQueryArgs( `${ NAMESPACE }/products/reviews/${ id }` );
+		const url = addQueryArgs( `wc/v3/products/reviews/${ id }` );
 		const response: {
 			headers: Map< string, string >;
 			data: ReviewObject;
--- a/plugins/woocommerce/client/blocks/assets/css/style.scss
+++ b/plugins/woocommerce/client/blocks/assets/css/style.scss
@@ -99,10 +99,11 @@ body.wc-block-product-gallery-modal-open {
 	}
 }
 
-.wc-block-grid__product-rating {
+.wc-block-grid__product-rating, .wp-block-woocommerce-product-review-rating {
 	display: block;
 
 	.wc-block-grid__product-rating__stars,
+	.wc-block-components-product-review-rating__stars,
 	.star-rating {
 		overflow: hidden;
 		position: relative;
--- a/plugins/woocommerce/client/blocks/assets/css/style.scss
+++ b/plugins/woocommerce/client/blocks/assets/css/style.scss
@@ -33,7 +33,9 @@ body.wc-block-product-gallery-modal-open {
 		}
 	}
 }
-.edit-post-visual-editor .editor-block-list__block .wc-block-grid__product-title,
+.edit-post-visual-editor
+	.editor-block-list__block
+	.wc-block-grid__product-title,
 .editor-styles-wrapper .wc-block-grid__product-title,
 .wc-block-grid__product-title {
 	font-family: inherit;
@@ -93,7 +95,8 @@ body.wc-block-product-gallery-modal-open {
 .has-7-columns,
 .has-8-columns,
 .has-9-columns {
-	.wc-block-grid__product-add-to-cart.wp-block-button .wp-block-button__link::after {
+	.wc-block-grid__product-add-to-cart.wp-block-button
+		.wp-block-button__link::after {
 		content: "";
 		margin: 0;
 	}
@@ -104,7 +107,7 @@ body.wc-block-product-gallery-modal-open {
 	display: block;
 
 	.wc-block-grid__product-rating__stars,
-	.wc-block-components-product-review-rating__stars,
+	.wc-block-product-review-rating__stars,
 	.star-rating {
 		overflow: hidden;
 		position: relative;
@@ -207,7 +210,9 @@ body.wc-block-product-gallery-modal-open {
 	.wc-block-grid__product-onsale,
 	.wc-block-components-product-title,
 	.wc-block-components-product-sale-badge {
-		font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen, Ubuntu, Cantarell, "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
+		font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto,
+			Oxygen, Ubuntu, Cantarell, "Fira Sans", "Droid Sans",
+			"Helvetica Neue", sans-serif;
 	}
 	.wc-block-grid__product-title::before {
 		display: none;
@@ -216,13 +221,16 @@ body.wc-block-product-gallery-modal-open {
 	.wc-block-components-product-sale-badge {
 		line-height: 1;
 	}
-	.editor-styles-wrapper .wp-block-button .wp-block-button__link:not(.has-text-color) {
+	.editor-styles-wrapper
+		.wp-block-button
+		.wp-block-button__link:not(.has-text-color) {
 		color: #fff;
 	}
 }
 
 .theme-twentytwenty {
-	$twentytwenty-headings: -apple-system, blinkmacsystemfont, "Helvetica Neue", helvetica, sans-serif;
+	$twentytwenty-headings: -apple-system, blinkmacsystemfont, "Helvetica Neue",
+		helvetica, sans-serif;
 	$twentytwenty-highlights-color: #cd2653;
 
 	.wc-block-grid__product-link {
@@ -288,12 +296,17 @@ body.wc-block-product-gallery-modal-open {
 	.wc-block-grid__products .wc-block-components-product-sale-badge {
 		position: static;
 	}
-	.wc-block-grid__products .wc-block-grid__product-image .wc-block-components-product-sale-badge {
+	.wc-block-grid__products
+		.wc-block-grid__product-image
+		.wc-block-components-product-sale-badge {
 		position: absolute;
 	}
 
 	// These styles are not applied to the All Products atomic block, so it can be positioned normally.
-	.wc-block-grid__products .wc-block-grid__product-onsale:not(.wc-block-components-product-sale-badge) {
+	.wc-block-grid__products
+		.wc-block-grid__product-onsale:not(
+			.wc-block-components-product-sale-badge
+		) {
 		position: absolute;
 		right: 4px;
 		top: 4px;
--- a/plugins/woocommerce/client/blocks/assets/css/style.scss
+++ b/plugins/woocommerce/client/blocks/assets/css/style.scss
@@ -99,7 +99,8 @@ body.wc-block-product-gallery-modal-open {
 	}
 }
 
-.wc-block-grid__product-rating, .wp-block-woocommerce-product-review-rating {
+.wc-block-grid__product-rating,
+.wp-block-woocommerce-product-review-rating {
 	display: block;
 
 	.wc-block-grid__product-rating__stars,
--- a/plugins/woocommerce/client/blocks/assets/css/style.scss
+++ b/plugins/woocommerce/client/blocks/assets/css/style.scss
@@ -33,9 +33,7 @@ body.wc-block-product-gallery-modal-open {
 		}
 	}
 }
-.edit-post-visual-editor
-	.editor-block-list__block
-	.wc-block-grid__product-title,
+.edit-post-visual-editor .editor-block-list__block .wc-block-grid__product-title,
 .editor-styles-wrapper .wc-block-grid__product-title,
 .wc-block-grid__product-title {
 	font-family: inherit;
@@ -95,8 +93,7 @@ body.wc-block-product-gallery-modal-open {
 .has-7-columns,
 .has-8-columns,
 .has-9-columns {
-	.wc-block-grid__product-add-to-cart.wp-block-button
-		.wp-block-button__link::after {
+	.wc-block-grid__product-add-to-cart.wp-block-button .wp-block-button__link::after {
 		content: "";
 		margin: 0;
 	}
@@ -210,9 +207,7 @@ body.wc-block-product-gallery-modal-open {
 	.wc-block-grid__product-onsale,
 	.wc-block-components-product-title,
 	.wc-block-components-product-sale-badge {
-		font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto,
-			Oxygen, Ubuntu, Cantarell, "Fira Sans", "Droid Sans",
-			"Helvetica Neue", sans-serif;
+		font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen, Ubuntu, Cantarell, "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
 	}
 	.wc-block-grid__product-title::before {
 		display: none;
@@ -221,16 +216,13 @@ body.wc-block-product-gallery-modal-open {
 	.wc-block-components-product-sale-badge {
 		line-height: 1;
 	}
-	.editor-styles-wrapper
-		.wp-block-button
-		.wp-block-button__link:not(.has-text-color) {
+	.editor-styles-wrapper .wp-block-button .wp-block-button__link:not(.has-text-color) {
 		color: #fff;
 	}
 }
 
 .theme-twentytwenty {
-	$twentytwenty-headings: -apple-system, blinkmacsystemfont, "Helvetica Neue",
-		helvetica, sans-serif;
+	$twentytwenty-headings: -apple-system, blinkmacsystemfont, "Helvetica Neue", helvetica, sans-serif;
 	$twentytwenty-highlights-color: #cd2653;
 
 	.wc-block-grid__product-link {
@@ -296,17 +288,12 @@ body.wc-block-product-gallery-modal-open {
 	.wc-block-grid__products .wc-block-components-product-sale-badge {
 		position: static;
 	}
-	.wc-block-grid__products
-		.wc-block-grid__product-image
-		.wc-block-components-product-sale-badge {
+	.wc-block-grid__products .wc-block-grid__product-image .wc-block-components-product-sale-badge {
 		position: absolute;
 	}
 
 	// These styles are not applied to the All Products atomic block, so it can be positioned normally.
-	.wc-block-grid__products
-		.wc-block-grid__product-onsale:not(
-			.wc-block-components-product-sale-badge
-		) {
+	.wc-block-grid__products .wc-block-grid__product-onsale:not(.wc-block-components-product-sale-badge) {
 		position: absolute;
 		right: 4px;
 		top: 4px;
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/block.json
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/block.json
@@ -5,7 +5,7 @@
 	"title": "Product Review Rating",
 	"category": "woocommerce",
 	"description": "Displays the rating of a product review.",
-	"textdomain": "default",
+	"textdomain": "woocommerce",
 	"usesContext": [ "commentId" ],
 	"attributes": {
 		"textAlign": {
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/block.json
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/block.json
@@ -1,0 +1,24 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/block.json",
+	"apiVersion": 3,
+	"name": "woocommerce/product-review-rating",
+	"title": "Product Review Rating",
+	"category": "woocommerce",
+	"description": "Displays the rating of a product review.",
+	"textdomain": "default",
+	"usesContext": [ "commentId" ],
+	"attributes": {
+		"textAlign": {
+			"type": "string"
+		}
+	},
+	"supports": {
+		"color": {
+			"gradients": true,
+			"__experimentalDefaultControls": {
+				"background": true,
+				"text": true
+			}
+		}
+	}
+}
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/block.json
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/block.json
@@ -4,6 +4,7 @@
 	"name": "woocommerce/product-review-rating",
 	"title": "Product Review Rating",
 	"category": "woocommerce",
+	"ancestor": [ "woocommerce/blockified-product-reviews" ],
 	"description": "Displays the rating of a product review.",
 	"textdomain": "woocommerce",
 	"usesContext": [ "commentId" ],
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/edit.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/edit.tsx
@@ -1,0 +1,94 @@
+/**
+ * External dependencies
+ */
+import clsx from 'clsx';
+import { __, sprintf } from '@wordpress/i18n';
+import { useEntityProp } from '@wordpress/core-data';
+import {
+	AlignmentToolbar,
+	BlockControls,
+	useBlockProps,
+} from '@wordpress/block-editor';
+import { useStyleProps } from '@woocommerce/base-hooks';
+import type { BlockEditProps } from '@wordpress/blocks';
+
+/**
+ * Renders the `woocommerce/product-review-rating` block on the editor.
+ *
+ * @param   {Object}  props                    React props.
+ * @param   {Object}  props.attributes         Block attributes.
+ * @param   {Object}  props.setAttributes      Function to set block attributes.
+ * @param   {Object}  props.context           Inherited context.
+ * @param   {string}  props.context.commentId The comment ID.
+ *
+ * @return {JSX.Element} React element.
+ */
+export default function Edit( {
+	context: { commentId },
+	attributes,
+	setAttributes,
+}: BlockEditProps< {
+	textAlign: string;
+} > & {
+	context: { commentId: string };
+} ) {
+	const { textAlign } = attributes;
+	const styleProps = useStyleProps( attributes );
+	const className = clsx(
+		styleProps.className,
+		'wc-block-components-product-review-rating',
+		{
+			[ `has-text-align-${ textAlign }` ]: textAlign,
+		}
+	);
+	const blockProps = useBlockProps( {
+		className,
+	} );
+	let [ rating ] = useEntityProp( 'root', 'comment', 'rating', commentId );
+	rating = rating ?? 4;
+
+	const starStyle = {
+		width: ( rating / 5 ) * 100 + '%',
+	};
+
+	const ratingText = sprintf(
+		/* translators: %f is referring to the average rating value */
+		__( 'Rated %f out of 5', 'woocommerce' ),
+		rating
+	);
+
+	const ratingHTML = {
+		__html: sprintf(
+			/* translators: %s is the rating value wrapped in HTML strong tags. */
+			__( 'Rated %s out of 5', 'woocommerce' ),
+			sprintf( '<strong class="rating">%f</strong>', rating )
+		),
+	};
+
+	return (
+		<>
+			<BlockControls>
+				<AlignmentToolbar
+					value={ attributes.textAlign }
+					onChange={ ( newAlign ) => {
+						setAttributes( { textAlign: newAlign || '' } );
+					} }
+				/>
+			</BlockControls>
+			<div { ...blockProps }>
+				<div
+					className={
+						'wc-block-components-product-review-rating__stars'
+					}
+					role="img"
+					aria-label={ ratingText }
+				>
+					<span
+						style={ starStyle }
+						dangerouslySetInnerHTML={ ratingHTML }
+					/>
+				</div>
+			</div>
+		</>
+	);
+}
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/edit.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/edit.tsx
@@ -9,20 +9,8 @@ import {
 	BlockControls,
 	useBlockProps,
 } from '@wordpress/block-editor';
-import { useStyleProps } from '@woocommerce/base-hooks';
 import type { BlockEditProps } from '@wordpress/blocks';
 
-/**
- * Renders the `woocommerce/product-review-rating` block on the editor.
- *
- * @param   {Object}  props                    React props.
- * @param   {Object}  props.attributes         Block attributes.
- * @param   {Object}  props.setAttributes      Function to set block attributes.
- * @param   {Object}  props.context           Inherited context.
- * @param   {string}  props.context.commentId The comment ID.
- *
- * @return {JSX.Element} React element.
- */
 export default function Edit( {
 	context: { commentId },
 	attributes,
@@ -33,14 +21,9 @@ export default function Edit( {
 	context: { commentId: string };
 } ) {
 	const { textAlign } = attributes;
-	const styleProps = useStyleProps( attributes );
-	const className = clsx(
-		styleProps.className,
-		'wc-block-components-product-review-rating',
-		{
-			[ `has-text-align-${ textAlign }` ]: textAlign,
-		}
-	);
+	const className = clsx( 'wc-block-product-review-rating', {
+		[ `has-text-align-${ textAlign }` ]: textAlign,
+	} );
 	const blockProps = useBlockProps( {
 		className,
 	} );
@@ -77,9 +60,7 @@ export default function Edit( {
 			</BlockControls>
 			<div { ...blockProps }>
 				<div
-					className={
-						'wc-block-components-product-review-rating__stars'
-					}
+					className="wc-block-product-review-rating__stars"
 					role="img"
 					aria-label={ ratingText }
 				>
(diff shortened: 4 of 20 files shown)


plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/block.json
```diff
@@ -0,0 +1,24 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/block.json",
+	"apiVersion": 3,
+	"name": "woocommerce/product-review-rating",
+	"title": "Product Review Rating",
+	"category": "woocommerce",
+	"description": "Displays the rating of a product review.",
+	"textdomain": "woocommerce",
+	"usesContext": [ "commentId" ],
```
louwie17, line 10: Sticking with `commentId` given we will keep the comments template component that exposes the comment id.

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/reviews-title/edit.tsx
```diff
@@ -0,0 +1,170 @@
...
+		} ),
+	} );
+
+	useEffect( () => {
+		if ( isSiteEditor ) {
+			setReviewsCount( 3 );
+			return;
+		}
+		resolveSelect( reviewsStore )
+			.getReviewsTotalCount( {
```
louwie17, line 114: Had to use the `reviewsStore` for the data as the WP comments API doesn't work for this.
This is the main reason I created a new block for this.

plugins/woocommerce/src/Blocks/BlockTypes/Reviews/ProductReviewsTitle.php
```diff
@@ -0,0 +1,98 @@
...
+	 */
+	protected $block_name = 'product-reviews-title';
+
+	/**
+	 * Render the block.
+	 *
+	 * @param array $attributes Block attributes.
+	 * @return string Rendered block content.
+	 */
+	protected function render( $attributes, $content, $block ) {
```
louwie17, line 85: This is based off of the comments-title render function, just making use of the actual `$product` entity.

imanish003 (commented):
Hey @louwie17, I've completed functional testing of the PR and identified a few minor issues that I'll detail below. Thanks for your work on this!

- Verified that third accordion contains the Blockified Product Reviews block as its child
- Verified that the Product Reviews Title block is present and displays text `3 reviews for "Product Title"`
- Toggles "Show Reviews Count" and "Show Product Title" correctly update the title in Editor side as well as on the front-end

> We may want to create a follow up to have an explicit "No review" title block

Makes sense. In my testing, when there are no reviews:
- And if Product Title is hidden, it shows "0 reviews" on Frontend
- When Reviews Count is hidden, it shows `Reviews for “Polo”` on Frontend

Which I agree, isn't best UX. 

> as in the current block it displays "There are no reviews yet." and "Be the first to review “Black and White”"

This doesn't seem to be the case for me, Am I missing something? Here is a screenshot of what I am seeing on Frontend:

image

#### Minor: Star rating is cut off

As you can see in the screenshot below, the stars are cut off or truncated. Are you also seeing this?
image

#### Minor: Accordions are appearing and disappearing on Frontend

I don't think it is directly related to this PR but I noticed on front end that at first all the accordions are appearing and after some time only one accordion remains which is description accordion. I was able to reproduce it for an album product but when I tried to reproduce it for a cap product, I was not able to reproduce it. So I am not sure if there is something wrong with my local setup, but in any case, I thought it's worth mentioning.

https://github.com/user-attachments/assets/ca8a109e-ba01-47de-8517-7574d3942484

### Unexpected behavior when adding this block to a product collection block

The video below demonstrates an unexpected behavior when adding this block to a product collection block. In the editor, when I add the block to a product collection block, only one product displays on the frontend if the block is part of the product template. Maybe there is a bug in Product Collection block? Are you able to reproduce this?

Sidenote: This seems to be working fine inside Single Product block.

https://github.com/user-attachments/assets/37ce9bb2-b551-43f0-a147-b7a3c2e14f43

louwie17:
Thanks for the review @imanish003 :) 

> This doesn't seem to be the case for me, Am I missing something? Here is a screenshot of what I am seeing on Frontend:

From the image it looks like your looking at the new block, when mentioning the `current block` here "current block it displays "There are no reviews yet." and "Be the first to review “Black and White”"" I was referring to the current non-blockified reviews version:
Screenshot 2025-03-19 at 09 20 57

> #### Minor: Star rating is cut off
> As you can see in the screenshot below, the stars are cut off or truncated. Are you also seeing this?

I fixed this as part of this commit: https://github.com/woocommerce/woocommerce/pull/56332/commits/7164e9988063801404c9bb58ae0c4093dc341800, as it was because of the fixed width I set the third column to, which wasn't big enough for the rating.
In hindsight, turning it into three columns is not the right approach, what I did instead is I moved the rating and the comment author block into a `row` block with justification set to "Space inbetween items". This way the star review has plenty of room and it should also scale better. 

> #### Minor: Accordions are appearing and disappearing on Frontend
> I don't think it is directly related to this PR but I noticed on front end that at first all the accordions are appearing and after some time only one accordion remains which is description accordion. I was able to reproduce it for an album product but when I tried to reproduce it for a cap product, I was not able to reproduce it. So I am not sure if there is something wrong with my local setup, but in any case, I thought it's worth mentioning.

I cannot reproduce this! What theme are you using? This kinda smells like a styling issue with the theme?

> ### Unexpected behavior when adding this block to a product collection block
> The video below demonstrates an unexpected behavior when adding this block to a product collection block. In the editor, when I add the block to a product collection block, only one product displays on the frontend if the block is part of the product template. Maybe there is a bug in Product Collection block? Are you able to reproduce this?

Interesting!! So I did some further digging and this turns out to be an issue with the product collection block I believe in particular when using a comment form.
If I create the same thing but in a query loop, it renders fine!
It also is not strictly related to the **Blockified product reviews** block, and just adding the **Comments form** block only within the product collection also triggers the issue.

I don't know if the product collection block does some parsing of the HTML? but I have a suspicion that somewhere some HTML is being modified and a closing tag is being removed. Which in turn causes some of the nested HTML to disappear and the interactivity API not being able to find some elements. 


https://github.com/user-attachments/assets/f95eb215-7ec5-478f-8310-56f4e86410ad

louwie17:
@imanish003 this should be ready for a re-review given the product collection issue is not related to this PR, and I addressed the rating overlap issue.

imanish003:
> From the image it looks like your looking at the new block, when mentioning the current block here "current block it displays "There are no reviews yet." and "Be the first to review “Black and White”"" I was referring to the current non-blockified reviews version:

Ah! My bad 🤦‍♂️

> I fixed this as part of this commit: https://github.com/woocommerce/woocommerce/commit/7164e9988063801404c9bb58ae0c4093dc341800, as it was because of the fixed width I set the third column to, which wasn't big enough for the rating.
> In hindsight, turning it into three columns is not the right approach, what I did instead is I moved the rating and the comment author block into a row block with justification set to "Space inbetween items". This way the star review has plenty of room and it should also scale better.

Great! It's working as expected now.

> I cannot reproduce this! What theme are you using? This kinda smells like a styling issue with the theme?

This is occurring for certain products such as the album and Hoodie with Logo, as shown in the demo video below. It is not reproducible across all products. Here are some environmental details in case they are helpful.

WC Version: 9.9.0-dev
WP Version: 6.7.2
Twenty Twenty-Five


https://github.com/user-attachments/assets/c21f24db-9c38-4d42-b55e-ae9a0938879d

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/edit.tsx
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * @param   {Object}  props                    React props.
+ * @param   {Object}  props.attributes         Block attributes.
+ * @param   {Object}  props.setAttributes      Function to set block attributes.
+ * @param   {Object}  props.context           Inherited context.
+ * @param   {string}  props.context.commentId The comment ID.
+ *
+ * @return {JSX.Element} React element.
+ */
+export default function Edit( {
```
imanish003: I see an ESLint warning. Do you see it too?

image

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/edit.tsx
```diff
@@ -0,0 +1,94 @@
...
+	context: { commentId },
+	attributes,
+	setAttributes,
+}: BlockEditProps< {
+	textAlign: string;
+} > & {
+	context: { commentId: string };
+} ) {
+	const { textAlign } = attributes;
+	const styleProps = useStyleProps( attributes );
```
imanish003: Out of curiosity, do we really need to use `useStyleProps` here? Isn't there any alternative present in Gutenberg APIs? For example - [__experimentalGetColorClassesAndStyles](https://github.com/WordPress/gutenberg/blob/1c698f8ccdcc8a86b91f9e7efef608ad66f3a46b/packages/block-editor/src/index.js#L8). TBH I have not looked into Gutenberg APIs for this but I am wondering if there is an API in Gutenberg which we can use for this purpose, instead of custom API.

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/edit.tsx
```diff
@@ -0,0 +1,94 @@
...
+	const className = clsx(
+		styleProps.className,
+		'wc-block-components-product-review-rating',
```
imanish003: Nit: Is there any reason to have components in this className? 

```suggestion
		'wc-block-product-review-rating',
```

louwie17:
> This is occurring for certain products such as the album and Hoodie with Logo, as shown in the demo video below. It is not reproducible across all products. Here are some environmental details in case they are helpful.
> 
> WC Version: 9.9.0-dev WP Version: 6.7.2 Twenty Twenty-Five

@imanish003 unfortunately this is not super helpful! Are you able to dig in a bit deeper on your end to see what might be going on? For example are the actual accordions disappearing? or is there a style that is changing that is overlapping the two accordions? Are you able to reproduce this on a fresh site? Is this only happening for you on this branch or also on `trunk`?
What happens if you move the **blockified product details** block down below the **columns** or **Product Details** block, is the same thing happening?

We could also jump on a call.

Btw I am also using the twenty twenty-five theme in the video below on a fresh site and it looks a lot different than yours, especially the classic product reviews block. Are you sure your not importing any other styles or have other plugins enabled? Notice how yours has the borders for the classic reviews block and mine doesn't:

https://github.com/user-attachments/assets/31877d4b-1e32-4e58-bee4-cc97ccdccb4e

> Is this only happening for you on this branch or also on `trunk`?

If this is also happening on `trunk` and every thing else tests well, could we approve this PR? and continue debugging the above issue and fix it in a separate PR?

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/edit.tsx
```diff
@@ -0,0 +1,94 @@
...
+					value={ attributes.textAlign }
+					onChange={ ( newAlign ) => {
+						setAttributes( { textAlign: newAlign || '' } );
+					} }
+				/>
+			</BlockControls>
+			<div { ...blockProps }>
+				<div
+					className={
+						'wc-block-components-product-review-rating__stars'
```
imanish003:
```suggestion
						'wc-block-product-review-rating__stars'
```

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/edit.tsx
```diff
@@ -0,0 +1,94 @@
...
+		'wc-block-components-product-review-rating',
+		{
+			[ `has-text-align-${ textAlign }` ]: textAlign,
+		}
+	);
+	const blockProps = useBlockProps( {
+		className,
+	} );
+	let [ rating ] = useEntityProp( 'root', 'comment', 'rating', commentId );
+	rating = rating ?? 4;
```
imanish003: qq: Will we refactor this code to be review-specific or will this remain same?

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Internal dependencies
+ */
+import metadata from './block.json';
+import edit from './edit';
+
+if ( isExperimentalBlocksEnabled() ) {
+	// @ts-expect-error metadata is not typed.
+	registerBlockType( metadata, {
```
imanish003, line 16: I think we can use `registerProductBlockType` here as well. Alternatively, we can set `"ancestor": [ "woocommerce/blockified-product-reviews" ],` property.

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+		edit,
+		icon: {
+			src: (
+				<Icon
+					icon={ starHalf }
+					className="wc-block-editor-components-block-icon"
+				/>
+			),
+		},
```
imanish003: FWIW, out of curiosity, I checked whether directly passing the icon works, and it seems that it does. I was just curious and looked at the [code in Gutenberg](https://github.com/WordPress/gutenberg/blob/1c698f8ccdcc8a86b91f9e7efef608ad66f3a46b/packages/blocks/src/api/utils.js#L94-L130).

```suggestion
		icon: (
			<Icon
				icon={ starHalf }
				className="wc-block-editor-components-block-icon"
			/>
		),
```

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/reviews-title/edit.tsx
```diff
@@ -0,0 +1,170 @@
...
+					onChange={ ( value ) =>
+						setAttributes( { showReviewsCount: value } )
+					}
+				/>
+			</PanelBody>
+		</InspectorControls>
+	);
+
+	const postTitle = isSiteEditor
+		? __( '“Product Title”', 'woocommerce' )
```
imanish003:
```suggestion
		? __( '"Product Title"', 'woocommerce' )
```

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/reviews-title/edit.tsx
```diff
@@ -0,0 +1,170 @@
...
+					checked={ showReviewsCount }
+					onChange={ ( value ) =>
+						setAttributes( { showReviewsCount: value } )
+					}
+				/>
+			</PanelBody>
+		</InspectorControls>
+	);
+
+	const postTitle = isSiteEditor
```
imanish003:
```suggestion
	const productTitle = isSiteEditor
```

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/reviews-title/edit.tsx
```diff
@@ -0,0 +1,170 @@
...
+				/* translators: %s: Post title. */
+				__( 'Reviews for %s', 'woocommerce' ),
+				postTitle
+			);
+		}
+	} else if ( reviewsCount === 1 ) {
+		placeholder = __( 'Review', 'woocommerce' );
+	} else {
+		placeholder = __( 'Reviews', 'woocommerce' );
+	}
```
imanish003: qq: Is it possible to avoid nested if-else structures and instead use a flatter if-else structure? I find this code a bit harder to understand. Maybe it's just me who feels this way 😅

plugins/woocommerce/src/Blocks/BlockTypes/Reviews/ProductReviewRating.php
```diff
@@ -0,0 +1,85 @@
...
+	 * @return string Rendered block content.
+	 */
+	protected function render( $attributes, $content, $block ) {
+		if ( ! isset( $block->context['commentId'] ) ) {
+			return '';
+		}
+
+		$rating = intval( get_comment_meta( $block->context['commentId'], 'rating', true ) );
+
+		$classes_and_styles = StyleAttributesUtils::get_classes_and_styles_by_attributes( $attributes, array(), array( 'extra_classes' ) );
```
imanish003: Is it necessary to use `StyleAttributesUtils`? Is it possible to use [get_block_wrapper_attributes](https://github.com/woocommerce/woocommerce/issues/51648) here? TBH I haven't looked deeply into it but I remembered that we were discussing about moving away from this.

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/reviews-title/style.scss
```diff
@@ -0,0 +1,4 @@
+
+.wp-block-woocommerce-product-reviews-title.has-background {
```
imanish003, line 2: Should we consider using a class with a WC prefix instead of a WP prefix here? This way, when third-party developers add some styles, they can rely on this specific class for this block?

```suggestion
.wc-block-product-reviews-title.has-background {
```

plugins/woocommerce/client/blocks/assets/css/style.scss
```diff
@@ -99,10 +99,12 @@ body.wc-block-product-gallery-modal-open {
 	}
 }
 
-.wc-block-grid__product-rating {
+.wc-block-grid__product-rating,
+.wp-block-woocommerce-product-review-rating {
```
imanish003, line 103: Similar to the other comment, Should we consider using a class with a WC prefix instead of a WP prefix here?

imanish003 (changes requested):
@louwie17, you’ve done an amazing job on this PR! Everything looks really good overall. I’ve shared a few minor comments and suggestions for you to consider. I’d love to hear your thoughts on them, and once we go over those little details, I’ll be happy to approve the PR! 🚀

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/edit.tsx
louwie17: Yeah I got the same thing, I will remove it actually, as it is not common for us to add these JS docs to begin with and the component is also typed already.
I had initially mimicked this from the GB approach I guess.

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/edit.tsx
louwie17: Which code are you referring to specifically?
We will continue to use the `useEntityProp` with the `commentId` as context, given we use the same context as part of the variations from the comment blocks.
Currently the above works fine if we provide a review id, and its not supported to use `useEntityProp( 'root', 'product-review', 'rating', commentId );` for example.

I may explore adding the support in the review data store for retrieving a single review https://woocommerce.github.io/woocommerce-rest-api-docs/#retrieve-a-product-review and use that instead.

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/index.tsx
louwie17: Yeah even the use of `Icon` is not necessary, and we can just use `icon: starHalf`.
I don't quite remember doing this explicitly myself, so I may of copied this from the average rating block: https://github.com/woocommerce/woocommerce/blob/99f04562c084808e02515a526664fdc0c77fa688/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/average-rating/index.tsx#L16

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/reviews-title/edit.tsx
louwie17: Hmm while I see your point, I am afraid it may not help, and if we do flatten the structure, we end up with a lot of large if statements. I also debated on doing some concatenation fun, but that would mess up the translation strings.

Btw I didn't write these if statements myself: https://github.com/WordPress/gutenberg/blob/317c348874fb5d4cf47297351aee4c9098cf15c6/packages/block-library/src/comments-title/edit.js#L138

louwie17: Let me move the logic to its own function actually and make use of early return statements, those may help

plugins/woocommerce/src/Blocks/BlockTypes/Reviews/ProductReviewRating.php
louwie17: Nice catch! This was also copy/paste from the `ProductAverageRating` class 🤦‍♂️

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/reviews-title/style.scss
louwie17: Hmm is this something we do in other places, or do we have some logic that auto adds a `wc-block` generated class?
To be honest, on order for me to do the above I have to add the `wc-block-*` class specifically in multiple places, I rather rely on the auto generated `wp-block` class that is based off of the block name.

plugins/woocommerce/client/blocks/assets/css/style.scss
louwie17: I replied above, but in my opinion "no". Unless we have a specific guideline that we follow to do this for all our WooCommerce blocks.
If not, I think relying on the block generated name is just as fine.

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/edit.tsx
louwie17: Removed the comment block as part of this commit: https://github.com/woocommerce/woocommerce/pull/56332/commits/ab16e279c648dd788b7cafea8172c02e9538e9fa

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/edit.tsx
louwie17: Removed it: https://github.com/woocommerce/woocommerce/pull/56332/commits/ab16e279c648dd788b7cafea8172c02e9538e9fa :)

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/edit.tsx
louwie17: Nope, removed it: https://github.com/woocommerce/woocommerce/pull/56332/commits/ab16e279c648dd788b7cafea8172c02e9538e9fa

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/edit.tsx
louwie17: I end up adding the `getReview` selector to the reviews data store and used that here instead of `useEntityProp`: https://github.com/woocommerce/woocommerce/pull/56332/commits/540e2ff7da385331ebedd2297b4b704daf518d41

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/index.tsx
louwie17: I made use of `ancestor` -> https://github.com/woocommerce/woocommerce/pull/56332/commits/68afb7b637ed741782dd89e04ef58e08f4c674ea

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/index.tsx
louwie17: Just passed icon directly: https://github.com/woocommerce/woocommerce/pull/56332/commits/1b60ec1bb62284ce683c744014552a7006ef5a77

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/reviews-title/edit.tsx
louwie17: I moved it to a function and made use of the early return statements to flatten the conditions, I did it both for the JS version and the PHP version: https://github.com/woocommerce/woocommerce/pull/56332/commits/346c7250db33f3a7958ed644dfa206f18c99f402
I hope this helps :)

plugins/woocommerce/src/Blocks/BlockTypes/Reviews/ProductReviewRating.php
louwie17: Removed: https://github.com/woocommerce/woocommerce/pull/56332/commits/e1ea5daa95f09cb62bbc5593cd86dd295d691a60

louwie17:
@imanish003 thanks for the detailed review, I addressed your suggestions and left a couple replies to some of your other questions, in particular around the `wc-block-*` class.

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/reviews-title/style.scss
imanish003: I see and I somewhat agree with you that it might be better to rely on the auto-generated class instead of specifying the class with the WC prefix in multiple places, as that could lead to more errors. I also asked about this on Slack (p1742547412664939-slack-C02FL3X7KR6).

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-rating/index.tsx
imanish003: Nice 🚀

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/reviews-title/edit.tsx
```diff
@@ -0,0 +1,187 @@
...
+function getProductReviewsTitle(
+	showReviewsCount: boolean,
+	reviewsCount: number,
+	showProductTitle: boolean,
+	productTitle: string
+) {
+	if ( showReviewsCount && showProductTitle ) {
+		return reviewsCount === 1
+			? sprintf(
+					/* translators: %s: Post title. */
```
imanish003: Nit: Can we update this in all places? 

```suggestion
					/* translators: %s: Product title. */
```

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/reviews-title/edit.tsx
imanish003: That looks much better to me. Thanks a lot for making this change 🙇

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/reviews-title/style.scss
imanish003: It seems that we have reached a consensus on Slack to use the auto-generated class. Therefore, I am resolving this comment.

packages/js/data/src/reviews/resolvers.ts
```diff
@@ -36,6 +36,23 @@ export function* getReviews( query: ReviewsQueryParams ) {
 	}
 }
 
+export function* getReview( id: number ) {
+	try {
+		const url = addQueryArgs( `${ NAMESPACE }/products/reviews/${ id }` );
```
imanish003: I checked, and here `NAMESPACE` points to `/wc-analytics`. Is there a reason we rely on this endpoint? As far as I understand, this endpoint is specifically for views related to analytics. Should we consider using `wc/v3` instead? I’m new to this part of the codebase, so I may have misunderstood something.

imanish003 (changes requested):
Thanks for making all the changes. Everything looks good to me now, except for a few more minor things—especially my comment about `NAMESPACE` constant.

packages/js/data/src/reviews/resolvers.ts
louwie17: Yeah I did notice the same thing, but turns out that the `wc-analytics` makes use of the REST Api v3 class in behind the scenes: https://github.com/woocommerce/woocommerce/blob/3dd60322d5490d7d86ad33e7d732f276df73a12d/plugins/woocommerce/src/Admin/API/ProductReviews.php#L18

Having said that it is probably worth refactoring this, so the review class uses `wc/v3` always and we make an exception for the analytics endpoint.
I would suggest I do this as a follow up, to unblock this work.

I did remove the use of namespace for now: https://github.com/woocommerce/woocommerce/pull/56332/commits/ea5ea6d0292585e884a044844cbca4be966ad224 and will update the `getReviews` function in a follow up.

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/reviews-title/edit.tsx
louwie17: Done, thanks: https://github.com/woocommerce/woocommerce/pull/56332/commits/f643843317f856cfad884a280fbb0fb194d3d1c7

louwie17:
@imanish003 should be good for a re-review, I also replied to your question around the analytics endpoint usage. I think we should update this as a follow up. The endpoint uses the `wc/v3` class in behind the scenes and only updates the prepare links as the difference: https://github.com/woocommerce/woocommerce/blob/3dd60322d5490d7d86ad33e7d732f276df73a12d/plugins/woocommerce/src/Admin/API/ProductReviews.php#L32

I initially debated on creating a new data store that uses the `wc/v3` endpoint only, but that would add additional confusion having two review data stores. I think this store should be updated to use `wc/v3` by default and use the analytics endpoint if needed.
But as suggested in the comment I think this should be a follow up.

packages/js/data/src/reviews/resolvers.ts
imanish003: I see! FWIW there is also a feature flag, and based on that flag's value, [some of the analytics API may not be loaded](https://github.com/woocommerce/woocommerce/blob/ea5ea6d0292585e884a044844cbca4be966ad224/plugins/woocommerce/src/Admin/API/Init.php#L124). However, it looks like the product reviews analytics API will always be loaded. So yeah, I think we can merge this as it is, and as you mentioned, you could maybe follow up in a new PR.

> I did remove the use of namespace for now: https://github.com/woocommerce/woocommerce/commit/ea5ea6d0292585e884a044844cbca4be966ad224 and will update the getReviews function in a follow up.

Nice! 🚀

imanish003:
> @imanish003 should be good for a re-review, I also replied to your question around the analytics endpoint usage. I think we should update this as a follow up. The endpoint uses the `wc/v3` class in behind the scenes and only updates the prepare links as the difference:
> 
> https://github.com/woocommerce/woocommerce/blob/3dd60322d5490d7d86ad33e7d732f276df73a12d/plugins/woocommerce/src/Admin/API/ProductReviews.php#L32
> 
> I initially debated on creating a new data store that uses the `wc/v3` endpoint only, but that would add additional confusion having two review data stores. I think this store should be updated to use `wc/v3` by default and use the analytics endpoint if needed. But as suggested in the comment I think this should be a follow up.

Yeah, this makes sense to me. I am going to do final testing before approving the PR.

plugins/woocommerce/src/Blocks/BlockTypes/Reviews/ProductReviewsTitle.php
```diff
@@ -0,0 +1,128 @@
...
+		$wrapper_attributes = get_block_wrapper_attributes( array( 'class' => $align_class_name ) );
+		$reviews_count      = $product->get_review_count();
+		$tag_name           = 'h2';
+		if ( isset( $attributes['level'] ) ) {
+			$tag_name = 'h' . $attributes['level'];
+		}
+
+		if ( 0 === $reviews_count ) {
+			return;
+		}
```
imanish003: I just wanted to confirm whether this is intentional because, according to the PR description, it should display text that says "0 reviews for...". However, based on this condition, it won't show anything at all on the front end. So, I wanted to clarify if this is intentional or not.

imanish003: If this is intentional, I believe we (should?) also update the editor and shouldn't display anything on the editor side. Here is a screenshot comparing the editor and the front endview.
image

imanish003 (approved):
Hey @louwie17, thank you so much for your patience and amazing work on this PR. I know there has been a lot of back and forth, and I truly appreciate your patience. Everything looks good to me now. I have tested the PR again, and everything is working as expected. I left one minor comment, but I am going to pre-approve the PR. I think once we address that comment, we can merge this PR. 🚀

plugins/woocommerce/src/Blocks/BlockTypes/Reviews/ProductReviewsTitle.php
louwie17: Good call out, I decided to still show the title on the front end as well https://github.com/woocommerce/woocommerce/pull/56332/commits/a46845693705fde017a6d9b88117a4de23b7fd92, I had mimicked this logic from the core comments title block as well.
The problem is that you can't really hide the title in the block editor, or I am not aware of a good way, that would still allow you to edit it.
If we do want to have the option to hide it, we could add a setting to "hide title if no reviews", we can keep that conversation going as a follow up. In either case things should be identical now.

@j111q & @ellenbauer do you think we need a setting for the reviews title to give users the option to hide it if there are 0 reviews? Or should we always show a title?

imanish003: Sure, it makes sense to continue the discussion and, if needed, follow up with the PR. FWIW I wanted to point out that in the screenshot above, the ratings block is showing "no reviews" on the editor side, and on the front end, it is not rendering anything at all similar to this block.

j111q: I actually wonder if we should NEVER show a title, for any of the accordion sections -- meaning we don't show the "Description" title or "Additional info" or "Reviews" title, ever. Because the Accordion title itself already serves as a title, if you get what I mean. 

Here's some nice Woo stores as an example:
- https://bario-neal.com/product/rittenhouse-chain/
- https://shop.thecoolhunter.net/product/silver-cocktail-coasters/

PR #56263, opened 2025-03-10
[Experimental] Add to Cart Button: show single add to cart text when inside the Add to Cart with Options block

Aljullu:
### Changes proposed in this Pull Request:

Closes https://github.com/woocommerce/woocommerce/issues/56218.

This PR updates the _Add to Cart Button_ block to display the single add to cart text when it's a descendant of the _Add to Cart with Options_ block.

### How to test the changes in this Pull Request:

0. Install and activate the WooCommerce Beta Tester plugin
0. Go to _WooCommerce Admin Test Helper_ > _Features_ and enable `experimental-blocks` and `blockified-add-to-cart`.
1. Go to _Appearance_ > _Editor_ > _Templates_ > _Single Product_ and update the _Add to Cart with Options_ block to the blockified version.
2. Switch between product types and verify the button always says "Add to cart" except for external products, in which case it says "Buy product".

[Enregistrament de pantalla del 2025-03-10 15-43-38.webm](https://github.com/user-attachments/assets/f3dbe45d-6520-4860-90b6-5eade91ec793)

3. In the frontend, visit the page of a grouped product, verify the cart button says "Add to cart" instead of "View products".

Before | After
--- | ---
imatge | imatge

4. Visit the page of an external product and verify the button says "Buy product" (or whatever is defined in the product settings).
5. Verify there are no regressions in the Add to Cart Button in other contexts. For example, by going to the _Shop_ page and verifying the button of the grouped product says "View products" instead of "Add to cart".

imatge

PR #56358, opened 2025-03-13
Prevent product button loading unnecessary dependencies

Aljullu:

(For Bug Fixes) Bug introduced in PR https://github.com/woocommerce/woocommerce/pull/56263.

This PR moves the `useBlockProps()` usage and the `@wordpress/block-editor` import from the Product Button block component to the edit one, ensuring we get rid of the unnecessary dependencies in the frontend which were introduced in #56263.

1. Open the _Network_ tab of your browser devtools (<kbd>F12</kbd>).
2. Go the _Cart_ or _Checkout_ page.
3. In the _Network_ tab verify there are no scripts loaded like `/wp-content/plugins/gutenberg/build/block-editor/index.min.js` or `/wp-content/plugins/gutenberg/build/components/index.min.js`.

Also repeat the tests from https://github.com/woocommerce/woocommerce/pull/56263 to make sure there are no regressions.

### DIFF
--- a/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/block.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/block.tsx
@@ -175,12 +175,10 @@ const AddToCartButtonPlaceholder = ( {
 	useEffect( () => {
 		if ( blockClientId ) {
 			registerListener( blockClientId );
-		}
-		return () => {
-			if ( blockClientId ) {
+			return () => {
 				unregisterListener( blockClientId );
-			}
-		};
+			};
+		}
 	}, [ blockClientId, registerListener, unregisterListener ] );
 
 	const buttonText =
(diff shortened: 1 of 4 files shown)


plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/block.tsx
```diff
@@ -165,20 +164,22 @@ const LoadingAddToCartButton = ( {
...
-
 	const {
 		current: currentProductType,
 		registerListener,
 		unregisterListener,
 	} = useProductTypeSelector();
 
 	useEffect( () => {
-		registerListener( blockClientId );
+		if ( blockClientId ) {
```
mdperez86, line 176:
```suggestion
		if ( blockClientId ) {
				registerListener( blockClientId );
				return () => {
						unregisterListener( blockClientId );
				};
		}
```

I prefer this way so the returned callback will be executed only if the blockClientId exists and not always.

mdperez86 (approved):
Nice job @Aljullu!

I left one comment that is not a blocker, also not a big deal. Feel free to merge this PR as soon as you need it.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/block.tsx
Aljullu: Good suggestion, merged!

Issue #52630, opened 2024-11-06
uri?add-to-cart=<productId> url is adding product twice to cart if url is redirected


alapiere:
### Prerequisites

- [x] I have carried out troubleshooting steps and I believe I have found a bug.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

When querying /checkout?add-to-cart=123 , product is added twice to cart instead of once

### Expected behavior

Product must be added only once even when url is redirected,
it must not be added to cart on redirected url

### Actual behavior

When querying /checkout?add-to-cart=123 ,
I'm redirected 301 to the correct url which is /checkout/?add-to-cart=123 ( "/" after checkout)

Product is added to cart on first 301 hit,
then again on the correct url,
--> leaving the product twice in cart

### Steps to reproduce

- go to /checkout?add-to-cart=123 (replace 'checkout' by your checkout slug if it's different)
- user is redirected to /checkout/?add-to-cart=123 
- product is added twice to cart

### WordPress Environment

<details><summary>System Status Report</summary>

```
### WordPress Environment ###

WordPress address (URL): [Redacted]
Site address (URL): [Redacted]
WC Version: 9.3.3
Legacy REST API Package Version: ✔ 1.0.4
Action Scheduler Version: ✔ 3.8.1
Log Directory Writable: ✔
WP Version: 6.6.2
WP Multisite: –
WP Memory Limit: 256 MB
WP Debug Mode: –
WP Cron: ✔
Language: fr_FR
External object cache: –

### Server Environment ###

Server Info: nginx/1.25.5
PHP Version: 8.1.30
PHP Post Max Size: 128 MB
PHP Time Limit: 300
PHP Max Input Vars: 10000
cURL Version: 7.68.0
OpenSSL/1.1.1f

SUHOSIN Installed: –
MySQL Version: 10.11.5-MariaDB-1:10.11.5+maria~ubu2004
Max Upload Size: 128 MB
Default Timezone is UTC: ✔
fsockopen/cURL: ✔
SoapClient: ✔
DOMDocument: ✔
GZip: ✔
Multibyte String: ✔
Remote Post: ✔
Remote Get: ✔

### Database ###

[REDACTED]

### Post Type Counts ###

acf-field: 83
acf-field-group: 13
acf-post-type: 4
acf-taxonomy: 2
apto_sort: 8
artist: 159
attachment: 4402
custom_css: 1
customize_changeset: 76
[...]

PR #56360, opened 2025-03-13
Fix early calls to page conditionals breaking cache

### LINKED ISSUE #56320: `woocommerce-cart` and `woocommerce-page` classes are missing from the short-code cart
### Prerequisites

### Describe the bug

When a store uses the short-code cart and a product is added to the cart via a direct link, like so: `http://wp.local/cart/?add-to-cart=19`, then the Cart page is missing some essential classes like `woocommerce-cart` and `woocommerce-page` from the `body`.

This seems to have broken after: https://github.com/woocommerce/woocommerce/pull/56167

### Expected behavior

`woocommerce-cart` and `woocommerce-page` should always be present in the `body` when the short-code Cart is used. 

### Actual behavior

`woocommerce-cart` and `woocommerce-page` are missing `body` when the short-code Cart is used. 

### Steps to reproduce

1. Activate a classic theme (Storefront or 2020).
2. Create a cart page using the `[woocommerce_cart]` short-code. 
3. Create a Simple Product and take note of its ID. 
4. Construct an add-to-cart link that redirects the user to the cart after a successful addition, like so: `http://wp.local/cart/?add-to-cart=19`, where 19 is the product ID. 
5. Inspect the `body` and notice that the `woocommerce-cart` and `woocommerce-page` classes are missing.
6. Refresh the page -- the `woocommerce-cart` and `woocommerce-page` classes are now added. 
### WordPress Environment

Fresh WordPress install with latest WooCommerce `trunk`/`nightly` + Storefront.

### Isolating the problem

mikejolley:

#56167 fixed is_cart/is_checkout and introduced a cache. This cache relies on page conditionals which are only available after `pre_get_posts`. This PR ensures the page conditionals return `false` before `pre_get_posts` and do not cache.

This is related to https://github.com/woocommerce/woocommerce/issues/52630 since add-to-cart running early is the culprit.

Closes #56320

1. Activate a classic theme (Storefront or 2020).
2. Create a cart page using the [woocommerce_cart] short-code.
3. Create a Simple Product and take note of its ID.
4. Construct an add-to-cart link that redirects the user to the cart after a successful addition, like so: http://wp.local/cart/?add-to-cart=19, where 19 is the product ID.
5. Inspect the body and check that the `woocommerce-cart` and `woocommerce-page` classes are present.

The above steps on blocks/shortcode cart pages with and without a block theme.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Blocks/Utils/CartCheckoutUtils.php
```diff
@@ -24,22 +24,25 @@ class CartCheckoutUtils {
...
-		} else {
-			global $post;
-			// Check page contents for block/shortcode.
-			self::$is_cart_page = is_a( $post, 'WP_Post' ) && ( wc_post_content_has_shortcode( 'woocommerce_cart' ) || self::has_block_variation( 'woocommerce/classic-shortcode', 'shortcode', 'cart', $post->post_content ) );
+		if ( null === self::$is_cart_page ) {
+			$page_id            = wc_get_page_id( 'cart' );
+			self::$is_cart_page = $page_id && is_page( $page_id );
+
+			// If the is_page check returned false, check the page contents for a cart block or shortcode.
+			if ( ! self::$is_cart_page && is_a( $post, 'WP_Post' ) ) {
```
nielslange: Should we use `$post instanceof WP_Post` instead of `is_a( $post, 'WP_Post' )` here? Since `instanceof` is an operator rather than a function, it generally performs better.

AI helped me to create the following snippet for benchmarking this:

```php
add_action('init', function () {
    global $post;
    $post = new WP_Post((object) ['ID' => 1]); // Mock WP_Post
    $iterations = 10_000_000; // 10 million for better averaging
    $trials = 5; // Run multiple trials

    function benchmark($callback, $iterations, $trials) {
        $times = [];
        for ($t = 0; $t < $trials; $t++) {
            $start = hrtime(true);
            for ($i = 0; $i < $iterations; $i++) {
                $callback();
            }
            $times[] = (hrtime(true) - $start) / 1e9; // Convert ns to seconds
        }
        return array_sum($times) / count($times); // Return average
    }

    $instanceofTime = benchmark(fn() => $post instanceof WP_Post, $iterations, $trials);
    $isaTime = benchmark(fn() => is_a($post, 'WP_Post'), $iterations, $trials);

    echo "<pre>instanceof: " . number_format($instanceofTime, 6) . "s\n";
    echo "is_a: " . number_format($isaTime, 6) . "s\n";
    echo "Difference: " . number_format(($isaTime - $instanceofTime) / $instanceofTime * 100, 2) . "%\n</pre>";
});
```

On my end, `instanceof` is about 30% faster than `is_a()`.

plugins/woocommerce/src/Blocks/Utils/CartCheckoutUtils.php
```diff
@@ -70,22 +73,25 @@ function ( $method ) {
...
-			self::$is_checkout_page = true;
-		} else {
-			global $post;
-			// Check page contents for block/shortcode.
-			self::$is_checkout_page = is_a( $post, 'WP_Post' ) && ( wc_post_content_has_shortcode( 'woocommerce_checkout' ) || self::has_block_variation( 'woocommerce/classic-shortcode', 'shortcode', 'checkout', $post->post_content ) );
+		if ( null === self::$is_checkout_page ) {
+			$page_id                = wc_get_page_id( 'checkout' );
+			self::$is_checkout_page = $page_id && is_page( $page_id );
+
+			// If the is_page check returned false, check the page contents for a cart block or shortcode.
```
nielslange: This seemed to have slipped though:

```suggestion
			// If the is_page check returned false, check the page contents for a checkout block or shortcode.
```

nielslange (commented):
Thanks for working on this, @mikejolley. When reviewing this, I initially wasn’t able to reproduce the issue. @PanosSynetos jumped on a call with me, and we figured out that this particular issue only affects classic themes. Panos updated the steps to reproduce and the testing instructions accordingly.  

While the change works as expected, I noticed a few things:  

### Performance

I wonder if we should use `instanceof` instead of `is_a()`, given that it performs **30% faster on average**. I shared a benchmark snippet.  

### DRY and Type Hinting
I noticed that both `is_cart_page()` and `is_checkout_page()` are fairly similar. What are your thoughts on using the **Template Method Pattern** here and making the parameter and return types explicit, e.g.:

```php
private static function is_woocommerce_page( string $page_type ): bool {
    [...]
}

public static function is_cart_page(): bool {
    return self::is_woocommerce_page( 'cart' );
}

public static function is_checkout_page(): bool {
    return self::is_woocommerce_page( 'checkout' );
}
```

### PHPDocs 

The two functions you've updated have the following PHPDocs:  

```php
/**
  * Returns true if:
  * - The cart page is being viewed.
  * - The page contains a cart block, cart shortcode, or classic shortcode block with the cart attribute.
  *
  * This function cannot be used accurately before the `pre_get_posts` action has been run.
  *
  * @return bool
  */
```
```php
/**
  * Returns true if:
  * - The checkout page is being viewed.
  * - The page contains a checkout block, checkout shortcode, or classic shortcode block with the checkout attribute.
  *
  * This function cannot be used accurately before the `pre_get_posts` action has been run.
  *
  * @return bool
  */
```

While not crucial, I wonder if we should also improve the PHPDocs now that we're modifying these functions?

That said, the PR itself works as expected and below, I've posted before and after screenshots.

<table>
<tr>
<td valign="top">Before:
<br><br>
Screenshot 2025-03-18 at 18 50 19
</td>
<td valign="top">After:
<br><br>
Screenshot 2025-03-18 at 18 41 13
</td>
</tr>
</table>

plugins/woocommerce/src/Blocks/Utils/CartCheckoutUtils.php
mikejolley: Good shout, I agree. I've changed the usage in this touched file in https://github.com/woocommerce/woocommerce/pull/56360/commits/e56c17fa1a50b70ca71a4ea6cd2b1eb6223aac63 The codebase has lots of `is_a` usage but this should be out of scope here.

mikejolley:
> I noticed that both is_cart_page() and is_checkout_page() are fairly similar. What are your thoughts on using the Template Method Pattern here and making the parameter and return types explicit, e.g.:

Given the similarities, I think you're right. As these are class methods we can create a private helper to share the logic, and the name of the shared method can be generic. We just need to take care of the shortcode/block naming as it differs per page. Implemented in https://github.com/woocommerce/woocommerce/pull/56360/commits/0c7020c127ff2908c3b74705dfe5e4a6367c52e3

Ill re-request review once I've double checked source.

plugins/woocommerce/src/Blocks/Utils/CartCheckoutUtils.php
nielslange: > The codebase has lots of is_a usage but this should be out of scope here.

Yes, changing all `is_a()` against `instanceof` across the codebase, should not be part of this PR. That said, given the performance impact, I'd say it's worth exploring how much faster WooCommerce could be if we'd change the existing sections. I've created https://github.com/woocommerce/woocommerce/issues/56495 for tracking this.

nielslange (approved):
Thanks for applying my suggestions, @mikejolley. Looking at `is_cart_page()` and `is_checkout_page()`, we could potentially further simplify these functions by moving the similar logic into `is_page_type()`. We could use [variable variables](https://www.php.net/manual/en/language.variables.variable.php) to do so. This is what I had in mind:

```php
private static function is_page_type( string $page_type ):bool|null {
  $property = "is_{$page_type}_page";

  if ( ! isset( self::$$property ) ) {
    ...
  }

  return self::$$property;
}
```

This change would simplify the other two functions:

```php
public static function is_cart_page():bool {
  return self::is_page_type( 'cart' );
}
```
```php
public static function is_checkout_page():bool {
  return self::is_page_type( 'checkout' );
}
```

That said, this approach relies on variable variables, which may reduce readability for some developers. I’m pre-approving this PR, given that it works as expected and that @woocommerce/somewherewarm is blocked by this PR. I'll leave it up to you whether you want to implement this suggestion.

mikejolley:
Thanks @nielslange I'd prefer to keep the caching part separate to `is_page_type` so I'll leave it how it is now. CI is passing, `WP_Post` should have been `\WP_Post` due to namespaces.

nielslange:
> Thanks @nielslange I'd prefer to keep the caching part separate to `is_page_type` so I'll leave it how it is now. CI is passing, `WP_Post` should have been `\WP_Post` due to namespaces.

Sounds good to keep it separated, and sorry for missing `\WP_Post`. 🫣

webdados:
Before this PR ([or this one](https://github.com/woocommerce/woocommerce/pull/54340) maybe), `is_checkout()` was working correctly on a custom function hooked to the `woocommerce_billing_fields` filter.

Now, in some edge cases that I still haven't been able to reproduce in any of my installs (but have 3 customers where I could confirm that problem exists), `is_checkout()` is returning false on the checkout page. What I've been able to determine is that `$post `was still not available when the function run, so of course it would be impossible to correctly detect the page.

PR #56366, opened 2025-03-13
[dev] Update release checks workflows

### LINKED ISSUE #51027: Run the daily-checks against the nightly package
More context: pf59Ax-1Pg-p2#comment-1532

Current state:
- the daily-checks are triggered on a schedule, via the `tests-daily-run.yml` workflow, that triggers `ci.yml` with the 'daily-checks' event. `ci.yml` then runs the configured checks as classic checks, running the build command and mapping the plugin dir in wp-env.
- the release-checks are triggered whenever a release is published or updated. `ci.yml` the configured checks but skips the build step, downloads the release package and includes it in the plugins list in wp-env config to be installed.

Proposed state:
Switch the daily-checks to use the nightly release instead, which is updated every night.

### Tasks
- [ ] Consolidate both the the release-checks and the nightly-checks in a single workflow, that should trigger ci.yml with similar parameters.
- [ ] Have the nightly-checks run on a schedule every night, after the nightly build is created.
- [ ] Remove `tests-daily-run.yml`.

adimoldovan:

Closes https://github.com/woocommerce/woocommerce/issues/51027

Improve the flexibility and clarity of the CI workflows, particularly around handling different types of test runs.
With this change the nightly build will be used in the nightly (former daily) checks. This means we'll be using the same workflow with an almost identical setup as we do for release checks, ensuring we're testing changes to the workflow on a daily basis preventing issues during the release.

Changes:
- Removed the release trigger in `ci.yml`, and added new inputs: `refName` and `artifactName`. The `refName` will be used to checkout the code in case it's set.
- Added a new workflow `tests-on-release.yml` for running release checks. It includes schedule, release, and workflow dispatch triggers. Uses ci.yml with nightly-checks or release-checks triggers.
- Changed the trigger input from daily-checks to nightly-checks.
- Minor tweaks on formatting and run names

Tested locally with [act](https://nektosact.com/usage/index.html).
Checks should run for this PR the same as before.
The new workflows cannot be tested until after merge.


.github/workflows/ci.yml
```diff
@@ -165,6 +163,8 @@ jobs:
     steps:
       - uses: 'actions/checkout@v4'
         name: 'Checkout'
+        with:
+          ref: ${{ inputs.refName != '' && inputs.refName || github.ref_name }}
```
kalessil: We will be better off falling back to `''` so the checkout action keeps doing its internal magic, which is currently stable.

kalessil: There are more similar places in this workflow that this comment covers, but I'll avoid duplicating them yet.

adimoldovan: Good point, it was failing on prs anyway with ref_name. Updated with [5b7700](https://github.com/woocommerce/woocommerce/pull/56366/commits/25b770027042c30e36af1f2cf57a0b315073ec8f)

.github/workflows/tests-on-release.yml
```diff
@@ -0,0 +1,25 @@
+name: 'Release checks run'
+run-name: ${{ github.event_name == 'schedule' && 'scheduled' || github.event_name == 'workflow_dispatch' && 'on demand'  || github.event_name == 'release' && 'release' || '' }} checks for ${{ inputs.tag != '' && inputs.tag || github.ref_name }}
```
kalessil, line 2: Worth adding `( ... )` for better clarity to visually isolate the possible values here.

.github/workflows/tests-daily-run.yml
```diff
@@ -1,7 +1,5 @@
-name: 'Daily tests run'
+name: 'Daily checks run'
```
kalessil, line 1: Don't know how much sense it makes domain-wise, but as I read the changes it looks like this workflow could trigger `.github/workflows/tests-on-demand.yml` instead.

.github/workflows/ci.yml
```diff
@@ -468,7 +469,7 @@ jobs:
 
     steps:
       - uses: 'actions/checkout@v4'
-
+
```
kalessil, line 472: Noted unneccesary spaces here.

kalessil (approved):
LGTM (review only)! Leaving some optional comments.

.github/workflows/tests-daily-run.yml
adimoldovan: The plan is to remove this workflow in a followup, once I validate the nightly works as expected. I only kept it for now while transitioning to the new release workflow that will use the nightly build.

kalessil: Sounds good!

PR #56367, opened 2025-03-13
[Store API] Line items updated on every request due to bug in `update_line_items_from_cart` logic

### LINKED ISSUE #56284: Cart item hashes are retrieved via WC_Order::get_meta_data() instead of WC_Order::get_meta()
### Prerequisites

### Describe the bug

I was investigating some really strange race conditions caused by the checkout block using the StoreAPI. Some items (fees, shipping, taxes) were added twice to the order under some rare conditions. 

When digging through the code I found that the [OrderController](https://github.com/woocommerce/woocommerce/blob/5d13ee5eaa1e3ae20c44c9e9c8cb019fc707d4d4/plugins/woocommerce/src/StoreApi/Utilities/OrderController.php#L706C3-L706C82) seems to remove/add all items except line items on each request (that increases the chance of the race condition to occur). That is because the hashes are retrieved via the wrong method. It needs to be `WC_Order::get_meta()` instead of `WC_Order::get_meta_data()`. 

```php 
if ( $order->get_meta_data( '_shipping_hash' ) !== $cart_hashes['shipping'] ) {
```

needs to be

```php 
if ( $order->get_meta( '_shipping_hash' ) !== $cart_hashes['shipping'] ) {
```

This has quite some big performance implications. 

What is still missing is some kind of locking mechanism - currently what may happen (due to multiple requests, e.g. [extensionCartUpdate](https://github.com/woocommerce/woocommerce-blocks/blob/trunk/docs/third-party-developers/extensibility/rest-api/extend-rest-api-update-cart.md) happening at the same time or still running when hitting the checkout button) is that the `OrderController::update_order_from_cart()` method is called by the cart endpoint right when the Checkout is already processed via the checkout route. This can lead to all kind of (very hard to debug) problems so either the Checkout Block needs to really block the submit button until all requests have finished or there needs to be some blocking mechanism within the API to make sure such concurrent requests targeting the same draft order may not happen.

Best,
Dennis

### Expected behavior

Retrieve meta data with the right getter :)

### Actual behavior

Retrieves all meta data but not the meta data for the expected key.

### Steps to reproduce

Please have a look at the code.

### WordPress Environment

Latest Woo, WP.

### Isolating the problem

mikejolley:

`update_line_items_from_cart` is supposed to check cart hashes to prevent recreating line items on every request. But due to using `$order->get_meta_data()` (which returns all meta as an array) instead of `$order->get_meta()` (which returns the value), this logic was broken.

Closes #56284

This bug is invisible in regular usage. It mainly impacts performance. CI should catch issues as it tests checkouts and Store API updates.

Optionally, smoke testing can be done by placing an order via block checkout and confirming line items exist in the order.

- Debugging output of the meta functions when making API requests
- Placing order via API
- Placing order via block checkout

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


wavvves (approved):
Nice report with good impact. Approved :)

PR #56369, opened 2025-03-13
[e2e-tests] Update @wordpress/e2e-* deps to wp-6.8

WunderBart:
## What?

Follow up to https://github.com/woocommerce/woocommerce/pull/56262.

Clean up and update @wordpress/e2e-* deps to wp-6.8.

## How to test

CI passes.


.syncpackrc
```diff
@@ -304,7 +304,7 @@
 			"packages": [
 				"**"
 			],
-			"pinVersion": "^1.15.0"
+			"pinVersion": "wp-6.8"
```
kalessil, line 282: For aligning with similar deps, we need to remove this section and add the dependencies under `"label": "@wordpress/* build-dependencies (always ahead of runtime-dependencies)",` section in `.syncpackrc`.

It's kind of an experiment, but it's still worth keeping things consolidated.

kalessil: PS: if it won't work then at least let's relocate this section near the referenced one.

WunderBart: I'm not sure how to move it to that section - what's the syntax for pinning versions of multiple packages? Or am I misreading it?

kalessil: Like this:

```diff
Subject: [PATCH] Syncpack
---
Index: .syncpackrc
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/.syncpackrc b/.syncpackrc
--- a/.syncpackrc	(revision 074ccb6f7bdc0a34dd0f541335af58309c992e58)
+++ b/.syncpackrc	(date 1741940923723)
@@ -298,15 +298,6 @@
 			"pinVersion": "10.17.0"
 		},
 		{
-			"dependencies": [
-				"@wordpress/e2e-test-utils-playwright"
-			],
-			"packages": [
-				"**"
-			],
-			"pinVersion": "^1.15.0"
-		},
-		{
 			"label": "Ensure @woocommerce/* dependencies are locked to the monorepo itself - except certain dependencies (migration pending).",
 			"dependencies": [
 				"@woocommerce/settings"
@@ -333,7 +324,8 @@
 		{
 			"label": "@wordpress/* build-dependencies (always ahead of runtime-dependencies)",
 			"dependencies": [
-				"@wordpress/dependency-extraction-webpack-plugin"
+				"@wordpress/dependency-extraction-webpack-plugin",
+				"@wordpress/e2e-test-utils-playwright"
 			],
 			"packages": [
 				"**"

```

WunderBart: Ah, so we're not pinning exact version, just making sure the version is consistent across the monorepo?

WunderBart: Hmm how does this pin the wp-6.8 version, though?

kalessil: It'll pin to the 'next' version, not 'wp-6.8'.

> PS: if it won't work then at least let's relocate this section near the referenced one.

This PS admits that `next` might not work for your purpose, but it is worth trying.

WunderBart: I'll relocate for now. We have too many v6.8 issues to handle ATM, so I don't want to experiment with this, if that's OK.

kalessil: Sounds good!

gigitux (approved):
LGTM! 🚀

PR #56373, opened 2025-03-13
Add/56340 support install options

### LINKED ISSUE #56340: RemoteSpec: Support `install_options`
To enable adding or updating options before or after installing a plugin during the core profiler, introduce the install_options property, which uses the following format.

```
[
	{
		"name": "option name",
		"value": "option value",
		"options": {
			"unserialize": true,
			"force_array": true,
			"autoload": true,
			"install_priority": "before|after"
		}
	}
]
```

### Options

Name | Description
-- | --
unserialize | Unserializes the data before executing update_option. This is only needed for an edge case where the “value” cannot be expressed in JSON.
force_array | When set to true, the “value” will be decoded as an associative array.
autoload | The autoload value for the option.
install_priority | When set to "before", the option will be added before installing the plugin.

moon0326:

Closes #56340

This PR adds support for `install_options` property in the extensions feed.

install_options format:

```
{
  "install_options": [
    {
      "name": "option_name",
      "value": "option_value",
      "options": {
        "force_array": "true|false",
        "autoload": "true|false"
      }
    }
  ]
}
```

1. Introduces `woocommerce_rest_api_plugins_install_logger` [hook](https://github.com/woocommerce/woocommerce/pull/56373/files#diff-38f216066764e2afb70406832279e29873dc34008334384bfd79011a95188bbbR253) to support adding/replacing PluginsInstallLogger logger used in `install_plugins` function.
2. Added [ProcessInstallOptions](https://github.com/woocommerce/woocommerce/pull/56373/files#diff-3830b4e0d9608b202a3f130f4cf5d59aa49e8122dfc0cda687aba3f29c9d6848) logger that listens to plugin install events and process `install_options`
3. Plugin install REST APIs used in the core profiler now accept `source` arg to distinguish where the installation was initiated. This value is used in https://github.com/woocommerce/woocommerce/pull/56373/files#diff-a8a2daf3b77de4010391a59b0feb7262d48a40150f73fefa0f10dd3f66fdbbbfR130 to replace the logger.

1. Create a new JN site with this branch. Do not start the core profiler yet.
5. Navigate to `Plugins`.
6. Install and activate [free-extensions-test.zip](https://github.com/user-attachments/files/19236759/free-extensions-test.zip). This plugin replaces the core profiler feed with [this JSON](https://gist.github.com/moon0326/a7556cbd5d517a7a14516693d9ea76fa)
7. Start the core profiler.
8. On the extensions page, you should see only "WooCommerce Payments"
9. Click on `Continue` to install the plugin.
10. Navigate to `Tools -> WCA Test Helper`
11. Search for the options listed in https://gist.github.com/moon0326/a7556cbd5d517a7a14516693d9ea76fa#file-gistfile1-txt-L9 and compare the values.
12. ~Delete the options for the next test.~ Use the CLI to remove the options. See https://github.com/woocommerce/woocommerce/issues/56657

### Async install test

1. Using the same site, navigate to `Plugins` and delete `WooCommerce Payments`
2. Navigate to the core profiler and open your browser insepctor.
3. Paste `wp.data.dispatch('wc/admin/onboarding').installAndActivatePluginsAsync(['woocommerce-payments'], 'core-profiler');` and hit enter.
4. Navigate to `WooCommerce -> Status -> Scheduled Actions`
5. Search for `woocommerce_plugins_install_and_activate_async_callback`.
6. Wait until the job finishes.
10. Navigate to `Tools -> WCA Test Helper`
11. Search for the options listed in https://gist.github.com/moon0326/a7556cbd5d517a7a14516693d9ea76fa#file-gistfile1-txt-L9 and compare the values.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 
Added support for `install_options` property in the extensions feed.
</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessInstallOptions.php
```diff
@@ -0,0 +1,186 @@
...
+			if ( ! isset( $options->$key ) ) {
+				$options->$key = $value;
+			}
+		}
+
+		if ( $options->force_array ) {
+			$install_option->value = json_decode( wp_json_encode( $install_option->value ), true );
+		}
+
+		update_option( $install_option->name, $install_option->value, $options->autoload ? 'yes' : 'no' );
```
chihsuan: (Not sure if it's necessary.) Given that the remote spec originates from an external source, should we validate or restrict the option name and value before updating it, even if it's considered a trusted source?

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessInstallOptions.php
```diff
@@ -0,0 +1,186 @@
...
+
+		$options = $install_option->options ?? new \stdClass();
+		foreach ( $default_options as $key => $value ) {
+			if ( ! isset( $options->$key ) ) {
+				$options->$key = $value;
+			}
+		}
+
+		if ( $options->force_array ) {
+			$install_option->value = json_decode( wp_json_encode( $install_option->value ), true );
```
chihsuan: If value is incorrect JSON string, a fatal error will be thrown. Should we handle this edge case gracefully?

chihsuan (commented):
> Introduces woocommerce_rest_api_plugins_install_logger [hook](https://github.com/woocommerce/woocommerce/pull/56373/files#diff-38f216066764e2afb70406832279e29873dc34008334384bfd79011a95188bbbR253) to support adding/replacing PluginsInstallLogger logger used in install_plugins function.

Thanks for working on this! @moon0326 

The current implementation works! But for me, using the logger for option installation seems a bit confusing. I believe logging and option installation should be distinct responsibilities. I wonder if there is a better way to handle this?

moon0326:
> > Introduces woocommerce_rest_api_plugins_install_logger [hook](https://github.com/woocommerce/woocommerce/pull/56373/files#diff-38f216066764e2afb70406832279e29873dc34008334384bfd79011a95188bbbR253) to support adding/replacing PluginsInstallLogger logger used in install_plugins function.
> 
> Thanks for working on this! @moon0326
> 
> The current implementation works! But for me, using the logger for option installation seems a bit confusing. I believe logging and option installation should be distinct responsibilities. I wonder if there is a better way to handle this?

Thank you @chihsuan 

Yeah, we should either consider renaming it to something else or add hooks to implement it. I'll look into it.

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessInstallOptions.php
moon0326: @chihsuan Good question. I think there's a chance that the remote spec could overwrite existing options. This one is tricky. I'll look into it.

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessInstallOptions.php
moon0326: @chihsuan Thank you! I'll look into that.

moon0326:
@chihsuan ✋ 

I have some updates.

1. I've removed Logger classes. 
2. Added `woocommerce_plugins_install_before` and `woocommerce_plugins_install_after` hooks in [Add woocommerce_plugins_install_before and woocommerce_plugins_instal…](https://github.com/woocommerce/woocommerce/pull/56373/commits/fab0d988f597d5c6fd5a8a52d4519b5a5c8a2329)
3. Added JSON error handling in [Take care of JSON decode error](https://github.com/woocommerce/woocommerce/pull/56373/commits/9129848adde0458849a4677cba64ae0447516f53)

Regarding the allowed option list you mentioned in https://github.com/woocommerce/woocommerce/pull/56373#discussion_r1994672549, I still have to think about it a bit more.

plugins/woocommerce/src/Internal/Admin/Events.php
```diff
@@ -123,6 +127,23 @@ final public static function instance() {
...
+		);
+
+		add_action(
+			'woocommerce_plugins_install_after',
+			function ( $slug, $source ) {
+				$this->install_options_for_core_profiler_plugin_install( $slug, $source, 'after' );
+			},
+			10,
+			2
+		);
```
chihsuan: I was curious if there is a specific reason why these actions are being added here instead of in OnboardingSetup Wizard.php?

packages/js/data/src/onboarding/actions.ts
```diff
@@ -527,7 +527,8 @@ export function* actionTask( id: string ) {
 }
 
 export function* installAndActivatePluginsAsync(
-	plugins: Partial< PluginNames >[]
+	plugins: Partial< PluginNames >[],
+	source?: string
```
chihsuan, line 533: I think it would be beneficial to include some documentation/comments somewhere in the code to explain the usage of the "source".

chihsuan (commented):
> I have some updates.
> 
> 1. I've removed Logger classes.
> 2. Added `woocommerce_plugins_install_before` and `woocommerce_plugins_install_after` hooks in [Add woocommerce_plugins_install_before and woocommerce_plugins_instal…](https://github.com/woocommerce/woocommerce/pull/56373/commits/fab0d988f597d5c6fd5a8a52d4519b5a5c8a2329)
> 3. Added JSON error handling in [Take care of JSON decode error](https://github.com/woocommerce/woocommerce/pull/56373/commits/9129848adde0458849a4677cba64ae0447516f53)
> 
> Regarding the allowed option list you mentioned in [#56373 (comment)](https://github.com/woocommerce/woocommerce/pull/56373#discussion_r1994672549), I still have to think about it a b

Thanks @moon0326, the changes look great! Just some minor comments. 

I'll test the PR once it's ready.

plugins/woocommerce/src/Internal/Admin/Events.php
moon0326: @chihsuan Good catch! I completely forgot about OnboardingSetupWizard.php class 🤦 I'll move it to there.

moon0326: @chihsuan Fixed it in [Move woocommerce_plugins_install_before and woocommerce_plugins_insta…](https://github.com/woocommerce/woocommerce/pull/56373/commits/be71be75ce9b082361cc185cc82b48b3ae5017e8)

packages/js/data/src/onboarding/actions.ts
moon0326: @chihsuan Good catch! Fixed it in [Document `source](https://github.com/woocommerce/woocommerce/pull/56373/commits/5d5643887262e8272b6928f5992d7af3cbf4f328)

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessInstallOptions.php
moon0326: @adrianduffell @chihsuan  ✋ 

I'm thinking that we should only allow new options to be added with `before,` and `after` can only update those newly added options for security reasons. Does this work for our use case? What do you think?

Or..

We can also prevent updates to options with an option_id lower than 350 to ensure that plugins cannot modify options added by WordPress. A fresh WordPress installation without plugins typically includes 317 options.

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessCoreProfilerPluginInstallOptions.php
```diff
@@ -0,0 +1,164 @@
...
+
+		if ( $options->force_array ) {
+			$install_option->value = json_decode( wp_json_encode( $install_option->value ), true );
+			// In case of JSON error, return early.
+			if ( json_last_error() === JSON_ERROR_NONE ) {
+				return;
+			}
+		}
+
+		$this->update_option( $install_option->name, $install_option->value, $options->autoload ? 'yes' : 'no' );
```
adrianduffell: I think it would be preferable to default the autoload to `null`, so that it uses WP's new "maybe autoload" mechanism. ie it would defer the decision to WordPress https://make.wordpress.org/core/2024/06/18/options-api-disabling-autoload-for-large-options/

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessInstallOptions.php
adrianduffell: > I'm thinking that we should only allow new options to be added with before, and after can only update those newly added options for security reasons. Does this work for our use case? What do you think?

@moon0326 This sounds quite good to me, security-wise. It would limit the possible damage in a worst case scenario that WooCommerce.com is compromised. Attacker would only be able to create new options.

However I think it limits the usefulness of `after` 🤔 What use-cases would you see for it under this approach? That said, I'm not too sure `after` is 100% needed -- potentially we could just remove support for `before` / `after` and always perform the option creation before the plugin is installed. We would just have to work with partners to ensure the plugin checks the existence of the option and preserves the value when the plugin sets defaults.

moon0326: @adrianduffell Good point! Let's go with only `before`

I thought it would be helpful to have before/after concept early since changing RemoteSpec features after release is difficult. It seems we don't have any solid use cases for `after`. I'll remove the priority concept.

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessCoreProfilerPluginInstallOptions.php
moon0326: @adrianduffell Sounds good! TIL. Fixed it in [Option can only be added. It should not allow updating existing options](https://github.com/woocommerce/woocommerce/pull/56373/commits/7af128e41333800248580147c37800318dd3d1bf)

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessInstallOptions.php
moon0326: @adrianduffell @chihsuan I've made a few changes.

1. Removed `install_priority`. All options are added before installing a plugin.
2. Replaced `update_option` with `add_option` so that only new options can be added.

[Remove install_priority property -- we always process options before …](https://github.com/woocommerce/woocommerce/pull/56373/commits/11ca87302e3f00d7a23cefb66d2026b1de2eb53e)
[Option can only be added. It should not allow updating existing options](https://github.com/woocommerce/woocommerce/pull/56373/commits/7af128e41333800248580147c37800318dd3d1bf)

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessCoreProfilerPluginInstallOptions.php
```diff
@@ -0,0 +1,118 @@
...
+		foreach ( $default_options as $key => $value ) {
+			if ( ! isset( $options->$key ) ) {
+				$options->$key = $value;
+			}
+		}
+
+		if ( $options->force_array ) {
+			$install_option->value = json_decode( wp_json_encode( $install_option->value ), true );
+			// In case of JSON error, return early.
+			if ( json_last_error() === JSON_ERROR_NONE ) {
```
chihsuan: Should we use `!==` here instead? Looks like `JSON_ERROR_NONE` means no error has occurred.

```suggestion
			if ( json_last_error() !== JSON_ERROR_NONE ) {
```

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessCoreProfilerPluginInstallOptions.php
```diff
@@ -0,0 +1,118 @@
...
+	/**
+	 * Updates an option in the WordPress database.
+	 *
+	 * @param string $name Option name.
+	 * @param mixed  $value Option value.
+	 * @param string $autoload Autoload option.
+	 *
+	 * @return void
+	 */
+	protected function add_option( string $name, $value, $autoload = null ) {
```
chihsuan, line 156: I wonder if we can use add_option() directly instead of creating a method wrapper for it. 🤔

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessCoreProfilerPluginInstallOptions.php
chihsuan: Also, do we need to add some logs when some unexpected behavior occurs?

chihsuan (commented):
I've tested the PR and left some comments. We should fix the JSON error checking logic. Other than that, the PR looks good to me.


#### Test environment

<details>

- WordPress: 6.7.2
- Theme: Twenty Twenty-Five Version: 1.1
- WooCommerce: Development version with this fix-specific branch
- PHP: 8.3
- Web Server: Nginx
- Browser: Brave
- OS: macOS Sequoia 15.3.1
</details>

#### Review checklist

| Parameter | Result |
| Approach | ✅ All good |
| Implementation | 👍 Almost good (left a comment) |
| Automated tests | ✅ All good |
| Naming | ✅ All good |
| Typing | ✅ All good |
| Code comments | ✅ All good |
| Changelog | ✅ All good |
| Documentation | ✅ All good |
| PR description | ✅ All good |
| Testing instructions | ✅ All good |
| Manual testing | ⚠️ Partial success |

#### Scenarios tested

| Scenario | Result |
| Extension install during Core Profiler with install_options | ⚠️  Options properly set except for array values |
| Async plugin installation with install_options | ⚠️  Options properly set except for array values |

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessCoreProfilerPluginInstallOptions.php
moon0326: @chihsuan Good catch! Fixed it in [Fix json error handling](https://github.com/woocommerce/woocommerce/pull/56373/commits/16171ae215b514e987af008f624e9ce7e04059cf) (Added logging as well)

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessCoreProfilerPluginInstallOptions.php
moon0326: @chihsuan Yeah..I don't really like it either, but it is needed for unit testing as `add_option` is a global function. It's tricky to properly mock it without a 3rd party composer package.

plugins/woocommerce/src/Admin/PluginsHelper.php
```diff
@@ -317,6 +318,13 @@ public static function install_plugins( $plugins, ?PluginsInstallLogger $logger
 				continue;
 			}
 
+			/**
+			 * Action triggered before a plugin is installed.
+			 *
+			 * @since 9.9
```
ilyasfoo: Since this is now targeting 9.8
```suggestion
			 * @since 9.8
```

plugins/woocommerce/src/Admin/PluginsHelper.php
```diff
@@ -369,6 +377,13 @@ public static function install_plugins( $plugins, ?PluginsInstallLogger $logger
 
 			$installed_plugins[] = $plugin;
 			$logger && $logger->installed( $plugin, $time[ $plugin ] );
+
+			/**
+			 * Action triggered after a plugin is installed.
+			 *
+			 * @since 9.9
```
ilyasfoo:
```suggestion
			 * @since 9.8
```

plugins/woocommerce/src/Internal/Admin/Events.php
```diff
@@ -8,6 +8,8 @@
 defined( 'ABSPATH' ) || exit;
 
 use Automattic\WooCommerce\Admin\Features\Features;
+use Automattic\WooCommerce\Admin\PluginsInstallLoggers\AsyncPluginsInstallLogger;
```
ilyasfoo: These imports aren't used in the file. Are they required?

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessCoreProfilerPluginInstallOptions.php
```diff
@@ -0,0 +1,130 @@
...
+	 *
+	 * @param object $install_option Install option object.
+	 */
+	protected function add_install_option( object $install_option ) {
+		$default_options = array(
+			'force_array' => false,
+			'autoload'    => false,
+		);
+
+		$options = $install_option->options ?? new \stdClass();
```
ilyasfoo: I think we need to ensure `$install_option->options` is an object here

ilyasfoo (commented):
This tests well, but I feel adding an option with an arbitrary name exposes merchants to some security risk. In the case of a network-level compromise, an attacker could override the specs. 

I think the simplest way to reduce the attack surface even more is to have a list of disallowed options (from ChatGPT):

```
$sensitive_options = array(
    'siteurl',              // The URL to your WordPress installation.
    'home',                 // The home URL of the site.
    'admin_email',          // Administrator email address.
    'wp_user_roles',        // Serialized roles and capabilities.
    'active_plugins',       // List of active plugins.
    'template',             // The current theme template.
    'stylesheet',           // The current theme stylesheet.
    'default_role',         // Default role for new users.
    'ftp_hostname',         // FTP server hostname.
    'ftp_username',         // FTP server username.
    'ftp_password',         // FTP server password.
    'ftp_port',             // FTP server port.
    'ftp_ssl',              // Whether to use FTP over SSL.
    'ftp_pasv',             // Whether to use passive FTP.
    'rewrite_rules',        // URL rewrite rules.
    'permalink_structure',  // Structure of permalinks.
    'cron',                 // Scheduled tasks (WP-Cron jobs).
    'upload_path',          // Filesystem path for uploads.
    'upload_url_path',      // URL path for uploads.
    'mailserver_url',       // Mail server hostname.
    'mailserver_login',     // Mail server login.
    'mailserver_pass',      // Mail server password.
    'mailserver_port'       // Mail server port.
);
```

Let me know what you think!

adrianduffell:
> I think the simplest way to reduce the attack surface even more is to have a list of disallowed options

I think that's good call! I can't see there being any use-case for this to change any of the WordPress options.

moon0326:
> > I think the simplest way to reduce the attack surface even more is to have a list of disallowed options
> 
> I think that's good call! I can't see there being any use-case for this to change any of the WordPress options.

The current implementation does not allow any modification to the existing options. The default options that come with WP are safe. However, I think disallowed list is a nice addition. I'll work on it 👍

moon0326:
@adrianduffell @ilyasfoo Added disallowed options in [Make the logger optional -- uses wc_get_logger if not provided](https://github.com/woocommerce/woocommerce/pull/56373/commits/90237eb35a1488e2aeaaf5812533e4a9be5df6b6) and [Add test case for disallowed option](https://github.com/woocommerce/woocommerce/pull/56373/commits/a572da6cdc9c6b4eb84f2e4ee36cd9ef498c064b)

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessCoreProfilerPluginInstallOptions.php
moon0326: Updated it in [Make sure options is an object](https://github.com/woocommerce/woocommerce/pull/56373/commits/a93945c8ffed3cc630e51448f0f03bb7259e2fcf).

plugins/woocommerce/src/Internal/Admin/Events.php
moon0326: @ilyasfoo good catch! Updated it in [Remove unused import](https://github.com/woocommerce/woocommerce/pull/56373/commits/54dfc264af638f2687cdb807eddc0bd3d5264cbd)

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessCoreProfilerPluginInstallOptions.php
```diff
@@ -0,0 +1,163 @@
+<?php
+
+declare( strict_types = 1);
```
ilyasfoo: Nit
```suggestion
declare( strict_types = 1 );
```

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessCoreProfilerPluginInstallOptions.php
```diff
@@ -0,0 +1,163 @@
...
+	 *
+	 * @return void
+	 */
+	protected function add_option( string $name, $value, $autoload = null ) {
+		if ( in_array( $name, self::DISALLOWED_OPTIONS, true ) ) {
+			$this->logger && $this->logger->error( 'Disallowed option: ' . $name );
+			return;
+		}
+		// phpcs:ignore WordPress.WP.DeprecatedParameters.Add_optionParam3Found
+		add_option( $name, $value, $autoload );
```
ilyasfoo: The third argument is actually deprecated, forth is autoload: https://developer.wordpress.org/reference/functions/add_option/

plugins/woocommerce/src/Admin/PluginsHelper.php
```diff
@@ -317,6 +318,13 @@ public static function install_plugins( $plugins, ?PluginsInstallLogger $logger
 				continue;
 			}
 
+			/**
+			 * Action triggered before a plugin is installed.
+			 *
+			 * @since 9.8
+			 */
+			do_action( 'woocommerce_plugins_install_before', $slug, $source );
```
ilyasfoo, line 326: Since we're adding the option only when installing a plugin, we will not support pre-installed plugins for certain hosts. Do we need to consider on activate? cc @adrianduffell

ilyasfoo (changes requested):
Thanks for addressing, @moon0326! I found a couple of issues in the following comments:

#### Review checklist

| Parameter  | Result |
| Approach  | ⚠️ Commented on pre-installed extensions use case |
| Implementation  | ⚠️ Commented on autoload value |
| Automated tests  | ✅ All good |
| Changelog  | ✅ All good |
| Documentation  | Not applicable |
| Addresses issue criteria  | ✅ All good |

#### Scenarios tested

| Adding options on core profiler extensions step  | ⚠️ Autoload not set properly |  |
| Adding options on `installAndActivatePluginsAsync`  | ⚠️ Autoload not set properly |  |
| Adding disallowed options using [custom extension](https://github.com/user-attachments/files/19440890/extensions-test-ftppassword-option.zip) | ✅ Works as expected | image  |

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessCoreProfilerPluginInstallOptions.php
moon0326: @ilyasfoo Thank you!

Updated it in [Fix add_option argument order](https://github.com/woocommerce/woocommerce/pull/56373/commits/6b30619e6edb17ee938c67f6a437e6d6126fb9b1)

adrianduffell (approved):
Nice, thanks for working on this @moon0326 and making the recent tweaks. Tested well per testing instructions and LGTM 🚀 

Screen Shot 2025-03-25 at 10 27 32 am

Also tested the following screnarios:

🆗 Plugins without `install_options` field will install without issue.
🆗 When a plugin is pre-installed, the option is not added.
🆗 When WooPayments is installed from task, the option is not added.

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessCoreProfilerPluginInstallOptions.php
```diff
@@ -0,0 +1,163 @@
...
+		if ( $options->force_array ) {
+			$install_option->value = json_decode( wp_json_encode( $install_option->value ), true );
+			// In case of JSON error, return early.
+			if ( json_last_error() !== JSON_ERROR_NONE ) {
+				$this->logger && $this->logger->error( 'Failed to decode JSON for install option value for ' . $install_option->name . ': ' . json_last_error_msg() );
+				return;
+			}
+		}
+
+		$this->add_option( $install_option->name, $install_option->value, $options->autoload ? 'yes' : null );
```
ilyasfoo: Autoload will never have a `no` value. Is this expected?

ilyasfoo (approved):
Nice, thanks! One last comment but pre-approving

| Parameter  | Result |
| Approach  | ✅  Not supporting on activate is expected (p1742872879657049/1742780270.580589-slack-C01SFMVEYAK)  |

#### Scenarios tested

| Adding options on core profiler extensions step  | ✅ Works as expected | image |
| Adding options on `installAndActivatePluginsAsync`  | ✅ Works as expected |  |

plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/ProcessCoreProfilerPluginInstallOptions.php
moon0326: @ilyasfoo You're correct. It's either yes or null. null now means "maybe autoload" according to https://make.wordpress.org/core/2024/06/18/options-api-disabling-autoload-for-large-options/

moon0326: @ilyasfoo I've made a quick change in [Allow autoload value no](https://github.com/woocommerce/woocommerce/pull/56373/commits/073547f6f659796244d99c8c8ac432937511817c)

~Update: Actually, we should also allow true and false.~

Allowing true and false as well in [Allow true and false](https://github.com/woocommerce/woocommerce/pull/56373/commits/048326d2075cbb07019c9b924345bc13ad40d080)

ilyasfoo: Thanks, @moon0326. 

Nit: Based on [the docs](https://developer.wordpress.org/reference/functions/add_option/), it mentioned the proper value would be boolean. I think it'll be good to normalize all to either boolean or null. What do you think?

> Accepts a boolean, or null to leave the decision up to default heuristics in WordPress. For backward compatibility 'yes' and 'no' are also accepted, though using these values is deprecated.

moon0326: @ilyasfoo Good call! Updated it in [Fix autoload value](https://github.com/woocommerce/woocommerce/pull/56373/commits/68ed0e5160fa77d93e3fd596ff67a9334d832e04)

ilyasfoo: Nice, LGTM! Thanks @moon0326

PR #56101, opened 2025-03-05
Add AJAX add to cart functionality for product pages

nielslange:
### Changes proposed in this Pull Request:

Closes WOOPRD-9.
Closes #56017.
Closes #36006. / Closes WOOPLUG-984.
Closes #42638. / Closes WOOPLUG-657.

This PR introduces the following changes:

- Add AJAX add to cart functionality for product pages.
- Introduce a new JavaScript file for AJAX add to cart handling.
- Add a new setting `woocommerce_enable_ajax_add_to_cart_product_pages` to toggle this feature. For now, this setting is false by default (see discussion in https://github.com/woocommerce/woocommerce/discussions/57955).
- Exclude external products from AJAX functionality.

These changes enable us to open the Mini-Cart drawer automatically, when a shopper adds a product to their cart while being on a single product page.

### Screenshots or screen recordings:

https://github.com/user-attachments/assets/c3ce1f92-6ce5-4b27-b0ca-56774b42c472

### How to test the changes in this Pull Request:

**ZIPs to test the changes:**
1. Based on this branch - option **disabled** by default: [woocommerce.zip](https://github.com/user-attachments/files/21663257/woocommerce.zip)
2. Based on [temporary branch](https://github.com/woocommerce/woocommerce/pull/60251) - option **enabled** by default. At first the feature will be released as disabled by default but this is to test the further steps: [woocommerce.zip](https://github.com/user-attachments/files/21663239/woocommerce.zip)

---

1. Go to `/wp-admin/admin.php?page=wc-settings&tab=products`.
2. Enable the setting `Enable AJAX add to cart buttons on product pages`.

##### Testing a simple product

1. Go to the product page of a simple product.
2. Click the `Add to Cart` button.
5. Verify the product was added to the cart and the page didn't reload.

##### Testing a variable product

1. Go to the product page of a variable product.
2. Click the `Add to Cart` button.
3. Verify the product was added to the cart and the page didn't reload.
6. Modify one of the attributes and verify the button text changes back from "X in cart" to "Add to cart".

##### Testing a grouped product

1. Go to the product page of a grouped product.
2. Select at least two products.
3. Click the `Add to Cart` button.
3. Verify the products were added to the cart and the page didn't reload.

##### Testing an external/affiliate product

1. Go to the product page of an external/affiliate product.
2. Click the `Add to Cart` button.
3. Verify a redirect to the external product happens.

#### Block theme

1. Go to `Apperance » Editor » Patterns » Header » Header`.
[...]

PR #56378, opened 2025-03-13
[Experimental] Blockified Add to Cart with Options block: make it possible to add variable products to cart

### LINKED ISSUE #55698: Blockified Add to Cart with Options block: make it possible to add variable products to cart
* Blocked by https://github.com/woocommerce/woocommerce/issues/55696.

Similar to https://github.com/woocommerce/woocommerce/issues/54416 but for variable products.

Once we have all the inner blocks of the blockified Add to Cart with Options block ready, we will need to make them actually work. This issue is focused exclusively on variable products.

As an outcome, it should be possible to add the blockified _Add to Cart with Options_ block in the Single Product template (or the Single Product block when a variable product is selected) and, from the frontend, add to cart without causing a page reload.

mdperez86:

Closes #55698

| Before | After |
| ------ | ----- |
|        |       |

6. Then click `Save` to publish the changes.
7. Go to the store and click on any variable product or create a new one if there are not any.
8. Select the attributes and add a quantity from the form.
9. Click add to cart.
10. Check that the correct variation and quantity is added to the cart.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/base/stores/cart-items.ts
```diff
@@ -8,10 +8,16 @@ import { store } from '@wordpress/interactivity';
  */
 import { triggerAddedToCartEvent } from './legacy-events';
 
+type Variation = {
+	attribute: string;
+	value: string;
+};
```
Aljullu: Wondering if we could unify this type...

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
```diff
@@ -1,9 +1,50 @@
 /**
  * External dependencies
  */
-import { store } from '@wordpress/interactivity';
+import { store, getContext } from '@wordpress/interactivity';
+
+type Attribute = {
+	attribute: string;
+	value: string;
+};
```
Aljullu: ... with this one. Or are they intentionally different?

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelector.php
```diff
@@ -308,7 +45,20 @@ protected function render( $attributes, $content, $block ): string {
...
+			if ( empty( $variation_attributes ) ) {
+				return '';
+			}
+
+			$variations = $this->get_variations_data( $product );
+			if ( empty( $variations ) ) {
+				return '';
+			}
+
+			add_filter( 'woocommerce_product_supports', array( $this, 'check_product_supports' ), 10, 3 );
```
Aljullu, line 59: Not a strong opinion, but if this filter is added to ensure the Product Button knows whether the product can be added to the cart without a page refresh, I’d lean towards modifying the Product Button render method instead.

So instead of checking `$product->supports( 'ajax_add_to_cart' )`, we could assume that adding to the cart without a page refresh is always possible when inside the blockified "Add to Cart with Options" block. What do you think?

Aljullu (commented):
Really exciting to see this working! :raised_hands:

I see the _Quantity_ is still not working properly, if I try to add more than one item, only one gets added. I guess that's a known issue for now. Besides that, I left two minor comments, I would like to re-review once this is rebased on `trunk`, but wanted to give early feedback to help keep things moving.

plugins/woocommerce/client/blocks/assets/js/base/stores/cart-items.ts
```diff
@@ -8,10 +8,16 @@ import { store } from '@wordpress/interactivity';
...
+
 type Item = {
 	key?: string;
 	id: number;
 	quantity: number;
+	variation: Variation[] | undefined;
```
samueljseay: Just a question, is this intentional that an item could have variation value of undefined? Otherwise I think it'd make more sense to be variation?: Variation[]

also why is it named `variation` if it holds an array of variations. should it be `variations`? or `variationAttributes`

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
```diff
@@ -1,9 +1,50 @@
...
+
+type Context = {
+	variation: Attribute[];
```
samueljseay: Nitpick: This var name seems confusing, should it be `variationAttributes`? it's just feeling a bit unclear right now.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
```diff
@@ -1,9 +1,50 @@
...
+};
 
 store( 'woocommerce/add-to-cart-with-options', {
-	state: {},
-	actions: {},
+	state: {
+		get variation() {
+			const { variation } = getContext< Context >();
```
samueljseay: Is there a need for this getter? what about having the context default to `[]`?

samueljseay:
@mdperez86 Just mentioning, that typically for PRs, we expect that you would write some detailed tests instructions to manually test the functionality, either to confirm that it hasn't introduced regressions or that it fixes bugs or behaviours intended to be fixed.

It's especially helpful for outside reviewers to be able to explore what you've implemented and get some context.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
```diff
@@ -1,9 +1,185 @@
...
+const { actions: wooAddToCartWithOptions } = store(
+	'woocommerce/add-to-cart-with-options',
+	{},
+	{
+		// Stores are locked to prevent 3PD usage until the API is stable.
+		lock: 'I acknowledge that using a private store means my plugin will inevitably break on the next store release.',
+	}
+);
+
+function setAttribute( name: string, value: string ) {
```
samueljseay: The type should be `( name: string, value?: string )`

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
```diff
@@ -1,9 +1,185 @@
...
+				}
+
+				context.tabIndex = -1;
+			},
+		},
+	}
+);
+
+type DropdownContext = Context & {
+	isSelected: 'selected' | undefined;
```
samueljseay: Why is this not a boolean?

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
```diff
@@ -1,9 +1,185 @@
...
+type Context = {
+	name: string;
+	selected: string | null;
+	option: Option;
+	options: Option[];
+};
+
+type PillsContext = Context & {
+	tabIndex?: number;
+	selected?: string;
```
samueljseay: This should be `selectedValue` I think?

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
```diff
@@ -1,9 +1,50 @@
...
+type Attribute = {
+	attribute: string;
+	value: string;
+};
+
+type Context = {
+	variation: Attribute[];
+};
 
 store( 'woocommerce/add-to-cart-with-options', {
```
samueljseay: If you change this line to capture the value then you can export a type like:

```
const addToCartWithOptionsStore = store(...
...

export type AddToCartWithOptionsStore = typeof addToCartWithOptionsStore;
```

Then when you call it later such as [here](https://github.com/woocommerce/woocommerce/pull/56378/files#diff-baaa04fb4f93cdad22fe635ee17fdc572b9878c182c6cab5abcd5f33e29f19dbR26)

you can pass it to the store as the generic arg: `store< AddToCartWithOptionsStore >` which will properly type the actions

samueljseay (changes requested):
Overall this looks good, The thing I'd like to see the most is the variables named more concisely as I've suggested because it took me a while to grok the intent where some variable names are vague. Small thing but it helps when someone new comes to take this over later.

mdperez86:
> @mdperez86 Just mentioning, that typically for PRs, we expect that you would write some detailed tests instructions to manually test the functionality, either to confirm that it hasn't introduced regressions or that it fixes bugs or behaviours intended to be fixed.
> 
> It's especially helpful for outside reviewers to be able to explore what you've implemented and get some context.

Yes, that is correct. As you can see this PR is yet in progress so I'll add the testing instructions as soon as it's ready to be tested or reviewed. I know I could create this PR as a Draft as well, but I relay on the Project Status instead.

mdperez86:
> Really exciting to see this working! 🙌
> 
> I see the _Quantity_ is still not working properly, if I try to add more than one item, only one gets added. I guess that's a known issue for now. Besides that, I left two minor comments, I would like to re-review once this is rebased on `trunk`, but wanted to give early feedback to help keep things moving.

No, I did not touch the quantity selector yet.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelector.php
mdperez86: So are we going to not consider this setting for this case?
image

plugins/woocommerce/client/blocks/assets/js/base/stores/cart-items.ts
mdperez86: >Just a question, is this intentional that an item could have variation value of undefined? Otherwise I think it'd make more sense to be variation?: Variation[]

I had a type check error with that notation so I changed it to `variation: Variation[] | undefined;` I did not research about why. But I do agree with you.

>also why is it named variation if it holds an array of variations. should it be variations? or variationAttributes

Very aligned with you, but this is because of the add to cart API contract definition. So in order to change that I'll have to change the API as well which runs out of the scope of this PR.

mdperez86: Maybe the thinking behind the `variation` name is that the array of attributes assigned to it represents one  product variation.

Perhaps what I did wrong here was to use the `Variation` name for the type instead of using `VariationAttribute` so this `variation: Variation[]` could be replaced by `variation: VariationAttribute[]`.

But in any case I do agree with you that it's confusing specially because `variation` is treated as an array and not as an object which was what I should expect.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
mdperez86: According to this https://github.com/woocommerce/woocommerce/pull/56378#discussion_r2003503954
Do you thing `variation: VariationAttribute[]` could make sense?

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
mdperez86: Yes it could be, but the value is never undefined but can be empty string.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
mdperez86: I don't like creating a type by inference since it is not descriptive. But in this case it make sense.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelector.php
Aljullu: My concern with `$product->supports( 'ajax_add_to_cart' )` is that so far it has been used in the context of product grids. Only simple products have it set to `true`. So I think in the past it has been used more like "can be added with AJAX from the product grid" instead of "can be added with AJAX in general". That's why I'm not 100% convinced to use it in the context of the _Add to Cart with Options_ block. Having said that, not a strong opinion, and it's something that we can change later on, so it shouldn't be a blocker.

Regarding the settings:

- Redirect to the cart page after successful addition: I agree we should support this one.
- Enable AJAX add to cart buttons on archives: IMO we shouldn't support this one, the _Add to Cart with Options_ block is not an archive.
- Enable AJAX add to cart buttons on product pages: we are looking into introducing this new setting in https://github.com/woocommerce/woocommerce/pull/56101. It will be nice to support it, but not something we should worry about now.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
samueljseay: As soon as you name the type it becomes descriptive :)

samueljseay (commented):
The code looks good to me :) 

During testing I saw confusing UX behaviour. Your testing steps don't cover it so I wanted to check if it's ok.

When I add items to cart it does indeed add the right amount of variation, but it navigates to site root and it feels super weird:



https://github.com/user-attachments/assets/7537b962-ce4d-4918-82c2-c0341e99d8a5

Aljullu (approved):

samueljseay (approved):

PR #56380, opened 2025-03-13
Refresh order counts on background hook

### LINKED ISSUE #54819: [Performance]: Create reusable interface for background indexing

joshuatf:

This PR implements order count cache refreshes on the new AS background hook introduced in https://github.com/woocommerce/action-scheduler/pull/1260.

Closes #54819 .

Screenshot 2025-04-17 at 5 33 03 PM

1. Clone and install/activate AS from `trunk` - https://github.com/woocommerce/action-scheduler
2. Navigate to Tools -> Scheduled Actions
3. Hover over the `action_scheduler_run_recurring_actions_schedule_hook` action and click "Run"
4. Make sure that there is now a pending action for `woocommerce_refresh_order_count_cache ` with the argument `shop_order` (and any other custom order types installed) scheduled 12 hours from now
5. Run that action and make sure that it completes without error in the completed actions log

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

## Summary by CodeRabbit

- **New Features**
  - Order count queries are now refreshed in the background, reducing immediate load and improving performance.
  - Automated periodic refreshes of order counts now occur every 12 hours.

- **Tests**
  - Added tests to verify that background scheduling for order count refreshes is functioning correctly.

- **Documentation**
  - Updated changelog to reflect the performance improvement in order count processing.


plugins/woocommerce/src/Caches/OrderCountCacheService.php
```diff
@@ -31,6 +33,38 @@ final public function init() {
...
+	public function register_background_actions() {
+		if ( ! function_exists( 'as_schedule_recurring_action' ) ) {
+			return;
+		}
+
+		$order_types = wc_get_order_types( 'order-count' );
+		foreach ( $order_types as $order_type ) {
+			if ( false === as_has_scheduled_action( 'woocommerce_order_count' ) ) {
+				$frequency = HOUR_IN_SECONDS * 12;
+				as_schedule_recurring_action( time() + $frequency, $frequency, 'woocommerce_order_count', array( $order_type ) );
```
joshuatf: A small "gotcha" with this: we need to set the first time to `time() + $frequency` since the scheduled recurring action does not get detected inside the hook.

If a consumer forgets to add this, the action will end up running every hour when the background scheduler hook runs.

plugins/woocommerce/src/Caches/OrderCountCacheService.php
```diff
@@ -31,6 +33,38 @@ final public function init() {
 		add_action( 'woocommerce_order_status_changed', array( $this, 'update_on_order_status_changed' ), 10, 4 );
 		add_action( 'woocommerce_before_trash_order', array( $this, 'update_on_order_trashed' ), 10, 2 );
 		add_action( 'woocommerce_before_delete_order', array( $this, 'update_on_order_deleted' ), 10, 2 );
+		add_action( 'woocommerce_order_count', array( $this, 'refresh_cache' ) );
```
joshuatf: It could just be me, but this took significantly longer on the consumer side to set up and test these hooks.  While the API in this PR is significantly simpler than https://github.com/woocommerce/woocommerce/pull/55746, this is the trade-off we need to consider.

plugins/woocommerce/src/Admin/Schedulers/BackgroundScheduler.php
```diff
@@ -0,0 +1,26 @@
...
+	/**
+	 * Action scheduler hook name.
+	 */
+	const HOOK_NAME = 'woocommerce_background_scheduler';
+
+	/**
+	 * Initialize the background scheduler.
+	 */
+	public function init() {
+		if ( function_exists( 'as_schedule_recurring_action' ) && false === as_has_scheduled_action( self::HOOK_NAME ) ) {
```
prettyboymp: My understanding is that we need to check that `as_has_scheduled_action()` exists as well and if not, fallback to `as_next_scheduled_action()`.  `as_has_scheduled_action()` was added in AS 3.3 and even though Woo includes it, there are edge cases where an older version is loaded.  See https://developer.woocommerce.com/2021/10/12/best-practices-for-deconflicting-different-versions-of-action-scheduler/

I'm not sure how long we need to support this.

prettyboymp: Unless we're able to implement caching in ActionScheduler (ref https://github.com/woocommerce/action-scheduler/issues/1240), we may want to cache the results here to make sure we're not adding that query to check on every page load.

plugins/woocommerce/src/Caches/OrderCountCacheService.php
```diff
@@ -31,6 +33,38 @@ final public function init() {
...
+	 * @return void
+	 */
+	public function register_background_actions() {
+		if ( ! function_exists( 'as_schedule_recurring_action' ) ) {
+			return;
+		}
+
+		$order_types = wc_get_order_types( 'order-count' );
+		foreach ( $order_types as $order_type ) {
+			if ( false === as_has_scheduled_action( 'woocommerce_order_count' ) ) {
```
prettyboymp: Won't this need to be a order_type specific check?

plugins/woocommerce/src/Caches/OrderCountCacheService.php
prettyboymp: I'm not sure I understand.  Setting the first time to `time()` will just run the action immediately the next time action scheduler runs, correct?  After that it should get rescheduled in `$frequency` seconds.  Or does `as_schedule_recurring_action()` work different than expected?

plugins/woocommerce/src/Caches/OrderCountCacheService.php
prettyboymp: It may be helpful to rename the hook `woocommerce_order_count` to something that make it more clear that it is refreshing the cache.

prettyboymp: While I prefer the more light weight implementation here and think it is more direct, I understand the concern.  

In the end, I would accept either approach.

plugins/woocommerce/src/Admin/Schedulers/BackgroundScheduler.php
joshuatf: Given our WP policy goes back ~1 year and this change occurred almost 4 years ago (and we bundle AS), I'm wondering if we can forego this check going forward.

plugins/woocommerce/src/Caches/OrderCountCacheService.php
joshuatf: Ah, with the argument?  Good call.  Should we switch over to using an order type in the hook name?  I recall you mentioning that checking existence of hooks with arguments is less performant (though maybe this is less of an issue since it's in the background).

plugins/woocommerce/src/Caches/OrderCountCacheService.php
joshuatf: I need to do a deeper dive into this to understand why if we go with this PR's approach, but I think what might be happening is:

1. The WC background scheduler hook runs
2. This schedule recurring action function runs and sets up an immediate (next async?) action
3. The scheduled action completes, however there is no call to `as_schedule_recurring_action` as this only runs in the background
4. On the next background scheduler hook, `as_schedule_recurring_action` runs again, but since the action has now completed and we're outside of the original scheduling, it immediately runs again believing it has not been previously scheduled.

joshuatf: Confirmed that this runs similar to what I expected above.  `as_schedule_recurring_action` is always looking to schedule an action at the specified time (typically `time()` now).  The recurring functions [check if there are running or pending hooks](https://github.com/woocommerce/action-scheduler/blob/d0a539eaab96430ff59eb70168d483b961e63d21/classes/ActionScheduler_ActionFactory.php#L179-L202) before scheduling the next time.

Because `as_schedule_recurring_action` only occurs inside of the background hook, this function is not being run when the action itself is being run.  It only runs on the next background hook when the action has already completed.  On the next background hook, AS will then look to see if there are any pending or running actions and since there are not, it will immediately schedule the next action.

plugins/woocommerce/src/Admin/Schedulers/BackgroundScheduler.php
joshuatf: I ended up removing these checks based on the above.  Let me know if you see any issues with that and we can discuss.

joshuatf: After looking at this more and discussing in Slack, I think we should keep some of these changes upstream in AS.  If we attempt to cache this here, the checks that help `as_schedule_recurring_action` create recurring actions will not work correctly since we're not consistently calling this function.

plugins/woocommerce/src/Caches/OrderCountCacheService.php
joshuatf: Ended up removing this check and letting AS handle this upstream by adding `true` for the `$unique` property.

plugins/woocommerce/src/Caches/OrderCountCacheService.php
joshuatf: Renamed this to `woocommerce_refresh_order_count_cache`.

joshuatf:
@prettyboymp Updated this to use the new hook introduced in https://github.com/woocommerce/action-scheduler.  This PR now just refreshes order count caches.

plugins/woocommerce/tests/php/src/Caching/OrderCountCacheServiceTest.php
```diff
@@ -135,4 +135,12 @@ public function test_count_on_multiple_status_changes() {
 		$this->assertEquals( $initial_count[ OrderInternalStatus::COMPLETED ], $count[ OrderInternalStatus::COMPLETED ] );
 		$this->assertEquals( $initial_count[ OrderInternalStatus::CANCELLED ] + 1, $count[ OrderInternalStatus::CANCELLED ] );
 	}
+
+	/**
+	 * Test that background actions are scheduled.
+	 */
+	public function test_background_actions_scheduled() {
+		$this->order_cache->schedule_background_actions();
```
prettyboymp: I think this needs to be the service instead of the cache instance.

prettyboymp (commented):
The changes look good.  We'll probably want to wait to merge until the next release of ActionScheduler is shipped and packaged with WooCommerce.   Though, it technically wouldn't break anything if this was included first - caches would just expire as they do now.

joshuatf:
I'd be in favor of merging this before the release since the new hook is already merged.  The refresh gracefully degrades and is harmless if the hook doesn't exist.

Leaving this PR open leaves us exposed to potential rebases and then needing to retest and pick up context around this PR again.

plugins/woocommerce/src/Caches/OrderCountCacheService.php
```diff
@@ -47,6 +47,32 @@ final public function init() {
...
+
+	/**
+	 * Refresh the cache for a given order type.
+	 *
+	 * @param string $order_type The order type.
+	 * @return void
+	 */
+	public function refresh_cache( $order_type ) {
+		$this->order_count_cache->remove( $order_type );
+		OrderUtil::get_count_for_type( $order_type );
```
prettyboymp, line 62: I think this is good enough as is for now, especially for an admin based cache.  But one thing we may want to consider around this pattern going forward is updating the cached value without deleting.  This will be important for heavy volume/load sites where requests may come in in the time in between deleting the cached value and generating the new value where all of those request end up stampeding the database attempting recalculate the counts at once because of the temporary cache miss.

prettyboymp (approved):
@joshuatf , I'm good with merging this how it is knowing that it is usable how it is and will just improve automatically in future releases.

prettyboymp:
@coderabbitai review

PR #56387, opened 2025-03-14
[e2e tests] Stablize product-tags-attributes spec

adimoldovan:

Tests in product-tags-attributes spec stared failing with `page.goto: net::ERR_ABORTED`. A goto action is performed before the previous page load triggered by a save/update button click was completed.

Fixed by making sure we wait for the initial save/update to complete. 
Other changes:
- used the restApi fixture instead of request to handle the test page
- the failing test required woocommerce_attribute_lookup_enabled setting to be enabled, but enabling it via the API was not enough, so a workaround to also change a random setting via the UI was performed. Instead of that, I changed the test to update via the UI the actual setting the test requires.

```
pnpm test:e2e product-tags-attributes.spec.js
```
Note: CYS failures with WP6.8 beta2 can be ignored.

### DIFF
--- a/plugins/woocommerce/changelog/e2e-fix-product-tag-attributes-flakyness
+++ b/plugins/woocommerce/changelog/e2e-fix-product-tag-attributes-flakyness
@@ -1,0 +1,4 @@
+Significance: patch
+Type: dev
+
+E2E tests: stabilize product-tags-attributes spec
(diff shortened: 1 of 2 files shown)


Stojdza (approved):
LGTM 👍

PR #56390, opened 2025-03-14
[Experimental] Don't display the Add to Cart with Options form for out of stock products and not-purchasable simple products

### LINKED ISSUE #56371: Blockified Add to Cart with Options block: don't display the form for out of stock products
In classic themes, when a product is out of stock, WooCommerce doesn't display the Add to Cart with Options form:

Image

However, in the blockified Add to Cart with Options block, we still display the form:

Image

This issue is about making the new blockified Add to Cart with Options block behave like the PHP templates, so the form isn't rendered for out of stock products.

Aljullu:

This PR makes it so we don't display the Add to Cart with Options template part when products are out of stock, or are not-purchasable simple products, matching the behavior of the PHP templates.

Closes https://github.com/woocommerce/woocommerce/issues/56371.

0. Install and activate the WooCommerce Beta Tester plugin.
0. Go to _Tools_ > _WCA Helper_ > _Features_ and enable `experimental-blocks` and `blockified-add-to-cart`.
1. Create a post or page, add the Single Product block and choose a simple product. (Alternatively, edit the Single Product template)
2. Update the _Add to Cart with Options_ block to the blockified version.
3. In the frontend, verify the Add to Cart form is displayed.

imatge

4. Now, edit the product and set it out of stock.
5. In the frontend, verify the Add to Cart form is not visible and instead an _Out of stock_ text is shown.

imatge

6. Edit the product again and remove its price, making it not-purchasable.
7. In the frontend, verify the Add to Cart form is not visible and the _Out of stock_ text is not visible either.

imatge

8. Edit the product again, add the price again and set is as on backorder.
9. In the frontend, verify the Add to Cart form displays an _On backorder_ text before the form.

imatge

10. Now create or edit a Grouped Product block. Set one of its child products as out of stock or on backorder. Verify the stock status is displayed below the price.

imatge

### DIFF
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-template/constants.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-template/constants.ts
@@ -3,6 +3,9 @@
  */
 import type { TemplateArray } from '@wordpress/blocks';
 
+/**
+ * Template definition for Grouped Product Item Template in the Add to Cart form.
+ */
 export const GROUPED_PRODUCT_ITEM_TEMPLATE: TemplateArray = [
 	[
 		'woocommerce/add-to-cart-with-options-grouped-product-selector-item',
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-template/constants.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-template/constants.ts
@@ -52,17 +52,34 @@ export const GROUPED_PRODUCT_ITEM_TEMPLATE: TemplateArray = [
 						},
 					],
 					[
-						'woocommerce/product-price',
+						'core/group',
 						{
-							isDescendentOfSingleProductBlock: true,
-							fontSize: 'medium',
-							textAlign: 'right',
+							layout: {
+								type: 'flex',
+								orientation: 'vertical',
+							},
 							style: {
-								typography: {
-									fontWeight: 400,
+								spacing: {
+									blockGap: '0',
 								},
 							},
 						},
+						[
+							[
+								'woocommerce/product-price',
+								{
+									isDescendentOfSingleProductBlock: true,
+									fontSize: 'medium',
+									textAlign: 'right',
+									style: {
+										typography: {
+											fontWeight: 400,
+										},
+									},
+								},
+							],
+							[ 'woocommerce/product-stock-indicator' ],
+						],
 					],
 				],
 			],
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
@@ -75,6 +75,8 @@ class AddToCartWithOptions extends AbstractBlock {
 		}
 
 		if ( ! $product->is_in_stock() ) {
+			$product = $previous_product;
+
 			return wc_get_stock_html( $product );
 		}
 
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
@@ -74,7 +74,7 @@ class AddToCartWithOptions extends AbstractBlock {
 			return '';
 		}
 
-		if ( ! $product->is_in_stock() ) {
+		if ( ProductType::SIMPLE === $product->get_type() && ! $product->is_in_stock() ) {
 			$product = $previous_product;
 
 			return wc_get_stock_html( $product );
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
@@ -68,7 +68,7 @@ class AddToCartWithOptions extends AbstractBlock {
 
 		$previous_product = $product;
 		$product          = wc_get_product( $post_id );
-		if ( ! $product instanceof \WC_Product ) {
+		if ( ! $product instanceof \WC_Product || ( ProductType::SIMPLE === $product->get_type() && ! $product->is_purchasable() ) ) {
 			$product = $previous_product;
 
 			return '';
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
@@ -68,16 +68,10 @@ class AddToCartWithOptions extends AbstractBlock {
 
 		$previous_product = $product;
 		$product          = wc_get_product( $post_id );
-		if ( ! $product instanceof \WC_Product || ( ProductType::SIMPLE === $product->get_type() && ! $product->is_purchasable() ) ) {
+		if ( ! $product instanceof \WC_Product ) {
 			$product = $previous_product;
 
 			return '';
-		}
-
-		if ( ProductType::SIMPLE === $product->get_type() && ! $product->is_in_stock() ) {
-			$product = $previous_product;
-
-			return wc_get_stock_html( $product );
 		}
 
 		wp_enqueue_script_module( $this->get_full_block_name() );
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
@@ -74,6 +74,10 @@ class AddToCartWithOptions extends AbstractBlock {
 			return '';
 		}
 
+		if ( ! $product->is_in_stock() ) {
+			return wc_get_stock_html( $product );
+		}
+
 		wp_enqueue_script_module( $this->get_full_block_name() );
 
 		$product_type = $product->get_type();
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsQuantitySelector.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsQuantitySelector.php
@@ -97,6 +97,12 @@ class AddToCartWithOptionsQuantitySelector extends AbstractBlock {
 	/**
 	 * Render the block.
 	 *
+	 * The selector is hidden for:
+	 * - Simple products that are out of stock.
+	 * - Not purchasable simple products.
+	 * - External products with URLs
+	 * - Products sold individually
+	 *
 	 * @param array    $attributes Block attributes.
 	 * @param string   $content Block content.
 	 * @param WP_Block $block Block instance.
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsQuantitySelector.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsQuantitySelector.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 namespace Automattic\WooCommerce\Blocks\BlockTypes;
 
 use Automattic\WooCommerce\Blocks\Utils\StyleAttributesUtils;
+use Automattic\WooCommerce\Enums\ProductType;
 
 /**
  * AddToCartWithOptionsQuantitySelector class.
@@ -116,14 +117,22 @@ class AddToCartWithOptionsQuantitySelector extends AbstractBlock {
 			return '';
 		}
 
-		wp_enqueue_script_module( $this->get_full_block_name() );
+		if ( ProductType::SIMPLE === $product->get_type() && ( ! $product->is_in_stock() || ! $product->is_purchasable() ) ) {
+			$product = $previous_product;
+
+			return '';
+		}
 
 		$is_external_product_with_url        = $product instanceof \WC_Product_External && $product->get_product_url();
 		$can_only_be_purchased_one_at_a_time = $product->is_sold_individually();
 
 		if ( $is_external_product_with_url || $can_only_be_purchased_one_at_a_time ) {
+			$product = $previous_product;
+
 			return '';
 		}
+
+		wp_enqueue_script_module( $this->get_full_block_name() );
 
 		$is_stepper_style = isset( $attributes['quantitySelectorStyle'] ) && 'stepper' === $attributes['quantitySelectorStyle'] && ! $product->is_sold_individually();
 
--- a/plugins/woocommerce/src/Blocks/BlockTypes/ProductStockIndicator.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/ProductStockIndicator.php
@@ -63,7 +63,10 @@ class ProductStockIndicator extends AbstractBlock {
 	}
 
 	/**
-	 * Include and render the block.
+	 * Renders the stock indicator block.
+	 *
+	 * This method handles both direct product context and global product context,
+	 * ensuring the stock indicator displays correctly in various template scenarios.
 	 *
 	 * @param array    $attributes Block attributes. Default empty array.
 	 * @param string   $content    Block content. Default empty string.
--- a/plugins/woocommerce/src/Blocks/BlockTypes/ProductStockIndicator.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/ProductStockIndicator.php
@@ -71,31 +71,38 @@ class ProductStockIndicator extends AbstractBlock {
 	 * @return string Rendered block type output.
 	 */
 	protected function render( $attributes, $content, $block ) {
+		global $product;
+
 		if ( ! empty( $content ) ) {
 			parent::register_block_type_assets();
 			$this->register_chunk_translations( [ $this->block_name ] );
 			return $content;
 		}
-		$post_id = isset( $block->context['postId'] ) ? $block->context['postId'] : '';
-		$product = wc_get_product( $post_id );
+		$post_id           = isset( $block->context['postId'] ) ? $block->context['postId'] : '';
+		$product_to_render = wc_get_product( $post_id );
 
-		if ( ! $product || in_array( $product->get_type(), $this->get_product_types_without_stock_indicator(), true ) ) {
+		// Use the global product if the product to render can't be retrieved from the context.
+		if ( ! $product_to_render instanceof WC_Product ) {
+			$product_to_render = $product;
+		}
+
+		if ( ! $product_to_render || in_array( $product_to_render->get_type(), $this->get_product_types_without_stock_indicator(), true ) ) {
 			return '';
 		}
 
-		$availability = ProductAvailabilityUtils::get_product_availability( $product );
+		$availability = ProductAvailabilityUtils::get_product_availability( $product_to_render );
 
 		if ( empty( $availability['availability'] ) ) {
 			return '';
 		}
 
-		$total_stock        = $product->get_stock_quantity();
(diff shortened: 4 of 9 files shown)


plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
```diff
@@ -68,12 +68,16 @@ protected function render( $attributes, $content, $block ) {
...
 		$product          = wc_get_product( $post_id );
-		if ( ! $product instanceof \WC_Product ) {
+		if ( ! $product instanceof \WC_Product || ( ProductType::SIMPLE === $product->get_type() && ! $product->is_purchasable() ) ) {
 			$product = $previous_product;
 
 			return '';
 		}
 
+		if ( ! $product->is_in_stock() ) {
+			return wc_get_stock_html( $product );
```
dinhtungdu: Should we restore the previous product here before returning?

Aljullu: Ups, true. Good catch! Fixed in 2141101d884cac48c1b567d70895e0a3a691cbb4.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
```diff
@@ -68,12 +68,18 @@ protected function render( $attributes, $content, $block ) {
...
 			$product = $previous_product;
 
 			return '';
 		}
 
+		if ( ! $product->is_in_stock() ) {
+			$product = $previous_product;
+
+			return wc_get_stock_html( $product );
+		}
```
dinhtungdu: I'm sorry for the incremental review, but after comparing the behavior introduced in this PR and the classic add-to-cart template, I think we should cover not only out-of-stock status but also backorder an other statuses.

The way it works for the classic template is:
- always displaying the availability text.
- displaying the add-to-cart form conditionally based on stock status.

Maybe we should create a block for availability text?

Aljullu: > I'm sorry for the incremental review

Not at all, thanks for noticing it! :pray: 

Instead of always displaying the Stock text, I added the _Product Stock Indicator_ block to the simple product template part (4506d463347c482ca6ea39d1714477a5ecfa3d7c).

I also took the chance to update the Grouped Product template part to display when child products are out of stock (f68d2defbf01ce1cf56dd4020935b3f64f91c7e7). I updated the PR testing steps accordingly.

plugins/woocommerce/tests/php/src/Blocks/BlockTypes/AddToCartWithOptions.php
```diff
@@ -87,6 +87,31 @@ public function test_product_type_add_to_cart_render() {
...
+
+		$this->assertStringContainsString( 'Add to cart', $markup, 'The Simple Product Add to Cart Button is visible for purchasable in stock products.' );
+
+		$product->set_stock_status( 'outofstock' );
+		$product->save();
+		$markup = do_blocks( '<!-- wp:woocommerce/single-product {"productId":' . $product_id . '} --><!-- wp:woocommerce/add-to-cart-with-options /--><!-- /wp:woocommerce/single-product -->' );
+
+		$this->assertStringNotContainsString( 'Add to cart', $markup, 'The Simple Product Add to Cart Button is not visible for out of stock products.' );
+		$this->assertStringContainsString( 'Out of stock', $markup, 'The stock indicator is visible for out of stock products.' );
+	}
```
dinhtungdu: Nice! Looks like we can replace quite a few E2E test cases with this setup. 👏🏽

dinhtungdu (commented):
Thanks for the update, I can see that we already have the Product Stock Indicator block. It makes sense to use it here, but I have some concerns with the current approach, mostly because of the limitation of that block:
1. The Product Stock Indicator block isn’t available in the inserter, so if merchants delete it, they can’t add it back normally. It’s also unable to move the block out of Add to Cart with Options block.
2. If we use that block, shouldn’t we remove the check for out-of-stock simple products added [here](https://github.com/woocommerce/woocommerce/blob/558e2709bd21fab04da9b240e3fb64ae31ac2ae3/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php#L77-L81)?
3. 2nd point leads to another question: Should the Product Stock Indicator be a part of the Add to Cart with Options block?

I’m not very familiar with the way the Add to Cart with Options block is saved and rendered. From my initial observation, the content of the Add to Cart with Options block is saved to some template parts. And on the frontend, we will get and render the corresponding template part for the current product type. With this in mind, I don’t think the stock indicator should be a part of the block.

Fixing/updating the Product Stock Indicator block (and atomic blocks, in general) is out of the scope of this PR, so I think we can reduce the scope of this PR to just hide the Add to Cart form for out-of-stock and non-purchasable products and address the stock indicator in other PRs. What do you think?

Aljullu:
Thanks for all the feedback!

> Should the Product Stock Indicator be a part of the Add to Cart with Options block?
> 
> I’m not very familiar with the way the Add to Cart with Options block is saved and rendered. From my initial observation, the content of the Add to Cart with Options block is saved to some template parts. And on the frontend, we will get and render the corresponding template part for the current product type. With this in mind, I don’t think the stock indicator should be a part of the block.

What you say is right, but I'm not sure I agree with your conclusion. I believe it's better to have the Stock Indicator as a block within the template part rather than forcing its render from the PHP side.

I understand that the block is a bit confusing because it's only visible for on-backorder products and not for in-stock ones. However, using a block instead of rendering it via PHP has some advantages, such as allowing users to customize the block styles or changing its alignment:

Default | Customized
--- | ---
imatge | imatge

With the _Add to Cart with Options_ project we want to give more customizability options and this seems like an easy win IMO.

> The Product Stock Indicator block isn’t available in the inserter, so if merchants delete it, they can’t add it back normally. It’s also unable to move the block out of Add to Cart with Options block.

Huh, good catch. That feels like a bug. I opened an issue about it here: https://github.com/woocommerce/woocommerce/issues/56490.

dinhtungdu:
> I believe it's better to have the Stock Indicator as a block within the template part rather than forcing its render from the PHP side.

@Aljullu  I also think that we should use block instead of rendering it from the PHP side. That was my original suggestion. What I was saying is: I think the Product Stock Indicator block doesn't need to be nested inside the Add to Cart with Options block. because for some case, like out of stock, we're returning early, which means that we don't render inner blocks in those case, and Product Stock Indicator is one of those inner blocks.

Currently, when the product is out of stock, we're rendering the message manually, not the block. So all of customization merchant made to that block won't be available on the front end when the product is out of stock.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
```diff
@@ -68,12 +68,18 @@ protected function render( $attributes, $content, $block ) {
...
 			$product = $previous_product;
 
 			return '';
 		}
 
+		if ( ProductType::SIMPLE === $product->get_type() && ! $product->is_in_stock() ) {
+			$product = $previous_product;
+
+			return wc_get_stock_html( $product );
+		}
```
dinhtungdu: As I mentioned previously, I'm wary about this logic as we're manually render the stock indicator here for out of stock case but use stock indicator block for other case. Besides, handling some specific cases for a specific product inside the Add to Cart with Options block is a code smell to me.

We have `woocommerce/add-to-cart-with-options-grouped-product-selector` and `woocommerce/add-to-cart-with-options-variation-selector`, what about creating one wrapper for simple product?

dinhtungdu: > what about creating one wrapper for simple product?

In a follow up PR, for sure.

Aljullu: Thanks again for the feedback and the review, @dinhtungdu!

> We have `woocommerce/add-to-cart-with-options-grouped-product-selector` and `woocommerce/add-to-cart-with-options-variation-selector`, what about creating one wrapper for simple product?

Those blocks are used to wrap the attributes selector and the child products selector, they are not wrappers of the entire form. The _Add to Cart Button_, for example, is outside of those blocks. So I'm not sure we could do something similar for Simple products.

> I'm wary about this logic as we're manually render the stock indicator here for out of stock case but use stock indicator block for other case.

Yeah, that's a great point. I just pushed one more commit (c96a3adb429704491a6b5b2f88431ebd1fcaa6a6) which changes how things are implemented. The end result is the same, and thus the testing instructions are the same, but in the code now we do:

1. We always display the template part, no matter if the product is purchasable or out of stock.
2. The blocks themselves take care of showing up or not, depending on the specific product.

This has the benefit that now everything is in blocks, so we aren't printing anything directly from PHP. It adds some complexity to the blocks rendering logic, but I think that's unavoidable, as we need the logic to handle unpurchasable and out of stock products somewhere.

plugins/woocommerce/tests/php/src/Blocks/BlockTypes/AddToCartWithOptions.php
```diff
@@ -97,6 +97,31 @@ public function test_product_type_add_to_cart_render() {
 		remove_action( 'woocommerce_custom_add_to_cart', array( $this, 'print_custom_product_type_add_to_cart_markup' ) );
 	}
 
+	/**
+	 * Tests that no Add to Cart button is displayed for out of stock products and not purchasable products.
+	 */
```
dinhtungdu: Suggested by Claude Sonnet 3.5
```suggestion
	/**
	 * Tests that no Add to Cart button is displayed for out of stock products and not purchasable products.
	 * 
	 * Verifies that:
	 * 1. Add to Cart button is hidden for not purchasable products
	 * 2. Add to Cart button is visible for in-stock purchasable products
	 * 3. Add to Cart button is hidden and stock indicator shows for out of stock products
	 *
	 * @covers AddToCartWithOptions::render
	 */
```

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsQuantitySelector.php
dinhtungdu, line 1: Claude suggests this comment for the `render` method:
```
	/**
	 * Render the block.
	 *
	 * The selector is hidden for:
	 * - Simple products that are out of stock
	 * - Not purchasable products
	 * - External products with URLs
	 * - Products sold individually
	 *
	 * @param array    $attributes Block attributes.
	 * @param string   $content    Block content.
	 * @param WP_Block $block      Block instance.
	 * @return string Rendered quantity selector or empty string if hidden.
	 */
```

plugins/woocommerce/src/Blocks/BlockTypes/ProductStockIndicator.php
dinhtungdu, line 1: Claude suggests this comment for the `render` method:
```
	/**
	 * Renders the stock indicator block.
	 * 
	 * This method handles both direct product context and global product context,
	 * ensuring the stock indicator displays correctly in various template scenarios.
	 *
	 * @param array    $attributes Block attributes.
	 * @param string   $content    Block content.
	 * @param WP_Block $block      Block instance.
	 * @return string Rendered block output.
	 */
```

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-template/constants.ts
dinhtungdu, line 1: Claude suggests this comment for group item template constant:
```
/**
 * Template definition for grouped product items in the Add to Cart form.
 *
 * This template structures the layout for each product in a grouped product, including:
 * - Product title (as a link)
 * - Price information
 * - Stock status indicator
 *
 * The layout uses a vertical flex container to stack price and stock status,
 * ensuring consistent spacing and alignment.
 *
 * @type {TemplateArray}
 */
export const GROUPED_PRODUCT_ITEM_TEMPLATE: TemplateArray = [
```

dinhtungdu (approved):
Thanks for addressing my feedback. This PR is looking good to me now. Great work!

I included a review using Claude Sonnet 3.5 at the end of my reviews. It recommended adding some code comments I attached to this review.

#### Screenshots

https://github.com/user-attachments/assets/a7f92011-2e70-491c-b91e-481ee6fdfd0f

#### Tested steps:

- Tested against Single Product template with blockified Add to Cart with Options block.
- Tested single product with various stock statuses: in stock, out of stock, on backorder, non purchasable.
- Tested grouped products with products have all stock statuses.

#### Test environment

```
### WordPress Environment ###

WC Version: 9.9.0-dev
Log Directory Writable: ✔
WP Version: 6.7.2
WP Multisite: –
WP Memory Limit: 256 MB
WP Debug Mode: ✔
WP Cron: ✔
Language: en_US
External object cache: –

### Server Environment ###

Server Info: nginx/1.25.4
PHP Version: 7.4.33
PHP Post Max Size: 8 MB
PHP Time Limit: 30
PHP Max Input Vars: 1000
cURL Version: 8.5.0
OpenSSL/3.1.4

SUHOSIN Installed: –
MySQL Version: 8.0.33
Max Upload Size: 2 MB
Default Timezone is UTC: ✔
fsockopen/cURL: ✔
SoapClient: ✔
DOMDocument: ✔
GZip: ✔
Multibyte String: ✔
Remote Post: ✔
Remote Get: ✔

### Database ###

[REDACTED]

### Post Type Counts ###

attachment: 24
page: 8
post: 3
product: 19
product_variation: 7
revision: 66
wp_font_face: 36
wp_font_family: 12
wp_global_styles: 2
wp_navigation: 1
wp_template: 2
wp_template_part: 3

### Security ###

Secure connection (HTTPS): ❌
Hide errors from visitors: ❌Error messages should not be shown to visitors.

### Active Plugins (3) ###

Laravel DD for Wordpress: by Peter Hegman – 1.0.1
WooCommerce: by Automattic – 9.9.0-dev

### Inactive Plugins (4) ###

Akismet Anti-spam: Spam Protection: by Automattic - Anti-spam Team – 5.3.6
Hello Dolly: by Matt Mullenweg – 1.7.2
WooCommerce: by Automattic – 9.7.1

### Must Use Plugins (1) ###

test.php: by  –

### Settings ###

Legacy API Enabled: –
Force SSL: –
Currency: VND (₫)
Thousand Separator: .
Decimal Separator: ,
Number of Decimals: 0
grouped (grouped)
simple (simple)
variable (variable)

featured (featured)
outofstock (outofstock)
rated-1 (rated-1)
rated-2 (rated-2)
rated-3 (rated-3)
rated-4 (rated-4)
rated-5 (rated-5)

HPOS feature enabled: ✔
HPOS data sync enabled: –

### Logging ###

Enabled: ✔
Retention period: 30 days
Level threshold: –
Log directory size: 56 KB

### WC Pages ###

Shop base: #7 - /?page_id=7
Cart: #8 - /?page_id=8 -  Contains the woocommerce/cart block
Checkout: #9 - /?page_id=9 -  Contains the woocommerce/checkout block
My account: #10 - /?page_id=10 -  Contains the [woocommerce_my_account] shortcode

### Theme ###

Name: Twenty Twenty-Five
Version: 1.1
Theme type: Block theme

### Templates ###

Overrides: /Users/tung/workspace/woocommerce/plugins/woocommerce/templates/block-notices/error.php
/Users/tung/workspace/woocommerce/plugins/woocommerce/templates/block-notices/notice.php
/Users/tung/workspace/woocommerce/plugins/woocommerce/templates/block-notices/success.php


### Admin ###

analytics
product-block-editor
experimental-blocks
coupons
core-profiler
customize-store
customer-effort-score-tracks
import-products-task
shipping-smart-defaults
shipping-setting-tour
homescreen
marketing
minified-js
mobile-app-banner
onboarding
onboarding-tasks
payment-gateway-suggestions
product-custom-fields
printful
remote-inbox-notifications
remote-free-extensions
shipping-label-banner
subscriptions
store-alerts
transient-notices
woo-mobile-welcome
wc-pay-promotion
wc-pay-welcome-page
launch-your-store
blockified-add-to-cart
disable-core-profiler-fallback

product-pre-publish-modal
settings
use-wp-horizon
beta-tester-slotfill-examples

Daily Cron: ✔ Next scheduled: 2025-03-24 10:10:56 +00:00
Options: ✔
Notes: 66
Onboarding: completed

### Action Scheduler ###

Complete: 272
Oldest: 2025-03-04 10:10:57 +0000
Newest: 2025-03-24 06:48:39 +0000

Failed: 1
Oldest: 2025-03-05 03:00:25 +0000
Newest: 2025-03-05 03:00:25 +0000

Pending: 1
Oldest: 2025-03-25 02:41:48 +0000
Newest: 2025-03-25 02:41:48 +0000


Generated at: 2025-03-24 06:57:35 +00:00
```
</details>

#### Review by Claude Sonnet 3.5

<details>
<summary>Details</summary>


**Core Changes**:
* Adds stock indicator block to simple and grouped product templates
* Hides quantity selector and add to cart button for out of stock/not purchasable simple products
* Moves stock status display logic from PHP to blocks for better customization

**Code Quality**:
* The changes follow WooCommerce coding standards
* Proper unit tests have been added to verify the functionality
* The implementation handles edge cases (backorders, grouped products, etc.)

**Notable Implementation Details**:
* The stock indicator is now implemented as a block rather than direct PHP rendering
* Product availability checks are done at the block level
* Template parts have been updated to include stock indicators in appropriate locations

**Potential Concerns**:
* As noted by @dinhtungdu in the PR comments, there was initially a concern about manually rendering stock indicators in PHP vs using blocks
* This was addressed in commit c96a3ad which moved the logic to the blocks themselves

**Testing Coverage**:
* Unit tests verify behavior for:
* Not purchasable products
* In stock products
* Out of stock products
* Stock indicator visibility

**Positive Aspects**:
* Matches PHP template behavior
* Improves customization capabilities through block-based approach
* Handles all product states appropriately (in stock, out of stock, backorder)
* Good test coverage

**Suggestions for Improvement**:
* Consider adding JSDoc comments for the new TypeScript constants
* Could benefit from some inline comments explaining the logic flow in more complex conditions

Overall, this PR represents a solid improvement to the WooCommerce block system, particularly in how it handles product availability states. The changes are well-tested and follow best practices.
The main architectural decision to move stock status handling to blocks rather than PHP rendering is sound and provides better flexibility for customization.
**Recommendation**: Approve with minor suggestions for documentation improvements.


</details>

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-template/constants.ts
Aljullu: I added it in eb35c69a5ac6526935fbc4c518f3a039f9e665a2, but kept it shorter. I'm not sure if the `@type` definition adds much considering it's a TypeScript file. I also removed the sentences that were describing the layout itself, it feels like something that will easily end up being outdated if we introduce changes to the template but forget to update the comment.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsQuantitySelector.php
Aljullu: Good call! Added in eb35c69a5ac6526935fbc4c518f3a039f9e665a2 as well.

plugins/woocommerce/src/Blocks/BlockTypes/ProductStockIndicator.php
Aljullu: Included in eb35c69a5ac6526935fbc4c518f3a039f9e665a2 too. :slightly_smiling_face:

plugins/woocommerce/tests/php/src/Blocks/BlockTypes/AddToCartWithOptions.php
Aljullu: Added in eb35c69a5ac6526935fbc4c518f3a039f9e665a2 as well.

Issue #55563, opened 2025-02-17
Pull request: Added wc_sanitize_phone_number() to remove invalid characters in 9.5.0 removes as well spaces.


vyskoczilova:
### Prerequisites

- [x] I have carried out troubleshooting steps and I believe I have found a bug.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

Hi! I'm the author of https://woocommerce.com/products/phone-validator-and-formatter/ plugin, and this [#52476](https://github.com/woocommerce/woocommerce/pull/52476/files) edit complicated formatting of phone way too much.

https://github.com/woocommerce/woocommerce/pull/52476/files#diff-08b704510240689138540f200fb74dfef365591b86b864148b8d7b44465d4dccR886 This function at this place stripes of all incorrect characters including "spaces" but spaces can be sometimes wanted in the phone number.

My plugin allows users to select in which format they want to store the phone number and excluding a space forced me to update the phone number after WooCommerce processes it :( Especially the woocommerce_checkout_update_order_meta action is kind of silly. Is there any better way to achieve this? Allow having space in the phone number or filter the allowed tags somehow? I've spent a whole day on solving this trouble and my solution works, but it's not nice at all to touch already stored number.

```
\add_filter('woocommerce_process_checkout_field_billing_phone', [$this, 'allow_spaces_in_phone_numbers'],);
\add_filter('woocommerce_process_checkout_field_shipping_phone', [$this, 'allow_spaces_in_phone_numbers'],);
\add_action('woocommerce_checkout_update_order_meta', [$this, 'store_phone_with_spaces']);

/**
	 * Allow spaces in phone numbers
	 *
	 * @param string $value Phone number.
	 *
	 * @return string
	 */
	public function allow_spaces_in_phone_numbers(string $value): string
	{
		return \preg_replace('/[^\d\s+]/', '', $value);
	}

	/**
	 * Store phone numbers with spaces
	 *
	 * @param int $order_id Order ID.
	 *
	 * @return void
	 * @throws Exception
	 */
	public function store_phone_with_spaces(int $order_id)
	{
        // phpcs:disable WordPress.Security.NonceVerification.Missing
		if (!empty($_POST['billing_phone']) && !empty($_POST['billing_country'])) {
			$phone = \phone_validator_and_formatter()->standardize(
				\wc_clean(\wp_unslash($_POST['billing_phone'])),
[...]

PR #56402, opened 2025-03-14
Fix phone number validation

### LINKED ISSUE #54492: [Enhancement]: Revise wc_sanitize_phone_number() to remove invalid characters only for comparison and not for data storage.

[Added wc_sanitize_phone_number() to remove invalid characters.](https://github.com/woocommerce/woocommerce/pull/52476#top)
#52476

Since the 8.5.1 release on December 18th, I have had a couple of clients complain that their customer billing phone numbers are no longer formatted. This was the only change in the release that I can see that works with phone numbers. It appears this new feature resaves the stripped phone number formatting and it is added to order emails, the orders themselves and even the user profile. Would it be possible to change the phone number for the validation process only and not save the stripped version to the user profile? The formatting for my clients has been the standard (###) ###-####.

_No response_

### Additional context

_No response_

### LINKED ISSUE #55065: `WC_Checkout::validate_posted_data()` no longer sets errors for invalid phone numbers
### Prerequisites

### Describe the bug

If I enter an invalid phone number in classic checkout, the UI properly detects it as invalid:

Image

But if I then do something to trigger `WC_Checkout::validate_posted_data()`, I do not get any error messages about the phone number being invalid.

This is due to a change made in this PR: https://github.com/woocommerce/woocommerce/pull/52476/files

Prior to calling `WC_Validation::is_phone( $data[ $key ] )`, the phone number is sanitized. That means you'll probably never get into a situation w here `WC_Validation::is_phone( $data[ $key ] )` returns `false`.

### Expected behavior

I expect the `WC_Checkout::validate_posted_data()` function to set this error:

```
array (
  'billing_phone_validation' =>
  array (
    0 => '<strong>Billing Phone</strong> is not a valid phone number.',
  ),
)
```

### Actual behavior

 `WC_Checkout::validate_posted_data()` does not set any phone errors at all. The phone number is effectively seen as valid.

### Steps to reproduce

I've put together a code snippet that can be used to help trigger the correct code at the correct time. It also adds logging to better track the result of `validate_checkout()`.

First install this code snippet on your site:

```php
add_action('woocommerce_checkout_update_order_review', function($checkout_data = '') {
    class TestCheckoutClass extends WC_Checkout {
        public function __construct(WC_Checkout $instance)
        {
            self::$instance = $instance;
        }

        public function validate_checkout(&$data, &$errors)
        {
            parent::validate_checkout($data, $errors);
        }
    }

    $posted_data = [];

    parse_str( $checkout_data, $posted_data );

    $errors = new \WP_Error();
    $checkout = new TestCheckoutClass(WC()->checkout());
    $checkout->validate_checkout($posted_data, $errors);

    error_log(var_export($errors->errors, true));
});
```

Open up a terminal window or similar to tail error logs.

Then go to checkout and enter an invalid phone number. Example:

Image

The UI should pick it up as being invalid.

Then perform any action that would trigger `update_order_review`. Toggling the "Deliver to a different address?" checkbox works.

You will not see any errors in the error log -- just an empty array.

If you manually comment out this line: https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/class-wc-checkout.php#L889 Then repeat the steps, you will get this in your error log:

```
array (
  'billing_phone_validation' =>
  array (
    0 => '<strong>Billing Phone</strong> is not a valid phone number.',
  ),
)
```

### WordPress Environment

```
### WordPress Environment ###

WC Version: 9.6.0
Action Scheduler Version: ✔ 3.9.0
Log Directory Writable: ✔
WP Version: 6.7.1
WP Multisite: –
WP Memory Limit: 512 MB
WP Debug Mode: ✔
WP Cron: ✔
Language: en_GB
External object cache: –

### Server Environment ###

Server Info: nginx/1.23.4
PHP Version: 8.2.25
PHP Post Max Size: 8 MB
PHP Time Limit: 3600
PHP Max Input Vars: 1000
cURL Version: 7.74.0
OpenSSL/1.1.1w

SUHOSIN Installed: –
MySQL Version: 8.0.33
Max Upload Size: 8 MB
Default Timezone is UTC: ✔
fsockopen/cURL: ✔
SoapClient: ✔
DOMDocument: ✔
GZip: ✔
Multibyte String: ✔
Remote Post: ✔
Remote Get: ✔

### Database ###

[REDACTED]

### Post Type Counts ###

attachment: 10
course: 1
customize_changeset: 1
elementor_library: 1
lesson: 3
nav_menu_item: 4
page: 18
post: 2
product: 29
product_variation: 6
revision: 58
sensei_email: 19
sfwd-courses: 1
sfwd-lessons: 4
shop_coupon: 4
shop_order: 295
shop_order_refund: 8
shop_subscription: 37
wc_membership_plan: 9
wc_memberships_team: 3
wc_order_email: 1
wc_order_status: 9
wc_pickup_location: 6
wc_product_retailer: 1
wc_product_tab: 2
wc_team_invitation: 1
wc_user_membership: 16
wc_voucher: 4
wc_voucher_template: 1
wp_global_styles: 1
wp_navigation: 1

### Security ###

Secure connection (HTTPS): ❌
Hide errors from visitors: ✔

### Active Plugins (7) ###

Ashley Tweaks: by  –
Debug Bar Console: by Drew Jaynes – 0.3.1
Debug Bar: by wordpressdotorg – 1.1.6
Email Log: by Sudar – 2.4.9
Query Monitor: by John Blackbourn – 3.17.1
Record Queries: by Ashley Gibson – 1.0
WooCommerce: by Automattic – 9.6.0

### Inactive Plugins (36) ###

Akismet Anti-spam: Spam Protection: by Automattic - Anti-spam Team – 5.3.5
Custom Product Tabs Lite for WooCommerce: by SkyVerge – 1.9.1
Gravity Forms: by Gravity Forms – 2.8.18
LearnDash Licensing & Management: by LearnDash – 1.3
LearnDash LMS: by LearnDash – 4.12.0
Mollie Payments for WooCommerce: by Mollie – 7.9.1
Sensei LMS: by Automattic – 4.24.5
Teams for WooCommerce Memberships: by SkyVerge – 1.7.5-dev.1 (update to version 1.7.5 is available)
WooCommerce Address Validation: by SkyVerge – 2.11.3-dev.1 (update to version 2.11.3 is available)
WooCommerce Admin Custom Order Fields: by SkyVerge – 1.17.1-dev.1 (update to version 1.17.1 is available)
WooCommerce Authorize.Net Gateway: by SkyVerge – 3.10.10-dev.1 (update to version 3.10.10 is available)
WooCommerce Checkout Add-Ons: by SkyVerge – 2.7.3
WooCommerce Cost of Goods: by SkyVerge – 2.13.2 (update to version 2.13.3 is available)
WooCommerce Customer/Order/Coupon CSV Import Suite: by SkyVerge – 3.12.1-dev.1 (update to version 3.12.2 is available)
WooCommerce Customer/Order/Coupon Export: by SkyVerge – 5.5.4-dev.1
WooCommerce CyberSource Gateway: by SkyVerge – 2.8.3
WooCommerce Elavon Converge Gateway: by SkyVerge – 2.14.3
WooCommerce Global Payments HPP: by SkyVerge – 3.3.1
WooCommerce Local Pickup Plus: by SkyVerge – 2.11.7
WooCommerce Measurement Price Calculator: by SkyVerge – 3.23.7-dev.1
WooCommerce Memberships: by SkyVerge – 1.26.10
WooCommerce Moneris Gateway: by SkyVerge – 3.4.4-dev.1
WooCommerce Order Status Control: by SkyVerge – 1.16.0
WooCommerce Order Status Manager: by SkyVerge – 1.15.4
WooCommerce PayPal Payments: by WooCommerce – 2.9.6
WooCommerce PDF Product Vouchers: by SkyVerge – 3.12.6
WooCommerce Print Invoices/Packing Lists: by SkyVerge – 3.13.8-dev.1
WooCommerce Product Add-Ons: by Woo – 7.2.1 (update to version 7.4.0 is available)
WooCommerce Product Documents: by SkyVerge – 1.15.2
WooCommerce Product Retailers: by SkyVerge – 1.18.2
WooCommerce Product Reviews Pro: by SkyVerge – 1.19.3-dev.1 (update to version 1.19.3 is available)
WooCommerce Sequential Order Numbers Pro: by SkyVerge – 1.21.4-dev.1
WooCommerce Stripe Gateway: by Stripe – 9.1.1
WooCommerce Subscriptions: by WooCommerce – 6.6.1 (update to version 7.1.0 is available)
WooCommerce Tab Manager: by SkyVerge – 1.16.1-dev.1 (update to version 1.16.1 is available)
WooCommerce Twilio SMS Notifications: by SkyVerge – 1.19.0-dev.1

### Dropin Plugins () ###

### Settings ###

Legacy API Enabled: –
Force SSL: –
Currency: USD ($)
Currency Position: left
Thousand Separator: ,
Decimal Separator: .
Number of Decimals: 2
grouped (grouped)
simple (simple)
subscription (subscription)
variable (variable)
variable subscription (variable-subscription)

featured (featured)
outofstock (outofstock)
rated-1 (rated-1)
rated-2 (rated-2)
rated-3 (rated-3)
rated-4 (rated-4)
rated-5 (rated-5)

HPOS feature enabled: ✔
HPOS data sync enabled: ✔

### Logging ###

Enabled: ✔
Retention period: 30 days
Level threshold: –
Log directory size: 309 KB

### WC Pages ###

Shop base: #6 - /shop/
Basket: #7 - /basket/ -  Contains the [woocommerce_cart] shortcode
Checkout: #8 - /checkout/ -  Contains the [woocommerce_checkout] shortcode
My account: #9 - /my-account/

### Theme ###

Name: Storefront
Version: 4.6.0 (update to version 4.6.1 is available)
Child Theme: ❌ – If you are modifying WooCommerce on a parent theme that you did not build
personally we recommend using a child theme. See: How to create a child theme

Theme type: Classic theme
WooCommerce Support: ✔

### Templates ###

Overrides: –

### Admin ###

analytics
product-block-editor
coupons
core-profiler
customize-store
customer-effort-score-tracks
import-products-task
shipping-smart-defaults
shipping-setting-tour
homescreen
marketing
mobile-app-banner
onboarding
onboarding-tasks
product-custom-fields
remote-inbox-notifications
remote-free-extensions
payment-gateway-suggestions
printful
shipping-label-banner
subscriptions
store-alerts
transient-notices
woo-mobile-welcome
wc-pay-promotion
wc-pay-welcome-page
launch-your-store

experimental-blocks
minified-js
product-pre-publish-modal
settings
blueprint
use-wp-horizon
blockified-add-to-cart

Daily Cron: ✔ Next scheduled: 2025-02-04 12:06:06 +00:00
Options: ✔
Notes: 84
Onboarding: skipped

### Action Scheduler ###

Complete: 298
Oldest: 2025-01-06 12:31:36 +0000
Newest: 2025-02-03 14:39:18 +0000

Failed: 161
Oldest: 2024-06-13 12:24:44 +0100
Newest: 2025-02-03 14:41:08 +0000

Pending: 9
Oldest: 2025-02-03 15:05:20 +0000
Newest: 2025-03-31 15:43:17 +0100


Generated at: 2025-02-03 14:52:01 +00:00
```
</details>

### Isolating the problem

wavvves:

This pull request fixes a bug where phone number sanitize would make validation always pass by removing all invalid chars.
This sanitize call was remove, and replace with a new format utility function `wc_remove_non_displayable_chars()` to remove useless non-displayable chars, avoiding copy-paste issues, as originally intended as fix on the original PR.

Closes #55065 
Closes #54492

(For Bug Fixes) Bug introduced in PR #52476

1. Place orders using Checkout Block using the following numbers (copy-paste it as it has invisible chars): `‪+31123456789‬`
2. Verify no phone validation errors show up.
3. After placing the order verify that on Order Confirmation, and Order details in admin do not display the non-visible chars. You can use this [tool](https://www.soscisurvey.de/tools/view-chars.php) to highlight them.
image

4. Place another order using a phone numbers with `# _ - + / ( ) .` and spaces:
  - `+1 (800) 123-4567`
  - `0044/20-7946.0958`
  - `+49 (0)30 1234-5678`
  - `#123-4567`
  - `+55 (11) 91234_5678`
5. Verify no phone number validation errors occur, and the the formatted number is safely shown on Order Confirmation and Order Details.
6. Smoke test with invalid phone numbers, eg `12345asdfg`

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


ralucaStan:
@vyskoczilova you mentioned taking a[ look at the PR](https://github.com/woocommerce/woocommerce/issues/55563#issuecomment-2688083371) when it was ready so I'm tagging you as well

ralucaStan:
Let's extend the testing steps to include the `-` character, and to check on how the phone details are shown in the order confirmation page and in the Admin @wavvves

plugins/woocommerce/includes/wc-formatting-functions.php
```diff
@@ -1568,6 +1568,51 @@ function wc_parse_relative_date_option( $raw_value ) {
...
+ *
+ * - **Non-breaking space (`U+00A0`)** – Useful for preventing line breaks in addresses.
+ * - **Word joiner (`U+2060`)** – Sometimes needed for proper text rendering in certain scripts.
+ *
+ * @param string $raw_value The input string to sanitize.
+ *
+ * @return string The sanitized string without problematic characters.
+ * @since 9.9.0
+ */
+function wc_remove_non_displayable_chars( string $raw_value ): string {
```
ralucaStan, line 1594: Curious why you created a separate function. 
https://github.com/woocommerce/woocommerce/blob/007190eca874b3fe8087c64d0643ee6f233cc49f/plugins/woocommerce/includes/wc-formatting-functions.php#L1051 

`wc_format_phone_number` already removes some unwanted characters so we could consolidate this in one place?

ralucaStan (commented):
Left a question about the implementation

plugins/woocommerce/includes/wc-formatting-functions.php
wavvves: That function is way more destructive, and includes is_phone() validation inside before actually doing something. **I did not find any usage for it within our codebase.**

The new function is not phone only, and designed to safely remove non-display chars ahead of validation to avoid issues like copy-pasting problems. If this PR is approved and no further problems arise, I believe it should be extended to other fields.

wavvves:
> Let's extend the testing steps to include the `-` character, and to check on how the phone details are shown in the order confirmation page and in the Admin @wavvves

@ralucaStan Order confirmation and admin checking were already in the testing instructions.
I expanded the valid number testing to include all permitted chars `(# _ - + / ( ) .` and spaces) with examples.

vyskoczilova:
@ralucaStan, from reading the code, I see it looks good; I will test it out with the RC candidate and let you know. I'm happy that I could drop the hacks around when 9.9.0 is released :) (or maybe you can ship it to 9.8.0 already?)

plugins/woocommerce/includes/wc-formatting-functions.php
```diff
@@ -1568,6 +1568,51 @@ function wc_parse_relative_date_option( $raw_value ) {
...
+ * - **Soft hyphen (`U+00AD`)** – Invisible unless text is broken across lines.
+ * - **Zero-width spaces & joiners (`U+200B–U+200D`)** – Invisible and can cause copy/paste issues.
+ * - **Directional markers (`U+200E–U+200F`, `U+202A–U+202E`)** – Can affect text rendering.
+ * - **Byte Order Mark (BOM) (`U+FEFF`)** – Can interfere with encoding.
+ * - **Interlinear annotation characters (`U+FFF9–U+FFFB`)** – Rarely used and unnecessary in checkout fields.
+ *
+ * It does **not** remove:
+ *
+ * - **Non-breaking space (`U+00A0`)** – Useful for preventing line breaks in addresses.
+ * - **Word joiner (`U+2060`)** – Sometimes needed for proper text rendering in certain scripts.
```
alexflorisca, line 1587: Is this ever needed for a phone number though?

plugins/woocommerce/includes/wc-formatting-functions.php
alexflorisca: What might be useful here is to have `is_phone` and other validation functions call this function on the input first

alexflorisca (commented):
Nice one Paulo, left a few inline comments. Wondering why the `wc_sanitize_phone_number` function doesn't remove these, as it feels like it should. The regex it uses `/[^\d+]/` removes anything that isn't a digit? Does `preg_replace` ignore those unicode chars?

I also wonder if we should create an issue to do this for other sanitization/validation functions too if we go ahead with this?

I thought about if there are any risks with stored phone numbers that already have those invisible characters, vs ones that don't. I don't see any problems there, I think it would be a welcome addition. If other extensions have implemented similar functionality to remove them, it shouldn't affect them.

Lastly, how did you come up with the list of character to be removed? Are there any others that could potentially be harmful?

plugins/woocommerce/includes/wc-formatting-functions.php
wavvves: We could, ahead of validation, but it would allow these to be persisted in the db, perpetuating copy-paste issues.

wavvves:
> why the `wc_sanitize_phone_number` function doesn't remove these, as it feels like it should. 
It does, but also removes a lot more. We don't want to fully sanitize before validation, since validation here only checks for allowed chars, it would always pass after sanitizing.

I'm keeping the old validation behaviour, while removing issues with invisible useless chars being copy-pasted.

> I also wonder if we should create an issue to do this for other sanitization/validation functions too if we go ahead with this?

I created a separate function so it can clean other fields. Do you think this would be better suited for frontend only?

> I thought about if there are any risks with stored phone numbers that already have those invisible characters, vs ones that don't. I don't see any problems there, I think it would be a welcome addition. If other extensions have implemented similar functionality to remove them, it shouldn't affect them.
> Lastly, how did you come up with the list of character to be removed? Are there any others that could potentially be harmful?

There's no harm in letting those pass, the only issue is causing validation fails leaving the user clueless as they are non displaying chars. is_phone checks if remaining chars exist, if those are invisible to the user it's hard to understand why the number is not being accepted.

plugins/woocommerce/includes/wc-formatting-functions.php
wavvves: This function is not intended to be phone number specific, actually both nbsp and word joiner are useful in phone number rendering, and can cripple validation. I believe these too have generic value over other fields too, and since they are not causing issues currently, should be kept.

I think we need to realize here that we don't have phone validation whatsoever. There's zero format checking, we just regex for valid chars. That is far from stating a phone number is valid. This PR fixes a regression where sanitizing would cause validation to always pass, while keeping a fix that allows non displaying chars to not interfere with validation when copy-pasting numbers.

plugins/woocommerce/includes/wc-formatting-functions.php
alexflorisca: It wouldn't if we strip them out before validation right?

wavvves: I mean, it really depends if we want to clean up data just for validation, or for saving too. I'd say, since they weren't allowed to be saved before ( the method is_phone whitelists chars), we strip them away before saving to keep things as before.

plugins/woocommerce/includes/wc-formatting-functions.php
ralucaStan: > while keeping a fix that allows non displaying chars to not interfere with validation when copy-pasting numbers.

@alexflorisca this is[ the fix Paulo](https://github.com/woocommerce/woocommerce/pull/52476/files#diff-317ef43b05d33cd872b5d83d972174ec9429f775984c3b187893cdf6bf89bbe0R235) is mentioning.

plugins/woocommerce/includes/wc-formatting-functions.php
ralucaStan: > The new function is not phone only, and designed to safely remove non-display chars ahead of validation to avoid issues like copy-pasting problems. If this PR is approved and no further problems arise, I believe it should be extended to other fields.

Thanks Paulo! Why didn't you apply this to the other fields from `validate_posted_data` in this PR?

ralucaStan (dismissed):
I think we should apply the function to the other fields as well.

I'm ok with keeping this PR just to fix the regression. But I think we should improve on our phone formating logic, as [mentioned here](https://github.com/woocommerce/woocommerce/issues/54492#issuecomment-2694298210)

@wavvves could you create a followup ticket for this?

plugins/woocommerce/includes/wc-formatting-functions.php
alexflorisca: To me it makes more sense to remove any hidden/useless characters as early as possible. Then that input can be passed to validation and if it passes to the database. Similar principle to "Escape input as early as possible"

alexflorisca:
> We don't want to fully sanitize before validation, since validation here only checks for allowed chars, it would always pass after sanitizing.

I see, so the problem we have is that we want to remove SOME characters before validation, and others after? Which I think is ok and what you've done here, I'm with you now. I think it might be useful to adda comment to says something like "This removes invisible characters which are problematic for validation. Further santization is done later with `wc_sanitize_phone_number`"

> I created a separate function so it can clean other fields. Do you think this would be better suited for frontend only?

No what I meant here is that we should run the `wc_remove_non_displayable_chars` function on other inputs too, not just phone number.

So i think if you add the comment and use the `wc_remove_non_displayable_chars` in the other POSTed data in the `validate_posted_data` function, that should be sweet 👍

ralucaStan:
@wavvves what is delaying this PR from getting closed?

plugins/woocommerce/includes/wc-formatting-functions.php
wavvves: > Thanks Paulo! Why didn't you apply this to the other fields from validate_posted_data in this PR?

I felt this would be out of scope of fixing the regression only. I will create an issue for extending this to other fields, or probably open a PR for it right away since it would only take a few minutes.

wavvves:
> @wavvves what is delaying this PR from getting closed?

CI was failing after trunk merges and I couldn't relate the failed tests with these changes. Updated with trunk and re-running now hoping it gets ✅ . If not, I'll probably temporarily revert my changes to confirm it as the cause.

wavvves:
I fixed the CI finally, while this PR was up some tests were introduced that passed because of the bug fixed here.

> @ralucaStan: could you create a followup ticket for this?

I created https://github.com/woocommerce/woocommerce/issues/56795 for improving phone formatting.

> @ralucaStan: I think we should apply the function to the other fields as well.

Opened https://github.com/woocommerce/woocommerce/pull/56796 to use this on other fields, and keep this regression fix scoped 👍🏼

alexflorisca:
Apparently there is an integrated AI review bot in github that i wanted to try out here 

@CodiumAI-Agent /review

alexflorisca:
Ok that was pretty bogus! Proper review coming up

alexflorisca (approved):
Thanks for adding the comments Paulo, I think that's clearer. Happy to address the other fields in a separate PR. Tests pass, I've res-tested and all works ok, so happy to approve from my end. Maybe wait for Raluca's review before merging too as she was quite involved here too.

PR #56453, opened 2025-03-17
Feat: Add apply_filter to allow the adding of filtering post clauses outside is_main_query

onstage2426:

Added a apply_filters to the value retrieved from the is_main_query() which is checked before adding the filter post clauses.

The reason for the change: To allow the application of the post clauses outside the main query. The attribute filter via the attribute lookup table and the price filter are applied via post clauses instead of the WP_Query->query_vars, so when you copy the query_vars from the main query than the attribute filter (only via attribute lookup table) and price filters are not applied and because these are locked behind the is_main_query there is no way to apply them. 

A other option would be to check if a query_var is present with a value instead of a filter to save a little perf. In the code example for testing you have for examlpe the check for wc_query === product_query. This is added by wc_query to the main_query. If that would be preffered i will change it to that.

1. New environment with default Wordpress theme and WooCommerce as only plugin.
2. Enable the attribute lookup table WooCommerce->Settings->Products->Advanced->Enable table usage
3. Create two products attributes in a single tax. pa_color with  term "red" and "green".
4. Create two test products, with different attribute (1. pa_color = red, 2. pa_color = green) and different prices ( 1. €‎100, 2. €‎200)
5. In the shop page use the filters to test two cases (?filter_color=red) and (?min_price=150)
6. Use the following snippit in the functions.php to see that the filters will not be applied when you copy the query_vars. (will retrieve 2 in both attribute and price test case.
```
add_action( "woocommerce_before_main_content", function () {
    $query_vars = $GLOBALS['wp_query']->query_vars;
    print_r( (new WP_Query($query_vars))->posts );
} );
```
7. Use the following snippit in functions.php to test the created apply_filters. This examle checks for wc_query === product_query which is set in the main query by wc_query class that filters the main_query.
```
add_filter( "woocommerce_enable_filter_post_clauses", function ($enabled, $wp_query) {
    if ($wp_query->get('wc_query') === 'product_query') {
        return true;
    }
    return $enabled;
}, 10, 2 );
```

Tested as above, because its only a filter breaking of existing environments is not possible. The only way of breaking it would than also be via a add_filter which is then developer caused.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Add filter to allow the adding of filtering post clauses outside "is_main_query".
</details>

<details>

#### Comment 

</details>


albarin:
Hi @Onstage2426, thanks so much for your contribution! 🙌 
After reviewing we have a few suggestions: what do you think about slightly modifying the filter name to `woocommerce_enable_post_clause_filtering`?

Also, the linter is failing because new hooks need to be documented, see: https://github.com/woocommerce/woocommerce/actions/runs/13900421163/job/39286856875?pr=56453#step:4:25. Would you mind adding some docs for both the hooks? You can see an example of the expected format [here](https://github.com/woocommerce/woocommerce/blob/ab382fc82188a5623819c2ff591b141cdff8b91d/plugins/woocommerce/src/Admin/PluginsHelper.php#L232-L239).

I've also completed the changelog details on the PR description so the changelog file is autogenerated.

Thanks a lot again!

onstage2426:
Hey @albarin, thanks for the response.
I made some changes as you indicated, i did skip the "since" param in the phpdoc comment.

albarin:
@Onstage2426  Thanks for the changes!
We need the `@since` param in the docs, otherwise the linter will keep failing, and we won't be able to merge:
https://github.com/woocommerce/woocommerce/actions/runs/14130826929/job/39700773132?pr=56453#step:4:27

You could add `@since 9.9.0`. Thanks!

onstage2426:
Hey @albarin, thanks for your patience with me.
The param has been added.

albarin (approved):

Issue #55856, opened 2025-02-26
Spike: Identify duplication introduced when building the feature plugin


ralucaStan:
Building the feature plugin added some technical debt where we were forced to override or replace certain functionality from core.

I’d like for us to spend some time identifying these and either consolidating them in a backwards compatible manner so that more of the codebase is shared by shortcode and blocks on the backend, or further separate them so that the shortcode experience could be easily decoupled

Areas with technical debts:
- `CartController` in core, e.g. `add_to_cart` method.
- coupons, shipping, creating the order, CheckoutController
- filters that duplicate things from the cart class core
- etc

context: pca54o-8R5-p2#comment-7335

mikejolley:
# Cart class consolidation (first pass)

This is a list of methods that either overlap, override each other, or could be abstracted and shared between
classes to consolidate logic and reduce complexity. This list is a combination of my own notes and AI analysis.

## Architectural Decisions

- `CartController` is a utility class that is used to control the cart for the Store API.
- `WC_Cart` is the core class that is used to control the cart for the rest of the site; this includes helper methods, session handling, and other cart state management.

As part of this refactor we should review the following:

- Whether we can consolidate the cart controller and WC_Cart methods
- Whether or not we can split shared functionality into shared classes
- If a larger refactor is required to reduce the scope of the WC_Cart class so it is more reusable, for example, by creating a new class to manage the cart state, which WC_Cart and CartController can interact with

## CartController: Cart Item Validation

```
// WC_Cart handles basic validation
public function check_cart_items()
public function check_cart_item_stock()

// While CartController has more extensive validation
public function validate_cart()
public function validate_cart_items()
public function validate_cart_item()
```

Suggestion: Unify validation logic into a separate CartValidator class that both can use.

```
class CartValidator {
    public function validateItem(array $item): void 
    public function validateItems(WC_Cart $cart): void 
    public function validateStock(WC_Product $product, int $quantity): void
    public function validateCoupons(array $coupons): void
}
```

Throw exceptions and allow the caller to handle them as needed e.g. WC_Cart can convert to notices.

## CartController: Coupon Handling

```
// WC_Cart
public function apply_coupon($coupon_code)

// CartController
public function apply_coupon($coupon_code)
```

The CartController overrides core coupon functionality to avoid notices being added when applying a coupon. It also alters validation logic and error handling.

Suggestion, either:

[...]

PR #56455, opened 2025-03-17
Adjusting the default quantity of the ProductButton block

### LINKED ISSUE #56454: Adjusting the default quantity of the ProductButton block
### Prerequisites

### Describe the bug

Currently, the `woocommerce_add_to_cart_quantity` hook is being used within the block’s [render function](https://github.com/woocommerce/woocommerce/blob/3f5f8fd80610a52a1f53fc17a11d4c90d6715c85/plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php#L156) instead of being applied at request time as [it used to run](https://github.com/woocommerce/woocommerce/blob/3f5f8fd80610a52a1f53fc17a11d4c90d6715c85/plugins/woocommerce/includes/class-wc-cart.php#L1050) for classic WooCommerce areas. This usage can potentially introduce risks and potential issues.

### Expected behavior

- The `woocommerce_loop_add_to_cart_args` hook should be the hook used for setting the initial quantity within the block’s render function. Similar to how the shortcodes work [here](https://github.com/woocommerce/woocommerce/blob/3f5f8fd80610a52a1f53fc17a11d4c90d6715c85/plugins/woocommerce/includes/wc-template-functions.php#L1397).
- The Store API CartController should be responsible for handling `woocommerce_add_to_cart_quantity` at request time. A related technical debt issue for maintaining backward compatibility for these hooks is documented [here](https://github.com/woocommerce/woocommerce/issues/55856).

### Actual behavior

The `woocommerce_add_to_cart_quantity` hook is currently applied during rendering instead of within the Store API’s add-to-cart request.

### Potential risks of using `woocommerce_add_to_cart_quantity` in rendering:

- **Validation concerns**: Classic WooCommerce areas might perform validation tasks (e.g., throwing exceptions). If these occur during rendering, they could cause the page to break unexpectedly.
- **Superglobal dependencies**: This hook might assume access to `$_REQUEST` or `$_POST`, which would be problematic during rendering, leading to unpredictable behavior.

### Screencasts
Here some examples of the current behavior:
- Classic theme: https://d.pr/v/KaB0XM
- Block theme: https://d.pr/v/IPT7cl

cc @gigitux 

### Isolating the problem

xristos3490:

This PR ensures that the default quantity used when rendering the ProductButton block is fetched in the same way as it is in the "[woocommerce_template_loop_add_to_cart](https://github.com/woocommerce/woocommerce/blob/6f05893904a1e535fb359a25cf86830ab3bde438/plugins/woocommerce/includes/wc-template-functions.php#L1430)" function used in classic themes. This is by utilizing the `woocommerce_loop_add_to_cart_args` hook.

The main takeaway here is that the `woocommerce_add_to_cart_quantity` hook has been removed from that block's rendering function and will instead be addressed as part of the future enhancements to the Store API (see #56454), if that decision is made. For now, a workaround for this hook can be achieved using the `woocommerce_add_cart_item ` alternative that functions in both the classic and the StoreAPI contexts.

### Risks of using `woocommerce_add_to_cart_quantity` during rendering addressed by this PR:
- **Validation issues**: Classic WooCommerce sections may carry out validation operations (like throwing exceptions). If these happen during rendering, they could unexpectedly disrupt the page.
- **Superglobal dependencies**: This hook may rely on $_REQUEST or $_POST, creating potential problems during rendering and resulting in erratic behavior.

Closes #56454 

### Prerequisites

Install the following PHP code for easier debugging:
```php
add_filter( 'woocommerce_add_to_cart_quantity',
	function( $quantity, $product_id ) {
		error_log( 'woocommerce_add_to_cart_quantity: ' . $product_id );
		return $quantity;
	},
	10,
	2
);
add_filter( 'woocommerce_add_cart_item',
	function( $data, $cart_id ) {
		error_log( 'woocommerce_add_cart_item: ' . $data['product_id'] );
		return $data;
	},
	10,
	2
);
add_filter( 'woocommerce_loop_add_to_cart_args',
	function( $args, $product ) {
		error_log( 'woocommerce_loop_add_to_cart_args: ' . $product->get_id() );
		return $args;
	},
	10,
	2
);
```

1. Use a classic theme (e.g., Storefront). 
2. Navigate to the Shop page. 
3. Check the logs to ensure that only `woocommerce_loop_add_to_cart_args` runs during the initial rendering. 
4. Click on a simple product's add-to-cart button and observe the `woocommerce_add_to_cart_quantity` and `woocommerce_add_cart_item` run. 
5. Now switch to a block theme (e.g., TT5). 
6. Navigate to the Shop page again.
7. Ensure that only `woocommerce_loop_add_to_cart_args` runs during the initial rendering (not `woocommerce_add_to_cart_quantity`).
8. Click on the same product's add-to-cart button and notice that only the `woocommerce_add_cart_item` hook runs.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/improve-product-button-default-qty
+++ b/plugins/woocommerce/changelog/improve-product-button-default-qty
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Default quantity for ProductButton block now fetched via `woocommerce_loop_add_to_cart_args` instead of `woocommerce_add_to_cart_quantity` hook.
--- a/plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
@@ -145,16 +145,27 @@ class ProductButton extends AbstractBlock {
 			)
 		);
 
-		$default_quantity = 1;
 		/**
-		* Filters the change the quantity to add to cart.
-		*
-		* @since 10.9.0
-		* @param number $default_quantity The default quantity.
-		* @param number $product_id The product id.
-		*/
-		$quantity_to_add = apply_filters( 'woocommerce_add_to_cart_quantity', $default_quantity, $product->get_id() );
+		 * Allow filtering of the add to cart button arguments.
+		 *
+		 * @since 9.7.0
+		 */
+		$args = apply_filters(
+			'woocommerce_loop_add_to_cart_args',
+			array(
+				'quantity'   => 1,
+				'class'      => $html_classes,
+				'attributes' => array(
+					'data-product_id'  => $product->get_id(),
+					'data-product_sku' => $product->get_sku(),
+					'aria-label'       => $product->add_to_cart_description(),
+					'rel'              => 'nofollow',
+				),
+			),
+			$product
+		);
 
+		$quantity_to_add                   = absint( $args['quantity'] );
 		$is_descendent_of_add_to_cart_form = isset( $block->context['woocommerce/isDescendantOfAddToCartWithOptions'] ) ? $block->context['woocommerce/isDescendantOfAddToCartWithOptions'] : false;
 		$add_to_cart_text                  = null !== $product->add_to_cart_text() ? $product->add_to_cart_text() : __( 'Add to cart', 'woocommerce' );
 		if ( $is_descendent_of_add_to_cart_form && null !== $product->single_add_to_cart_text() ) {
@@ -167,25 +178,6 @@ class ProductButton extends AbstractBlock {
 			'addToCartText'   => $add_to_cart_text,
 			'tempQuantity'    => $number_of_items_in_cart,
 			'animationStatus' => 'IDLE',
-		);
-
-		/**
-		 * Allow filtering of the add to cart button arguments.
-		 *
-		 * @since 9.7.0
-		 */
-		$args = apply_filters(
-			'woocommerce_loop_add_to_cart_args',
-			array(
-				'class'      => $html_classes,
-				'attributes' => array(
-					'data-product_id'  => $product->get_id(),
-					'data-product_sku' => $product->get_sku(),
-					'aria-label'       => $product->add_to_cart_description(),
-					'rel'              => 'nofollow',
-				),
-			),
-			$product
 		);
 
 		if ( isset( $args['attributes']['aria-label'] ) ) {


xristos3490:
**Note to the reviewer:** I think this change might come across as breaking, and if so, we could block this PR until the hooks parity in the Store API is in place. Considering how long this has been out, I believe it is a bold step toward an improved structure.

gigitux:
Thanks, @xristos3490, for your work!  

I understand this is a POC, but I have a few questions:  

> **Validation issues:** Classic WooCommerce sections may perform validation (e.g., throwing exceptions). If these occur during rendering, they could unexpectedly disrupt the page.  
> **Superglobal dependencies:** This hook may rely on `$_REQUEST` or `$_POST`, which could cause issues during rendering and lead to erratic behavior.  

Could you expand on these points?  

> The **woocommerce_add_to_cart_quantity** hook has been removed from this block’s rendering function and may be reconsidered as part of future Store API enhancements ([woocommerce/woocommerce#56454](https://github.com/woocommerce/woocommerce/issues/56454)).  

Is this necessary? The Store API already allows developers to specify the quantity via the request body. It should be up to the consumer to interpret and pass the correct value.  

For example, here’s how the Product Button handles it:  
https://github.com/woocommerce/woocommerce/blob/f5ec121fea6882c6d7e607c3c94e12f6cfb86143/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts#L119-L122

xristos3490:
Hey @gigitux! 

> Is this necessary? The Store API already allows developers to specify the quantity via the request body. It should be up to the consumer to interpret and pass the correct value.

Yes, that's true. However, the current approach seems to conflict with the classic Woo areas. This PR offers the same functionality by utilizing the `woocommerce_loop_add_to_cart` hook instead of the `woocommerce_add_to_cart_quantity` hook.

> Could you expand on these points?

For sure, the most important risk here is that the woocommerce_add_to_cart_quantity hook is being used within the block’s [render function](https://github.com/woocommerce/woocommerce/blob/3f5f8fd80610a52a1f53fc17a11d4c90d6715c85/plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php#L156) instead of being applied at request time as [it used to run](https://github.com/woocommerce/woocommerce/blob/3f5f8fd80610a52a1f53fc17a11d4c90d6715c85/plugins/woocommerce/includes/class-wc-cart.php#L1050) for classic WooCommerce areas. Here's an example of how one can use this hook in a classic theme that will break the block theme archive pages: 
```php
add_filter( 'woocommerce_add_to_cart_quantity',
	function( $quantity, $product_id ) {
		if ( $quantity < 10 ) {
			throw new Exception( 'Quantity must be at least 10' );
		}

		return $quantity;
	},
	10,
	2
);
```
By using the filter above 3PDs could control add-to-cart and error handling in the same hook. Here's how it looks in Storefront (a non-block theme):
https://d.pr/v/5feScr

At the same time, the TT5, is now broken:
Screenshot 2025-03-18 at 10 58 30 AM

Furthermore, the same hook can still work in the add-to-cart button on a single product page within a block theme, which further highlights a conflict here: https://d.pr/v/apN6hO

Does this help?

gigitux:
@xristos3490 Thank you for providing more context 🙇 

I see the issue here, and I think that it is something that we should discuss given it impacts the new Add To Cart block too (cc @mdperez86 @Aljullu).

My main concern is that this change is a breaking change, and also I'm afraid that we can lose some compatibility with 3rd party extensions.

Aljullu:
> I see the issue here, and I think that it is something that we should discuss given it impacts the new Add To Cart block too (cc @mdperez86 @Aljullu).

The impact on the new _Add to Cart with Options_ block will be smaller because in that scenario we will have the _Quantity Selector_ block, so the quantity will come from the input in that block instead of the PHP filter. The only exception will be if the _Quantity Selector_ block has been removed, in which case the form will "fall back" to the default quantity.

> My main concern is that this change is a breaking change, and also I'm afraid that we can lose some compatibility with 3rd party extensions.

That's my concern as well. I'm not sure if there is a way to fix this in a non-breaking way. :confused: 

@xristos3490 to better understand: is this change required to get extensions working or it was more like a "polish" task to make the filters consistent between old product grids and the new blocks?

xristos3490:
> is this change required to get extensions working or it was more like a "polish" task to make the filters consistent between old product grids and the new blocks?

No blockers here. This is a "polish" task aimed at improving the structure for 3PDs. I understand how breaking this can be, so what if we block this PR until we have a clearer picture on https://github.com/woocommerce/woocommerce/issues/55856? cc @Aljullu  @gigitux

gigitux:
> > is this change required to get extensions working or it was more like a "polish" task to make the filters consistent between old product grids and the new blocks?
> 
> No blockers here. This is a "polish" task aimed at improving the structure for 3PDs. I understand how breaking this can be, so what if we block this PR until we have a clearer picture on #55856? cc @Aljullu @gigitux

Sounds great to me! Thanks for raising this issue! 🙇

xristos3490:
Closing this as stale.

PR #56459, opened 2025-03-17
Remove obsolete block registration mechanism used for Product Gallery only

kmanijak:

Accidentally, I found out there's the whole block registration mechanism implemented in parallel to the one we're using widely:
- we use `registerProductBlockType` from `plugins/woocommerce/client/blocks/assets/js/atomic/utils/register-product-block-type.ts`
- while there was also `BlockRegistrationManager` in `plugins/woocommerce/client/blocks/assets/js/atomic/utils/blocks-registration-manager/blocks-registration-manager.ts` which, even though was kept in atomic utils, seemed to be tightly coupled and used only by Product Gallery block. Its settings were directly imported and used within util, not the other way around.

That means, along with registration we're doing in `index.tsx` [here](https://github.com/woocommerce/woocommerce/blob/5d659576172a81963a49f4581908b7f7cae04721/plugins/woocommerce/client/blocks/assets/js/blocks/product-gallery/index.tsx#L20) there was a parallel mechanism registering and unregistering the block when in Editor. It was unnecessary use of resources impacting the whole Editor no matter if Product Gallery was in use or not, hence removal.

Make sure the block is available correctly in correct context

#### Single Product template

3. Also, try inserting the block somewhere else in the template
4. **Expected**: The block is available and can be inserted

#### Single Product block

1. Create new post
2. Try inserting Product Gallery block
3. **Expected**: The block is NOT available in the inserter
4. Insert Single Product block and choose some product
5. Replace the Product Image block with the Product Gallery (beta) block.
6. **Expected**: The block is available and can be inserted

Focus on Thumbnail block, play with the size control and make sure it's reflected in Editor

Save and go to the frontend

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/atomic/utils/blocks-registration-manager/blocks-with-restriction.ts
```diff
@@ -1,42 +0,0 @@
...
-		blockSettings: ProductGalleryBlockSettings,
-		allowedTemplates: {
-			'single-product': true,
-		},
-		allowedTemplateParts: {
-			'product-gallery': true,
-		},
-		availableInPostOrPageEditor: false,
-		isVariationBlock: false,
-	},
```
cpapazoglou, line 41: Seems that we shouldn't care about these restrictions 👍

https://github.com/woocommerce/woocommerce/blob/99f04562c084808e02515a526664fdc0c77fa688/plugins/woocommerce/client/blocks/assets/js/blocks/product-gallery/block.json#L21

cpapazoglou (approved):
Tested it and works well:

✅ I can edit an existing Product Gallery block in the site editor
✅ I can add a new one
✅ I can't add a Product Gallery block in a post or page
✅ I can add a Product Gallery block inside a Single Product block, in a post or page
✅ `BlockRegistrationManager` wasn't used anywhere else
✅ All references to `plugins/woocommerce/client/blocks/assets/js/atomic/utils/blocks-registration-manager/index.ts` have being removed
✅ Registering the Product Gallery block has [been simplified](https://github.com/woocommerce/woocommerce/pull/56459/files#diff-f322b9067ea3e4b28a05cbb5bf9ee44e5f0c0c4ad7ed5cecd98c76f58334f7d2R10-R21)
💪 420 LoC less

plugins/woocommerce/client/blocks/assets/js/atomic/utils/blocks-registration-manager/blocks-with-restriction.ts
kmanijak: Yeah. The `registerProductBlockType` from `plugins/woocommerce/client/blocks/assets/js/atomic/utils/register-product-block-type.ts` takes care of this:

```
allowedTemplates: {
			'single-product': true,
		},
```

so basically making the block available globally without ancestors in this particular template.

And this part:

```
		allowedTemplateParts: {
			'product-gallery': true,
		},
```

is no longer needed as we got rid of this template part some time ago. And to be fair, I think this was probably the main reason the whole mechanism was even implemented and running.

PR #56473, opened 2025-03-18
Move email sender options to the email template editor

### LINKED ISSUE #56123: Move the email sender email and name to the email editor
Image

Do this only if the feature flag is enabled `woocommerce_feature_block_email_editor_enabled`. Otherwise, keep the old behaviour. 

We want to keep the settings only in one place and that is the email editor. 

Add a notice banner to inform users that the settings have been moved into the Template editor. Add the message: "Email sender options have been moved. You can access these settings via Email Template." Use the default WordPress info message above the emails list, if user closes the notice, do not show it again.

Image

lysyjan:

Closes #56123.

- This PR moves the email sender options from `WooCommerce -> Settings -> Emails` to the new email template editor, which is behind a feature flag `block_email_editor`.
- When the feature flag `block_email_editor` is active, a user can see a notice in the `WooCommerce -> Settings -> Emails`.

| Before | After |
| ------ | ----- |
|Screenshot 2025-03-17 at 20 24 30 | Screenshot 2025-03-17 at 20 30 34|  
|Screenshot 2025-03-17 at 20 31 33 | Screenshot 2025-03-17 at 20 31 00 |
|Screenshot 2025-03-17 at 20 31 50|Screenshot 2025-03-17 at 20 38 38|
||Screenshot 2025-03-19 at 10 28 17|
||Screenshot 2025-03-19 at 10 28 36|

1. Check that saving the sender name and email address still works in `WooCommerce -> Settings -> Emails` when the feature `block_email_editor` is disabled.
2. Enable the feature flag `block_email_editor`
  a. A notice about moving the sender options is visible only in `Settings` and only for the `Emails` tab
  b. The notice is closable
  c. The notice is not visible anymore when the feature flag `block_email_editor` is deactivated
3. The sender options can be set in the email template editor after opening the sidebar related to the template
  a. When an invalid email address is typed, the input should display an HTML validation warning.
  b. Saving an invalid email address is not allowed, and a notice about an invalid email is displayed after clicking on save.

### Changelog entry

Because this feature is still behind the feature flag I think we don't need changelog here.
<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/templates/style.scss
+++ b/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/templates/style.scss
@@ -3,7 +3,7 @@
 		margin-top: 0;
 	}
 	p {
-		color: #757575;
+		color: $gray-700;
 	}
 }
 .woocommerce-email-sidebar-template-settings-sender-options-input {
--- a/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/templates/style.scss
+++ b/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/templates/style.scss
@@ -1,0 +1,12 @@
+.woocommerce-email-sidebar-template-settings-sender-options {
+	h2 {
+		margin-top: 0;
+	}
+	p {
+		color: #757575;
+	}
+}
+.woocommerce-email-sidebar-template-settings-sender-options-input {
+	margin-bottom: 10px;
+	width: 100%;
+}
--- a/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/templates/template_sender_panel.tsx
+++ b/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/templates/template_sender_panel.tsx
@@ -4,38 +4,43 @@
 import { __ } from '@wordpress/i18n';
 import { Panel, PanelBody, PanelRow, TextControl } from '@wordpress/components';
 import { useEntityProp } from '@wordpress/core-data';
-import { useRef } from '@wordpress/element';
+import { useCallback, useRef } from '@wordpress/element';
 
 function TemplateSenderPanel() {
 	const [ woocommerce_template_data, setWoocommerceTemplateData ] =
 		useEntityProp( 'postType', 'wp_template', 'woocommerce_data' );
 	const emailInputRef = useRef< HTMLInputElement >( null );
 
-	const handleFromNameChange = ( value: string ) => {
-		setWoocommerceTemplateData( {
-			...woocommerce_template_data,
-			sender_settings: {
-				...woocommerce_template_data?.sender_settings,
-				from_name: value,
-			},
-		} );
-	};
+	const handleFromNameChange = useCallback(
+		( value: string ) => {
+			setWoocommerceTemplateData( {
+				...woocommerce_template_data,
+				sender_settings: {
+					...woocommerce_template_data?.sender_settings,
+					from_name: value,
+				},
+			} );
+		},
+		[ woocommerce_template_data, setWoocommerceTemplateData ]
+	);
+	const handleFromAddressChange = useCallback(
+		( value: string ) => {
+			setWoocommerceTemplateData( {
+				...woocommerce_template_data,
+				sender_settings: {
+					...woocommerce_template_data?.sender_settings,
+					from_address: value,
+				},
+			} );
 
-	const handleFromAddressChange = ( value: string ) => {
-		setWoocommerceTemplateData( {
-			...woocommerce_template_data,
-			sender_settings: {
-				...woocommerce_template_data?.sender_settings,
-				from_address: value,
-			},
-		} );
-
-		// Use HTML5 validation
-		if ( emailInputRef.current ) {
-			emailInputRef.current.checkValidity();
-			emailInputRef.current.reportValidity();
-		}
-	};
+			// Use HTML5 validation
+			if ( emailInputRef.current ) {
+				emailInputRef.current.checkValidity();
+				emailInputRef.current.reportValidity();
+			}
+		},
+		[ woocommerce_template_data, setWoocommerceTemplateData ]
+	);
 
 	return (
 		<Panel className="woocommerce-email-sidebar-template-settings-sender-options">
--- a/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/templates/template_sender_panel.tsx
+++ b/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/templates/template_sender_panel.tsx
@@ -54,7 +54,8 @@ function TemplateSenderPanel() {
 				<PanelRow>
 					<TextControl
 						className="woocommerce-email-sidebar-template-settings-sender-options-input"
-						label={ __( '"from" name', 'woocommerce' ) }
+						/* translators: Label for the sender's `“from” name` in email settings. */
+						label={ __( '“from” name', 'woocommerce' ) }
 						name="from_name"
 						type="text"
 						value={
@@ -68,7 +69,8 @@ function TemplateSenderPanel() {
 					<TextControl
 						ref={ emailInputRef }
 						className="woocommerce-email-sidebar-template-settings-sender-options-input"
-						label={ __( '"from" email', 'woocommerce' ) }
+						/* translators: Label for the sender's `“from” email` in email settings. */
+						label={ __( '“from” email', 'woocommerce' ) }
 						name="from_email"
 						type="email"
 						value={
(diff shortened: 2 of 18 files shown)


plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/templates/template_sender_panel.tsx
```diff
@@ -0,0 +1,49 @@
...
+								'This is how your sender name and email address would appear in outgoing WooCommerce emails.',
+								'woocommerce'
+							) }
+						</p>
+					</div>
+				</PanelRow>
+				<PanelRow>
+					<TextControl
+						className="woocommerce-email-sidebar-template-settings-sender-options-input"
+						label={ __( '“from” name', 'woocommerce' ) }
```
costasovo, line 63: This translation would perhaps use a comment for translators. Maybe something like:
```
/* translators: Label for the sender's name in email settings. */
```
I would add it also for “from” email

plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/templates/style.scss
```diff
@@ -0,0 +1,12 @@
+.woocommerce-email-sidebar-template-settings-sender-options {
+	h2 {
+		margin-top: 0;
+	}
+	p {
+		color: #757575;
```
costasovo: I think you can also use `$gray-700` variable for that color.

plugins/woocommerce/src/Internal/EmailEditor/EmailTemplates/TemplateApiController.php
```diff
@@ -0,0 +1,67 @@
...
+		return Builder::object(
+			array(
+				'sender_settings' => Builder::object(
+					array(
+						'preheader'   => Builder::string(),
+						'preview_url' => Builder::string(),
+					)
+				),
+			)
+		)->to_array();
```
costasovo, line 83: Oh! Nice. I didn't realize we brought the builder with the package. 👍

plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/templates/template_sender_panel.tsx
```diff
@@ -3,8 +3,32 @@
...
+
+	const handleFromAddressChange = ( value: string ) => {
+		setWoocommerceTemplateData( {
+			...woocommerce_template_data,
+			sender_settings: {
+				...woocommerce_template_data?.sender_settings,
+				from_address: value,
+			},
+		} );
+	};
```
costasovo: We could add useCallback to avoid redefining the functions on each re-render.

plugins/woocommerce/src/Internal/EmailEditor/EmailTemplates/TemplateApiController.php
```diff
@@ -0,0 +1,67 @@
...
+	 * Update WooCommerce specific data we store with Template.
+	 *
+	 * @param array              $data - WP_Block_Template data.
+	 * @param \WP_Block_Template $template_post - WP_Block_Template object.
+	 */
+	public function save_template_data( array $data, \WP_Block_Template $template_post ): void {
+		if ( WooEmailTemplate::TEMPLATE_SLUG === $template_post->slug && isset( $data['sender_settings'] ) ) {
+			update_option( 'woocommerce_email_from_name', $data['sender_settings']['from_name'] );
+			update_option( 'woocommerce_email_from_address', $data['sender_settings']['from_address'] );
+		}
```
costasovo: Just out of curiosity. I see we don't check the `from_name` key is present in `$data['sender_settings']['from_name']`. Does the API do this via the schema? I assume it does, but I'm not sure.

plugins/woocommerce/includes/admin/settings/class-wc-settings-emails.php
```diff
@@ -277,6 +277,8 @@ protected function get_settings_for_default_section() {
 		$body_text_color_setting_in_palette   = $reorder_colors ? $body_text_color_setting : null;
 		$footer_text_color_setting_in_palette = $reorder_colors ? $footer_text_color_setting : null;
 
+		$block_email_editor_enabled = get_option( 'woocommerce_feature_block_email_editor_enabled', 'no' );
```
costasovo: I think we should rather use  `FeaturesUtil::feature_is_enabled( 'block_email_editor' )` for this check, and we get `true` or `false`. I think it is better than checking the `no` string as it may theoretically be '0' or something else.

costasovo (commented):
@lysyjan Great job! I left a few comments, but they are just tiny details. Thanks for adding the unit test. There is also one question or possible issue I run into during testing. Please see below:

### Testing performed

✅ When the feature flag is disabled saving of sender's name and address works as expected
✅ After enabling the flag, I can see the notice and the sender's name and inputs are no longer displayed. Note: I noticed that the link Email Template goes to the email listing. I think that's fine for now because as for now we can't link the template mode directly.
❓  I noticed that when I dismiss the notice and switch between tabs in the settings the notice pops up again. I would expect it to stay hidden. Is that expected?
✅ I tested saving the sender's email and name settings in the editor and it works as expected.
✅ I tested adding an invalid email and validation worked as expected.
✅ After disabling the feature flag the notice is not displayed any more.

**Video:**
I started with a feature flag `off` then set it 'on` and towards the end I set it again `off` to verify that the notice is gone. I didn't record the validation testing so I'm providing a screenshot.

https://github.com/user-attachments/assets/32d32175-42fc-4f17-8912-267b5c00c116

**Screenshot of validation message:**

Screenshot 2025-03-20 at 11 46 53

I also noticed a warning about the script size, but I consider that fine as it is a new script, and the test obviously checks the percentage. Overall size IMO grew accordingly.

Screenshot 2025-03-20 at 11 53 38

plugins/woocommerce/includes/admin/settings/class-wc-settings-emails.php
lysyjan: Thanks for this. I had it in my mind, but in the end, I forgot.

plugins/woocommerce/src/Internal/EmailEditor/EmailTemplates/TemplateApiController.php
lysyjan: This is good point. The schema does not check it because the endpoint should update data only for the specific template only. I will add a code with some checks.

lysyjan:
@costasovo, that was a good catch with the non-dismissable notice. I fixed it.
New commits start at commit [a37ad7a](https://github.com/woocommerce/woocommerce/pull/56473/commits/a37ad7a62dd099e92db8bc0940438099bab5623b)

plugins/woocommerce/includes/admin/class-wc-admin-notices.php
```diff
@@ -673,7 +673,7 @@ public static function email_sender_options_notice() {
 			 */
 			$notice_dismissed = apply_filters(
 				'woocommerce_hide_email_sender_options_notice',
-				get_user_meta( get_current_user_id(), 'dismissed_block_email_editor_notice', true )
+				get_user_meta( get_current_user_id(), 'dismissed_email_sender_options_notice', true )
```
costasovo: I missed this in the first review. I see that in the newer code in the [src](https://github.com/search?q=repo%3Awoocommerce%2Fwoocommerce+get_user_meta+path%3A%2F%5Eplugins%5C%2Fwoocommerce%5C%2Fsrc%5C%2F%2F+&type=code) the user meta is often prefixed `woocommerce_`. Perhaps we could do that as well so that it is less conflict-prone.

costasovo (changes requested):
@lysyjan Thanks for the update.

I added one comment with a suggestion, but when I started testing I noticed a visual issue in the settings > email page.

When the I activate the feature flag, I see that the preview is partially over the emails table.
It is possible that the issue was present in the first testing I did, but I had a smaller window so I missed it.

I found this was caused by commit [87235c0b01cccaf6e98fc6c4f5f5eaa6a8cb65a6](https://github.com/woocommerce/woocommerce/pull/56473/commits/87235c0b01cccaf6e98fc6c4f5f5eaa6a8cb65a6)

Screenshot 2025-03-21 at 11 22 03

I think we need to address the issue.

lysyjan:
@costasovo I added a new commit renaming the option. I also rebased the branch and cannot replicate the issue on the email settings page anymore.

costasovo (approved):
@lysyjan Thank you for updating the name of the option!

I wasn't able to replicate the issue with the preview. So it was probably only a glitch on my site.

### Testing performed

Video of basic testing:

https://github.com/user-attachments/assets/97f7526c-fc02-4091-9eab-f3cbe54405b7

Screenshot of received email with updated sender values:
Screenshot 2025-03-30 at 16 14 47

Issue #31728, opened 2022-01-25
coupons for individual use isnt working as it should


pipin3000:
### Prerequisites

- [X] I have carried out troubleshooting steps and I believe I have found a bug.
- [X] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

I am creating two discount codes for individual use only which shouldn't allow users to use more than one discount code at a time but if a user types in another discount code, the first discount code is simply overridden and the new one is applied.

### Expected behavior

sorry coupon has already been applied and cannot be used in conjunction with other coupons.

### Actual behavior

Original discount code is over ridden and the new one is applied.

### Steps to reproduce

https://www.staging39.myconsumables.com/

Add items to the basket

at basket use discount code "test10" & "test20". Both of these are set for individual use only.

### WordPress Environment ###

WordPress address (URL): https://www.staging39.myconsumables.com
Site address (URL): https://www.staging39.myconsumables.com
WC Version: 6.1.1
REST API Version: ✔ 6.1.1
WC Blocks Version: ✔ 6.5.2
Action Scheduler Version: ✔ 3.4.0
WC Admin Version: ✔ 3.0.3
Log Directory Writable: ✔
WP Version: 5.8.3
WP Multisite: –
WP Memory Limit: 768 MB
WP Debug Mode: –
WP Cron: –
Language: en_GB
External object cache: –

### Server Environment ###

Server Info: Apache
PHP Version: 7.4.27
PHP Post Max Size: 256 MB
PHP Time Limit: 120
PHP Max Input Vars: 3000
cURL Version: 7.66.0
OpenSSL/1.1.1l-fips

SUHOSIN Installed: –
MySQL Version: 5.7.32-35-log
Max Upload Size: 256 MB
Default Timezone is UTC: ✔
fsockopen/cURL: ✔
SoapClient: ✔
DOMDocument: ✔
GZip: ✔
Multibyte String: ✔
Remote Post: ✔
Remote Get: ✔

### Database ###

WC Database Version: 6.1.1
WC Database Prefix: pka_
Total Database Size: 143.06MB
Database Data Size: 96.11MB
Database Index Size: 46.95MB
pka_woocommerce_sessions: Data: 1.17MB + Index: 0.02MB + Engine InnoDB
pka_woocommerce_api_keys: Data: 0.02MB + Index: 0.03MB + Engine InnoDB
[...]

PR #56488, opened 2025-03-18
Update coupon error messages - explicitly mention coupon code

### LINKED ISSUE #56487: [Enhancement]: Mention coupon code in coupon error messages

Ideally, we'd like to add a new field to error responses returned from the POST v3/order (order creation) when the creation fails due to an invalid coupon. This would be used in the WooCommerce mobile apps/POS.

However, since the mobile apps communicate through the jetpack tunnel in some cases (`public-api.wordpress.com/rest/v1.1/jetpack-blogs/[SITE_ID]/rest-api/`) this approach isn't currently viable, because the proxy strips away all fields from the response except of `code` and `message`.

Therefore, an alternative solution is to explicitly mention the coupon code in the error message itself.

_No response_

### Additional context

pdfdoF-6Ia-p2

malinajirka:

This PR improves the error messages shown when applying or adding a coupon to an order fails. Previously, most messages did not explicitly mention the coupon code that caused the failure, making it hard to re-use the API in other clients of the API such as the mobile apps. The goal of this PR is to include the specific coupon code in all relevant error messages.

Additionally, some of the modified lines had pre-existing suppressed lint warnings. To satisfy all lint checks, those issues have also been addressed as part of this PR. Most of the additional changes involve properly escaping inputs such as price, coupon code, and URL.

⚠️ [As explained in this comment, this PR introduces HTML entities to all the coupon error messages - specifically HTML tags for quotes](https://github.com/woocommerce/woocommerce/pull/56488#discussion_r2010207091). I've tested the front-end and we seem to be decoding them properly so this shouldn't cause any regressions AFAICT. Let me know what you think, thanks!

🧪 The tests are failing because of the above HTML entities - I'll update them when we make a final decision whether the quotes should be as part of the translatable string or part of the coupon token.

📓 [There is a known issue related to coupons with `Individual use only` flag on.](https://github.com/woocommerce/woocommerce/issues/31728)

📓 I noticed the HTML tags, such as `<a>` are ignored on the UI. I guess they were added historically and we never removed them?

Closes #56487 .

### Invalid Coupon  
1. Add any product to the cart.  
2. Click **View Cart**.  
3. Enter random text into the **Coupon code** input and click **Apply coupon**.  
4. A well-formatted error message should be displayed, indicating the coupon is invalid.

---

### $10 Off Coupon  
1. In **WP Admin**, go to **Marketing → Coupons**.  
2. Click **Create Coupon**.  
3. Set the **Coupon Amount** to `$10`.  
4. Set the **Coupon Code** to `ten-off`.  
5. Click **Publish**.  
6. On the store's front-end, add any product to the cart.  
7. Click **View Cart**.  
8. Enter `ten-off` in the **Coupon code** input and click **Apply coupon**.  
9. The coupon should be successfully applied, and a $10 discount should be reflected in the cart total.

---

### 🧪 Additional Validations  

Repeat the above steps with the following coupon restrictions and verify that appropriate error messages are shown:

- **Minimum Spend**:  
  - Set a minimum spend on the coupon.  
  - Try applying it with a cart total below the limit.  
  - A clear error message should indicate that the minimum spend requirement is not met.

- **Maximum Spend**:  
  - Set a maximum spend on the coupon.  
  - Try applying it with a cart total above the limit.  
  - A proper error message should be shown.

- **Specific Products**:  
  - Restrict the coupon to specific products.  
  - Add a product *not* in the allowed list and attempt to apply the coupon.  
  - A relevant error message should be displayed.

- **Allowed Emails – Not Set**:  
  - Restrict the coupon to specific email addresses.  
  - Try applying the coupon without setting a billing email.  
  - A proper error message should be shown.

- **Allowed Emails – Mismatch**:  
  - Set allowed emails on the coupon.  
  - Use a billing email that does *not* match.  
  - An appropriate error message should be displayed.

- **Usage Limit: 1 per Coupon**:  
  - Set the usage limit to **1**.  
  - Apply the coupon once—verify it works.  
  - Create a new order and apply the same coupon again—verify a clear error is shown due to the usage limit being exceeded.
  - 📓 🐛 This one breaks the layout a bit.

Test any other scenarios you can think of.

I've tested all of the above on both the UI as well as using a rest-api testing tool.


plugins/woocommerce/includes/class-wc-coupon.php
```diff
@@ -1021,61 +1021,67 @@ public function get_coupon_message( $msg_code ) {
 	public function get_coupon_error( $err_code ) {
 		switch ( $err_code ) {
 			case self::E_WC_COUPON_INVALID_FILTERED:
-				$err = __( 'Coupon is not valid.', 'woocommerce' );
+				/* translators: %s: coupon code */
+				$err = sprintf( esc_html__( 'Coupon %s is not valid.', 'woocommerce' ), '"' . esc_html( $this->get_code() ) . '"' );
```
coreymckrill:
```suggestion
				$err = sprintf(
					esc_html__( 'Coupon "%s" is not valid.', 'woocommerce' ),
					esc_html( $this->get_code() )
				);
```

One thing I would recommend is making the `sprintf` multi-line. It's way easier to read that way. Also, I think the quotes would be better in the translatable string rather than in the token. Some languages use other characters besides quotes for things like this.

plugins/woocommerce/includes/class-wc-coupon.php
```diff
@@ -1021,61 +1021,67 @@ public function get_coupon_message( $msg_code ) {
...
+				/* translators: %s: coupon code */
+				$err = sprintf( esc_html__( 'Sorry, coupon %s is not applicable to your cart contents.', 'woocommerce' ), '"' . esc_html( $this->get_code() ) . '"' );
 				break;
 			case self::E_WC_COUPON_USAGE_LIMIT_COUPON_STUCK:
 				if ( is_user_logged_in() && wc_get_page_id( 'myaccount' ) > 0 && ! WC()->is_store_api_request() ) {
-					/* translators: %s: myaccount page link. */
-					$err = sprintf( __( 'Coupon usage limit has been reached. If you were using this coupon just now but your order was not complete, you can retry or cancel the order by going to the <a href="%s">my account page</a>.', 'woocommerce' ), wc_get_endpoint_url( 'orders', '', wc_get_page_permalink( 'myaccount' ) ) );
+					/* translators: %1$s: coupon code, %2$s: myaccount page link. */
+					$err = sprintf( wp_kses_data( __( 'Usage limit for coupon "%1$s" has been reached. If you were using this coupon just now but your order was not complete, you can retry or cancel the order by going to the <a href="%2$s">my account page</a>.', 'woocommerce' ) ), esc_html( $this->get_code() ), wc_get_endpoint_url( 'orders', '', wc_get_page_permalink( 'myaccount' ) ) );
```
coreymckrill:
```suggestion
					$err = sprintf( wp_kses_data( __( 'Usage limit for coupon "%1$s" has been reached. If you were using this coupon just now but your order was not complete, you can retry or cancel the order by going to the <a href="%2$s">my account page</a>.', 'woocommerce' ) ), esc_html( $this->get_code() ), esc_attr( wc_get_endpoint_url( 'orders', '', wc_get_page_permalink( 'myaccount' ) ) ) );
```

malinajirka: Thanks, fixed in 5fc1b38c35b57af32c81171ee3caf90a409dc0b0.

plugins/woocommerce/includes/class-wc-coupon.php
malinajirka: > Some languages use other characters besides quotes for things like this.

That's a good point, I didn't realize that. 

I actually had them in the string originally, but moved them out since `esc_html__` was replacing them with `&quot;`. 

This seems fine for some of the respones, eg.

Before changes
```
{
"error": "woocommerce_rest_invalid_coupon",
"message": "The minimum spend for this coupon is <span class=\"woocommerce-Price-amount amount\"><bdi><span class=\"woocommerce-Price-currencySymbol\">&#36;</span>50.00</bdi></span>."
}
```

After changes
```
{
"error": "woocommerce_rest_invalid_coupon",
"message": "The minimum spend for coupon &quot;20_dollars_off_cart_min_amount_50&quot; is <span class=\"woocommerce-Price-amount amount\"><bdi><span class=\"woocommerce-Price-currencySymbol\">&#036;</span>50.00</bdi></span>."
}
```

Since the message already contains html entities it seems ok to assume the client will decode them. (I personally believe this doesn't belong to REST API, but that's another story)

However, for other messages, this assumption seems quite risky, for example:

Before changes
```
{
"error": "woocommerce_rest_invalid_coupon",
"message": "Coupon \"wrong_coupon_code\" does not exist!"
}
```

After changes
```
{
"error": "woocommerce_rest_invalid_coupon",
"message": "Coupon &quot;wrong_coupon_code&quot; does not exist!"
}
```


Wdyt @coreymckrill ?

coreymckrill: 🤔 What makes the second example risky? Is it that the `&quot;` HTML entity might be interpreted as part of the coupon code?

malinajirka: Just the fact that the message contains a HTML entity now even though it didn't contain it before => the client needs to interpret it as HTML instead of just a plain text.

coreymckrill: Ah, I see what you mean. I think it's a fairly common pattern in our APIs to have HTML entities like this because of escaping, so it may be worth adapting the client to be able to handle it. But a couple of other options:

* Use `wp_kses` with an empty array for the "allowed tags" parameter, so that all HTML tags get stripped (but quotes don't get converted)
* Don't wrap the coupon code in quotes at all

malinajirka: 👍 Let's keep it simple and remove the quotes. They don't seem necessary anyway and will reduce risks of introducing some issues in translations.

malinajirka: Actually, removing the quotes might not be the best option - we are using them across the codebase, even for successful messages. Considering that in my tests the quotes didn't cause any troubles on the UI, let's continue escaping them as we have been escaping them. The worst case scenario, they won't be de-escaped which is not a great UX but also not a critical error.

coreymckrill:
@malinajirka There's some e2e tests failing, I suspect it is related to the quote thing. Either the test doesn't wrap the variable in quotes, and/or maybe it should be expecting the HTML entities instead of actual quotes?

malinajirka:
> @malinajirka There's some e2e tests failing, I suspect it is related to the quote thing. Either the test doesn't wrap the variable in quotes, and/or maybe it should be expecting the HTML entities instead of actual quotes?

I was fixing these on Friday but ran out of time, sorry I should have commented here that it's still on my todo-list.

Either case, I have looked into it today and there seems to be a bug in my code - one of the UI tests is failing and looking at the video, the behavior is truly not expected => the run displays `Coupon successfully applied` even though the same coupon is already applied to the order and `Coupon code ... already applied!` error should have been displayed instead. 

I'll continue looking into this. When I ran the same test manually on my store, it seems to work as expected. Sharing here in case you'd have some ideas, what might be going on.

malinajirka:
@coreymckrill All the tests related to this PR are fixed now. There is one failing test, but it doesn't seem to be related to this PR since it's failing in other PRs as well (eg. [here](https://github.com/woocommerce/woocommerce/actions/runs/14168660647/job/39687400305)).

Let me know how it all looks to you. I'm AFK for the next two weeks starting tomorrow, but I'll try to check on this PR in case there are some minor suggestions for updates. Of course, feel free to make them yourself.

Thank you for all the tips and feedback so far 🙇!!!

Konamiman:
Hi @malinajirka, just to clarify, is this PR completed? Asking because the PR description mentions pending changes in the tests.

staskus:
Thanks for taking a look at this @Konamiman. Yes, [the failing tests were addressed](https://github.com/woocommerce/woocommerce/pull/56488#issuecomment-2765766568) a few days after the description was written.

Konamiman:
I have tested the scenarios listed in the pull request description and then some more:

- Coupon is already applied to the order
- Coupon has expired
- Coupon is not valid for sale items
- Coupon can't be used with other coupons
- Coupon has excluded products
- Coupon has excluded categories

Konamiman (approved):

malinajirka:
Thank you both for helping wrap this up!

senadir:
Hey @malinajirka, tests are still broken for this PR, which tests do you update? Searching `Coupon code already applied!` in the codebase surfaces 8 other places that need updates. Not sure why they didn't block this PR.

senadir:
Please also ping @woocommerce/rubik when you're working on Cart/Checkout related work :)

PR #56511, opened 2025-03-19
Fix error in new hook names

Dekadinious:
## Description

Fix incorrect hook names in the deprecated filter hooks mapping. The keys `woocommerce_get_order_refund_get_amount` and `woocommerce_get_order_refund_get_reason` don't match the actual filters fired by WooCommerce — they should be `woocommerce_order_refund_get_amount` and `woocommerce_order_refund_get_reason` (without the extra `get_` prefix).

Because the constructor registers listeners on these keys, the mapping is completely dead — anyone using the old deprecated hooks `woocommerce_refund_amount` / `woocommerce_refund_reason` gets neither a deprecation notice nor their callback executed.

## Testing instructions

1. Add the following snippet to a plugin or theme's `functions.php`:

```php
add_filter( 'woocommerce_refund_amount', function ( $amount ) {
    error_log( 'Deprecated refund amount filter fired: ' . $amount );
    return $amount;
}, 10, 1 );
```

2. Create a test order and issue a refund.

**Before fix:** No deprecation notice is logged and the callback never fires — the deprecated hook is silently ignored because the mapping key (`woocommerce_get_order_refund_get_amount`) doesn't match any real filter, so the listener registered in the constructor never triggers.

**After fix:** A deprecation notice is logged (the old hook is correctly mapped from `woocommerce_order_refund_get_amount`) and the callback fires as expected.

### Changelog

Added in `plugins/woocommerce/changelog/fix-56511-deprecated-refund-hook-names`.


jorgeatorres (changes requested):
Hi @Dekadinious!

Thank you for your contribution. This seems correct based on how the filters changed from 2.6.x to 3.0.x, but even if the change is small we'll need a changelog entry and proper testing instructions. A small PHP snippet showing how the "deprecated" filter gets translated into the correct one would be enough.

Looking forward to hearing back from you.

Dekadinious:
Hey @jorgeatorres, sorry for the long delay on this — life got in the way. I've added a changelog entry and updated the description with testing instructions and a PHP snippet showing the fix in action. Let me know if there's anything else needed!

jorgeatorres (approved):
Hey @Dekadinious, no worries at all! Really appreciate you coming back to this and taking the time to add the changelog entry and the testing instructions. Thanks for your contribution!

PR #56516, opened 2025-03-19
Add setting options data store for managing settings

chihsuan:

This PR introduces a new `settings-options` data store that provides a centralized way to manage WooCommerce settings. The store handles settings groups, individual settings, unsaved changes, and error states.

#### Key features

- Manage settings groups and individual settings
- Handle unsaved changes with edit states
- Batch update capabilities
- Properly handle errors for failed updates

#### Why this is needed?

We've deprecated the `option` data store in favor of creating individual endpoints for each usage. However, having a new endpoint for each setting is ineffective. I was trying to [migrate the core profiler to use dedicated endpoints for saving store settings](https://github.com/woocommerce/woocommerce/pull/56347) but realized that we can reuse the existing [WC `setting options` API](https://woocommerce.github.io/woocommerce-rest-api-docs/#setting-options).

With this new data store, we can easily manage settings groups and individual settings. It's also useful for new modern Settings pages except for replacing the existing Settings/Options store.

#### Why introduce a new data store rather than use the existing `settings` store?

The existing `settings` store was introduced for the same purpose. However, it has many issues:

1. The method signatures are confusing. For example, `select( settingsStore ).getSettings('general').general;` is not very intuitive and `wp.data.select( wc.data.settingsStore ).getSetting( 'general', 'woocommerce_default_country' )`
2. It's mixed with other unrelated settings because we preload all wcSettings.admin to the store, so you can use `resolveSelect( settingsStore ).getSettings( 'wc_admin').shopUrl` to get the store URL, but you can't update it since it's setting in WC > Settings.
3. It doesn't handle errors correctly. For example, when updating settings in a batch and some fail, it doesn't return the error. Instead, it considers API success.

Besides, I [tried to enhance the existing `settings` store](https://github.com/woocommerce/woocommerce/pull/56478), but it's pretty complicated since we also need to ensure backward compatibility. I believe it's easier to create a new data store for this and deprecate the existing one in the future.

**This PR doesn't add/update any new user-facing features. I'll have some follow-up PRs to migrate the code to use this new data store.**

Review the [REAMDE.md](https://github.com/woocommerce/woocommerce/blob/add/settings-option-store/packages/js/data/src/setting-options/README.md)

1. Create a fresh site 
2. Go to any WooCommerce Admin page
3. Open dev console and test the following commands:

#### Basic Operations

```js
wp.data.resolveSelect( wc.data.experimentalSettingOptionsStore ).getGroups().then(console.log)
// should print an array of groups

wp.data.resolveSelect( wc.data.experimentalSettingOptionsStore ).getGroup( 'general' ).then(console.log)
// should print an object of general group

wp.data.resolveSelect( wc.data.experimentalSettingOptionsStore ).getSettings( 'general' ).then(console.log)
// should print an object of general settings (e.g. woocommerce_default_country...)
```

#### Single Setting Operations

```js
wp.data.resolveSelect( wc.data.experimentalSettingOptionsStore ).getSetting( 'general', 'woocommerce_currency' ).then(console.log)
// Should print the woocommerce_currency setting object

wp.data.resolveSelect( wc.data.experimentalSettingOptionsStore ).getSettingValue( 'general', 'woocommerce_currency' ).then(console.log)
// Should print the woocommerce_currency setting value
```

#### Batch Operations

```js
// Update multiple settings at once
wp.data.dispatch( wc.data.experimentalSettingOptionsStore ).editSettings( 'general', { woocommerce_currency: 'JPY', 'woocommerce_default_country': 'GB'  } )

wp.data.select( wc.data.experimentalSettingOptionsStore ).getEditedSettingIds( 'general')
// Should return  ['woocommerce_currency', 'woocommerce_default_country']

wp.data.dispatch( wc.data.experimentalSettingOptionsStore ).saveEditedSettingsGroup( 'general' );
// Confirm it send the API request to update the settings and return 200 status in network tab
```

#### Direct Save Operations

```js
wp.data.dispatch( wc.data.experimentalSettingOptionsStore ).saveSettingsGroup( 'general', { woocommerce_currency: 'USD', 'woocommerce_default_country': 'US:CA'  } )
// Confirm it send the API request to update the settings and return 200 status in network tab

wp.data.dispatch( wc.data.experimentalSettingOptionsStore ).saveSetting( 'general', 'woocommerce_currency', 'JPY' )
// Confirm it send the API request to update the settings and return 200 status in network tab
```

#### Error Handling

```js
wp.data.dispatch( wc.data.experimentalSettingOptionsStore ).saveSetting( 'general', 'woocommerce_currency', '' )
// Confirm it throw an error in console

wp.data.select( wc.data.experimentalSettingOptionsStore ).getGroupError( 'general' )
// Confirm it return the object with key `woocommerce_currency`

wp.data.select( wc.data.experimentalSettingOptionsStore ).getSettingError( 'general', 'woocommerce_currency')
// Confirm it return the error object
```

#### Testing Concurrent Operations

```js
// Refresh the page to reset the state
// Test 1: Concurrent save and fetch
Promise.all([
  wp.data.dispatch( wc.data.experimentalSettingOptionsStore ).saveSetting( 'general', 'woocommerce_currency', 'JPY' ),
  wp.data.resolveSelect( wc.data.experimentalSettingOptionsStore ).getSetting( 'general', 'woocommerce_currency' )]
).then(console.log);
// Check network tab to observe the request order (screenshot below) and final state

// Test 2: Multiple concurrent saves
Promise.all([ 
  wp.data.dispatch( wc.data.experimentalSettingOptionsStore ).saveSetting( 'general', 'woocommerce_currency', 'JPY' ),
  wp.data.dispatch( wc.data.experimentalSettingOptionsStore ).saveSetting( 'general', 'woocommerce_currency', 'USD' )
]).then(console.log);
// Verify final currency value and check network requests

// Test 3: Edit during save
wp.data.dispatch( wc.data.experimentalSettingOptionsStore ).saveSetting( 'general', 'woocommerce_currency', 'JPY' );
wp.data.dispatch( wc.data.experimentalSettingOptionsStore ).editSetting( 'general', 'woocommerce_currency', 'USD' );
// Check if edit is preserved or overwritten by save response
wp.data.select( wc.data.experimentalSettingOptionsStore ).getEditedSettingIds( 'general' )
// Should return ['woocommerce_currency']
```

❌
Screenshot 2025-03-26 at 14 35 33

✅ 
Screenshot 2025-03-26 at 14 36 45

### Expected Results for Concurrent Operations

1. When running save and fetch concurrently, the fetch should return the most up-to-date value
2. When running multiple saves concurrently, the last save operation should be reflected in the final state
3. When editing during a save operation, the edit should not be lost due to the save response

### Things to Watch For:

- Check Network tab to observe request timing and order
- Monitor console for any errors or race conditions
- Verify final state matches expected values

Please try different combinations of dispatching actions and select the data you want to test.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


packages/js/data/src/setting-options/actions.ts
```diff
@@ -0,0 +1,361 @@
+/**
+ * External dependencies
+ */
+import apiFetch from '@wordpress/api-fetch';
+import { DispatchFromMap } from '@automattic/data-stores';
```
rjchow: I suspect the types in `@automattic/data-stores` are from before `@wordpress/data` started exporting its own types, so using the types from there will result in conflicts at a later time - do you know if that's the case?

packages/js/data/src/setting-options/actions.ts
```diff
@@ -0,0 +1,361 @@
...
+	BatchSettingsError,
+	SettingsState,
+	SettingUpdate,
+	SettingsUpdateObject,
+} from './types';
+import { NAMESPACE } from '../constants';
+
+type WPDataRegistry = ReturnType< typeof createRegistry >;
+type AllSelectors = typeof import('./selectors');
+type CurriedState< F > = F extends (
```
rjchow, line 31: I thiiiiiiiink you can get this from `@wordpress/data`

packages/js/data/src/setting-options/actions.ts
```diff
@@ -0,0 +1,361 @@
...
+	state: SettingsState,
+	...args: infer P
+) => infer R
+	? ( ...args: P ) => R
+	: F;
+type Selectors = {
```
rjchow: This too

packages/js/data/src/setting-options/selectors.ts
```diff
@@ -0,0 +1,213 @@
...
+ * @param {string}        groupId - The ID of the group to get.
+ * @return {SettingsGroup | undefined} The group if found, otherwise undefined.
+ */
+export const getGroup = (
+	state: SettingsState,
+	groupId: string
+): SettingsGroup | undefined =>
+	state.groups.find( ( group ) => group.id === groupId );
+
+// Ensure we have a consistent empty object to return when there are no settings.
```
rjchow, line 47: Is this for checking for referential equality?

packages/js/data/src/setting-options/tests/actions.test.ts
```diff
@@ -0,0 +1,615 @@
...
+			await expect(
+				registry.dispatch( STORE_NAME ).saveSettingsGroup( groupId )
+			).rejects.toThrow( 'Failed to update some settings' );
+
+			// Verify successful update
+			expect( store.getState().settings[ groupId ]?.setting1 ).toEqual(
+				mockResults.update[ 0 ]
+			);
+
+			// Verify error state
```
rjchow, line 404: In conjunction with the other comment about not caching error values, if that's accepted then we should also verify that the returned value is correct here

rjchow (commented):
- after an error in saving, e.g saving woocommerce_currency to '', the getSetting and getSettingsValue returns the invalid value - I think the cache should not be saved if the saving fails?
- should updateSettings be stageSettingUpdate? I think its slightly unintuitive that an imperative verb like 'update' doesn't persist the settings by default, and requires a parameter to make it save
- when doing a refresh after an error occurs, this weird thing shows up in the console not really sure if its an issue

image

Otherwise, really nice job! I think the API is quite comprehensively thought out and the test suite is very good. I wonder if we should mark this experimental somehow while we iterate on it? I reckon we might learn that we'll need some changes when we work on the modern Settings work

packages/js/data/src/setting-options/types.ts
```diff
@@ -0,0 +1,160 @@
+/**
+ * Represents the value of a setting. It's unknown because any value can be stored.
+ */
+export type SettingValue = unknown;
```
rjchow, line 4: Gut feel that having it as unknown might result in a bunch of type overrides but I'm happy to leave it alone for now and see if it proves true

packages/js/data/src/setting-options/actions.ts
chihsuan: We only export ActionDispatchers without overriding @wordpress/data types, so I believe using the types from there is safe. But actually, I think we don't need this. It was needed for the old way of doing things. I've removed it!

packages/js/data/src/setting-options/actions.ts
chihsuan: Thanks! Updated in 4711d41d253a6383ce37626d6423f75e1cc7bbc2.

packages/js/data/src/setting-options/selectors.ts
chihsuan: Yes, selectors need to return a scalar or a consistent reference. https://developer.wordpress.org/news/2024/03/how-to-work-effectively-with-the-useselect-hook

packages/js/data/src/setting-options/types.ts
chihsuan: It appears that the majority of settings values are strings, although `wc_admin` utilizes `string[]` for a few. Additionally, it seems that any type of value can be registered so I left it as unknown for now.

chihsuan:
Thanks for the review! @rjchow 🙏 

> * after an error in saving, e.g saving woocommerce_currency to '', the getSetting and getSettingsValue returns the invalid value - I think the cache should not be saved if the saving fails?

If a user changes a setting and the save operation fails, I think the edited state should still be cached instead of clearing it so that the user can see the edited value with the error message. Does this make sense? I think it's common for users to edit an input field, and then the save operation fails, but they still see the edited value.

> * should updateSettings be stageSettingUpdate? I think its slightly unintuitive that an imperative verb like 'update' doesn't persist the settings by default, and requires a parameter to make it save

Sounds good! What do you think if we rename `updateSettings` to `editSettings`? WP core data has similar pattern, e.g. `saveEntityRecord` and `editEntityRecord`. I'm not very sure if this is a good idea from a natural language perspective. :thinking:

> * when doing a refresh after an error occurs, this weird thing shows up in the console not really sure if its an issue
> 
> image
> 

I'll test more and see if I can reproduce it. I remember that if you have redux devtools open, it will show some weird things because of the extension cache.

chihsuan:
> I wonder if we should mark this experimental somehow while we iterate on it? I reckon we might learn that we'll need some changes when we work on the modern Settings work


Good idea, I'll export the store as experimental and add a note to the readme.

rjchow:
> > * after an error in saving, e.g saving woocommerce_currency to '', the getSetting and getSettingsValue returns the invalid value - I think the cache should not be saved if the saving fails?
> 
> If a user changes a setting and the save operation fails, I think the edited state should still be cached instead of clearing it so that the user can see the edited value with the error message. Does this make sense? I think it's common for users to edit an input field, and then the save operation fails, but they still see the edited value.

I think this use case makes sense - do we need to handle the situation where multiple instances of select/dispatch on the same store is being used at the same time?


> > * should updateSettings be stageSettingUpdate? I think its slightly unintuitive that an imperative verb like 'update' doesn't persist the settings by default, and requires a parameter to make it save
> 
> Sounds good! What do you think if we rename `updateSettings` to `editSettings`? WP core data has similar pattern, e.g. `saveEntityRecord` and `editEntityRecord`. I'm not very sure if this is a good idea from a natural language perspective. 🤔
 
I think that's fine!

chihsuan:
> > If a user changes a setting and the save operation fails, I think the edited state should still be cached instead of clearing it so that the user can see the edited value with the error message. Does this make sense? I think it's common for users to edit an input field, and then the save operation fails, but they still see the edited value.
> 
> I think this use case makes sense - do we need to handle the situation where multiple instances of select/dispatch on the same store is being used at the same time?

Hmm, good point. I'll need to think about this.

> > Sounds good! What do you think if we rename `updateSettings` to `editSettings`? WP core data has similar pattern, e.g. `saveEntityRecord` and `editEntityRecord`. I'm not very sure if this is a good idea from a natural language perspective. 🤔
> 
> I think that's fine!

Ok, I've renamed `updateSetting*` to `editSetting*`.

packages/js/data/src/setting-options/actions.ts
```diff
@@ -0,0 +1,345 @@
...
+			const successfulUpdates: Setting[] = [];
+			const errors: Array< { id: string; error: APIError } > = [];
+
+			results.update.forEach( ( result ) => {
+				if (
+					'error' in result &&
+					result.error &&
+					typeof result.error === 'object' &&
+					'code' in result.error
+				) {
```
psealock: Is every error guaranteed to always have a `code` property? I'm wondering if this check is overly specific and may miss an error.

psealock (approved):
This is really nice @chihsuan, thanks for making it happen.

I tested on JN instance by calling selectors and dispatches via the console. Just a small question, but no real issues pop out - this is testing well.

packages/js/data/src/setting-options/actions.ts
chihsuan: I confirmed that the WC API ensures every error includes a code property. But your concern is valid. The check might be too stringent. I've removed it in 323a842f9659051a0e997db3c10cfa1e6fb97c79.

chihsuan:
Hey @rjchow, I've made some changes: 

1.  Modified the logic to return original settings or edited values based on the `includeEdits` option so developers can decide whether to get edited values.
2. Added new actions to save settings directly instead of using the { save } option in edit* actions.
3. Integrated lock actions from core data to prevent concurrent operations on the same setting. I imported the `createLocksActions` directly from core data build folder. Alternatively, we can copy the code from Core Data to our package.
4. Some naming changes to make it more consistent and easier to understand.

Could you review the changes again? Let me know if you have any questions or feedback. 🙏 

Thanks!

> when doing a refresh after an error occurs, this weird thing shows up in the console not really sure if its an issue

For this, I couldn't reproduce it. I don't think we cache any data in the store so every time we refresh, it should be like a fresh state. 🤔

rjchow (commented):
I think there's some issues with the cache, but otherwise it looks good and almost ready to merge
```
wp.data.dispatch( wc.data.experimentalSettingOptionsStore ).editSettings( 'general', { woocommerce_currency: 'AED', 'woocommerce_default_country': 'GB'  } )
Promise { <state>: "pending" }

wp.data.dispatch( wc.data.experimentalSettingOptionsStore ).saveEditedSettingsGroup( 'general' );
Promise { <state>: "pending" }

wp.data.resolveSelect( wc.data.experimentalSettingOptionsStore ).getSetting( 'general', 'woocommerce_currency' ).then(console.log)
Object { id: "woocommerce_currency", label: "Currency", description: "This controls what currency prices are listed at in the catalog and which currency gateways will take payments in.", type: "select", default: "USD", options: {…}, tip: "This controls what currency prices are listed at in the catalog and which currency gateways will take payments in.", value: "JPY", group_id: "general", _links: {…} }

Promise { <state>: "fulfilled", <value>: undefined }

wp.data.resolveSelect( wc.data.experimentalSettingOptionsStore ).getSettingValue( 'general', 'woocommerce_currency' ).then(console.log)
Promise { <state>: "pending" }

AED
wp.data.resolveSelect( wc.data.experimentalSettingOptionsStore ).getSetting( 'general', 'woocommerce_currency' ).then(console.log)
Object { id: "woocommerce_currency", label: "Currency", description: "This controls what currency prices are listed at in the catalog and which currency gateways will take payments in.", type: "select", default: "USD", options: {…}, tip: "This controls what currency prices are listed at in the catalog and which currency gateways will take payments in.", value: "JPY", group_id: "general", _links: {…} }

Promise { <state>: "fulfilled", <value>: undefined }
```

chihsuan:
> I think there's some issues with the cache, but otherwise it looks good and almost ready to merge
> 
> ```
> wp.data.dispatch( wc.data.experimentalSettingOptionsStore ).editSettings( 'general', { woocommerce_currency: 'AED', 'woocommerce_default_country': 'GB'  } )
> Promise { <state>: "pending" }
> 
> wp.data.dispatch( wc.data.experimentalSettingOptionsStore ).saveEditedSettingsGroup( 'general' );
> Promise { <state>: "pending" }
> 
> wp.data.resolveSelect( wc.data.experimentalSettingOptionsStore ).getSetting( 'general', 'woocommerce_currency' ).then(console.log)
> Object { id: "woocommerce_currency", label: "Currency", description: "This controls what currency prices are listed at in the catalog and which currency gateways will take payments in.", type: "select", default: "USD", options: {…}, tip: "This controls what currency prices are listed at in the catalog and which currency gateways will take payments in.", value: "JPY", group_id: "general", _links: {…} }
> 
> Promise { <state>: "fulfilled", <value>: undefined }
> 
> wp.data.resolveSelect( wc.data.experimentalSettingOptionsStore ).getSettingValue( 'general', 'woocommerce_currency' ).then(console.log)
> Promise { <state>: "pending" }
> 
> AED
> wp.data.resolveSelect( wc.data.experimentalSettingOptionsStore ).getSetting( 'general', 'woocommerce_currency' ).then(console.log)
> Object { id: "woocommerce_currency", label: "Currency", description: "This controls what currency prices are listed at in the catalog and which currency gateways will take payments in.", type: "select", default: "USD", options: {…}, tip: "This controls what currency prices are listed at in the catalog and which currency gateways will take payments in.", value: "JPY", group_id: "general", _links: {…} }
> 
> Promise { <state>: "fulfilled", <value>: undefined }
> ```

Good catch! @rjchow Confirmed that the issue is related to the selector cache not being correctly invalidated. I've fixed the selectors' memoization and added some tests for it in https://github.com/woocommerce/woocommerce/pull/56516/commits/63d8d3851b116272bb6aaf5be43fabb1389e6a77.

Screenshot 2025-03-27 at 14 35 12

rjchow (approved):
Thanks! Magnificent work!

Issue #50009, opened 2024-07-26
[Enhancement]: Add support for ActionScheduler's priority argument + support for setting the webhook priority


ahegyes:
### Describe the solution you'd like

Last year, ActionScheduler added support for a priority argument to their `as_schedule_single_action` and `as_schedule_recurring_action` functions. However, the `WC_Action_Queue` does not yet support those.

Recently encountered the situation that a website recieved 20k+ new users within a span of 2-3 days. On every new user being created, there were 2 webhooks configured which ended up going into the AS queue under the `woocommerce_deliver_webhook_async` hook. 

However, that created a serious backlog of tasks to be processed that caused WC Subscription renewals to not be processed on time. In the grand scheme of things, notifying 3rd-parties of new customers via webhooks is much less important than maintaining the stream of revenue via subscription renewals. 

Ideally, by adding support for this property, there would also be some mechanism to tweak the priorities. Ideas:

* "local" / "core" features like payment processing and email notifications should receive a higher priority by default than things like webhook notifications
* a new setting field for each webhook to configure a priority manually that is different from the default (the store in question has over 100 webhooks so it might be nice to decide a prioritization scheme)

barryhughes:
Thanks, we've discussed this a little internally, but haven't so far been able to prioritize it. However, we definitely recognize both that the current queue implementation is lagging behind what Action Scheduler now offers, and that things are a bit fragmented within the plugin in terms of where and when the queue is even used (vs direct calls to AS).

pmcpinto:
Reviewed and confirmed this is still valid. Marking as ready for engineering.

PR #56525, opened 2025-03-19
Performance: expose action scheduler priority flag

kalessil:

Context: HACK week activities, https://github.com/woocommerce/woocommerce/issues/50009

Here we are taking the very first step and exposing the priority flag for ActionScheduler API in a backward-compatible way with a reference to the exposed flag.

Testing scenario 1 (default queue):

- Using a local testing site, delete `woocommerce_marketplace_suggestions` entry from the options table
- Navigate to `WooCommerce -> Orders` in the admin area (it's one of pages triggering the marketplace data refresh)
- Evaluate entries in action scheduler tables for hook `woocommerce_update_marketplace_suggestions` (priority column should be `1` instead of the default `10`)
- Refresh the page and ensure the option `woocommerce_marketplace_suggestions` is populated

Testing scenario 2 (custom queue):

- Patch `\WC_Action_Queue` and remove `, QueueWithPrioritiesInterface` to simulate existing custom queue
- Perform testing steps from scenario 1 - the only difference should be priority column value for A-S table (`10`)


plugins/woocommerce/includes/queue/class-wc-action-queue.php
```diff
@@ -16,52 +18,57 @@
  * A job queue using WordPress actions.
  *
  * @version 3.5.0
+ *
+ * TODO: backward compatibility
```
kalessil: Context: `woocommerce_queue_class` filter allows the specification of a replacement. We'll need an aggregate that is aware of priorities support by the queue instance and route priority flag accordingly.

Note: new interface, queue implementation, and the aggregate should be under `src`.

plugins/woocommerce/tests/php/src/Proxies/ClassThatDependsOnLegacyCodeTest.php
```diff
@@ -113,7 +114,7 @@ public function test_static_mocks_can_be_used_via_injected_legacy_proxy_and_wooc
 	 */
 	public function test_get_instance_of_can_be_used_via_injected_legacy_proxy_and_woocommerce_object( $method_to_use ) {
 		$instance = $this->sut->$method_to_use( \WC_Queue_Interface::class, 34 );
-		$this->assertInstanceOf( \WC_Action_Queue::class, $instance );
+		$this->assertInstanceOf( QueueProxy::class, $instance );
```
kalessil: TODO: assert interface instance, not the implementation.

plugins/woocommerce/tests/php/src/Proxies/LegacyProxyTest.php
```diff
@@ -86,7 +87,7 @@ public function test_get_instance_of_class_with_load_method_gets_an_instance_of_
 	 */
 	public function test_get_instance_of_wc_queue_interface_gets_an_instance_of_the_appropriate_class() {
 		$instance = $this->sut->get_instance_of( \WC_Queue_Interface::class, 34 );
-		$this->assertInstanceOf( \WC_Action_Queue::class, $instance );
+		$this->assertInstanceOf( QueueProxy::class, $instance );
```
kalessil: TODO: assert interface instance, not the implementation.

plugins/woocommerce/src/Admin/API/OnboardingPlugins.php
```diff
@@ -155,7 +156,12 @@ public function install_and_activate_async( WP_REST_Request $request ) {
 		$plugins = $request->get_param( 'plugins' );
 		$job_id  = uniqid();
 
-		WC()->queue()->add( 'woocommerce_plugins_install_and_activate_async_callback', array( $plugins, $job_id ) );
+		WC()->queue()->add(
+			'woocommerce_plugins_install_and_activate_async_callback',
+			array( $plugins, $job_id ),
+			'',
+			ActionQueuePriority::URGENT
```
kalessil: `ActionQueuePriority::URGENT` - toying around, not sure what this action is supposed to do.

kalessil:
Adding @barryhughes (to keep in the loop) in case subscriptions could leverage prios in the near future.

plugins/woocommerce/includes/queue/class-wc-action-queue.php
```diff
@@ -17,31 +21,58 @@
...
+	/**
+	 * Enqueue an action to run one time, as soon as possible
+	 *
+	 * @param string $hook     The hook to trigger.
+	 * @param array  $args     Arguments to pass when the hook triggers.
+	 * @param string $group    The group to assign this job to.
+	 * @param int    $priority Action priority.
+	 * @return string The action ID
+	 */
+	public function add_with_priority( string $hook, array $args = array(), string $group = '', int $priority = ActionQueuePriority::NORMAL ) {
```
kalessil, line 47: TODO: tests for new methods (in progress).

plugins/woocommerce/includes/admin/marketplace-suggestions/class-wc-marketplace-suggestions.php
```diff
@@ -197,10 +200,14 @@ public static function get_suggestions_api_data() {
...
+			$queue = WC()->queue();
+			$next  = $queue->get_next( 'woocommerce_update_marketplace_suggestions', array(), WC_Marketplace_Updater::JOBS_GROUP );
 			if ( ! $next ) {
-				WC()->queue()->cancel_all( 'woocommerce_update_marketplace_suggestions' );
-				WC()->queue()->schedule_single( time(), 'woocommerce_update_marketplace_suggestions' );
+				$queue->cancel_all( 'woocommerce_update_marketplace_suggestions', array(), WC_Marketplace_Updater::JOBS_GROUP );
+
+				( $queue instanceof QueueWithPrioritiesInterface )
+					? $queue->add_with_priority( 'woocommerce_update_marketplace_suggestions', array(), WC_Marketplace_Updater::JOBS_GROUP, ActionQueuePriority::HIGH )
+					: $queue->add( 'woocommerce_update_marketplace_suggestions', array(), WC_Marketplace_Updater::JOBS_GROUP );
```
barryhughes, line 210: Commenting here not because of a problem with these lines, as such, but just because they are quite a nice illustration of the design problem I'm trying to describe.

So, I think the primary challenge is that there is no guarantee that we are working with a priority-aware queue implementation. That is, the only thing we can be sure of when we call `WC()->queue()` is that we'll get back a type that satisfies the original `WC_Queue_Interface` ... and, of course, you know all this, or we wouldn't have the ternary expression here 🙃 

Looking to the future, what if we want to add support for other capabilities—for instance, uniqueness (which is also something Action Scheduler supports, and which solves a common race condition)? If we follow the same pattern as we have in this PR, we might then update `WC_Action_Queue` to additionally implement `QueueWithUniqueActionsInterface` ... but to safely consume these capabilities, we'd have a pretty messy situation.

This all leads me to ask:

1. Should we introduce support for priorities alongside support for uniqueness (ie, make both changes available in the same release)? That might be a little better, though it doesn't solve the broader problem I've described (because there might be yet other capabilities we want to add support for in the future).
2. Is there a different pattern we could use? What is potentially unlocked if we "ditch" `WC()->queue()` and `WC_Queue_Interface` and introduce a completely new queue interface that would exist in parallel with the current solution?

This could all have been discussed already 🙌🏼 ... and if that analysis has happened and we've decided to move ahead with this anyway, I'm of course fine with that, but I figured it was worth pausing and posing the question.

plugins/woocommerce/includes/queue/class-wc-action-queue.php
```diff
@@ -17,31 +21,58 @@
...
+	 *
+	 * @param int    $timestamp When the job will run.
+	 * @param string $hook      The hook to trigger.
+	 * @param array  $args      Arguments to pass when the hook triggers.
+	 * @param string $group     The group to assign this job to.
+	 * @param int    $priority  Action priority.
+	 * @return string The action ID
+	 */
+	public function schedule_single_with_priority( int $timestamp, string $hook, array $args = array(), string $group = '', int $priority = ActionQueuePriority::NORMAL ) {
+		return as_schedule_single_action( $timestamp, $hook, $args, $group, false, $priority );
```
barryhughes, line 75: In theory, we work on the principle that we have access to whatever version of Action Scheduler has been bundled with the plugin.

So, that means we *should* have access to priorities and, building on my earlier comment, we should also be able to work with the concept of unique actions—and I realize this PR is not concerned with uniqueness, but bear with me here 🙃 

In practice, [this may not be so](https://developer.woocommerce.com/2021/10/12/best-practices-for-deconflicting-different-versions-of-action-scheduler/), and we may find our code running alongside a much older version of Action Scheduler that doesn't support modern features. Now, the chances of this problem occurring are presumably reducing with every passing year but, at the same time, WooCommerce is installed on a huge number of sites and the gold standard is not to destabilize them unnecessarily (just for reference, this problem caused us to issue a [fix release](https://developer.woocommerce.com/2024/04/15/woocommerce-8-8-1-advisory/) only last year).

When it comes to support for priorities, this probably doesn't matter _too_ much, because the worst case is that the priority may be ignored. However, this could still be confusing for merchants or for support. Somewhere—not necessarily here—should we at least check to see if priorities are supported, and log a warning if they are not?

Finally, the reason I mentioned uniqueness was to build on my earlier comment: in a future where we also want to add support for that, it is *far* more important to get this sort of thing right (because if a naive attempt is made to register a unique action in an environment where they aren't supported, we potentially end up with a flood of actions). Again, I don't *think* the stakes are quite so high with priorities but, on the other hand, it's also hard to predict exactly where things will go wrong.

barryhughes (commented):
> In case subscriptions could leverage prios in the near future.

Thank you!

Over on Subscriptions, we use Action Scheduler directly (so far as I'm aware we haven't and don't use `WC()->queue()`) ... though it could definitely be worth reconsidering this in the future.

This is an interesting problem though, and I'm stoked to see some work happening in this area. I left a few comments/questions (mini walls of text, apologies for that!).

plugins/woocommerce/includes/admin/marketplace-suggestions/class-wc-marketplace-suggestions.php
joshuatf: Taking a step back from this code and looking the `WC_Queue` as a whole, what benefit does it serve over directly calling AS functions?

I believe it's a thin wrapper around the interface that served a purpose when we were unsure which version of AS was being loaded.  But since AS is bundled in core and [always uses the latest version](https://developer.woocommerce.com/2021/10/12/best-practices-for-deconflicting-different-versions-of-action-scheduler/) (as of `v3.2.0` ~4 years ago), could we forego a lot of these interfaces and rely directly on AS upstream?  

Otherwise, we'll continue to need to update in both AS and core whenever these types of updates occur and need to break or introduce new interfaces.

plugins/woocommerce/includes/admin/marketplace-suggestions/class-wc-marketplace-suggestions.php
```diff
@@ -197,10 +200,14 @@ public static function get_suggestions_api_data() {
...
-			$next = WC()->queue()->get_next( 'woocommerce_update_marketplace_suggestions' );
+			$queue = WC()->queue();
+			$next  = $queue->get_next( 'woocommerce_update_marketplace_suggestions', array(), WC_Marketplace_Updater::JOBS_GROUP );
 			if ( ! $next ) {
-				WC()->queue()->cancel_all( 'woocommerce_update_marketplace_suggestions' );
-				WC()->queue()->schedule_single( time(), 'woocommerce_update_marketplace_suggestions' );
+				$queue->cancel_all( 'woocommerce_update_marketplace_suggestions', array(), WC_Marketplace_Updater::JOBS_GROUP );
+
+				( $queue instanceof QueueWithPrioritiesInterface )
+					? $queue->add_with_priority( 'woocommerce_update_marketplace_suggestions', array(), WC_Marketplace_Updater::JOBS_GROUP, ActionQueuePriority::HIGH )
```
joshuatf, line 209: Just double checking, but this will potentially have 2 overlapped `woocommerce_update_marketplace_suggestions` actions at the same time since the group is different now, right?

I don't necessarily think that's an issue, but wanted to call it out just in case.

joshuatf (commented):
Thanks for putting this together @kalessil!

Left some comments/questions around the benefits of our `WC_Queue` implementation.  This is something I would like to move away from as it will continue to be a maintenance burden long-term.

plugins/woocommerce/includes/admin/marketplace-suggestions/class-wc-marketplace-suggestions.php
barryhughes: > What benefit does it serve over directly calling AS functions?

Many of these decisions pre-date my time on this project, so, take the following with a healthy pinch of salt 🙌🏼 ... but my assumption is that, originally, it meant we were abstracted from Action Scheduler itself. 

This opens up the potential for A/S to be swapped out with something else entirely (perhaps a solution based on SQS, MQ, etc, which might provide different benefits for merchants operating at certain scales). I think that's smart. 

It also means that, in theory, extensions using the queue abstraction are shielded from changes to the underlying library. In other words, it feels reasonable to take the stance that our use of A/S is a private implementation detail, and third party code shouldn't assume it is available to them *just* because they are building on WooCommerce. If this approach was widely accepted, we could probably imagine various ecosystem benefits (a reduction in various sorts of conflicts and oversights).

However, to an extent, that ship has already sailed. In today's WooCommerce, we have a mix of uses: the queue abstraction being called from some places, and A/S directly in yet others (predominantly, though not exclusively, in code that was merged in from the WC Admin project). So, for an at-scale deployment where there's a need to implement an alternative queuing system, it would really need to be done by registering a new storage backend for A/S itself (which is possible—in fact, it already ships with three different store implementations—but this approach may not always be problem free).

Apologies for the use of internal/private links, but an outline of this situation can also be found at p7bje6-62u-p2 and a further alternative besides the one you noted is outlined at p7bje6-6pd-p2. In short, though, I agree completely that taking a step back from this code and looking at ~the WC_Queue as a whole~ *queuing/handling of async workloads within WC* is exactly what we need to do.

barryhughes: This may read like a set of nits, but, I think it's worth clarifying:

> I believe it's a thin wrapper around the interface that served a purpose when we were unsure which version of AS was being loaded.

I don't think that's *quite* right. We were never "actively" unsure of the version being loaded (even if we should have been).

That is, the promise that the latest version is always the one that will be loaded is not a new one. That was the same promise we had even *before* 3.2.1.

> But since AS is bundled in core and [always uses the latest version](https://developer.woocommerce.com/2021/10/12/best-practices-for-deconflicting-different-versions-of-action-scheduler/) (as of v3.2.0 ~4 years ago)

This is the nit-picky bit. As the post you linked to calls out, this *also* impacted 3.2.0, it was actually addressed (or, *partially* addressed) in the subsequent release, hence the recommendation to adopt at least 3.2.1.

I'd also suggest that, as was the case prior to 3.2.1, just because we don't know of cases where the promise can be broken, doesn't mean they don't exist. Problems in software tend to stick around. As I linked to in my earlier comment, we shipped a fix release only last year because of this.

plugins/woocommerce/includes/queue/class-wc-action-queue.php
kalessil: > just for reference, this problem caused us to issue a [fix release](https://developer.woocommerce.com/2024/04/15/woocommerce-8-8-1-advisory/) only last year

I see and agree that this PR, in its current form, is likely to encounter the same issue.

> When it comes to support for priorities, this probably doesn't matter too much, because the worst case is that the priority may be ignored. However, this could still be confusing for merchants or for support. Somewhere—not necessarily here—should we at least check to see if priorities are supported, and log a warning if they are not?

To me, verifying that the loaded version of Action Scheduler supports priorities and issuing a warning if not sounds reasonable. Checking how we can detect this.

> Finally, the reason I mentioned uniqueness was to build on my earlier comment: in a future where we also want to add support for that, it is far more important to get this sort of thing right (because if a naive attempt is made to register a unique action in an environment where they aren't supported, we potentially end up with a flood of actions). Again, I don't think the stakes are quite so high with priorities but, on the other hand, it's also hard to predict exactly where things will go wrong.

Even if uniqueness is not a focus for this PR, I appreciate that you flagged potential issues - it allows us to understand better pitfalls around how we use Action Scheduler.

plugins/woocommerce/includes/admin/marketplace-suggestions/class-wc-marketplace-suggestions.php
kalessil: > Should we introduce support for priorities alongside support for uniqueness (ie, make both changes available in the same release)? That might be a little better, though it doesn't solve the broader problem I've described (because there might be yet other capabilities we want to add support for in the future).

Valid point, I'll experiment more with the 'capabilities-driven' approach as we align a bit more on the bigger picture.

> Is there a different pattern we could use? What is potentially unlocked if we "ditch" WC()->queue() and WC_Queue_Interface and introduce an entirely new queue interface that would exist in parallel with the current solution?

That would be a backward compatibility break: [here](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/queue/class-wc-queue.php#L63), we are allowing builders to inject a custom queue implementation via the `woocommerce_queue_class` hook.

This hook is basically what holds up the re-warming of the queue implementation (new interfaces, new implementation, which can be capabilities-driven or something similar).

kalessil: > Taking a step back from this code and looking the WC_Queue as a whole, what benefit does it serve over directly calling AS functions?

It's a backward compatibility break as per my previous comment.

plugins/woocommerce/includes/admin/marketplace-suggestions/class-wc-marketplace-suggestions.php
kalessil: Correct. Practically, the action might be executed twice after the update, but later it gets back to normal execution frequency.

plugins/woocommerce/includes/admin/marketplace-suggestions/class-wc-marketplace-suggestions.php
barryhughes: > > What is potentially unlocked if we "ditch" WC()->queue()
> 
> That would be a backward compatibility break: [here](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/queue/class-wc-queue.php#L63)

I'm not sure if it has to break back-compat. 

`WC()->queue()` can continue to supply an instance of `WC_Action_Queue`, and that could continue to satisfy `WC_Queue_Interface`, but it could be updated to delegate the actual work to an instance of hypothetical `WC_Action_Queue_Mark_2` (which might even be internally namespaced). At the same time, direct access to that same new-and-improved queue handler could be obtained directly through some other channel.

Apologies for any confusion that might have been caused by my use of the word "ditching". I definitely did not mean to suggest removing any parts of the existing public API surface. Also not pushing for us to do what I outlined: just noting the possibility.

joshuatf: >It's a backward compatibility break as per my previous comment.

Even if this is a backward compatibility issue, we need to be okay choosing to deprecate and no longer maintain code.  If this is adding extra layers of backwards compatibility to other plugins that have not been updated in 4 years, I'm not sure the onus is on WooCommerce to indefinitely support those plugins.

>That would be a backward compatibility break: [here](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/queue/class-wc-queue.php#L63), we are allowing builders to inject a custom queue implementation via the woocommerce_queue_class hook.

I couldn't find any instances of this hook in the wild on GitHub.  Not to say that there's not any, but again this wrapper becomes a maintenance burden and we have to weigh out the benefits as we have limited time to maintain a high number of APIs, especially when ones already exist that meet our needs (direct use of Action Scheduler).

kalessil:
Thanks for the inputs, everyone!

After several conversations (including here), it looks like the alignment favors using the Action scheduler directly and letting it care about compatibility (after previous compatibility topics addressed, we should be good).

Therefore, I'm abandoning this PR. It goes in another direction, and we'll end up with one more abstraction, which would only complicate the core maintenance.

PR #56527, opened 2025-03-19
[e2e tests] Temporary skip broken e2e tests

adimoldovan:

Skip some e2e tests that are broken. They are already fixed in trunk with https://github.com/woocommerce/woocommerce/pull/56208

E2E tests should pass in CI


kmanijak (approved):
CI is green! 🟢 

Approving and merging as per p1742415303431679/1742397877.855609-slack-C01DT6U03HC

PR #56528, opened 2025-03-19
Store API - Ensure shipping hashes change when choosing shipping methods so orders get updated

### LINKED ISSUE #56523: Order totals are incorrect when shipping method changes in cart
### Prerequisites

### Describe the bug

Assuming you have a Flat Rate method for US that costs $10 and a Flat Rate method for the rest of the world that costs $20 and the store is based in the US. When you add a product to the cart and go to the checkout block, then the $10 Flat Rate is visible. If you change the shipping details to another country, then the $20 Flat Rate is selected and the order totals are correct (including the $20 rate). When you place the order, the order totals that are posted in the order are incorrect as they include the initial $10 Flat Rate. 

https://github.com/user-attachments/assets/98476863-7366-4e8c-a677-c5fe875dba18

### Expected behavior

The order totals should be correct, based on the selected Shipping Method.

### Actual behavior

The order totals are not correct. 

### Steps to reproduce

1. Create a US based store.
2. Create a US Flat Rate for $10.
3. Create a Rest of the World Flat Rate for $20.
4. View the store as logged out user or via incognito window. 
5. Add a product to the cart -- see that the $10 Flat Rate is selected.
6. Fill in a non-US address. 
7. See that the $20 Flat Rate is selected. 
8. Confirm that order totals are correct and place the order. 
9. In the order confirmation screen and in the WC Admin, totals are incorrect. 

### WordPress Environment

Fresh WooCommerce install with WooCommerce nightly enabled. 

### Isolating the problem

mikejolley:

In https://github.com/woocommerce/woocommerce/pull/56367 we fixed some logic so that cart hashes were correctly used to detect if the order needed to be updated or not. The logic in this function was broken previously.

Uncaught in tests, because of a separate bug in the `get_cart_hashes) method, the hash would not change after changing shipping method selection on checkout. This caused the order to contain stale data.

The fix is to include chosen methods as part of the hash.

Closes #56523

(For Bug Fixes) Bug introduced in PR https://github.com/woocommerce/woocommerce/pull/56367

1. On a store, enable a shipping method, and local pickup
2. Add something to cart and go to block checkout
3. Fill out the checkout form, then toggle shipping -> local pickup.
4. Place order
5. Confirm Local Pickup was used and you can see a collection address in the confirmation

I've updated one of the e2e tests to check that the toggle works.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/fix-56523-order-total-update
+++ b/plugins/woocommerce/changelog/fix-56523-order-total-update
@@ -1,0 +1,5 @@
+Significance: patch
+Type: fix
+Comment: Regression in 56367
+
+
--- a/plugins/woocommerce/client/blocks/tests/e2e/tests/checkout/checkout-block.shopper.block_theme.spec.ts
+++ b/plugins/woocommerce/client/blocks/tests/e2e/tests/checkout/checkout-block.shopper.block_theme.spec.ts
@@ -160,7 +160,7 @@ test.describe( 'Shopper → Local pickup', () => {
 		}
 	);
 
-	test( 'Switching between local pickup and shipping does not affect the address', async ( {
+	test( 'Switching between local pickup and shipping does not affect the address and is used for the order', async ( {
 		page,
 		frontendUtils,
 		checkoutPageObject,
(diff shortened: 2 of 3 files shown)


jimjasson (approved):
Thanks for working on this @mikejolley! 

I ran a test with your fix and confirmed that now, when I change the selected Shipping Method in the checkout block, then my selection persists after the order is submitted. Before this fix, always the Shipping Method that was initially selected was passed to the order. 

Good call on improving the E2E test as well, to avoid similar regressions in the future.💯

wavvves (approved):
Nice one Mike, approving 👍🏼 

I noticed that md5(wp_json_encode()) is a common pattern to hash things around, I'm opening a GH discussion because there are more performant alternatives around this

PR #56543, opened 2025-03-20
Improvements and fixes when recounting product category and tag terms

### LINKED ISSUE #44462: Term Counts miscalculated when Product Settings are saved
### Prerequisites

### Describe the bug

Our team has been investigating a rather subtle bug causing an issue with certain taxonomy term counts; in this case causing our site menus to disappear whenever WooCommerce Product 'Inventory' settings are saved (but, I think this might cause a different effect depending on each store, depending on whether there are any term IDs matching term taxonomy IDs).

This appears to not have been reported, and - if proven to be correct, might have been causing subtle issues for a while.
We’ve had multiple developers look into this and have confirmed the same thing - I think there is a low level bug here.

image (65)
image (64)
image (63)

1. wc_recount_all_terms is called whenever WooCommerce product settings are updated, which triggers a recount of ‘product category’ terms.
2. This retrieves all ‘product_cat’ taxonomies (valid) and passes these to _wc_term_recount, with arguments true for callback, and false for terms_are_term_taxonomy_ids (no issue here), screenshot attached.
3. The second screenshot shows the _wc_term_recount function. Since the callback argument is true, _update_post_term_count is called. NOTE - this is where the problem is. _update_post_term_count expects a list of ‘Term Taxonomy IDs’, not Term IDs. You can see in this function, that WordPress updates based on the ‘term_taxonomy_id’ field, which in our case is causing a ‘nav_menu’ taxonomy count to be calculated incorrectly, as it has the same ‘term taxonomy id’ as a ‘term id’ of a completely unrelated term.
4. In the case of WooCommerce, it should just be updating ‘product_cat’ term taxonomies, but since term IDs are passed in to the above function, other terms are recalulcated, and since $taxonomy is set to ‘product_cat’, it’ll recalculate terms incorrectly, in our case setting some of our nav_menu counts to 0, which results in menus not rendering.

### Expected behavior

Only product category terms are recalculated.

### Actual behavior

Terms with a matching term_id are updated incorrectly.

### Steps to reproduce

1. Update a nav_menu or non-product category term to have a term_taxonomy_id (ie term_taxonomy_id = 1950) that matches a product category 'term id' (term_id = 1950, term_taxonomy_id = 123)
2. Save WooCommerce Product settings, triggering a recount of product categories
3. Note that the 'count' field within wp_term_taxonomy for term_taxonomy_id = 1950 has been reset, even though it is not a product category term.

### WordPress Environment ###
WC Version: 8.4.0
REST API Version: ✔ 8.4.0
WC Blocks Version: ✔ 11.6.2
Action Scheduler Version: ✔ 3.7.0
Log Directory Writable: ✔
WP Version: ❌ 6.4.2 - There is a newer version of WordPress available (6.4.3)
WP Multisite: –
WP Memory Limit: 7 GB
WP Debug Mode: –
WP Cron: –
Language: en_US
External object cache: –

### Database ###

WC Database Version: 8.4.0
WC Database Prefix: wp_

exclude-from-catalog (exclude-from-catalog-2)
featured (featured)
featured (featured-2)
outofstock (outofstock)
outofstock (outofstock-3)
outofstock (outofstock-2)
rated-1 (rated-1)
rated-1 (rated-1-2)
rated-2 (rated-2)
rated-3 (rated-3-3)
rated-3 (rated-3-4)
rated-3 (rated-3)
rated-3 (rated-3-2)
rated-4 (rated-4)
rated-4 (rated-4-2)
rated-5 (rated-5-3)
rated-5 (rated-5)
rated-5 (rated-5-2)

Connected to Woo.com: –
Enforce Approved Product Download Directories: –
HPOS feature screen enabled: ✔
HPOS feature enabled: ✔
HPOS data sync enabled: –

Generated at: 2024-02-09 01:32:49 +10:00
`

### Isolating the problem

- [X] This bug happens with a default WordPress theme active, or [Storefront](https://woo.com/products/storefront/).

### LINKED ISSUE #54680: Spike: Likely Bug in wc_recount_all_terms()
## Issue

The function `_update_post_term_count()` is called with a term_id of `0` once for each top level product category.  The same happens for product tags.  See the image below of a local test site with 64 product categories and 199 tags:

Image

## Steps to Reproduce
* Load a site with multiple product categories.
* Setup Query monitor so you can view queries run.
* Go to WP-Admin -> WooCommerce -> Settings -> Products -> Inventory
* Check/Uncheck any option just so you can save the page settings.
* Review the Database Queries

## Expected Cause

The function `_wc_term_recount()` has been around for awhile (renamed from `_woocommerce_term_recount()`) and has some different parameters making it behave slightly different and accept different format `$terms` parameters.

Most of the time when `_wc_term_recount()` is used, the `$terms` parameter is an array in the formation of `$term_id => $parent_id`.  The call to `_update_post_term_count()` receives this array as is and only uses the values of the array which is just `$parent_id`s.  At this point, I am unsure if that is the intentional.

If that is correct, and we do want to run `_update_post_term_count()` against all of the `parent_id`s, then we should first run `array_unique` against the set and filter out the `0` ID to prevent unneeded updates.

Otherwise, we should determine if we should be calling `_update_post_term_count()` against the individual `term_id`s instead and pass the parameter through `array_keys()`.

It would also make sense to move this call until after the handling to deal with cases where `$terms_are_term_taxonomy_ids == true`.

Relevant Links:
* 37de4f62cbdc6b9a96b79ad875e06273a72c1248. Possible fix for https://github.com/woocommerce/woocommerce/issues/1882 That moved callback usage
* cb0b92750ba4a292ec8afa530c7033e9179d1a37 Extraction of `_wc_term_recount()` calls into `wc_recount_all_terms()` where `$callback = true`

coreymckrill:

This makes a few changes to the system that calculates how many products are assigned to each Product Category and Product Tag taxonomy term. The main changes are:

* Fixing a subtle bug that made it so that the counts could be incorrect for some terms retrieved with the `get_terms` function when multiple taxonomies were specified.
* Fixing a bug where the built-in WP callback for counting terms, `_update_post_term_count`, was being given the wrong IDs (term IDs instead of term taxonomy IDs in some cases, parent term IDs in other cases), which caused queries to be run, and counts updated for _the wrong terms_.
* Reworking the flow of the code to avoid unnecessary queries running during certain times when terms are being recounted.

Fixes #44462
Fixes #54680

Some parts of this changeset are difficult to test manually. In those cases I've added unit tests to try and ensure that no regressions are caused.

* ~Follow the "Steps to reproduce" in #44462 and make sure only the correct count field gets reset.~ Smoke test to ensure that updating posts and categories, products and product categories/tags results in accurate counts for those terms.
* Follow the "Steps to reproduce" in #54680 and note that Query Monitor no longer shows any duplicate queries relating to term counts after saving the settings in WooCommerce -> Settings -> Products -> Inventory.
* Create a few product categories. Have at least one be a child of another category. Create a few products. Assign various categories to these products. Make sure you're working with a small enough number that you can easily figure out what the product count should be for each category. Visit the front end of your shop and spot check the category archive pages (e.g. `http://localhost/product-category/category-1`) for a few categories and make sure the "Showing all X results" header is correct for these categories.
* Now mark one of your products as "Out of stock" and then go to WooCommerce -> Settings -> Products -> Inventory and check the "Hide out of stock items from the catalog" box and save your settings. Go back and visit the category archive page for one of the categories assigned to the out of stock product. It should be changed to not take that product into account anymore.

* During development, these changes were tested using wp-env with WP 6.7.2, in both PHP 7.4 and PHP 8.1.
* This PR adds several unit tests that verify the functionality in the relevant changes.

### DIFF
--- a/plugins/woocommerce/changelog/fix-54680-recount-terms
+++ b/plugins/woocommerce/changelog/fix-54680-recount-terms
@@ -1,0 +1,4 @@
+Significance: minor
+Type: fix
+
+Ensure term counts are calculated correctly when the term count method is not given a list of term taxonomy IDs. Also improve performance by avoiding unnecessary queries while counting terms.
--- a/plugins/woocommerce/includes/admin/settings/class-wc-settings-products.php
+++ b/plugins/woocommerce/includes/admin/settings/class-wc-settings-products.php
@@ -490,7 +490,7 @@ class WC_Settings_Products extends WC_Settings_Page {
 		 * Product->Inventory has a setting `Out of stock visibility`.
 		 * Because of this, we need to recount the terms to keep them in-sync.
 		 */
-		WC()->call_function( 'wc_recount_all_terms' );
+		WC()->call_function( 'wc_recount_all_terms', false );
 
 		$this->do_update_options_action();
 	}
--- a/plugins/woocommerce/includes/wc-term-functions.php
+++ b/plugins/woocommerce/includes/wc-term-functions.php
@@ -658,7 +658,8 @@ function wc_get_product_visibility_term_ids() {
 
 	static $term_ids = array();
 
-	if ( count( $term_ids ) > 0 ) {
+	// The static variable doesn't work well with unit tests.
+	if ( count( $term_ids ) > 0 && ! class_exists( 'WC_Unit_Tests_Bootstrap' ) ) {
 		return $term_ids;
 	}
 
--- a/plugins/woocommerce/includes/wc-term-functions.php
+++ b/plugins/woocommerce/includes/wc-term-functions.php
@@ -526,7 +526,17 @@ function wc_change_term_counts( $terms, $taxonomies ) {
 		return $terms;
 	}
 
-	if ( ! isset( $taxonomies[0] ) || ! in_array( $taxonomies[0], apply_filters( 'woocommerce_change_term_counts', array( 'product_cat', 'product_tag' ) ), true ) ) {
+	/**
+	 * Filter which product taxonomies should have their term counts overridden to take catalog visibility into account.
+	 *
+	 * @since 2.1.0
+	 *
+	 * @param array $valid_taxonomies List of taxonomy slugs.
+	 */
+	$valid_taxonomies   = apply_filters( 'woocommerce_change_term_counts', array( 'product_cat', 'product_tag' ) );
+	$current_taxonomies = array_intersect( (array) $taxonomies, $valid_taxonomies );
+
+	if ( empty( $current_taxonomies ) ) {
 		return $terms;
 	}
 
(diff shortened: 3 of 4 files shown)


albarin:
Looking good so far, I've tested everything except the first point:

> Follow the "Steps to reproduce" in https://github.com/woocommerce/woocommerce/issues/44462 and make sure only the correct count field gets reset.

I'm having trouble understanding this step 😅 could you maybe clarify? 🙏 

> Update a nav_menu or non-product category term to have a term_taxonomy_id (ie term_taxonomy_id = 1950) that matches a product category 'term id' (term_id = 1950, term_taxonomy_id = 123)

Thanks!

coreymckrill:
@albarin 🤔 That one is a bit tricky. I confirmed that what the reporter of the issue said is true, that we were giving the `_update_post_term_count` function `term_id`'s instead of `taxonomy_term_id`'s. But I am also not able to replicate the issue as the instructions are written. Perhaps instead we should just do some smoke testing to make sure taxonomy term counts seem accurate?

albarin:
> Perhaps instead we should just do some smoke testing to make sure taxonomy term counts seem accurate?

Thanks @coreymckrill, that's what I did 👌 Every other point of the testing instructions worked well 😄 
Approving! ✅

albarin (approved):

PR #56562, opened 2025-03-20
Always load preview data in the editor for Checkout

senadir:
This PR fixes an issue in which real cart ended up being loaded in the editor, something causing mixed results, or showing the address in the editor, which is not what should happen.

This also fixes a side issue in which a request is made to PUT checkout in the editor because no PHP default payment is set, this would cause a draft order to be created.

Around the codebase, we had places where we would patch in `isEditor` and return preview data instead of actual data, but in the data store, we would always fetch data.

This however, caused anyone who reaches to `getCartData` to trigger a request to Store API, and plenty of components did (ones that couldn't subscribe to `useStoreCart`). So in this PR, I fixed it at the root of it, which will be future proof.

For the extra PUT request, an easy solution would have been to bail early in `pushChanges` but a better solution would be to correct the default state there.

Closes WOOPLUG-3236

1. Go to Checkout, fill out your address, place order, all should be good. This just ensures you have an address in your account. Skip if you placed orders before logged in.
2. Once an order is placed, add an item to your cart
3. Go to Checkout editor page, ensure all fields are empty.
4. Open network tab, ensure there is no request being made to cart or checkout endpoints.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Fix a bug in which real address was being shown on the editor page for Checkout.
</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/local-pickup-select/index.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/local-pickup-select/index.tsx
@@ -6,8 +6,7 @@ import {
 	RadioControlOptionType,
 } from '@woocommerce/blocks-components';
 import { CartShippingPackageShippingRate } from '@woocommerce/types';
-import { cartStore } from '@woocommerce/block-data';
-import { useSelect } from '@wordpress/data';
+import { useShippingData } from '@woocommerce/base-context';
 
 interface LocalPickupSelectProps {
 	title?: string | undefined;
@@ -33,9 +32,8 @@ export const LocalPickupSelect = ( {
 	renderPickupLocation,
 	packageCount,
 }: LocalPickupSelectProps ) => {
-	const internalPackageCount = useSelect(
-		( select ) => select( cartStore )?.getCartData()?.shippingRates?.length
-	);
+	const { shippingRates } = useShippingData();
+	const internalPackageCount = shippingRates?.length || 1;
 	// Hacky way to check if there are multiple packages, this way is borrowed from  `assets/js/base/components/cart-checkout/shipping-rates-control-package/index.tsx`
 	// We have no built-in way of checking if other extensions have added packages.
 	const multiplePackages =
--- a/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/shipping-rates-control-package/index.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/shipping-rates-control-package/index.tsx
@@ -29,11 +29,10 @@ export const ShippingRatesControlPackage = ( {
 	showItems,
 	highlightChecked = false,
 }: PackageProps ): ReactElement => {
-	const { selectShippingRate, isSelectingRate } = useShippingData();
+	const { selectShippingRate, isSelectingRate, shippingRates } =
+		useShippingData();
 
-	const internalPackageCount = useSelect(
-		( select ) => select( cartStore )?.getCartData()?.shippingRates?.length
-	);
+	const internalPackageCount = shippingRates?.length || 1;
 
 	const packageClass = 'wc-block-components-shipping-rates-control__package';
 	const [ instanceCount, setInstanceCount ] = useState( 0 );
--- a/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/shipping-rates-control-package/index.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/shipping-rates-control-package/index.tsx
@@ -8,8 +8,6 @@ import { useCallback, useEffect, useMemo, useState } from '@wordpress/element';
 import { useShippingData } from '@woocommerce/base-context/hooks';
 import { sanitizeHTML } from '@woocommerce/utils';
 import type { ReactElement } from 'react';
-import { useSelect } from '@wordpress/data';
-import { cartStore } from '@woocommerce/block-data';
 
 /**
  * Internal dependencies
--- a/plugins/woocommerce/client/blocks/assets/js/base/context/hooks/cart/use-store-cart.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/base/context/hooks/cart/use-store-cart.ts
@@ -23,7 +23,6 @@ import { useSelect } from '@wordpress/data';
 import { decodeEntities } from '@wordpress/html-entities';
 import type {
 	StoreCart,
-	CartResponse,
 	CartResponseTotals,
 	CartResponseFeeItem,
 	CartResponseBillingAddress,
--- a/plugins/woocommerce/client/blocks/assets/js/base/context/hooks/cart/use-store-cart.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/base/context/hooks/cart/use-store-cart.ts
@@ -36,7 +36,6 @@ import { emptyHiddenAddressFields } from '@woocommerce/base-utils';
 /**
  * Internal dependencies
  */
-import { useEditorContext } from '../../providers/editor-context';
 import { useStoreCartEventListeners } from './use-store-cart-event-listeners';
 
 declare module '@wordpress/html-entities' {
@@ -140,11 +139,6 @@ export const useStoreCart = (
 	options: { shouldSelect: boolean } = { shouldSelect: true }
 ): StoreCart => {
 	const { shouldSelect } = options;
-	const { isEditor, previewData } = useEditorContext();
-	const previewCart = previewData?.previewCart as unknown as CartResponse & {
-		receiveCart?: ( cart: CartResponse ) => void;
-		receiveCartContents?: ( cart: CartResponse ) => void;
-	};
 	const currentResults = useRef();
 	const billingAddressRef = useRef( defaultBillingAddress );
 	const shippingAddressRef = useRef( defaultShippingAddress );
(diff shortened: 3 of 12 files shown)


plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/local-pickup-select/index.tsx
```diff
@@ -33,9 +32,8 @@ export const LocalPickupSelect = ( {
 	renderPickupLocation,
 	packageCount,
 }: LocalPickupSelectProps ) => {
-	const internalPackageCount = useSelect(
-		( select ) => select( cartStore )?.getCartData()?.shippingRates?.length
```
senadir, line 37: No need to be doing this.

plugins/woocommerce/client/blocks/assets/js/base/context/hooks/cart/use-store-cart.ts
```diff
@@ -158,35 +152,6 @@ export const useStoreCart = (
...
-							? previewCart.receiveCart
-							: () => undefined,
-					receiveCartContents:
-						typeof previewCart?.receiveCartContents === 'function'
-							? previewCart.receiveCartContents
-							: () => undefined,
-				};
-			}
-
 			const store = select( cartStore );
```
senadir, line 154: No longer needed

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-shipping-address-block/block.tsx
```diff
@@ -53,7 +53,16 @@ const Block = (): JSX.Element => {
...
+		// Only sync if any values are different
+		const needsSync = ! Object.keys( syncValues ).every(
+			( key ) =>
+				syncValues[ key as keyof BillingAddress ] ===
+				billingAddress[ key as keyof BillingAddress ]
+		);
+
+		if ( needsSync ) {
+			setBillingAddress( syncValues );
+		}
```
senadir: This was causing an extra empty request to sync.

mikejolley:
@senadir I will review this tomorrow. There are merge conflicts in the meantime.

plugins/woocommerce/client/blocks/assets/js/base/context/hooks/cart/use-store-cart.ts
ralucaStan: Hi Nadir, I'm looking at your PR and I need some context about why this isn't needed. You mentioned in the PR description that the PR is fixing the root cause, but you're not describing your approach and why it is future proof.

I'd really appreciate an updated description with a few words about the approach in your fix.

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-shipping-address-block/block.tsx
```diff
@@ -53,7 +53,16 @@ const Block = (): JSX.Element => {
 			delete syncValues.company;
 		}
 
-		setBillingAddress( syncValues );
+		// Only sync if any values are different
+		const needsSync = ! Object.keys( syncValues ).every(
+			( key ) =>
+				syncValues[ key as keyof BillingAddress ] ===
```
ralucaStan: Some things come to mind when checking for billing address object equality:
- do we have arrays or objects for the value, because deep equality won't work; I'm thinking country filed, maybe additional fields?
- you can check for keys length and return true if they are not equal, a small optimization
-

plugins/woocommerce/client/blocks/assets/js/data/utils/is-editor.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * External dependencies
+ */
+import { getPath } from '@wordpress/url';
+
+/**
+ * Returns true if the current page is in the editor.
+ */
+export const isEditor = (): boolean => {
```
ralucaStan, line 9: I'm curious if there isn't already another way used to detect this. I saw this approach https://github.com/woocommerce/woocommerce/blob/7a636875210947e899b23675402097229245b285/plugins/woocommerce/client/blocks/assets/js/data/payment/utils/check-payment-methods.ts#L46 but I think we should check for `core/edit-post` as well. WDYT?

ralucaStan (commented):
I've tested the PR and works as described. Checked the cart block and with or without the local pickup enabled. 

I left some comments. I understood the approach in the end, but it would be good to add a couple of words about it in the PR description .Thank you Nadir.

plugins/woocommerce/client/blocks/assets/js/data/utils/is-editor.ts
senadir: That was my first option but ended up being too early to call it in a data store, hence why I switched to this static link approach which I'm not 100% confident about. If a frontend checkout page has `post.php` in it, which is something I never saw, then it would break.

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-shipping-address-block/block.tsx
senadir: addresses are flat objects so a quick surface level iteration on them is fine. additional fields are top level values here.

plugins/woocommerce/client/blocks/assets/js/base/context/hooks/cart/use-store-cart.ts
senadir: I think it's explained in this section:

> Around the codebase, we had places where we would patch in isEditor and return preview data instead of actual data, but in the data store, we would always fetch data.

The above part is not needed because it's redundant, the same hook will call `getCartData` a few lines after that, which does check for isEditor and return previewCart, so no need to do that at the hook level.

Fixing it at the root cause means going to where the actual `fetch` request is happening instead of trying to avoid doing it down the road, and simply return preview data at that level, this means anyone calling `getCartData` would resolve to the previewData always.

senadir:
Failing checks seem to be from trunk

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/shipping-rates-control-package/index.tsx
```diff
@@ -29,11 +27,10 @@ export const ShippingRatesControlPackage = ( {
...
 	highlightChecked = false,
 }: PackageProps ): ReactElement => {
-	const { selectShippingRate, isSelectingRate } = useShippingData();
+	const { selectShippingRate, isSelectingRate, shippingRates } =
+		useShippingData();
 
-	const internalPackageCount = useSelect(
-		( select ) => select( cartStore )?.getCartData()?.shippingRates?.length
-	);
+	const internalPackageCount = shippingRates?.length || 1;
```
mikejolley, line 33: I'm concerned about this. `internalPackageCount` makes me suspicious that `getCartData()?.shippingRates` are packages, whereas `useShippingData(). shippingRates` are the rates. Can you confirm?

We may need to test this against a plugin with packages, such as subscriptions.

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-shipping-address-block/block.tsx
mikejolley: Would there be value in abstracting this into a helper somewhere?

plugins/woocommerce/client/blocks/assets/js/data/cart/resolvers.ts
```diff
@@ -3,20 +3,27 @@
...
+import { isEditor } from '../utils';
 
 /**
  * Resolver for retrieving all cart data.
  */
 export const getCartData =
 	() =>
 	async ( { dispatch }: { dispatch: CartDispatchFromMap } ) => {
+		if ( isEditor() ) {
+			dispatch.receiveCart( previewCart );
```
mikejolley, line 23: 👍🏻

plugins/woocommerce/client/blocks/assets/js/data/utils/is-editor.ts
mikejolley: Not 100% about this one.

Could we pass something from server instead?

mikejolley (changes requested):
When testing I noticed if you switch from shipping to pickup we do 2 API requests to `select-shipping-rate`. Should this be avoided in the editor as well?

As for the changes and testing steps, this worked as intended. I checked a logged in account with previous orders and current active session on the frontend.

But I think there may be a logic error I am requesting changes until you've able to confirm suspicions.

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/shipping-rates-control-package/index.tsx
senadir: I was as concerned but they are the same thing, `useShippingData().shippingRates` just returns `cartState.shippingRates`. Second, I was concerted this is avoiding subscribing to it, but it uses useSelect, so it's the same.

senadir:
> When testing I noticed if you switch from shipping to pickup we do 2 API requests to select-shipping-rate. Should this be avoided in the editor as well?

Yep I missed that local pickup is selectable, I will cancel this somewhere early, maybe at select level.

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-shipping-address-block/block.tsx
senadir: It's only used there so no need?

plugins/woocommerce/client/blocks/assets/js/data/utils/is-editor.ts
senadir: We can pass something from the server however the server logic is still basically the same, this is how we check for siteEditor in server
```
'site-editor.php' === $pagenow
```

The logic I added matches was Gutenberg uses https://github.com/WordPress/gutenberg/blob/bb9272766d1ef923f9cbb12013ffdd2788a08cb7/packages/edit-post/src/deprecated.js#L24-L25

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-shipping-address-block/block.tsx
ralucaStan: thank you Nadir for confirming those they are flat objects. Did you consider optimizing based on the key length check?
```
  if (Object.keys(syncValues).length !== Object.keys(billingAddress).length) {
            return true;
        }
       ```

senadir: I didn't think that's needed mainly because they both have the same number of fields always, we I added that just in case.

senadir:
ready for you @mikejolley

plugins/woocommerce/client/blocks/assets/js/data/utils/is-editor.ts
mikejolley: I was thinking rather than targeting the editor we could target WP Admin which is easier to detect. That shouldn't impact the frontend store.

mikejolley (approved):
Thanks @senadir I did another frontend test and checked the Local Pickup toggle which if fixed now. If you satisfied with feedback, I think this is good to go if CI is 🟢

PR #56567, opened 2025-03-20
[Email Editor] Generate email editor post for WooCommerce transactional emails

### LINKED ISSUE #56109: Generate transactional emails with blocks
This task is related to redesigning the email settings page: https://github.com/woocommerce/woocommerce/issues/56106

Image

We need to generate the emails in the database so that they can be displayed. Create all the transactional emails that we need and ensure that they are properly registered. 

Make sure they are sent using the right actions.

Once this is done, remove the top-level menu item for Woo Emails.

Add a button to the settings page using a button. If the feature switch is enabled and the emails are not in the database. In the future, we will use the button if we update the emails to re-generate them. 

Make the emails enabled by default and make the name and description the same as the current emails:

Image

1. Create block-based initial content for each core transactional email. When generating the emails, use this pre-made content as a source for content when generating email posts.
2. The minimal content would contain only the WooCommerce placeholder block. This could also work as the default content for third-party emails.
3. For core emails, we can add a [condition to the current template so that it skips the text part](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/templates/emails/customer-processing-order.php#L31-L49). With the condition the part we capture would contain only the order data. We can grab the text part and place it as blocks into our block-based initial content for the particular emails.
4. Transactional emails would either have a unique initial content, or we would use the default if the special one is not defined.
5. Remove the Select control from the email editor sidebar and ensure we can easily fetch Woo transactional emails with related email editor posts.

Here is an example of how such a content initial content could look like:

```
<!-- wp:heading -->
<h2 class="wp-block-heading">Hi <!--[woocommerce/shopper-name]--></h2>
<!-- /wp:heading -->

<!-- wp:paragraph -->
<p>We’ve received your order and will let you know when it’s on its way to you!</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p>Here’s a reminder of what you’ve ordered:</p>
<!-- /wp:paragraph -->

<!-- wp:woo/email-content {"lock":{"move":false,"remove":true}} -->
<div class="wp-block-woo-email-content">##WOO_CONTENT##</div>
<!-- /wp:woo/email-content -->

<!-- wp:paragraph -->
<p>{$additional_content_defined_by_user_in_the_current_settings}</p>
<!-- /wp:paragraph -->
```

Place that in PHP, and for texts, we use translation functions __

The idea is about the initial content we place in the woo_email posts. The minimal content is the Woo Placeholder Block, but we adjust the current PHP template files to skip some texts and create a set of the initial contents, one per the core transactional emails. These initial contents would already contain text blocks plus the Woo block. Users could modify texts above and below the order table, which is currently not editable and part of the block. See [this picture](https://github.com/user-attachments/assets/9f53d56f-f37b-421c-83dc-087e3ecf9b3f). We would move the texts from the red boxes out of Woo Block to the email content.

triple0t:

Closes #56109 

### Email Editor Block Templates Implementation

This PR introduces block-based email templates for WooCommerce transactional emails, improving the customization experience and making email templates more maintainable.

#### Changes

- Added initial block-based email templates for all core WooCommerce transactional emails
- Implemented personalization tags for dynamic content replacement
- Refactored email template generation to use transients for version tracking
- Enhanced email content handling and editor integration
- Improved structure and dynamic content handling in email templates
- Added support for block-based email customization
- Removed the top-level `WP Admin > Woo Emails` menu
- Removed the template selector from the Email editor settings sidebar panel
- Update Woo Content block placeholder preview

#### Email Templates Added

- Admin emails:
  - New order
  - Failed order
- Customer emails:
  - New account
  - Reset password
  - Order confirmation
  - Processing order
  - Completed order
  - Failed order
  - On-hold order
  - Refunded order
  - Customer note
  - Invoice

#### Technical Details

- All new templates use the block editor format with proper WordPress block syntax
- For this PR, we generate the initial email posts when `/wp-admin/admin.php?page=wc-settings&tab=email` page is visited
- Initial block email content templates are stored in `plugins/woocommerce/templates/emails/block/`
- We use `plugins/woocommerce/templates/emails/block/general-block-email.php` to render the dynamic part of the email template
- Added support for personalization tags using the `<!--[woocommerce/tag-name]-->` syntax
- Templates maintain backward compatibility with existing email template structure
- Email posts are associated with WooCommerce transactional emails with the option: `woocommerce_email_templates_WC-EMAIL-ID_post_id`
- We ensure the templates are generated with transient-based version tracking. option: `wc_email_editor_initial_templates_generated`
- Version 9.9.0 targeted for these changes

**Email editor**
Screenshot 2025-03-20 at 7 45 29 PM
Screenshot 2025-03-20 at 7 45 39 PM

**Email client**
Screenshot 2025-03-20 at 7 49 41 PM

Screenshot 2025-03-20 at 9 30 14 PM

NOTE: Testing this PR may be a little difficult without the `Woo emails` UI. You can temporarily add it back for your test.
Add
```php
  'public'       => false,
  'show_ui'      => true, 
  'show_in_menu' => true,
  'show_in_rest' => true,
```
to https://github.com/woocommerce/woocommerce/blob/b647643298086579dc6a60c9e1a60065c945a5aa/plugins/woocommerce/src/Internal/EmailEditor/Integration.php#L98-L109

1. Enable the `block_email_editor` experimental flag (in DB, set `woocommerce_feature_block_email_editor_enabled` option to `yes` Or using wp-cli `wp option add woocommerce_feature_block_email_editor_enabled yes`).
2. Visit `/wp-admin/admin.php?page=wc-settings&tab=email` for automatic email post-generation
3. Create a new order to test admin new order email
4. Process an order through different statuses to test customer emails
5. Add customer notes to verify note emails
6. Test account-related emails (new account, reset password)
7. Verify personalization tags are properly replaced in all emails
8. Check email template customization in the block editor

To retest template generation
* Delete transient `wp transient delete wc_email_editor_initial_templates_generated`
* Delete all or some email posts association with `SELECT * FROM `wp_options` WHERE `option_name` LIKE '%woocommerce_email_templates_%'` 
* Delete the required email editor posts
* Visit `/wp-admin/admin.php?page=wc-settings&tab=email` for automatic email post-generation

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


costasovo:
@triple0t Thanks for the PR. Awesome job. I've performed testing and had a few questions. As a next step I'll look into the code changes.

### Testing performed

1) I had no email posts.

Screenshot 2025-03-21 at 12 57 03

2) After visiting WooCommerce Email Settings, 12 emails were generated.
Screenshot 2025-03-21 at 13 50 21

❓: I noticed that the "Order details" email is `draft` all the other emails have state `publish`. Is that on purpose?  The "Order details" email is triggered manually from the order's admin screen. It can't be enabled or disabled. I would expect it being published as well.

❓  I've experienced some issues with the generating because initially, I had some leftovers in the options table, and I had `woocommerce_email_templates_customer_processing_order_post_id`, which pointed to a nonexistent email. The email was skipped and was not generated. Perhaps we could make the generating more robust and check whether the records point to existing `woo_email` posts. What do you think?

After generating, I created a new account and order and played with order statuses and customer notes, etc., in the admin.
Here is the recording:

https://github.com/user-attachments/assets/086a6f5f-3a46-4eb3-a245-bea180ed4048

I worked as expected. There were two cases where I didn't get the email (both around 1:45 in the video):
1) When I switched from `processing` to `on hold`. When switching to `on hold` from a different state I got the email.
2) When I switched from `completed` to `cancelled`. When switching to `cancelled` from a different state, I got the email.

I assume that's expected behavior, as it might not make sense to send an email for some transitions.

I also deleted transients, posts and option rows with a post-to-email association a couple of times and re-generating worked as expected.

❓ I was thinking that in the future, we may put the generating process into a background job, so that it is not dependent on visiting the emails page. What do you think?

plugins/woocommerce/src/Internal/EmailEditor/WCTransactionalEmails/WCEmailTemplateGenerator.php
```diff
@@ -0,0 +1,150 @@
...
+	 * @var WCEmailTemplateManager
+	 */
+	private $template_manager;
+	private $default_templates = [];
+
+	public function __construct() {
+		$this->template_manager = WCEmailTemplateManager::get_instance();
+	}
+
+	public function init() {
```
costasovo: I would suggest using a different method name. AFAIK the init will be triggered in case we grab the service from DI, so I find it potentially dangerous.

plugins/woocommerce/src/Internal/EmailEditor/WCTransactionalEmails/WCEmailTemplateManager.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\Internal\EmailEditor\WCTransactionalEmails;
+
+use Automattic\WooCommerce\Internal\EmailEditor\Integration;
+
+class WCEmailTemplateManager {
```
costasovo: I would suggest a different name. It is a sort of template for the content, but I think this might be easily confused with the main template, which carries the header and footer. What about `WCTransactionalEmailPostsManager` or something similar?

plugins/woocommerce/src/Internal/EmailEditor/WCTransactionalEmails/WCEmailTemplateGenerator.php
```diff
@@ -0,0 +1,150 @@
...
+
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\Internal\EmailEditor\WCTransactionalEmails;
+
+use Automattic\WooCommerce\Internal\EmailEditor\Integration;
+use Automattic\WooCommerce\Internal\EmailEditor\BlockEmailRenderer;
+use Automattic\WooCommerce\Internal\EmailEditor\EmailTemplates\WooEmailTemplate;
+
+class WCEmailTemplateGenerator {
```
costasovo: What about `WCTransactionalEmailPostsGenerator` to avoid confusion with the wp_template posts?

plugins/woocommerce/includes/emails/class-wc-email-cancelled-order.php
```diff
@@ -142,6 +142,23 @@ public function get_content_plain() {
 			);
 		}
 
+		/**
+		 * Get block editor email template content.
+		 *
+		 * @return string
+		 */
+		public function get_block_editor_email_template_content() {
```
costasovo, line 150: I'm not sure this is a good idea. This will force changes in emails added by third parties who respect the WC_Email class ([example](https://github.com/woocommerce/woocommerce-subscriptions/blob/c7da0607a98681f555925126d2445e80b18e32aa/includes/payment-retry/emails/class-wcs-email-payment-retry.php#L17)) to add this method for the support.

What do we want to achieve here? The commit message doesn't give much insight here. Can't we do it in the patent WC_Email class?

plugins/woocommerce/src/Internal/EmailEditor/WCTransactionalEmails/WCEmailTemplateGenerator.php
```diff
@@ -14,60 +14,145 @@ class WCEmailTemplateGenerator {
...
-			],
-			// other email types...
-		];
+		$core_transactional_emails = WCTransactionalEmails::$core_transactional_emails;
+
+		$wc_emails = \WC_Emails::instance();
+		$email_types = $wc_emails->get_emails();
+		$email_types = array_filter( $email_types, function( $email ) use ( $core_transactional_emails ) {
+			return in_array( $email->id, $core_transactional_emails );
+		} );
```
costasovo: Just an idea. I think we could also support 3rd party emails by having universal email content that would contain only the woo_placeholder block. So when generating the email posts, we could check if there is a post content for the type, and if not, we would use the universal one. We could then allow 3rd party integrators to provide their own block content via a filter.
For now, I think core emails are fine, but this way, we can add basic support for correctly registered 3rd party emails soon.

plugins/woocommerce/src/Internal/EmailEditor/WCTransactionalEmails/WCEmailTemplateGenerator.php
```diff
@@ -80,15 +165,19 @@ public function generate_email_templates() {
...
 	private function generate_single_template($email_type, $template_data) {
-		// Skip if template already exists
-		if ($this->template_manager->template_exists($email_type)) {
-			return;
-		}
-
 		$post_data = [
 			'post_type' => Integration::EMAIL_POST_TYPE,
-			'post_status' => 'publish',
+			'post_status' => $template_data['enabled'] ? 'publish' : 'draft',
```
costasovo: I see! That's why the `Order details` aka `customer_invoice ` email is a draft. This one is manually sent so it will never be set as enabled. You can test if the email is manual by calling `WC_Email::is_manual` then perhaps the logic could be `$template_data['enabled'] || $template_data['manual'] ? 'publish' : 'draft'`

triple0t:
Thank you for the quick review, @costasovo 

>  I noticed that the "Order details" email is draft all the other emails have state publish. Is that on purpose? The "Order details" email is triggered manually from the order's admin screen. It can't be enabled or disabled. I would expect it being published as well.

That's true. We aren't doing anything with the email post status right now, so we can enable all emails. 
The idea is that if a user intentionally disables a WooCommerce transactional email, the email post should reflect that status. I think we are planning to do that in another issue so I will update this PR.

>  I've experienced some issues with the generating because initially, I had some leftovers in the options table, and I had woocommerce_email_templates_customer_processing_order_post_id, which pointed to a nonexistent email. The email was skipped and was not generated. Perhaps we could make the generating more robust and check whether the records point to existing woo_email posts. What do you think?

I initially implemented this to check if the email post exists but switched it to using the options table for performance reasons. It wasn't very practical to run get_post 10 to 12 times. I added a [hook](https://github.com/woocommerce/woocommerce/blob/34bc33ca0ed53913541cb1a6e5c69b57c87c424b/plugins/woocommerce/src/Internal/EmailEditor/Integration.php#L86) to notify us when a post is deleted and to perform options value [cleanup](https://github.com/woocommerce/woocommerce/blob/34bc33ca0ed53913541cb1a6e5c69b57c87c424b/plugins/woocommerce/src/Internal/EmailEditor/Integration.php#L164) on post delete.
We shouldn't experience the same issue again.


> I assume that's expected behavior, as it might not make sense to send an email for some transitions.

I believe so, yes.

>  I was thinking that in the future, we may put the generating process into a background job, so that it is not dependent on visiting the emails page. What do you think?

Yes, please. I agree with you.

plugins/woocommerce/includes/admin/class-wc-admin.php
```diff
@@ -7,6 +7,8 @@
  * @version  2.6.0
  */
 
+declare(strict_types=1);
```
costasovo, line 10: Is that intentional? I have a concern that in such an old class we may run into issues because some variables in WP tend to be sometimes string sometimse an integer...

costasovo (commented):
@triple0t Wow, such an incredible amount of work and functionality added! Great job! I think it is good overall. I just have a couple of questions and suggestions.

I also think it would be great to add some unit tests for the classes that generate the emails.

plugins/woocommerce/src/Internal/EmailEditor/WCTransactionalEmails/WCEmailTemplateGenerator.php
triple0t: Updated [here](https://github.com/woocommerce/woocommerce/pull/56567/commits/75a6a46927eedc62c2fd6aac6c02ba4e538b181b)

plugins/woocommerce/includes/emails/class-wc-email-cancelled-order.php
triple0t: > I'm not sure this is a good idea. This will force changes in emails added by third parties who respect the WC_Email class ([example](https://github.com/woocommerce/woocommerce-subscriptions/blob/c7da0607a98681f555925126d2445e80b18e32aa/includes/payment-retry/emails/class-wcs-email-payment-retry.php#L17)) to add this method for the support.

Not really. It will not affect third-party emails without the email editor posts.

> What do we want to achieve here? The commit message doesn't give much insight here. Can't we do it in the patent WC_Email class?

We execute this [method](https://github.com/woocommerce/woocommerce/blob/34bc33ca0ed53913541cb1a6e5c69b57c87c424b/plugins/woocommerce/src/Internal/EmailEditor/WooContentProcessor.php#L110) when we want to get only the [order details part](https://github.com/woocommerce/woocommerce/blob/34bc33ca0ed53913541cb1a6e5c69b57c87c424b/plugins/woocommerce/templates/emails/block/general-block-email.php) of the template for the Woo content placeholder.

plugins/woocommerce/src/Internal/EmailEditor/WCTransactionalEmails/WCEmailTemplateGenerator.php
triple0t: I share the same thoughts here. I would love to add a filter here to support creating initial email posts for third-party extensions, but I recall Pavel asking us to hold off on doing that for now. I may add it, but mark it as experimental.

plugins/woocommerce/src/Internal/EmailEditor/WCTransactionalEmails/WCEmailTemplateGenerator.php
triple0t: Noted. Will do.

plugins/woocommerce/includes/admin/class-wc-admin.php
triple0t: Yes... but we don't have any type-hint here, so it should be fine.

plugins/woocommerce/src/Internal/EmailEditor/WCTransactionalEmails/WCEmailTemplateManager.php
triple0t: Done in https://github.com/woocommerce/woocommerce/pull/56567/commits/3ac5319bdce5c108d4d1d682e0bc9569e173eb8e

plugins/woocommerce/src/Internal/EmailEditor/WCTransactionalEmails/WCEmailTemplateGenerator.php
triple0t: Done in https://github.com/woocommerce/woocommerce/pull/56567/commits/3ac5319bdce5c108d4d1d682e0bc9569e173eb8e

plugins/woocommerce/src/Internal/EmailEditor/WCTransactionalEmails/WCEmailTemplateGenerator.php
triple0t: Done in https://github.com/woocommerce/woocommerce/pull/56567/commits/d9cf3bd20d4013f88a4b52766364afe876311c43

plugins/woocommerce/src/Internal/EmailEditor/WCTransactionalEmails/WCEmailTemplateGenerator.php
triple0t: I added support for loading an initial block email content [here](https://github.com/woocommerce/woocommerce/pull/56567/commits/8f42f47f268afdefb2a36382e05e4364dd98250b) when not provided by the theme/plugin developer and added a [filter](https://github.com/woocommerce/woocommerce/pull/56567/commits/d9cf3bd20d4013f88a4b52766364afe876311c43#) for including an email ID when fetching the list of transactional emails.

triple0t:
Thank you for the in-depth review, @costasovo 
I have addressed your feedback and added more test cases to the PR.

**Updates**
* Renamed the WCEmailTemplateGenerator and WCEmailTemplateManager classes
* Added a filter for including an email ID when fetching the list of transactional emails. (`woocommerce_transactional_emails_for_block_editor`)
* Allow the process of generating the initial block email contents to run from any WooCommerce admin pages
* Refactor WCTransactionalEmailPostsGenerator class to only generate the email content when required compared to the previous implementation where all emails were generated on init.
* Added support for loading an initial block email content when not provided in `theme/woocommerce/emails/block/email-id.php`
* Add more test cases.

**Third-party Integration/Extension** 

The following steps must be taken to allow the block editor to generate and manage extension emails.
1. Ensure the email is registered using the `woocommerce_email_classes` filter
2. Add the email ID to the required list using the `woocommerce_transactional_emails_for_block_editor` filter
3. Add an initial block email content to `theme/woocommerce/emails/block/email-id.php` (content should be a valid WordPress Block syntax)
4. Clear the transient `wc_email_editor_initial_templates_generated` and visit any WooCommerce admin page.

Note: Block content customization is done using Personalization tags; examples [here](https://github.com/woocommerce/woocommerce/tree/75a6a46927eedc62c2fd6aac6c02ba4e538b181b/plugins/woocommerce/templates/emails/block). We are still in the process of building the integration points for third-party Personalization tag integration, i.e., allowing extension developers to register custom Personalization tags for their own email contents.

plugins/woocommerce/src/Internal/EmailEditor/WCTransactionalEmails/WCTransactionalEmailPostsGenerator.php
```diff
@@ -109,15 +110,41 @@ function ( $acc, $email ) {
...
+		try {
+			$template_html = wc_get_template_html(
+				$template_name,
+				array(
+					'order'         => $email->object,
+					'sent_to_admin' => true,
+					'plain_text'    => false,
+					'email'         => $email,
+				)
+			);
```
costasovo: Do we need to provide these data? At the time of generating email posts we use personalized tags. What order are we passing there?

plugins/woocommerce/src/Internal/EmailEditor/WCTransactionalEmails/WCTransactionalEmailPostsGenerator.php
```diff
@@ -72,7 +72,7 @@ final public function init() {
 	 * This function initializes the default templates for the core transactional emails.
 	 * It fetches all the emails from WooCommerce and filters them to include only the core transactional emails.
 	 */
-	private function init_default_templates() {
+	public function init_default_templates() {
```
costasovo: Could we use `init_default_transactional_emails` or `init_default_email_contents` or just something else but templates? I see it is used in many other places. Theoretically, it is a template for the content. Maybe we could at least say they are content templates. I would like to not use templates with this context to avoid confusion with the main email template. In the future, we will add multiple email templates and I'm afraid it will be confusing what a template is.

plugins/woocommerce/includes/emails/class-wc-email-cancelled-order.php
costasovo: @triple0t Thanks for explaining! I think I understand now.

I thought we could achieve something similar by modifying the original email templates where we could use a feature switch and comment and skip the text parts we want to move to blocks and keep only the main content. But I see we would get basically what is now the general email content in many cases.

My only concern now is what happens if a third party extends the WC_Email class and is listed in WC_Emails. We will create the default content post from the fallback block content, right? So it will have a post. Then when sending we will replace the Woo_block with the general email content from the main WC_Email:get_block_editor_email_template_content, and we will ignore the third-party HTML template. 
So basically, a third party needs to provide a block template to be able to add some customized content. Or am I missing something? 

I thought that with the approach mentioned above, we could still grab the full content from the HTML templates (including text parts, etc.). from third-party emails and use that to replace the Woo placeholder block. That way, we could keep sending reasonable content also for the email added by a third party.

costasovo (commented):
@triple0t Thank you for the updates! Kudos for adding the fallback post content!

I have a couple of additional questions. Please take a look at the comments.

plugins/woocommerce/includes/emails/class-wc-email-cancelled-order.php
triple0t: > I thought we could achieve something similar by modifying the original email templates where we could use a feature switch and comment and skip the text parts we want to move to blocks and keep only the main content. But I see we would get basically what is now the general email content in many cases.

I thought of that approach as well but didn't like it because it's not very clean.
It has the following limitations
1. We have no control over the email content when the user has overridden the transaction email template
2. When a user has overridden the transactional email template, what we set and display on the email editor may differ from what is rendered in the email client.
3. It makes extending the WooCommerce email templates complicated. For example, now that you have multiple if-else statements, you don't know what they are used for or if you even need them in your custom template. 
4. It makes it difficult for us to implement the project's second phase, which is creating a block version of the WooCommerce order details. We would still have to do something similar to what I've done here.

> My only concern now is what happens if a third party extends the WC_Email class and is listed in WC_Emails. We will create the default content post from the fallback block content, right?

No, we don't create the email post by default. Unless the extender adds the email ID to the required list using the `woocommerce_transactional_emails_for_block_editor` filter. Steps [here](https://github.com/woocommerce/woocommerce/pull/56567#issuecomment-2747115685).
That is only when we create the email post and use the fallback block content when not provided.

costasovo: > I thought of that approach as well but didn't like it because it's not very clean.
It has the following limitations

@triple0t Thanks for the explanation and the reasons. I forgot about the template override, and it makes perfect sense for the second phase and overall, your approach seems more reasonable 👍 

> No, we don't create the email post by default. Unless the extender adds the email ID to the required list using the woocommerce_transactional_emails_for_block_editor filter. Steps https://github.com/woocommerce/woocommerce/pull/56567#issuecomment-2747115685.
That is only when we create the email post and use the fallback block content when not provided.

Oh! Sorry, I missed that. So, it will be opt-in for the third party. Nice! Now it clicked! So, in that case I think that in the new email listing, the emails without the post should still open the old email manage screen. I can update it in my branch. What do you think?

Sorry for the many questions, and thanks for your patience! I wanted to properly understand the overall idea. I could not make or failed to make the full picture properly from the code and commit messages.

plugins/woocommerce/src/Internal/EmailEditor/WCTransactionalEmails/WCTransactionalEmailPostsGenerator.php
triple0t: You're right. We don't need this here. Removed in https://github.com/woocommerce/woocommerce/pull/56567/commits/759a1c1fc95f0748344f7d71eb8e001521285545

plugins/woocommerce/src/Internal/EmailEditor/WCTransactionalEmails/WCTransactionalEmailPostsGenerator.php
triple0t: Noted. Implemented here https://github.com/woocommerce/woocommerce/pull/56567/commits/759a1c1fc95f0748344f7d71eb8e001521285545

plugins/woocommerce/includes/emails/class-wc-email-cancelled-order.php
triple0t: > Oh! Sorry, I missed that. So, it will be opt-in for the third party.

@costasovo, Yes, that is correct.

> So, in that case I think that in the new email listing, the emails without the post should still open the old email manage screen. I can update it in my branch. What do you think?

I think that's fine. It would make it easy to still manage other emails.

plugins/woocommerce/src/Internal/EmailEditor/WooContentProcessor.php
```diff
@@ -88,7 +107,7 @@ private function capture_woo_content( \WC_Email $wc_email ): string {
 		remove_all_filters( 'woocommerce_email_header' );
 		remove_all_filters( 'woocommerce_email_footer' );
 
-		$woo_content = $wc_email->get_content_html();
+		$woo_content = $wc_email->get_block_editor_email_template_content();
```
costasovo: After the previous conversation I realized that now, with the new `get_block_editor_email_template_content` and copied template with the main content we no longer need to disable and re-enable `woocommerce_email_header` and `woocommerce_email_footer` filters because they are no longer included in the template that is generated via `get_block_editor_email_template_content`. We actually don't event need the method as all we need to call is `$wc_email->get_block_editor_email_template_content();`

I tested it locally and it seems to work without them. This is a nice side-effect of your idea with the copy of general email content. 👍 

I'm sorry that I didn't realize it in the first or second review. Could you please remove the manipulation with filters?

plugins/woocommerce/src/Internal/EmailEditor/WCTransactionalEmails/WCTransactionalEmailPostsGenerator.php
```diff
@@ -0,0 +1,273 @@
...
+		if ( is_wp_error( $template_html ) || empty( $template_html ) || $has_template_error ) {
+			$default_template_name = 'emails/block/default-block-content.php';
+			$template_html         = wc_get_template_html(
+				$default_template_name,
+				array(
+					'order'         => $email->object,
+					'sent_to_admin' => true,
+					'plain_text'    => false,
+					'email'         => $email,
+				)
```
costasovo: I missed this place before. I assume we can use an empty array here the same as the above.

costasovo (commented):
@triple0t, Thank you for addressing my comments and for your patience! 
I have two more comments with tiny issues, but I think we should rather address them now so that we don't forget.

plugins/woocommerce/src/Internal/EmailEditor/WooContentProcessor.php
triple0t: No problem. I missed this during the last cleanup as well. Done in https://github.com/woocommerce/woocommerce/pull/56567/commits/472982c792130e4a62e92253f593a64595afa7d6

plugins/woocommerce/src/Internal/EmailEditor/WCTransactionalEmails/WCTransactionalEmailPostsGenerator.php
triple0t: Thanks for catching it. Done in https://github.com/woocommerce/woocommerce/pull/56567/commits/472982c792130e4a62e92253f593a64595afa7d6

costasovo (approved):
@triple0t Great job!

I performed another round of manual testing and it works as expected.

Here is a screenshot of the generated emails:

Screenshot 2025-03-25 at 16 52 22

Here is the recording of testing various emails:

https://github.com/user-attachments/assets/f0bcafb7-3786-4197-bd0d-c9a83ff4994b


I see there are a couple of jobs failing. I guess a rebase might help with that.

PR #56584, opened 2025-03-21
[e2e-tests] Introduce `wpCoreVersion` worker fixture for Blocks tests

WunderBart:
## What?

This PR introduces a new worker-scoped fixture wpCoreVersion that detects the WordPress version being tested against and allows tests to adapt their behavior for backwards compatibility when functionality differs across WordPress versions.

> [!NOTE]
> This PR also skips some tests as agreed in p1742489145338069/1742483773.434129-slack-C02FL3X7KR6

### Key features:

- Integrated with the existing test environment (can be shared with any test and fixture).
- The version check/request happens only once per worker process (no extra delays in execution).
- Version is parsed to a number (e.g., "6.4.1" → 6.4) for simpler numeric comparisons.

### Example usage:

**Conditional actions**
```js
test('some feature', async ({ wpCoreVersion, page }) => {
    if (wpCoreVersion < 6.8) {
        // Handle pre-6.8 behavior
    } else {
        // Handle current behavior
    }
});
```

**Skipping tests**
```js
test('some feature', async ({ wpCoreVersion, page }) => {
    test.skip(wpCoreVersion < 6.8, 'This test requires WordPress 6.8 or higher');
    
    // Test code that only works in WP 6.8+
    await page.click('.new-feature-button');
});

// You can also skip at the describe level
test.describe('some block', () => {
    test.skip(({ wpCoreVersion }) => wpCoreVersion < 6.8, 'These tests require WordPress 6.8 or higher');
    
    test('feature A', async ({ page }) => {
        // Test code
    });

    test('feature B', async ({ page }) => {
        // Test code
    });
});
```

## How to test

All Blocks E2E tests should be passing for WP 6.6, 6.7 and 6.8.

### DIFF
--- a/plugins/woocommerce/client/blocks/tests/e2e/tests/classic-template/classic-template.block_theme.spec.ts
+++ b/plugins/woocommerce/client/blocks/tests/e2e/tests/classic-template/classic-template.block_theme.spec.ts
@@ -230,6 +230,7 @@ test.describe( `${ blockData.name } Block `, () => {
 	test( `is still available after resetting a modified WC template`, async ( {
 		admin,
 		editor,
+		wpCoreVersion,
 	} ) => {
 		await admin.visitSiteEditor( {
 			postId: `woocommerce/woocommerce//single-product`,
@@ -271,9 +272,7 @@ test.describe( `${ blockData.name } Block `, () => {
 		const editButton = editor.page.getByRole( 'menuitem', {
 			name: 'Edit',
 		} );
-
-		// Keep WP v6.7 compatibility.
-		if ( await editButton.isHidden() ) {
+		if ( wpCoreVersion >= 6.8 ) {
 			await actionsButton.click();
 		}
 
--- a/plugins/woocommerce/client/blocks/tests/e2e/tests/templates/template-customization.block_theme.spec.ts
+++ b/plugins/woocommerce/client/blocks/tests/e2e/tests/templates/template-customization.block_theme.spec.ts
@@ -13,6 +13,11 @@ import {
 import { CUSTOMIZABLE_WC_TEMPLATES } from './constants';
 
 test.describe( 'Template customization', () => {
+	test.skip(
+		( { wpCoreVersion } ) => wpCoreVersion <= 6.6,
+		'Skipping on WP 6.6 and below'
+	);
+
 	CUSTOMIZABLE_WC_TEMPLATES.forEach( ( testData ) => {
 		const userText = `Hello World in the ${ testData.templateName } template`;
 		const fallbackTemplateUserText = `Hello World in the fallback ${ testData.templateName } template`;
--- a/plugins/woocommerce/client/blocks/tests/e2e/utils/test.ts
+++ b/plugins/woocommerce/client/blocks/tests/e2e/utils/test.ts
@@ -115,13 +115,14 @@ const test = base.extend<
 	},
 	{
 		requestUtils: RequestUtils;
+		wpCoreVersion: number;
 	}
 >( {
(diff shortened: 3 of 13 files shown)


plugins/woocommerce/client/blocks/tests/e2e/tests/templates/template-customization.block_theme.spec.ts
```diff
@@ -13,6 +13,11 @@ import {
 import { CUSTOMIZABLE_WC_TEMPLATES } from './constants';
 
 test.describe( 'Template customization', () => {
+	test.skip(
+		( { wpCoreVersion } ) => wpCoreVersion <= 6.6,
+		'Skipping on WP 6.6 and below'
+	);
+
```
WunderBart, line 20: Template tests (this and below) skipped as per p1742489145338069/1742483773.434129-slack-C02FL3X7KR6.

WunderBart:
Running the daily workflow to verify 6.6, 6.7 and 6.8 tests pass against this branch here: https://github.com/woocommerce/woocommerce/actions/runs/13993652307

plugins/woocommerce/client/blocks/tests/e2e/tests/classic-template/classic-template.block_theme.spec.ts
```diff
@@ -271,9 +272,7 @@ test.describe( `${ blockData.name } Block `, () => {
 		const editButton = editor.page.getByRole( 'menuitem', {
 			name: 'Edit',
 		} );
-
-		// Keep WP v6.7 compatibility.
-		if ( await editButton.isHidden() ) {
+		if ( wpCoreVersion >= 6.8 ) {
```
gigitux, line 275: NIT: Should we improve this logic targeting the WordPress - 1 version? In this way, we can easily refactor things when old WordPress version aren't supported anymore.

plugins/woocommerce/client/blocks/tests/e2e/utils/test.ts
```diff
@@ -168,6 +169,22 @@ const test = base.extend<
 		},
 		{ scope: 'worker', auto: true },
 	],
+	wpCoreVersion: [
+		async ( {}, use ) => {
+			const output = await wpCLI( 'core version' );
+			const version = output.stdout.trim().split( '\n' ).at( -1 ) ?? '';
+			const parsedVersion = Number.parseFloat( version );
```
gigitux: WordPress versions should always have only two numbers, correct? I'm asking because this code doesn’t catch ‘6.10’.

gigitux (approved):
The PR LGTM! Thanks for working on this! 🙇 


Would it be better to target always WordPress -1 in the skip logic? This way, refactoring becomes easier when older WordPress versions are no longer supported. If you agree, should it make sense to add this rule to our test code guidelines?

plugins/woocommerce/client/blocks/tests/e2e/utils/test.ts
WunderBart: Great point, this will start failing when we reach x.10 and above. I'll make sure it works as expected.

edit: WP version never gets over x.9 - after that it's usually bumped a major.

WunderBart: Added an explanation in https://github.com/woocommerce/woocommerce/pull/56584/commits/2fc0965248bb1056d15e8bf171932d271adfe3c9

WunderBart:
All the block tests agains `latest-1` (6.6) passed: https://github.com/woocommerce/woocommerce/actions/runs/13993652307. Merging!

PR #56588, opened 2025-03-21
[NOX In-Context] Add Business Verification step

### LINKED ISSUE #56171: [WC Payments NOX in-context flows] Copy and refactor the onboarding wizard
### Description

We need to copy and refactor the WooPayments onboarding wizard into the modal step (fourth step). This includes both our data collection step and the Stripe embedded KYC component. We will use the backend logic of the WooPayments client, only the UI will be copied/forked.

### Acceptance criteria

- The data collection form is present and has the same behavior as in the client (including persisting choices)
- The embedded KYC component is present and has the same behavior as in the client
- We prefill the business location country with the one selected in the Payments Settings page (maybe also prefill the "Store Go Live" timeframe based on Coming soon mode)
- We have two sub-routes that we can target through URLs: one for the data collection sub-step, the other for the embedded KYC sub-step
- Failure to use the embedded component is handled by prompting the user to start the off-site Stripe KYC with a return URL to the modal step route; the rest of the failures are handled the same way they are in the client
- The backend logic remains the same and resides in the WooPayments client
- We use the latest WC component versions
- The step content has the right responsive behavior
- The JS code is organized into its proper place and directory structure

### Designs
Design post: pfHfG4-Mn-p2
Figma: xO8YdpKfaVoP2d8NvweZ5I-fi-6516_128513

### Dev notes

Spike: pe4Hk0-23D-p2
Slack discussion: p1741181729264379-slack-C0800LPNCET

### Additional context
PRD: 1wMfX8CAjDnOp24tcU-PADB0pOt3cz7mGdyZYeRBxx2Y-gdoc

cc @timmy5685 @orcungogus

mordeth:

This PR introduces the Business Verification step part of NOX In-context flows. It includes both pre-KYC data collection and embedded onboarding KYC.

Closes #56171 .

[Screenshot 2025-04-01 at 0 35 10](https://www.loom.com/share/2eebe582d3f24a49a4923a739887c2d7)

1. Check out this PR locally.
2. Ensure your site is already connected to WordPress.com.
3. If you have any WooPayments accounts connected, reset the connection.
4. Apply the diff provided in [this comment](https://github.com/woocommerce/woocommerce/pull/56723#issuecomment-2754995345) to open the modal when clicking “Complete Setup”.
5. Click “Complete Setup”.
- If you haven’t selected payment methods (PMs) yet, you’ll be prompted to do so.
- Select a few payment methods and click Continue.
6. You’ll proceed to the Test Account Creation step. Click “Next step” (note: this step isn’t implemented yet).
7. After being redirected to the Activate Payments step:
- The country from your WooCommerce store should be pre-selected.
- The country dropdown should list only countries supported by WooPayments.
- Select United States.
8. Next, you’ll see the field: “What type of legal entity is your business?”
- Select “Individual”.
9. A new field should immediately appear:
- “What type of goods or services does your business sell?”
- Verify the search functionality in the dropdown is working correctly.
- Select any value.
10. Two more fields should appear:
- “What is your estimated annual Ecommerce revenue (USD)?”
- “What is the estimated timeline for taking your store live?”
11. Before selecting values for those two fields, click Continue.
- You should see validation errors displayed below the unselected fields.
12. Select values for both fields.
13. Before clicking Continue, refresh the page.
- Your previous selections should be preserved.
14. Click Continue, you should be redirected to the embedded KYC onboarding.
15. Fill out all required KYC details.
16. After completing a validated section, refresh the page, data should remain persisted.
17. Once the embedded KYC onboarding is successfully completed, you should be redirected to the next onboarding step.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/steps/business-verification/steps/embedded-kyc.tsx
```diff
@@ -0,0 +1,140 @@
...
+		// To-Do: Track step change.
+	};
+
+	const handleOnExit = async () => {
+		setFinalizingAccount( true );
+
+		try {
+			const response = await finalizeOnboarding(
+				currentStep?.actions?.kyc_session_finish?.href ?? '',
+				'NOX' // To-Do: Replace with the correct source.
```
mordeth: @vladolaru What would be the correct source?

vladolaru: @mordeth the supported sources are [here](https://github.com/Automattic/woocommerce-payments/blob/0024a37df240a6c9cb60deb32f145d4fb596ab99/includes/class-wc-payments-onboarding-service.php#L30-L30). But best not to send anything if you don't have reliable data on hand - it is up to the backend to fill in the gaps.

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/components/custom-select-control/index.tsx
```diff
@@ -0,0 +1,224 @@
+/**
+ * This is a copy of the CustomSelectControl component, found here:
+ * https://github.com/WordPress/gutenberg/tree/7aa042605ff42bb437e650c39132c0aa8eb4ef95/packages/components/src/custom-select-control
```
elazzabi, line 3: I know this was copied over, but I think it's worth noting that the component supports generic props now https://github.com/WordPress/gutenberg/pull/63985.

We can add an issue to track this and potentially use the component directly from `gutenberg`

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/steps/business-verification/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+		...( currentStep?.context?.self_assessment ?? {} ),
+	};
+
+	const handleStepChange = () => window.scroll( 0, 0 );
+
+	return (
+		<div className="settings-payments-onboarding-modal__step-business-verification">
+			<WooPaymentsStepHeader onClose={ () => {} } />
+			<div className="settings-payments-onboarding-modal__step-business-verification-content">
+				<BusinessVerificationContextProvider
```
elazzabi, line 45: I think that using the context here might be overkill. From the implementation, I see it's for holding some data in state, which can be transferred to `BusinessVerificationStep` directly.

Also, with the current implementation of this file, we are not taking full advantage of the context to avoid "props waterfalls" (continuously passing props from parents to children). Our component is completely flat, which makes it easy to pass props directly to children (and avoiding the context).

Example without context:
```tsx
const data = {
	setTouch,
	setError,
	setData,
};

<Stepper
	onStepChange={ handleStepChange }
	onExit={ () => {} }
>
	<Step name="business">
		<OnboardingForm>
			<BusinessDetails data={data} />
		</OnboardingForm>
	</Step>
	<Step name="embedded" showHeading={ false }>
		<EmbeddedKyc data={data} />
	</Step>
</Stepper>
```

I think the context makes more sense if we had the `Stepper` defined like inside the woopayments provider:
```jsx
<Stepper
	onStepChange={ handleStepChange }
	onExit={ () => {} }
	steps={ [
		{
			id: "business",
			// etc..
		}
	] }
>
```

Given the steps aren't flat, we can't pass data via props, and the context makes it easier. But for now, I don't believe it's necessary.

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/steps/business-verification/steps/embedded-kyc.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+		if ( ! continueKyc ) {
+			checkEligibility();
+		} else {
+			setIsEligible( false );
+		}
+	}, [ continueKyc, data ] );
+
+	const handleStepChange = () => {
+		// To-Do: Track step change.
```
elazzabi: Is this a correct To-Do that we need to complete? 🤔

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/steps/business-verification/style.scss
```diff
@@ -0,0 +1,213 @@
...
+$gutenberg-blueberry-focus: #1d35b4;
+
+.settings-payments-onboarding-modal__step#business_verification {
+	display: flex;
+	flex-direction: column;
+	flex: 1;
+	overflow: hidden;
+}
+
+.settings-payments-onboarding-modal__step-business-verification {
```
elazzabi, line 14: I think these two selectors look the same:

`.settings-payments-onboarding-modal__step#business_verification`
`.settings-payments-onboarding-modal__step-business-verification`

We should find a better way to differentiate between the two 😅

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/components/banner-notice/index.tsx
dmallory42, line 1: Some of the components we've added here (`BannerNotice`, `GroupedSelectControl` and `StripeSpinner`) seem like they could be useful in other places, not just in this form. Have we considered whether any of them would be a good fit for adding to `@woocommerce/components`? 

We don't need to handle it in this PR of course (it can be a separate issue that we can tackle outside of the project) but I wanted to raise it to see whether we think it would be worthwhile. 🙂

elazzabi (changes requested):
Reviewed the code changes and left a few comments. Nice work!

I'll do the UI testing next

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/steps/business-verification/components/embedded/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+	useEffect( () => {
+		const initializeStripe = async () => {
+			try {
+				const accountSession = await createKycAccountSession(
+					onboardingData,
+					currentStep?.actions?.kyc_session?.href ?? '',
+					isPoEligible
+				);
+
+				// To-Do: Track the embedded component redirection event.
```
dmallory42, line 67: Do we have an issue for this? If not, maybe it is worth creating one to make sure we handle it 🙂 

```suggestion
// @todo: Track the embedded component redirection event
```

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/steps/business-verification/components/form/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+	GroupedSelectField,
+	GroupedSelectFieldProps,
+	SelectField,
+	SelectFieldProps,
+	TextField,
+	TextFieldProps,
+} from './fields';
+import { useBusinessVerificationContext } from '../../data/business-verification-context';
+import { OnboardingFields } from '../../types';
+import { completeSubStep, isPreKYCComplete } from '../../utils';
```
dmallory42: Nitpick: It looks like in other places in the code we use `Kyc` (e.g. `AccountKycSession`) rather than `KYC` - what do you think about using `isPreKycComplete` instead here?

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/steps/business-verification/steps/embedded-kyc.tsx
dmallory42: Could we not use `wcadmin-settings-page` here? Or would we want to add a new source, specifically for NOX?

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/steps/business-verification/style.scss
```diff
@@ -0,0 +1,213 @@
...
+				max-width: 615px;
+				width: 100%;
+			}
+
+			&__cta {
+				display: block;
+				width: 100%;
+				height: 40px; // Matching the updated WP Component. We can remove this when we update Components version.
+				margin-top: $gap-large;
+				background: var(--wp-components-color-accent, $gutenberg-blueberry); // override the MOX CTA to use Gutenberg Blueberry.
```
dmallory42, line 133: I wonder about whether we should fall back to the WP theme colors rather than using our own blueberry colours here? Wouldn't it look out of place if the merchant changed their theme?

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/steps/business-verification/utils/index.ts
```diff
@@ -0,0 +1,226 @@
...
+								type.key
+						  ]
+						: businessTypeDescriptionStrings.generic[ type.key ],
+				} ) ),
+			} ) )
+			.sort( ( a, b ) => a.name.localeCompare( b.name ) ) || []
+	);
+};
+
+/**
```
dmallory42: Nitpick: I wonder whether we should take this opportunity to extract the API request functions to a separate file (like `actions.tsx`)? Looking at it now, I'm not sure the utils file is the best place for them.

plugins/woocommerce/client/admin/client/settings-payments/onboarding/types.ts
```diff
@@ -4,6 +4,14 @@
 import { type ReactNode } from 'react';
 import { type RecommendedPaymentMethod } from '@woocommerce/data';
 
+/**
+ * Internal dependencies
+ */
+import {
+	Country,
+	MccsDisplayTreeItem,
+} from './providers/woopayments/steps/business-verification/types'; // To-do: Maybe move to @woocommerce/data
```
dmallory42, line 13: Agree with this TODO, IMO we can do it as a follow up issue rather than as part of this PR because it's already quite large. 😄

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/steps/business-verification/style.scss
dmallory42: What is the difference between the two? (as in, where would they both be used?) 

I guess one way of doing it is trying to have a hierarchy of gradually more and more specific selectors.

dmallory42:
Screenshot 2025-03-28 at 11 07 05

A small thing I noticed while testing: The font color is inconsistent between the `"What type of goods or services does your business sell?"`field and all the other ones.

dmallory42 (changes requested):
Thanks for working on this! It's already looking great. I left a few comments to take a look at, I don't think any of htem are blocking, however.

I did experience a problem on step 14 of the testing, I saw an error when trying to load the embedded onboarding, and it looks like it was because of a 404 error on this route: `/wp-json/http://localhost:8082/wp-json/wc-admin/settings/payments/woopayments/onboarding/step/business_verification/kyc_session?self_assessment%5Bbusiness_name%5D=WooCommerce%20Payments%20Dev&self_assessment%5Bmcc%5D=most_popular__clothing_and_apparel&self_assessment%5Bsite%5D=https%3A%2F%2Fwcpay.test&self_assessment%5Bcountry%5D=US&self_assessment%5Bbusiness_type%5D=individual&self_assessment%5Bannual_revenue%5D=from_1m_to_20m&self_assessment%5Bgo_live_timeframe%5D=from_3_to_6months&progressive=false&_locale=user`

Screenshot 2025-03-28 at 11 09 31

Let me know if I missed something and I can retest!

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/steps/business-verification/index.tsx
mordeth: Personally, I wouldn’t consider it overkill, using context is essentially the same as defining the states in a data object and passing it down to children. It’s just a different way of organizing the same logic, but the context definition looks much better. :)

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/steps/business-verification/style.scss
mordeth: I agree that could be confusing, the html structure is as follows:
```
<div id=""business_verificaiton" class="settings-payments-onboarding-modal__step">
   <div class="settings-payments-onboarding-modal__step-business-verification">
   </div>
</div>
```

I plan to change the first css selector to `#business_verification.settings-payments-onboarding-modal__step` which should sort the confusion. WDYT?

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/steps/business-verification/components/form/index.tsx
mordeth: Good catch, updated!

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/steps/business-verification/style.scss
mordeth: Great question! That’s actually how button backgrounds are currently implemented on the WooPayments side, they don’t follow the WordPress theme colors. @orcungogus, should the button colors within the modal align with the WP theme styles?

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/steps/business-verification/style.scss
dmallory42: Yeah, I think that reads easier!

mordeth: Updated with https://github.com/woocommerce/woocommerce/pull/56588/commits/ddb547f51273e97f38f207685ef2600527831f0e

plugins/woocommerce/client/admin/client/settings-payments/onboarding/types.ts
mordeth: Sounds good! I’ll take care of this in a follow-up PR.

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/steps/business-verification/steps/embedded-kyc.tsx
mordeth: Yes, I have create an issue to handle the tracking implementation. https://github.com/woocommerce/woocommerce/issues/56826

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/steps/business-verification/components/embedded/index.tsx
mordeth: Yes, I have created an issue for it https://github.com/woocommerce/woocommerce/issues/56826.

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/components/custom-select-control/index.tsx
mordeth: Thank you for the reminder! I have created an issue to replace the forked CustomSelectControl with the one from `@wordpress/components` https://github.com/woocommerce/woocommerce/issues/56827.

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/steps/business-verification/utils/index.ts
mordeth: That makes sense! I have move all the API request functions to an external file in https://github.com/woocommerce/woocommerce/pull/56588/commits/ebe28001789671b793a9e97ea3c29921e903aa1e

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/components/banner-notice/index.tsx
mordeth: I’m not certain if any of these will be used elsewhere, but we can definitely address exporting them in a follow-up PR.

dmallory42: Yeah, I guess we can address it if we see a need to use them elsewhere in future.

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/steps/business-verification/steps/business-details.tsx
dmallory42, line 1: This is nitpicky but having steps inside a step feels like it might get confusing. Is there a way we can rethink that? I'm not suggesting making any big change to the hierarchy, but maybe using a different terminology for the individual parts of a specific step.

mordeth: That makes sense, thanks! Renamed the folder to `sections` to avoid any confusion.

dmallory42 (approved):
Thanks for all your work on this! Tested again after the latest round of changes and everything is working as expected. Let's :shipit: !

oaratovskyi (approved):
Manual testing only - works as expected 🚢

PR #56590, opened 2025-03-21
Import missing class in GenericStatsController

szepeviktor:
Added 1 class import. Because it was missing, causing a fatal error.

Please, please, please use **static analysis** while developing WooCommerce.


chihsuan (approved):
Thanks for the fix, looks good!

PR #56613, opened 2025-03-21
Error handling and propagation on My subscriptions refresh

anaspk:

Closes # .

| Before | After |
| ------ | ----- |
|        | Screenshot 2025-03-26 at 3 57 39 PM |
|        | Screenshot 2025-03-26 at 4 07 38 PM |

1. Checkout this branch on your local environment.
2. Build assets for WooCommerce Admin: `pnpm --filter='@woocommerce/plugin-woocommerce' watch:build:admin`
3. Ensure that your WC Core development site is connected to your local WooCommerce.com development environment.
4. In the database of your WC Core development environment, in `wp_options` table,  corrupt the values of `access_token` and `access_token_secret` for the option_name `woocommerce_helper_data`.
5. Click the Refresh button on [My subscriptions page](http://wc-core.test:8888/wp-admin/admin.php?page=wc-admin&tab=my-subscriptions&path=%2Fextensions).
6. Notice that an error message is displayed indicating that the Authentication has failed.
7. Revert the changes you made to `woocommerce_helper_data` to make it corrupt in step 4 above.
8. On your local WooCommerce.com development environment, in `api-rate-limiting-check.php` , update the rate limit for /subscriptions endpoint to some low value (e.g. `3`) from the default value `120`.
9. Click the Refresh button on [My subscriptions page](http://wc-core.test:8888/wp-admin/admin.php?page=wc-admin&tab=my-subscriptions&path=%2Fextensions).
10. Notice that an error message is displayed indicating that the WooCommerce.com API returned HTTP status code 429.
11. While the subscriptions refresh is failing due to either of the above errors, browse different areas of the site to make sure that nothing else is broken due to the changes in this PR e.g. none of the exceptions are left unhandled in any of the flows.
12. Also try refreshing the [My subscriptions page](http://wc-core.test:8888/wp-admin/admin.php?page=wc-admin&tab=my-subscriptions&path=%2Fextensions) completely (by clicking the Refresh button on the browser) to make sure that it still works as expected.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

Handle and display any errors that occur while refreshing WooCommerce.com subscriptions list

</details>


plugins/woocommerce/includes/admin/helper/class-wc-helper.php
```diff
@@ -1691,43 +1717,53 @@ public static function get_subscriptions() {
...
+				throw new Exception( $request->get_error_message() );
+			}
 
-		if ( wp_remote_retrieve_response_code( $request ) !== 200 ) {
-			set_transient( $cache_key, array(), 15 * MINUTE_IN_SECONDS );
-			return array();
-		}
+			$code = wp_remote_retrieve_response_code( $request );
+			if ( 200 !== $code ) {
+				// translators: %d: HTTP status code.
```
thilinah: In case of a non 200 response, we need to set the transient to an empty array. Like in the previous implementation. ( E.g `set_transient( $cache_key, array(), 15 * MINUTE_IN_SECONDS );`).

This logic was there to prevent a store from keep calling WCCOM in short intervals when subscription request is failing, for an example due to rate-limiter.

plugins/woocommerce/includes/admin/helper/class-wc-helper.php
```diff
@@ -1589,27 +1605,36 @@ public static function get_product_usage_notice_rules() {
...
+			if ( is_wp_error( $request ) ) {
+				throw new Exception( $request->get_error_message() );
+			}
 
-		$data = json_decode( wp_remote_retrieve_body( $request ), true );
-		if ( empty( $data ) || ! is_array( $data ) ) {
-			$data = array();
-		}
+			$code = wp_remote_retrieve_response_code( $request );
+			if ( 200 !== $code ) {
```
thilinah, line 1624: Same issue is also here: https://github.com/woocommerce/woocommerce/pull/56613/files#r2011573478

plugins/woocommerce/includes/admin/helper/class-wc-helper.php
```diff
@@ -1691,43 +1717,53 @@ public static function get_subscriptions() {
...
+			}
 
-		if ( wp_remote_retrieve_response_code( $request ) !== 200 ) {
-			set_transient( $cache_key, array(), 15 * MINUTE_IN_SECONDS );
-			return array();
-		}
+			$code = wp_remote_retrieve_response_code( $request );
+			if ( 200 !== $code ) {
+				// translators: %d: HTTP status code.
+				throw new Exception( sprintf( __( 'WooCommerce.com API returned HTTP status code %d.', 'woocommerce' ), $code ) );
```
thilinah: The most common error states for subscription request is 429 and 403. For these response codes, rather than showing a generic status code, based error, we can make it more descriptive.

E.g: 
- for 429: You have exceeded the request limit. Please try again after a few minutes.
- for 403: We're having trouble authenticating your store. Please try again. If the issue persists, disconnect your store from WooCommerce.com and reconnect.

hesahesa: Agree with @thilinah that we probably should make the message more descriptive and not show the HTTP status code in the notice?

plugins/woocommerce/includes/admin/helper/class-wc-helper.php
anaspk: Added it back in https://github.com/woocommerce/woocommerce/pull/56613/commits/c98b6ed5a368362dbd0a9b3929417bf9c0abcbab

plugins/woocommerce/includes/admin/helper/class-wc-helper.php
anaspk: Added it back in https://github.com/woocommerce/woocommerce/pull/56613/commits/c98b6ed5a368362dbd0a9b3929417bf9c0abcbab

plugins/woocommerce/includes/admin/helper/class-wc-helper.php
anaspk: Thanks for the suggestions @thilinah , I have updated the error messages in https://github.com/woocommerce/woocommerce/pull/56613/commits/c98b6ed5a368362dbd0a9b3929417bf9c0abcbab and https://github.com/woocommerce/woocommerce/pull/56613/commits/7704452854db9354562328c7145b50b04f4c8e4f 

@hesahesa I think it is fine to show the HTTP status code in case of less common errors as it will help with debugging both the user themselves and also our HE if they contact support. WDYT?

anaspk:
Thanks for the review @thilinah , I have addressed your feedback and also updated the screenshots in issue description with the updated error messages. Please take another look when you have time.

plugins/woocommerce/includes/admin/helper/class-wc-helper.php
hesahesa: @anaspk yes i think on a less common HTTP status code we can show it in the notice

plugins/woocommerce/client/admin/client/marketplace/components/my-subscriptions/table/actions/refresh-button.tsx
```diff
@@ -44,7 +44,7 @@ export function RefreshButton() {
 							'Error refreshing subscriptions: %s',
 							'woocommerce'
 						),
-						error.message
+						error.data.message
```
hesahesa, line 47: i'm trying to understand why we need the `error.data` here?

the `error` is an instance of this, right? https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch#exceptions or is it a different thing?

anaspk: @hesahesa I think the structure of response inside this error object is whatever is returned from backend. In our case, we return response using [`wp_send_json_error` here](https://github.com/woocommerce/woocommerce/pull/56613/files#diff-4cf137656b8d29bec4423f31fb2781a2d67d2866f172c2620ae3a5d4a2d75294R152). `wp_send_json_error` [automatically wraps](https://developer.wordpress.org/reference/functions/wp_send_json_error/#more-information) whatever `value` we provide it in a top level `data` key. That's what we are unwrapping here.

thilinah:
@anaspk I guess "Try again" link is not introduced on this PR. But seems like there is an additional space before the link.

Screenshot 2025-03-27 at 17 35 46
Screenshot 2025-03-27 at 17 42 32

Also, we might not need this link on error. If it's hard to be removed without affecting other errors, maybe we should at least fix the spacing and the page refresh happens after clicking the link.

anaspk:
@thilinah The space was coming from the flex gap applied to the whole notice component via the CSS.

Screenshot 2025-03-28 at 11 58 27 AM

I have removed the Try again button in https://github.com/woocommerce/woocommerce/pull/56613/commits/dfceadb67c2e32f9e6bc53fbb7a686781b861d8e . Now it looks like this:

Screenshot 2025-03-28 at 12 00 58 PM

thilinah (approved):
Thanks @anaspk, LGTM, tested and works as expected.

PR #56615, opened 2025-03-21
Product Image Block: Support Inner Blocks

### LINKED ISSUE #55996: Spike: Blockify the Product Image Block's On-Sale Badge
One of the features our Product Image block is missing is the ability to style the on-sale badge. Rather than trying to add support for this, we're going to utilize our existing On-Sale Badge block instead.

**Acceptance Criteria**

- Product Image block should support inner blocks.
- On-Sale Badge block should be useable within the Product Image block and work correctly with Product Collection blocks.
- Replaced on-sale toggle with On-Sale Badge block automatically added with Product Image.
- Product Collection templates should add the On-Sale Badge block with correct default positioning.
- Container blocks (group, row, column, grid) should be usable within Product Image blocks for customizing the layout.

ObliviousHarmony:

This pull request adds support for inner blocks to the Product Image block and migrates the sale badge into one. This allows users to have access to the full customizability of the On-Sale Badge block as well as place any other block on top of the product image.

Closes https://github.com/woocommerce/woocommerce/issues/55996

| Before | After |
| ------ | ----- |
|        |       |

1.
2.
3.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/edit.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/edit.tsx
@@ -168,7 +168,9 @@ const Edit = ( {
 				</PanelBody>
 			</InspectorControls>
 			<Block { ...{ isEditor: true, ...attributes, ...context } }>
-				<InnerBlocks />
+				<div className="wc-block-components-product-image-inner-blocks">
+					<InnerBlocks />
+				</div>
 			</Block>
 		</div>
 	);
--- a/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/edit.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/edit.tsx
@@ -2,14 +2,17 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { InspectorControls, useBlockProps } from '@wordpress/block-editor';
+import {
+	InnerBlocks,
+	InspectorControls,
+	useBlockProps,
+} from '@wordpress/block-editor';
 import { createInterpolateElement, useEffect } from '@wordpress/element';
 import { getAdminLink, getSettingWithCoercion } from '@woocommerce/settings';
 import { isBoolean } from '@woocommerce/types';
 import type { BlockEditProps } from '@wordpress/blocks';
 import { ProductQueryContext as Context } from '@woocommerce/blocks/product-query/types';
 import {
-	Disabled,
 	PanelBody,
 	ToggleControl,
 	// eslint-disable-next-line @typescript-eslint/ban-ts-comment
@@ -164,9 +167,9 @@ const Edit = ( {
 					</ToggleGroupControl>
 				</PanelBody>
 			</InspectorControls>
-			<Disabled>
-				<Block { ...{ ...attributes, ...context } } />
-			</Disabled>
+			<Block { ...{ isEditor: true, ...attributes, ...context } }>
+				<InnerBlocks />
+			</Block>
 		</div>
 	);
 };
--- a/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/edit.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/edit.tsx
@@ -34,6 +34,7 @@ import { BLOCK_ICON as icon } from './constants';
 import { title, description } from './block.json';
 import { BlockAttributes, ImageSizing } from './types';
 import { ImageSizeSettings } from './image-size-settings';
+import { useIsDescendentOfSingleProductBlock } from '../shared/use-is-descendent-of-single-product-block';
 
 type SaleBadgeAlignProps = 'left' | 'center' | 'right';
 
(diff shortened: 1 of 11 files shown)


ObliviousHarmony:
Howdy @albarin and @gigitux!

Since it looks like you have some experience working with the Product Image block, could you take a look at this PR and let me know what you think?

My biggest concern at the moment is that this might introduce style incompatibilities because I moved the sale badge out of the scope of `.wc-block-components-product-image`. It's now nested under `.wc-block-components-product-image-inner-blocks`. Any themes targeting the badge might be affected depending on the styles and the selectors used. Something to consider as well is that the block now supports runtime style generation and that might cause conflicts as well.

I'm also concerned about the way that the Product Image block's inner content overflows. Currently you can place blocks inside that stretch outside of the bounds of the parent and starts to cover up other content. I'm not sure if this should be considered "normal" or not.

Lastly, do you think it might make more sense to add `layout` support to the Product Image block rather than relying on an inner Row block? This feels "more correct" but at the same time I am worried that we're adding too many runtime styles that might conflict with existing theme styles.

gigitux:
Thanks for the ping!

Nice work! The PR looks good. I tested the block, and everything works as expected. I still have some concerns about not reusing the Product Gallery Large Image block. To clarify, I don’t want to impose my perspective—I just want to ensure the team is aligned on the direction: could it make sense to include a designer in the conversation? 

> My biggest concern at the moment is that this might introduce style incompatibilities because I moved the sale badge out of the scope of `.wc-block-components-product-image`. It's now nested under `.wc-block-components-product-image-inner-blocks`. Any themes targeting the badge might be affected depending on the styles and the selectors used. Something to consider as well is that the block now supports runtime style generation and that might cause conflicts as well.

Does rendering the old HTML structure when `$parsed_attributes['showSaleBadge']` is present fix the issue?

> Lastly, do you think it might make more sense to add layout support to the Product Image block rather than relying on an inner Row block? This feels "more correct" but at the same time I am worried that we're adding too many runtime styles that might conflict with existing theme styles.

Not entirely sure about this—perhaps it’d be worth double-checking with a designer. For context, the Product Gallery block uses a group block:

https://github.com/woocommerce/woocommerce/blob/315fece578b40ae36b01e749d93881bca5ad2a45/plugins/woocommerce/client/blocks/assets/js/blocks/product-gallery/edit.tsx#L35-L52

plugins/woocommerce/src/Blocks/BlockTypes/ProductImage.php
```diff
@@ -235,20 +228,31 @@ protected function render( $attributes, $content, $block ) {
...
-				$wrapper_attributes,
-				$this->render_anchor(
-					$product,
-					$this->render_on_sale_badge( $product, $parsed_attributes ),
-					$this->render_image( $product, $parsed_attributes ),
-					$parsed_attributes
-				)
-			);
+		if ( ! $product instanceof \WC_Product ) {
+			return;
```
dinhtungdu, line 232: We should return an empty string here to be aligned the doc above.

plugins/woocommerce/src/Blocks/BlockTypes/ProductImage.php
```diff
@@ -235,20 +228,31 @@ protected function render( $attributes, $content, $block ) {
...
+				%2$s
+				<div class="wc-block-components-product-image-inner-blocks">
+					%3$s
+				</div>
+			</div>',
+			$wrapper_attributes,
+			$this->render_anchor(
+				$product,
+				$sale_badge,
+				$this->render_image( $product, $parsed_attributes ),
```
dinhtungdu, line 252: I think we follow the cover block here to manually style the image, we can set the `object-fit` property to `cover` and control the dimension using the wrapper element. We can also leverage the Supports API here to eliminate some of the manual style processing.

dinhtungdu: Talking about block support, we need move block supports declaration to `block.json` and remove `get_block_type_supports()` (which does nothing since the introduction of Product Image `block.json` file).

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/edit.tsx
dinhtungdu, line 1: With the inner blocks support, do we need a default template?

dinhtungdu: And do we need to sett allowed block?

dinhtungdu (commented):
Apart from my review comments, here is my opinion on your concerns [here](https://github.com/woocommerce/woocommerce/pull/56615#issuecomment-2759580694):

> My biggest concern at the moment is that this might introduce style incompatibilities because I moved the sale badge out of the scope of .wc-block-components-product-image. 

To me it's fine, as we still support the `showSaleBadge` attribute, if that attribute exists, the HMTL output remains unchanged. For newer structure, I expect that third parties need to update their codebase if necessary.

> I'm also concerned about the way that the Product Image block's inner content overflows. Currently you can place blocks inside that stretch outside of the bounds of the parent and starts to cover up other content. I'm not sure if this should be considered "normal" or not.

I'd look at how Cover block is doing first then mimic the approach.

> Lastly, do you think it might make more sense to add layout support to the Product Image block rather than relying on an inner Row block? This feels "more correct" but at the same time I am worried that we're adding too many runtime styles that might conflict with existing theme styles.

For me, yes. From my observation, I notice a frequent use case that merchant put a quick view button and a add to wish list button when hovering on the product image. Those button can be in the middle but some set them aside. Having a layout support give merchant more tool to achieve their design.

sunyatasattva:
Closing in favor of https://github.com/woocommerce/woocommerce/pull/57586

PR #56621, opened 2025-03-23
New Block: Product Specifications block to display weight, dimensions, and attributes

### LINKED ISSUE #55726: [Product Details Block] Product additional information block
**Description**

As part of the blockifying process of product details we want need to create a block to display the additional information of a product.

Current additional information tab PHP file: https://github.com/woocommerce/woocommerce/blob/8ee74ede0ebc1957089af08207ae2a1a8250610e/plugins/woocommerce/templates/single-product/tabs/additional-information.php#L1
Exposes a couple hooks:
- `woocommerce_product_additional_information_heading`
- `woocommerce_product_additional_information`

Currently the additional information tab displays a couple different things using the product attributes template ( [src](https://github.com/woocommerce/woocommerce/blob/5f43827f7f55a73658e2f564cf107ed79cc78b14/plugins/woocommerce/templates/single-product/product-attributes.php#L1) ):
- Weight ( if entered, and `wc_product_enable_dimensions_display` filter is true )
- Dimensions ( if entered, and `wc_product_enable_dimensions_display` filter is true )
- Attributes ( if entered and `wc_attributes_array_filter_visible` filter is true )

The above list can be extended using `woocommerce_display_product_attributes`

**Acceptance criteria**

- [x] Have an investigation done on how plugins may extend the additional information section, and create a backwards compatibility API for this ( if needed ).
- [ ] Create a "Product Additional Information" or a "Product Attributes" block ( @louwie17 leans towards the latter to keep it granular and more modular )
- [ ] Update the product details block default template to use the new additional information or attributes block

imanish003:

This PR introduces a new experimental Product ~Attributes~ Specifications block (`woocommerce/product-specifications`) that displays product details such as **weight**, **dimensions**, and **attributes**. This block provides merchants with a flexible way to showcase additional product information. This is supposed to replace existing Additional Information tab on the single product page.

The block includes several key features: It is context-aware, displaying placeholder text in the editor when no product context is available in case of Single Product template and showing a warning when placed in an invalid query loop. It uses server-side rendering with PHP render callback to ensure accurate and localized output on the frontend. Additionally, it is integrated with the Product Details block, being automatically included in the Additional Information section.

**Compatibility Layer**

To keep this PR focused and manageable, I plan to implement the compatibility layer in a **follow-up PR**. However, I welcome your thoughts on whether we should include hook support in the current PR instead. Since this would require additional discussion and consideration, I believe addressing it in a separate PR would help maintain clarity and focus.

Below are the hooks we're planning to support in the compatibility layer:

| **Hook Name**                                      | **Support?** | **Notes**                                                                        |
| -------------------------------------------------- | ------------ | -------------------------------------------------------------------------------- |
| woocommerce_product_additional_information         | ✅ Yes       | Core hook for adding content. Add to compatibility layer for frontend rendering. |
| woocommerce_display_product_attributes             | ✅ Yes       | Filter used to modify attribute data. Render filtered attributes on frontend.    |
| wc_product_enable_dimensions_display               | ✅ Yes       | Respect this filter when deciding to show dimensions.                            |
| woocommerce_product_additional_information_heading | ❌ No        | Too rarely used, can be replaced by block-based heading in editor.               |

Closes #55726
Closes WOOPLUG-3845

**Inside Single Product Block**

Image

**Inside Single Product Template**

Image

**Prerequisites:**

If you are testing on Production build, you will need to enable experimental blocks feature flag. (Not required for Dev build)

- Have the WooCommerce Beta Tester plugin activated
- Go to Tools > WCA Test Helper > Features and enable the experimental blocks feature flag

#### Scenario 1: Inside Single Product template

1. Go to "Single Product" template
2. Add "Product Specifications" block
3. Verify that it's showing placeholder text as shown in screenshot below: "This block displays product attributes including dimensions and weight. When viewing a product page, the attributes will automatically appear here."
   Image

4. Save and go to frontend of any product
5. Verify that it's showing attributes table as expected, for example with attributes like depending on current product attributes:
   - Color: Blue, Green, Red
   - Size: Large, Medium, Small
     Image

#### Scenario 2: Inside Single Product block

1. Create a new post or edit existing one
2. Add "Single Product" block
3. Add "Product Specifications" block inside "Single Product" block
4. Verify that it's showing product attributes as expected depending on selected product.
5. Save and go to frontend
6. Verify that it's showing product attributes as expected.

Here is screenshot of Editor side and Frontend side:
Image

#### Scenario 3: Product Specifications block is used in Product Collection block

1. Create a new post or edit existing one
2. Add "Product Collection" block
3. Add "Product Specifications" block inside "Product Collection" block
4. Verify that it's showing product attributes as expected if product has attributes.
5. Save and go to frontend
6. Verify that it's showing product attributes as expected.

#### Scenario 4: Should not be available in Block Inserter in Invalid context

1. Create a new post or edit existing one
2. Try to add "Product Specifications" block and verify that it's not available in Block Inserter

#### Scenario 5: Block is available in Product Details block

1. Go to Single Product template
2. Add "Blockified Product Details" block
3. Verify that Product Specifications block is available in Product Details block by default under "Additional Information" accordion.
   Image

#### Scenario 6: Testing controls

1. Verify that following inspector controls are working on Editor side as well as Frontend side:
   1. Dimensions
      1. Padding
      2. Margin
   2. Advanced
      1. Additional CSS class(es)
   3. Typography
      1. Size
      2. Line height

#### Scenario 7: Query Loop Block

1. Create a new post or edit existing one
2. Add "Query Loop" block
3. Add "Product Specifications" block inside "Query Loop" block
4. Verify that you can see following warning message: "The Product Specifications block requires a product context. When used in a Query Loop, the Query Loop must be configured to display products."
   Image

5. Set "Post type" to "Product" using Inspector control
6. Verify that Product Specifications block is displayed correctly
7. Publish the post and view the frontend. Verify that Product Specifications block is displayed correctly.

### Changelog entry

<details>

#### Significance

- [ ] Patch
- [ ] Minor
- [x] Major

#### Type

- [x] Add - Adds functionality

#### Message 
New Block: Product Specifications block to display weight, dimensions, and attributes
</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/blocks/product-attributes/style.scss
imanish003, line 1: For now, I have copied the CSS from the legacy styles, but I think we will need to introduce a few inspector controls to manage the style of this block. For example, users should be able to change the border width and border type, and they should also be able to set the alternative row color. Currently, we have it set to a grayish tone, but users should be able to change it using inspector controls. Let me know your thoughts on this, and I believe we can introduce these changes in the next iteration or follow-up PR.

imanish003: Related to this, I also considered using core table as an inner block. However, the table block allows users to change the table data. I'm not sure yet if we want to allow users to change attributes dynamically. We need to be careful here, as it could lead to users mistakenly updating attribute data. I just wanted to mention that this might be something we may consider in future.

plugins/woocommerce/client/blocks/assets/js/blocks/product-attributes/edit.tsx
```diff
@@ -0,0 +1,168 @@
...
+		};
+	}, [] );
+
+	const { results, isLoading } = useCollection< Product >( {
+		namespace: '/wc/v3',
+		resourceName: 'products',
+		resourceValues: [ Number( postId ) ],
+		shouldSelect: isSpecificProductContext,
+	} );
+	const product = results as unknown as Product;
```
imanish003: This is the best option that I found to fetch weight and dimensions of a product on editor site.

plugins/woocommerce/src/Blocks/BlockTypes/ProductAttributes.php
imanish003, line 1: Here, I attempt to replicate the [wc_display_product_attributes](https://github.com/woocommerce/woocommerce/blob/f0e12ba75d48b5858fe03fad3342e88553fe6dc5/plugins/woocommerce/includes/wc-template-functions.php#L3814) function that was used for the additional information tab. Also, This will enable us to easily incorporate previous hooks for the compatibility layer.

plugins/woocommerce/client/blocks/assets/js/blocks/product-attributes/edit.tsx
louwie17: You should be able to use the products data store for this, as it hits the same endpoint as above.
You can do this instead:
```
import { productsStore } from "woocommerce/data";
const { product } = useSelect( ( select ) => {
  const { getProduct } = select( productsStore );
  return {
    product: getProduct( postId )
  }
});
```

plugins/woocommerce/client/blocks/assets/js/blocks/product-attributes/edit.tsx
```diff
@@ -0,0 +1,168 @@
...
+		if ( validDimensions.length === 0 ) return null;
+
+		return `${ validDimensions.join( ' × ' ) } ${ dimensionUnit }`;
+	};
+
+	const formattedDimensions = product?.dimensions
+		? getFormattedDimensions( product.dimensions )
+		: null;
+
+	return isSpecificProductContext ? (
```
louwie17: Currently we are rendering this whether the product has attributes or not, this can cause this weird empty state when a product doesn't have dimensions or attributes:
Screenshot 2025-03-24 at 14 53 32

plugins/woocommerce/client/blocks/assets/js/blocks/product-attributes/block.json
```diff
@@ -0,0 +1,35 @@
...
+	"ancestor": [
+		"woocommerce/single-product",
+		"woocommerce/product-template",
+		"core/post-template"
+	],
+	"attributes": {
+		"fontSize": {
+			"type": "string",
+			"default": "medium"
+		}
```
louwie17: We should add an attribute or multiple attributes to configure the display of dimensions and weight.
In my mind a user should be able to display either weight, dimensions, or attributes by themselves.
Display them all together if they exist, or mix-match.

For example I could easily see a scenario where a user wants to display the dimensions separately from the attributes, they may actually create an accordion just for dimensions and weight.
Displaying them always together doesn't allow for this.
Making it configurable, allows users to re-use this block in multiple places, but display either attributes in one place and dimensions and weight in another.

plugins/woocommerce/client/blocks/assets/js/blocks/product-attributes/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+import icon from './icon';
+import './style.scss';
+
+const blockConfig = {
+	...metadata,
+	icon,
+	edit,
+};
+
+if ( isExperimentalBlocksEnabled() ) {
```
louwie17: Where you thinking of removing this as part of this PR, or as a follow up? ( I am good with either ).

plugins/woocommerce/src/Blocks/BlockTypes/ProductAttributes.php
```diff
@@ -0,0 +1,115 @@
...
+					$value = make_clickable( esc_html( $value ) );
+				}
+			}
+
+			$product_attributes[ 'attribute_' . sanitize_title_with_dashes( $attribute->get_name() ) ] = array(
+				'label' => wc_attribute_label( $attribute->get_name() ),
+				'value' => wpautop( wptexturize( implode( ', ', $values ) ) ),
+			);
+		}
+
```
louwie17: Lets add a condition here to return an empty string when a product doesn't have any attributes or dimensions. Otherwise we still render an empty table.
The previous implementation also rendered nothing: https://github.com/woocommerce/woocommerce/blob/5f43827f7f55a73658e2f564cf107ed79cc78b14/plugins/woocommerce/templates/single-product/product-attributes.php#L22-L24

plugins/woocommerce/client/blocks/assets/js/blocks/product-attributes/edit.tsx
```diff
@@ -0,0 +1,168 @@
...
+	const blockProps = useBlockProps();
+	return (
+		<div { ...blockProps }>
+			<p>
+				{ __(
+					'This block displays product attributes including dimensions and weight. When viewing a product page, the attributes will automatically appear here.',
+					'woocommerce'
+				) }
+			</p>
+		</div>
```
louwie17: Instead of showing this as a place holder I would like to see an actual table with placeholder text depending on the block settings.

louwie17: See the order placeholder tables for examples:
Screenshot 2025-03-24 at 15 13 20

louwie17: The only downside to the above example is that they do use an hardcoded table.
I would actually like us to make use of the core table block if at all possible.

louwie17 (changes requested):
Thanks for working on this @imanish003, it tested well for the most part. I did just notice an issue when a product doesn't have any attributes or dimensions, which we currently don't handle very well within the editor it-self.

I also have a couple suggestions, which in hindsight I should of made more clear within the issue it-self. Given this block is currently under the experimental flag, I did be ok addressing these suggestions as follow ups and just address the empty state and some of the smaller items that I mentioned inline as part of this PR.
My suggestions:
- We should add an attribute setting to allow to user to toggle the display of either only attributes, or only dimensions, or only weight, or combine any of them ( see the terms list screenshot as an example below, it provides several toggle options )
- I would like us to inherit styles and settings from the core table block for this component, so that a user can change the styling of the table or add headers/footers ( see table demo below )
     - The one thing I am not entirely sure on, is if we are best to just handle these options manually within our block or use the block as a nested child or create a variation thereof. 

Terms list example:
Screenshot 2025-03-24 at 15 34 36

Table demo:
https://github.com/user-attachments/assets/86fe3076-60a8-4173-ac7c-f2e314332d3c

plugins/woocommerce/client/blocks/assets/js/blocks/product-attributes/block.json
dinhtungdu: I opened an issue for this: [WOOPLUG-3844: Add conditional options for displaying weight, dimensions, and attributes](https://linear.app/a8c/issue/WOOPLUG-3844/add-conditional-options-for-displaying-weight-dimensions-and)

plugins/woocommerce/client/blocks/assets/js/blocks/product-attributes/edit.tsx
dinhtungdu: Let's use hard code table for now then try to replace it with the Table block later. Edit: Addressed in https://github.com/woocommerce/woocommerce/pull/56621/commits/bd96b64d633ec8e2a7f2233ae6dc765213dd0a06

plugins/woocommerce/client/blocks/assets/js/blocks/product-attributes/edit.tsx
dinhtungdu: Addressed in https://github.com/woocommerce/woocommerce/pull/56621/commits/20a4e2636cd68817ee825e058688dfbe42bd4c3f

plugins/woocommerce/client/blocks/assets/js/blocks/product-attributes/index.tsx
dinhtungdu: Let's do it in this PR.

dinhtungdu:
Changes in recent commits:
- Renaming block to `Product Specifications`.
- Rendering the place holder data in the Site Editor.
- Using product store to retrieve product data.
- Removing the redundant experimental block check.
- Reduce specificity of block style.
- Use table as the block container element to match the markup between frontend and editor.

@louwie17 as you reviewed this PR, I requested a review from you, please let me know if you don't have capacity for that.

gigitux (approved):
Thanks for working on this, @imanish003, and thanks to @dinhtungdu for wrapping it up!

The PR looks good to me. I believe we’ll continue iterating on this block in future PRs.

One note: regarding the use of the `<table>` element — have we decided to stick with it? With this new block, we might have an opportunity to modernize the markup structure and make it more SEO-friendly.

dinhtungdu:
> regarding the use of the `<table>` element — have we decided to stick with it? With this new block, we might have an opportunity to modernize the markup structure and make it more SEO-friendly.

AFAIK, no. I will think more about it and also share this question with our designers for UX direction.

louwie17 (approved):
Thanks for addressing my previous suggestions @dinhtungdu 👍 
I didn't actually test this though, but just looked over the new changes. Mostly as I see @gigitux already did and also because this block is experimental. 

I will note that looking at the changes this product empty state may still happen: https://github.com/woocommerce/woocommerce/pull/56621#discussion_r2010231758 ? 
I am happy if that is a follow up, but just wanted to call it out as it may be an edge case ( using this block in a context aware state with a product that doesn't contain any attributes or dimensions ).

dinhtungdu:
> I will note that looking at the changes this product empty state may still happen: [https://github.com/woocommerce/woocommerce/pull/56621#discussion_r2010231758](https://github.com/woocommerce/woocommerce/pull/56621#discussion_r2010231758) ?  I am happy if that is a follow up, but just wanted to call it out as it may be an edge case ( using this block in a context aware state with a product that doesn't contain any attributes or dimensions ).

@louwie17 I tagged you in the comment thread on this topic: [WOOPLUG-3223](https://linear.app/a8c/issue/WOOPLUG-3223/[product-details-block]-product-specifications-block#comment-689484f9).

PR #56646, opened 2025-03-24
Fix `get_last_modified` method in ProductQuery to return 0 when query returns null

gigitux:

This PR fixes a potential crash in the `get_last_modified()` method when the database has no products. Previously, the method would attempt to pass NULL to `strtotime()` when no products existed, which could cause PHP errors.

Closes # .

| Before | After |
| ------ | ----- |
| image |   image|

1. Ensure that your store has 0 products in the database (using a new site would be great).
2. Open the browser network tool.
3. Visit the Single Product template.
4. Ensure that `wp-json/wc/store/v1/products?include=0&_locale=user` doesn't return 500 error.
5. Ensure that `Product SKU` and `Product Rating` doesn't crash 

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Fix get_last_modified method in ProductQuery to return 0 when query returns null.

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/56646-fix-avoid-crash-store-api-when-store-0-product
+++ b/plugins/woocommerce/changelog/56646-fix-avoid-crash-store-api-when-store-0-product
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix get_last_modified method in ProductQuery to return 0 when query returns null.
--- a/plugins/woocommerce/src/StoreApi/Routes/V1/Products.php
+++ b/plugins/woocommerce/src/StoreApi/Routes/V1/Products.php
@@ -87,8 +87,11 @@ class Products extends AbstractRoute {
 			$query_results = $product_query->get_results( $request );
 		}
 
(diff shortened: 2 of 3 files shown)


louwie17 (approved):
Thanks for the quick and easy fix on this @gigitux, this tested well and code looks good!
Having said that it looks like we only use this function here: https://github.com/woocommerce/woocommerce/blob/bbd5b7ab8ebf42b32a7c048c92f7d9b0553f4df3/plugins/woocommerce/src/StoreApi/Routes/V1/Products.php#L91
when setting the `Last-Modified` header. But when looking at the request I don't see the `Last-Modified` header at all 🤷‍♂️ 
Also it doesn't seem like passing a integer to the `Last-Modified` header is valid: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Last-Modified

gigitux:
Thanks for noticing this! I think that you're right and I updated the logic. @louwie17, can you take another look before the merge?

louwie17 (approved):
Thanks for updating the logic that change looks good! This still tested well, thanks!

PR #56647, opened 2025-03-24
Confirm before activating the plugin after connecting a subscription to WooCommerce.com

thilinah:

When connecting a subscription via in-app subscriptions list, the plugin is activated without a confirmation from the user.

This can lead to activation of a plugin which can break the store. 

This PR introduce a confirmation dialog for the user to decide whether the related plugin should be activated after connecting the subscriptions to WooCommerce.com. 

Closes # .

| Before | After |
| ------ | ----- |
|    Screenshot 2025-03-25 at 19 09 40 |   Screenshot 2025-03-25 at 14 56 14 |

1.  Create the build from this PR and install it on customer store.
2. Ensure the store is connected to WooCommerce.com.
3. Make sure you have a WooCommerce.com extension installed but not connected. (The plugin should not be activated).
4. Click connect and notice the confirmation dialog to activate the plugin.
5. The plugin should be activated only if the user confirmed the action.
6. When connecting a subscription the confirmation dialog should not be shown in following cases:
    - When the plugin is already active on site.
    - When connecting a theme.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

Show a confirmation dialog before activating the plugin when connecting a subscription to WooCommerce.com

</details>

<details>

#### Comment 

</details>


nirolph (commented):
Thanks for tackling this @thilinah. 
Asking the user for confirmation before activating a deactivated extension makes sense.

However, the dialog only shows up for about a second for me before closing and performing the connection. It doesn't wait for my input and it looks like it connects the license (confirmed by the new entry in `wp_woocommerce_helper_products`) without activating the plugin.

Is there something I'm missing?

https://github.com/user-attachments/assets/a5fba900-8492-4f07-9b68-c360cf1e780a

thilinah:
> However, the dialog only shows up for about a second for me before closing and performing the connection. It doesn't wait for my input and it looks like it connects the license (confirmed by the new entry in `wp_woocommerce_helper_products`) without activating the plugin.

Fixed with https://github.com/woocommerce/woocommerce/pull/56647/commits/ca3e93020b9953d28ee7b144dc9bc6435cfcd281

KokkieH (commented):
It works as expected regarding the plugin being activated or not, but I notice some UI issues:

If I click "No" in the modal, the connect button remains stuck in a busy state.

If I refresh, the subscription status is no longer "not connected", the button shows "Renew", and on WCCOM the product shows as connected. So the connection still completed without activating the plugin (I assume that's intended), but the state of the Connect button didn't update automatically like it does when clicking Yes in the modal.

https://github.com/user-attachments/assets/13c15bb7-d5d6-4b2e-87eb-77d11f8383a4

### A11Y
The Yes/No buttons don't have a visual focus state if navigating via keyboard. It does when VoiceOver is active, but that's because VO adds its own focus state on top of anything defined in the website styles, and not all keyboard nav users are also using screen readers at the same time.

https://github.com/user-attachments/assets/006f11bb-43b4-45ee-9875-cf62290786cc

I also notice the "Not connected" chip is coded as a button and receive focus so the tooltip will appear to keyboard users, but when using VoiceOver, only the button text is being read out, and the tooltip content isn't read out at all.

That's out of scope for this PR, but it's defnitely something we need to improve.

https://github.com/user-attachments/assets/6ce4b5e5-08f7-4380-ab1b-9e7a87101921

thilinah:
Thanks for the review @KokkieH.

> If I click "No" in the modal, the connect button remains stuck in a busy state.

This is fixed with: https://github.com/woocommerce/woocommerce/pull/56647/commits/b187052c9c26a2d7a8ba0687e7f3a58a88d7b650. Also made some improvements to how loading state is shown.

> The Yes/No buttons don't have a visual focus state if navigating via keyboard. It does when VoiceOver is active, but that's because VO adds its own focus state on top of anything defined in the website styles, and not all keyboard nav users are also using screen readers at the same time.

I think this is because how the primary button focus state is shown on all the medals in in-app screens. I've added a class that makes "No" button behave the same way it does in other dialogs. But the primary button need to be fixed on all the dialogs and it needs to done on another PR.

Now the activate dialog looks same as any other modals:
Screenshot 2025-04-11 at 10 47 40

Screenshot 2025-04-11 at 11 10 12

> If I refresh, the subscription status is no longer "not connected", the button shows "Renew"

This is the default behaviour. If your subscription connected but cancelled (auto-renew disabled), it'll show a Renew button. But when the subscription is `cancelled` but not connected, `Connect` button takes precedence over `Renew`

KokkieH:
> This is the default behaviour. If your subscription connected but cancelled (auto-renew disabled), it'll show a Renew button. But when the subscription is cancelled but not connected, Connect button takes precedence over Renew

Yep, that part I got. I mostly mentioned it to indicate the connection succeeded, even though the plugin wasn't activated :)

KokkieH (approved):
Thanks for the changes.

All the test cases work as described.

PR #56652, opened 2025-03-24
Prevent count cache errors when order hooks are run in different orders

### LINKED ISSUE #56650: Order count cache is incorrect when a status is immediately transitioned
### Prerequisites

### Describe the bug

When creating an order that immediately sets a status to one of the "non-default" statuses (checkout draft, pending, etc), the order count cache increases the order count for both statuses since both the `woocommerce_new_order` and `woocommerce_order_status_changed` hooks are fired.

This issue is compounded by the fact that the hooks are run in different orders in the frontend and backend.

Image

Introduced in https://github.com/woocommerce/woocommerce/pull/54034

### Expected behavior

Expected a new order to only increase the cache count by 1 in the respective status.

### Actual behavior

Cache count incorrectly increments by 2 orders if order status is different.

### Steps to reproduce

1. Go to the admin orders screen
2. Make note of the order counts
3. Click "Add order"
4. Set the order status to complete
5. Click "Create"
6. Note that the order count has incorrectly increased by 2

### WordPress Environment

n/a

### Isolating the problem

joshuatf:

This PR fixes the order count becoming out of sync as order hooks fire in different orders and potentially multiple times.

Closes #56650 .

(For Bug Fixes) Bug introduced in PR #54034.

| Before | After |
| ------ | ----- |
|   Screenshot 2025-03-24 at 3 22 34 PM  |   Screenshot 2025-03-24 at 3 22 25 PM |

For all tests, install an object persistent cache plugin like [Docket Cache](https://wordpress.org/plugins/docket-cache/).

**Admin orders without status changes**
1. Go to WooCommerce -> Orders
2. Take note of the order counts
3. Click "Add order"
4. Keep the order status as "Pending"
5. Click "Create"
6. Make sure that "Pending" has increased by 1 and no other order counts have

**Admin orders with status changes**
1. Go to WooCommerce -> Orders
2. Take note of the order counts
3. Click "Add order"
4. Set the status to "Complete" or any non-pending status
5. Click "Create"
6. Make sure that "Complete" has increased by 1 and no other order counts have increased or decreased

**Customer placed order**
1. Add various payment gateways from as a merchant that have different initial order statuses
2. Take note of the order counts
3. As a customer, add an item to the cart
4. Checkout as a customer
5. In the admin, make sure order counts are reflected correctly

**Programmatically placed order (dev only)**
1. Make sure tests pass
6. Attempt to create orders without status changes and persist
7. Make sure that order counts are correctly reflected
8. Attempt to create an order and set a status one or more times before persisting
9. Make sure that order counts are correctly reflected

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/src/Caches/OrderCountCacheService.php
+++ b/plugins/woocommerce/src/Caches/OrderCountCacheService.php
@@ -10,6 +10,8 @@ use Automattic\WooCommerce\Utilities\OrderUtil;
 
 /**
  * A service class to help with updates to the aggregate orders cache.
+ *
+ * @internal
  */
 class OrderCountCacheService {
 
@@ -19,6 +21,20 @@ class OrderCountCacheService {
 	 * @var OrderCountCache
 	 */
 	private $order_count_cache;
+
+	/**
+	 * Array of order ids with their last transitioned status as key value pairs.
+	 *
+	 * @var array
+	 */
+	private $order_statuses = array();
+
+	/**
+	 * Array of order ids with their initial status as key value pairs.
+	 *
+	 * @var array
+	 */
+	private $initial_order_statuses = array();
 
 	/**
 	 * Class initialization, invoked by the DI container.
@@ -44,6 +60,18 @@ class OrderCountCacheService {
 			return;
 		}
 
+		// If the order status was updated, we need to increment the order count cache for the
+		// initial status that was errantly decremented on order status change.
+		if ( isset( $this->initial_order_statuses[ $order_id ] ) ) {
+			$this->order_count_cache->increment( $order->get_type(), $this->get_prefixed_status( $this->initial_order_statuses[ $order_id ] ) );
+		}
+
+		// If the order status count has already been incremented, we can skip incrementing it again.
+		if ( isset( $this->order_statuses[ $order->get_id() ] ) && $this->order_statuses[ $order->get_id() ] === $order->get_status() ) {
+			return;
+		}
+
+		$this->order_statuses[ $order_id ] = $order->get_status();
 		$this->order_count_cache->increment( $order->get_type(), $this->get_prefixed_status( $order->get_status() ) );
 	}
 
@@ -94,6 +122,15 @@ class OrderCountCacheService {
 			return;
 		}
 
+		if ( isset( $this->order_statuses[ $order_id ] ) && $this->order_statuses[ $order_id ] === $next_status ) {
+			return;
+		}
+
+		if ( ! isset( $this->initial_order_statuses[ $order_id ] ) ) {
+			$this->initial_order_statuses[ $order_id ] = $previous_status;
+		}
+
+		$this->order_statuses[ $order_id ] = $next_status;
 		$this->order_count_cache->decrement( $order->get_type(), $this->get_prefixed_status( $previous_status ) );
 		$this->order_count_cache->increment( $order->get_type(), $this->get_prefixed_status( $next_status ) );
 	}
--- a/plugins/woocommerce/src/Caches/OrderCountCacheService.php
+++ b/plugins/woocommerce/src/Caches/OrderCountCacheService.php
@@ -122,10 +122,12 @@ class OrderCountCacheService {
 			return;
 		}
 
+		// If the order status count has already been incremented, we can skip incrementing it again.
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/src/Caches/OrderCountCacheService.php
```diff
@@ -44,6 +60,18 @@ public function update_on_new_order( $order_id, $order ) {
 			return;
 		}
 
+		// If the order status was updated, we need to increment the order count cache for the
+		// initial status that was errantly decremented on order status change.
+		if ( isset( $this->initial_order_statuses[ $order_id ] ) ) {
```
joshuatf, line 65: Because `woocommerce_new_order` can fire before or after `woocommerce_order_status_changed` and `woocommerce_order_status_changed` can fire multiple times, we want to make sure we didn't accidentally decrement a status count that had never been persisted.

joshuatf (commented):
This is a bit hacky as the hooks have somewhat unexpected behavior.  The alternative to this is to directly integrate `OrderCountCache` into `WC_Order`.  Given there is existing caching in that class, this is not unprecedented.

plugins/woocommerce/src/Caches/OrderCountCacheService.php
```diff
@@ -10,6 +10,8 @@
 
 /**
  * A service class to help with updates to the aggregate orders cache.
+ *
+ * @internal
  */
 class OrderCountCacheService {
```
kalessil, line 16: Since we are marking it internal, how about making it final?

plugins/woocommerce/src/Caches/OrderCountCacheService.php
```diff
@@ -44,6 +60,18 @@ public function update_on_new_order( $order_id, $order ) {
...
 		}
 
+		// If the order status was updated, we need to increment the order count cache for the
+		// initial status that was errantly decremented on order status change.
+		if ( isset( $this->initial_order_statuses[ $order_id ] ) ) {
+			$this->order_count_cache->increment( $order->get_type(), $this->get_prefixed_status( $this->initial_order_statuses[ $order_id ] ) );
+		}
+
+		// If the order status count has already been incremented, we can skip incrementing it again.
+		if ( isset( $this->order_statuses[ $order->get_id() ] ) && $this->order_statuses[ $order->get_id() ] === $order->get_status() ) {
```
kalessil, line 70: Guess we could use `$order_id` over `$order->get_id()` for consistency?

kalessil: Also there is potential to extract repeating `$this->get_prefixed_status( ... )` in `update_on_order_status_changed` and `update_on_new_order` into local variables to simplify the codebase a bit.

kalessil (commented):
Did a quick review and leaving some comments - will do manual testing tomorrow.

kalessil:
Manual testing results:

- create an order as 'completed' from the admin area: :lady_beetle: - order counted twice
Screenshot from 2025-03-26 09-44-16

- change the order status to cancelled: :heavy_check_mark: 
Screenshot from 2025-03-26 09-47-19

- place an order from storefront: :heavy_check_mark: 
Screenshot from 2025-03-26 09-51-02

joshuatf:
Thanks for testing, @kalessil!  Was able to reproduce and found [this relevant tracs issue](https://core.trac.wordpress.org/ticket/62236) in WP core.

It seems we can't use negative numbers (surprising since there's no mention of this in [the API docs](https://developer.wordpress.org/reference/functions/wp_cache_decr/)), so the initial decrement in status change was not successful, which may explain why we also didn't initially find issues in https://github.com/woocommerce/woocommerce/pull/54034.

Pushed a change in https://github.com/woocommerce/woocommerce/pull/56652/commits/21d33d3db77da834fb6322a51ebfadaa98fc0c6b that checks for decrement success before re-incrementing the number in new orders.

https://github.com/user-attachments/assets/b6d630f1-d320-47e3-b33b-dbee6119063b

kalessil (approved):
Works like a charm, preapproving!

PR #56654, opened 2025-03-24
Update override-loop-template-show-quantities-next-add-cart.md

rossanafmenezes:
We are using the wrong apostrophe throughout several docs which is rendering incorrectly in the front end

CleanShot 2025-03-24 at 15 07 38@2x

Just changed the text to  `theme's`


nerrad:
Looks like this got addressed separately somewhere else. The docs currently show the expected smart quote.
CleanShot 2026-08-12 at 11 55 51@2x

PR #56460, opened 2025-03-17
Re-disable "hide shipping costs" option and ensure shipping is selected by default in Checkout block.

opr:
### Changes proposed in this Pull Request:

Partially revert https://github.com/woocommerce/woocommerce/pull/53466 due to bugs found in testing

Specifically the "Hide shipping costs until an address is entered" is disabled again if local pickup is on, and the overrides to force the option to `false` are back in place.

This PR also updates E2E tests to account for the disabled checkbox, and tests that a _shipping_ method is selected by default when first visiting the checkout page.

(For Bug Fixes) Bug introduced in PR #53466 .

### Screenshots or screen recordings:

| Before | After |
| ------ | ----- |
| image | image |
| https://www.loom.com/share/6e0455ac8e5d40f1a1bf33723eec1648 | https://www.loom.com/share/05227fd3b6eb48709f3250e2c94dadb3 |

### How to test the changes in this Pull Request:

Using the [WooCommerce Testing Instructions Guide](https://github.com/woocommerce/woocommerce/wiki/Writing-high-quality-testing-instructions), include your detailed testing instructions:

1. Go to WooCommerce -> Settings -> Shipping -> Local Pickup. Disable local pickup.
2. Go to WooCommerce -> Settings -> Shipping -> Shipping Settings. Observe that the "Hide shipping costs until an address is entered" setting can be checked/unchecked and is not disabled.
3. Go to WooCommerce -> Settings -> Shipping -> Local Pickup. Enable local pickup and set up some methods.
4. Go to WooCommerce -> Settings -> Shipping -> Shipping Settings. Observe that the "Hide shipping costs until an address is entered" setting is disabled.
5. Go to WooCommerce -> Settings -> Shipping and add some shipping methods to a zone that covers your store's location.
6. Go to WooCommerce -> Settings -> General -> Default customer location. Set this to "Shop/country region".
7. In an incognito window, add an item to your cart and go to the cart block.
8. In the sidebar, ensure you see the shipping options and the local pickup options. Ensure the shipping option is selected by default.
9. Go to the Checkout block. Ensure the shipping option is selected by default.
10. Ensure you can check out and that the order confirmation screen shows the correct shipping method.
11. Check the order in the dashboard and ensure the correct shipping method was recorded.
12. Repeat step 7, 8, and 9.
13. Change the selected shipping method to local pickup. Ensure the sidebar updates with the updated information.
14. Ensure you can check out and that the order confirmation screen shows the correct local pickup method.
[...]

PR #56663, opened 2025-03-25
Remove calls to disable third party shipping plugin in shipping E2E tests

opr:

https://github.com/woocommerce/woocommerce/pull/56460 added many E2E tests, some of which disabled the third party "Woo Collection" shipping method. This PR disables the test helper plugin as a whole so this is no longer necessary. The plugin is not needed for these tests. If another test needs the helper plugin it can be enabled, like so: https://github.com/woocommerce/woocommerce/blob/6a7d3f4d6d7570978dd5380bc3f05fc12c1912de/plugins/woocommerce/client/blocks/tests/e2e/tests/checkout/checkout-block.merchant.block_theme.spec.ts#L104-L106

Note, I created a follow up issue as this test plugin may not be needed. https://github.com/woocommerce/woocommerce/issues/56664

(For Bug Fixes) Bug introduced in PR https://github.com/woocommerce/woocommerce/pull/56460

1. Check E2E tests in CI pass.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/blocks/tests/e2e/tests/cart/cart-checkout-block-shipping.block_theme.spec.ts
+++ b/plugins/woocommerce/client/blocks/tests/e2e/tests/cart/cart-checkout-block-shipping.block_theme.spec.ts
@@ -45,8 +45,7 @@ test.describe( 'Shopper → Shipping', () => {
 		await shippingUtils.enableShippingCostsRequireAddress();
 	} );
 
-	test.beforeEach( async ( { admin, requestUtils } ) => {
-		await requestUtils.deactivatePlugin( 'woocommerce-blocks-test-helper' );
+	test.beforeEach( async ( { admin } ) => {
 		await admin.visitAdminPage( 'admin.php?page=wc-settings&tab=general' );
 		await admin.page
 			.getByLabel( 'Default customer location' )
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/client/blocks/tests/e2e/tests/cart/cart-checkout-block-shipping.block_theme.spec.ts
```diff
@@ -45,7 +45,8 @@ test.describe( 'Shopper → Shipping', () => {
 		await shippingUtils.enableShippingCostsRequireAddress();
 	} );
 
-	test.beforeEach( async ( { admin } ) => {
+	test.beforeEach( async ( { admin, requestUtils } ) => {
+		await requestUtils.deactivatePlugin( 'woocommerce-blocks-test-helper' );
```
WunderBart: This should not be necessary - the plugin is initially disabled, no?

WunderBart (approved):
Left just one comment [there](https://github.com/woocommerce/woocommerce/pull/56663/files#r2012303837).

Apart from that, it looks good to me! 🚀

plugins/woocommerce/client/blocks/tests/e2e/tests/cart/cart-checkout-block-shipping.block_theme.spec.ts
opr: Yes, _shouldn't_ be necessary, but when testing locally in the [original PR](https://github.com/woocommerce/woocommerce/pull/56460) it _was_ enabled which is why each test disabled it. This is possibly because of bad state from previous test runs?

If we are confident that the plugin is disabled in CI I would be OK with removing it.

PR #56668, opened 2025-03-25
Update some block dependencies to allow DataViews to work

louwie17:

This updates two dependencies with the blocks package to prevent the build process from using an older version of `@floating-ui/react-dom`. As this broke the use of dataviews in this PR: https://github.com/woocommerce/woocommerce/pull/56641

1. Make sure all the GH actions are green
2. Create a new post and change the editor to **Code Editor**
3. Paste `<!-- wp:woocommerce/handpicked-products {"editMode":false,"products":[]} /-->`
4. Select a product and change some of the settings, the block should work correctly
5. Now also remove the product afterwards, it should display a list of sample products and work correctly

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/update-blocks_dependencies
+++ b/plugins/woocommerce/changelog/update-blocks_dependencies
@@ -1,0 +1,4 @@
+Significance: minor
+Type: update
+
+Update @wordpress/components and @wordpress/server-side-render dependencies in block package.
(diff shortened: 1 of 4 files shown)


roykho (approved):
Great work! For my curiosity, we're using `wp-6.6` is because of `L-1`?

When specific product is selected:
CleanShot 2025-03-25 at 12 17 19

When product is removed:
CleanShot 2025-03-25 at 12 17 05

louwie17:
> Great work! For my curiosity, we're using `wp-6.6` is because of `L-1`?

Good question, I believe `L-1` is the last two WP versions, so it is technically suppose to be 6.5, until 6.8 gets released. I updated it to 6.6 as this is what we have currently set in syncpack as the default for all the WP dependencies: https://github.com/woocommerce/woocommerce/blob/df495f62037815b8d7b8ea3ae6687058fbaaf877/.syncpackrc#L311

Most of the WP dependencies though are only used for development tasks, unit tests and types, given it uses the `window.wp` for most of them, which is the same case for the two I updated above.

PR #56680, opened 2025-03-25
[e2e-tests] Fix skipped C&C tests

### LINKED ISSUE #56522: [e2e-tests] Investigate `fixme` Cart & Checkout tests
## What

Let's investigate the 4 tests that we had to skip because they started permafailing on `trunk`: https://github.com/woocommerce/woocommerce/pull/56517

WunderBart:
## What?

Closes: #56522

## How to test

CI should be green.

### DIFF
--- a/plugins/woocommerce/changelog/e2e-fix-skipped-cart-and-checkout-tests
+++ b/plugins/woocommerce/changelog/e2e-fix-skipped-cart-and-checkout-tests
@@ -1,0 +1,4 @@
+Significance: patch
+Type: dev
+
+Unskip and fix tests skipped in https://github.com/woocommerce/woocommerce/pull/56517
--- a/plugins/woocommerce/client/blocks/tests/e2e/tests/checkout/checkout-block.shopper.block_theme.spec.ts
+++ b/plugins/woocommerce/client/blocks/tests/e2e/tests/checkout/checkout-block.shopper.block_theme.spec.ts
@@ -142,26 +142,27 @@ test.describe( 'Shopper → Local pickup', () => {
 		} );
 	} );
 
-	test.fixme(
-		'The shopper can choose a local pickup option',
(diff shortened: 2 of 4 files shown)


WunderBart:
The failing tests are the ones that are currently failing on trunk as well (fixed in https://github.com/woocommerce/woocommerce/pull/56663). The tests unskipped in this PR are all passing now.

alexflorisca (approved):
Thank you for fixing these and unskipping. Noticed you made some optimisations too by using `this.page` instead of passing the page object around 👍  Changes are quite hard to follow because of spacing I think, whole blocks seem to be deleted then added again but I didn't spot any other major changes? Correct me if I'm wrong...

Asked Cursor for any improvements here and suggested refactoring the pickup location selectors into its own function that can be re-used. Not a blocker, if you feel inclined to do so.

```js
async selectAndVerifyPickupLocation(locationName: string) {
    await this.page.getByRole('radio', { name: 'Pickup' }).click();
    await expect(this.page.getByLabel(locationName).last()).toBeVisible();
    await this.page.getByLabel(locationName).last().check();
}

// Then in the test:
await checkoutPageObject.selectAndVerifyPickupLocation('Testing');
```

Nothing blocking here so approving.

WunderBart:
> Changes are quite hard to follow because of spacing I think, whole blocks seem to be deleted then added again but I didn't spot any other major changes? Correct me if I'm wrong...

Yeah, the changes are hard to follow because of the reindentation, but the good news is that there was only one thing failing the tests and it was [this line](https://github.com/woocommerce/woocommerce/pull/56680/files#diff-a2f3066dd0f5dfb8fdd5d2f7452e3aa5f8506a32877205a50af79ac87540adccR560). The issue was the strict mode violation for the email, where there was more than one element visible on the page.

PR #56687, opened 2025-03-25
Product Gallery: use server side rendering.

cpapazoglou:

- Changed how `state` is created, the changes are required to benefit from server side rendering which in turn fixes the layout shift.
- Also, simplified `get_product_gallery_image_data` function to return just an array with all images

Related to: p1742369862965689-slack-C02FL3X7KR6

| Before | After |
| ------ | ----- |
|image|CleanShot 2025-03-19 at 20 38 37|

1. Add the Product Gallery ( Beta ) block in a single product template
2. Visit a product with more than 4 images ( `hoodie` )
3. Interact with the gallery and test the following:
 - zoom on hover
 - next / previous buttons
 - thumbnnail click selects the related image
 - large image click opens the popup with the related image in focus
 - thumbnail scrolling

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 
Product Gallery ( Beta ): Improves layout shift by rendering the large image from the server and then hydrating the gallery.
</details>

<details>

#### Comment 

</details>


cpapazoglou:
> As noted in our DMs the previous performance implementation which was replaced by loading=lazy attributes to conditionally load in only necessary imagery only works on Chrome, and not Firefox or Safari.

@tjcafferkey with all the thumbnails now loaded by default, this PR is improving the UX in Chrome at least. How do you feel proceeding with this and iterating on thumbnails sizes and lazy-loading large images?

tjcafferkey:
I think we need to reconsider this approach. Having given it some thought I don't think state is the correct way forward here because a user can have multiple gallery blocks on a single page. Here is a summary of the differences between state and context in the WP iAPI which Grok has generated for me, sums it up quite nicely:

>In the WordPress Interactivity API, state is global data managed within a store, shared across all instances of a block type or even between different blocks, making it ideal for synchronized or persistent values. Context, on the other hand, is local data scoped to a specific block instance, defined via data-wp-context in the HTML and accessed with getContext(), perfect for instance-specific information that shouldn’t affect other blocks. Essentially, use state for shared, cross-block data and context for isolated, per-instance data.

As a result, if we add multiple Single Product blocks to a page or post, and add the new gallery into them then this happens:

Screenshot 2025-03-26 at 12-11-15 Single Product – Woo

kmanijak:
As per @tjcafferkey comment this is not the way to go. I'm closing this PR for now and will be seeking some alternative.

PR #56697, opened 2025-03-26
Reset analytics settings at setup in `analytics-data.spec.js`

rodelgc:

This PR makes `analytics-data.spec.js`, particularly the `analytics settings` test, more stable on external environments by resetting the analytics settings during test setup. One reason why this test is failing on external environments is because the Analytics > Settings page is sometimes left with non-default values, resulting to Playwright not being able to locate objects that have default values (such as the Default Date Range setting) and failing assertions.

1. Log in to our Pressable and WPCOM test sites.
2. Go to their Analytics > Settings pages.
3. The image below shows the defaults of the three Analytics settings touched by the test. In the Pressable & WPCOM test sites, check and uncheck values such that they don't match the defaults anymore. Easiest way would be to uncheck all the options.
    Screenshot 2025-03-26 at 2 30 52 PM
4. Run the following test against our Pressable and WPCOM test sites, and make sure they both pass:
    ```bash
    # Set the E2E_ENV_KEY
    pnpm test:e2e:pressable analytics-data.spec.js -g 'analytics settings'
    pnpm test:e2e:wpcom analytics-data.spec.js -g 'analytics settings'
    ```
    Without this fix, they would fail.

Ran the entire `analytics-data.spec.js` file (not just the 'analytics settings' test) against the following environments:
- localhost (wp-env) with WP 6.7
- localhost (wp-env) with WP 6.8-RC1 (installed the WP version by setting `core` to the download url of WP 6.8 RC1) 
- Pressable site
- WPCOM site

Tested the error reporting of the API requests I added by:
- Changing [this](https://github.com/woocommerce/woocommerce/blob/638abf7d11ca8fcca4df98d9370f5d4baf79c83f/plugins/woocommerce/tests/e2e-pw/tests/analytics/analytics-data.spec.js#L171) request URL to get a 404 error and see if the Playwright HTML report could display an understandable error message. It didn't at first, but it did when I added a catch block to make the error message more readable and easier to point to the failing step.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/tests/e2e-pw/tests/analytics/analytics-data.spec.js
+++ b/plugins/woocommerce/tests/e2e-pw/tests/analytics/analytics-data.spec.js
@@ -165,18 +165,22 @@ test.beforeAll( async ( { browser, restApi } ) => {
 		} );
 
 	// Reset Analytics Settings to their default values.
-	// Setting options can be found at plugins/woocommerce/src/Internal/Admin/Settings.php.
-	await wcAdminApi.post( 'options', {
-		woocommerce_excluded_report_order_statuses: [
-			'pending',
-			'cancelled',
-			'failed',
-		],
-	} );
-	await wcAdminApi.post( 'options', {
+	await wcAdminApi
+		.put( 'options', {
+			woocommerce_excluded_report_order_statuses: [
+				'pending',
+				'cancelled',
+				'failed',
+			],
+		} )
+		.then( ( response ) => {
+			console.log( response );
+		} )
+		.catch( ( error ) => console.log( error ) );
+	await wcAdminApi.put( 'options', {
 		woocommerce_actionable_order_statuses: [ 'processing', 'on-hold' ],
 	} );
-	await wcAdminApi.post( 'options', {
+	await wcAdminApi.put( 'options', {
 		woocommerce_default_date_range: 'period=month&compare=previous_year',
 	} );
 
--- a/plugins/woocommerce/tests/e2e-pw/tests/analytics/analytics-data.spec.js
+++ b/plugins/woocommerce/tests/e2e-pw/tests/analytics/analytics-data.spec.js
@@ -4,6 +4,7 @@
 import { expect, tags, test as baseTest } from '../../fixtures/fixtures';
 import { WC_ADMIN_API_PATH, WC_API_PATH } from '../../utils/api-client';
 import { ADMIN_STATE_PATH } from '../../playwright.config';
+import { describe } from 'node:test';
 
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/tests/e2e-pw/tests/analytics/analytics-data.spec.js
```diff
@@ -164,6 +164,80 @@ test.beforeAll( async ( { browser, restApi } ) => {
...
+			}
+		)
+		.then( ( response ) => {
+			expect( response.data.value ).toEqual( [
+				'pending',
+				'cancelled',
+				'failed',
+			] );
+		} )
+		.catch( ( error ) => {
```
rodelgc, line 183: I thought of catching the error here and exposing its details and making the error more descriptive. When I used `/wc_admins/` instead of `/wc_admin` in the request URL in order to force an error, the HTML report looked confusing:

- The BeforeAll section and all of its steps were marked as ✅  even though obviously this request had failed. So it was difficult to look exactly at which step the error occurred.
- Even though the error message was shown on the HTML report, it was vague:
    ```js
     "code": "rest_setting_setting_group_invalid",
     "message": "Invalid setting group.",
    ```
    
When I added a catch block to format the error a little bit, the exact step in the BeforeAll hook section is now identifiable and correctly marked as ❌ in the HTML report.

plugins/woocommerce/tests/e2e-pw/tests/analytics/analytics-data.spec.js
```diff
@@ -164,6 +164,80 @@ test.beforeAll( async ( { browser, restApi } ) => {
 			orderIds = response.data.create.map( ( order ) => order.id );
 		} );
 
+	// Reset Analytics Settings to their default values.
+	// Reset 'Excluded statuses' to default values.
+	await restApi
+		.post(
+			'wc-analytics/settings/wc_admin/woocommerce_excluded_report_order_statuses',
```
rodelgc, line 171: Aside from this test, the other one that uses the `wc-analytics` endpoint is `customer-list.spec.js` but only as a wait criteria and not as an endpoint in a `restApi` request. So I didn't add a `WC_ANALYTICS` constant in `utils/api-client.js` but let me know if you think I should, and I'll add it on the next commit.

adimoldovan (approved):
Thanks for the update @rodelgc
It looks good, and tests and passed every time.

Issue #56447, opened 2025-03-17
Improve Webpack script module build


samueljseay:
In line with improving PHP script module support we should simplify the script modules Webpack build to allow Webpack to find all the frontend files needed to be built instead of listing them manually like we do now. 

Proposed solution:

* Declare `viewScriptModule` on `block.json` for interactivity API powered blocks
* Have a script detect all viewScriptModule entries to generate an entry list for Webpack
* Have script modules build also build styles for those blocks, remove any workaround code from the existing script builds to ensure that these blocks are built entirely in the new webpack config.

PR #56758, opened 2025-03-27
Improve script module support for interactivity API blocks

### LINKED ISSUE #56446: Improve script module support for blocks
Blocks that use scripts (versus script modules) rely on script support built into AbstractBlock. As we are moving many new blocks to interactivity API **with** script modules, we need a better suited class or mechanism for supporting script modules.<br><br>I'm proposing that we create a new class like AbstractScriptModuleBlock that is designed to support script modules.<br><br>It should:

* Load styles for editor and frontend like AbstractBlock does
* call enqueue_script_module as lazy as possible (perhaps simplest is just if render returns non empty string)
* Load the combined asset.php file for all script module blocks

samueljseay:

Closes #56446 and #56447 

* We improve the JS build for interactive blocks via Webpack and some scripts to auto detect entry points
* We use block.json to declare styles and scripts to enqueue
* To disable the existing asset and enqueue code in AbstractBlock we introduce a trait

N/A

Smoke test of each block that has moved to the new class:

1. Render the block frontend, smoke test functionality and styling
2. Render the block editor side, smoke test functionality and styling
3. Open console on both front-end and editor side, ensure no 404 happens trying to load any assets of these blocks
**Note** you may see failing requests for `product-description.css`, I've addressed that in https://github.com/woocommerce/woocommerce/pull/56886

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Introduce AbstractInteractiveBlock class with improved script module support.

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
```diff
@@ -8,11 +8,11 @@
 use Automattic\WooCommerce\Blocks\Utils\StyleAttributesUtils;
 use Automattic\WooCommerce\Enums\ProductType;
 use Automattic\WooCommerce\Blocks\Utils\BlockTemplateUtils;
-
+use Automattic\WooCommerce\Blocks\BlockTypes\AbstractInteractivityAPIBlock;
```
Aljullu: Is this needed?

plugins/woocommerce/client/blocks/tests/e2e/tests/product-filters/attribute-filter-editor.block_theme.spec.ts
```diff
@@ -49,7 +49,7 @@ test.describe( `${ blockData.name }`, () => {
 
 		await expect( block ).toBeVisible();
 
-		await block.click();
+		await editor.selectBlocks( block );
```
Aljullu, line 52: Do you know why this change was needed?

plugins/woocommerce/client/blocks/bin/webpack-interactivity-entries.js
```diff
@@ -0,0 +1,60 @@
...
+		const fullPath = path.join( dir, entry.name );
+		if ( entry.isDirectory() ) {
+			results = results.concat(
+				findInteractivityBlockAssets( fullPath, additionalPatterns )
+			);
+		} else if ( entry.isFile() && entry.name === 'block.json' ) {
+			// parse the json file and determine if its a block that supports interactivity.
+			const blockJson = JSON.parse( fs.readFileSync( fullPath, 'utf8' ) );
+
+			if ( blockJson.supports && blockJson.supports.interactivity ) {
```
mdperez86: Solved nicely by looking at this support! Great job.

plugins/woocommerce/src/Blocks/BlockTypes/AbstractInteractivityAPIBlock.php
```diff
@@ -0,0 +1,106 @@
...
+		$result                     = $this->render( $render_callback_attributes, $content, $block );
+
+		if ( ! empty( $result ) && ! is_admin() && ! WC()->is_rest_api_request() ) {
+			$this->enqueue_assets( $render_callback_attributes, $content, $block );
+		}
+
+		$this->enqueue_data( $attributes );
+
+		return $result;
+	}
```
Aljullu: No need to do it in this PR, but do you think the changes in this function could be done directly into `AbstractProduct`? This would keep both classes more similar to each other.

plugins/woocommerce/client/blocks/package.json
```diff
@@ -64,7 +64,7 @@
 		"lint:ci": "pnpm run lint:js && pnpm run lint:css",
 		"lint:css": "stylelint '**/*.scss'",
 		"lint:css-fix": "stylelint '**/*.scss' --fix",
-		"lint:js": "wp-scripts lint-js --ext=js,ts,tsx",
+		"lint:js": "WP_EXPERIMENTAL_MODULES=true wp-scripts lint-js --ext=js,ts,tsx ",
```
mdperez86: There is a space at the end of the script definition.

plugins/woocommerce/src/Blocks/BlockTypes/AbstractInteractivityAPIBlock.php
```diff
@@ -0,0 +1,106 @@
...
+	 *
+	 * If your block does not have a frontend stylesheet you can disable this method by returning null.
+	 *
+	 * @return string[]
+	 */
+	protected function get_legacy_editor_styles() {
+		$this->asset_api->register_style( 'wc-blocks-style-' . $this->block_name, $this->asset_api->get_block_asset_build_path( $this->block_name, 'css' ), [], 'all', true );
+
+		return 'wc-blocks-style-' . $this->block_name;
+	}
```
Aljullu: I might be confused, but `wc-blocks-style-<block_name>` is actually the frontend stylesheet, no? Mentioning it because the "legacy editor style" name is a bit confusing for me. :confused:

plugins/woocommerce/src/Blocks/BlockTypes/AbstractInteractivityAPIBlock.php
```diff
@@ -0,0 +1,106 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Automattic\WooCommerce\Blocks\BlockTypes;
+
+/**
+ * AbstractInteractivityAPIBlock class.
+ */
+abstract class AbstractInteractivityAPIBlock extends AbstractBlock {
```
mdperez86: Considering that the iAPI is a feature that we can add/remove to/from a block and not a block type specialization I think a better fit for this concept is to use a trait instead of an abstract class. This will allow us to plugin or out this specific block support and others if we follow the composition pattern instead of the inheritance one for similar cases.

Aljullu (commented):
Thanks for looking into improving the way we build styles, @samueljseay! I left some inline comments in the code, and something I noticed when testing the PR:

When going to `/wp-admin/post-new.php`, I can see two identical files loaded for each iAPI block. For example:

* `/plugins/woocommerce/assets/client/blocks/product-collection.css`
* `/plugins/woocommerce/assets/client/blocks/woocommerce/product-collection-style.css`

Can you reproduce as well? I compared and I think this is a regression. :thinking:

louwie17:
I just tested the Accordion blocks, which still tested well, thanks for all the iAPI work!

roykho:
For product filters, the rating stars are missing in the frontend.

CleanShot 2025-04-01 at 11 59 09

plugins/woocommerce/src/Blocks/BlockTypes/AbstractInteractivityAPIBlock.php
samueljseay: A trait would be fine maybe as a final design, but for this it's really done to handle technical debt where we support the inherited methods of AbstractBlock and override some where relevant. With a trait you can't enforce any type constraints of the original class and you can't extend methods in other classes.

This is not an ideal solution as it stands but done with the thought of juggling existing technical debt.

Also I also considered calling it AbstractScriptModuleBlock. I would be ok with that as a class name.

plugins/woocommerce/client/blocks/tests/e2e/tests/product-filters/attribute-filter-editor.block_theme.spec.ts
samueljseay: This was to fix these e2e tests. you shouldn't use `locator.click` to select a block because you might click anywhere on a block and it could cause any child block to be selected (which is what was happening here). Maybe by luck this wasn't happening before because of exact screen layout or pixels where the block was being clicked. 

Another more robust way might be to open the block list and click there, but this helper is what is used in Gutenberg

plugins/woocommerce/src/Blocks/BlockTypes/AbstractInteractivityAPIBlock.php
samueljseay: Into `AbstractBlock` you mean? this class is not just about solving styles, but to handle script modules. All non script module blocks do not want this code.

Sorry if my description might need some work on the PR, but the idea here is to separate out this functionality because its necessary for non script module blocks to continue functioning as they do. And we extend the original class also to reduce backwards compat issues with any iAPI blocks that depend on "lifecycle" methods from the original class.

In future when all blocks are script module blocks  we would:

1. Not extend the old class any more
2. Build all blocks stylesheets better
3. Remove any distinction between script module or non script module blocks

plugins/woocommerce/src/Blocks/BlockTypes/AbstractInteractivityAPIBlock.php
samueljseay: @Aljullu remember that both are bundled into one stylesheet right now! Neither answer is correct but the point is we're only loading this in the editor because now I build a frontend stylesheet just for frontend styles.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
samueljseay: Thanks this can be removed

samueljseay:
> Thanks for looking into improving the way we build styles

Just mentioning, styles are a small thing here we had to handle its true, but in reality this PR is more about script module support which iAPI blocks must have. 

> Can you reproduce as well? I compared and I think this is a regression.

This is intentional to handle technical debt. I mentioned this in the comment: https://github.com/woocommerce/woocommerce/pull/56758/files#diff-4bd41065f61f07cbf5f63ebfb0f3f4be7dc0dc7ea431ea35428347d05cc560abR80-R83

In the editor we load the old stylesheet which bundles editor styles and frontend styles :C

but in frontend we now only load the frontend styles, an improvement where it counts. 

The issue I linked in the comment discusses the fact we need to fix this bundling of styles together because now we load one more stylesheet in the editor. Its far from ideal, but in reality in most cases its not a tonne of extra styles.

All that said, we could look at a fast-follow PR after this one to try remedy this situation, one way would be to (just for iAPI blocks) fix the styles build so they only build editor side styles in the editor stylesheet (and we name it more appropriately like `-editor-styles` for clarity.


EDIT: Actually maybe we could just disable loading the new frontend stylesheet in editor for now 😅

samueljseay:
> For product filters, the rating stars are missing in the frontend.

Thanks for spotting! I'll see if I can reproduce this.

plugins/woocommerce/client/blocks/bin/webpack-interactivity-entries.js
kmanijak: I'm marking some more blocks with `supports.interactivity` in this PR: https://github.com/woocommerce/woocommerce/pull/56661.

Just FYI, no action needed Sam.

kmanijak: Hmm.. or maybe there is.

You're marking blocks that actually hold the `frontend.ts` script with `supports.interactivity: true`. And those you refactored to extend from `AbstractInteractivityAPIBlock`.

I'm marking _all_ the blocks that use any directives even if they're supposed to be used as a child of interactive block and don't have `frontend.ts` 🤔 I don't think that matters much but it may cause confusion in PHP files once we forget about this discussion 😄

kmanijak (commented):
Smoke tested Product Collection, all seems to be working fine! 💪

As always, amazing job @samueljseay!

plugins/woocommerce/client/blocks/tests/e2e/tests/product-filters/attribute-filter-editor.block_theme.spec.ts
Aljullu: It makes sense, thanks!

plugins/woocommerce/src/Blocks/BlockTypes/AbstractInteractivityAPIBlock.php
Aljullu: Sorry if I wasn't clear, I meant only the changes inside `render_callback()`. If I'm understanding it right, the same changes could be done to `AbstractBlock`, no? Then, the other methods will still be different (`enqueue_assets()`, `enqueue_data()`, etc.), but I _think_ there isn't anything blocking `render_callback()` to be the same. The changes you applied here could work in `AbstractBlock`.

But definitely not something that needs to be handled in this PR.

plugins/woocommerce/src/Blocks/BlockTypes/AbstractInteractivityAPIBlock.php
Aljullu: So if I'm understanding it right, with this branch we generate two files for each block:

* `/plugins/woocommerce/assets/client/blocks/woocommerce/<block_name>-style.css`: only frontend styles.
* `/plugins/woocommerce/assets/client/blocks/<block_name>.css`: frontend + editor styles bundled.

Is that correct? If so, I think the naming could be improved:

* Why one of these files is inside a `woocommerce` subfolder?
* Could we change the `-style` suffix to `-frontend` to match what we do with JS files?

What do you think?

---

Answering your other comment from below:

> EDIT: Actually maybe we could just disable loading the new frontend stylesheet in editor for now 😅

That would work! If I'm understanding it right, and all `/plugins/woocommerce/assets/client/blocks/woocommerce/<block_name>-style.css` are either a subset of or equal to `/plugins/woocommerce/assets/client/blocks/<block_name>.css`, loading only the latter in the editor should work.

Aljullu: Btw, I noticed this happens in `trunk` too, but it looks like we are generating the same CSS files twice in two different locations:

* `/plugins/woocommerce/assets/client/blocks/`
* `/plugins/woocommerce/client/blocks/build/`

Can you reproduce too, @samueljseay? I can open a separate issue to investigate that.

cpapazoglou (commented):
Smoke tested the `Product Gallery` block and couldn't find any regression. Good job 💪

plugins/woocommerce/src/Blocks/BlockTypes/AbstractInteractivityAPIBlock.php
luisherranz: The editor will still use scripts, it's just the frontend part that will use script modules. Therefore, maybe a better name (more generic) would be just `AbstractInteractiveBlock`.

plugins/woocommerce/client/blocks/bin/webpack-interactivity-entries.js
samueljseay: @kmanijak thanks for the heads up. I think this is totally ok at first glance. It is totally fine for a block to support interactivity yet not have a frontend.ts file and ok not to extend this new class either 

> I don't think that matters much but it may cause confusion in PHP files once we forget about this discussion
This could be an argument for alternative naming of the class like originally I considered calling it `AbstractScriptModuleBlock`... 🤔

plugins/woocommerce/src/Blocks/BlockTypes/AbstractInteractivityAPIBlock.php
samueljseay: @luisherranz Yep totally I can get behind that.

plugins/woocommerce/src/Blocks/BlockTypes/AbstractInteractivityAPIBlock.php
samueljseay: @Aljullu oh sure, I can see your perspective on that :) 

>  If I'm understanding it right, the same changes could be done to AbstractBlock, no?

I'll have a think about that, I don't mind making those kind of changes here if it makes sense :)

plugins/woocommerce/src/Blocks/BlockTypes/AbstractInteractivityAPIBlock.php
samueljseay: Resolved in https://github.com/woocommerce/woocommerce/pull/56758/commits/08c99e96594daee46a84c9071ce9c1a25dce7d8b

plugins/woocommerce/src/Blocks/BlockTypes/AbstractInteractivityAPIBlock.php
samueljseay: I think it's most likely reasonable we can port these changes from render callback. It *may* in some cases cause issues if one of the existing classes does something like render an empty block but still depend on a script or style being loaded. This would likely be the kind of thing that's implicit and not by design, but I really wouldn't be surprised if it exists.

For that reason I definitely want to do this as a follow up PR so I can properly test and confirm I don't regress anything since this change would touch a lot of blocks and styling regressions especially won't likely be covered by testing.

samueljseay:
Rating stars issue is solved in https://github.com/woocommerce/woocommerce/pull/56758/commits/210437ae83c14339c990a6f82d93974de71380c3 

It's an interesting case and right now it won't come up too often on interactive blocks, but I'm not sure of a nice way to fix it. It's something we should think of more broadly when building the interactive blocks going forward. We don't want to continue building all styles loaded from something like `index.tsx` because the whole point is to separate out frontend styles and editor only styles and not load both. To solve it here I've manually imported the frontend styles from that component to the `styles.scss` for the block.

plugins/woocommerce/client/blocks/bin/webpack-interactivity-entries.js
samueljseay: @kmanijak I was actually a little bit wrong about the block.json spec for interactive support,

I've changed the detection now to look for either: `"interactivity: true"` or `"interactivity": { "interactive": true }` 

Not a huge deal, but I think for all the blocks we marked as `clientNavigation: false` it would be nice to also explicitly add `interactive: false` unless they have directives.

plugins/woocommerce/src/Blocks/BlockTypes/AbstractInteractivityAPIBlock.php
samueljseay: Still on my list to investigate that one further. Thanks for raising it!

plugins/woocommerce/client/blocks/bin/webpack-interactivity-entries.js
kmanijak: Wait, that's new, isn't it? I'm just checking [documentation](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-supports/#interactivity) and indeed it's there!

> Not a huge deal, but I think for all the blocks we marked as clientNavigation: false it would be nice to also explicitly add interactive: false unless they have directives.

Yeah, makes sense.

kmanijak: It's not new at all 🤦 no idea how I did miss it 😅

plugins/woocommerce/src/Blocks/BlockTypes/AbstractInteractivityAPIBlock.php
samueljseay: @Aljullu I think this is expected, but still a bit unfortunate because I was hoping to remove this when we merged blocks into the woocommerce plugin. 

We still do a build of the assets within block library (`/plugins/woocommerce/client/blocks/build/`) then we do an [rsync copy](https://github.com/woocommerce/woocommerce/blob/4b6a1b60e6109b0790051f1bcf0e542ea6d91da3/plugins/woocommerce/package.json#L21) of them into the `/plugins/woocommerce/assets/client/blocks/` directory.

I also double checked just for sanity the prod release and still only the assets in `/plugins/woocommerce/assets/client/blocks/` are shipped with the plugin.

samueljseay: > Why one of these files is inside a woocommerce subfolder?

This was a convention in the entry point name, I based them on the block name to nest them under a separate folder from the non-script module scripts but also so we could register the script from the block name. It's not really a big deal if we don't have this convention though, it's just convenient.

> Could we change the -style suffix to -frontend to match what we do with JS files?

I think its better to match block json semantics here, which I will adjust a little bit in this PR now that I've looked closer at block json semantics.

I was a little wrong about the purpose of `style.scss` in our codebase. Block json has three types of styles:

* `style` (which is both editor and f/e) 
* `viewStyle` (f/e only), 
* `editorStyle` (editor only),

We don't use [f/e only styles right now](https://github.com/woocommerce/woocommerce/blob/4b6a1b60e6109b0790051f1bcf0e542ea6d91da3/plugins/woocommerce/src/Blocks/BlockTypes/AbstractBlock.php#L223-L228). Most of the time we use a combination of `style`(both)  and `editorStyle` (editor only)

So I want to keep it like:

`<blockname>-style`
`<blockname>-view`
`<blockname>-editor`

I won't introduce building styles for `view.scss` here because its not a convention we've used, but its not exactly clear if everyone using a `style.scss` understood that this was a file designed to be loaded in both editor and f/e. I can see that in most cases you could live with using either a combo of style/view or style/editor and it wouldn't matter too much.

samueljseay:
@Aljullu I was wondering if you'd mind taking another look (and a quick smoke test of add to cart with options if you can)

I realized that actually the way styles are in AbstractBlock right now styles are like this:

* styles imported in index.tsx for editor and frontend
* editor styles get combined into a single stylesheet thats loaded for all blocks
*  styles.scss is generated as standalone stylesheet with name `block-name.css`

So I made some adjustments. I moved all style generation for interactive blocks into the new script module build. Now for AbstractInteractiveBlock I'm moving away from importing stylesheets in JS (long term this is good to align with Gutenberg and also tie closer to block json semantics). 

Styles I generate for interactive blocks now:

* styles.scss -> loaded in both editor and f/e with name like `woocommerce/block-name-style.css`
* editor.scss -> loaded just in editor and f/e with name like `woocommerce/block-name-editor.css`

both of these are loaded via block settings of `styles` and `editorStyles`

Later we could introduce view.scss (just loaded in f/e via `viewStyle`) but no one really uses that concept in the blocks right now.

samueljseay:
@luisherranz Would love for you to have a high level review of everything here and see if anything problematic stands out to you.

plugins/woocommerce/src/Blocks/BlockTypes/AbstractInteractivityAPIBlock.php
Aljullu: > For that reason I definitely want to do this as a follow up PR so I can properly test and confirm I don't regress anything since this change would touch a lot of blocks and styling regressions especially won't likely be covered by testing.

Sure, it makes sense!

plugins/woocommerce/src/Blocks/BlockTypes/AbstractInteractivityAPIBlock.php
Aljullu: > So I want to keep it like:
> 
> `<blockname>-style`
> `<blockname>-view`
> `<blockname>-editor`

Good points. Using these suffixes sounds good to me. It might be worth documenting it somewhere in the future so we keep it consistent with Gutenberg.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsGroupedProductSelector.php
```diff
@@ -3,10 +3,12 @@
 
 namespace Automattic\WooCommerce\Blocks\BlockTypes;
 
+use Automattic\WooCommerce\Blocks\BlockTypes\AbstractInteractiveBlock;
```
Aljullu: Is this line needed? `AbstractInteractiveBlock` is already inside the used namespace, no?

Aljullu (commented):
Thanks for the updates, @samueljseay! Overall it's looking good, but I noticed a couple of issues so far:

1. In the existing _Add to Cart with Options_ block (not the blockified one), the skeleton component is not visible:

imatge

2. I see some 404 requests when opening the post editor or the Site Editor:

imatge

What's weird is that those blocks have CSS files (ie: https://github.com/woocommerce/woocommerce/blob/8b25ba0dac1810006851f98bf9cee91d93985020/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-cta/style.scss). So the styles are missing in the editor:

imatge

Can you reproduce too?

samueljseay:
@Aljullu Thanks, I didn't see that when testing the add to cart with options. I'll investigate.

@luisherranz Don't worry about reviewing this right away. @dinhtungdu and I did some discovery yesterday and we found that actually we could use block.json as is with `viewScriptModule`, and remove `AbstractInteractiveBlock`. I realize this isn't high priority, so I won't spend too much time on it, but I don't want to introduce a new class if we don't actually need it.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsGroupedProductSelector.php
samueljseay: 🤦🏻‍♂️ I think cursor gets a bit excited and adds these use statements, then I don't always see it added it. Thanks for spotting.

samueljseay:
Skeleton loader css not loading is fixed.
404s on scripts and styles fixed.
Note that if we want to load styles from a component right now we need to import them in the stylesheet rather than in the JS, as I am determined we move away from importing scss in the JS to move closer to Gutenberg standards.

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
```diff
@@ -79,6 +68,7 @@ public function dequeue_add_to_cart_scripts() {
 	 * @return string Rendered block type output.
 	 */
 	protected function render( $attributes, $content, $block ) {
+		// wp_enqueue_script_module( $this->get_full_block_name() );
```
Aljullu: Are we keeping this line on purpose?

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartItemsBlock.php
```diff
@@ -1,14 +1,37 @@
...
+	 * @param array    $attributes Block attributes.
+	 * @param string   $content    Block content.
+	 * @param WP_Block $block      Block instance.
+	 * @return string Rendered block.
+	 */
+	protected function render( $attributes, $content, $block ) {
+		ob_start();
+		?>
+		<div class="wc-block-mini-cart__items" data-wp-interactive="<?php echo esc_attr( $this->get_full_block_name() ); ?>">
+			<p>Hello World</p>
```
Aljullu: Is this expected? :thinking:

plugins/woocommerce/src/Blocks/BlockTypes/EnableBlockJsonAssetsTrait.php
```diff
@@ -0,0 +1,36 @@
...
+
+trait EnableBlockJsonAssetsTrait {
+
+	/**
+	 * Disable the script handle for this block type. We use block.json to load the script.
+	 *
+	 * @param string|null $key The key of the script to get.
+	 * @return null
+	 */
+	// phpcs:ignore
```
Aljullu, line 14: Could we make this ignore specific to the rules that need to be ignored?

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
```diff
@@ -109,7 +109,9 @@ const { state } = store< Store >(
 
 				// Todo: Use the module exports instead of `store()` once the
 				// woocommerce store is public.
+				console.log( 'hello loading stores' );
 				yield import( '@woocommerce/stores/woocommerce/cart' );
+				console.log( 'hello loaded stores' );
```
Aljullu: This probably needs to be removed. :slightly_smiling_face:

Aljullu (approved):
Thanks for the updates, @samueljseay! This is testing well on my end. I also like the use of a trait to handle when we load the styles and scripts from `block.json`. From my side, this is good to go. I left some minor inline comments for code which I think needs to be removed, but besides that, LGTM.

<details>

* Theme: Twenty Twenty-Five
* Store size: small
</details>

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartItemsBlock.php
samueljseay: Haha whoops pushed changes from an unrelated branch 🤦‍♂️

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
samueljseay: No. Thanka for spotting. Leaving mess behind today 🫠

plugins/woocommerce/src/Blocks/BlockTypes/EnableBlockJsonAssetsTrait.php
samueljseay: Yep good point

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
samueljseay: Yes sorry for the pollution from another branch 🥲

luisherranz:
> @luisherranz Don't worry about reviewing this right away. @dinhtungdu and I did some discovery yesterday and we found that actually we could use block.json as is with viewScriptModule, and remove AbstractInteractiveBlock. I realize this isn't high priority, so I won't spend too much time on it, but I don't want to introduce a new class if we don't actually need it.

So, what's the plan then? Close this PR and open another? Or make the changes in this pull request?  Or leave everything as it is for the moment?

samueljseay:
@luisherranz sorry I haven't updated the comments to reflect the state. 

>  Or make the changes in this pull request?

Yes this is done, it is good to go now!

I spent a little time on the new approach and it worked great.  We don't introduce any new classes. I just introduced a trait to deal with some tech debt. but I moved the asset handles into the block json and it works 🎉 

The only thing I was still working on is just ensuring that CI is passing.

plugins/woocommerce/src/Blocks/BlockTypes/EnableBlockJsonAssetsTrait.php
samueljseay: Couldn't fix this in the end. phpcs seems to get confused about other lint errors that don't exist, ie that the arg wasn't commented in the comment block. Gave up. Wasn't worth the time.

samueljseay:
@luisherranz hope it's ok, but I'm going to go ahead and merge this now. When last we talked I remember you didn't have too strong feelings.

Now I think it's even less controversial from your side and still improves the situation of script modules and gets us a little closer to core.

luisherranz:
Perfect, thanks Sam 🙂

PR #56768, opened 2025-03-27
Avoid fatal error when filter attributes are passed as array in the URL

### LINKED ISSUE #55100: [Fatal] PHP Fatal error: Uncaught TypeError: explode(): Argument #2 ($string) must be of type string, array given in | class-wc-query.php:928
### Prerequisites

### Describe the bug

witnessed in `9.6.0`

[link to code](https://github.com/woocommerce/woocommerce/blob/6f5032a0277a0eb006a442407d9d09c0c370b8be/plugins/woocommerce/includes/class-wc-query.php#L928)

```
PHP Fatal error:  Uncaught TypeError: explode(): Argument #2 ($string) must be of type string, array given in /../plugins/woocommerce/9.6.0/includes/class-wc-query.php:928
Stack trace:
#0 /../plugins/woocommerce/9.6.0/includes/class-wc-query.php(928): explode(',', Array)
#1 /../plugins/woocommerce/9.6.0/includes/class-wc-query.php(556): WC_Query::get_layered_nav_chosen_attributes()
#2 /../core/6.7.1/wp-includes/class-wp-hook.php(324): WC_Query->product_query_post_clauses(Array, Object(WP_Query))
#3 /../core/6.7.1/wp-includes/plugin.php(256): WP_Hook->apply_filters(Array, Array)
#4 /../core/6.7.1/wp-includes/class-wp-query.php(2955): apply_filters_ref_array('posts_clauses', Array)
#5 /../core/6.7.1/wp-includes/class-wp-query.php(3852): WP_Query->get_posts()
#6 /../core/6.7.1/wp-includes/class-wp.php(696): WP_Query->query(Array)
#7 /../core/6.7.1/wp-includes/class-wp.php(816): WP->query_posts()
#8 /../core/6.7.1/wp-includes/functions.php(1336): WP->main('')
#9 /../core/6.7.1/wp-blog-header.php(16): wp()
#10 /../core/6.7.1/index.php(17): require('/../core...')
#11 {main}
  thrown in /../plugins/woocommerce/9.6.0/includes/class-wc-query.php on line 928
``` 

### Expected behavior

no fatals expected

### Actual behavior

fatal errors witnessed

### Steps to reproduce

We don't have steps to reproduce, but observed 122 instances of this problem on Fri Jan 31, 2025 across 1 site. Hopefully, someone can dig in and apply a defensive check if appropriate. - To be updated after investigation.

### WordPress Environment

N/A

### Isolating the problem

albarin:

This PR introduces checks to ensure the attribute filters are passed as strings, not arrays.
They are not processed as valid filters if they are passed as arrays from the URL.

Closes https://github.com/woocommerce/woocommerce/issues/55100

1. While on `trunk`, go to `Appearance > Themes > Customize`, edit the `Product Catalog` template, and add an attribute filter (like `Colors` or `Size`).
2. Save and go to `/shop`.
3. In the Shop page, select some attribute terms in the filter.
4. You'll see the URL change to reflect the selected terms. For example: `/shop/?filter_color=green,yellow&query_type_color=or`.
5. Change the `filter_color` in the URL to be an array: `/shop/?filter_color[]=green&filter_color[]=yellow&query_type_color=or` and see you get a fatal error:
`Fatal error: Uncaught TypeError: explode(): Argument #2 ($string) must be of type string, array given in /app/public/wp-content/plugins/woocommerce/includes/class-wc-query.php:928`
6. Check out to this PR branch, refresh, and check that you no longer see the error.
7. Go again to the `/shop` page, smoke test the attribute filters, and make sure the product results make sense with what's selected.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Avoid fatal error when filter attributes are passed as array in the URL
</details>

<details>

#### Comment 

</details>


jorgeatorres (commented):
@albarin: This looks good to me for preventing the fatal but I'm wondering _why_ the attributes would be passed as an array in the first place. Do you know if we have some code that would be doing that? I really doubt someone was manually changing URL parameters to produce the error. Can you think of any situation where the filtering might end up producing an incorrect URL?

Also, the code expects the value to be a comma-separated list of terms. Maybe an alternative to ignoring the filter when the value is an array would be to handle both possible value types: if it's a string we split it just like right now. If it's an array, we consider each item as a term (and sanitize individually).
Not sure if that's worth pursuing, though. What do you think?

albarin:
@jorgeatorres 

> @albarin: This looks good to me for preventing the fatal but I'm wondering _why_ the attributes would be passed as an array in the first place. Do you know if we have some code that would be doing that? I really doubt someone was manually changing URL parameters to produce the error. Can you think of any situation where the filtering might end up producing an incorrect URL?

I've been wondering the same thing 😅 I couldn't find anything in core that could produce those URLs with arrays. I thought that maybe it was some advanced filtering plugin, and I did quite a bit of research. I tested 3 or 4 of those plugins, but I couldn't really reproduce it. Maybe it's some kind of custom code; it's hard to say, and the stack trace does not help much, either.

> Also, the code expects the value to be a comma-separated list of terms. Maybe an alternative to ignoring the filter when the value is an array would be to handle both possible value types: if it's a string we split it just like right now. If it's an array, we consider each item as a term (and sanitize individually). Not sure if that's worth pursuing, though. What do you think?

I did try that, in fact, I have a branch with that partially implemented, but I got into a rabbit hole when having array params in the URL and interacting with the filters blocks, the URLs ended up being a mess and I though it was not work investing more time just to support that when I wasn't even able to reproduce in any way.

jorgeatorres:
> Maybe it's some kind of custom code; it's hard to say, and the stack trace does not help much, either.

Yeah, given the URL is already that way, the stack trace isn't super useful to guess what happened. Thanks for doing the research.

> but I got into a rabbit hole when having array params in the URL and interacting with the filters blocks, the URLs ended up being a mess

Fair enough. For what it's worth I wasn't suggesting we kept the arrays in the URL, just that if we found one, we could process it and turn it into a plain string in the URL, but that would also require some changes somewhere, which might be complicated.

I'm good with the fix as is, but if you want to add anything, let me know. Otherwise, I'll do another round of testing + reviewing.

jorgeatorres (approved):
LGTM. Thank you @albarin!

Issue #56485, opened 2025-03-18
Fatal Error On override Related Products Columns and Items. WooCommerce 9.80 Beta 1


BackuPs:
### Prerequisites

- [x] I have carried out troubleshooting steps and I believe I have found a bug.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

In Woo 9.80 Beta 1  The single product produces a Fatal error which does not happen in 9.7.1 or before.

Fatal error: Uncaught TypeError: array_slice(): Argument #3 ($length) must be of type ?int, string given in /var/docs/public/wp-content/plugins/woocommerce/includes/wc-product-functions.php:1154 Stack trace: #0 /var/docs/public/wp-content/plugins/woocommerce/includes/wc-product-functions.php(1154): array_slice() #1 /var/docs/public/wp-content/plugins/woocommerce/includes/wc-template-functions.php(2285): wc_get_related_products() #2 /var/docs/public/wp-content/themes/CustomTheme/framework/functions/woocommerce.php(523): woocommerce_related_products() #3 /var/docs/public/wp-includes/class-wp-hook.php(324): theme_woocommerce_output_related_products() #4 /var/docs/public/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters() #5 /var/docs/public/wp-includes/plugin.php(517): WP_Hook->do_action() #6 /var/docs/public/wp-content/plugins/woocommerce/templates/content-single-product.php(72): do_action() #7 /var/docs/public/wp-includes/template.php(812): require('...') #8 /var/docs/public/wp-content/plugins/woocommerce/includes/wc-core-functions.php(285): load_template() #9 /var/docs/public/wp-content/themes/CustomTheme/woocommerce/single-product.php(36): wc_get_template_part() #10 /var/docs/public/wp-includes/template-loader.php(106): include('...') #11 /var/docs/public/wp-blog-header.php(19): require_once('...') #12 /var/docs/public/index.php(17): require('...') #13 {main} thrown in /var/docs/public/wp-content/plugins/woocommerce/includes/wc-product-functions.php on line 115

$Liimit when submitted of type string after applying the filter woocommerce_output_related_products_args produced the error.

Here is the code used 

		$output = "";
		$related_count = '2';$related_columns = '2';
		$args = array(
			'posts_per_page' => $related_count,
			'columns' => $related_columns,
			'orderby' => 'rand'
		);
		ob_start();
[...]

PR #56790, opened 2025-03-28
wc_get_related_products: Typecheck against limit parameter

tjcafferkey:

Improve type handling for limit parameter in `wc_get_related_products()`

Adds better type handling and developer feedback for the limit parameter in `wc_get_related_products()`:

1. Adds strict type checking using `is_numeric()` before casting
2. Forces integer casting to ensure `array_slice()` receives correct type
3. Adds error logging when non-numeric values are passed to help developers identify misuse
4. Maintains existing fallback behavior to prevent breaking changes

The error logging includes:
- The type of value that was passed
- The actual value for debugging
- Source identification for filtering logs

Example log message:
"Invalid limit type passed to wc_get_related_products. Expected numeric, got string with value: 'all'"

This change helps prevent the fatal error:
`"TypeError: array_slice(): Argument #3 ($length) must be of type ?int, string given"`

Related https://github.com/woocommerce/woocommerce/issues/56485

| Before | After |
| ------ | ----- |
|        |       |

1. Ensure CI passes
2. Test that `wc_get_related_products( $product->get_id(), 'hello' )` returns an empty array
3. Test that `wc_get_related_products( $product->get_id(), '5' )` returns related products but logs an error
4. Test that `wc_get_related_products( $product->get_id(), 5 )` returns related products

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


cpapazoglou (changes requested):
My understanding:
- related products isn't a critical feature
- the block itself can be present in critical flows though ( eg single product page )

I consider it fair:
✅ not throwing a fatal error, but instead logging an error.
✅ casting to int for numeric values, generally WP uses a lot of strings that are casted to bool or int.
❌ not showing the related products if limit isn't numeric. Showing related products with the default limit might falsely seem as the feature is working. Since it's not critical to actually display something and given we log an error, I would return early.

tjcafferkey:
@cpapazoglou thanks for your feedback, just so I am clear you think:

* If the `$limit` value is a string, try casting it to a number
* If after this, the value is still not numeric then return early.

This should cover us for those who are using it correctly (`$limit = 5`) and those who aren't (`$limit = '5'`)

Have I understood correctly?

cpapazoglou:
> @cpapazoglou thanks for your feedback, just so I am clear you think:
> 
> * If the `$limit` value is a string, try casting it to a number
> * If after this, the value is still not numeric then return early.
> 
> This should cover us for those who are using it correctly (`$limit = 5`) and those who aren't (`$limit = '5'`)
> 
> Have I understood correctly?

`If the `$limit` value is a string, try casting it to a number`

Only, if `is_numeric($limit) = true`. If `is_numeric($limit) = false`, then return early instead of defaulting to `5`.

This will still work for `$limit = 5` and `$limit = '5'` and will safely fail for:

- `$limit = 'not_a_number'`
- `$limit = bool`

plugins/woocommerce/includes/wc-product-functions.php
```diff
@@ -1097,7 +1097,19 @@ function wc_get_product_backorder_options() {
...
+			sprintf(
+				'Invalid limit type passed to wc_get_related_products. Expected numeric, got %s with value: %s',
+				gettype( $limit ),
+				wp_json_encode( $limit )
+			),
+			array( 'source' => 'wc_get_related_products' )
+		);
+	}
 
+	$limit          = is_numeric( $limit ) ? (int) $limit : 5;
```
cpapazoglou:
```suggestion
		  return array(); // Return early.
		);
	}

	$limit          = (int) $limit;
```

tjcafferkey:
@cpapazoglou I've pushed some changes that maintains its original usage (accepts '5' but logs error) and also doesnt fatal error. I've included some tests for this too.

plugins/woocommerce/changelog/update-wc-get-related-products-limit
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: update
+
+Logs error if $limit isnt numeric and avoids a fatal error by returning an empty array
```
cpapazoglou, line 4:
```suggestion
Related Products: Logs error if $limit isn't numeric and avoids a fatal error by returning an empty array
```

plugins/woocommerce/tests/php/includes/wc-product-functions-test.php
```diff
@@ -240,4 +240,48 @@ public function test_wc_scheduled_sales_sale_end() {
...
+		wp_set_object_terms( $related_product3->get_id(), $category_term['term_id'], 'product_cat' );
+
+		// Save all products.
+		$main_product->save();
+		$related_product1->save();
+		$related_product2->save();
+		$related_product3->save();
+
+		// Get related products with a limit of 2.
+		$related_products = wc_get_related_products( $main_product->get_id(), 2 );
```
cpapazoglou, line 269: Let's also test with a string numeric value of `'2'`.

cpapazoglou (approved):
I like this approach:

- it educates devs about the parameter being `int`
- it allows numeric values to avoid regressions
- avoids fatals if it's not numeric

Tested it and works well:

✅ int value, displayed that number of products
✅ numeric string, displayed that number of products & logged and error
✅ non numeric, didn't display Related Products & logged and error

Feel free to add another test case for numeric strings and 🚢 !

plugins/woocommerce/tests/php/includes/wc-product-functions-test.php
tjcafferkey: https://github.com/woocommerce/woocommerce/pull/56790/commits/aacc83da95c5963f9001bec23c0ef5c8f208fa59

Issue #55700, opened 2025-02-20
Blockified Add to Cart with Options block: work on a new API for extensions that hook into the form using PHP hooks


Aljullu:
Even though we will try to support extensions that use PHP hooks in the _Add to Cart with Options_ form (https://github.com/woocommerce/woocommerce/issues/55699), we would like to offer a more intentional API so extensions can leverage the block paradigm in their extensions. This issue is about thinking the shape of the API we will recommend and how extensions will be able to use it while still using PHP hooks for classic themes and PHP templates.

This might rely on [Block Hooks](https://make.wordpress.org/core/2023/10/15/introducing-block-hooks-for-dynamic-blocks/).

helgatheviking:
Following! I already have the front end blocks, but need a way to register them for the backend (and in the way that they only show when my product type is being edited)

Aljullu:
I'm going to close this issue as we will be working on this in a more granular way, case by case, instead of in a meta issue like this one. To clarify: we will still work on the _Add to Cart + Options_ block extensibility, it's just that this GitHub issue isn't coupled to any specific task. 🙂

Issue #56682, opened 2025-03-25
Blockified Add to Cart with Options block: don't render hooks content if simple product is out of stock or not purchasable


Aljullu:
Since https://github.com/woocommerce/woocommerce/pull/56390, we hide the quantity selector and add to cart button for simple products which are out of stock or not purchasable. We should also hide the content added by PHP filters.

## Steps to reproduce

0. Install and activate the WooCommerce Beta Tester plugin.
0. Go to _Tools_ > _WCA Helper_ > _Features_ and enable `experimental-blocks` and `blockified-add-to-cart`.
1. Install [WooCommerce Product Add-Ons](https://woocommerce.com/products/product-add-ons/).
2. Create an out of stock product with a product add-on.
3. Create a post or page, add the _Single Product_ block and choose the product you just edited.
4. In the frontend, notice the add-on form element is shown but the quantity selector and add to cart button aren't.

![Image](https://github.com/user-attachments/assets/9c3a80db-b19b-4930-8290-f29cacdcefa8)

Expected behavior: the add-on form element should be hidden as well.

Issue #55701, opened 2025-02-20
Blockified Add to Cart with Options block: work on a new API for extensions that register 3rd-party product types


Aljullu:
For extensions that register 3rd-party product types, we currently display a placeholder in the editor and the PHP template in the frontend. However, we will need to create an intentional API so extensions can register their own template parts and use blocks for the forms of their 3rd-party block types. This issue is about thinking the shape of the API we will recommend.

## Tasks

* [x] Introduce hook that allows extensions to register template parts: [#56799](https://github.com/woocommerce/woocommerce/pull/56799)
* [x] Display the templates in the frontend: [#56799](https://github.com/woocommerce/woocommerce/pull/56799)
  * [ ] Add automated tests
* [ ] Display the templates in the editor when editing the _Add to Cart with Options_ [#60914](https://github.com/woocommerce/woocommerce/pull/60914)
  * [ ] Add automated tests
* [ ] Display the templates in the Site Editor > Patterns > Add to Cart with Options [#60914](https://github.com/woocommerce/woocommerce/pull/60914)
  * [ ] Add automated tests

helgatheviking:
😍 following

Aljullu:
I'm going to close this issue as we will be working on this in a more granular way, case by case, instead of in a meta issue like this one. To clarify: we will still work on the _Add to Cart + Options_ block extensibility, it's just that this GitHub issue isn't coupled to any specific task. 🙂

PR #56799, opened 2025-03-28
Dev/blockified Display add to cart block template custom product types

helgatheviking:
This introduces a filter for plugins to declare a template part for blockified add to cart for custom product types. It would be used like so from another plugin:

```
add_filter( 'woocommerce_taco_add_to_cart_block_template_part', function() {
	return plugin_dir_path( __FILE__ ) . '/templates/taco-product-add-to-cart-with-options.html';
}, 10, 2 );
```

*Update* with the `__experimental_` prefix

```
add_filter( '__experimental_woocommerce_taco_add_to_cart_with_options_block_template_part', function() {
	return plugin_dir_path( __FILE__ ) . '/templates/taco-product-add-to-cart-with-options.html';
}, 10, 2 );
```

NB: This is _only_ related to the front-end display.

here's a [complete mini-plugin](https://gist.github.com/helgatheviking/f49ebaf44b1fe70b4ab8e993336491f9) that creates a custom product type and it's relevant block template for testing this PR. (updated with the experimental filter prefix)

Result for the display of my "taco" type product is the custom taco add to cart template!
image

related to #55700 cc @Aljullu

### DIFF
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
@@ -242,13 +242,17 @@ class AddToCartWithOptions extends AbstractBlock {
 			ob_start();
 
 			if ( in_array( $product_type, array( ProductType::SIMPLE, ProductType::EXTERNAL, ProductType::VARIABLE, ProductType::GROUPED ), true ) ) {
+
 				$add_to_cart_fn = 'woocommerce_' . $product_type . '_add_to_cart';
+				remove_action( 'woocommerce_' . $product_type . '_add_to_cart', $add_to_cart_fn, 30 );
+
 				/**
 				 * Trigger the single product add to cart action that prints the markup.
 				 *
 				 * @since 9.9.0
 				 */
 				do_action( 'woocommerce_' . $product->get_type() . '_add_to_cart' );
+				add_action( 'woocommerce_' . $product_type . '_add_to_cart', $add_to_cart_fn, 30 );
 			}
 
 			$form_html = $form_html . ob_get_clean();
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
@@ -158,20 +158,6 @@ class AddToCartWithOptions extends AbstractBlock {
 					 * @since 1.5.0
 					 */
 					do_action( 'woocommerce_before_add_to_cart_button' );
-				} elseif ( ProductType::EXTERNAL === $product_type ) {
-					/**
-					 * Hook: woocommerce_before_add_to_cart_button.
-					 *
-					 * @since 1.5.0
-					 */
-					do_action( 'woocommerce_before_add_to_cart_button' );
-				} elseif ( ProductType::GROUPED === $product_type ) {
-					/**
-					 * Hook: woocommerce_before_add_to_cart_button.
-					 *
-					 * @since 1.5.0
-					 */
-					do_action( 'woocommerce_before_add_to_cart_button' );
 				} elseif ( ProductType::VARIABLE === $product_type ) {
 					/**
 					 * Hook: woocommerce_before_variations_form.
@@ -179,6 +165,13 @@ class AddToCartWithOptions extends AbstractBlock {
 					 * @since 2.4.0
 					 */
 					do_action( 'woocommerce_before_variations_form' );
+				} else {
+					/**
+					 * Hook: woocommerce_before_add_to_cart_button.
+					 *
+					 * @since 1.5.0
+					 */
+					do_action( 'woocommerce_before_add_to_cart_button' );
 				}
 				$hooks_before = ob_get_clean();
 
@@ -196,20 +189,6 @@ class AddToCartWithOptions extends AbstractBlock {
 					 * @since 1.5.0
 					 */
 					do_action( 'woocommerce_after_add_to_cart_button' );
-				} elseif ( ProductType::EXTERNAL === $product_type ) {
-					/**
-					 * Hook: woocommerce_after_add_to_cart_button.
-					 *
-					 * @since 1.5.0
-					 */
-					do_action( 'woocommerce_after_add_to_cart_button' );
-				} elseif ( ProductType::GROUPED === $product_type ) {
-					/**
-					 * Hook: woocommerce_after_add_to_cart_button.
-					 *
-					 * @since 1.5.0
-					 */
-					do_action( 'woocommerce_after_add_to_cart_button' );
 				} elseif ( ProductType::VARIABLE === $product_type ) {
 					/**
 					 * Hook: woocommerce_after_variations_form.
@@ -217,7 +196,15 @@ class AddToCartWithOptions extends AbstractBlock {
 					 * @since 2.4.0
 					 */
 					do_action( 'woocommerce_after_variations_form' );
+				} else {
+					/**
+					 * Hook: woocommerce_after_add_to_cart_button.
+					 *
+					 * @since 1.5.0
+					 */
+					do_action( 'woocommerce_after_add_to_cart_button' );
 				}
+				
 				$hooks_after = ob_get_clean();
 			}
 
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
@@ -87,9 +87,9 @@ class AddToCartWithOptions extends AbstractBlock {
 		 * @param boolean $supports Is the type supported
 		 * @param string $product_type The product type 
 		 */
-		$template_path = apply_filters( 'woocommerce_' . $product_type . '_add_to_cart_block_template_part', Package::get_path() . 'templates/' . BlockTemplateUtils::DIRECTORY_NAMES['TEMPLATE_PARTS'] . '/' . $slug . '.html', $product_type );
+		$template_part_path = apply_filters( 'woocommerce_' . $product_type . '_add_to_cart_with_options_block_template_part', Package::get_path() . 'templates/' . BlockTemplateUtils::DIRECTORY_NAMES['TEMPLATE_PARTS'] . '/' . $slug . '.html', $product_type );
 
-		if ( file_exists( $template_path ) ) {
+		if ( file_exists( $template_part_path ) ) {
 
 			$template_part_contents = '';
 			// Determine if we need to load the template part from the DB, the theme or WooCommerce in that order.
@@ -108,7 +108,7 @@ class AddToCartWithOptions extends AbstractBlock {
 			}
 
 			if ( '' === $template_part_contents ) {
-				$template_part_contents = file_get_contents( $template_path ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
+				$template_part_contents = file_get_contents( $template_part_path ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
 			}
 
 			$classes_and_styles = StyleAttributesUtils::get_classes_and_styles_by_attributes( $attributes, array(), array( 'extra_classes' ) );
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
@@ -80,14 +80,18 @@ class AddToCartWithOptions extends AbstractBlock {
 
 		$slug = $product_type . '-product-add-to-cart-with-options';
 
-		/**
-		 * Filter to declare product type's cart block template is supported.
-		 * 
-		 * @since x.x.x
-		 * @param boolean $supports Is the type supported
-		 * @param string $product_type The product type 
-		 */
-		$template_part_path = apply_filters( 'woocommerce_' . $product_type . '_add_to_cart_with_options_block_template_part', Package::get_path() . 'templates/' . BlockTemplateUtils::DIRECTORY_NAMES['TEMPLATE_PARTS'] . '/' . $slug . '.html', $product_type );
+		if ( in_array( $product_type, array( ProductType::SIMPLE, ProductType::EXTERNAL, ProductType::VARIABLE, ProductType::GROUPED ), true ) ) {
+			$template_part_path = Package::get_path() . 'templates/' . BlockTemplateUtils::DIRECTORY_NAMES['TEMPLATE_PARTS'] . '/' . $slug . '.html';
+		} else {
+			/**
+			 * Filter to declare product type's cart block template is supported.
+			 * 
+			 * @since x.x.x
+			 * @param boolean $supports Is the type supported
+			 * @param string $product_type The product type 
+			 */
+			$template_part_path = apply_filters( 'woocommerce_' . $product_type . '_add_to_cart_with_options_block_template_part', Package::get_path() . 'templates/' . BlockTemplateUtils::DIRECTORY_NAMES['TEMPLATE_PARTS'] . '/' . $slug . '.html', $product_type );
+		}
 
 		if ( file_exists( $template_part_path ) ) {
 
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
@@ -90,7 +90,7 @@ class AddToCartWithOptions extends AbstractBlock {
 			 * @param boolean $supports Is the type supported
 			 * @param string $product_type The product type 
 			 */
-			$template_part_path = apply_filters( 'woocommerce_' . $product_type . '_add_to_cart_with_options_block_template_part', false, $product_type );
+			$template_part_path = apply_filters( '__experimental_woocommerce_' . $product_type . '_add_to_cart_with_options_block_template_part', false, $product_type );
 		}
 
 		if ( file_exists( $template_part_path ) ) {
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
@@ -93,7 +93,7 @@ class AddToCartWithOptions extends AbstractBlock {
 			$template_part_path = apply_filters( '__experimental_woocommerce_' . $product_type . '_add_to_cart_with_options_block_template_part', false, $product_type );
 		}
 
-		if ( file_exists( $template_part_path ) ) {
+		if ( is_string( $template_part_path) && file_exists( $template_part_path ) ) {
 
 			$template_part_contents = '';
 			// Determine if we need to load the template part from the DB, the theme or WooCommerce in that order.
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
@@ -86,7 +86,7 @@ class AddToCartWithOptions extends AbstractBlock {
 			/**
 			 * Filter to declare product type's cart block template is supported.
 			 * 
-			 * @since x.x.x
+			 * @since 9.9.0
 			 * @param mixed string|boolean The template part path if it exists
 			 * @param string $product_type The product type 
 			 */
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
@@ -90,7 +90,7 @@ class AddToCartWithOptions extends AbstractBlock {
 			 * @param boolean $supports Is the type supported
 			 * @param string $product_type The product type 
 			 */
-			$template_part_path = apply_filters( 'woocommerce_' . $product_type . '_add_to_cart_with_options_block_template_part', Package::get_path() . 'templates/' . BlockTemplateUtils::DIRECTORY_NAMES['TEMPLATE_PARTS'] . '/' . $slug . '.html', $product_type );
+			$template_part_path = apply_filters( 'woocommerce_' . $product_type . '_add_to_cart_with_options_block_template_part', false, $product_type );
 		}
 
 		if ( file_exists( $template_part_path ) ) {
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
@@ -158,6 +158,20 @@ class AddToCartWithOptions extends AbstractBlock {
 					 * @since 1.5.0
 					 */
 					do_action( 'woocommerce_before_add_to_cart_button' );
+				} elseif ( ProductType::EXTERNAL === $product_type ) {
+					/**
+					 * Hook: woocommerce_before_add_to_cart_button.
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
```diff
@@ -146,27 +158,20 @@ protected function render( $attributes, $content, $block ) {
...
-					do_action( 'woocommerce_before_add_to_cart_button' );
-				} elseif ( ProductType::GROUPED === $product_type ) {
+					do_action( 'woocommerce_before_variations_form' );
+				} else {
 					/**
 					 * Hook: woocommerce_before_add_to_cart_button.
 					 *
 					 * @since 1.5.0
 					 */
 					do_action( 'woocommerce_before_add_to_cart_button' );
```
Aljullu: I wouldn't run this filter for all product types because we can't anticipate if they, in fact, have a product button. If an extension wants this hook or any other to run before/after their template part, they can do it manually (they can use `render_block` or a similar hook). Otherwise, removing the hook for extension that don't want it to run becomes much harder.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
```diff
@@ -78,11 +78,23 @@ protected function render( $attributes, $content, $block ) {
...
+		$slug = $product_type . '-product-add-to-cart-with-options';
+
+		/**
+		 * Filter to declare product type's cart block template is supported.
+		 * 
+		 * @since x.x.x
+		 * @param boolean $supports Is the type supported
+		 * @param string $product_type The product type 
+		 */
+		$template_path = apply_filters( 'woocommerce_' . $product_type . '_add_to_cart_block_template_part', Package::get_path() . 'templates/' . BlockTemplateUtils::DIRECTORY_NAMES['TEMPLATE_PARTS'] . '/' . $slug . '.html', $product_type );
```
Aljullu: I would restrict this filter to custom product types, only. What do you think?

For WC core product types, themes can use the default paths. And we would like to prevent extensions from modifying the path to avoid them breaking themes.

So something like this:

```PHP
if ( in_array( $product_type, array( ProductType::SIMPLE, ProductType::EXTERNAL, ProductType::VARIABLE, ProductType::GROUPED ), true ) ) {
	$template_part_path = Package::get_path() . 'templates/' . BlockTemplateUtils::DIRECTORY_NAMES['TEMPLATE_PARTS'] . '/' . $slug . '.html';
} else {
	/**
	 * Filter to get the template part path for a product type's Add to Cart with Options block.
	 * Returns the path to the template file for the given product type
	 * or false if no template part is supported.
	 *
	 * @since x.x.x
	 * @param string $template_part_path The path to the template file for the given product type.
	 */
	$template_part_path = apply_filters( 'woocommerce_' . $product_type . '_add_to_cart_with_options_block_template_part', false );
}
```

I would also rename the filter to `woocommerce_<product_type>_add_to_cart_with_options_block_template_part`. I know it's very long, but given that everywhere we use this name for the block, it makes sense to me to keep it the same in the filter name.

Aljullu (commented):
Thanks for opening this PR and for the taco plugin to make it easy to test! :raised_hands: 

I left a couple of suggestions, looking forward to knowing what you think.

Besides that, there will be two follow-up tasks as part of https://github.com/woocommerce/woocommerce/issues/55701:

1. We will need to add [automated tests](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/tests/php/src/Blocks/BlockTypes/AddToCartWithOptions.php).
2. We will need to register the template parts (probably using the [BlockTemplatesRegistry](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Blocks/BlockTemplatesRegistry.php#L67)) to make sure they show up in the Site Editor.

We can handle them in separate PRs, so not blockers. :slightly_smiling_face:

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
helgatheviking: Ok, that makes sense mostly. I guess I, myself, don't show the add to cart button if my type isn't purchasable. 

I also ran into issues [here](https://github.com/woocommerce/woocommerce/pull/56799/files#diff-cda6779a411f29de14b217e7ac5f8e1ddd839d7600f9d48edcaa6e9202531ad1R237) with core calling functions I didn't want (that's why I wrapped it in that exists check) so don't want to make things hard to remove.

So should I revert this entire chunk?

Aljullu: Oh, right. That line

```PHP
add_action( 'woocommerce_' . $product_type . '_add_to_cart', 'woocommerce_' . $product_type . '_add_to_cart', 30 );
```

was assuming it was a WC core product type so the function would always exist. With your changes, custom product types might reach this line too, so the check you added looks like a good protection to me. Another option would be to wrap everything in an if:

```PHP
if ( in_array( $product_type, array( ProductType::SIMPLE, ProductType::EXTERNAL, ProductType::VARIABLE, ProductType::GROUPED ), true ) ) {
	$add_to_cart_fn = 'woocommerce_' . $product_type . '_add_to_cart';
	remove_action( 'woocommerce_' . $product_type . '_add_to_cart', $add_to_cart_fn, 30 );

	/**
	 * Trigger the single product add to cart action that prints the markup.
	 *
	 * @since 9.9.0
	 */
	do_action( 'woocommerce_' . $product->get_type() . '_add_to_cart' );
	add_action( 'woocommerce_' . $product_type . '_add_to_cart', $add_to_cart_fn, 30 );
}
```

I think if a custom product type has registered a custom template part, there is no point in running `woocommerce_<product_type>_add_to_cart`, no? :thinking:

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
helgatheviking: Well as a plugin dev I am probably going to always come down on the side of "extend all the things" :) In the past I have needed to modify template parts from another plugin, though not the entire add to cart template. But we could always start with your proposal and then switch if someone comes in with a need to override on of the core templates.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
helgatheviking: >I think if a custom product type has registered a custom template part, there is no point in running `woocommerce_<product_type>_add_to_cart`, no?

I would agree. I didn't entirely get why you were calling that hook at all. I do use that hook (currently) to register some scripts specific to product types, but for block support I would expect to use new approaches.

Aljullu: We are using in WC core product types because some extensions are using them to add extra content, ie when a product is out of stock. But in the case of product types, I think it shouldn't be needed.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
Aljullu: Good points, I see what you mean. I would lean on the conservative side for now, thought. I think using the inner PHP hooks or inner blocks will be a "safer way" to extend the form than completely replacing the template part path. In the future, we can revisit this decision if there is enough demand for adding this filter for core product types.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
helgatheviking: If I look at my 2 plugins: Mix and Match and Name Your Price... a mix and match container bundle can be name your price enabled... and the nyp price input would appear on the `woocommerce_before_add_to_cart_button` so I would still need a way to enable the compatibility layer for my custom product type. Should I handle that with a compat layer filter on my side?

Aljullu: I don't have a strong opinion yet, but I was leaning towards extensions handling the compatibility layer on their own because it gives them more flexibility about when to enable or disable it. For example, for simple products we want to disable it when the product is out of stock or not purchasable (https://github.com/woocommerce/woocommerce/issues/56682). I would expect extensions introducing custom product types might have similar needs, that's why I thought it would make sense to let extensions handle the compatibility layer themselves.

Another option would be to introduce an extra hook that allows extensions to decide whether the compatibility layer is enabled or disabled for certain product types. :thinking: 

What do you think?

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
helgatheviking: ok, pushed some updates :)

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
```diff
@@ -78,11 +78,27 @@ protected function render( $attributes, $content, $block ) {
...
+			$template_part_path = Package::get_path() . 'templates/' . BlockTemplateUtils::DIRECTORY_NAMES['TEMPLATE_PARTS'] . '/' . $slug . '.html';
+		} else {
+			/**
+			 * Filter to declare product type's cart block template is supported.
+			 * 
+			 * @since x.x.x
+			 * @param boolean $supports Is the type supported
+			 * @param string $product_type The product type 
+			 */
+			$template_part_path = apply_filters( 'woocommerce_' . $product_type . '_add_to_cart_with_options_block_template_part', Package::get_path() . 'templates/' . BlockTemplateUtils::DIRECTORY_NAMES['TEMPLATE_PARTS'] . '/' . $slug . '.html', $product_type );
```
Aljullu: We know the default path (`Package::get_path() . 'templates/' . BlockTemplateUtils::DIRECTORY_NAMES['TEMPLATE_PARTS'] . '/' . $slug . '.html'`) will never return a file because WC core won't include templates for 3rd-party product types. So I think we could simply do:

```suggestion
			$template_part_path = apply_filters( '__experimental_woocommerce_' . $product_type . '_add_to_cart_with_options_block_template_part', false, $product_type );
```

I would also like to prefix the hook as `__experimental` for now. At least until we get it working in the editor side. :slightly_smiling_face:

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
```diff
@@ -78,11 +78,27 @@ protected function render( $attributes, $content, $block ) {
...
+		$slug = $product_type . '-product-add-to-cart-with-options';
+
 		if ( in_array( $product_type, array( ProductType::SIMPLE, ProductType::EXTERNAL, ProductType::VARIABLE, ProductType::GROUPED ), true ) ) {
+			$template_part_path = Package::get_path() . 'templates/' . BlockTemplateUtils::DIRECTORY_NAMES['TEMPLATE_PARTS'] . '/' . $slug . '.html';
+		} else {
+			/**
+			 * Filter to declare product type's cart block template is supported.
+			 * 
+			 * @since x.x.x
+			 * @param boolean $supports Is the type supported
```
Aljullu: I guess this `@param` needs to be updated? :slightly_smiling_face: Same for the `@since` version.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
```diff
@@ -225,14 +241,15 @@ protected function render( $attributes, $content, $block ) {
...
-			add_action( 'woocommerce_' . $product_type . '_add_to_cart', 'woocommerce_' . $product_type . '_add_to_cart', 30 );
+			if ( in_array( $product_type, array( ProductType::SIMPLE, ProductType::EXTERNAL, ProductType::VARIABLE, ProductType::GROUPED ), true ) ) {
+				$add_to_cart_fn = 'woocommerce_' . $product_type . '_add_to_cart';
+				/**
+				 * Trigger the single product add to cart action that prints the markup.
+				 *
+				 * @since 9.9.0
+				 */
+				do_action( 'woocommerce_' . $product->get_type() . '_add_to_cart' );
+			}
```
Aljullu: We need to keep the `remove_action()` and `add_action()` so we don't run the WC core actions when rendering this block. Otherwise, the form appears twice:

imatge

Aljullu (commented):
Thanks for the updates, @helgatheviking! This is looking great. I left some minor additional comments inline.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
helgatheviking: What `@since` should we use?

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
```diff
@@ -78,11 +78,27 @@ protected function render( $attributes, $content, $block ) {
...
+			 * Filter to declare product type's cart block template is supported.
+			 * 
+			 * @since 9.9.0
+			 * @param mixed string|boolean The template part path if it exists
+			 * @param string $product_type The product type 
+			 */
+			$template_part_path = apply_filters( '__experimental_woocommerce_' . $product_type . '_add_to_cart_with_options_block_template_part', false, $product_type );
+		}
+
+		if ( file_exists( $template_part_path ) ) {
```
Aljullu: We should probably add a check to make sure `$template_part_path`is a string. Otherwise, it might be `false` and cause an error with `file_exists( false )`.

Aljullu (commented):
Thanks for the updates, @helgatheviking! This is looking good on my end!

In order to merge, there are some things pending:

* There needs to be a changelog file (it can be created with `pnpm --filter='@woocommerce/plugin-woocommerce' changelog add`).
* I see there are also some minor [linting issues](https://github.com/woocommerce/woocommerce/actions/runs/14495438094/job/40715208076?pr=56799).
* I wonder if you could update the testing steps (and maybe the sample plugin)? For example, [`woocommerce_taco_add_to_cart_block_template_part`](https://gist.github.com/helgatheviking/f49ebaf44b1fe70b4ab8e993336491f9#file-wc-blockified-add-to-cart-test-php-L65) would need to be `__experimental_woocommerce_taco_add_to_cart_block_template_part`.

Thanks in advance!

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
helgatheviking: How can we confirm this? Granted, I am only quick testing at `https://onlinephp.io/` but `file_exists(false)` isn't throwing any errors with php 7+ (or even 5.6).

helgatheviking: Never mind... ran it locally and boom fatal error. :sweat_smile: Fix incoming.

helgatheviking:
@Aljullu I don't know if it's directly related to this PR, but I did just notice [this bit about `do_block()` not relaying context](https://github.com/woocommerce/woocommerce/blob/ad5cc8d5e8f010d670e6c80bc75db3dc57ce0f55/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php#L233-L237) since I was reworking some of my own code and running into the same problem. 

Core is manually calling `set_is_descendant_of_add_to_cart_with_options_context` just before `do_blocks` and removing it immediately after, but I think we'll need a more consistent way for other blocks to do the same so context can flow from the `add-to-cart-with-options` block to any child blocks.

Aljullu (commented):
Great job, @helgatheviking! Thanks for handling this task. Everything looks good on my end now, and it's testing correctly.

The testing steps I followed:

**To enable the block:**

2. Go to Tools WA Test Helper > Features and enable `experimental-blocks` and `blockified-add-to-cart`.
3. Edit the Single Product template, click on the _Add to Cart with Options_ block and upgrade to the blockified version in the sidebar.

**To create a taco product:***

1. Create a `taco` folder in `/wp-content/plugins/`.
2. Paste the three files from your sample repo.
3. In wp-admin, enable the plugin.
4. Create a new product with the taco type.
5. In the frontend, verify the correct template is loaded:

imatge

@helgatheviking I see there is one last linting error. Would you be able to fix it?

<details>

* Theme: Twenty Twenty-Five
* Store size: small
</details>

Aljullu:
> @Aljullu I don't know if it's directly related to this PR, but I did just notice [this bit about `do_block()` not relaying context](https://github.com/woocommerce/woocommerce/blob/ad5cc8d5e8f010d670e6c80bc75db3dc57ce0f55/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php#L233-L237) since I was reworking some of my own code and running into the same problem. 
> 
> Core is manually calling `set_is_descendant_of_add_to_cart_with_options_context` just before `do_blocks` and removing it immediately after, but I think we'll need a more consistent way for other blocks to do the same so context can flow from the `add-to-cart-with-options` block to any child blocks.

Right, that's a limitation in Gutenberg, we don't have a way to "pass down" the context to blocks rendered inside a template part. Out of curiosity, what's the use-case for this?

I hadn't tried it, but one solution might be to use the same approach we take and hook the context before rendering the _Add to Cart with Options_ block, and removing it after render. You could use the [`render_block_{$this->name}` filter](https://developer.wordpress.org/reference/hooks/render_block_this-name/) for this.

helgatheviking:
@Aljullu In Mix and Match I will have a grid of products (or table) similar to the Group Products Selector with Selector Item template. Inside the child product the product/post ID context is the _current_ child product's ID, but I also need a way to reference the original parent product ID. I had been setting it as a new context `containerId`. 

First time through, I inserted my own MNM block that had all my child blocks : selections/options, prompt, reset, status. I dabbled with removing my wrapping block and trying to use the add to cart with options block but yeah ran into the lost context. It would be nice if the add to cart with options' available context was automatically available, but I am not sure how to do that.

helgatheviking:
Woo could automatically flow any context with something like the following, what do you think?

Replace [these lines](https://github.com/woocommerce/woocommerce/blob/f850b2a35f926cdf69845f0eaec54e590475ad88/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php#L228-L232)

```
// Because we are printing the template part using do_blocks, context from the outside is lost.
// This filter is used to add the isDescendantOfAddToCartWithOptions context back.
add_filter( 'render_block_context', array( $this, 'set_is_descendant_of_add_to_cart_with_options_context' ), 10, 2 );
$template_part_blocks = do_blocks( $template_part_contents );
remove_filter( 'render_block_context', array( $this, 'set_is_descendant_of_add_to_cart_with_options_context' ) );
```

with

```

/**
 * Because we are printing the template part using do_blocks, context from the outside is lost.
 * The following is a workaround to persist the context for any child block.
 */
$filter_block_context = static function ( $context ) use ( $block ) {
	$context['woocommerce/isDescendantOfAddToCartWithOptions'] = true;
	return array_merge( $block->context, $context );
};

add_filter( 'render_block_context', $filter_block_context );

$template_part_blocks = do_blocks( $template_part_contents );

remove_filter( 'render_block_context', $filter_block_context );
```

Update: that doesn't quite work as I had hoped. I'm struggling to get the child items to set a _new_ context for hte `postId` when doing this. :thinking:

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
```diff
@@ -78,11 +78,27 @@ protected function render( $attributes, $content, $block ) {
...
+			 * Filter to declare product type's cart block template is supported.
+			 *
+			 * @since 9.9.0
+			 * @param mixed string|boolean The template part path if it exists
+			 * @param string $product_type The product type
+			 */
+			$template_part_path = apply_filters( '__experimental_woocommerce_' . $product_type . '_add_to_cart_with_options_block_template_part', false, $product_type );
+		}
+
+		if ( is_string( $template_part_path) && file_exists( $template_part_path ) ) {
```
helgatheviking: @Aljullu This is the lint error 

>96 | ERROR | Expected 1 spaces before closing parenthesis; 0 found (PEAR.Functions.FunctionCallSignature.SpaceBeforeCloseBracket)

I don't understand it though... it's saying there's no space before the closing `)` ? when there definitely looks to be one.

helgatheviking: NM... I see the missing space is in the `is_string()` check

helgatheviking (commented):
linting fixed

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
helgatheviking: I'm partial to core handling things to make them more universal, but as long as it's clear that we have to handle the compat layer from our side, it should be relatively straightforward to do.

Aljullu:
> @Aljullu In Mix and Match I will have a grid of products (or table) similar to the Group Products Selector with Selector Item template. Inside the child product the product/post ID context is the _current_ child product's ID, but I also need a way to reference the original parent product ID. I had been setting it as a new context `containerId`. 

I might not be fully understanding, but could you use the `global $product`? While it's not as clean as the block context, we use it in several places, so the `$product` is always the product data being rendered (either the parent or the child product data). 

For example, in this block, we set the global `$post` and `$product` to make sure children products have the correct globals. That's useful because some Gutenberg blocks like _Product Title_/_Post Title_ read from the global.

https://github.com/woocommerce/woocommerce/blob/381d06e0cd5368938c5f61c0e28ba0a763fb8ec7/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsGroupedProductSelectorItemTemplate.php#L32-L40

And in this block we try to get the product ID from the context, but if it's not available, we fall back to the global:

https://github.com/woocommerce/woocommerce/blob/381d06e0cd5368938c5f61c0e28ba0a763fb8ec7/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsGroupedProductSelectorItemCTA.php#L109-L117

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
Aljullu: Makes sense. I would go with this route for now. I think with good documentation it shouldn't be too difficult for extensions to do, and it allows more flexibility. But if we see this being problematic, we can revisit this approach. Does it sound good?

Aljullu:
Btw, all checks are passing, @helgatheviking. Do you think we should go ahead and merge this PR?

helgatheviking:
The global product becomes the current product object when you are inside that loop of child products though. I need to always know the current product _and_ the parent (so the grouped product). I've been having trouble getting the parent product so that why I started passing the parent's product ID as a persistent context. 

And yeah, I think we look good to merge! :partying_face:

Aljullu:
> The global product becomes the current product object when you are inside that loop of child products though. I need to always know the current product and the parent (so the grouped product). I've been having trouble getting the parent product so that why I started passing the parent's product ID as a persistent context.

Gotcha. I don't have a clear idea of how to solve it, my understanding is that the blocks structure will be something like this:

```
Single Product template
--- (other blocks)
--- Add to Cart with Options
------ Mix and Match wrapper block
--------- Product grid / table
--- (other blocks)
```

Unfortunately, the context from _Add to Cart with Options_ is lost in _Mix and Match wrapper block_. But if you only need the parent product data, it will be in `global $product`. So you can save it in the _Mix and Match wrapper block_ context and access it from the descendants. Would that work?

Aljullu (approved):

Aljullu:
> And yeah, I think we look good to merge! 🥳

Merged! Thanks for your help with this, @helgatheviking! :raised_hands:

PR #56816, opened 2025-03-28
Enable shop managers to create customers via REST

### LINKED ISSUE #56343: Allow shop managers to create new customer user accounts
Right now, the Shop Manager role is limited in what it can do with users because of security implications. It can edit/modify user details _if_ the user role is customer or subscriber (with a filter to allow other roles as well), but it cannot create new users. This presents some limitations, particularly for apps utilizing the REST API, where you may want to create a new account after a customer has made a purchase.

The proposed solution for this is to create a new meta capability, `create_customers`, which would be assigned to the Shop Manager role. This capability could be checked within the REST API context or any other context to allow a shop manager to create a new customer user account, without introducing all the other implications that come along with WP's built-in `create_users` capability.

coreymckrill:

Before this, shop managers could update data on existing customer user accounts, but could not create new customer accounts. This PR changes that, by introducing a new capability, `create_customers`, that is assigned to both shop manager and administrator roles. That capability is checked in the `wc_rest_check_user_permissions` function, so that shop managers can create users via REST API requests. However, other work will need to be done to allow shop managers to create customers in other contexts, such as the WP Admin web UI.

This also updates several unit tests to cover permission checks for various customer-related REST operations using different roles (admin, shop manager, editor).

Fixes #56343

1. Set up your test site with a user with the Shop Manager role.
2. Give that user access to the WC REST API by generating either an API key with read/write, or an app password.
3. Using your API client of choice (my current favorite is [Yaak](https://yaak.app/)), send a request to create a new customer using your shop manager credentials
    ```
    curl -X POST 'http://localhost:8888/wp-json/wc/v3/customers' \
      --header 'Content-Type: multipart/form-data' \
      --form 'context=edit' \
      --form 'email=test@example.com' \
      --user '{key or username}:{secret or app password}'
    ```
4. The request should succeed and return a new user object. Note the new user should have a `role` property with a value of `customer`.
    * If you attempt this on trunk, this request should fail with a permission error message.
5. Try to do nefarious things like create a new user with the administrator role, with something like `--form 'role=administrator'` or `--form 'roles=administrator'`. The request should still succeed, but you should see that the customer record created still has a role of `customer` rather than whatever you specified. (This is the same behavior that already existed if an admin user tried to create an admin account using the customer endpoint.)
6. The changes in this PR should only affect the ability of a shop manager to _create_ a customer account. Other functionality, like updating existing customer data (should succeed) or deleting an existing customer (should fail) should not be changed by this.

* During development, testing was done using a wp-env test site running WP 6.7 and PHP 8.1.
* Unit tests were expanded to cover the case introduced in this PR, of shop managers being able to create new customer accounts, but not accounts with other roles.


albarin (approved):
- Confirmed shop managers can now create customers.
- Confirmed they cannot create users with roles other than customer.
- Confirmed they can update customers but not delete.

PR #56866, opened 2025-03-31
[dev] CI: PHP 8.2 environment for core PHP-tests.

kalessil:

- PHP 8.2 environment for core PHP testing jobs.

- CI-checks shall pass

### DIFF
--- a/plugins/woocommerce/package.json
+++ b/plugins/woocommerce/package.json
@@ -109,7 +109,7 @@
 			},
 			"tests": [
 				{
-					"name": "PHP: 8.1 WP: latest",
+					"name": "PHP: 8.2 WP: latest",
 					"testType": "unit:php",
 					"command": "test:php:env",
 					"shardingArguments": [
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/package.json
```diff
@@ -128,14 +128,14 @@
...
-							"phpVersion": "8.1",
+							"phpVersion": "8.2",
 							"wpVersion": "latest"
 						}
 					},
 					"events": [
 						"pull_request",
 						"push",
-						"php-unit-php-8.8-wp-latest"
+						"php-unit-php-8.2-wp-latest"
```
kalessil, line 138: @adimoldovan: not sure about this event, should I update it somewhere else (search points to no original value)?

adimoldovan: Nope, it's not referenced anywhere. It only exists for convenience, in case one wants to run only this job on demand.

kalessil: Ok, just updating it then.

PR #56879, opened 2025-03-31
Update method for generating order months dropdown

coreymckrill:

In #55510 we changed how the Months filter dropdown on the Orders list table is generated so that it uses a much more efficient query. However, it introduced a bug that could happen at the end of the month when the following month had fewer days in it. This was surfaced via a unit test, fortunately. In this PR we are making the method more robust so that the number of days in the month doesn't matter, and also improving the behavior during an edge case where all existing orders are in the future, so that the _furthest_ future order date is represented in the dropdown, and not just the _closest_ future date.

Bug introduced in PR #55510

It probably isn't realistic to try and spoof the current date on your local dev environment in order to test this. Instead, just follow the testing instructions in #55510 and make sure there aren't any regressions. Also ensure the unit tests pass.

* Tested in a wp-env environment running WP 6.7 and PHP 8.1
* Unit test that caught the failure was updated to take into account the changes in this PR. All unit tests in the file are passing in the local environment.

### DIFF
--- a/plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
+++ b/plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
@@ -85,7 +85,7 @@ class ListTable extends WP_List_Table {
 	 * Init method, invoked by DI container.
 	 *
 	 * @internal This method is not intended to be used directly (except for testing).
-	 * @param PageController      $page_controller Page controller instance for this request.
+	 * @param PageController $page_controller Page controller instance for this request.
 	 */
 	final public function init( PageController $page_controller ) {
 		$this->page_controller = $page_controller;
@@ -861,7 +861,7 @@ class ListTable extends WP_List_Table {
 		if ( $start > $end ) {
 			$last_year_month_gmt = $wpdb->get_row(
 				$wpdb->prepare(
-				'
+					'
 					SELECT YEAR( t.date_created_gmt ) AS year,
 					       MONTH( t.date_created_gmt ) AS month
 					FROM %i t
--- a/plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
+++ b/plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
@@ -808,7 +808,7 @@ class ListTable extends WP_List_Table {
 	 *
 	 * @return \stdClass[]
 	 */
-	public function get_months_filter_options(): array {
+	protected function get_months_filter_options(): array {
 		global $wpdb;
 
 		$orders_table   = esc_sql( OrdersTableDataStore::get_orders_table_name() );
--- a/plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
+++ b/plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
@@ -7,6 +7,7 @@ use Automattic\WooCommerce\Internal\DataStores\Orders\CustomOrdersTableControlle
 use Automattic\WooCommerce\Internal\DataStores\Orders\OrdersTableDataStore;
 use Automattic\WooCommerce\Caches\OrderCountCache;
 use Automattic\WooCommerce\Utilities\OrderUtil;
+use DatePeriod;
 use WC_Order;
 use WP_List_Table;
 use WP_Screen;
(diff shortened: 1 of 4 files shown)


plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
```diff
@@ -811,25 +812,33 @@ private function months_filter() {
...
 
 		$first_year_month_gmt = $wpdb->get_row(
-			// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
 			$wpdb->prepare(
-				"
+				'
 					SELECT YEAR( t.date_created_gmt ) AS year,
 					       MONTH( t.date_created_gmt ) AS month
-					FROM $orders_table t
+					FROM %i t
```
jorgeatorres: `%i` sadly doesn't work with WPCOM's wpdb, so I guess we shouldn't rely on it.

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
```diff
@@ -842,16 +851,41 @@ protected function get_months_filter_options(): array {
...
-			$start = new \WC_DateTime( 'now', wp_timezone() );
+			$start = $this_month;
 		}
 
-		$end     = new \WC_DateTime( 'now', wp_timezone() );
+		$end     = $this_month;
 		$options = array();
 
-		// If, somehow, the oldest order date is in the future, swap the start and end of the range.
+		// If, somehow, the oldest order date is in the future, we need to find the order furthest into the future as well.
```
jorgeatorres: I believe we'll need the information about the order furthest into the future _regardless_ of whether the the oldest order is in the future or not.

For example, let's say you have 3 orders: one in the past (say, 2024-12-25), and two in the future (2025-12-25) and (2026-12-25). We would detect that the oldest one is in fact _not_ in the future and so we'll miss adding year/month options for the other orders.

I believe the correct solution would be to fetch both the dates of the oldest and newest order and make sure that `$first_year_month` and `$last_year_month` satisfy this (pseudo-code):

```
$first_year_month = $oldest_order_year_month; // alternatively min( $today_year_month, $oldest_order_year_month) if we want to include the current year/month and upwards.

$last_year_month  = max( $today_year_month, $newest_order_year_month );
```

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
```diff
@@ -811,25 +812,33 @@ private function months_filter() {
 	protected function get_months_filter_options(): array {
 		global $wpdb;
 
+		$this_month = new \WC_DateTime(
+			sprintf(
+				'%s-%s-01',
+				gmdate( 'Y' ),
+				gmdate( 'm' )
```
jorgeatorres: Given `gmdate()` returns time in UTC, are we sure that `$this_month` is always correct?

Here's my thinking: for simplicity, let's assume there are no orders and let's say current time is 2025-12-31 23:00:00 UTC. `gmdate()` would return `2025` as year and `12` as month, so `$this_month` gets set to `2025-12-01` (with the assumption that the date is in my timezone). This, in turn, would result in the option `2025/12` being added to the dropdown, but if my timezone is, say, UTC+2, the option we need is actually `2026/01`.

I might be missing something (timezones are tricky!) so apologies if that's the case.

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
```diff
@@ -842,16 +851,41 @@ protected function get_months_filter_options(): array {
...
+			);
+
+			$end = new \WC_DateTime(
+				sprintf(
+					'%s-%s-01',
+					$last_year_month_gmt->year,
+					$last_year_month_gmt->month
+				)
+			);
+			$end->setTimezone( wp_timezone() ); // Adjust date and time to reflect site timezone.
```
jorgeatorres: Maybe similar to the above? I feel we should take the time into account too.

jorgeatorres (changes requested):
Hey @coreymckrill!

Thanks for the quick PR 💯. I left some feedback around some possible edge cases (and a WPCOM issue). Please let me know what you think.

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
coreymckrill: Dangit, I should have known there was a reason we weren't already using it. Fixed in 5c9aa738df

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
coreymckrill: This is a great point. In 5c9aa738df I simplified everything so that we always get the min and max, and then if one end of the range or the other does not include the current month, use that instead. Hopefully this makes things more robust and also avoids weird edge date pitfalls with the unit tests. But I'm not entirely sure about the performance implications of the new query. I'll ping Joshua about it in a separate comment.

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
coreymckrill: I changed this in 5c9aa738df, but just to clarify, if you create a DateTime object in UTC and then afterward use the setTimezone method to change to the site's timezone, it will adjust the date/time forwards or backwards as necessary. But if you create a DateTime object and include the timezone as the second parameter, it assumes that the date/time you're giving it is already in that timezone. So what I had above, using gmdate, but then also setting the timezone, was probably wrong.

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
```diff
@@ -808,69 +808,83 @@ private function months_filter() {
...
 			$wpdb->prepare(
 				"
-					SELECT YEAR( t.date_created_gmt ) AS year,
-					       MONTH( t.date_created_gmt ) AS month
-					FROM $orders_table t
+					SELECT MIN( t.date_created_gmt ) as min_date_gmt,
+					       MAX( t.date_created_gmt ) as max_date_gmt
+					FROM `{$orders_table}` t
 					WHERE type = %s
 					AND status != %s
```
coreymckrill, line 823: @joshuatf do you have thoughts about the performance of this query? Or, do you still have a site available with a high order count that you could use to test this with? If not, I can try to spin something up. (This is a further revision of the work done in #55510)

jorgeatorres (approved):
LGTM ❤️ .

Issue #50714, opened 2024-08-15
[Enhancement]: Improvements in TTFB


joshuatf:
### Describe the solution you'd like

With WooCommerce enabled, we see an increase in almost 2x of TTFB for frontend pages.  This issue involves looking into the causes of those increases and gathering possible solutions to decrease that time.

While the primary goal of this spike is to reduce the overall time, things to look for in this spike include:

* Code that is being run indiscriminately on every page
* Unoptimized or poor performing code
* Packages that significantly add to the overall time
* Hooks, methods, or classes that are common culprits and should be used sparingly

### With WooCommerce

<img width="834" alt="Screenshot 2024-08-15 at 1 41 42 PM" src="https://github.com/user-attachments/assets/6d260cb7-bf04-436c-b0f3-cb471f3af4b7">

### Without WooCommerce

<img width="834" alt="Screenshot 2024-08-15 at 1 42 29 PM" src="https://github.com/user-attachments/assets/8e05f797-7432-43ea-9d0c-c979d7cbb4a1">

### Additional context

Originally documented in p55Cj4-3m3-p2.

This post notes that the median values for all Atomic sites was consistent with the 2x TTFB times seen above.  These statistics were limited to logged out users on the frontend homepage of sites.

joshuatf:
## TTFB Analysis

To analyze the performance of the page, I used a couple different techniques:

1. Xdebug profiling with Qcachegrind to visualize the data.
2. TTFB response time with cURL (`time_starttransfer`), using the median response time over 100 requests.  Packages were then manually included or excluded from the app to demonstrate their effect on performance.

Each test used the exact same criteria that follows.  This is admittedly a narrow view on the total performance, but more closely matches the original environment where this issue was originally reported.

* Non-logged in user homepage request
* Only the WooCommerce plugin
* Twenty Twenty-Four theme
* 20 products
* Default homepage template

All times below are in milliseconds and any percentages represent a difference in time from the baseline (no plugins active).
### Initial findings

Below were the result from the initial tests, but something isn't quite right.  How is it possible that cURL was reporting only `21ms` when WooCommerce is active, but WordPress alone takes `30ms`?

| Environment                                 | Fastest  | Slowest  | Median   |
| ------------------------------------------- | -------- | -------- | -------- |
| Baseline (no plugins)                       | 28.532ms | 74.118ms | 29.072ms |
| WooCommerce Active                          | 45.569ms | 90.788ms | 46.907ms |
[...]

PR #56888, opened 2025-04-01
 Replace AssetDataRegistry usage in product filters with wp_interactivity_config

samueljseay:

1. Introduce a new class with static methods that will allow adding shared config under the `woocommerce` namespace, that can be accessed via `@wordpress/interactivity`'s `getConfig`
2. Migrate usage of `getSetting`/`AssetDataRegistry` in product filters to use wp_interactivity_config
3. Some small clean ups.

N/A

This will require some smoke testing of the product filters:

0. Add a product collection to a post or page, then nest the Product Filters block in it.

1. Especially ensure that price filter functions as it did before (ie displaying formatted prices in the min and max fields)
2. Ensure that when any filter is updated it navigates as before
3. Look in dev console to ensure there are no unexpected console errors

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/inner-blocks/attribute-filter/inspector.tsx
```diff
@@ -27,7 +27,7 @@ import {
 import { sortOrderOptions } from './constants';
 import { BlockAttributes, EditProps } from './types';
 import { getAttributeFromId } from './utils';
-import { getInnerBlockByName } from '../../utils';
+import { getInnerBlockByName } from '../../utils/get-inner-block-by-name';
```
samueljseay, line 30: I removed the index file of exports, which is not ideal when mixing some utils used both frontend and backend.

plugins/woocommerce/src/Blocks/BlockTypes/ClassicTemplate.php
```diff
@@ -165,7 +171,7 @@ protected function render_order_received() {
 
 		echo '<div class="wp-block-group">';
 
-		echo sprintf(
+		printf(
```
samueljseay, line 172: phpcs auto-formatting is doing this.

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilters.php
```diff
@@ -35,17 +35,18 @@ protected function enqueue_data( array $attributes = array() ) {
 		global $pagenow;
 		parent::enqueue_data( $attributes );
 
-		$this->asset_data_registry->add( 'isBlockTheme', wp_is_block_theme() );
-		$this->asset_data_registry->add( 'isProductArchive', is_shop() || is_product_taxonomy() );
-		$this->asset_data_registry->add( 'isSiteEditor', 'site-editor.php' === $pagenow );
```
samueljseay, line 40: isSiteEditor and isWidgetEditor do not appear to be used, is it ok to remove them?

plugins/woocommerce/src/Blocks/SharedInteractivityConfig.php
dinhtungdu, line 1: This class looks like a utility class for me, so I think it should live under Utils directory. Maybe with another name, like a more general `InteractivityUtils`.

dinhtungdu: If you agree that we should rename the class, then we'll also need to rename some method, like `init` to `initSharedConfig`, and `add` to `addSharedConfig`.

plugins/woocommerce/src/Blocks/Domain/Bootstrap.php
```diff
@@ -141,6 +141,9 @@ function () {
 			$this->container->get( Installer::class )->init();
 			$this->container->get( GoogleAnalytics::class )->init();
 			$this->container->get( is_admin() ? CheckoutFieldsAdmin::class : CheckoutFieldsFrontend::class )->init();
+
+			// Note that DI is not needed for one-time initialization of shared settings.
+			add_action( 'init', [ SharedInteractivityConfig::class, 'init' ] );
```
dinhtungdu: By doing this, we're always initializing the config, even if the page doesn't have any interactive block. Wdyt about calling it in the `AbstractInteractivityAPIBlock::render_callback()`?

plugins/woocommerce/src/Blocks/BlockTypes/ClassicTemplate.php
```diff
@@ -53,6 +54,11 @@ protected function enqueue_data( array $attributes = [] ) {
 		// Indicate to interactivity powered components that this block is on the page,
 		// and needs refresh to update data.
 		$this->asset_data_registry->add( 'needsRefreshForInteractivityAPI', true );
```
dinhtungdu: It looks safe to remove this line for me.

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilters.php
dinhtungdu: They were used in the past, but logic using them got removed I think. It's safe to remove them for me, too.

plugins/woocommerce/src/Blocks/BlockTypes/ProductQuery.php
```diff
@@ -184,6 +184,11 @@ public function update_query( $pre_render, $parsed_block ) {
 				'needsRefreshForInteractivityAPI',
```
dinhtungdu: Same as my comment above, I think we can remove this setting.

dinhtungdu (commented):
I smoked test the Product Filters block and didn't found any regression except one with rating filter that reported in the umbrella PR:

image

Beside review comments, this is looking good to me. Thanks for working on this!

plugins/woocommerce/src/Blocks/SharedInteractivityConfig.php
samueljseay: @dinhtungdu yeah sure. I am happy to rename. :)

plugins/woocommerce/src/Blocks/Domain/Bootstrap.php
samueljseay: @dinhtungdu Originally I had it in render callback, and it's totally open to discussion because I can see upside and downside.

if it's here, it's only ever called exactly once so no redundant method calls to `init`. Probably small for performance though.

Also it allows 3PDs to use the config regardless of if they use our block classes, and I think that's more consistent with how `AssetDataRegistry` behaved? 

All that said I think your point is just as good too. I felt that either could be argued, but I'm happy to lean towards your preference here.

dinhtungdu: > if it's here, it's only ever called exactly once so no redundant method calls to init. Probably small for performance though.

We short circuit if the method is called, so the cost of redundant calls is a boolean check, which is cheap. But if we initialize the config every time, the execution time for initialization is bigger than the boolean check. So performance wise, I still think only initializing the config when needed is better. Not to mention that the config may grow in the future and many Woo sites aren't using block themes or interactive blocks.

> Also it allows 3PDs to use the config regardless of if they use our block classes, and I think that's more consistent with how `AssetDataRegistry` behaved?

Do you mean `wc-settings`? If yes, it's only initialized and enqueued when a block needs it. Currently, Cart, Checkout, and Mini Cart use it on the front end.
About 3PDs, we need to think how we design the extensibility (and document it). But for now, with the public util class, they can still init the config if needed.

Comparing between site performance and developer experience, I think the former has more weight. We're doing everything we can to reduce TTFB in https://github.com/woocommerce/woocommerce/issues/50714, so we should be aligned with the effort there too.

samueljseay: Yep agree. Resolved.

plugins/woocommerce/src/Blocks/SharedInteractivityConfig.php
samueljseay: After some thought, I don't think it makes sense to have the word util in the class name and this class belongs to blocks really since the config will be shared amongst blocks. Its still replacing the purpose of AssetDataRegistry, which admittedly had a really confusing name.

The class is now called InteractivityAPIConfig. I'm still not super happy with it. The config is really specific to woo blocks. I will add the method renames to indicate that its shared config though because that makes sense imo.

samueljseay: Renamed methods here https://github.com/woocommerce/woocommerce/pull/56888/commits/b7e594f309999c89cf43df21602cf9c257334770

plugins/woocommerce/src/Blocks/BlockTypes/AbstractBlock.php
```diff
@@ -95,6 +114,7 @@ public function render_callback( $attributes = [], $content = '', $block = null
 			$this->register_block_type_assets();
 			$this->enqueue_assets( $render_callback_attributes, $content, $block );
 		}
+		$this->interactivity_api_config->initialize_shared_config();
```
luisherranz, line 117: Are we sure that it is necessary to serialize the default settings every time there is a Woo block in the front end, whatever that block may be?

plugins/woocommerce/src/Blocks/InteractivityAPIConfig.php
```diff
@@ -0,0 +1,104 @@
...
+	 * Interface for adding block settings to wp_interactivity_config. New settings will be merged with existing settings.
+	 *
+	 * @param array $data The data to add to the settings.
+	 */
+	public function add_shared_config( $data ) {
+		if ( ! $this->core_config_registered ) {
+			$this->initialize_shared_config();
+		}
+
+		wp_interactivity_config( $this->settings_namespace, $data );
```
luisherranz, line 102: What is the benefit of a method for adding settings to the namespace of `woocommerce` instead of doing it more explicitly with `wp_interactivity_config( 'woocommerce', $data );`?

luisherranz (commented):
A couple of questions.

plugins/woocommerce/src/Blocks/BlockTypes/AbstractBlock.php
samueljseay: We are not sure of that no. We can remove this and get blocks to explicitly do this instead.

plugins/woocommerce/src/Blocks/InteractivityAPIConfig.php
samueljseay: @luisherranz I don't think there is really a benefit now. Maybe at some point I had a good reason for this, but now I'm happy to remove it. the class can just be responsible to setup the few settings that get used a lot.

samueljseay:
Gonna reopen this against trunk

plugins/woocommerce/src/Blocks/BlockTypes/AbstractBlock.php
luisherranz: It would still be good to expose a global way to serialize those settings if it's something that multiple blocks need to do. But perhaps we could combine it with the same API we need to create for the state:

https://github.com/woocommerce/woocommerce/blob/381d06e0cd5368938c5f61c0e28ba0a763fb8ec7/plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php#L293-L309

samueljseay: @luisherranz I've moved over into a new PR and opted to explicitly init when needed. if multiple blocks call it still happens just once, but I wonder if you're looking for something else, we can discuss there https://github.com/woocommerce/woocommerce/pull/57469/files#diff-cdfe42b347ee28e297ddff155cd570f732c92f9f1b07371bc09a1eb83fccc78bR43

I certainly expected it will be needed by multiple blocks.

Issue #40118, opened 2023-09-11
Remove Puppeteer E2E Package Dependencies


ObliviousHarmony:
By removing these packages from our PNPM workspace and the related dependencies we will improve the performance of `pnpm install` quite significantly.

```[tasklist]
### Acceptance Criteria
- [ ] Remove any uses of the packages
- [ ] Remove the packages from `dependencies` and `devDependencies`
- [ ] Remove the packages from the PNPM workspace
```

kalessil:
Done as part of #40117.

Issue #40117, opened 2023-09-11
Deprecate Puppeteer E2E Packages


ObliviousHarmony:
As per [NPM's instructions](https://href.li/?https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions), we should deprecate the legacy E2E packages. We have migrated to Playwright and are no longer supporting these packages.

kalessil:
DevAd sync: p1749465274191259-slack-C02NVV8LQLW
PRs: as per the issue timeline (linked on GitHub).

kalessil:
Closing: deprecations are performed, and linked PRs are merged.

Also starting backlog cleanup thread: p1749628841462039-slack-C07JJG089M0

PR #56892, opened 2025-04-01
[dev] CI: CI-workflow tweaks (API, Metrics, Blocks E2E start conditions)

kalessil:

- Refine start conditions for Core API, Metrics, Blocks E2E jobs for running in PRs (narrows down and aims for a more relevant set of running jobs).

- CI-checks shall pass

### DIFF
--- a/plugins/woocommerce/package.json
+++ b/plugins/woocommerce/package.json
@@ -547,6 +547,11 @@
 						"tests/metrics/**",
 						".wp-env.json"
 					],
+					"onlyForDependencies": [
+						"@woocommerce/admin-library",
+						"@woocommerce/block-library",
+						"@woocommerce/classic-assets"
+					],
 					"testEnv": {
 						"start": "env:test"
 					},
@@ -581,6 +586,7 @@
 						"templates/**",
 						"patterns/**"
 					],
+					"onlyForDependencies": [],
 					"testEnv": {
 						"start": "env:start:blocks"
 					},
@@ -718,6 +724,7 @@
 		"@woocommerce/api": "workspace:*",
 		"@woocommerce/e2e-utils-playwright": "workspace:*",
 		"@woocommerce/eslint-plugin": "workspace:*",
+		"@wordpress/browserslist-config": "next",
 		"@wordpress/e2e-test-utils-playwright": "wp-6.8",
 		"@wordpress/env": "10.17.0",
 		"@wordpress/scripts": "30.6.0",
@@ -754,8 +761,7 @@
 	"dependencies": {
 		"@woocommerce/admin-library": "workspace:*",
 		"@woocommerce/block-library": "workspace:*",
-		"@woocommerce/classic-assets": "workspace:*",
-		"@wordpress/browserslist-config": "next"
+		"@woocommerce/classic-assets": "workspace:*"
 	},
 	"nodemonConfig": {
 		"delay": 2500,
--- a/plugins/woocommerce/package.json
+++ b/plugins/woocommerce/package.json
@@ -445,7 +445,7 @@
 					}
 				},
 				{
-					"name": "Core API tests (HPOS:on)",
+					"name": "Core API tests",
 					"testType": "api",
 					"command": "test:api",
 					"optional": false,
@@ -472,7 +472,7 @@
 					"events": [
 						"pull_request",
 						"push",
-						"core-api-hpos-enabled"
+						"core-api"
 					],
 					"report": {
 						"resultsBlobName": "core-api-report",
--- a/plugins/woocommerce/package.json
+++ b/plugins/woocommerce/package.json
@@ -499,7 +499,6 @@
 					],
 					"onlyForDependencies": [],
 					"events": [
-						"pull_request",
 						"push",
 						"core-api-hpos-disabled"
 					],
--- a/plugins/woocommerce/package.json
+++ b/plugins/woocommerce/package.json
@@ -463,7 +463,9 @@
 						"woocommerce.php",
 						"uninstall.php"
 					],
-					"onlyForDependencies": [],
+					"onlyForDependencies": [
+						"@woocommerce/e2e-utils-playwright"
+					],
 					"testEnv": {
 						"start": "env:test"
 					},
@@ -497,7 +499,9 @@
 						"woocommerce.php",
 						"uninstall.php"
 					],
-					"onlyForDependencies": [],
+					"onlyForDependencies": [
+						"@woocommerce/e2e-utils-playwright"
+					],
 					"events": [
 						"pull_request",
 						"push",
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/package.json
```diff
@@ -445,34 +445,34 @@
 					}
 				},
 				{
-					"name": "Core API tests",
+					"name": "Core API tests (HPOS:on)",
```
kalessil: Changes (Core API tests (HPOS:on)):
- do not track patterns/templates changes in PRs as those unlikely to impact REST (trunk push will cover us)
- changes in other packages are irrelevant, except  one exception

plugins/woocommerce/package.json
```diff
@@ -481,25 +481,29 @@
 					}
 				},
 				{
-					"name": "Core API tests - HPOS disabled",
+					"name": "Core API tests (HPOS:off)",
```
kalessil, line 482: Changes (Core API tests (HPOS:off)):
- run in PRs
- do not track patterns/templates changes in PRs as those unlikely to impact REST (trunk push will cover us)
- changes in other packages are irrelevant, except one exception

plugins/woocommerce/package.json
```diff
@@ -547,6 +550,11 @@
 						"tests/metrics/**",
 						".wp-env.json"
 					],
+					"onlyForDependencies": [
```
kalessil, line 548: Changes (Metrics):
- trigger only on intermediate prod-deps changes (other packages should cover each other with plain JS tests and running the job on trunk merge)

plugins/woocommerce/package.json
```diff
@@ -577,9 +585,14 @@
 						"--shard=10/10"
 					],
 					"changes": [
-						"src/Blocks/**",
-						"templates/**",
-						"patterns/**"
+						"patterns/**/*.php",
```
kalessil, line 583: Changes (Blocks e2e tests):
- narrow down patterns a bit more specifically
- trigger only on changes in two specific packages (other packages should cover each other with plain JS tests and running the job on trunk merge)

plugins/woocommerce/package.json
adimoldovan: Nitpick: HPOS on is default, so there's no good reason to have it explicit in title. We don't have it for any other job.

plugins/woocommerce/package.json
```diff
@@ -445,34 +445,34 @@
...
 						"tests/e2e-pw/tests/api-tests/**",
 						"tests/e2e-pw/*.js",
 						"tests/e2e-pw/*.sh",
 						".wp-env.json",
 						"woocommerce.php",
 						"uninstall.php"
 					],
+					"onlyForDependencies": [
+						"@woocommerce/e2e-utils-playwright"
+					],
```
adimoldovan: This dependency is not actually impacting the API tests.

plugins/woocommerce/package.json
```diff
@@ -445,34 +445,34 @@
...
 					"testEnv": {
 						"start": "env:test"
 					},
 					"events": [
 						"pull_request",
 						"push",
-						"core-api"
+						"core-api-hpos-enabled"
```
adimoldovan: Same nitpick on HPOS being on by default

plugins/woocommerce/package.json
```diff
@@ -481,25 +481,29 @@
...
+						"tests/e2e-pw/tests/api-tests/**",
+						"tests/e2e-pw/*.js",
+						"tests/e2e-pw/*.sh",
 						".wp-env.json",
 						"woocommerce.php",
 						"uninstall.php"
 					],
+					"onlyForDependencies": [
+						"@woocommerce/e2e-utils-playwright"
+					],
```
adimoldovan: Same as above, @woocommerce/e2e-utils-playwright is not required.

plugins/woocommerce/package.json
```diff
@@ -481,25 +481,29 @@
...
 					"events": [
+						"pull_request",
```
adimoldovan: Why is this important to run on PR?

plugins/woocommerce/package.json
```diff
@@ -577,9 +585,14 @@
...
-						"templates/**",
-						"patterns/**"
+						"patterns/**/*.php",
+						"src/Blocks/**/*.php",
+						"templates/**/*.php",
+						"templates/**/*.html"
+					],
+					"onlyForDependencies": [
+						"@woocommerce/block-library",
+						"@woocommerce/e2e-utils"
```
adimoldovan: "@woocommerce/e2e-utils" is no longer used for long time now, interesting it's a dependency in blocks project. I'm sure it's not used though and can be removed.

adimoldovan: see #40117 and #40118

kalessil: Blocks E2Es are heavily using it, I double-checked as had similar thoughts.

plugins/woocommerce/package.json
kalessil: Following my gut feeling in this case, can revert if it's overkill.

plugins/woocommerce/package.json
kalessil: Reverted in 2be5a1dc02f9175c909193682d242f51e5daad02

plugins/woocommerce/package.json
kalessil: Reverted in 2be5a1dc02f9175c909193682d242f51e5daad02

plugins/woocommerce/package.json
kalessil: Super! Dropped in 9b85a86e7b9a8112bfe57bb524ce14dc4ec07cad

plugins/woocommerce/package.json
kalessil: Super! Dropped in 9b85a86e7b9a8112bfe57bb524ce14dc4ec07cad

plugins/woocommerce/package.json
kalessil: But regarding the deprecation path, let's drop this dependency from this config.

kalessil: Here we go: 280251c800ed999db9c4b76e7ef6c8f768b199d2

adimoldovan: > Blocks E2Es are heavily using it, I double-checked as had similar thoughts.

TIL! that package is old and using Puppeteer, blocks should not be using it.

adimoldovan: Actually, blocks are not using that package after all, `@woocommerce/e2e-utils` is mapped to `tests/e2e/utils`: https://github.com/woocommerce/woocommerce/blob/f4d80351eda1dab106db285827cf99ae5432103f/plugins/woocommerce/client/blocks/tsconfig.base.json#L68

plugins/woocommerce/package.json
adimoldovan: I think it's an overkill. We'll be protected on push and the daily is running some e2e with HPOS off as well.

kalessil: Ok, reverting!

kalessil: Reverted in 50ca9a070c35774664a86844f74a13b6ddda4f3f

PR #56912, opened 2025-04-01
POC: Add Regular and Sale fields to Product Editor

### LINKED ISSUE #56844: [ POC Product Editor DataForm ] Migrate the regular and sale price over
Parent Github issue - #56834

imanish003:

Closes #56844

| Before | After |
| ------ | ----- |
|        |       |

1.
2.
3.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/packages/js/product-editor/src/data-form/components/column/index.tsx
+++ b/packages/js/product-editor/src/data-form/components/column/index.tsx
@@ -1,0 +1,83 @@
+/**
+ * External dependencies
+ */
+import { Product } from '@woocommerce/data';
+import { Template } from '@wordpress/blocks';
+import { useSelect, useDispatch } from '@wordpress/data';
+import { store as coreDataStore } from '@wordpress/core-data';
+import { createElement, useMemo } from '@wordpress/element';
+import { DataForm } from '@wordpress/dataviews';
+
+/**
+ * Internal dependencies
+ */
+import { useDataFormProductFields } from '../use-data-form-product-fields';
+
+type ProductColumnProps = {
+	columnTemplate: Template;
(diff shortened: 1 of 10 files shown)


louwie17 (approved):
Nice, this looks good!! I like how you split up the Columns!

Btw on a side note: Not something that needs to be addressed as part of this PR, but we will have to try to come up with a better solution of still supporting the passed in attributes like the `Label`, also many other fields support more attributes.
And the `disableConditions` which the sale price uses here for example: https://github.com/woocommerce/woocommerce/blob/89068601d334953e2904ecf56f528fc271c7b9ec/plugins/woocommerce/src/Internal/Features/ProductBlockEditor/ProductTemplates/SimpleProductTemplate.php#L266-L276

I was leaning towards either wrapping the `Edit` function in a context that we can use, within the `Edit` similar to the `useBlockContext`.

imanish003:
> Btw on a side note: Not something that needs to be addressed as part of this PR, but we will have to try to come up with a better solution of still supporting the passed in attributes like the Label, also many other fields support more attributes.
And the disableConditions which the sale price uses here for example:

Yeah, I was also thinking about this while working on this PR. However, since this is still a POC, I chose to ignore it. But sure, we definitely need to address all of this.

> I was leaning towards either wrapping the Edit function in a context that we can use, within the Edit similar to the useBlockContext.

I like the idea. It would be great to have context somewhere at the top of the tree, allowing all the fields and child blocks to consume that context.

PR #56927, opened 2025-04-01
Implement a way for developers to register an address provider in PHP.

opr:

This PR implements the server-side registration system for address autocomplete providers in WooCommerce Blocks. It adds a new abstract class and filter that allows developers to register address autocomplete providers (which will be `WC_Address_Provider`s) through a filter.

Closes: WOOPLUG-3615

Key additions:
- Create New `AddressProviderServiceProvider` service class for Woo core DI Container
- Create new `AddressProviderController` class for handling registrations of providers
- New abstract `WC_Address_Provider` class
- New filter to add providers.
- Unit tests for the new functionality

1. Create a PHP snippet with the following code:
```php
class WP_Provider extends WC_Address_Provider {	
		$this->id   = 'wp-provider';
		$this->name = 'WordPress provider';
	}
};

class WC_Provider extends WC_Address_Provider {
		$this->id   = 'woo-provider';
		$this->name = 'Woo provider';
	}
};


add_filter( 'woocommerce_address_providers', function( $providers ) {
	$providers[] = 'WP_Provider';
	$providers[] = 'WC_Provider';
	return $providers;
} );

```

2. Ensure no errors are shown while using the site, try accessing different areas of the admin, e.g. shipping settings, or browsing customers, orders, and products.
3. Ensure no errors occur while going through the checkout process. (Add an item to cart, go to the Cart block, go to the Checkout block, place the order).

- PHP Unit tests added covering:
  - Filter application
  - Multiple filter application
  - Overriding earlier filter applications

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Blocks/Domain/Bootstrap.php
```diff
@@ -403,6 +404,14 @@ function () {
 				return new BlockTemplatesController();
 			}
 		);
+		if ( Constants::is_defined( 'WC_EXPERIMENTAL_ADDRESS_AUTOCOMPLETE' ) && Constants::get_constant( 'WC_EXPERIMENTAL_ADDRESS_AUTOCOMPLETE' ) ) {
```
senadir: We should just use `Features::is_enabled( 'experimental-blocks' )` when gating things.

plugins/woocommerce/src/Blocks/Domain/Services/AddressAutocomplete.php
```diff
@@ -0,0 +1,97 @@
+<?php
+declare( strict_types=1 );
+namespace Automattic\WooCommerce\Blocks\Domain\Services;
+
+/**
+ * Service class for managing address autocomplete providers.
+ */
+class AddressAutocomplete {
```
senadir: We discussed this in our meetup @opr but we now also want to support maps block, even if we're not going to do that right now, we should give that space to ourself later on. This is an internal class, so we can rename it later but I think we should move from `AddressAutocomplete` to something more generic that includes both a map and address search. It's a safe assumption that any service that allows you to embed a map, will also provide address search, as searching for an address is a prerequisite for a map. Things like mapbox, Apple Maps, and Google Maps, all provide address search.

I think we can rename this to something like AddressProvider, and remove the mention of autocomplete or mention it alongside a map

plugins/woocommerce/src/Blocks/Domain/Services/AddressAutocomplete.php
```diff
@@ -0,0 +1,97 @@
...
+	private $providers = [];
+
+	/**
+	 * Register a new address autocomplete provider.
+	 *
+	 * @param string $id   The provider ID.
+	 * @param string $name The provider name.
+	 * @return bool True if registration was successful, false otherwise.
+	 */
+	public function register_provider( string $id, string $name ): bool {
```
senadir: Should we match other providers in Woo and ask people to provide a class? this is what Shipping and Payments do, they extend an abstract class `WC_Payment_Gateway` and `WC_Shipping_Method`. We can also match how things are registered the same way, which is via a filter. This helps race conditions in which developers don't need to time out when they register things, they just append the class to a filter and we call that when we need it.

senadir (changes requested):
Thank you Thomas, I left some feedback, mainly in 2 places:

- We match the current behavior of how payments and shipping are registered, nothing wrong with that approach and it served us well.
- No need for an unregister method, people can just filter it out of the filter.
- We move to a class instead of passing strings, this give us space for future methods that won't break backward compatibility.
- Rename things slighlty to accomade the fact that we will support maps in the future.

plugins/woocommerce/src/Blocks/Domain/Bootstrap.php
opr: Updated in 0c551e564843e0e4d0c1bf04efc66d690ab85645

plugins/woocommerce/src/Blocks/Domain/Services/AddressAutocomplete.php
opr: I thought about it, and I think the classes we want the users to register are `AddressProvider`s so it makes sense for our class to be called something like `AddressProviderService`. Updated 👌🏼

plugins/woocommerce/src/Blocks/Domain/Services/AddressAutocomplete.php
opr: I think this is fine, we do give up the error checking and duplicate provider checking though don't we, if we're just operating on the result of a filter.

senadir: You can still do that at initiation step, not registration step, which is a fine solution, we don't want to break often and early, only break as little as possible.

senadir (commented):
Thank you for the update, Thomas! I have even more feedback, can you believe it? anything to keep you busy.

First, any reason why you initiate the class, add stuff to the object (not the class?), and then pass it back?

If we're matching WC_Payment_Gateway and WC_Shipping_Method, they only pass back the class name, all the data happens on the class level. It's also important to assign translatable strings at the __construct level and not earlier, to not trigger a `translation_too_early` error.

https://github.com/woocommerce/woocommerce/blob/01fef596d4c96f1aa3b7e8643a4321ec2b51f763/plugins/woocommerce/includes/shipping/free-shipping/class-wc-shipping-free-shipping.php#L23-L65

When it comes to the actual filter, shipping one passes an id => class name and payments is a simple classes array, I'm not sure which is best.

plugins/woocommerce/src/Blocks/Domain/Services/AddressAutocomplete.php
opr: Sounds good, will probably get to that in the next PR then when we actually use it.

plugins/woocommerce/src/Blocks/Domain/Services/AddressProviderService.php
```diff
@@ -0,0 +1,102 @@
...
+	/**
+	 * Get all registered providers.
+	 *
+	 * @return WC_Address_Provider[] array of WC_Address_Providers.
+	 */
+	public function get_registered_providers(): array {
+		/**
+		 * Filter the registered address providers.
+		 *
+		 * @since 10.2.0
```
tjcafferkey: I think this should be `9.9.0` like the abstract class `WC_Address_Provider`

plugins/woocommerce/src/Blocks/Domain/Services/AddressProviderService.php
```diff
@@ -0,0 +1,102 @@
...
+
+		$providers = [];
+
+		foreach ( $provider_class_names as $provider_class_name ) {
+			// Ensure the class exists and is a valid WC_Address_Provider subclass.
+			if ( class_exists( $provider_class_name ) && is_subclass_of( $provider_class_name, WC_Address_Provider::class ) ) {
+				$provider_instance = new $provider_class_name();
+
+				// Validate the instance has the necessary properties.
+				if ( ! empty( $provider_instance->id ) && ! empty( $provider_instance->name ) ) {
```
tjcafferkey: Do we want to ensure IDs are unique? Whats the consequence of providers with duplicate IDs, even if an edge-case? Especially since `is_provider_available()` queries by ID.

plugins/woocommerce/src/Blocks/Domain/Services/AddressProviderService.php
```diff
@@ -0,0 +1,102 @@
...
+		}
+
+		// If the class names haven't changed, return the cached instances.
+		if ( $this->cached_provider_class_names === $provider_class_names && ! empty( $this->cached_providers ) ) {
+			return $this->cached_providers;
+		}
+
+		$providers = [];
+
+		foreach ( $provider_class_names as $provider_class_name ) {
```
tjcafferkey: Shall we just `continue;` if `$provider_class_name` is not a string?

plugins/woocommerce/tests/php/src/Blocks/Domain/Services/AddressProviderServiceTest.php
```diff
@@ -0,0 +1,435 @@
...
+
+use Mockery\Adapter\Phpunit\MockeryTestCase;
+use Automattic\WooCommerce\Blocks\Domain\Services\AddressProviderService;
+use Automattic\WooCommerce\Blocks\Package;
+use WC_Address_Provider;
+
+/**
+ * Tests for Address Provider Service functionality
+ */
+class AddressProviderServiceTest extends MockeryTestCase {
```
tjcafferkey: If we do ensure IDs are unique we can write an additional test case for this.

plugins/woocommerce/src/Blocks/Domain/Services/AddressProviderService.php
opr: Good point, we did have that check in the previous iteration, I'll reimplement for the new approach

plugins/woocommerce/src/Blocks/Domain/Services/AddressProviderService.php
opr: Thanks for the suggestion, I refactored to fail early and only add the provider if it passes all tests. Also added unit tests for this.

plugins/woocommerce/tests/php/src/Blocks/Domain/Services/AddressProviderServiceTest.php
opr: Done in 2648eab807f887a13f718c495357f4730de5eb85

tjcafferkey (approved):
Thanks for addressing my feedback @opr! I've retested this and everything is working well. I even tried passing invalid/incorrect data types to trigger some errors/fatals if possible but wasn't able to which is good. Also appreciate the test coverage.

Approving, but feel free to wait for Nadir's approval if you feel like its necessary.

senadir (approved):
I gave this a final review, thank you taking the feedback and adding this Thomas!

PR #56976, opened 2025-04-02
Move `supports` definition to block.json

kmanijak:

Some of the blocks had `supports` listed in JS and passed over to `registerBlockType` function rather than having them in `block.json`. That means `supports` aren't visible to frontend (PHP) as it relies on `block.json` content.

This PR moves the JS-defined `supports` to block.json.

The main reason they were kept in JS was usage of  `typeof __experimentalGetSpacingClassesAndStyles === 'function' &&` condition as a form of a feature flag for some `supports` properties. [`__experimentalGetSpacingClassesAndStyles` function](https://github.com/WordPress/gutenberg/blob/a32cb4b49c25f92cce1c191d8a4b265336d9522c/packages/block-editor/src/index.js#L12) was exported in Gutenberg 4 years ago, meaning that condition is always true for WordPress versions WooCommerce supports and condition is useless.
In some other cases `supports` was simply stored in JS just like that. Along with that change I removed `get_block_type_supports` from PHP block classes as this is used to get `supports` for frontend. It was supposed to be kept in sync with JS definition. With the migration to `block.json` this is no longer needed as `supports` from metadata will be used.

In addition, this PR includes removal of `shared/config` from Product Elements. With the migration of block definition to `block.json` files, the shared config didn't make much sense anymore.

Part of WOOPLUG-2055

| Before | After |
| ------ | ----- |
|        |       |

1. Create new post
2. Add Product Collection and add couple of blocks inside
    - Rating
    - SKU
3. Now focus on blocks and verify
    - **Expected:** they have style options in Inspector Controls (set of options should differ between blocks as that depends on configuration. I reused the existing config per block)
    - **Expected:** play round with options, they should be applied to block (we're aware some options are not applied correctly but it's not a purpose of this PR to test and fix them. Raise a problem if for example block doesn't react to ANY controls)
4. Save and go to frontend
5. **Expected:** applied styles are displayed here as well

Example: SKU

Options | Editor view | Frontend view
-- | -- | --
image | image | image

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


gigitux:
Thanks for working on this! I noticed a different UI with this PR:

| Header | Header |
|--------|--------|
|image |image | 

Regarding the style, I didn't find any regression!

kmanijak:
> Thanks for working on this! I noticed a different UI with this PR:

Good catch! I added category to remaining blocks in https://github.com/woocommerce/woocommerce/pull/56976/commits/361b4174a11ca116e43451b01f72dd29a0063d67.

Or do you mean missing Woo logo? I don't see it even on `trunk`. And TBH I'm not sure where/how is it injected 🤔

gigitux (approved):
Thanks for working on this! Everything looks great!

image

PR #56981, opened 2025-04-02
Fix: Prevent null offset warning in has_block_including_patterns method found in BlockTemplateUtils.php

Varan170:
### Description

Please refer to the attached PDF to view the error I was experiencing. Or if you want to view the errors directly, I have attached images to the screenshots or screen recordings section.

In the function `has_block_including_patterns` (Line 721), a warning was being triggered: "Trying to access array offset on value of type null," due to a lack of checks for a possibly null block pattern. This commit adds validation to ensure that the block pattern exists and contains the specified content before attempting to parse it. This change improves stability and eliminates warnings when specific block patterns are not registered or when their content is not set.

The method was added during [this commit](https://github.com/woocommerce/woocommerce/commit/e6b1cac49c0d252df930425c7936b12e38bb2a8b).

I believed that a little more validation in this method would fix the error, and I was correct to think so.

I have attached a video demonstration showing that this change removes the error: [Video Demonstration of Bug](https://www.youtube.com/watch?v=guqxQKkCURc).

I believe the error is an edge case stemming from the migration of the site from my university's server to my local environment (since this was originally developed as an assignment). It may be influenced by variations in server configurations or the backup process through different plugins(like WPVivid). Strengthening the handling of such cases can enhance the WooCommerce plugin’s robustness.

- I have checked to ensure there are no other open [Pull Requests](https://github.com/woocommerce/woocommerce/pulls) for the same update/change.
- Following the above guidelines will result in quick merges and clear and detailed feedback when appropriate.

### Changes Proposed in This Pull Request:

This change adds validation checks in the `has_block_including_patterns` method to prevent null offset warnings. It ensures that patterns are validated before accessing their contents.

### Screenshots or Screen Recordings:

- [Wordpress woocommerce error.pdf](https://github.com/user-attachments/files/19567076/Wordpress.woocommerce.error.pdf)
- [Video Demonstration of Bug](https://www.youtube.com/watch?v=guqxQKkCURc)
- Image version of PDF to show the errors directly in the pull request.

Wordpress woocommerce error_page-0001
Wordpress woocommerce error_page-0002
Wordpress woocommerce error_page-0003
Wordpress woocommerce error_page-0004
Wordpress woocommerce error_page-0005
Wordpress woocommerce error_page-0006
Wordpress woocommerce error_page-0007
Wordpress woocommerce error_page-0008

### How to Test the Changes in This Pull Request: (thanks again to @Aljullu and @opr for helping with this.)(this assumes you have a working installation of WordPress and WooCommerce)

1. Add a /templates/single-product.html file to your theme with these contents:

<main class="wp-block-group">
	
	

	
	<div class="wp-block-columns alignwide">
		
		<div class="wp-block-column" style="flex-basis:512px">
			
		</div>
		

		
		<div class="wp-block-column">
			

			

			

			

			

			
			<div class="wp-block-woocommerce-product-meta">
				
				<div class="wp-block-group">
					

					

					

				</div>
				
			</div>
			
		</div>
		
	</div>
	

	

	
</main>

2. Enable a block theme, and then go to Editor -> Appearance -> Templates.
3. Find the single product template and reset it (click the 3 dots and then reset). If the reset option is not there, that's fine.
4. Go to the single product page.
5. Confirm that the issue is fixed and the single product page displays correctly with no PHP warnings.

### Changelog Entry

- [ ] This Pull Request does not require a changelog entry. 

<details>

#### Significance

- [x] Patch

#### Type

- [x] Fix - Fixes an existing bug

#### Message 

Fix: Prevent null offset warning in BlockTemplateUtils.php by adding validation checks for block patterns.

</details>

<details>

#### Comment 

</details>


opr:
@Varan170 Hi there, thanks for your contribution! I noticed you've uploaded a PDF to the PR. Would you mind sharing the content directly in the PR description or as plain text/markdown files instead? This makes it easier for us to review and gives context directly in the PR rather than in a separate file.

Thanks!

Varan170:
> @Varan170 Hi there, thanks for your contribution! I noticed you've uploaded a PDF to the PR. Would you mind sharing the content directly in the PR description or as plain text/markdown files instead? This makes it easier for us to review and gives context directly in the PR rather than in a separate file.
> 
> Thanks!

@opr Hi, Thank you for your feedback! I have uploaded all relevant pages from the PDF as images directly in the pull request description, which I believe provides a clearer context for the error I was experiencing and also makes your work easier. 
If you need any further information or additional modifications, please let me know. I'm looking forward to your review!

opr:
@Varan170 thanks for that. I am a little unfamiliar with this area of code, but would love to help review your PR.

Can you help me with some specific reproduction instructions? What steps do I need to take to get the error to occur?

Also, your PR will need a changelog entry. I see you filled the details in the PR body but didn't check the `Automatically create a changelog entry from the details below.`. I have done that for you now so it should be added to the PR. I also removed the comment, the comment is used when we _don't_ want to include an entry. I think generating a changelog file for this is the right move.

Adding the changelog adds a commit to the branch, so you would need to make sure your local branch is in sync with the remote before making any further changes.

Once I've got reproduction instructions I'll test this fully. 🙇🏼

plugins/woocommerce/src/Blocks/Utils/BlockTemplateUtils.php
```diff
@@ -718,11 +718,12 @@ public static function has_block_including_patterns( $block_names, $blocks ) {
 			) {
 				$registry       = WP_Block_Patterns_Registry::get_instance();
 				$pattern        = $registry->get_registered( $block['attrs']['slug'] );
```
opr:
```suggestion
				$registry = WP_Block_Patterns_Registry::get_instance();
				$pattern  = $registry->get_registered( $block['attrs']['slug'] );
```

plugins/woocommerce/src/Blocks/Utils/BlockTemplateUtils.php
```diff
@@ -718,11 +718,12 @@ public static function has_block_including_patterns( $block_names, $blocks ) {
...
-
-				if ( self::has_block_including_patterns( $block_names, $pattern_blocks ) ) {
-					return true;
-				}
+				if ( $pattern && isset( $pattern['content'] ) ) {
+                	$pattern_blocks = parse_blocks( $pattern['content'] );
+	                if ( self::has_block_including_patterns( $block_names, $pattern_blocks ) ) {
+	                    return true;
+	                }
+            	}
```
opr:
```suggestion
				if ( isset( $pattern['content'] ) ) {
					$pattern_blocks = parse_blocks( $pattern['content'] );
					if ( self::has_block_including_patterns( $block_names, $pattern_blocks ) ) {
						return true;
					}
				}
```

Varan170:
> @Varan170 thanks for that. I am a little unfamiliar with this area of code, but would love to help review your PR.
> 
> Can you help me with some specific reproduction instructions? What steps do I need to take to get the error to occur?
> 
> Also, your PR will need a changelog entry. I see you filled the details in the PR body but didn't check the `Automatically create a changelog entry from the details below.`. I have done that for you now so it should be added to the PR. I also removed the comment, the comment is used when we _don't_ want to include an entry. I think generating a changelog file for this is the right move.
> 
> Adding the changelog adds a commit to the branch, so you would need to make sure your local branch is in sync with the remote before making any further changes.
> 
> Once I've got reproduction instructions I'll test this fully. 🙇🏼

Hi @opr,

Thank you so much for your willingness to help me with this! I really appreciate your guidance as I navigate my first contribution to a public project. I’m currently an IT student, and while I have a good general understanding of programming, I have limited experience with PHP.

I have added detailed reproduction instructions for the error in the pull request description in the section *How to reproduce the error*. After further testing on a different Windows 11 PC, I believe the issue I encountered is an edge case related to migration from my university's server to a local setup or by the WPVivid Backup plugin or some other unknow issue. However, it’s still something that needs to be addressed.

I also commented on the original commit to try to bring the attention of the person who created the has_block_including_patterns method, as I think it would be beneficial for them to be aware of this issue.

If you have any trouble reproducing the error or following the instructions, please let me know. 
Thanks again for your support.

plugins/woocommerce/src/Blocks/Utils/BlockTemplateUtils.php
opr: The initial `if ( $pattern` check isn't necessary, since that's covered by the `isset` call.

opr (changes requested):
@Varan170 I appreciate you adding the steps, and I see this is your first PR so I would like to suggest some improvements for them, hope you don't mind.

Generally we wouldn't expect to see instructions on setting up webservers etc. it's assumed that someone testing a PR would have a site and WooCommerce store set up, so it's safe to skip those parts. We also wouldn't expect to have to import people's databases to test issues (unless the bug specifically related to the import!) it should be possible for a tester to reproduce the bug themself. That way we rule out issues caused by individual sites misconfigurations.

I have a way to test the bug without requiring the import, thanks to @Aljullu for helping with this.

1. Go to `wp-content/plugins/woocommerce-monorepo/plugins/woocommerce/templates/templates/blockified/single-product.html` and add `<!-- wp:pattern {"slug":"missing-pattern"} /-->` somewhere in the block.
2. Enable a block theme, and then go to Editor -> Appearance -> Templates.
3. Find the single product template and reset it (click the 3 dots and then reset). If the reset option is not there, that's fine.
4. Go to the single product page.
5. See the warning.

I've tested your fix and it works, so that's great 👏🏼  I've just suggested some changes to satisfy the lint rules, and also noted that one check can be removed as it's covered by the subsequent `isset` check.

Let me know if you have any questions.

Varan170:
> @Varan170 I appreciate you adding the steps, and I see this is your first PR so I would like to suggest some improvements for them, hope you don't mind.
> 
> Generally we wouldn't expect to see instructions on setting up webservers etc. it's assumed that someone testing a PR would have a site and WooCommerce store set up, so it's safe to skip those parts. We also wouldn't expect to have to import people's databases to test issues (unless the bug specifically related to the import!) it should be possible for a tester to reproduce the bug themself. That way we rule out issues caused by individual sites misconfigurations.
> 
> I have a way to test the bug without requiring the import, thanks to @Aljullu for helping with this.
> 
> 1. Go to `wp-content/plugins/woocommerce-monorepo/plugins/woocommerce/templates/templates/blockified/single-product.html` and add `<!-- wp:pattern {"slug":"missing-pattern"} /-->` somewhere in the block.
> 2. Enable a block theme, and then go to Editor -> Appearance -> Templates.
> 3. Find the single product template and reset it (click the 3 dots and then reset). If the reset option is not there, that's fine.
> 4. Go to the single product page.
> 5. See the warning.
> 
> I've tested your fix and it works, so that's great 👏🏼 I've just suggested some changes to satisfy the lint rules, and also noted that one check can be removed as it's covered by the subsequent `isset` check.
> 
> Let me know if you have any questions.

Thanks you very much @opr and @Aljullu for the help. I have updated the PR Description with the changes you suggested as I was able to reproduce the error myself as well on a fresh wordpress and woocommerce install and was able to apply the fix to see the correct result. I agree we can fix the code to pass the lint rules. I am glad that working together have been able to fix this issue which feels very rewarding.

plugins/woocommerce/src/Blocks/Utils/BlockTemplateUtils.php
Varan170: > The initial `if ( $pattern` check isn't necessary, since that's covered by the `isset` call.

I have tested this change and I agree with you. Thanks for the help with pass the lint tests as well. I will commit this now.

opr (approved):
@Varan170 thanks for making the changes!

The testing instructions look better, I just edited them a little more, we also usually write instructions to check if the issue is fixed, we wouldn't usually include a step that essentially asks the tester to check it's broken, hope that makes sense?

I'll approve this PR but I've tagged Albert to give it a second review since he's more familiar with this code area.

Thanks again and congrats on your first PR 👏🏼

Aljullu (approved):
Thanks for working on this, @Varan170! :raised_hands: And thanks for the initial review, @opr!

Code changes look good to me, I tested and it seems to work fine as well:

Before | After
--- | ---
imatge | imatge

One small suggestion about the testing steps, though:

> 1. Go to wp-content/plugins/woocommerce/templates/templates/blockified/single-product.html and add `<!-- wp:pattern {"slug":"missing-pattern"} /-->` somewhere in the block.

After thinking about this, I would prefer if we can change this step to something like:

1. Add a `/templates/single-product.html` file to your theme with this contents:

```HTML
<!-- wp:template-part {"slug":"header"} /-->
<!-- wp:pattern {"slug":"missing-pattern"} /-->
<!-- wp:group {"tagName":"main","layout":{"inherit":true,"type":"constrained"}} -->
<main class="wp-block-group">
	<!-- wp:woocommerce/breadcrumbs /-->
	<!-- wp:woocommerce/store-notices /-->

	<!-- wp:columns {"align":"wide"} -->
	<div class="wp-block-columns alignwide">
		<!-- wp:column {"width":"512px"} -->
		<div class="wp-block-column" style="flex-basis:512px">
			<!-- wp:woocommerce/product-image-gallery /-->
		</div>
		<!-- /wp:column -->

		<!-- wp:column -->
		<div class="wp-block-column">
			<!-- wp:post-title {"level": 1, "__woocommerceNamespace":"woocommerce/product-query/product-title"} /-->

			<!-- wp:woocommerce/product-rating {"isDescendentOfSingleProductTemplate":true} /-->

			<!-- wp:woocommerce/product-price {"isDescendentOfSingleProductTemplate":true, "fontSize":"large"} /-->

			<!-- wp:post-excerpt {"__woocommerceNamespace":"woocommerce/product-query/product-summary", "excerptLength":100} /-->

			<!-- wp:woocommerce/add-to-cart-form /-->

			<!-- wp:woocommerce/product-meta -->
			<div class="wp-block-woocommerce-product-meta">
				<!-- wp:group {"layout":{"type":"flex","flexWrap":"nowrap"}} -->
				<div class="wp-block-group">
					<!-- wp:woocommerce/product-sku {"isDescendentOfSingleProductTemplate":true} /-->

					<!-- wp:post-terms {"term":"product_cat","prefix":"Category: "} /-->

					<!-- wp:post-terms {"term":"product_tag","prefix":"Tags: "} /-->

				</div>
				<!-- /wp:group -->
			</div>
			<!-- /wp:woocommerce/product-meta -->
		</div>
		<!-- /wp:column -->
	</div>
	<!-- /wp:columns -->

	<!-- wp:woocommerce/product-details {"align":"wide","className":"is-style-minimal"} /-->

	<!-- wp:pattern {"slug":"woocommerce-blocks/related-products"} /-->
</main>
<!-- /wp:group -->

<!-- wp:template-part {"slug":"footer"} /-->
```

While the step we currently have works, WooCommerce doesn't officially support users modifying templates under `/wp-content/plugins/woocommerce/templates/`, while themes having their own templates is something that _is supported_. So I feel it's better to write the steps for a user-flow that we support.

But besides that, LGTM. :slightly_smiling_face: 

<details>

* WordPress 6.7.2 local dev environment
* Plugins: WooCommerce
* Theme: Twenty Twenty-Five
* Store creation: April 6th 2025
* Store size: small
</details>

Aljullu:
Thanks for the PR, @Varan170! I have merged it an it will be available in WooCommerce 9.9! :partying_face:

Varan170:
> Thanks for the PR, @Varan170! I have merged it an it will be available in WooCommerce 9.9! 🥳

Cheers @Aljullu , I am glad I could help. And thanks again to @opr as well for helping me with my first contribution to a public project as well.

PR #56985, opened 2025-04-02
Updates to support offline payment methods.

### LINKED ISSUE #56364: NOX - Update Offline method pages as per the new design
The settings pages for the offline payment methods - bank transfer, check, and cash on delivery - are now out of sync with the new Payment settings design. We should update them to make the UX consistent.

### Acceptance criteria

* Align all components, typestyles, and spacing with the Gutenberg style.
* Apply a two-column layout.
* Remove the top-level tabbed navigation.

#### Notes

* Back button + page title will be handled separately in https://github.com/woocommerce/woocommerce/issues/55738.
* White background will be handled separately by @woocommerce/ghidorah.

#### Before

Image

#### After

Image

### Additional context

* pfHfG4-XF-p2
* LK697bTM8sHMU95BS4owef-fi-1325_144844

dmallory42:

Closes #56364  

**Before**
Screenshot 2025-04-29 at 17 11 10

Screenshot 2025-04-29 at 17 11 20

Screenshot 2025-04-29 at 17 11 27

**After**
Screenshot 2025-04-29 at 18 21 53

Screenshot 2025-04-29 at 18 22 33

1. Spin up new JN site with this branch.
2. Ensure the new payment settings experience is enabled.
3. Go to `Settings > Payments > Offline Methods`.
4. Go to the `Direct bank transfer` method, verify the form loads. Make some changes to the form and save. Verify the form reloads and a successful `POST` request is sent. Reload the page and verify that your changes persisted.
5. Disable the new payment settings experience and go to the _old_ version of the direct bank transfer page. Verify that the changes you made using the new form are correctly persisted.
6. Repeat steps 4-5 with the other offline methods, cash on delivery and check. 

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/includes/rest-api/Controllers/Version3/class-wc-rest-payment-gateways-controller.php
```diff
@@ -223,4 +223,52 @@ public function get_item_schema() {
...
 	}
+
+	/**
+	 * Validate multiselect based settings (with support for nested options).
+	 *
+	 * @param array|string $values  The submitted values.
+	 * @param array        $setting The field settings.
+	 * @return array|WP_Error
+	 */
+	public function validate_setting_multiselect_field( $values, $setting ) {
```
dmallory42, line 234: This fixes an issue where the validation of options keys wasn't working as expected (because the array was multidirectional, but it was expecting a flat array)

oaratovskyi:
**Testing-only review** (code review will follow soon)

Issues found:

- Cash on delivery can't be opened
image
- Different columns in bank account details
image
- Can't add bank account. Filled in the data but it wasn't saved. Tried to load a gif it's too big, will add to Slack.
image
  - do we need the country? no country in old version of the page
  - view/edit, delete ellipsis behavior and styling isn't consistent with what we have in NOX settings page (should hide when clicked)
- When editing bank account details saved from the old page, it's impossible to save
image
  - IBAN field wasn't required in old version, now it is
  - required fields errors are not visible to user, they are not red
- all those pages, shouldn't be centered, see the design 
image 
image

And this separately:
- Need to check out with the design, but I guess we need it consistent with other pages
image
  - navigation bar should probably be hidden (or not - since it's not hidden in WooPayments settings)
  - top level text should be settings (it's everywhere like that) and the back button should be below (like in Fraud page)
  image

packages/js/data/src/payment-gateways/types.ts
```diff
@@ -7,26 +7,32 @@ export type SettingDefinition = {
...
-	order: number | '';
-	enabled: boolean;
-	method_title: string;
-	method_description: string;
-	method_supports: string[];
-	settings: SettingDefinition;
-	settings_url: string;
-	needs_setup?: boolean;
-};
+export type PaymentGateway< TSettings = Record< string, SettingDefinition > > =
```
oaratovskyi, line 18: It should work fine, but I wonder if this can break some existing functionality that relies on this type? Could you double check please? 🙏

plugins/woocommerce/client/admin/client/settings-payments/components/bank-accounts-list/bank-account-modal.tsx
```diff
@@ -41,22 +46,18 @@ export const BankAccountModal = ( {
 	onSave,
 	defaultCountry,
 }: Props ) => {
+	const countries = window.wcSettings.countries;
```
oaratovskyi, line 50: Do we need those countries? I see that there is no countries in the old version. As far as I get, countries influence what fields are required, right? Could you link me please to the requirement from product/design about it, I seem to have missed it

plugins/woocommerce/client/admin/client/settings-payments/components/bank-accounts-list/types.ts
```diff
@@ -1,9 +1,7 @@
 export interface BankAccount {
-	id: string;
```
oaratovskyi, line 2: not sure what is the cause, but bank account creation does not work. Maybe due to this id absense?

plugins/woocommerce/client/admin/client/settings-payments/components/bank-accounts-list/utils.ts
```diff
@@ -1,10 +1,32 @@
...
  *
- * @param country The selected country code.
+ * @return {string} The label for the sort code field.
  */
-export const getDefaultRoutingField = ( country: string ) => {
-	if ( country === 'US' ) return 'routing_number';
-	if ( country === 'AU' ) return 'sort_code';
-	return 'iban';
+export const getSortcodeLabel = ( country: string ) => {
+	if ( country === 'AU' ) return __( 'BSB', 'woocommerce' );
```
oaratovskyi: I suggest to change this to `switch` 🙂

plugins/woocommerce/client/admin/client/settings-payments/components/payment-settings-layout/payment-settings-layout.scss
```diff
@@ -0,0 +1,22 @@
+.payment-settings-layout {
+	margin: 24px auto 0;
+	max-width: 1032px;
+	display: flex;
+	flex-flow: column;
+
+	// TODO: Add mobile styling (row instead of column).
```
oaratovskyi: spotted that, adding the comment, so it's not missed 🙈

plugins/woocommerce/client/admin/client/settings-payments/components/payment-settings-layout/payment-settings-layout.scss
```diff
@@ -0,0 +1,22 @@
+.payment-settings-layout {
+	margin: 24px auto 0;
+	max-width: 1032px;
```
oaratovskyi: is that so in design? number seems unusual

plugins/woocommerce/client/admin/client/settings-payments/components/payment-settings-layout/payment-settings-layout.tsx
```diff
@@ -0,0 +1,14 @@
+/**
+ * External dependencies
+ */
+
+/**
+ * Internal dependencies
+ */
+import './payment-settings-layout.scss';
+
+export const PaymentSettingsLayout = ( {
```
oaratovskyi: could you add some small comment about this component? It's not self-explanatory for me

plugins/woocommerce/client/admin/client/settings-payments/components/gateway-settings-form/gateway-settings-form.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Internal dependencies
+ */
+
+interface GatewaySettingsFormProps {
+	children: React.ReactNode;
+}
+
+export const GatewaySettingsForm: React.FC< GatewaySettingsFormProps > = ( {
```
oaratovskyi:
```suggestion
export const PaymentGatewaySettingsForm: React.FC< GatewaySettingsFormProps > = ( {
```

plugins/woocommerce/client/admin/client/settings-payments/components/payment-settings-layout/payment-settings-layout.tsx
oaratovskyi: maybe PaymentGatewaySettingsLayout

plugins/woocommerce/client/admin/client/settings-payments/components/field-placeholder/field-placeholder.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { Placeholder } from '@wordpress/components';
+
+/**
+ * Internal dependencies
+ */
+import './field-placeholder.scss';
+
+export type FieldPlaceholderSize = 'small' | 'medium' | 'large';
+
+export const FieldPlaceholder = ( {
```
oaratovskyi, line 13:
```suggestion
export const PaymentGatewayFieldPlaceholder = ( {
```

plugins/woocommerce/client/admin/client/settings-payments/components/payment-settings-section/payment-settings-section.scss
```diff
@@ -0,0 +1,60 @@
...
+		flex-flow: row;
+	}
+
+	&__details {
+		flex: 0 1 auto;
+		margin-bottom: 24px;
+		max-width: 328px;
+
+		@include breakpoint( ">782px" ) {
+			flex: 0 0 33%;
```
oaratovskyi: please add comment about this behaviour

plugins/woocommerce/client/admin/client/settings-payments/components/payment-settings-section/payment-settings-section.scss
```diff
@@ -0,0 +1,60 @@
...
+		}
+
+		h2 {
+			font-size: 16px;
+			line-height: 24px;
+		}
+
+		p {
+			font-size: 13px;
+			line-height: 17.89px;
```
oaratovskyi: unusual number too! Is that so in design?

plugins/woocommerce/client/admin/client/settings-payments/components/payment-settings-section/payment-settings-section.tsx
```diff
@@ -0,0 +1,28 @@
+/**
+ * External dependencies
+ */
+
+/**
+ * Internal dependencies
+ */
+import './payment-settings-section.scss';
+
+export const PaymentSettingsSection = ( {
```
oaratovskyi:
```suggestion
export const PaymentGatewaySettingsSection = ( {
```

plugins/woocommerce/client/admin/client/settings-payments/offline/settings-payments-bacs.tsx
```diff
@@ -0,0 +1,251 @@
...
+import '../settings-payments-body.scss';
+import './settings-payments-offline-method.scss';
+import { FieldPlaceholder } from '~/settings-payments/components/field-placeholder';
+import { PaymentSettingsSection } from '~/settings-payments/components/payment-settings-section';
+import { BankAccountsList } from '~/settings-payments/components/bank-accounts-list';
+import { PaymentSettingsLayout } from '~/settings-payments/components/payment-settings-layout';
+import { GatewaySettingsForm } from '~/settings-payments/components/gateway-settings-form';
+import { BankAccount } from '~/settings-payments/components/bank-accounts-list/types';
+
+export const SettingsPaymentsBacs = () => {
```
oaratovskyi, line 27: add please a comment about the component

oaratovskyi: maybe call it direct bank transfer so it's more readable?

plugins/woocommerce/client/admin/client/settings-payments/offline/settings-payments-bacs.tsx
```diff
@@ -0,0 +1,251 @@
...
+					{ isLoadingAccounts ? (
+						<FieldPlaceholder size="large" />
+					) : (
+						<BankAccountsList
+							accounts={ accounts }
+							onChange={ setAccounts }
+							defaultCountry={ storeCountryCode }
+						/>
+					) }
+				</PaymentSettingsSection>
```
oaratovskyi:
```suggestion
				</PaymentGatewaySettingsSection>
```
I suggest changing this and according styling - so it's more specific, otherwise folks can this it's related to the payment settings page

plugins/woocommerce/client/admin/client/settings-payments/offline/settings-payments-cod.tsx
```diff
@@ -0,0 +1,250 @@
...
+							label={ __(
+								'Enable for shipping methods',
+								'woocommerce'
+							) }
+							help={ __(
+								'Select shipping methods for which this payment method is enabled.',
+								'woocommerce'
+							) }
+							options={
+								codSettings.settings.enable_for_methods?.options
```
oaratovskyi: this does not work locally 🐛  idea: maybe move it into useEffect and use ready data here?

oaratovskyi (changes requested):
Thanks for this big work! It's in a good shape, but I believe we should fix it to make it work before merging
___
#### Review checklist
| Parameter  | Result |
| Approach  | ✅ All good |
| Implementation  | 👍  Almost good (left comments) |
| Automated tests  | ⚠️ Would be great to add some test to client components and newly introduced backend function |
| Naming  | 👍  Almost good (left comments) |
| Code comments  | 👍  Almost good (left a comment) |
| PR description  | ✅ Good |
| Testing instructions  | ✅ Good |
| Manual testing  | ⚠️ Faced some challenges during testing, see comment above | 

#### Scenarios tested
| Scenario  | Result | 
| Cash on delivery gateway  | ⚠️ Does not fully work. Need to address the shipping methods part |
| Check payments gateway  | ✅ Functionality worked correctly | 
| Direct bank transfer gateway  | ⚠️ Does not fully work. Need to address bank account adding part |

packages/js/data/src/payment-gateways/types.ts
dmallory42: The previous definition is incorrect 🙂 so this is really a bug fix to match the actual structure which gets returned. It was defining it like this (example):

```
settings: {
  default: string;
  description: string;
  id: string;
  label: string;
  options?: Record< string, Record< string, string > >;
  placeholder: string;
  tip: string;
  type: string;
  value: string;
  value: string | string[];
  is_dismissed: string;
}
``` 

when it should be more like:

```
settings: {
  "setting_1": {
    default: string;
    description: string;
    id: string;
    label: string;
    options?: Record< string, Record< string, string > >;
    placeholder: string;
    tip: string;
    type: string;
    value: string;
    value: string | string[];
    is_dismissed: string;
 },
 "setting_2": {
    default: string;
    description: string;
    id: string;
    label: string;
    options?: Record< string, Record< string, string > >;
    placeholder: string;
    tip: string;
    type: string;
    value: string;
    value: string | string[];
    is_dismissed: string;
 },
}
```

I will double check it didn't break anything else though, thanks for mentioning it

plugins/woocommerce/client/admin/client/settings-payments/components/bank-accounts-list/bank-account-modal.tsx
dmallory42: Yes, country selection is a requirement from the designs by Magda (I didn't add it in the previous PR because it was just easier to address it in this one) - here's the P2 post which outlines the requirements: pfHfG4-17j-p2 🙂

plugins/woocommerce/client/admin/client/settings-payments/components/field-placeholder/field-placeholder.tsx
dmallory42: This can be used generically for any field, so IMO `FieldPlaceholder` is an appropriate name for it.

oaratovskyi: OK, agree 🙂

plugins/woocommerce/client/admin/client/settings-payments/components/bank-accounts-list/utils.ts
dmallory42: Good suggestion. Updated in c91f889324!

plugins/woocommerce/client/admin/client/settings-payments/components/payment-settings-layout/payment-settings-layout.scss
dmallory42: Handled in 370a44a843!

plugins/woocommerce/client/admin/client/settings-payments/components/payment-settings-layout/payment-settings-layout.scss
dmallory42: Have removed this constraint in 370a44a843. Good spot!

plugins/woocommerce/client/admin/client/settings-payments/components/payment-settings-layout/payment-settings-layout.tsx
dmallory42: 370a44a843 sorts out all these components. My original intention was to create a lightweight structure that can be used for any form page. In this commit, I took your advice about these components not being clear and reduced them to just one component with sub-sections, which I think is a lot clearer. I also added a comment explaining it.

plugins/woocommerce/client/admin/client/settings-payments/components/payment-settings-section/payment-settings-section.scss
dmallory42: I think this one was AI-generated weirdness. It should be a 20px line height. Fixed now!

plugins/woocommerce/client/admin/client/settings-payments/offline/settings-payments-bacs.tsx
dmallory42: Updated in d0d09a984c. I think I prefer `bacs` because it is the same identifier used elsewhere in the code (more searchable), but I added a comment referring to it as direct bank transfer as well.

plugins/woocommerce/client/admin/client/settings-payments/offline/settings-payments-bacs.tsx
dmallory42: Removed in 370a44a843242f748568e6b9369d4d13cad5559b

plugins/woocommerce/client/admin/client/settings-payments/components/payment-settings-section/payment-settings-section.tsx
dmallory42: Removed in 370a44a843242f748568e6b9369d4d13cad5559b

plugins/woocommerce/client/admin/client/settings-payments/components/gateway-settings-form/gateway-settings-form.tsx
dmallory42: Resolved by 370a44a843242f748568e6b9369d4d13cad5559b

dmallory42:
Thank you for the detailed review, @oaratovskyi! I have addressed your feedback and think this is in much better shape now. I will work on adding some automated tests, but this is ready for another round of review.

@rogermattic, I think this is also ready for a design review when you have time. 🙂 You can test it by just spinning up a JN site with this branch and going to the `Settings > Payments > Offline Methods` page and then checking each offline methods settings page. Let me know if you need anything else!

oaratovskyi (commented):
Thanks, Daniel! It's in much better shape now, almost good!

Some items to sort:
- styling in cash on delivery shipping methods 
image
- Wrong check payments header title
image
- direct bank transfer Country is not saved

https://github.com/user-attachments/assets/0de68e97-3c02-4f6d-869a-766ad6770fdf

- "dirty" reloader when changed some data, saved and try to reload page with Ctrl +R

https://github.com/user-attachments/assets/6e8dafc2-d9b8-438e-a2c4-62ef88348364

dmallory42:
Thanks for the further reviews, @oaratovskyi - regarding your comments:

> styling in cash on delivery shipping methods

I think that sadly this won't be possible to address in this PR. We were able to identify a component, `TreeSelectControl` which provides the same functionality as the design, but the appearance is slightly different - while we can update it in future, I think we should do this as a follow-up issue since we'll possibly need to create a new component for it.

> Wrong check payments header title

Fixed. Thanks! 

> direct bank transfer Country is not saved

Great catch! Fixed - note that this relies on saving a new option value, so it will only work for newly added/saved bank accounts (legacy bank accounts will default to the store country until updated).

> dirty" reloader when changed some data, saved and try to reload page with Ctrl +R

I tried to fix this, but it's complicated because it's caused by the WP `mainform` which wraps the whole page. I think we can fix it, but I'd prefer to handle it in a new issue.

oaratovskyi (approved):
OK, thanks for addressing those! 
It's in a very good shape now - approving 🟢 
Please create small issues to address the follow ups with the styling and dirty data to address it next week 🙏

dmallory42:
I just pushed a small commit which hides the navbar on the offline methods pages as mentioned in the design feedback. 🙂

rogermattic:
Hey @dmallory42 ! Sorry for blocking it for longer! Thanks for working on this! The pages look awesome <3 Here are some comments and notes that I wanted to share, also available in Figma ( 96Xr930vUUS29z5ta3CsD3-fi-936_1926 ):

QA May 8, 2025

dmallory42:
Thanks for the great feedback @rogermattic - I have fixed most of it and I will leave comments on Figma for anything I couldn't address here so we can create issues for them.

Edit: follow-up issue created here: https://github.com/woocommerce/woocommerce/issues/57879

daledupreez:
@dmallory4, @woocommerce/moltres, and @rogermattic  we tripped over a subtle downstream problem stemming from these changes in QIT tests for the WooCommerce Stripe gateway. Prior to these changes, the original implementation created a `<label>` element for both the `title` and `label` attributes returned from the server, but we're now creating only a label for the `label` attribute. We tripped over this because the activation tests were using the label from the `title`.

I am not sure how major that is from a functional or usability perspective, as the act of clicking on the title section in the older settings screen is not something I'd normally expect to toggle a checkbox! This definitely feels like a "feature" of the old implementation, but I want us to make a conscious call as to what we will (and will not) support.

The pre-existing code that generates the label is here: https://github.com/woocommerce/woocommerce/blob/c46ae897aa0e2bb0f3290416967f60d7246b1e23/plugins/woocommerce/includes/abstracts/abstract-wc-settings-api.php#L461

dmallory42:
Hi @daledupreez, thanks for pointing this out. I'm a bit unsure how these changes had that knock-on effect, are you able to identify which changes in the PR could have caused that? As far as I know, this PR did not intend to change how labels get rendered in other places - it implements new React settings pages specifically for the 3 offline method settings pages, the intention was not to change/break anything downstream.

I'm happy to wrangle a fix for that!

daledupreez:
@dmallory42, the TL;DR is that the activation e2e tests were navigating directly to the CoD setup page (i.e. `page=wc-settings&tab=checkout&section=cod`), and then trying to get the main checkbox via `getByLabel( 'Enable/Disable' )`, which is the label generated for the `title` attribute in current versions, but is not present in the new UI page. The code was also trying to get the virtual order checkbox to toggle it, but I see we've changed the structure there to use the `title` as the label, and have the secondary text in a `<span>`.

You can see my draft fix for this in https://github.com/Automattic/compatibility-dashboard/pull/1332 - I suspect that may be easier to grok.

alopezari:
Thanks for creating that draft fix @daledupreez!

Just wanted to ask for confirmation that there shouldn't be negative impact by accepting this change on how labels are generated, as it wasn't an actual requirement but a consequence of a recent implementation.

I think this might be a question for @dmallory42?

dmallory42:
Thank you @daledupreez, that draft PR clarifies it for me! I think I was a little confused by the mention of the WooCommerce Stripe gateway as I was worried I'd inadvertantly broken something there! I understand now that the QIT tests were trying to access those elements by labels which have now changed.

> Just wanted to ask for confirmation that there shouldn't be negative impact by accepting this change on how labels are generated, as it wasn't an actual requirement but a consequence of a recent implementation.

> I think this might be a question for @dmallory42?

To confirm, the changes implemented in this PR are mainly a continuation of prior work to 'modernise' the payment settings pages, with an implementation that uses React components instead of the old method of injecting the page content via PHP. The changes in labelling are based on the design requirements which are referenced [here](https://github.com/woocommerce/woocommerce/issues/56364), but they shouldn't affect anything functionally - the settings page should allow a merchant to do exactly the same things as they could previously. 

So, in summary I think updating the tests to reflect the changes should be the only action we need to take here - and it shouldn't have a knock-on negative impact, because the way it was implemented means that only pages which are specifically 'marked' as having a React-powered version are generated in this way. Although, we should remember to check the QIT tests if we work on any pages in future. 

In case it is useful for the tests, we use the feature flag, `reactify-classic-payments-settings` as a way of checking whether to render the new pages or to use the 'classic' settings variations.

alopezari:
Awesome, thank you very much for the confirmation and the detailed explanation, @dmallory42!

I'll review your PR, @daledupreez!

PR #57013, opened 2025-04-03
Fix tax rate API support for non-Latin class names

ismaeldcom:

Closes WOOPLUG-2071

- Update `create_or_update_tax` to properly handle field keys, ensuring the switch cases are executed accordingly.
- Add a test for tax classes with non-Latin characters

Using `$key` instead of `$field` ends up using only the default case, which causes the mentioned issue, as it runs `wc_clean` on the class, removing any encoded character and using `standard` class as a fallback.

> [!NOTE]
> This instructions relies on WC REST API, if you don't have it configured, please check [the instructions.](https://woocommerce.github.io/woocommerce-rest-api-docs/?shell#authentication). And WC tax to be enabled from **WooCommerce → Settings**

1. Add a tax class with non-Latin characters via POST request to `https://<local-server>/wp-json/wc/v3/taxes/classes` with:

```jsonc
{
  "name": "µ"  // or any other non-Latin char/word.
}
```

2. Add a tax rate with the previous class via POST request to `https://<local-server>/wp-json/wc/v3/taxes` with:

```jsonc
{
  "class": "%c2%b5" // or the returned slug from the previous request.
}
```

3. Confirm that the returned tax rate has the new class instead of `standard`.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 
Fix tax rate API support for non-Latin class names.
</details>

<details>

#### Comment 

</details>


plugins/woocommerce/tests/php/includes/rest-api/Controllers/Version3/class-wc-rest-taxes-controller-tests.php
```diff
@@ -277,12 +277,39 @@ public function test_can_get_taxes_filtering_by_class( $class ) {
...
 		$this->assertEquals( array( $tax_ids_by_class[ $class ] ), $ids );
 	}
+
+	/**
+	 * @testdox Tax rates with non-Latin characters in tax class names are properly created and associated with the correct class.
+	 */
+	public function test_can_create_tax_rate_with_non_latin_tax_class() {
+		wp_set_current_user( $this->user );
+
+		$tax_class_name = '∑';
```
samnajian, line 295: Question: do we need to check and add test for other none-Latin class names? such as Arabic, Persian, Hebrew, or east Asian characters?

plugins/woocommerce/changelog/57013-fix-45677-tax-rate-api-support-non-latin-class
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix tax rate API support for non-Latin class names.
```
samnajian: Let's keep the empty line at the end.

plugins/woocommerce/tests/php/includes/rest-api/Controllers/Version3/class-wc-rest-taxes-controller-tests.php
ismaeldcom: Not needed, any non-ASCII/non-Latin produces the same URL-encoded output.

plugins/woocommerce/changelog/57013-fix-45677-tax-rate-api-support-non-latin-class
ismaeldcom: I've made the changelog entry using the GH action from the PR description, should I just add an extra new line there? 🤔

samnajian:
@ismaeldcom 
This looks good to me and works as expected, I was wondering if it shouldn't be accepting the none-URL-encoded characters as well for the tax class name for the second request (to https://<local-server>/wp-json/wc/v3/taxes )?

plugins/woocommerce/changelog/57013-fix-45677-tax-rate-api-support-non-latin-class
samnajian: We should probably fix it on the GH Action level? ( in a separate issue )

ismaeldcom: Add it manually in 3447830

ismaeldcom:
>I was wondering if it shouldn't be accepting the none-URL-encoded characters as well for the tax class name for the second request (to https:///wp-json/wc/v3/taxes )?

I think we should do that too, but I didn't want to change something like that without having the full context. I'm not 100% sure there's a reason not to do it.

samnajian:
> I think we should do that too, but I didn't want to change something like that without having the full context. I'm not 100% sure there's a reason not to do it.

@ismaeldcom I'm fine with delegating it to another issue as it's essentially not part of the current issue, but wouldn't be too much if you decided to do it in this PR as well, so I'd leave this decision to you.

ismaeldcom:
I'll have another look at it to see if we have a reason to encode the slug, and if not it might cause an issue with existing ones.

ismaeldcom:
Hey @samnajian, after taking another look and making a few tests, I think we should proceed with this fix, discarding the idea of supporting non-Latin class slugs, at least for now.

We're sanitizing every slug in the codebase, and doing it at multiple places like API requests, SQL queries, and HTML params. Supporting it will require a major code update and migration of existing tax classes.

samnajian:
> We're sanitizing every slug in the codebase, and doing it at multiple places like API requests, SQL queries, and HTML params. Supporting it will require a major code update and migration of existing tax classes.

@ismaeldcom let's do it like this.
Would you mind creating an issue for it?

samnajian (approved):
Great work.
LGTM.
It would be great to have an issue for [this](https://github.com/woocommerce/woocommerce/pull/57013#issuecomment-2809293773).

ismaeldcom:
Thanks for the review @samnajian. Issued created #57340 🙂

PR #57025, opened 2025-04-03
Use string instead of number for email logo width default value

### LINKED ISSUE #56752: Value of `woocommerce_email_header_image_width` email setting is a string in some external environments instead of a number.
### Prerequisites

### Describe the bug

`woocommerce_email_header_image_width` is a number in our default wp-env localhost testing environment, while it's a string on our Pressable testing environment.

#### On wp-env localhost:
Image

#### On Pressable
Image

### Expected behavior

The value of `woocommerce_email_header_image_width` should consistently be a number across all environments.

### Actual behavior

The value of `woocommerce_email_header_image_width` is a string on Pressable.

### Steps to reproduce

1. In a Pressable site, make an authenticated request to `/wp-json/wc/v3/settings/email/woocommerce_email_header_image_width`.
2. Notice that the `value` in the response body is a numerical string, instead of a number.

### WordPress Environment

WP 6.7
WC 9.8 Beta

### Isolating the problem

NeosinneR:

This PR changes the `woocommerce_email_header_image_width` type from number to string. The reason is that there is no harm in using a string, and it's more difficult to ensure that the value is everywhere a number, when in the database, it's saved in the LONGTEXT column.

Closes #56752, closes WOOPLUG-3638.

(For Bug Fixes) Bug introduced in PR #54897.

1. Go to **Settings > Permalinks** and ensure that anything other than `Plain` is chosen (otherwise REST API doesn't work). 
2. Go to **User > Profile** and create a new Application password (note the password).
3. Locally, you can test that `curl -u "admin:APPLICATION_PASSWORD" "http://localhost:8888/wp-json/wc/v3/settings/email/woocommerce_email_header_image_width"` return strings for both `value` and `default`.

I've tested locally that both values are strings.

### DIFF
--- a/plugins/woocommerce/includes/admin/settings/class-wc-settings-emails.php
+++ b/plugins/woocommerce/includes/admin/settings/class-wc-settings-emails.php
@@ -149,7 +149,7 @@ class WC_Settings_Emails extends WC_Settings_Page {
 				'title'    => __( 'Logo width (px)', 'woocommerce' ),
 				'id'       => 'woocommerce_email_header_image_width',
 				'desc_tip' => '',
-				'default'  => 120,
+				'default'  => '120',
 				'type'     => 'number',
 			);
 			$header_alignment           = array(
--- a/plugins/woocommerce/changelog/wooplug-3638-value-of-woocommerce_email_header_image_width-email-setting
+++ b/plugins/woocommerce/changelog/wooplug-3638-value-of-woocommerce_email_header_image_width-email-setting
@@ -1,0 +1,5 @@
+Significance: patch
+Type: fix
+Comment: Only changes the default value type from number to string for email logo width. No visible or functional change
+
+
--- a/plugins/woocommerce/tests/e2e-pw/tests/api-tests/settings/settings-crud.test.js
+++ b/plugins/woocommerce/tests/e2e-pw/tests/api-tests/settings/settings-crud.test.js
@@ -1634,7 +1634,7 @@ test.describe( 'Settings API tests: CRUD', () => {
 						id: 'woocommerce_email_header_image_width',
 						label: 'Logo width (px)',
 						type: 'number',
-						default: 120,
+						default: '120',
 						value: expect.anything(), // value could be number or string depending on environment
 					} ),
 				] )


plugins/woocommerce/includes/admin/settings/class-wc-settings-emails.php
```diff
@@ -149,7 +149,7 @@ protected function get_settings_for_default_section() {
 				'title'    => __( 'Logo width (px)', 'woocommerce' ),
 				'id'       => 'woocommerce_email_header_image_width',
 				'desc_tip' => '',
-				'default'  => 120,
+				'default'  => '120',
 				'type'     => 'number',
```
websupporter, line 153: shouldnt the `type` be `string` then?

NeosinneR: This is used for `<input type="HERE">` - I used number so there is HTML validation for this being a number and also it adds the arrows to the input and allows changing the value with arrows. 
Screenshot 2025-04-04 at 09 59 06

websupporter: So that type field is used to determine the input type?

websupporter:
Have we explored what it would take to ensure that always numbers are returned?

plugins/woocommerce/includes/admin/settings/class-wc-settings-emails.php
NeosinneR: Yes. So you can also use anything there, e.g. `my_custom_component` and render whatever there. Or you can use any of the predefined types, which will render the setting row for you https://github.com/woocommerce/woocommerce/blob/6376d46d9c2aca9d097f4a37669e032424094818/plugins/woocommerce/includes/admin/class-wc-admin-settings.php#L333-L367

NeosinneR:
> Have we explored what it would take to ensure that always numbers are returned?

@websupporter not really, because I don't necessarily see the value of it being a number all the time. The way it's implemented now implies it should be a number, because in https://github.com/woocommerce/woocommerce/blob/5a787b309bcc16be3fc6a546f4e4dead1fcd13ad/plugins/woocommerce/templates/emails/email-styles.php#L167-L169 we append `px`, but number-like is good enough, even if the type is string. Not sure if it's worth trying to force the number there.

websupporter:
Reading the issue again:
> Expected behavior
> The value of woocommerce_email_header_image_width should consistently be a number across all environments.

I think we should investigate why it is a string in Pressable and see how we can ensure it is numeric. Changing the default to a string while claiming the type is a number; that feels inconsistent to me and atm I cant evaluate whether  it solves the issue.

NeosinneR:
@websupporter the `default` being a number (`120` instead of `'120'`) might have been the wrong assumption from the start. So while the issue is reported to expect it to be number in all environments, it's based on (my) wrong assumption from the start.

Looking at other number-like settings in Core, all of them use `'type' => 'number'` while having the default value string:

https://github.com/woocommerce/woocommerce/blob/5a787b309bcc16be3fc6a546f4e4dead1fcd13ad/plugins/woocommerce/includes/admin/settings/class-wc-settings-general.php#L282-L288
https://github.com/woocommerce/woocommerce/blob/5a787b309bcc16be3fc6a546f4e4dead1fcd13ad/plugins/woocommerce/includes/admin/settings/class-wc-settings-products.php#L250-L259
https://github.com/woocommerce/woocommerce/blob/5a787b309bcc16be3fc6a546f4e4dead1fcd13ad/plugins/woocommerce/includes/admin/settings/class-wc-settings-products.php#L298-L307

So this PR makes it more consistent with the rest of the Core settings. It goes against the expected behaviour from the issue, but I don't think that should necessarily be a driving factor. Does this make it clearer?

websupporter (approved):
lgtm. thank you @NeosinneR

PR #57037, opened 2025-04-04
Fix additional fields crashing and formatting issues

### LINKED ISSUE #56958: Custom schema validation crashes block
Using additional fields with a JSON Schema validation, a form step would crash if a field has a value and has custom validation.

### Steps to reproduce

* Be an a dev build (`pnpm --filter=@woocommerce/plugin-woocommerce watch:build`) to enable schema validation.
* Add the following field:

```php
		array(
			'id'         => 'plugin-namespace/alt-email-2',
			'label'      => 'Alternative Email',
			'location'   => 'contact',
			'type'       => 'text',
			'required'   => true,
			'validation' => array(
				'type'         => 'string',
				'format'       => 'email',
				'not'          => array(
					'const' => array( '$data' => '3/customer/billing_address/email' ),
				),
				'errorMessage' => array(
					'format' => 'Please enter a valid email address.',
					'not'  => 'Please enter a different email address to your contact email.',
				),
			),
		)
	);
```

* Go to checkout, add a value to this field.
* Notice that checkout crashes.

### Expected behavior

Block shouldnt crash and validation should trigger an error.

### Actual behavior

Block has an infinite loop and crashes.

### LINKED ISSUE #56960: A required but conditionally hidden select would trigger an error on load in Checkout
Via schema conditional fields, a conditionally hidden but required select would mount with an error in Checkout, likely due to persist additional fields work.

### Steps to reproduce:

* Be on a dev build (`pnpm --filter=@woocommerce/plugin-woocommerce watch:build`)
* Add the following field somewhere:

```
		array(
			'id'       => 'plugin-namespace/leave-with-neighbor',
			'label'    => __( 'Which neighbor should we leave it with if unable to hide?', 'woocommerce' ),
			'location' => 'contact',
			'type'     => 'select',
			'required' => true,
			'hidden'   => array(
				'cart' => array(
					'properties' => array(
						'needs_shipping' => array(
							'type'  => 'boolean',
							'const' => true,
						),
					),
				),
			),
			'options'  => array(
				array(
					'label' => 'Neighbor to the left',
					'value' => 'left',
				),
				array(
					'label' => 'Neighbor to the right',
					'value' => 'right',
				),
				array(
					'label' => 'Neighbor across the road',
					'value' => 'across',
				),
				array(
					'label' => 'Do not leave with a neighbor',
					'value' => 'none',
				),
			),
		)
	);
```

* Go to checkout with a shippable product.
* Notice that an error is shown on contact information step

### Expected behavior

* There should be no error if the field is hidden.

### Current behavior

* Checkout mount with an error an order cannot be placed.

### Extra context

This is because the PUT request is sending `""` for that field once it's hidden, and `""` is not accepted in Store API for select, possible solutions:

* Start accepting `""` if a field is hidden.
* Don't send anything if a field is empty (but we need to make sure the value is cleared from the order so it doesn't make it to admin).

### LINKED ISSUE #56987: Checkout pushes invalid additional fields
## Problem

Checkout seems to perform a PUT request to persist additional fields even if they have an error on them. 

## **Steps to Reproduce**

1. Be on a dev build (`pnpm --filter=@woocommerce/plugin-woocommerce watch:build`)
2. Add the following field:

```php
		array(
			'id'         => 'plugin-namespace/alt-email-2',
			'label'      => 'Alternative Email',
			'location'   => 'contact',
			'type'       => 'text',
			'required'   => true,
			'validation' => array(
				'type'   => 'string',
				'format' => 'email',
				'not'    => array(
					'const' => array( '$data' => '3/customer/billing_address/email' ),
				)
			),
		)
	);
```

3. On Checkout, add an email to alternative Email.
4. Repeat the same email in Email
5. Make sure you do that in this order, filling alternative email last would break checkout (bug reported here [WOOPLUG-3758](https://linear.app/a8c/issue/WOOPLUG-3758/custom-schema-validation-crashes-block))
6. Notice that a request is made to the server regardless, which will be rejected.

## Expectation

### LINKED ISSUE #56993: Email validation mismatch between client and server is additional fields
in additional fields, if you define a string format as `email`, you get a mismatching behavior in which a value that's valid on client is invalid on server, this happens when you have a domain without a TLD (`test@localhost`).

### Steps to reproduce

1. Be on a dev build (`pnpm --filter=@woocommerce/plugin-woocommerce watch:build`)
2. Add the following field:

```
		array(
			'id'         => 'plugin-namespace/alt-email-2',
			'label'      => 'Alternative Email',
			'location'   => 'contact',
			'type'       => 'text',
			'required'   => true,
			'validation' => array(
				'type'         => 'string',
				'format'       => 'email',
				'errorMessage' => array(
					'format' => 'Please enter a valid email address.',
				),
			),
		)
	);
```

3. On checkout, type `text@localhost` on that field.
4. Notice it get pushed to server but is rejected.

### Expected behavior

Field should not be pushed, or if pushed, should not return error from server.

### Current behavior

fields return error from server

### Context

This is because [Opis uses](https://github.com/opis/json-schema/blob/master/src/Formats/MiscFormats.php#L48) PHP's FILTER_VALIDATE_EMAIL that doesn't allow emails without TLD and AJV uses a [different regex one that allows them ](https://github.com/ajv-validator/ajv-formats/blob/master/src/formats.ts#L119C26-L119C40)

### LINKED ISSUE #56995: Checkout would revert to customer address after additional field update
On Checkout, if you land with an address 1, update it to address 2, refresh, see values persisted fine, but then change an additional field, after refreshing, notice that address 1 is back instead of address 2.

### Steps to reproduce

1. On Checkout, logged in, make sure you have a full address (remember it) and email. Ideally you place an order with this address.
2. Add the following field:

```php
		array(
			'id'       => 'plugin-namespace/job-function-2',
			'location' => 'contact',
			'required' => true,
			'type'     => 'select',
			'options'  => array(
				array(
					'label' => 'Director',
					'value' => 'director',
				),
				array(
					'label' => 'Engineering',
					'value' => 'engineering',
				),
				array(
				),
				array(
					'label' => 'Other',
					'value' => 'other',
				),
			),
		)
	);
```

3. In checkout, without touching that additional field, edit your address including email and shipping or billing or both.
4. Refresh page, make sure the address you changed persisted.
5. Modify the field you just added, wait for it to persist.
6. Refresh page, notice the new addresses you added are gone, and old address is reverted.

### Expected behavior

New address should have persisted.

### Current behavior

Additional fields updates revert address to what's in customer profile.

### Context

This seems to only happen for PUT requests, even thou the returned cart has the correct value, after refresh it's gone, I suspect cart is stale at that moment.

### LINKED ISSUE #56996: Hidden additional fields are still present on the final order
In Checkout block, if you have an additional field that was visible, got a value persisted, and then became hidden, its value will still be present on the final order.

### Steps to reproduce:

1. Be on a dev build to trigger schema fields: (`pnpm --filter=@woocommerce/plugin-woocommerce watch:build`).
2. Add the following field

```php
	array(
		'location' => 'contact',
		'required' => true,
		'type'     => 'select',
		'hidden'   => array(
			'customer' => array(
				'properties' => array(
					'billing_address' => array(
						'properties' => array(
							'email' => array(
								'pattern' => '^[a-zA-Z0-9._%+-]+@(gmail|hotmail|icloud)\.com$',
							),
						),
					),
				),
			),
		),
		'options'  => array(
			array(
				'label' => 'Director',
				'value' => 'director',
			),
			array(
				'label' => 'Engineering',
				'value' => 'engineering',
			),
			array(
			),
			array(
				'label' => 'Other',
				'value' => 'other',
			),
		),
	)
);
```

3. On Checkout, have an email with something like a8c or automattic domain.
4. "What's your role" field will be visible.
5. Select a value, remember it.
6. Refresh page, make sure value is still there.
7. Change your domain to gmail or hotmail, notice the field is now hidden.
8. Place order.
9. See that the value you selected is still there.

### Expected behavior

Hidden fields should be removed from order and shouldn't make it to order

### Current behavior

Hidden fields stay persisted

### Context

This is because we ignore hidden values on PHP, even the process of unsetting a value that's hidden is ignored.

Also, this seems to only happen for contact and order step, doesn't happen for country 🤔

### LINKED ISSUE #56997: Schema validation via formats isn't working
It seems that I mistakenly disabled validating additional fields via format, os if you pass in `format` keyword, it's ignored.

### Steps to reproduce

1. Be on a dev build (`pnpm --filter=@woocommerce/plugin-woocommerce watch:build`)
2. Add the following field:

```
		array(
			'id'         => 'plugin-namespace/alt-email-2',
			'label'      => 'Alternative Email',
			'location'   => 'contact',
			'type'       => 'text',
			'required'   => true,
			'validation' => array(
				'type'         => 'string',
				'format'       => 'email',
				'errorMessage' => 'Please enter a valid email address.',
			),
		)
	);
```

3. On checkout, type `text` on that field.
4. Focus away from it.
5. Notice it get pushed with to server and no error is present on client, but the server will reject it.

### Expected behavior

Field should not be pushed, and an error client side is shown.

### Current behavior

field has no error and passes fine.

senadir:
This PR fixes a bunch of issues that happened in additional fields, some related to schema validations, other to persist additional fields.

Closes https://github.com/woocommerce/woocommerce/issues/56997
Closes https://github.com/woocommerce/woocommerce/issues/56996
Closes https://github.com/woocommerce/woocommerce/issues/56987
Closes https://github.com/woocommerce/woocommerce/issues/56960
Closes https://github.com/woocommerce/woocommerce/issues/56958
Closes https://github.com/woocommerce/woocommerce/issues/56993
Closes #56995

Given this PR fixes 5 things, I will try to cover all 5 of them as unified as possible:

Be on a dev build (pnpm --filter=@woocommerce/plugin-woocommerce watch:build)
Add the following field somewhere in your code:
```php
add_action(
	'woocommerce_init',
	function () {
			array(
				'id'         => 'plugin-namespace/alt-email',
				'label'      => 'Alternative Email',
				'location'   => 'contact',
				'type'       => 'text',
				'required'   => true,
				'validation' => array(
					'type'         => 'string',
					'format'       => 'email',
					'not'          => array(
						'const' => array( '$data' => '3/customer/billing_address/email' ),
					),
					'errorMessage' => array(
						'format' => 'Please enter a valid email address.',
						'not'    => 'Please enter a different email address to your contact email.',
					),
				),
			)
		);

			array(
				'location' => 'contact',
				'required' => true,
				'type'     => 'select',
				'hidden'   => array(
					'customer' => array(
						'properties' => array(
							'billing_address' => array(
								'properties' => array(
									'email' => array(
										'pattern' => '^[a-zA-Z0-9._%+-]+@(gmail|hotmail|icloud)\.com$',
									),
								),
							),
						),
					),
				),
				'options'  => array(
					array(
						'label' => 'Director',
						'value' => 'director',
					),
					array(
						'label' => 'Engineering',
						'value' => 'engineering',
					),
					array(
					),
					array(
						'label' => 'Other',
						'value' => 'other',
					),
				),
			)
		);
	}
);
```
1. On an incognito window, go to checkout with a product.
2. You should see email, alternative email, and "role" select, and no error from server (https://github.com/woocommerce/woocommerce/issues/56960)
3. Input an invalid email `test.com` in the alternative email, and blur the field, notice that it shows an error (https://github.com/woocommerce/woocommerce/issues/56997)
4. Notice that it doesn't send a request to the server if the field is invalid (https://github.com/woocommerce/woocommerce/issues/56987)
5. In the email field, enter a gmail email. Add the same email in the alternative email field, make sure it doesn't crash (https://github.com/woocommerce/woocommerce/issues/56958).
6. Change your email domain to something else (a8c.com), notice the error is gone from the alternative email field, and a new select field is visible.
7. Select a value for the new select field, refresh the page, make sure it persisted. Also make sure the email did not revert from its previous value.
8. Change the alternative email to the same one as the primary email, an error would be shown.
9. Change the select to some other value, make sure it persists, refresh the page, it should still have persisted. (no issue marked for this, treating it as enhancement).
10. Change your primary email back to hotmail or gmail but different from alternative, notice that the select will now be hidden, and no validation errors are present.
11. Place the order. In the order confirmation screen, make sure `What is your main role at your company?` is not present because it was hidden. (https://github.com/woocommerce/woocommerce/issues/56996)

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Fix logic bugs in additional fields validation and persisting logic.
</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/form/form.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/form/form.tsx
@@ -121,17 +121,17 @@ const Form = <
 
 		// Previous errors are cleared when they're no longer present.
 		if ( previousErrors ) {
+			const errorsToClear: string[] = [];
 			Object.entries( previousErrors ).forEach( ( [ key ] ) => {
 				const inputRef = inputsRef.current[ key ];
 
 				// If error was previously set but no longer exists, clear it.
 				if ( ! ( key in errors ) ) {
-					dispatch( validationStore ).clearValidationError(
-						`${ addressType }_${ key }`
-					);
+					errorsToClear.push( `${ addressType }_${ key }` );
 					inputRef?.setErrorMessage( '' );
 				}
 			} );
+			dispatch( validationStore ).clearValidationErrors( errorsToClear );
 		}
 	}, [ errors, previousErrors, addressType, values ] );
 
--- a/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/form/form.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/form/form.tsx
@@ -25,7 +25,7 @@ import {
 } from '@woocommerce/settings';
 import { isNull } from '@woocommerce/types';
 import { useInstanceId } from '@wordpress/compose';
-import { dispatch } from '@wordpress/data';
+import { dispatch, select } from '@wordpress/data';
 import { useEffect, useRef } from '@wordpress/element';
 import { decodeEntities } from '@wordpress/html-entities';
 import isShallowEqual from '@wordpress/is-shallow-equal';
@@ -103,21 +103,28 @@ const Form = <
 				return;
 			}
 			inputRef?.setErrorMessage( error );
+			const hasValidationError = select(
+				validationStore
+			).getValidationError( `${ addressType }_${ key }` );
+
+			// Check if this field already has a validation error, prevents up from surfacing already hidden errors.
+			if ( hasValidationError ) {
+				return;
+			}
 			dispatch( validationStore ).setValidationErrors( {
 				[ `${ addressType }_${ key }` ]: {
 					message: error,
 					hidden: !! inputRef?.isFocused(),
 				},
 			} );
-			if ( ! inputRef?.isFocused() ) {
-				inputRef?.revalidate();
-			}
 		} );
 
+		// Previous errors are cleared when they're no longer present.
 		if ( previousErrors ) {
 			Object.entries( previousErrors ).forEach( ( [ key ] ) => {
 				const inputRef = inputsRef.current[ key ];
-				// If error was previously set but is now cleared
+
+				// If error was previously set but no longer exists, clear it.
 				if ( ! ( key in errors ) ) {
 					dispatch( validationStore ).clearValidationError(
 						`${ addressType }_${ key }`
--- a/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/form/form.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/form/form.tsx
@@ -131,7 +131,11 @@ const Form = <
 					inputRef?.setErrorMessage( '' );
 				}
 			} );
-			dispatch( validationStore ).clearValidationErrors( errorsToClear );
+			if ( errorsToClear.length ) {
+				dispatch( validationStore ).clearValidationErrors(
+					errorsToClear
+				);
+			}
 		}
 	}, [ errors, previousErrors, addressType, values ] );
 
--- a/plugins/woocommerce/client/blocks/assets/js/data/checkout/push-changes.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/data/checkout/push-changes.ts
@@ -14,7 +14,7 @@ import { STORE_KEY as CHECKOUT_STORE_KEY } from './constants';
 import { STORE_KEY as PAYMENT_STORE_KEY } from '../payment/constants';
 import { processErrorResponse } from '../utils';
 import { CheckoutPutData } from './types';
-import { validateAdditionalFields } from './utils';
+import { hasValidationError, validateAdditionalFields } from './utils';
 
 // This is used to track and cache the local state of push changes.
 const localState = {
@@ -83,14 +83,30 @@ const updateCheckoutData = (): void => {
 		localState.doingPush = false;
 		return;
 	}
-
 	// Figure out which additional fields have changed and only send those to the server
 	const changedFields = Object.keys( newCheckoutData.additionalFields )
 		.filter( ( key ) => {
-			return (
-				localState.checkoutData.additionalFields[ key ] !==
+			// Fields with errors should be ignored
+			if ( hasValidationError( key ) ) {
+				return false;
+			}
+
+			// Fields that are not present in the original checkout data and have an empty value should be ignored (happens when a field is hidden on mount).
+			if (
+				! ( key in localState.checkoutData.additionalFields ) &&
+				newCheckoutData.additionalFields[ key ] === ''
+			) {
+				return false;
+			}
+
+			// Fields that have not changed should be ignored
+			if (
+				localState.checkoutData.additionalFields[ key ] ===
 				newCheckoutData.additionalFields[ key ]
-			);
+			) {
+				return false;
+			}
+			return true;
 		} )
 		.reduce( ( acc: AdditionalValues, key ) => {
 			acc[ key ] = newCheckoutData.additionalFields[ key ];
--- a/plugins/woocommerce/client/blocks/assets/js/data/checkout/utils.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/data/checkout/utils.ts
@@ -249,25 +249,30 @@ export const validateAdditionalFields = (
 	// Check each additional field for validation errors
 	// The validation store prefixes ads a prefix depending on the field location
 	for ( const fieldKey of Object.keys( additionalFields ) ) {
-		let prefix = '';
-		if ( CONTACT_FORM_KEYS.includes( fieldKey as keyof ContactForm ) ) {
-			prefix = 'contact_';
-		} else if (
-			ORDER_FORM_KEYS.includes( fieldKey as keyof AdditionalValues )
-		) {
-			prefix = 'order_';
-		} else {
-			return false;
-		}
-
-		const error = select( validationStore ).getValidationError(
-			`${ prefix }${ fieldKey }`
-		);
-
-		if ( error && ! error.hidden ) {
+		if ( hasValidationError( fieldKey ) ) {
 			return false;
 		}
 	}
 
 	return true;
+};
+
+export const hasValidationError = ( fieldKey: string ) => {
+	let prefix = '';
+	if ( CONTACT_FORM_KEYS.includes( fieldKey as keyof ContactForm ) ) {
+		prefix = 'contact_';
+	} else if (
+		ORDER_FORM_KEYS.includes( fieldKey as keyof AdditionalValues )
+	) {
+		prefix = 'order_';
+	} else {
+		return false;
+	}
+	const error = select( validationStore ).getValidationError(
+		`${ prefix }${ fieldKey }`
+	);
+	if ( error ) {
+		return true;
+	}
+	return false;
 };
--- a/plugins/woocommerce/client/blocks/assets/js/data/checkout/utils.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/data/checkout/utils.ts
@@ -22,7 +22,13 @@ import {
 } from '@wordpress/data/build-types/types';
 import { checkoutStore } from '@woocommerce/block-data';
 import { select } from '@wordpress/data';
-import type { AdditionalValues, ContactForm } from '@woocommerce/settings';
+import type {
+	ContactForm,
+	ContactFormValues,
+	FormFields,
+	OrderForm,
+	OrderFormValues,
+} from '@woocommerce/settings';
 
 /**
  * Internal dependencies
@@ -242,9 +248,7 @@ export const hasValidationError = ( fieldKey: string ) => {
 	let prefix = '';
 	if ( CONTACT_FORM_KEYS.includes( fieldKey as keyof ContactForm ) ) {
 		prefix = 'contact_';
-	} else if (
-		ORDER_FORM_KEYS.includes( fieldKey as keyof AdditionalValues )
-	) {
+	} else if ( ORDER_FORM_KEYS.includes( fieldKey as keyof OrderForm ) ) {
 		prefix = 'order_';
 	} else {
 		return false;
--- a/plugins/woocommerce/client/blocks/assets/js/data/checkout/utils.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/data/checkout/utils.ts
@@ -244,7 +244,9 @@ export const getPaymentResultFromCheckoutResponse = (
 	return paymentResult;
 };
 
-export const hasValidationError = ( fieldKey: string ) => {
+export const hasValidationError = (
+	fieldKey: keyof ContactForm | keyof OrderForm
+) => {
 	let prefix = '';
 	if ( CONTACT_FORM_KEYS.includes( fieldKey as keyof ContactForm ) ) {
 		prefix = 'contact_';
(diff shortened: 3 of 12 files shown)


plugins/woocommerce/client/blocks/assets/js/utils/schema-parser/index.ts
```diff
@@ -9,13 +9,17 @@ const ajv = new Ajv( {
...
+	validateFormats: true,
 	strictSchema: false,
 	strict: false,
 	messages: true,
 } );
 
-addFormats( ajv );
+addFormats( ajv, {
+	mode: 'fast',
+	formats: [ 'date', 'time', 'email', 'uri' ],
```
senadir: right now those are the only validation we will support, we can support more later but didn't see a reason for.

plugins/woocommerce/src/StoreApi/Schemas/V1/CheckoutSchema.php
```diff
@@ -243,7 +243,7 @@ protected function get_checkout_response( \WC_Order $order, PaymentResult $payme
 			'shipping_address'   => (object) $this->shipping_address_schema->get_item_response( $order ),
 			'payment_method'     => $order->get_payment_method(),
 			'payment_result'     => $payment_result,
-			'additional_fields'  => $this->get_additional_fields_response( $order ),
+			'additional_fields'  => (object) $this->get_additional_fields_response( $order ),
```
senadir, line 246: when no fields are present, this would be an array, causing logic issues down the road.

plugins/woocommerce/src/StoreApi/Schemas/V1/CheckoutSchema.php
```diff
@@ -336,7 +336,7 @@ function ( $option ) {
 					},
 					$field['options']
 				);
-				if ( true !== $field['required'] ) {
+				if ( true !== $field['required'] || $this->additional_fields_controller->is_conditional_field( $field ) ) {
```
senadir, line 339: Basically also allow an `""` for conditional selects, this allows something to be hidden and unset.

tjcafferkey:
Really appreciate the thorough testing instructions @senadir, thanks for that. They're a great learning opportunity for those new to the codebase especially since we have two new team members.

I'm yet to review the code changes but reporting that each step except the last one worked as you described. Looks like the Select input value was still present on the order confirmation screen even though it was hidden on the checkout screen.

CleanShot 2025-04-14 at 14 32 26

Also, do we already have test coverage for this functionality or is that still to come?

Thanks!

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/form/form.tsx
```diff
@@ -101,21 +101,28 @@ const Form = <
...
 
+		// Previous errors are cleared when they're no longer present.
 		if ( previousErrors ) {
 			Object.entries( previousErrors ).forEach( ( [ key ] ) => {
 				const inputRef = inputsRef.current[ key ];
-				// If error was previously set but is now cleared
+
+				// If error was previously set but no longer exists, clear it.
 				if ( ! ( key in errors ) ) {
 					dispatch( validationStore ).clearValidationError(
```
tjcafferkey: Nit: I notice we have [clearValidationErrors](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/data/validation/actions.ts#L19-L27) which could reduce the number of dispatches.

plugins/woocommerce/client/blocks/assets/js/data/checkout/push-changes.ts
```diff
@@ -83,14 +83,30 @@ const updateCheckoutData = (): void => {
 		localState.doingPush = false;
 		return;
 	}
-
 	// Figure out which additional fields have changed and only send those to the server
 	const changedFields = Object.keys( newCheckoutData.additionalFields )
```
tjcafferkey, line 87: Not specific to the changes here but just reading `updateCheckoutData` I noticed that if validation fails [we set the local cache, to be the current state](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/data/checkout/push-changes.ts#L107-L111). However, [if the check on `activePaymentMethod` fails, then we don't](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/data/checkout/push-changes.ts#L82-L85) update the local cache at all. Just curious as to why?

plugins/woocommerce/client/blocks/assets/js/data/checkout/utils.ts
```diff
@@ -249,25 +249,30 @@ export const validateAdditionalFields = (
...
 
 	return true;
 };
+
+export const hasValidationError = ( fieldKey: string ) => {
+	let prefix = '';
+	if ( CONTACT_FORM_KEYS.includes( fieldKey as keyof ContactForm ) ) {
+		prefix = 'contact_';
+	} else if (
+		ORDER_FORM_KEYS.includes( fieldKey as keyof AdditionalValues )
```
tjcafferkey: I'm getting a TS error for this line

plugins/woocommerce/src/StoreApi/Schemas/V1/CheckoutSchema.php
tjcafferkey: why dont we just cast to `(object)` in the method prior to returning? would probably provide more clarity in the code?

senadir: Common pattern to just cast at the schema level (last step) to avoid any short-circuit doing something else, but that function only has one returned, moved it there.

plugins/woocommerce/client/blocks/assets/js/data/checkout/push-changes.ts
senadir: Sadly I'm not sure, this was written by Alex, my best guess is that selecting payment failing is shouldn't update the fields cache, as they would trigger from different paths, I will digg more into it later on.

mikejolley:
These issues were blocking me from adding correct email validation in https://github.com/woocommerce/woocommerce/issues/56993 so I've pushed that work here instead. I've added JS tests for the change to email validation. I will update the PR description also.

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/form/form.tsx
mikejolley: The loop seems necessary here to call `inputRef?.setErrorMessage( '' )` for each error as well.

mikejolley:
The final testing step is failing due to `$this->additional_fields_controller->sync_order_additional_fields_with_customer( $order, wc()->customer );`. I am investigating.

mikejolley:
> Really appreciate the thorough testing instructions @senadir, thanks for that. They're a great learning opportunity for those new to the codebase especially since we have two new team members.
> 
> I'm yet to review the code changes but reporting that each step except the last one worked as you described. Looks like the Select input value was still present on the order confirmation screen even though it was hidden on the checkout screen.
> 
> CleanShot 2025-04-14 at 14 32 26

Fixed in https://github.com/woocommerce/woocommerce/pull/57037/commits/4a222546c7ab5f1b2336e20539109efdf0e34220

tjcafferkey:
@senadir @mikejolley just checking these testing steps are correct/up to date with the changes as I'm seeing some inconsistencies:

**Step 2**: Where it says specifically I shouldn't see the select input, I am doing:

CleanShot 2025-04-28 at 14 29 08

**Step 4**: Where it says it shouldn't send a request to the server, it is doing:

CleanShot 2025-04-28 at 14 36 49

I didn't test any further just to make sure I'm either missing something, or the steps/code need reviewing.

tjcafferkey:
@mikejolley this tests fine apart from step 2 per[ my comment here](https://github.com/woocommerce/woocommerce/pull/57037#issuecomment-2835353813) which still occurs.

According to the testing steps I shouldn't initially see the select box upon loading the checkout, but I do.

mikejolley:
@tjcafferkey updated testing steps after confirming with Nadir it should be visible on entry.

tjcafferkey (approved):
Thanks for addressing my feedback/comments. Considering the update in testing steps this now LGTM.

PR #57053, opened 2025-04-06
[Product Gallery (Beta) Block]: Correct the visibility of the gallery slider arrows

### LINKED ISSUE #57047: Product Gallery (beta): "next" button enabled when product image identical to only gallery image
In the new Product Gallery block, if a product has a product image and a single gallery image, both of which point to the same image, the "next" arrow button is enabled (clickable) when the page first loads.

I've confirmed the bug on a fresh WordPress installation running WordPress 6.7.2 and WooCommerce 9.7.1.

Image

https://github.com/user-attachments/assets/882fd632-3032-45c4-86d0-c54562b8dc56

aatanasovdev:

Closes: https://github.com/woocommerce/woocommerce/issues/57047

| Before | After |
| ------ | ----- |
|  before |  after |

1. Ensure the `Product Gallery (Beta)` block is added to the Product template in the Site Editor as shown in the **Site Editor** screenshot below. 
2. Go to edit a product. 
3. Set a featured image.
4. Set the same image to the gallery as shown in the **Front-end** screenshot.
5. The slider arrows shouldn't be visible on the front-end.
6.  Set a different gallery image, the arrows should be visible.

**Site Editor**
product-category-site-editor

**Front-end**
back-end-images

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

Ensure the arrows of the Product Gallery slider don't appear when the featured image and the gallery image are the same.

</details>


Konamiman:
Hi @aatanasovdev, thanks for your contribution. Could you please extend the testing instructions to include how to configure the frontend sot hat the block is used? Thank you!

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryLargeImageNextPrevious.php
```diff
@@ -55,6 +55,12 @@ protected function render( $attributes, $content, $block ) {
 			return null;
 		}
 
+		$featured_image_id = $product->get_image_id();
+
+		if( $featured_image_id && count( $product_gallery) === 1 && (int) $product_gallery[0] === (int) $featured_image_id ) {
+			return null;
+		}
+
```
kmanijak, line 61: While it does address the case that was raised I think we could provide more generic solution to the problem. Thumbnails react correctly to the number of images as the case of overlap between featured image and gallery images is solved in utils used in other blocks.

I suggest we go this way and utilise the util so all the inner blocks share the same logic of getting images.

This is suggestion to replace lines 52-63 (GitHub doesn't allow marking suggestions to unchanged lines)

```suggestion
		$product_gallery_data       = ProductGalleryUtils::get_product_gallery_image_data( $product );
		$product_gallery_images  = $product_gallery_data['images'];
		// Don't show the arrows block if there is only one image.
		if ( count( $product_gallery_images ) <= 1 ) {
			return '';
		}
```

`get_product_gallery_image_data` does all the things to return unique visible images so that would be a nicer solution 🙌

kmanijak (changes requested):
Thanks a lot for your contribution! I left a comment suggesting more generic solution 🙌

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryLargeImageNextPrevious.php
aatanasovdev: @kmanijak, I agree, the suggested solution is better. I've implemented it.

aatanasovdev:
Hi @Konamiman and @kmanijak,

> Hi @aatanasovdev, thanks for your contribution. Could you please extend the testing instructions to include how to configure the frontend sot hat the block is used? Thank you!

Sure, I've updated the description and provided a screenshot. 

> Thanks a lot for your contribution! I left a comment suggesting more generic solution 🙌

Thanks, the feedback has been addressed.

kmanijak (approved):
Thanks for addressing the feedback! I tried multiple combinations and it works as expected:
- ✅ only featured image (result: no arrows, single image)
- ✅ featured image with the same image in gallery (result: no arrows, single image)
- ✅ featured image with the same image in gallery and other image (result: arrows appear, two images show up)
- ✅ no featured image with image in gallery (result: no arrows, single image)
- ✅ no featured image with multiple images in gallery (result: arrows appear, multiple images show up)
- ☑️ no images (result: placeholder appears but it has limited size, I think it should fit the container but it's not the scope of this PR - images below)

With some image (see the container size) | Without image (smaller placeholder)
-- | --
image | image

PR #57062, opened 2025-04-07
MIC: Create base infrastructure and DB tables for Back in Stock Notifications

### LINKED ISSUE #57031: MIC: Create base infrastructure and DB tables for Back in Stock Notifications
The goal of this issue is to:

* Introduce the base infrastructure as a Service Provider in Core and;
* Create new DB tables to accommodate stock notifications, metadata, and activity logs.
* Create a new WC_Data model for the stock notifications entity.
* Created dedicated data stores for notifications, their metadata, and logs.
* Write unit tests.

### Details on database tables

We aim to draw significant inspiration from the Back In Stock Notifications extension. However, this doesn't mean we should carry over the structure as it currently is; our goal should be to achieve better performance and eliminate the need to support multiple "types" of notifications, as the extension currently accommodates.

xristos3490:

Closes #57031

This PR is part of a broader effort to introduce back-in-stock notifications. It establishes the essential data models needed to manage customer notifications for restocked products.

## Database

Two new DB tables are introduced:
- `wc_stock_notifications`: Main table for storing notifications.
- `wc_stock_notificationmeta`: Custom metadata table.

## Specification

- The DB schema is stored in the main datastore and is running unconditionally into the `WC_Install::install()` method.
- A new `WC_Data` model named "Notification" will handle the notifications in the code.
- The notifications data store receives one sub-datastore for metadata through DI. It is exposed in the container to hook it to the available datastores.
- A controller class is also exposed in the container to manage the initialization of hooks (i.e., hooking the datastore).
- The notification datastore feature a `query()` method that constructs an SQL command using arguments.
- All classes are managed by the `StockNotificationsServiceProvider`.

### Query Performance

The DB indexes on the main table are created based on the current flows of the Back In Stock Notifications extension. Generally, the most common arguments for querying notifications would be via:
- product_status_attempt (product_id, status, date_last_attempt_gmt, id): Crucial for batch processor
_Having id at the end helps with ORDER BY id and makes it a covering index for SELECT id._
- user_lookup (user_id, product_id, status): For registered users checking "my active subscriptions for product X."
- email_lookup (user_email(100), product_id, status): For guests, or "is this email already subscribed to product X with active status?"

Indexes may change as we advance into specific areas of the feature in future PRs.

### Notes/Open Questions

- Should the notifications data model be public like `WC_Stock_Notification()` looks easier and could be accompanied with helpers like `wc_get_stock_notification(s)()`. This feels unclear, so one reason I chose an internal class for now is to reduce potential conflicts. However, introducing some 3PD hooks here would essentially expose internal class instances.

- Unit tests for CRUD flows.
- A code review, particularly regarding the DB footprint and performance, would be appreciated.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsTableDataStore.php
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	/**
+	 * Get the stock notifications meta table name.
+	 *
+	 * @return string
+	 */
+	public static function get_meta_table_name() {
+		global $wpdb;
+		return $wpdb->prefix . 'woocommerce_stock_notificationmeta';
```
peterfabian: I remember we ran into some issues historically with table names being too long, so it would be better to switch to `wc` instead of `woocommerce` in the table names.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -0,0 +1,384 @@
...
+		$max_index_length         = $this->database_util->get_max_index_length();
+
+		$sql = "
+CREATE TABLE $notifications_table_name (
+	id bigint(20) unsigned NOT NULL AUTO_INCREMENT,
+	product_id bigint(20) unsigned NOT NULL,
+	user_id bigint(20) unsigned NOT NULL,
+	user_email varchar(100) NOT NULL,
+	status varchar(20) NOT NULL DEFAULT 'pending',
+	date_created timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
peterfabian: I know it's a timestamp, so kinda explicit already, but maybe we can add `_gmt` to these columns to make it explicit in the context of WP that it's a GMT datetime.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -0,0 +1,384 @@
...
+	KEY meta_key (meta_key($max_index_length))
+) $collate;
+CREATE TABLE $logs_table_name (
+	id bigint(20) unsigned NOT NULL AUTO_INCREMENT,
+	notification_id bigint(20) unsigned NOT NULL,
+	action varchar(200) NOT NULL,
+	user_id bigint(20) unsigned NOT NULL,
+	user_email varchar(100) NOT NULL,
+	ip_address VARCHAR(45) NULL,
+	date_logged timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
peterfabian: Potentially the same here with `_gmt`.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -0,0 +1,384 @@
...
+
+namespace Automattic\WooCommerce\Internal\DataStores\StockNotifications;
+
+use Automattic\WooCommerce\Internal\Utilities\DatabaseUtil;
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * The Stock Notifications Data Store.
+ */
+class StockNotificationsDataStore implements \WC_Object_Data_Store_Interface {
```
peterfabian, line 18: I thought about the data store(s) a bit more and thought we might be able to get the best of both worlds:
- nice abstraction and encapsulation, and
- separation of concerns  
 
...if we used an umbrella data store as an abstraction for the stock notifications and dependency-inject 2 sub-data store classes for meta and log. Wdyt?

peterfabian (commented):
Looking good so far, just wanted to add a couple of thoughts while this is fresh.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
xristos3490: Completely agree; the primary data store is becoming unnecessarily large.

xristos3490:
Hey @peterfabian! I have addressed the feedback, created the data models, and added unit tests. I also updated the description to capture the changes in this PR! Let me know what you think :)

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -0,0 +1,534 @@
...
+
+	/**
+	 * Get the stock notifications logs table name.
+	 *
+	 * @return string
+	 */
+	public static function get_logs_table_name(): string {
+		global $wpdb;
+		return $wpdb->prefix . 'wc_stock_notifications_logs';
+	}
```
peterfabian: Interesting decision to include these methods here. Why not encapsulate them into individual data stores? Perhaps also define db schema for each table in the respective data store etc?

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -0,0 +1,534 @@
...
+		global $wpdb;
+
+		if ( 0 === $notification->get_id() ) {
+			return new \WP_Error( 'invalid_stock_notification', 'Invalid notification ID.' );
+		}
+
+		$changes = $notification->get_changes();
+		$result  = 0;
+
+		if ( array_intersect( array( 'product_id', 'user_id', 'user_email', 'status', 'is_queued', 'date_modified', 'date_subscribed', 'date_notified' ), array_keys( $changes ) ) ) {
```
peterfabian: Perhaps a helper method/var like `notification_changed` might help with readability.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -0,0 +1,534 @@
...
+	 */
+	public function delete( &$notification, $args = array() ) {
+		global $wpdb;
+
+		$wpdb->delete( $this->get_table_name(), array( 'id' => $notification->get_id() ), array( '%d' ) );
+
+		$wpdb->delete( $this->get_meta_table_name(), array( 'notification_id' => $notification->get_id() ), array( '%d' ) );
+
+		$wpdb->delete( $this->get_logs_table_name(), array( 'notification_id' => $notification->get_id() ), array( '%d' ) );
+	}
```
peterfabian: This could be also handed off to the sub-data-store classes.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -0,0 +1,534 @@
...
+	 */
+	private function after_meta_change( &$notification ): bool {
+
+		$current_time      = time();
+		$current_date_time = new \WC_DateTime( "@$current_time", new \DateTimeZone( 'UTC' ) );
+
+		$should_save =
+			$notification->get_id() > 0
+			&& $notification->get_date_modified( 'edit' ) < $current_date_time
+			&& empty( $notification->get_changes() );
```
peterfabian, line 393: Can you expand on the logic here a bit? Do you assume that if there are any changes, someone else will call `save` on the instance of Notification?

peterfabian (commented):
This is looking pretty good so far! 
Adding my initial thoughts so we can get the conversation going.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
xristos3490: > Do you assume that if there are any changes, someone else will call save on the instance of Notification?

Exactly! The rationale here is:
- Changes in the notification object will effectively require a `save()`, or;
- If there are no changes, it means that meta has changed, and thus the last modified date needs updating. 

A possible drawback is that if someone seeking to update a list of props and meta starts by setting or updating the meta before another prop changes, the system will perform two saves.

It's inspired by the HPOS metadata handling [here](https://github.com/woocommerce/woocommerce/blob/13b8feed35c4f204a1d9c84860336738a05afee1/plugins/woocommerce/src/Internal/DataStores/Orders/OrdersTableDataStore.php#L3338).

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
xristos3490: Considering that this isn't used elsewhere, and more importantly, the check below for the date_modified is somewhat relevant, I'm inclined to leave this as it is. Any objections?

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
xristos3490: Good point! 

I have moved the table names to each data store here https://github.com/woocommerce/woocommerce/pull/57062/commits/cfd37341c84c71eaf5a5bcae6526e6c6c82fe7d7

As for the schema, to move this I'd need to bring the database utility instance into the metadata data store because it requires the max index size for the meta_key index. Considering the complexity of adding a DI or just a parameter, I'm inclined to leave the schema as it is, similar to [the HPOS one](https://github.com/woocommerce/woocommerce/blob/ece5ba3ff7c381c30a73c42e634595ea72f1490d/plugins/woocommerce/src/Internal/DataStores/Orders/OrdersTableDataStore.php#L3131). On the contrary, I could say it is better for readability this way. Thoughts?

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
xristos3490: Great point! Updated here https://github.com/woocommerce/woocommerce/pull/57062/commits/5fa91854e7b0113fc79adefb5396b41bdf364e76

plugins/woocommerce/tests/php/src/Internal/StockNotifications/NotificationTests.php
```diff
@@ -0,0 +1,34 @@
+<?php // phpcs:ignore Suin.Classes.PSR4
```
kalessil: I'm curious to know why we disable this rule here.

plugins/woocommerce/tests/php/src/Internal/StockNotifications/FactoryTests.php
```diff
@@ -0,0 +1,30 @@
+<?php // phpcs:ignore Suin.Classes.PSR4
```
kalessil: Same here.

plugins/woocommerce/tests/php/src/Internal/DataStores/StockNotifications/StockNotificationsDataStoreTests.php
```diff
@@ -0,0 +1,535 @@
...
+		$notification_2->set_product_id( 2 );
+		$notification_2->set_user_id( 2 );
+		$notification_2->save();
+
+		$notifications = $this->data_store->query(
+			array(
+				'product_id' => 1,
+			)
+		);
+		$this->assertEquals( 1, count( $notifications ) );
```
kalessil: Nit: PHPUnit has `assertCount` method.

kalessil (commented):
I don't have deep context here, so leaving some generic comments and nits.

plugins/woocommerce/tests/php/src/Internal/DataStores/StockNotifications/StockNotificationsDataStoreTests.php
xristos3490: Nice catch! Done in https://github.com/woocommerce/woocommerce/pull/57062/commits/c67f1333cdc4097033bcb82e016d03300e3e87f3

plugins/woocommerce/tests/php/src/Internal/StockNotifications/NotificationTests.php
xristos3490: Thanks for noticing; these aren't needed, actually, and it's a leftover from [an issue](https://github.com/suin/phpcs-psr4-sniff/pull/6) I had with my local suin sniff in pre-commit hooks. Removed here https://github.com/woocommerce/woocommerce/pull/57062/commits/b078debd25f0608e62613fdd1e9a7da42eb00597

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsActivityLogsDataStore.php
```diff
@@ -0,0 +1,149 @@
...
+		return $wpdb->prefix . 'wc_stock_notifications_logs';
+	}
+
+	/**
+	 * Create an activity log.
+	 *
+	 * @param array $args Additional arguments.
+	 * @return int|false The log ID or false if the log was not created.
+	 */
+	public function create( $args ) {
```
peterfabian: I think return types and arg types should be added to methods where possible (i.e. except for union types that are only supported in PHP 8.0+).

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -0,0 +1,532 @@
...
+ * The Stock Notifications Data Store.
+ */
+class StockNotificationsDataStore implements \WC_Object_Data_Store_Interface {
+
+	/**
+	 * The database util object to use.
+	 *
+	 * @var DatabaseUtil
+	 */
+	protected $database_util;
```
peterfabian: You can add type declarations for better type safety.

plugins/woocommerce/tests/php/src/Internal/DataStores/StockNotifications/StockNotificationsActivityLogsDataStoreTests.php
```diff
@@ -0,0 +1,294 @@
...
+
+/**
+ * Class StockNotificationsActivityLogsDataStoreTests.
+ */
+class StockNotificationsActivityLogsDataStoreTests extends \WC_Unit_Test_Case {
+
+	/**
+	 * The data store instance.
+	 *
+	 * @var StockNotificationsActivityLogsDataStore
```
peterfabian: This tests the main data store, not the Activity Log store, right?

plugins/woocommerce/tests/php/src/Internal/DataStores/StockNotifications/StockNotificationsDataStoreTests.php
```diff
@@ -0,0 +1,535 @@
...
+
+		// Set all properties.
+		$notification->set_product_id( 1 );
+		$notification->set_user_id( 1 );
+		$notification->set_user_email( 'test@test.com' );
+		$notification->set_status( 'active' );
+		$notification->set_date_created( '2024-01-01 00:00:00' );
+		$notification->set_date_modified( '2024-01-01 00:00:00' );
+		$notification->set_date_subscribed( '2024-01-01 00:00:00' );
+		$notification->set_date_notified( '2024-01-01 00:00:00' );
```
peterfabian: Maybe worth using different dates here to ensure they're not mixed up?

plugins/woocommerce/tests/php/src/Internal/DataStores/StockNotifications/StockNotificationsDataStoreTests.php
```diff
@@ -0,0 +1,535 @@
...
+	 * Test validation requirements for creating a notification.
+	 */
+	public function test_create_notification_validation() {
+		$notification = new Notification();
+
+		// Test missing product_id.
+		$result = $notification->save();
+		$this->assertInstanceOf( \WP_Error::class, $result );
+		$this->assertEquals( 'stock_notification_validation_error', $result->get_error_code() );
+		$this->assertEquals( 'Product ID is required', $result->get_error_message() );
```
peterfabian: Not a biggie, but will this work with different language setting?

plugins/woocommerce/tests/php/src/Internal/DataStores/StockNotifications/StockNotificationsDataStoreTests.php
```diff
@@ -0,0 +1,535 @@
...
+		$this->assertEquals( 2, $notification->get_user_id() );
+		$this->assertEquals( 'test2@test.com', $notification->get_user_email() );
+		$this->assertEquals( 'active', $notification->get_status() );
+		$this->assertEquals( '2024-01-02 00:00:00', $notification->get_date_subscribed()->format( 'Y-m-d H:i:s' ) );
+		$this->assertEquals( '2024-01-02 00:00:00', $notification->get_date_notified()->format( 'Y-m-d H:i:s' ) );
+		$this->assertTrue( $notification->is_queued() );
+
+		// Verify modified date is updated.
+		$this->assertEquals( '2024-01-01 00:00:00', $notification->get_date_created()->format( 'Y-m-d H:i:s' ) );
+		$this->assertNotEquals( '2024-01-01 00:00:00', $notification->get_date_modified()->format( 'Y-m-d H:i:s' ) );
```
peterfabian, line 184: This is slightly confusing to me. I understand that in the first save, the date modified is retained because it was explicitly updated, but shouldn't it change to the time of the second save (on line 160)?

plugins/woocommerce/tests/php/src/Internal/DataStores/StockNotifications/StockNotificationsDataStoreTests.php
```diff
@@ -0,0 +1,535 @@
...
+		$notification = new Notification();
+		$notification->set_product_id( 1 );
+		$notification->set_user_id( 1 );
+		$notification->add_meta_data( 'test_meta', 'test_value' );
+		$notification->save();
+
+		$notification = new Notification( $notification->get_id() );
+		$this->assertEquals( 'test_value', $notification->get_meta( 'test_meta' ) );
+		$notification->delete_meta_data( 'test_meta' );
+		$notification->save();
```
peterfabian, line 277: Might be good to check non-existence also before re-reading it below.

plugins/woocommerce/tests/php/src/Internal/DataStores/StockNotifications/StockNotificationsDataStoreTests.php
```diff
@@ -0,0 +1,535 @@
...
+
+		$notification = new Notification( $notification->get_id() );
+		$this->assertEquals( '2024-01-01 00:00:00', $notification->get_date_modified()->format( 'Y-m-d H:i:s' ) );
+		$this->assertEquals( '', $notification->get_meta( 'test_meta' ) );
+		$notification->add_meta_data( 'test_meta', 'test_value' );
+		$notification->save();
+
+		$notification = new Notification( $notification->get_id() );
+		$this->assertNotEquals( '2024-01-01 00:00:00', $notification->get_date_modified()->format( 'Y-m-d H:i:s' ) );
```
peterfabian, line 302: This actually shows that modified date isn't updated when meta is added, which contradicts the comment above the method, no?

Same for update & delete.

peterfabian (commented):
Please see a bit more feedback below from the first pass of the code.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
peterfabian: Nice!

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
peterfabian: Not a big problem, let's leave it.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
peterfabian: I agree that this pattern is used also in HPOS, so probably fine to leave it as is. 

As a side note, wouldn't it be possible to create an `init` method taking `DatabaseUtil` as a parameter via DI in `StockNotificationsMetaDataStore` and remove it from the `StockNotificationsDataStore`? Or would the DI container not be able to handle it?

plugins/woocommerce/tests/php/src/Internal/DataStores/StockNotifications/StockNotificationsDataStoreTests.php
xristos3490: Yes, it should change. Notice that this checks the "Not-equals"; I could do a "check with delta" here, but I kept it simple.

plugins/woocommerce/tests/php/src/Internal/DataStores/StockNotifications/StockNotificationsDataStoreTests.php
xristos3490: Similar to above, this checks that the modified date is not equal to the previous one. Does this clarify things? 🤔

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsActivityLogsDataStore.php
xristos3490: Good call! I added them [here](https://github.com/woocommerce/woocommerce/pull/57062/commits/ce0a0942efbc40c3023bcb9290b9efaebe3e7267) and [here](https://github.com/woocommerce/woocommerce/pull/57062/commits/5c6c0133702a90ef6f29a36a7e6198e668b43a16), but not for every part, as the other data stores need to comply with the `WC_Object_Data_Store_Interface` and `CustomMetaDataStore` interfaces.

plugins/woocommerce/tests/php/src/Internal/DataStores/StockNotifications/StockNotificationsActivityLogsDataStoreTests.php
xristos3490: Yup, done in https://github.com/woocommerce/woocommerce/pull/57062/commits/252af8614fe2f1ca0dd791cf34418f30ea01c6be

plugins/woocommerce/tests/php/src/Internal/DataStores/StockNotifications/StockNotificationsDataStoreTests.php
xristos3490: Good call, added here https://github.com/woocommerce/woocommerce/pull/57062/commits/bf64d98486b08141396a0b9719ce58c1081b41d5

plugins/woocommerce/tests/php/src/Internal/DataStores/StockNotifications/StockNotificationsDataStoreTests.php
xristos3490: Yup, I think it's a good idea to avoid using texts here. Updated in https://github.com/woocommerce/woocommerce/pull/57062/commits/e9087ae936272a870ec067e562415ac5defa5864

plugins/woocommerce/tests/php/src/Internal/DataStores/StockNotifications/StockNotificationsDataStoreTests.php
xristos3490: Good call, added in https://github.com/woocommerce/woocommerce/pull/57062/commits/156b4580eade80e13961c22fcd0aa3e3729d34e7

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
xristos3490: Done in https://github.com/woocommerce/woocommerce/pull/57062/commits/5c6c0133702a90ef6f29a36a7e6198e668b43a16

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
xristos3490: Good question, I first thought the same honestly. 

It seems that the only way to DI into interfaces is to expose that interface _in the container_ and create the instance. I'm not entirely sure if we can add more dependencies here, as there isn't another case where a DI is injected with another DI; it seems like it's not the conventional approach in the container.

plugins/woocommerce/tests/php/src/Internal/DataStores/StockNotifications/StockNotificationsDataStoreTests.php
peterfabian: Certainly! Apologies, I missed the **Not**Equals part 🤦‍♂️

plugins/woocommerce/tests/php/src/Internal/DataStores/StockNotifications/StockNotificationsDataStoreTests.php
peterfabian: Thanks, I somehow missed that. Then it makes sense.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
peterfabian: Since we're going with a different approach, let's move on, there are more important aspect here that come later.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsActivityLogsDataStore.php
peterfabian: Thanks, makes sense.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsActivityLogsDataStore.php
```diff
@@ -0,0 +1,149 @@
...
+	 * @param array $args Additional arguments.
+	 * @return int|false The log ID or false if the log was not created.
+	 */
+	public function create( array $args ) {
+		global $wpdb;
+
+		$table = $this->get_table_name();
+		$data  = array(
+			'notification_id' => absint( $args['notification_id'] ),
+			'action'          => sanitize_text_field( $args['action'] ),
```
prettyboymp: Do we need to sanitize these at this low of level?  Where will the caller of `Notification::log()` be getting this data?

prettyboymp (changes requested):
@xristos3490 I noticed the schema here doesn't match some of the latest p2 proposal.  Is this PR still in progress?

xristos3490:
Hey @prettyboymp! 

> I noticed the schema here doesn't match some of the latest p2 proposal. Is this PR still in progress?

That's correct! Considering the active discussion in the p2 proposal, I'm marking this as a draft for now to avoid confusion.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsActivityLogsDataStore.php
xristos3490: I agree this might be overkill. Logs should be added with hardcoded parameters during notification actions, so this may not be necessary at such a low level.

By the way, we are currently exploring the idea of dropping logs for i1, so I'll return here either to remove or fix this one :)

xristos3490: Returning here, we ended up removing the log table entirely.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -0,0 +1,487 @@
...
+		global $wpdb;
+
+		$args = wp_parse_args(
+			$args,
+			array(
+				'status'     => '',
+				'product_id' => array(),
+				'user_id'    => 0,
+				'user_email' => '',
+				'is_queued'  => '',
```
prettyboymp: Doesn't this need to be updated to reflect that `is_queued` is now rolled into `status`?

plugins/woocommerce/src/Internal/StockNotifications/Enums/NotificationStatus.php
```diff
@@ -0,0 +1,52 @@
...
+declare( strict_types = 1 );
+
+namespace Automattic\WooCommerce\Internal\StockNotifications\Enums;
+
+/**
+ * Enum class for all the notification statuses.
+ */
+final class NotificationStatus {
+	/**
+	 * The notification has been created but not yet confirmed.
```
prettyboymp: What do you think about expanding on these comments a little bit more?  What does confirmed mean?

prettyboymp (changes requested):
Schema changes look good, but it looks like some other changes are missing around the switch to status.

One thing to think about, not necessarily to be solved in this PR, is how we deal with processes that need to load multiple Notifications.  The current WC_Data constructor makes it so that every time an instance is created, it performs two separate read queries (main table and meta) to load its data.  This means that loading 10 Notifications requires 21 queries.  One to retrieve the initial IDs, then 2 for each of the 10 Notifications.

This is a problem with the current way WC_Data works that we need to address across WooCommerce, but I wanted to bring it up in case it is something easier to fix for Notifications specifically before we have to consider backward compatibility.

plugins/woocommerce/src/Internal/StockNotifications/Notification.php
```diff
@@ -0,0 +1,371 @@
...
+			$status = NotificationStatus::PENDING;
+		}
+
+		$this->set_prop( 'status', $status );
+	}
+
+	/**
+	 * Set the date created.
+	 *
+	 * @param string $date_created Date created.
```
prettyboymp: The type here says `string`, but it will accept other types and the datastore specifically uses `time() :int`. Can we update the docblocks to reflect that accepted types?

xristos3490:
Thanks for the review, @prettyboymp!

> The current WC_Data constructor makes it so that every time an instance is created, it performs two separate read queries (main table and meta) to load its data. This means that loading 10 Notifications requires 21 queries. One to retrieve the initial IDs, then 2 for each of the 10 Notifications.

Thanks for bringing this up! The truth is we followed the WC_Data interface for consistency with other Woo entities, but indeed, when fetching notifications in bulk, this can get ugly fast. I can think of cases like the admin list tables and the My Account page showing the user's current subscriptions, so we definitely need to improve. 

Although the WC_Data can be improved to bring benefits across all entities (that's a reason we opted for this interface in the first place), I found it interesting [how admin notes are working](https://github.com/woocommerce/woocommerce/blob/51f0e4da1fa1ff15209282058602d02db0e8ef4c/plugins/woocommerce/src/Admin/Notes/Note.php#L79-L83). I pushed an attempt [[here](https://github.com/woocommerce/woocommerce/pull/57062/commits/e6365ad1a9e581522571fe91599d464ce1c7ea4f)] that will reduce the queries for 10 notifications from 21 to 1. The WC_Data's "maybe_read_meta_data" can help here to lazy load meta only if needed. 

Having said that, I still see a problem concerning instantiating a large number of notifications, but this shouldn't happen, at least not in any core processes we are about to implement going forward. Curious to hear your thoughts.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
xristos3490: Yes! Apparently, I missed some more places, fixed in https://github.com/woocommerce/woocommerce/pull/57062/commits/a76be4e3e4cb289699658e2c85b8ade7884320c3

plugins/woocommerce/src/Internal/StockNotifications/Enums/NotificationStatus.php
xristos3490: Absolutely! Expanded here https://github.com/woocommerce/woocommerce/pull/57062/commits/8cddd736c37e83efa6f5ead07ceffef0667c5403

plugins/woocommerce/src/Internal/StockNotifications/Notification.php
xristos3490: Done in https://github.com/woocommerce/woocommerce/pull/57062/commits/5667dde7e83f2c5b66f2ba5afc50935b76dfac7a

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -0,0 +1,492 @@
...
+		if ( 0 === $notification->get_id() ) {
+			return new \WP_Error( 'invalid_stock_notification', 'Invalid notification ID.' );
+		}
+
+		$changes = $notification->get_changes();
+		$result  = 0;
+
+		if ( array_intersect( array( 'product_id', 'user_id', 'user_email', 'status', 'date_modified', 'date_confirmed', 'date_last_attempt', 'date_notified', 'date_cancelled', 'cancellation_source' ), array_keys( $changes ) ) ) {
+
+			if ( ! in_array( 'date_modified', array_keys( $changes ), true ) ) {
```
kalessil: Hint: `array_key_exists`

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -0,0 +1,492 @@
...
+		$current_date_time = new \WC_DateTime( "@$current_time", new \DateTimeZone( 'UTC' ) );
+
+		$should_save =
+			$notification->get_id() > 0
+			&& $notification->get_date_modified( 'edit' ) < $current_date_time
+			&& empty( $notification->get_changes() );
+
+		if ( $should_save ) {
+			$notification->set_date_modified( $current_time );
+			$notification->save();
```
kalessil: We should check what `$notification->save();` returned here and decide is we return true/false.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -0,0 +1,492 @@
...
+		// WHERE clauses.
+		$where        = array();
+		$where_values = array();
+		if ( $args['status'] ) {
+			$where[]        = 'status = %s';
+			$where_values[] = esc_sql( $args['status'] );
+		}
+
+		if ( ! empty( $args['product_id'] ) ) {
+			$product_ids  = array_map( 'absint', is_array( $args['product_id'] ) ? $args['product_id'] : array( $args['product_id'] ) );
```
kalessil: Nit: `is_array( $args['product_id'] ) ? $args['product_id'] : array( $args['product_id'] )` can be simplified to `(array) $args['product_id']` https://3v4l.org/cWr1N

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -0,0 +1,492 @@
...
+			return array();
+		}
+
+		if ( 'objects' === $args['return'] ) {
+			$notifications = array();
+			foreach ( $results as $result ) {
+				$notifications[] = new Notification( $result );
+			}
+
+			return $notifications;
```
kalessil: For the consistency we can refactor this block to something like:
```php
return array_map(
    function ( $result ) {
         ...
    },
    array_values ( $results )
);
```

plugins/woocommerce/src/Internal/StockNotifications/Controller.php
```diff
@@ -0,0 +1,33 @@
...
+declare( strict_types = 1 );
+
+namespace Automattic\WooCommerce\Internal\StockNotifications;
+
+use Automattic\WooCommerce\Internal\DataStores\StockNotifications\StockNotificationsDataStore;
+
+/**
+ * The controller for the stock notifications.
+ */
+class Controller {
```
kalessil: A bit confused by the naming, what's the thinking behind picking 'Controller here'?

plugins/woocommerce/src/Internal/StockNotifications/Enums/NotificationCancellationSource.php
```diff
@@ -0,0 +1,37 @@
...
+ * Notification cancellation source enum.
+ */
+class NotificationCancellationSource {
+
+	/**
+	 * Admin cancellation source.
+	 *
+	 * @var string
+	 */
+	const ADMIN = 'admin';
```
kalessil: We can add 'public' modifiers for both constants to clearly state our intentions.

plugins/woocommerce/src/Internal/StockNotifications/Enums/NotificationCancellationSource.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Automattic\WooCommerce\Internal\StockNotifications\Enums;
+
+/**
+ * Notification cancellation source enum.
+ */
+class NotificationCancellationSource {
```
kalessil: We normally mark the new enums as `final`.

plugins/woocommerce/src/Internal/StockNotifications/Enums/NotificationStatus.php
```diff
@@ -0,0 +1,62 @@
...
+final class NotificationStatus {
+
+	/**
+	 * Status: 'pending'.
+	 * Initial state when Double Opt-In (DOI) is active, awaiting user email verification.
+	 * Not eligible for "back in stock" notifications until confirmed.
+	 *
+	 * @var string
+	 */
+	const PENDING = 'pending';
```
kalessil: Similar here: can be `public`.

plugins/woocommerce/src/Internal/StockNotifications/Factory.php
```diff
@@ -0,0 +1,39 @@
...
+ */
+class Factory {
+
+	/**
+	 * Get the notification object.
+	 *
+	 * @param  int $notification_id Notification ID to get.
+	 * @return Notification|bool
+	 */
+	public static function get_notification( $notification_id ) {
```
kalessil: Guess we could add `int` type before `$notification_id`.

plugins/woocommerce/src/Internal/StockNotifications/Notification.php
```diff
@@ -0,0 +1,374 @@
...
+	| Setters
+	|--------------------------------------------------------------------------
+	*/
+
+	/**
+	 * Set the product ID.
+	 *
+	 * @param int $product_id Product ID.
+	 */
+	public function set_product_id( $product_id ) {
```
kalessil: For the setters in this class, where applicable, let's add types declarations for the parameters.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -0,0 +1,492 @@
...
+	user_id bigint(20) unsigned NOT NULL,
+	user_email varchar(100) NOT NULL,
+	status varchar(20) NOT NULL DEFAULT 'pending',
+	date_created_gmt datetime NULL,
+	date_modified_gmt datetime NULL,
+	date_confirmed_gmt datetime NULL,
+	date_last_attempt_gmt datetime NULL,
+	date_notified_gmt datetime NULL,
+	date_cancelled_gmt datetime NULL,
+	cancellation_source varchar(255) NULL,
```
kalessil: Since we have an enum for the possible values, can we revisit the column length?

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -0,0 +1,492 @@
...
+		$table_name       = $this->get_table_name();
+		$meta_table_name  = $this->get_meta_table_name();
+		$max_index_length = $this->database_util->get_max_index_length();
+
+		$sql = "
+CREATE TABLE $table_name (
+	id bigint(20) unsigned NOT NULL AUTO_INCREMENT,
+	product_id bigint(20) unsigned NOT NULL,
+	user_id bigint(20) unsigned NOT NULL,
+	user_email varchar(100) NOT NULL,
```
kalessil, line 87: Nice, like that this is consistent with the users table!

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -0,0 +1,492 @@
...
+	date_modified_gmt datetime NULL,
+	date_confirmed_gmt datetime NULL,
+	date_last_attempt_gmt datetime NULL,
+	date_notified_gmt datetime NULL,
+	date_cancelled_gmt datetime NULL,
+	cancellation_source varchar(255) NULL,
+	PRIMARY KEY  (id),
+	KEY product_status_attempt (product_id, status, date_last_attempt_gmt, id),
+	KEY user_lookup (user_id, product_id, status),
+	KEY email_lookup (user_email(100), product_id, status)
```
kalessil: Do we need `user_email(100)` here (perhaps just `user_email` will work)?

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -0,0 +1,492 @@
...
+	 * @param Notification $notification The data object to delete.
+	 * @param array        $args         Additional arguments.
+	 * @return void
+	 */
+	public function delete( &$notification, $args = array() ) {
+		global $wpdb;
+
+		$wpdb->delete( $this->get_table_name(), array( 'id' => $notification->get_id() ), array( '%d' ) );
+
+		$this->data_store_meta->delete_by_notification_id( $notification->get_id() );
```
kalessil: We should check what the previous deletion returns before executing this one.

kalessil (changes requested):
Great work @xristos3490, I like how the PR is evolving!

Leaving comments of mixed types (performance, reliability, type safety, and conventions) as the final review round.

plugins/woocommerce/src/Internal/StockNotifications/Controller.php
xristos3490: Totally understand your skepticism here! My aim is to create a single entry point for this entire feature right from this file, allowing us to have just one instance to call from the main class-woocommerce.php. [This follow-up PR](https://github.com/woocommerce/woocommerce/pull/58044/files#diff-a6a04d8818a9b2a33b30f1414861c51e7c3d67d593b34c2350387b2e09049971) introducing the first "sub" controller may help clarify what I have in mind.

I’m also thinking about renaming it to `Integration.php`, much like how [the EmailEditor is set up](https://github.com/woocommerce/woocommerce/blob/0a25b2f633d57f71d4dfd1f86922b2430e32dad5/plugins/woocommerce/src/Internal/EmailEditor/Integration.php#L24).

Curious to hear your thoughts.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
xristos3490: Oops! Great catch! I intended to set that to `30`, but for some reason, it ended up being `255`.

Fixed in https://github.com/woocommerce/woocommerce/pull/57062/commits/ff0ed615a8a4c3fba9e675120ac1e48f6cf10912

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
xristos3490: Nice! Done in https://github.com/woocommerce/woocommerce/commit/ff0ed615a8a4c3fba9e675120ac1e48f6cf10912

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
xristos3490: Done in https://github.com/woocommerce/woocommerce/pull/57062/commits/5b877fca297f37f120519d887bcd06c7bd47d885

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
xristos3490: Thanks 👌 

Done in https://github.com/woocommerce/woocommerce/pull/57062/commits/5b877fca297f37f120519d887bcd06c7bd47d885

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
xristos3490: I appreciate the attention to detail here! Done in https://github.com/woocommerce/woocommerce/pull/57062/commits/b9d40138a26585adfb0fccbed4a76f231b34c3c1

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
xristos3490: Tiny mistake; 😅 fixed in https://github.com/woocommerce/woocommerce/pull/57062/commits/1ec94e92cc53257f2cbe18e98f6f2bf2c897a1a5

plugins/woocommerce/src/Internal/StockNotifications/Enums/NotificationCancellationSource.php
xristos3490: Done in https://github.com/woocommerce/woocommerce/pull/57062/commits/d4529640569c57a4a116d867912beaa8c83bafd6; also includes the `public` consts

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
xristos3490: Done in https://github.com/woocommerce/woocommerce/pull/57062/commits/49e550d1efc636ece22f6eb58560e636849fe75b

plugins/woocommerce/src/Internal/StockNotifications/Factory.php
xristos3490: Yup! Done in https://github.com/woocommerce/woocommerce/pull/57062/commits/dc06327f8c577c9a4f31a26f7a6d64f582415868

plugins/woocommerce/src/Internal/StockNotifications/Controller.php
kalessil: > Totally understand your skepticism here! 

No skepticism, just trying to better understand details.

> I’m also thinking about renaming it to Integration.php, much like how [the EmailEditor is set up](https://github.com/woocommerce/woocommerce/blob/0a25b2f633d57f71d4dfd1f86922b2430e32dad5/plugins/woocommerce/src/Internal/EmailEditor/Integration.php#L24).

The integration paths are different for BiS and EE. BIS is part of core, as per this PR, so I would discourage adopting the naming.

Generalizing, we are dealing with the `Observer` pattern, so I'd pick a name (containing) `Subscriber` in this case.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -0,0 +1,495 @@
...
+	 * @param Notification $notification The data object to delete.
+	 * @param array        $args         Additional arguments.
+	 * @return void
+	 */
+	public function delete( &$notification, $args = array() ) {
+		global $wpdb;
+
+		$deleted = $wpdb->delete( $this->get_table_name(), array( 'id' => $notification->get_id() ), array( '%d' ) );
+
+		if ( is_int( $deleted ) && $deleted > 0 ) {
```
kalessil: That's neat! Guess we can simplify this to `if ( $deleted > 0 ) {`.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -0,0 +1,495 @@
...
+
+		$should_save =
+			$notification->get_id() > 0
+			&& $notification->get_date_modified( 'edit' ) < $current_date_time
+			&& empty( $notification->get_changes() );
+
+		if ( $should_save ) {
+			$notification->set_date_modified( $current_time );
+			$saved = $notification->save();
+			return is_int( $saved ) && $saved > 0;
```
kalessil: With the latest changes, it seems to be a change in behaviour. WDYT about `return ! is_wp_error( $saved );`?

kalessil (approved):
Thank you for promptly addressing my comments @xristos3490!

Leaving two final comments and pre-approving from my end (CI-passes and the code looks good).

plugins/woocommerce/src/Internal/StockNotifications/Controller.php
xristos3490: > The integration paths are different for BiS and EE. BIS is part of core, as per this PR, so I would discourage adopting the naming.

Gotcha, @kalessil! Makes sense. 

> Generalizing, we are dealing with the Observer pattern, so I'd pick a name (containing) Subscriber in this case.

This class aims to ~~encapsulate~~ initiate all elements of this feature, such as frontend forms, admin UIs, stock-events-subscribers, etc. I'm a bit unsure how the "Subscriber" naming fits here. 

How about a more generic name like `StockNotifications` or `StockNotificationsModule`? I'd appreciate a bit more conceptual feedback to guide me as I add other classes/controllers. Am I thinking this all wrong?

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
xristos3490: Yup, the `is_int` isn't really needed here, since `false` would effectively become zero in this case. Done in https://github.com/woocommerce/woocommerce/pull/57062/commits/4b66f897d2f7b514744f5d72e0ba9ee9b0ae2587

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
xristos3490: Neat! Done in https://github.com/woocommerce/woocommerce/pull/57062/commits/4b66f897d2f7b514744f5d72e0ba9ee9b0ae2587

plugins/woocommerce/src/Internal/StockNotifications/Controller.php
xristos3490: @kalessil I realized my previous question might have been a bit vague. To clarify, would a (very)rough example like the following be in line with Core?

```php
class StockNotificationsController {

    final public function init() {

        $container = wc_get_container();
        $container->get( Emails\EmailManager::class );                 // Constructor adds hooks
        $container->get( EventHandlers\StockChangeHandler::class );    // Constructor adds hooks

        if ( is_request('frontend') ) {
            $container->get( Frontend\ProductPageIntegration::class ); // Constructor adds hooks
            $container->get( Frontend\MyAccountIntegration::class );   // Constructor adds hooks
            $container->get( Frontend\FormHandlerService::class );     // Constructor adds hooks
        }

        if ( is_request('admin') ) {
            $container->get( Admin\AdminMenuController::class );       // Constructor adds hooks
            $container->get( Admin\SettingsPage::class );              // Constructor adds hooks
        }

        add_action( 
            WC_SEND_STOCK_NOTIFICATIONS_JOB_HOOK, 
            function( $args ) {
                $container = wc_get_container();
                $processor = $container->get( AsyncTasks\NotificationsProcessor::class ); // Or via a static bridge (?)
                $processor->process_notifications( $args );
            }
        );
    }
}
```
_* Assume an arbitrary function "is_request()"_

kalessil: > How about a more generic name like StockNotifications

With multiple responsibilities, this is nice option 👍

Issue #56518, opened 2025-03-19
[Store Breadcrumbs]: Custom font size isn't applied on the frontend


gigitux:
### Prerequisites

- [x] I have carried out troubleshooting steps and I believe I have found a bug.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

The Typography → Custom size setting in the Store Breadcrumbs block does not apply correctly on the frontend. While the custom size appears to be set in the editor, it does not reflect properly when the page is rendered. This affects the consistency of typography styling between the editor and the live site.

https://github.com/user-attachments/assets/a571eeb0-9d0e-471c-a584-3a07b2348ec8

### Expected behavior

The custom font size should be correctly applied on the frontend

### Actual behavior

The font size does not reflect the custom setting on the frontend.

### Steps to reproduce

1. Edit the Product Catalog Template.
2. Click on the Store Breadcrumbs block.
3. Edit the Typography → Custom size setting.
4. See that the font size changes on the editor side.
5. Save the template.
6. Visit the `/shop` page.
7. The custom font size isn't applied.

### WordPress Environment

-

### Isolating the problem

- [x] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [x] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/products/storefront/).
- [x] I can reproduce this bug consistently using the steps above.

aniruddhapw:
i am working on it

gigitux:
Thanks @aniruddhapw! Could you please assign the issue to yourself? Moreover, feel free to ping us if you need some guidances.

aniruddhapw:
i cant assign it to myself can you do that @gigitux

gigitux:
@aniruddhapw Are you currently working on this issue? Please let us know so we can unassign it if needed.

aniruddhapw:
@gigitux  you can unassign it, i tried but didn't found the  problem yet

PR #57071, opened 2025-04-07
Breadcrumbs block: fix custom font size applied on the frontend

gigitux:

Breadcrumbs block had default attributes set for “font Size” and “alignment”. This is an antipattern because blocks should be style-agnostic and only themes should style them. For this reason, the custom font size setting didn't work.

⚠️  This PR could bring unexpected layout changes for stores that have themes that override Woo default templates and don't have templates stored in the database: should we publish a dev advisory? 

Closes WOOPLUG-3506 (https://github.com/woocommerce/woocommerce/issues/56518).

1. Edit the Product Catalog Template.
2. Click on the Store Breadcrumbs block.
3. Edit the Typography → Custom size setting.
4. See that the font size changes on the editor side.
5. Save the template.
6. Visit the `/shop` page.
7. Ensure that the custom font size is applied

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

Breadcrumbs block: fix custom font size applied on the frontend

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/57071-wooplug-3506-store-breadcrumbs-custom-font-size-isnt-applied-on-the
+++ b/plugins/woocommerce/changelog/57071-wooplug-3506-store-breadcrumbs-custom-font-size-isnt-applied-on-the
@@ -1,0 +1,4 @@
+Significance: minor
+Type: fix
+
+Breadcrumbs block: fix custom font size applied on the frontend
--- a/plugins/woocommerce/src/Blocks/BlockTypes/Breadcrumbs.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/Breadcrumbs.php
@@ -5,7 +5,7 @@ namespace Automattic\WooCommerce\Blocks\BlockTypes;
 use Automattic\WooCommerce\Blocks\Utils\StyleAttributesUtils;
 
 /**
- * CatalogSorting class.
+ * Breadcrumbs class.
  */
 class Breadcrumbs extends AbstractBlock {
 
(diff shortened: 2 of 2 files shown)


Aljullu (commented):
> ⚠️  This PR could bring unexpected layout changes for stores that have themes that override Woo default templates and don't have templates stored in the database: should we publish a dev advisory? 

Good point, I'm concerned about this as well. This is how the change looks on my dev site:

Before | After
--- | ---
imatge | imatge

I wonder if we could use CSS for that. So we have some default styles that set the font size to small and wide alignment. Something like this:

```CSS
.wc-block-breadcrumbs {
  font-size: var(--wp--preset--font-size--small);
  max-width: 100%;
}
```

If the user or theme defines other styles, they will take priority over the CSS. But if they are not defined, we fall back to how things were rendered until now. What do you think?

gigitux:
I don’t have a strong preference and I’m happy to proceed with this change. That said, wouldn’t it be better to publish a dev advisory and encourage theme developers to update their themes? Otherwise, we’ll be stuck maintaining this CSS indefinitely.

Aljullu:
> I don’t have a strong preference and I’m happy to proceed with this change. That said, wouldn’t it be better to publish a dev advisory and encourage theme developers to update their themes? Otherwise, we’ll be stuck maintaining this CSS indefinitely.

I'm fine with either approach, but IMO the change is too visible to be released only with a developer advisory. Besides themes, any store where the merchant has modified the templates will be affected, which I fear might have an impact on the confidence users have in WooCommerce updates.

I understand that with the alternative approach we will need to maintain the CSS file indefinitely, but unfortunately, I think that's the price we have to pay to guarantee backwards compatibility and reliable updates. Another thing to keep in mind is that at some point in the future the _Breadcrumbs_ block will be updated to a blockified one, so the current one will likely be deprecated.

As a sidenote, I wonder if we could have some kind of linting that prevents us from creating blocks with certain attribute defaults like `fontSize`, `align`, etc. This way we would make sure we don't introduce this kind of issues again in the future. What do you think?

gigitux:
> Besides themes, any store where the merchant has modified the templates will be affected

I didn't think about this case... good point! I'm going to add the CSS that you suggested. 

> As a sidenote, I wonder if we could have some kind of linting that prevents us from creating blocks with certain attribute defaults like fontSize, align, etc. This way we would make sure we don't introduce this kind of issues again in the future. What do you think?

It would be great, do you think that we should open an issue?

Aljullu:
> It would be great, do you think that we should open an issue?

Yes! I just created it: https://github.com/woocommerce/woocommerce/issues/57108 (WOOPLUG-3842). :slightly_smiling_face:

gigitux:
With a2fe99c047455001f1384d08393b55d74f8ca95e, I revisited the approach. As we discussed on Slack, it is pretty complicated to remove the default value from the `block.json`without impacting current stores.

Essentially, both `StyleAttributesUtils` and `get_block_wrapper_attributes` return the `has-small-font-size` class. which applies a `font-size` with `!important`, overriding the `style` attribute. For this reason, the style attribute is ignored.

I re-visited the logic ensuring that when the style attribute is present, the class isn't applied.

Aljullu (approved):
I like this approach, good idea! I tried setting a huge font size in `trunk`:

imatge

And this is how it rendered in `trunk` and in this branch:

`trunk` | This branch
--- | ---
imatge | imatge

<details>

* WordPress 6.8-RC3 local dev environment
* Plugins: WooCommerce, WooCommerce Beta Tester, WordPress Beta Tester
* Theme: Twenty Twenty-Five
* Store creation: April 6th 2025
* Store size: small
</details>

PR #52582, opened 2024-11-05
 CYS: direct all users to the non-AI flow 

Aljullu:
### Changes proposed in this Pull Request:

This PR includes the minimal required changes to direct all users to the non-AI flow. While a lot of other code can be removed, that will be handled as a different task: https://github.com/woocommerce/woocommerce/issues/52549.

Closes #52548.

### How to test the changes in this Pull Request:

0. Generate a ZIP from this branch or use [this one](https://github.com/user-attachments/files/17633249/woocommerce.zip) (when testing the release, simply use the release ZIP)
0. If needed, restart the Customize Your Store flow by installing the WooCommerce Beta Tester plugin and going to `wp-admin/tools.php?page=woocommerce-admin-test-helper` > Tools > "Reset Customize Your Store".
1. In a WordPress.com site, go to WooCommerce > Customize Your Store.
2. Verify there is no option to enter the AI flow:
imatge
3. Click on _Customize your store_. Verify you aren't requested to input a store description (which was the AI flow).
4. Go back to `/wp-admin/admin.php?page=wc-admin&path=%2Fcustomize-store%2Fintro`.
5. Click on _Use the store designer_. Verify you aren't requested to input a store description (which was the AI flow).
6. In a non WordPress.com site, go to WooCommerce > Customize Your Store.
7. Do some smoke testing to verify nothing is broken.

Issue #52549, opened 2024-11-04
CYS: remove AI flow specific code


Aljullu:
After https://github.com/woocommerce/woocommerce/issues/52548 is merged, all users will be directed to the non-AI flow, that means we will be able to remove the AI flow specific code.

This issue is about taking care of that. It includes backend and frontend code. We will need to make sure while removing that code that we don't introduce any backwards compatibility issue.

gigitux:
Hey @Aljullu, I created two PRs to remove the AI flow-specific code.

- JS Side: https://github.com/woocommerce/woocommerce/pull/57564
- PHP Side: https://github.com/woocommerce/woocommerce/pull/57575


I added you as a reviewer because you have more context regarding this project. I would suggest testing both PRs at the same time. This is a build that includes #57564 and #57575: [woocommerce.zip](https://github.com/user-attachments/files/19965017/woocommerce.zip)

PR #57075, opened 2025-04-07
Remove WooAI from the monorepo

emdashcodes:

See pdqkbG-5nb-p2

This PR removes the WooAI plugin from the monorepo, as we are sunsetting it.

1. Verify that the WooAI plugin directory and all its contents have been removed
2. Run `pnpm build `
3. See that no messages related to woo-ai appear

### Changelog entry

<details>

This does not update the main WooCommerce plugin, and instead deprecates one of the other plugins in the monorepo.

</details>


kalessil (approved):
LGTM overall!

Also it would be great to update `.github/project-pr-labeler.yml` as well - it references the plugin:

```yaml
'plugin: woo-ai':
  - plugins/woo-ai/**/*
```

emdashcodes:
Thanks! Updated that as well.

peterfabian (dismissed):
Looks good, thanks for looking into this.

There's also the JS package that complemented the woo-ai plugin in https://github.com/woocommerce/woocommerce/tree/trunk/packages/js/ai that would need to be removed, too (after we double check nobody uses it, but I don't think that's the case).

emdashcodes:
@peterfabian

It looks like the [Design with AI](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/admin/client/customize-store/design-with-ai/services.ts#L6) feature in WooCommerce Admin is currently using it to generate a Jetpack token.

This ties into my question on the P2, since that feature is also using the text-completions endpoint. Should we update the Design feature now as part of this sunset, or leave the package and endpoint in place for the time being?

nerrad:
@peterfabian can you please help get this through to completion with @emdashcodes  - it looks like we were almost there 3 weeks ago and now it's stalled. 

@Aljullu I think the "Design with AI" functionality was only enabled when in the WP Cloud (Atomic) environment so it should be safe to just remove it?

Aljullu:
> @Aljullu I think the "Design with AI" functionality was only enabled when in the WP Cloud (Atomic) environment so it should be safe to just remove it? 

I think so, yes. Since https://github.com/woocommerce/woocommerce/pull/52582 all users are redirected to the non-AI flow, so nothing under `/design-with-ai` is ever run. cc @gigitux just to double-check, as you also worked in CYS, in case I'm missing something or this is used somewhere else.

This might be a good opportunity to clean up the AI flow from CYS, which we never had time to do: https://github.com/woocommerce/woocommerce/issues/52549.

gigitux:
> > @Aljullu I think the "Design with AI" functionality was only enabled when in the WP Cloud (Atomic) environment so it should be safe to just remove it?
> 
> I think so, yes. Since #52582 all users are redirected to the non-AI flow, so nothing under `/design-with-ai` is ever run. cc @gigitux just to double-check, as you also worked in CYS, in case I'm missing something or this is used somewhere else.
> 
> This might be a good opportunity to clean up the AI flow from CYS, which we never had time to do: #52549.

Yeah, you're right, @Aljullu. Even in our internal documentation (PbPjnJ-26s-p2) we mention that AI flow isn't available from WooCommerce 9.5. 

I'm happy to proceed with clean up. I think that we can merge this PR as it is and create other two PRs to:
- remove the CYS AI flow.
- remove the `@woocommerce/ai`.

I'm happy to take care of both.

emdashcodes:
@gigitux I rebased this, so it should be ready to go 👍 

> I'm happy to take care of both.

Thank you! Let me know if you need anything from me otherwise.

gigitux (approved):
LGTM! Thanks for working on this! :shipit:

emdashcodes:
@gigitux Could we actually hold off for a moment on merging this? Per pdtR3Z-4bB-p2, we might end up pushing a deprecation notice for our Woo on Atomic notice, and adding something to the plugin temporarily will make this easier.

So I might prep that update in another PR so we push a notice, and then we can plan on removing the plugin after that.

peterfabian (approved):
> I think that we can merge this PR as it is and create other two PRs to:
> 
> - remove the CYS AI flow.
> - remove the @woocommerce/ai.
> I'm happy to take care of both.

I'm happy with the course of action. @emdashcodes Thanks for taking a step back and going with the user notice first instead of removing it right away.

gigitux:
@emdashcodes Are you planning to continue to work on this PR? I removed the CYS AI flow, so once this PR will be merged I will take care to delete `@woocommerce/ai`.

emdashcodes:
@gigitux Yup, I am actually starting the removal of Woo AI from WP.com today (it's been 2 weeks since our deprecation notice). I'll get this PR rebased and merged now that we no longer need it.

PR #57119, opened 2025-04-09
[dev] Fix plugincheck job for nightly checks

### LINKED ISSUE #57064: Fix plugin-check failure in nightly-checks
Since moving to nightly-checks with woocommerce/woocommerce#56366, the job is [failing](https://github.com/woocommerce/woocommerce/actions/runs/14300051828/job/40088400344#step:9:60):

```
Error: Invalid plugin slug: Plugin with slug woocommerce is not installed.
```

adimoldovan:

Closes https://github.com/woocommerce/woocommerce/issues/57064

The nightly build is using a different plugin name and this causes the pluginchecks job to fail:

```
Error: Invalid plugin slug: Plugin with slug woocommerce is not installed.
```

Install [woocommerce-trunk-nightly.zip](https://github.com/woocommerce/woocommerce/releases/tag/nightly)

Run the plugin check - expected to fail with `Invalid plugin slug`

```
pnpm test:plugincheck
```

Run the plugin check with PLUGIN_SLUG env variable 

```
PLUGIN_SLUG=woocommerce-trunk-nightly pnpm test:plugincheck
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -233,7 +233,7 @@ jobs:
           LAST_FAILED_RUN: ${{ vars.LAST_FAILED_RUN }}
           GITHUB_BASE_SHA: ${{ github.event.pull_request.base.sha }}
           WC_VERSION: ${{ inputs.wc-version }}
-          PLUGIN_NAME: ${{ inputs.refName == 'nightly' && 'woocommerce-trunk-nightly' || '' }} # required by test:plugincheck
+          PLUGIN_SLUG: ${{ inputs.refName == 'nightly' && 'woocommerce-trunk-nightly' || '' }} # required by test:plugincheck
         run: |
           lastRunFile="${{ steps.download-last-run-info.outputs.download-path }}/last-run__${{ strategy.job-index }}/.last-run.json"
           lastRunFileDest="$ARTIFACTS_PATH/.last-run.json"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -233,6 +233,7 @@ jobs:
           LAST_FAILED_RUN: ${{ vars.LAST_FAILED_RUN }}
           GITHUB_BASE_SHA: ${{ github.event.pull_request.base.sha }}
           WC_VERSION: ${{ inputs.wc-version }}
+          PLUGIN_NAME: ${{ inputs.refName == 'nightly' && 'woocommerce-trunk-nightly' || '' }} # required by test:plugincheck
         run: |
           lastRunFile="${{ steps.download-last-run-info.outputs.download-path }}/last-run__${{ strategy.job-index }}/.last-run.json"
           lastRunFileDest="$ARTIFACTS_PATH/.last-run.json"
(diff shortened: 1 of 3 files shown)


Konamiman:
Hi @adimoldovan, sorry I don't understand the testing instructions. You mean running `pnpm test:plugincheck` before and after the fix is applied? And why is the nightly zip file specifically needed?

adimoldovan:
> Hi @adimoldovan, sorry I don't understand the testing instructions. You mean running `pnpm test:plugincheck` before and after the fix is applied? And why is the nightly zip file specifically needed?

@Konamiman the command should be ran after the fix. The second one is supposed to be with `PLUGIN_SLUG=woocommerce-trunk-nightly` set (sorry, I missed adding this in the instructions, updated now).
This check is failing in our nightly checks which are using the nightly zip. With the nightly zip the plugin's slug is `woocommerce-trunk-nightly` instead of `woocommerce`.

Konamiman:
@adimoldovan Is there any built-in way to run `pnpm` directly on a JN site withWooCommerce installed from the linked zip file? I tried to [install pnpm](https://pnpm.io/installation) but the download gets stuck for some reason.

adimoldovan:
> @adimoldovan Is there any built-in way to run `pnpm` directly on a JN site withWooCommerce installed from the linked zip file? I tried to [install pnpm](https://pnpm.io/installation) but the download gets stuck for some reason.

@Konamiman I don't know of any way of doing that. 
If you're trying to test this, why not using the `wp-env`? To install the nightly, you have 2 options:
1. update the `.wp-env.json` by replacing the "." in `env.tests.plugins` with the url to the nightly. 
2. create an override config `.wp-env.override.json` in the same dir. You can use this content for the override:

```
{
  "env": {
    "tests": {
      "plugins": [
        "https://github.com/woocommerce/woocommerce/releases/download/nightly/woocommerce-trunk-nightly.zip",
        "https://downloads.wordpress.org/plugin/akismet.zip",
        "https://github.com/WP-API/Basic-Auth/archive/master.zip",
        "https://downloads.wordpress.org/plugin/wp-mail-logging.zip"
      ]
    }
  }
}

```

Then running `pnpm env:restart` should destroy any existing env, update and then start the new env.

.github/workflows/ci.yml
```diff
@@ -233,6 +233,7 @@ jobs:
           LAST_FAILED_RUN: ${{ vars.LAST_FAILED_RUN }}
           GITHUB_BASE_SHA: ${{ github.event.pull_request.base.sha }}
           WC_VERSION: ${{ inputs.wc-version }}
+          PLUGIN_SLUG: ${{ inputs.refName == 'nightly' && 'woocommerce-trunk-nightly' || '' }} # required by test:plugincheck
```
joshuatf, line 236: Optional, but would it be worth it to explicitly set the fallback to the default `woocommerce` so we don't need to provide defaults anywhere else if this gets reused?

```suggestion
          PLUGIN_SLUG: ${{ inputs.refName == 'nightly' && 'woocommerce-trunk-nightly' || 'woocommerce' }} # required by test:plugincheck
```

joshuatf (approved):
This is testing well for me!  Checked with the original test without setting the `PLUGIN_SLUG` and received the expected error:

Screenshot 2025-04-15 at 4 04 54 PM

With the correct slug:
Screenshot 2025-04-15 at 3 57 15 PM

Interestingly, if I download the nightly build from [this link](https://github.com/woocommerce/woocommerce/releases/download/nightly/woocommerce-trunk-nightly.zip) and then unzip or install, it gets added to the correct path at `woocommerce/`.  But I'm guessing that wp-env unzips to the same name as the zip folder.

Left one optional comment, but otherwise LGTM 🚢

PR #57128, opened 2025-04-09
Always run validation for postcode and email in Checkout block

senadir:

This PR fixes an issue in which Checkout validation for postcode and email would not work. This was introduced in https://github.com/woocommerce/woocommerce/pull/55254 and is only visible if you don't have additional fields installed, likely why our E2E tests didn't catch it.

Closes WOOPLUG-3857

(For Bug Fixes) Bug introduced in PR https://github.com/woocommerce/woocommerce/pull/55254

1. Go to checkout, type an invalid email, notice an inline error.
2. In shipping address, type an invalid postcode `abc` for country United States
3. Notice an inline error.
4. Fix the postcode, notice the error is gone and a request is sent to the server.
5. In Cart, try changing the address.
6. Make sure the zipcode error is shown for invalid zipcodes, and is gone once you input a valid one.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Fix issue in which client side validation in Checkout didn't work.

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/blocks/assets/js/base/hooks/use-schema-parser.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/base/hooks/use-schema-parser.ts
@@ -142,7 +142,7 @@ export const useSchemaParser = < T extends FormType | 'global' >(
 	}
 	return {
 		parser: null,
-		data: null,
+		data,
 	};
 };
 
--- a/plugins/woocommerce/changelog/57128-fix-always-validate-fields-on-client
+++ b/plugins/woocommerce/changelog/57128-fix-always-validate-fields-on-client
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix issue in which client side validation in Checkout didn't work.
--- a/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/form/form.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/form/form.tsx
@@ -91,7 +91,11 @@ const Form = <
 
 	const { errors, previousErrors } = useFormValidation(
 		formFields,
-		addressType
+		addressType,
+		// Temporary override for shipping calculator address form.
+		addressType === 'shipping' || addressType === 'billing'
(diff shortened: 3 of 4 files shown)


tjcafferkey:
Thanks for catching this @senadir 

>is only visible if you don't have additional fields installed, likely why our E2E tests didn't catch it.

Can we follow-up with some test coverage for this as well please?

senadir:
Will look into adding some JS tests for this.

plugins/woocommerce/client/blocks/assets/js/base/hooks/use-schema-parser.ts
```diff
@@ -142,7 +142,7 @@ export const useSchemaParser = < T extends FormType | 'global' >(
 	}
 	return {
 		parser: null,
-		data: null,
+		data,
 	};
```
nielslange, line 146: Do to your change, we could simplify the following part from
```ts
if ( window.schemaParser ) {
	return {
		parser: window.schemaParser,
		data,
	};
}
return {
	parser: null,
	data,
};
```
to
```ts
return {
	parser: window.schemaParser ?? null,
	data,
};
```

nielslange (approved):
Thanks for fixing this, @senadir. The fix works as expected. I left a suggestion on simplifying the return values. Before approving this PR, we should have tests in place as mentioned in https://github.com/woocommerce/woocommerce/pull/57128#issuecomment-2791855456.

nielslange (changes requested):
I've initially approved this PR without noticing that we should add tests. To prevent that this PR accidentally gets merged, I'm changing the status from "Approve" to "Request changes".

nielslange (approved):
As the PR works as expected, and it's needed for a CFE today, I'll approve it again. The tests, as requested in https://github.com/woocommerce/woocommerce/pull/57128#issuecomment-2791855456 will be part of a follow-up PR.

senadir:
This PR initially regressed the shipping calculator form, those were caught with E2E tests, I fixed them now.

PR #55299, opened 2025-02-10
[Store API] Load cart during rest_api_init for Store API requests

mikejolley:
### Submission Review Guidelines:

-   I have followed the [WooCommerce Contributing Guidelines](https://github.com/woocommerce/woocommerce/blob/trunk/.github/CONTRIBUTING.md) and the [WordPress Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/).
-   I have checked to ensure there aren't other open [Pull Requests](https://github.com/woocommerce/woocommerce/pulls) for the same update/change.
-   I have reviewed my code for [security best practices](https://developer.wordpress.org/apis/security/).
-   Following the above guidelines will result in quick merges and clear and detailed feedback when appropriate.

### Changes proposed in this Pull Request:

I have extracted this change from https://github.com/woocommerce/woocommerce/pull/55047/files#diff-06f7ff3b0e7a657b9691cff165a10019e8b3b648b2fc115f56c46d43c14ad9bd to reduce the scope of the other PR.

We have some upcoming use-cases where we need to access the cart early when the Store API request is being generated. To that end, this PR moves the Cart Token handling and cart init to `rest_api_init`.

### How to test the changes in this Pull Request:

The cart token is not something you can test without a REST API client. This test is aimed at developers.

1. Using REST API client,  post to `store.local/wp-json/wc/store/v1/cart` and add an item to cart, e.g. 

```json
{
	"id": 16
}
```

2. GET `store.local/wp-json/wc/store/v1/cart`. Product is there.
3. Look at headers and copy the cart token.
4. Clear cookies.
5. GET `store.local/wp-json/wc/store/v1/cart`. Cart is empty.
6. Add `cart-token` header with the token you copied earlier.
7. GET `store.local/wp-json/wc/store/v1/cart`. Cart has the product again.

### Changelog entry

-   [ ] Automatically create a changelog entry from the details below.

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Details</summary>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

-   [ ] Fix - Fixes an existing bug
-   [ ] Add - Adds functionality
-   [ ] Update - Update existing functionality
-   [ ] Dev - Development related task
-   [ ] Tweak - A minor adjustment to the codebase
-   [ ] Performance - Address performance issues
-   [ ] Enhancement - Improvement to existing functionality

#### Message 

</details>

<details>

[...]

PR #57152, opened 2025-04-10
[Store API] Revert #55299 to handle sessions and cart tokens when the request is available

mikejolley:

This PR reverts the changes in https://github.com/woocommerce/woocommerce/pull/55299/files so that session is loaded when a `WP_Rest_Request` object is available, rather than early during `rest_api_init`.

Because of this change, `validate_callback` under the checkout route needed some modifications so that it runs later still, has access to cart, and handles sanitized data since it no longer has access to the unsanitized values. 

This is to fix an inconsistent bug in WooPay where sometimes the session would not be loaded and cause an email validation error blocking checkout. Its still not 100% certain why this occurred, but this PR is being prepped ready in case we want to fix it this way.

(For Bug Fixes) Bug introduced in PR #55299

WooPay team confirmed this seems to mitigate their issue, but on the core side:

1. [Install checkout fields tester](https://github.com/woocommerce/additional-checkout-fields-tester)
2. Confirm you can checkout with additional fields present and filled
3. Via Store API, add an item to the cart, then POST to checkout. Check that you get error messages about missing fields in the request for additional fields and gradually populate them to check that errors are accurate. You should end up with a request something like this to cover additional fields:

```
{
	"billing_address": {
		"first_name": "Steve",
		"last_name": "Stevenson",
		"email": "test@test.com",
		"address_1": "41 Some Street",
		"city": "Townford",
		"postcode": "CB25 6FG",
		"country": "GB",
		"plugin-namespace/gov-id": "12345",
		"plugin-namespace/confirm-gov-id": "12345"
	},
	"shipping_address": {
		"first_name": "Steve",
		"last_name": "Stevenson",
		"email": "test@test.com",
		"address_1": "41 Some Street",
		"city": "Townford",
		"postcode": "CB25 6FG",
		"country": "GB",
		"plugin-namespace/gov-id": "12345",
		"plugin-namespace/confirm-gov-id": "12345"
	},
	"payment_method": "bacs",
	"additional_fields": {
		"plugin-namespace/alt-email": "test@test.com",
		"plugin-namespace/job-function": "director"
	}
}
```

The above needs to be checked against a dev build and a production build, since this has some experimental features.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/src/StoreApi/Authentication.php
+++ b/plugins/woocommerce/src/StoreApi/Authentication.php
@@ -21,20 +21,9 @@ class Authentication {
 		add_filter( 'rest_authentication_errors', array( $this, 'check_authentication' ) );
 		add_filter( 'rest_authentication_errors', array( $this, 'opt_in_checkout_endpoint' ), 9, 1 );
 		add_action( 'set_logged_in_cookie', array( $this, 'set_logged_in_cookie' ) );
-		add_filter( 'rest_pre_serve_request', array( $this, 'send_cors_headers' ), 10, 1 );
+		add_filter( 'rest_pre_serve_request', array( $this, 'send_cors_headers' ), 10, 3 );
 		add_filter( 'rest_allowed_cors_headers', array( $this, 'allowed_cors_headers' ) );
 		add_filter( 'rest_exposed_cors_headers', array( $this, 'exposed_cors_headers' ) );
-
-		// If cart has a valid token, override the core session class.
-		// Validate returns early if no token, so no performance penalty.
-		if ( JsonWebToken::validate( $this->get_cart_token(), $this->get_cart_token_secret() ) ) {
-			add_filter(
-				'woocommerce_session_handler',
-				function () {
-					return SessionHandler::class;
-				}
-			);
-		}
 
 		// Remove the default CORS headers--we will add our own.
 		remove_filter( 'rest_pre_serve_request', 'rest_send_cors_headers' );
@@ -75,10 +64,12 @@ class Authentication {
 	 *
 	 * Users of valid Cart Tokens are also allowed access from any origin.
 	 *
-	 * @param bool $value  Whether the request has already been served.
+	 * @param bool             $value  Whether the request has already been served.
+	 * @param \WP_REST_Server  $server The REST server instance.
+	 * @param \WP_REST_Request $request The REST request instance.
 	 * @return bool
 	 */
-	public function send_cors_headers( $value ) {
+	public function send_cors_headers( $value, $server, $request ) {
 		$origin = get_http_origin();
 
 		if ( 'null' !== $origin ) {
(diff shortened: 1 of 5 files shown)


plugins/woocommerce/src/StoreApi/Routes/V1/Checkout.php
```diff
@@ -86,7 +86,6 @@ public function get_args() {
 				'methods'             => \WP_REST_Server::CREATABLE,
 				'callback'            => [ $this, 'get_response' ],
 				'permission_callback' => '__return_true',
-				'validate_callback'   => [ $this, 'validate_callback' ],
```
mikejolley, line 89: These validate callbacks run early, before params are sanitised. This was one of the reasons we wanted the cart to be loaded earlier for all store API requests.

plugins/woocommerce/src/StoreApi/Routes/V1/Checkout.php
```diff
@@ -219,14 +219,6 @@ protected function get_route_response( \WP_REST_Request $request ) {
 	 * @return true|\WP_Error
 	 */
 	public function validate_callback( $request ) {
-		/**
-		 * The request is cloned to avoid modifying the original request object when sanitizing params.
-		 * Un-sanitized params are used to see if required fields had values. Sanitized params are used to
-		 * validate field values.
-		 */
-		$sanitized_request = clone $request;
-		$sanitized_request->sanitize_params();
```
mikejolley, line 228: Moving the validation to the request handler means its loading after sanitization of params. Therefore this is not needed.

senadir (approved):
Thank you Mike, I tested this against conditional additional fields and it's working correctly.

plugins/woocommerce/src/StoreApi/Authentication.php
```diff
@@ -75,10 +64,12 @@ public function exposed_cors_headers( $exposed_headers ) {
...
 	 * Users of valid Cart Tokens are also allowed access from any origin.
 	 *
-	 * @param bool $value  Whether the request has already been served.
+	 * @param bool             $value  Whether the request has already been served.
+	 * @param \WP_REST_Server  $server The REST server instance.
+	 * @param \WP_REST_Request $request The REST request instance.
 	 * @return bool
 	 */
-	public function send_cors_headers( $value ) {
+	public function send_cors_headers( $value, $server, $request ) {
```
tjcafferkey, line 72: @mikejolley just double checking these params against the `apply_filters` implementation [here](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/rest-api/class-wp-rest-server.php#L515).

Looks like the second param is passed as a `WP_HTTP_Response` object so using `$server` appears misleading if I'm understanding this correctly, even if we overwrite it later anyway this still could potentially cause confusion.

Also, is there an opportunity to skip `$server = rest_get_server();` since `$server` is passed as the 4th param to the filter?

mikejolley: Yeah you are correct. Fixed in https://github.com/woocommerce/woocommerce/pull/57963

PR #57172, opened 2025-04-11
Prevent infinite redirect loop when cart shortcode is present on checkout page

### LINKED ISSUE #57082: Prevent endless redirect on checkout page when cart is empty
### Prerequisites

### Describe the bug

It seems like there could be cases where wc_get_cart_url() ends up returning the checkout page URL causing endless redirect on checkout page if the cart is empty. 

In 9509714-zen, we saw the checkout page had two blank lines before the checkout shortcode causing this:

```
// blank line
// blank line
<!-- wp:woocommerce/classic-shortcode {"shortcode":"checkout"} /-->
``` 

This seems to be taken care of in the latest version of `woocommerce/src/Blocks/Utils/CartCheckoutUtils.php` file however if you simply put a classic cart shortcode on the checkout page - it causes infinite redirect.

### Expected behavior

Prevent infinite redirect regardless the content of checkout page. 

### Actual behavior

NA

### Steps to reproduce

Add [woocommerce_cart] to the checkout page in addition to checkout shortcode and see it being redirected to itself indefinitely when cart is empty. 

### WordPress Environment

NA

### Isolating the problem

mikejolley:

This PR fixes an infinite redirect loop that occurs when the cart shortcode is present on the checkout page and the cart is empty. Currently, if a merchant adds both the cart and checkout shortcodes on the same page, it can cause an endless redirect when customers try to access the checkout with an empty cart.

The fix adds a check to prevent the redirect if we're already on what WooCommerce considers to be the cart page, preventing the infinite loop while maintaining the expected behavior of redirecting empty carts from checkout to cart in normal scenarios.

Closes #57082.

1. Create a new page or edit your checkout page
2. Add both the cart and checkout shortcodes to the page:
   ```
   [woocommerce_cart]
   [woocommerce_checkout]
   ```
3. Empty your cart if it has items
4. Visit the checkout page
5. **Expected:** The page should redirect once to the cart page and stay there
6. **Previous behavior:** The page would enter an infinite redirect loop between cart and checkout

Additional test cases:
1. Test normal checkout with items in cart
   - Add an item to cart
   - Visit checkout
   - **Expected:** Checkout should work normally with no redirects

2. Test normal empty cart behavior
   - Empty your cart
   - Visit a regular checkout page (without cart shortcode)
   - **Expected:** Should redirect once to cart page

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/wooplug-3832-prevent-endless-redirect-on-checkout-page-when-cart-is-empty
+++ b/plugins/woocommerce/changelog/wooplug-3832-prevent-endless-redirect-on-checkout-page-when-cart-is-empty
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Prevent infinite redirect loop when cart shortcode is present on checkout page with empty cart.
--- a/plugins/woocommerce/includes/wc-template-functions.php
+++ b/plugins/woocommerce/includes/wc-template-functions.php
@@ -32,8 +32,11 @@ function wc_template_redirect() {
 
 	// When on the checkout with an empty cart, redirect to cart page.
 	if ( is_page( wc_get_page_id( 'checkout' ) ) && wc_get_page_id( 'checkout' ) !== wc_get_page_id( 'cart' ) && WC()->cart->is_empty() && empty( $wp->query_vars['order-pay'] ) && ! isset( $wp->query_vars['order-received'] ) && ! is_customize_preview() && apply_filters( 'woocommerce_checkout_redirect_empty_cart', true ) ) {
-		wp_safe_redirect( wc_get_cart_url() );
-		exit;
+		// Only redirect if we're not already on the cart page to prevent infinite loops.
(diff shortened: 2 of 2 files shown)


tjcafferkey:
@mikejolley do you think we can add a test case for this if we don't already have coverage?

mikejolley:
@tjcafferkey this area of core does not have coverage at all. Cursor had a go but failed during the workshop :) There is no rush for this one, if we need to spend time adding coverage we can work on it when I get back.

tjcafferkey:
@mikejolley okay lets just create a follow-up so we can track it, thanks!

senadir (approved):
Such a no brainer change Mike! working well.

mikejolley:
Logged https://github.com/woocommerce/woocommerce/issues/57453

PR #57179, opened 2025-04-11
Fix undefined array key "title" warning in shipping zone methods

tjcafferkey:

This PR addresses a PHP warning by properly checking if the 'title' key exists in the instance settings array before accessing it, preventing "undefined array key" errors in the REST API shipping zone methods controller.

Closes [WOOPLUG-3730](https://linear.app/a8c/issue/WOOPLUG-3730/warning-undefined-array-key-title-in-shipping-methods)

| Before | After |
| ------ | ----- |
|        |       |

1. Add the below snippet to your store. It will register a shipping method without a title, and inject to JS on the frontend to fetch the data from the endpoint. Ensure that you update the shipping zone ID (`2` in the snippet below) with the correct value from your store.
2. Go into your shipping settings and add this as a method to your zone
3. Optional: Undo my change and load the homepage, check your network tab to ensure request was sent. Then check logs for the error before re-implementing the change.
4. Load homepage,  check your network tab to ensure request was sent. Then check logs to ensure the error is no longer present. Also check the response in the network request to check that the `title` is "Custom Shipping Method Title"

```php
function fetch_shipping_methods() {
    ?>
    <script>
    document.addEventListener('DOMContentLoaded', function() {
        wp.apiFetch({
            path: '/wc/v3/shipping/zones/2/methods', // Use path instead of full URL
            method: 'GET',
            headers: {
                'Content-Type': 'application/json',
            }
        })
        .then(data => {
            console.log('Shipping methods:', data);
        })
        .catch(error => {
            console.error('Error fetching shipping methods:', error);
        });
    });
    </script>
    <?php
}
add_action('wp_footer', 'fetch_shipping_methods');

add_action( 'woocommerce_shipping_init', 'my_custom_shipping_method_init' );
function my_custom_shipping_method_init() {
    if ( ! class_exists( 'WC_My_Custom_Shipping_Method' ) ) {
        class WC_My_Custom_Shipping_Method extends WC_Shipping_Method {
            public function __construct( $instance_id = 0 ) {
                $this->id                 = 'my_custom_shipping';
                $this->instance_id        = absint( $instance_id );
                $this->method_title       = __( 'Custom Shipping Method Title', 'woocommerce' );
                $this->method_description = __( 'Custom shipping method', 'woocommerce' );
                $this->supports           = array( 'shipping-zones', 'instance-settings' );
                $this->init();
            }

            public function init() {
                $this->init_form_fields();
                $this->init_settings();
                $this->enabled = $this->get_option( 'enabled', 'yes' );
                $this->title   = $this->get_option( 'title', $this->method_title );
                $this->cost    = $this->get_option( 'cost', 10 );
                add_action( 'woocommerce_update_options_shipping_' . $this->id, array( $this, 'process_admin_options' ) );
            }

            public function init_form_fields() {
                $this->instance_form_fields = array(
                    'enabled' => array(
                        'title'   => __( 'Enable', 'woocommerce' ),
                        'type'    => 'checkbox',
                        'label'   => __( 'Enable this method', 'woocommerce' ),
                        'default' => 'yes',
                    ),
                    // 'title' => array(
                    //     'title'   => __( 'Title', 'woocommerce' ),
                    //     'type'    => 'text',
                    //     'default' => __( 'Custom Shipping', 'woocommerce' ),
                    // ),
                    'cost' => array(
                        'title'   => __( 'Cost', 'woocommerce' ),
                        'type'    => 'number',
                        'default' => '10',
                    ),
                );
            }

            public function calculate_shipping( $package = array() ) {
                $this->add_rate( array(
                    'id'    => $this->get_rate_id(),
                    'label' => $this->title,
                    'cost'  => $this->cost,
                ) );
            }
        }
    }
}

// Register the shipping method
add_filter( 'woocommerce_shipping_methods', 'add_my_custom_shipping_method' );
function add_my_custom_shipping_method( $methods ) {
    $methods['my_custom_shipping'] = 'WC_My_Custom_Shipping_Method';
    return $methods;
}
```

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-shipping-zone-methods-v2-controller.php
```diff
@@ -352,7 +352,7 @@ public function prepare_item_for_response( $item, $request ) {
 		$method = array(
 			'id'                 => $item->instance_id,
 			'instance_id'        => $item->instance_id,
-			'title'              => $item->instance_settings['title'],
+			'title'              => isset( $item->instance_settings['title'] ) ? $item->instance_settings['title'] : $item->method_title,
```
opr: You can null coalesce this
```suggestion
			'title'              => $item->instance_settings['title'] ?? $item->method_title,
```

opr (changes requested):
Thanks @tjcafferkey. ~The testing instructions are OK, but I'd recommend moving the JS part out and having it as a separate snippet. The tester should then enter it onto their console.~ 

Actually, this is fine. My concern was using plugins like code snippets wouldn't parse the closing PHP tags properly, but it's fine.

I'd also mention that the endpoint needs to have the ID changed to match the zone ID you added the method to.

E.g. when adding the method, check the URL and use that ID in the snippet.

`/wp-admin/admin.php?page=wc-settings&tab=shipping&zone_id=4`.

Left one comment about null coalesce. I think this is fine because if `$item->instance_settings['title']` doesn't exist (either because `instance_settings` doesn't have a 'title' key or if `instance_settings` itself is null), the expression will evaluate to the fallback value: `$item->method_title`.

The null coalesce operator handles both:

- When the left operand is null
- When the left operand is an undefined array key

Moreover, if `$item->instance_settings['title']` exists but contains an empty string, the null coalesce operator will _not_ fall back to `$item->method_title` which I think is desired, if the title has been purposely set to an empty string we should respect that.

Let me know if you agree, and I'll approve if you make the changes/provide a solid reason why we shouldn't change it.

plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-shipping-zone-methods-v2-controller.php
tjcafferkey: https://github.com/woocommerce/woocommerce/pull/57179/commits/1b9239762fee4d24dde5cfb107cfe9c86fd0b9d9

tjcafferkey:
>I'd also mention that the endpoint needs to have the ID changed to match the zone ID you added the method to.

Agreed - updated

>Left one comment about null coalesce.

Agreed - updated

>Let me know if you agree, and I'll approve if you make the changes/provide a solid reason why we shouldn't change it.

Also agree with this statement

opr (approved):
Thanks, re-tested and it's working well.

PR #57181, opened 2025-04-11
Blocks: Add Payment Method Icons Block

mikachan:

This PR adds a new block, Payment Method Icons, which lists all the default card icons available with WooPayments, plus any other active payment method that has an icon. This block is to be used with the new Woo Starter Theme. Here is the block in the theme design:

Screenshot 2025-06-19 at 13 35 39

Closes WOOPLUG-4698.

image

image

1. Go to the Editor (Appearance - Editor).
2. Ensure you have some payment methods enabled via WooPayments.
3. Find and insert the Payment Method Icons block.
4. Ensure that you can adjust the number of icons and they show correctly in the Editor and frontend.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


senadir:
Hey @mikachan if this is going to be used all over the site, I suggest you render this PHP-side using the PHP registry.

- It doesn't load heavy JS packages in the home page.
- It doesn't cause payments scripts to be loaded in home page.
- It works with payment methods that aren't integrated.

senadir:
I took a look at the code and it's indeed PHP rendered but still relies on the global registry in the editor to statically save the payment methods, which I still think is not ideal here because what happens when a new gateway is enabled or disabled, the merchant will need to save again. Just straight up reading from the PHP payment registry is good, possibly leaving the option to override certain gateways in the editor, but keeping the source of truth being the gateway registry.
Also making sure you're using the gateway title, so Credit Card instead of WooPayments, which is a merchant name, not a customer one.
One last thing which is that doing this PHP side will not support express payment buttons like Apple Pay and Google Pay, that need to be figured out maybe.

plugins/woocommerce/client/blocks/assets/js/blocks/payment-methods/style.scss
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	.wc-block-payment-methods__list-item {
+		display: flex;
+		align-items: center;
+		justify-content: center;
+		min-height: 24px;
+		padding: 12px;
+		border: 1px solid currentColor;
+		border-radius: 4px;
```
MaggieCabrera: Should we just use the button element class instead to let the block inherit from the theme? maybe use the outline class if needed too

MaggieCabrera:
I like this a lot. Maybe I'd suggest adding a limit of icons to show in case some sites have like a million payment types? but that's not a blocker for me. Also, I don't think we need to get this 100% right from the get go, we can iterate and nudge extenders to use a consistent way to provide an icon that can show in this block.

plugins/woocommerce/client/blocks/assets/js/blocks/payment-methods/style.scss
mikachan: I ended up removing this border as the icons included in WooPayments already have a border, and I think these are the ones used in the theme design.

mikachan:
Thanks @MaggieCabrera!

> Maybe I'd suggest adding a limit of icons to show in case some sites have like a million payment types? 

Yes, good idea. The block currently only supports 5 icons, but I've added a "number of icons" control anyway so users can display however many icons they want from the default list.

> Also, I don't think we need to get this 100% right from the get go, we can iterate and nudge extenders to use a consistent way to provide an icon that can show in this block.

I agree. It doesn't seem clear yet how these icons should be surfaced, but there is already an `icons` array from WooPayments, which seems ideal, but it's currently an empty array. For now, I've hardcoded the default cards that are required as part of WooPayments. I'm hoping that once we have the basic block, we can continue to improve it and make it more flexible as more icons become available.

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/block.json
```diff
@@ -0,0 +1,27 @@
...
+	"title": "Payment Method Icons",
+	"description": "Display icons for available payment methods.",
+	"category": "woocommerce",
+	"keywords": [ "woocommerce", "payments", "payment methods" ],
+	"textdomain": "woocommerce",
+	"attributes": {
+		"className": {
+			"type": "string",
+			"default": ""
+		},
```
kmanijak: This seems redundant. I don't think it's used anywhere and also, custom classnames are handled natively in advanced section of block.

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/block.json
```diff
@@ -0,0 +1,27 @@
+{
+	"name": "woocommerce/payment-method-icons",
+	"version": "1.0.0",
+	"title": "Payment Method Icons",
+	"description": "Display icons for available payment methods.",
```
kmanijak, line 5: If it only lists icons available via WooPayments, we should probably clarify it:

```suggestion
	"description": "Display icons for available WooPayments methods.",
```

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/edit.tsx
```diff
@@ -0,0 +1,97 @@
...
+							onChange={ ( value ) =>
+								setAttributes( { numberOfIcons: value } )
+							}
+							min={ 1 }
+							max={ availableTypes.length }
+							help={ __(
+								'Choose how many icons to display.',
+								'woocommerce'
+							) }
+						/>
```
kmanijak: I know it extends the PR scope but what about adding `0` as an option which would mean: show all. This pattern is used in many places, e.g. in Query Loop's Max Pages To show where `0` means show them all.

In current state user always has to adjust the value when adding additional method.

Also, in case you add it, I'd suggest making `0` a default value.

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/edit.tsx
```diff
@@ -0,0 +1,97 @@
...
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { useBlockProps, InspectorControls } from '@wordpress/block-editor';
+import { PanelBody, RangeControl } from '@wordpress/components';
+
+/**
+ * Internal dependencies
+ */
+import './style.scss';
```
kmanijak: `style.scss` is also imported in `index.tsx`. Please remove one import.

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+import edit from './edit';
+import './style.scss';
+
+registerBlockType( metadata.name, {
+	...( metadata as BlockConfiguration ),
+	icon: {
+		src: (
+			<Icon
+				icon={ paymentMethodsIcon }
+				className="wc-block-editor-components-block-icon"
```
kmanijak: I don't think this class is used so what about removing it?

```suggestion
```

plugins/woocommerce/src/Blocks/BlockTypes/PaymentMethodIcons.php
```diff
@@ -0,0 +1,168 @@
...
+
+	/**
+	 * Extra data passed through from server to client for block.
+	 *
+	 * @param array $attributes  Any attributes that currently are available from the block.
+	 */
+	protected function enqueue_data( array $attributes = [] ) {
+		parent::enqueue_data( $attributes );
+		$this->asset_data_registry->add( 'paymentMethodIcons', $this->get_enabled_card_types() );
+		$this->asset_data_registry->add( 'wooPaymentsEnabled', $this->is_woo_payments_enabled() );
```
kmanijak: Is this data used anywhere? I see `get_enabled_card_types` and `is_woo_payments_enabled` methods are called elsewhere anyway so we could probably avoid setting the data in registry. Unless there's a purpose for it I'm missing.

plugins/woocommerce/src/Blocks/BlockTypes/PaymentMethodIcons.php
```diff
@@ -0,0 +1,168 @@
...
+			}
+
+			$output .= '<div class="wp-block-woocommerce-payment-method-icons__item">';
+			$output .= '<span class="wp-block-woocommerce-payment-method-icons__icon" style="background-image: url(\'' . esc_url( $card_data['icon'] ) . '\');" role="img" aria-label="' . esc_attr( $card_data['name'] ) . '"></span>';
+
+			$output .= '</div>';
+		}
+
+		return $output;
+	}
```
kmanijak: It may be me but I find this function a bit hard to understand. I'd suggest the following but let me know if that feels better for you:

```suggestion
	private function render_card_types( $attributes ) {
		$output             = '';
		$enabled_card_types = $this->get_enabled_card_types();
		$number_of_icons    =  min( intval( $attributes['numberOfIcons'] ?? 5 ), count( $enabled_card_types ) );
		
		for ( $i = 0; $i <= $number_of_icons; $i++ ) {
		    $output .= '<div class="wp-block-woocommerce-payment-method-icons__item">';
			$output .= '<span class="wp-block-woocommerce-payment-method-icons__icon" style="background-image: url(\'' . esc_url( $card_data['icon'] ) . '\');" role="img" aria-label="' . esc_attr( $card_data['name'] ) . '"></span>';
             $output .= '</div>';
		}

		return $output;
	}
```

Basically it switches from `foreach` to `for` loop which removes the need for `max( 1, ...` and also we don't need to track the iterator manually which `if else break`. Let me know what you think 🙌

plugins/woocommerce/src/Blocks/BlockTypes/PaymentMethodIcons.php
```diff
@@ -0,0 +1,168 @@
...
+				'name' => 'Discover',
+				'icon' => $this->get_card_type_icon_url( 'discover' ),
+			),
+			'jcb'        => array(
+				'name' => 'JCB',
+				'icon' => $this->get_card_type_icon_url( 'jcb' ),
+			),
+		);
+
+		return $card_types;
```
kmanijak, line 157: Ia this supposed to be hardcoded list?

kmanijak: Ah, I see this in description:

> Currently, it looks like getPaymentMethods returns null for the icons attribute, so for now I've added all the default required cards from WooPayments.

so I assume it it expected.

kmanijak (changes requested):
For now I only reviewed the code as I couldn't make WooPayments to work in my local env. I'll test the feature tomorrow 🙌

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/edit.tsx
```diff
@@ -0,0 +1,97 @@
...
+				<div className="wp-block-woocommerce-payment-method-icons">
+					{ availableTypes.slice( 0, iconsToShow ).map( ( type ) => (
+						<CardPreview key={ type } type={ type } />
+					) ) }
+				</div>
+			</div>
+		);
+	}
+	return (
+		<div>
```
kmanijak: We should pass block props to this div as well so it gets outline and behaviour of the block in editor.

```suggestion
		<div { ...blockProps }>
```

Currently it's not interactive:

https://github.com/user-attachments/assets/b346577a-200d-40e6-adde-9d9604cdd822

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/edit.tsx
```diff
@@ -0,0 +1,97 @@
...
+	};
+	setAttributes: ( attributes: Record< string, unknown > ) => void;
+} ) => {
+	const blockProps = useBlockProps();
+	const { numberOfIcons } = attributes;
+	const isWooPaymentsEnabled = window.wcSettings.wooPaymentsEnabled;
+
+	if ( isWooPaymentsEnabled ) {
+		const { paymentMethodIcons } = window.wcSettings as {
+			paymentMethodIcons: Record< string, { icon: string } >;
```
kmanijak: After checking the `wcSettings` object and my WooPayments config, I don't think `paymentMethodIcons` is a correct field to use. IMO we should be using `wcSettings.paymentMethodData.woocommerce_payments.paymentMethodsConfig` which includes all the enabled methods and each of them have `icon` field.

I checked and object seems to be reflecting WooPayments config:

Case | Only 3 methods enable | Many methods enabled
-- | -- | --
WooPayments | image |  image 
`paymentMethodsConfig` | image | image

`paymentMethodIcons` seems to just have 5 hardcoded icons.

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/block.json
```diff
@@ -0,0 +1,27 @@
...
+			"type": "number",
+			"default": 5
+		}
+	},
+	"supports": {
+		"spacing": {
+			"margin": true,
+			"padding": true
+		}
+	},
```
kmanijak, line 20: Padding and margin are only respected in the editor:

Editor | Frontend
-- | --
image | image

Most likely this is due the missing mapping in `plugins/woocommerce/src/Blocks/BlockTypes/PaymentMethodIcons.php` from attributes to styles. I guess it's not ideal but we're using  internal util `StyleAttributesUtils::get_classes_and_styles_by_attributes` to properly map `supports` attributes into classes and styles.

kmanijak (changes requested):
Hey! I made another round of review, this time with proper testing and left couple more comments. The main one would be to switch the data source in Editor to one that actually reflects the currently enabled payments methods (`wcSettings.paymentMethodData.woocommerce_payments.paymentMethodsConfig`). This data is also available on the frontend so we should look if there's a way of reading `wcSettings` somehow on the server side (I don't know the solution to this though), so we get the correct payment methods to be rendered on the frontend as well.

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/edit.tsx
mikachan: The problem with using `paymentMethodsConfig` is that the icon for `card` is the generic card icon:

image

And for this block we want to list all the card brands that are available, which I'm not sure is currently possible. But I would love to know if it is possible!

We could use this to display the other payment methods though, in addition to cards.

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/block.json
mikachan: Yes! Removed in https://github.com/woocommerce/woocommerce/pull/57181/commits/b18273c36c4d5dc6a16c9b1d7b3f1f8cad8b9df9.

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/block.json
mikachan: That makes sense. I'd like for this block to handle multiple payment gateways eventually, but at the moment for simplicity, it only handles WooPayments. Updated the description in https://github.com/woocommerce/woocommerce/pull/57181/commits/ab437e3dfa1eb513b086453fc43c22067872c467.

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/edit.tsx
mikachan: Nice idea! Done in https://github.com/woocommerce/woocommerce/pull/57181/commits/69b023e23a84c06a7a022f36c043500458931f87.

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/edit.tsx
mikachan: Done in https://github.com/woocommerce/woocommerce/pull/57181/commits/46e21c87fd72b37cd626e28b994837b44f10c8df.

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/index.tsx
mikachan: Done in https://github.com/woocommerce/woocommerce/pull/57181/commits/aceeb34fe9fdec47eb175056299ae31b593e3ca3.

plugins/woocommerce/src/Blocks/BlockTypes/PaymentMethodIcons.php
mikachan: I added these so we could access this data in the Editor (e.g. through `window.wcSettings.wooPaymentsEnabled`). However, I'm not sure how things like this are usually handled and maybe there's a better way.

plugins/woocommerce/src/Blocks/BlockTypes/PaymentMethodIcons.php
mikachan: That makes sense, thanks! Done in https://github.com/woocommerce/woocommerce/pull/57181/commits/2993debd51e3d2045e5bb955bf2fcae8ac2a0d35.

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/edit.tsx
mikachan: Done in https://github.com/woocommerce/woocommerce/pull/57181/commits/573fd9a0cdf67c5959b8385092102bdf0ad459b3.

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/edit.tsx
mikachan: I've made it so any other payment method icons are included, as well as the card icons: https://github.com/woocommerce/woocommerce/pull/57181/commits/9ba8bb665cd89d1f1e3bd8529397002a2a8bd4d2.

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/block.json
mikachan: Good spot, thanks! Fixed in https://github.com/woocommerce/woocommerce/pull/57181/commits/b5b09d220c7e86bb925abb1e5d6c9557e0db2d1f.

plugins/woocommerce/src/Blocks/BlockTypes/PaymentMethodIcons.php
```diff
@@ -0,0 +1,222 @@
...
+	private function is_woopayments_enabled() {
+		if ( ! class_exists( 'WooCommerce' ) ) {
+			return false;
+		}
+
+		$payment_gateways = WC()->payment_gateways->get_available_payment_gateways();
+
+		return isset( $payment_gateways['woocommerce_payments'] ) &&
+				$payment_gateways['woocommerce_payments']->enabled === 'yes';
+	}
```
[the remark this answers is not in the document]
mikachan: Done in https://github.com/woocommerce/woocommerce/pull/57181/commits/cb01954396d106596be28d0c4c0c02e56b1bc746.

plugins/woocommerce/client/blocks/bin/webpack-entries.js
```diff
@@ -79,6 +79,9 @@ const blocks = {
 	},
 	'classic-shortcode': {},
 	'page-content-wrapper': {},
+	'payment-method-icons': {
+		isExperimental: true,
```
mikachan: Should this block be experimental? I marked it as such because it's a new block, but I'm not sure how this is usually decided.

plugins/woocommerce/src/Blocks/BlockTypes/PaymentMethodIcons.php
mikachan: I ended up removing `wooPaymentsEnabled` and using `getPaymentMethods()` to check if WooPayments is enabled. I'm still using `paymentMethodIcons` to provide the card icons so they're consistently available on both the server and client side.

mikachan:
Thank you for the reviews, @kmanijak! 🙇 I think I've addressed all your feedback. One of the main changes from the recent updates is that the other non-card payment methods from WooPayments should now also be included in the list of icons.

plugins/woocommerce/client/blocks/bin/webpack-entries.js
kmanijak: It's decided case by case by ourselves. Simple blocks often skip the experimental phase. We usually use it for complex blocks that require multiple PRs to achieve stable state.

In case of this block the criterium would be:
- if it can be used by users after this PR and it would display payment method deterministically - no need to mark it experimental
- if you consider this PR as foundation for block and more PRs is needed to make block production ready then it should be experimental.

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/edit.tsx
```diff
@@ -0,0 +1,134 @@
...
+};
+
+const Edit = ( {
+	attributes,
+	setAttributes,
+}: {
+	attributes: {
+		numberOfIcons: number;
+	};
+	setAttributes: ( attributes: Record< string, unknown > ) => void;
```
kmanijak: Minor but you can make it more strict by extracting attributes type and reusing it in `setAttributes`

```suggestion
type BlockAttributes = {
  numberOfIcons: number;
}

...

	attributes: BlockAttributes,
	setAttributes: ( attributes: Partial<BlockAttributes> ) => void;
```

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/edit.tsx
```diff
@@ -0,0 +1,134 @@
...
+			.map( ( type ) => {
+				if ( ! cardIcons[ type ] || ! cardIcons[ type ].icon ) {
+					return null;
+				}
+				return {
+					type,
+					icon: cardIcons[ type ].icon,
+				};
+			} )
+			.filter( Boolean );
```
kmanijak: What about converting `map + filter` (two loops) into `reduce` (single loop)?

```suggestion
const availableCardIcons = Object.keys( cardIcons ).reduce( ( acc, type ) => {
    if ( ! cardIcons[ type ] || !cardIcons[ type ].icon) {
        return acc;
    }
    
    return [ ...acc, {
        type,
        icon: cardIcons[ type ].icon,
    } ];
}, [] );
```

kmanijak (commented):
Amazing improvements! Thanks for addressing my feedback!

I'm happy we can see payments methods more deterministically appearing in Editor and frontend! I confirm:
- number of icons setting works correctly
- margin and padding work correctly

Payment methods appear reflect the settings

Config | Editor | Frontend
-- | -- | --
image | image | image
image | image | image

One thing to consider it the ordering - seems like due to `sort` in `otherPaymentMethods` the order in editor is different from the frontend. Minor but just mentioning it.

Also, I'm not fully convinced about the card payment methods. In WooPayments config I see more card support icons and I'm curious if sticking to five icons of ours is fine or should be look reach ou to WooPayments folks asking if there's source of those methods somewhere. Feel free to reach out on Slack so we can clarify this aspect 🙌 
image

plugins/woocommerce/client/blocks/bin/webpack-entries.js
mikachan: I see, thank you! In that case, I don't think this block needs to be experimental.

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/edit.tsx
mikachan: Nice, thank you! Done in https://github.com/woocommerce/woocommerce/pull/57181/commits/e10e1bc4d3da0cb727ccce091d40818f7e21211c.

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/edit.tsx
mikachan: Looks good! Done in https://github.com/woocommerce/woocommerce/pull/57181/commits/294348556594e7b579e713201516d2f8c4336aeb.

mikachan:
Thanks again, @kmanijak!

> One thing to consider it the ordering - seems like due to sort in otherPaymentMethods the order in editor is different from the frontend. Minor but just mentioning it.

I've ended up getting and sorting the payment methods server-side with `WC()->payment_gateways->get_available_payment_gateways()`. These are now available in the registry via `wcSettings.availablePaymentMethods`, which means the methods are consistent (and in the same order) across the Editor and the frontend. 

> Also, I'm not fully convinced about the card payment methods. In WooPayments config I see more card support icons and I'm curious if sticking to five icons of ours is fine or should be look reach ou to WooPayments folks asking if there's source of those methods somewhere. 

Yes, I'm also not sure what to do here! Currently, the list of card icons is taken from the card icons available in https://github.com/Automattic/woocommerce-payments/tree/develop/assets/images/payment-method-icons, which is why the list is different to what is shown in Settings > Payments. I chose these icons as that's what is used on the checkout, but there are also card icons listed in https://github.com/Automattic/woocommerce-payments/tree/develop/assets/images/cards. And there's also a couple of hardcoded lists that are different (like [here](https://github.com/Automattic/woocommerce-payments/blob/develop/client/settings/payment-methods-list/payment-method.tsx#L74) and [here](https://github.com/woocommerce/woocommerce/blob/trunk/packages/js/onboarding/src/components/WooPaymentsMethodsLogos/index.tsx#L36)). I'll reach out on Slack and see what's best here. Thank you!

.github/workflows/ci.yml
```diff
@@ -282,6 +282,10 @@ jobs:
           # Finally, run the tests
           pnpm --filter="${{ matrix.projectName }}" ${{ matrix.command }} $lastRunFlag
 
+      - name: 'Print debug log'
```
kalessil: @mikachan : can we get some context on this change, please? I'm wondering if this should be part of any existing artifact (steps below) or if it's a temporary troubleshooting measure related to the PR itself.

mikachan: It's just a temporary solution to try to debug why all the block e2e tests are failing on this PR. I will remove it before merging, although it doesn't seem to be helping me much anyway! I assume I've added something to cause a fatal error so the e2e tests can't run, but I'm unable to reproduce locally and it's difficult to debug, which is why I wanted to see the debug log.

kalessil: Gotcha, thanks for clarifying! If the log can be of use (it doesn't sound like it's helping, though), feel free to include it with other job run artifacts.

mikachan:
I've updated this PR with:
- After discussing the list of card brands with the WooPayments team, I learnt that this list is very unlikely to change, so it is OK to be hardcoded. I've copied over the brand icons from the WooPayment plugin to prevent any file or filename changes from breaking this block.
- Made the block description and some code comments less specific to WooPayments, as the block now includes other gateway payment method icons if they're available.

I think this is ready for a final review! cc @kmanijak if you get a chance to take another look 🙇

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/edit.tsx
```diff
@@ -0,0 +1,119 @@
...
+	icon,
+}: {
+	type: string | undefined;
+	icon: string | undefined;
+} ) => {
+	let CardIcon = null;
+
+	if ( type && icon ) {
+		CardIcon = (
+			<div className="wp-block-woocommerce-payment-method-icons__item">
```
kmanijak: We usually mark WooCommerce blocks with `wc` prefix - sorry I didn't spot that earlier!

```suggestion
			<div className="wc-block-payment-method-icons__item">
```

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/edit.tsx
```diff
@@ -0,0 +1,119 @@
...
+				<span
+					className="wp-block-woocommerce-payment-method-icons__icon"
```
kmanijak:
```suggestion
					className="wc-block-payment-method-icons__icon"
```

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+import './style.scss';
+
+registerBlockType( metadata.name, {
+	...( metadata as BlockConfiguration ),
+	icon: {
+		src: <Icon icon={ paymentMethodsIcon } />,
+	},
+	edit,
+	save: () => null,
+} );
```
kmanijak: Minor but `registerBlockType` also allows for:

```suggestion
registerBlockType( metadata, {
	icon: {
		src: <Icon icon={ paymentMethodsIcon } />,
	},
	edit,
	save: () => null,
} );
```

which feels simpler for me but I don't have strong opinion, just sharing the option.

plugins/woocommerce/src/Blocks/BlockTypes/PaymentMethodIcons.php
```diff
@@ -0,0 +1,221 @@
...
+	 * @param WP_Block $block      Block instance.
+	 * @return string Rendered block type output.
+	 */
+	protected function render( $attributes, $content, $block ) {
+		$enabled_card_types    = $this->get_enabled_card_types();
+		$other_payment_methods = $this->get_other_payment_method_icons();
+
+		if ( empty( $enabled_card_types ) && empty( $other_payment_methods ) ) {
+			return '';
+		}
```
kmanijak: Super minor but seems it could be replaced with `get_available_payment_methods` for simplicity

```suggestion
		$payment_methods = $get_available_payment_methods();

		if ( empty( $payment_methods ) ) {
			return '';
		}
```

plugins/woocommerce/src/Blocks/BlockTypes/PaymentMethodIcons.php
```diff
@@ -0,0 +1,221 @@
...
+
+		$all_payment_methods = $this->get_available_payment_methods();
+		$number_of_icons     = $attributes['numberOfIcons'] ?? 0;
+		$number_of_icons     = 0 === $number_of_icons ? count( $all_payment_methods ) : max( 0, min( intval( $number_of_icons ), count( $all_payment_methods ) ) );
+
+		if ( ! empty( $all_payment_methods ) ) {
+			for ( $i = 0; $i < $number_of_icons; $i++ ) {
+				$payment_method = $all_payment_methods[ $i ];
+				$output        .= '<div class="wp-block-woocommerce-payment-method-icons__item">';
+				$output        .= '<span class="wp-block-woocommerce-payment-method-icons__icon" style="background-image: url(\'' . \esc_url( $payment_method['icon'] ) . '\');" role="img" aria-label="' . \esc_attr( $payment_method['name'] ) . '"></span>';
```
kmanijak:
```suggestion
				$output        .= '<div class="wc-block-payment-method-icons__item">';
				$output        .= '<span class="wc-block-payment-method-icons__icon" style="background-image: url(\'' . \esc_url( $payment_method['icon'] ) . '\');" role="img" aria-label="' . \esc_attr( $payment_method['name'] ) . '"></span>';
```

kmanijak (changes requested):
I really appreciate reaching out to Woo Payments team! <3 And amazing effort overall, I like this block more and more! 😄 

But while testing I found discrepancies between editor and frontend, there are missing icons on the frontend while they were there with the previous implementation.


This is a default WooPayments config after activation:

Config | Editor | Frontend
-- | -- | --
image | image | image

And here's some edited config:

Config | Editor | Frontend
-- | -- | --
image | image | image

So seems like "Buy now, pay later" options are not appearing on the frontend 😞 

Otherwise, I left some minor code comments (apologies for back and forth, I didn't spot classnames not following usual convention 😅 ). But overall I think once we clarify BNPY situation:
- if not available on the frontend we should also hide them on editor
- or find a way to display them on the frontend
I'll be happy to approve the PR! 💪

plugins/woocommerce/client/blocks/assets/js/blocks/payment-method-icons/index.tsx
mikachan: Looks good to me!

plugins/woocommerce/src/Blocks/BlockTypes/PaymentMethodIcons.php
mikachan: Yes, thank you! I missed this when refactoring.

mikachan:
Thanks @kmanijak!

> But while testing I found discrepancies between editor and frontend, there are missing icons on the frontend while they were there with the previous implementation.

I think this is because the Editor and the frontend are using different contexts to show the available payment methods from `get_available_payment_gateways()`. I've noticed the same discrepancies, where I've enabled multiple BNPL methods, but only the ones I'd expect to be available in my country are shown on the frontend, and they are all shown in the Editor.

I'm not sure how to work around this but I'll look into it. Would you expect the Editor to only show the payment methods available to the current user?

kmanijak:
> I think this is because the Editor and the frontend are using different contexts to show the available payment methods from get_available_payment_gateways(). I've noticed the same discrepancies, where I've enabled multiple BNPL methods, but only the ones I'd expect to be available in my country are shown on the frontend, and they are all shown in the Editor.
> I'm not sure how to work around this but I'll look into it. Would you expect the Editor to only show the payment methods available to the current user?

So you mean:
- Editor shows all available methods _from merchant's perspective_ (basically what's in WooPayments config)
- Frontend shows available methods _from shopper's perspective_ (what's available for them)?

I didn't think of it! If that's the case I'm totally ok with it! Please confirm and I'll be happy to approve 😄

kmanijak (changes requested):
I wanted to give it final round of testing before approving given the explanation you gave and I found out an issue! 😞 

Before setting up payments, when you use block, the frontend shows critical error:

Editor | Frontend
-- | --
image | image

When I remove the block, it works correctly:

Editor | Frontend
-- | --
image | image

It's a bit of an edge case but nothing prevents setting up the layout and using the block before actually setting up the payments and all the other stuff so this should be fixed before merging. 🙌

mikachan:
> So you mean:

>Editor shows all available methods from merchant's perspective (basically what's in WooPayments config)
>Frontend shows available methods from shopper's perspective (what's available for them)?

Yes, exactly!

> Before setting up payments, when you use block, the frontend shows critical error:

Good spot! Looks like this was from one of the recent changes, `get_available_payment_methods()` was being called incorrectly. This should be fixed now.

kmanijak (approved):
I did a final round of testing and all looks good!

- Blocks works correctly before WooPayments setup
- Blocks works correctly with WooPayments displaying:
    - all the available methods in editor (from merchant's POV)
    - available methods in frontend (from buyer's POV which may be a subset)
- Block respects inspector controls settings.

Appreciate the responsiveness and addressing all of the feedback! Great work!

Let's 🚢 it!

PR #57216, opened 2025-04-14
Fix coming soon breaks other page

chihsuan:

https://wordpress.org/support/topic/coming-soon-apply-to-store-pages-only-breaks-other-checkout-page/

We improved the coming soon matching logic for store pages in https://github.com/woocommerce/woocommerce/pull/54906 to fix the issue. It has covered most of the cases, but there is still a edge case that are not covered: When the checkout page is set to a non-default page.

The issue is because we're using `page_slug` to match the coming soon page, but when the checkout page is set to a non-default page, the `page_slug` to match the coming soon page is not correct.

This PR fixes the issue by removing the slug check and only use the page ID to match the coming soon page which is more reliable. I also added a test to cover this case. 

(For Bug Fixes) Bug introduced in PR https://github.com/woocommerce/woocommerce/pull/54906

1. Enable coming soon mode `/wp-admin/admin.php?page=wc-settings&tab=site-visibility` and turn on `Apply to store pages only`
2. Create a new page
3. Set the checkout page to the new page `/wp-admin/admin.php?page=wc-settings&tab=advanced`
4. Open an incognito window and go to the new checkout page in the frontend 
5. You should see the coming soon page
6. Go to the default checkout page in the frontend
7. You should see the checkout page (not the coming soon page)

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/src/Admin/WCAdminHelper.php
+++ b/plugins/woocommerce/src/Admin/WCAdminHelper.php
@@ -149,7 +149,7 @@ class WCAdminHelper {
 		$store_pages = apply_filters( 'woocommerce_store_pages', $store_pages );
 
 		foreach ( $store_pages as $page_slug => $page_id ) {
-			if ( is_page( $page_slug ) || ( $page_id > 0 && is_page( $page_id ) ) ) {
+			if ( $page_id > 0 && is_page( $page_id ) ) {
 				return true;
 			}
 		}
--- a/plugins/woocommerce/changelog/fix-coming-soon-breaks-other-page
+++ b/plugins/woocommerce/changelog/fix-coming-soon-breaks-other-page
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix coming soon breaks other page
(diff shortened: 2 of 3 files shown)


jorgeatorres (approved):
This tests well for me (thanks @chihsuan!), but I'll let @louwie17 give the final approval before merging.

plugins/woocommerce/src/Admin/WCAdminHelper.php
```diff
@@ -149,7 +149,7 @@ public static function is_current_page_store_page() {
 		$store_pages = apply_filters( 'woocommerce_store_pages', $store_pages );
 
 		foreach ( $store_pages as $page_slug => $page_id ) {
-			if ( is_page( $page_slug ) || ( $page_id > 0 && is_page( $page_id ) ) ) {
+			if ( $page_id > 0 && is_page( $page_id ) ) {
```
louwie17, line 152: Lets leave this as is, as it fixes the issue and I don't want to accidently introduce something new.
But it does look like instead of doing a foreach we could of also done:
`if ( is_page( array_values( $store_pages ) ) { return true; }` as `is_page` also supports an array: https://developer.wordpress.org/reference/functions/is_page/

louwie17 (approved):
Tested well, thanks for addressing this @chihsuan :)

PR #57225, opened 2025-04-14
Accept the `created_via` param to the `POST /orders` endpoint

### LINKED ISSUE #56670: Accept the `created_via` field in the `POST /wp-json/wc/v3/orders` endpoint request

albarin:

This PR allows setting the `created_via` parameters for the `POST /wp-json/wc/v3/orders` endpoint.
When not sent, `created_via` will be set to the default `rest-api`.

Closes #56670

1. Go to `wp-admin/admin.php?page=wc-settings&tab=advanced&section=keys` and add a new key with `Read/Write` permissions.
2. Make the following request using the consumer key and secret generated in the previous step, adding the `created_via` param in the body:
```bash
curl --insecure --request POST \
  --url https://[STORE_URL]/wp-json/wc/v3/orders \
  -u [KEY]:[SECRET] \
  --header 'Content-Type: application/json' \
  --data '{
  "created_via": "[somevalue]"
}'
```
3. Confirm the created order has the value you chose for the `created_via` parameter.
4. Run the same request again, but without the `created_via`.
5. Confirm the created order has the value `rest-api` in the `created_via` parameter.
6. Run a PUT request to update one of the orders you just created and try to update the `created_via` field:
```bash
curl --insecure --request PUT \
  --url https://[STORE_URL]/wp-json/wc/v3/orders/[ORDER-ID] \
  -u [KEY]:[SECRET] \
  --header 'Content-Type: application/json' \
  --data '{
  "created_via": "[a new value]"
}'
```
7. Make sure the `created_via` field is **NOT updated** and it still has the original value.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Accept the "created_via" param to the "POST /orders" endpoint.
</details>

<details>

#### Comment 

</details>


malinajirka:
Thanks @albarin! I've tested the behavior, including requests through the jetpack tunnel, and everything seems to be working as expected.

I also tested some malicious entries and I'm wondering if the current behavior is expected or we need to do some extra sanitization. Both SQL injection and XSS kept the entries exactly as they were - I mean, I couldn't find an actual issue, since the code wasn't interpreted, but I'm wondering if it'd be safer to just escape these immediately.

albarin:
> I also tested some malicious entries and I'm wondering if the current behavior is expected or we need to do some extra sanitization. Both SQL injection and XSS kept the entries exactly as they were - I mean, I couldn't find an actual issue, since the code wasn't interpreted, but I'm wondering if it'd be safer to just escape these immediately.

@malinajirka Thanks, I think you are right, I've added some extra sanitization.

Also, I realised something was missing: we agreed to not allow updating `created_via` on `PUT /orders`, so I've pushed that change here https://github.com/woocommerce/woocommerce/pull/57225/commits/08053dc41f068eba428a6250624e55270e659184 and updated the testing steps. Could you have another look?

malinajirka (approved):
I've tested the changes and it seems to work as expected, thank you! It should still be reviewed by someone with more PHP experience.

plugins/woocommerce/includes/rest-api/Controllers/Version3/class-wc-rest-orders-controller.php
```diff
@@ -118,6 +118,12 @@ protected function prepare_object_for_database( $request, $creating = false ) {
 
 			if ( ! is_null( $value ) ) {
 				switch ( $key ) {
+					case 'created_via':
```
kalessil, line 122: I am puzzled a bit by the removal of `'readonly' => true` in controllers V1 and V2 without replicating the changes from this file in those controllers. What's our thinking behind this approach?

kalessil (commented):
LGTM! Manual and unit tests are passing :heavy_check_mark: 

Leaving a question to build more context on how we updating core REST APIs.

kalessil (approved):

plugins/woocommerce/includes/rest-api/Controllers/Version3/class-wc-rest-orders-controller.php
albarin: Oops, you are right, that didn't make sense, I've reverted the change on https://github.com/woocommerce/woocommerce/pull/57225/commits/084ddf6cc8675a84ac40e0f615e8d99c40ba7d09

kalessil: Looks good, thanks! We are good to merge.

PR #49643, opened 2024-07-17
Add description and delivery time to shipping methods in the Cart and Checkout blocks

opr:
### Changes proposed in this Pull Request:

Adds two new properties to the data of a `WC_Shipping_Rate`: `delivery_time` and `description`.

Both are strings, and they will be used to display a delivery time, and a description of the shipping rate in the Checkout block.

The design is here, however we are not going to implement the lighter colour of the delivery time due to issues with different coloured themes.

image

The code for displaying these properties already exists in the Checkout block, but it was never used as these properties didn't exist.

Closes #42667

### How to test the changes in this Pull Request:

Using the [WooCommerce Testing Instructions Guide](https://github.com/woocommerce/woocommerce/wiki/Writing-high-quality-testing-instructions), include your detailed testing instructions:

1. Add several shipping methods to your store. A mix of free and flate rate shipping options should be used. Local pickup _can_ be added, but isn't relevant to this, as it depends on the extension's use of `delivery_time` and `description` props, rather than anything we control.
2. Add the following code using a plugin like [Code Snippets](https://wordpress.org/plugins/code-snippets/)
```php
function custom_shipping_description( $rates, $package ) {
    foreach ( $rates as $rate_id => $rate ) {
		if ( 'pickup_location' === $rate->method_id ) {
			continue;
		}
		$rate->description = 'This is a custom description for ' . $rate->label . '.';
		$rate->delivery_time = 'This will be delivered in ' . rand( 1, 5 ) . ' days.';
	}
    return $rates;
}
add_filter( 'woocommerce_package_rates', 'custom_shipping_description', 10, 2 );
```
4. Ensure you have `WooCommerce -> Settings -> Shipping -> Shipping settings -> Enable the shipping calculator on the cart page` checked.
5. Add an item to your cart and go to the Cart block.
6. Ensure you see the rate descriptions and delivery times in the shipping calculator, like so:
image

7. Go to the Checkout block. Ensure you see the descriptions and delivery times in the shipping methods block, like so:
image

8. Change the text in the snippet above to be quite long, e.g. like so:
```php
function custom_shipping_description( $rates, $package ) {
    foreach ( $rates as $rate_id => $rate ) {
		if ( 'pickup_location' === $rate->method_id ) {
			continue;
		}
		$rate->description = 'This is a custom description for ' . $rate->label . '. It is quite a long, descriptive description that should flow onto multiple lines, for testing purposes.';
[...]

PR #57229, opened 2025-04-14
Fix longer shipping description being misaligned in Firefox

straku:

Closes [WOOPLUG-3060](https://linear.app/a8c/issue/WOOPLUG-3060/shipping-description-in-checkout-block-misaligned-in-firefox)

Bug introduced in https://github.com/woocommerce/woocommerce/pull/49643

`RadioControl` component relied on `display: table` for layout of the label, value and description (primary and secondary). When using table display, it is important for the last element in the tree to have `display: table-cell` property (it wasn't causing any issues in Chrome or Safari, but Firefox had troubles). In the PR mentioned above, overwrite for the styles was applied at the call site level, changing `display: table-cell` to `display: flex`, it was done to change the layout from horizontal and vertical and control the order of the items.

What I am doing here:
- remove style overwrites and add native support to the `RadioControl` component to control description layout
- while I was there I figured it's probably a wise move to remove our reliance on `display: table` and use standard `flex` instead as it will prevent similar (difficult to debug) problems in the future
- I also noticed that the default styles for `RadioControl` are a bit broken (not enough space for the input itself, [see storybook story build from trunk to see it](https://woocommerce.github.io/woocommerce/?path=/docs/woocommerce-blocks_external-components-radiocontrol--docs)), so I fixed it too

| Before | After |
|--------|-------|
| Screenshot 2025-04-14 at 12 00 37 | Screenshot 2025-04-14 at 11 59 57 |

Using the [WooCommerce Testing Instructions Guide](https://github.com/woocommerce/woocommerce/wiki/Writing-high-quality-testing-instructions), include your detailed testing instructions.

Acknowledging the fact that I mostly copied the below steps from @opr original PR. Thanks 😊

Please perform steps below on at least two browsers:
- **Firefox is a must** (the issue was only reproducible there)
- Chrome or Safari

1. Add several shipping methods to your store. A mix of free and flat rate shipping options should be used. Local pickup _can_ be added, but isn't relevant to this, as it depends on the extension's use of `delivery_time` and `description` props, rather than anything we control.
2. Add the following code using a plugin like [Code Snippets](https://wordpress.org/plugins/code-snippets/)
```php
function custom_shipping_description( $rates, $package ) {
    foreach ( $rates as $rate_id => $rate ) {
		if ( 'pickup_location' === $rate->method_id ) {
			continue;
		}
		$rate->description = 'This is a custom description for ' . $rate->label . '. It is quite a long, descriptive description that should flow onto multiple lines, for testing purposes.';
		$rate->delivery_time = 'This will be delivered in ' . rand( 1, 5 ) . ' days.';
	}
    return $rates;
}
add_filter( 'woocommerce_package_rates', 'custom_shipping_description', 10, 2 );
```
4. Add an item to your cart and process to the Checkout block. Ensure it looks OK with the text on multiple lines. There should be no widows/runts (single words on their own line at the end of a block of text) in the description or delivery time.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message

Fix wrong alignment of longer shipping option descriptions in Firefox

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/blocks/packages/components/radio-control/index.tsx
+++ b/plugins/woocommerce/client/blocks/packages/components/radio-control/index.tsx
@@ -20,6 +20,7 @@ const RadioControl = ( {
 	options = [],
 	disabled = false,
 	highlightChecked = false,
+	descriptionStackingDirection = 'row',
 }: RadioControlProps ): JSX.Element | null => {
 	const instanceId = useInstanceId( RadioControl );
 	const radioControlId = id || instanceId;
@@ -62,6 +63,9 @@ const RadioControl = ( {
 						}
 					} }
 					disabled={ disabled }
+					descriptionStackingDirection={
+						descriptionStackingDirection
+					}
 				/>
 			) ) }
 		</div>
--- a/plugins/woocommerce/client/blocks/packages/components/radio-control/option-layout.tsx
+++ b/plugins/woocommerce/client/blocks/packages/components/radio-control/option-layout.tsx
@@ -41,8 +41,8 @@ const OptionLayout = ( {
 					className={ clsx(
 						'wc-block-components-radio-control__description-group',
 						{
-							'wc-block-components-radio-control__description-group--row':
-								descriptionStackingDirection === 'row',
+							'wc-block-components-radio-control__description-group--column':
+								descriptionStackingDirection === 'column',
 						}
 					) }
 				>
--- a/plugins/woocommerce/client/blocks/packages/components/radio-control/option-layout.tsx
+++ b/plugins/woocommerce/client/blocks/packages/components/radio-control/option-layout.tsx
@@ -1,4 +1,9 @@
 /**
+ * External dependencies
+ */
+import clsx from 'clsx';
+
+/**
  * Internal dependencies
  */
 import type { RadioControlOptionLayout } from './types';
@@ -9,6 +14,7 @@ const OptionLayout = ( {
 	description,
 	secondaryDescription,
 	id,
+	descriptionStackingDirection,
 }: RadioControlOptionLayout ): JSX.Element => {
 	return (
 		<div className="wc-block-components-radio-control__option-layout">
@@ -31,7 +37,12 @@ const OptionLayout = ( {
 				) }
 			</div>
 			{ ( description || secondaryDescription ) && (
-				<div className="wc-block-components-radio-control__description-group">
+				<div
+					className={ clsx(
+						'wc-block-components-radio-control__description-group',
+						`wc-block-components-radio-control__description-group--${ descriptionStackingDirection }`
+					) }
+				>
 					{ description && (
 						<span
 							id={ id && `${ id }__description` }
--- a/plugins/woocommerce/client/blocks/packages/components/radio-control/option-layout.tsx
+++ b/plugins/woocommerce/client/blocks/packages/components/radio-control/option-layout.tsx
@@ -40,7 +40,10 @@ const OptionLayout = ( {
 				<div
 					className={ clsx(
 						'wc-block-components-radio-control__description-group',
-						`wc-block-components-radio-control__description-group--${ descriptionStackingDirection }`
+						{
+							'wc-block-components-radio-control__description-group--row':
+								descriptionStackingDirection === 'row',
+						}
 					) }
 				>
 					{ description && (
--- a/plugins/woocommerce/client/blocks/packages/components/radio-control/style.scss
+++ b/plugins/woocommerce/client/blocks/packages/components/radio-control/style.scss
@@ -32,7 +32,8 @@
 		}
 
 		// This rule removes the fake border-top from the selected element to prevent a double border.
-		&.wc-block-components-radio-control-accordion-option--checked-option-highlighted + div.wc-block-components-radio-control-accordion-option::after {
+		&.wc-block-components-radio-control-accordion-option--checked-option-highlighted
+			+ div.wc-block-components-radio-control-accordion-option::after {
 			display: none;
 		}
 	}
@@ -84,7 +85,8 @@
 
 	// Remove the fake border-top from the item after the selected element, this is to prevent a double border with the
 	// selected element's box-shadow.
-	.wc-block-components-radio-control__option--checked-option-highlighted + .wc-block-components-radio-control__option::after {
+	.wc-block-components-radio-control__option--checked-option-highlighted
+		+ .wc-block-components-radio-control__option::after {
 		display: none;
 	}
 
@@ -137,7 +139,7 @@
 }
 
 .wc-block-components-radio-control__option
-.wc-block-components-radio-control__option-layout {
+	.wc-block-components-radio-control__option-layout {
 	&::after {
 		display: none;
 	}
@@ -159,8 +161,8 @@
 // There is no variant handling row stacking as it is applied by default.
 .wc-block-components-radio-control__description-group--column {
 	flex-direction: column;
-	
-	// Use balance as this is not a good candidate for pretty. Pretty is 
+
+	// Use balance as this is not a good candidate for pretty. Pretty is
 	// intended  for body copy which this is not.
 	text-wrap: balance;
 
@@ -263,7 +265,7 @@
 
 .theme-twentytwentyone {
 	.wc-block-components-radio-control
-	.wc-block-components-radio-control__input {
+		.wc-block-components-radio-control__input {
 		&:checked {
 			border-width: 2px;
 
--- a/plugins/woocommerce/client/blocks/packages/components/radio-control/style.scss
+++ b/plugins/woocommerce/client/blocks/packages/components/radio-control/style.scss
@@ -121,7 +121,11 @@
 	display: block;
 	margin: em($gap) 0;
 	margin-top: 0;
-	padding: 0 0 0 em($gap-larger);
+	// padding-left here is aligned with the size of the radio input element,
+	// we need to reserve at least this amount of space for the label to not
+	// overflow the input. Input has position: absolute set, so it doesn't
+	// take any space on its own.
+	padding: 0 calc($gap + em(24px)) 0 em($gap-huge);
 	position: relative;
 	cursor: pointer;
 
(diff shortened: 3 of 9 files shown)


tjcafferkey:
I have no idea if this is expected in the dev env yet, nor have I looked where these values are coming from but I'm experiencing some weird issue with the expected delivery time when switching between each select option.

https://github.com/user-attachments/assets/b63a722a-be7f-48ce-a359-37e9f0ff4494

tjcafferkey:
>I have no idea if this is expected in the dev env yet, nor have I looked where these values are coming from but I'm experiencing some weird issue with the expected delivery time when switching between each select option.

🤦🏻 its defined in your snippet so its expected behavior. how did i miss that.

plugins/woocommerce/client/blocks/packages/components/radio-control/option-layout.tsx
```diff
@@ -31,7 +37,12 @@ const OptionLayout = ( {
 				) }
 			</div>
 			{ ( description || secondaryDescription ) && (
-				<div className="wc-block-components-radio-control__description-group">
+				<div
+					className={ clsx(
```
tjcafferkey, line 41: Nit (not blocking): Seems a really simple reason to introduce another dependency when you could just do:

`wc-block-components-radio-control__description-group wc-block-components-radio-control__description-group--${ descriptionStackingDirection }`

plugins/woocommerce/client/blocks/packages/components/radio-control/index.tsx
```diff
@@ -20,6 +20,7 @@ const RadioControl = ( {
 	options = [],
 	disabled = false,
 	highlightChecked = false,
+	descriptionStackingDirection = 'row',
```
tjcafferkey, line 23: Curious as to why we're defaulting to `row` if we've introduced this prop to use `column`? `column` as a default kind of makes more sense to me but just curious as to the reason, does `row` more accurately reflect the behavior prior to this pr?

tjcafferkey (approved):
Thank you for fixing this @straku! This tests well for me, I've left a few comments but none that should block this so I'm approving. 👏🏻 

I tested using the testing steps define in the description as well as with local pickup as an option since the component is also used here it looks like.

CleanShot 2025-04-15 at 08 56 43@2x

```
### WordPress Environment ###

WC Version: 9.9.0-dev
Log Directory Writable: ✔
WP Version: 6.7.2
WP Multisite: –
WP Memory Limit: 256 MB
WP Debug Mode: ✔
WP Cron: ✔
Language: en_US
External object cache: –

### Server Environment ###

Server Info: nginx/1.26.1
PHP Version: 8.1.23
PHP Post Max Size: 1,000 MB
PHP Time Limit: 1200
PHP Max Input Vars: 4000
cURL Version: 8.7.1
(SecureTransport) LibreSSL/3.3.6

SUHOSIN Installed: –
MySQL Version: 8.0.35
Max Upload Size: 300 MB
Default Timezone is UTC: ✔
fsockopen/cURL: ✔
SoapClient: ✔
DOMDocument: ✔
GZip: ✔
Multibyte String: ✔
Remote Post: ✔
Remote Get: ✔

### Database ###

[REDACTED]

### Post Type Counts ###

attachment: 23
page: 8
post: 2
product: 18
product_variation: 7
revision: 7
shop_order_placehold: 6
wp_font_face: 36
wp_font_family: 12
wp_global_styles: 1
wp_navigation: 1
wp_template: 1

### Security ###

Secure connection (HTTPS): ✔
Hide errors from visitors: ✔

### Active Plugins (6) ###

Code Snippets: by Code Snippets Pro – 3.6.8
Gutenberg: by Gutenberg Team – 20.6.0
WooPayments: by WooCommerce – 9.1.0 (update to version 9.2.0 is available)
WooCommerce: by Automattic – 9.9.0-dev

### Inactive Plugins (8) ###

Akismet Anti-spam: Spam Protection: by Automattic - Anti-spam Team – 5.3.7
Core Rollback: by Andy Fragen – 1.3.7
Jetpack: by Automattic – 14.5
SQL Buddy: by Delicious Brains – 1.0.0
Transients Manager: by WPBeginner – 2.0.7
WooCommerce Additional Variation Images: by WooCommerce – 2.3.2
WordPress Beta Tester: by WordPress Upgrade/Install Team – 3.6.3
WP Downgrade | Specific Core Version: by Reisetiger – 1.2.6

### Dropin Plugins () ###

### Settings ###

Legacy API Enabled: –
Force SSL: –
Currency: GBP (£)
Currency Position: left
Thousand Separator: ,
Decimal Separator: .
Number of Decimals: 2
grouped (grouped)
simple (simple)
variable (variable)

featured (featured)
outofstock (outofstock)
rated-1 (rated-1)
rated-2 (rated-2)
rated-3 (rated-3)
rated-4 (rated-4)
rated-5 (rated-5)

HPOS feature enabled: ✔
HPOS data sync enabled: –

### Logging ###

Enabled: ✔
Retention period: 30 days
Level threshold: –
Log directory size: 521 KB

### WC Pages ###

Shop base: #7 - /shop/

### Theme ###

Name: Twenty Twenty-Five
Version: 1.0 (update to version 1.1 is available)
Theme type: Block theme

### Templates ###

Overrides: /Users/tom/A8c/woocommerce/plugins/woocommerce/templates/block-notices/error.php
/Users/tom/A8c/woocommerce/plugins/woocommerce/templates/block-notices/notice.php
/Users/tom/A8c/woocommerce/plugins/woocommerce/templates/block-notices/success.php


### WooPayments ###

Version: 9.1.0
Connected to WPCOM: ❌ No
Logging: Disabled

### Admin ###

analytics
product-block-editor
experimental-blocks
coupons
core-profiler
customize-store
customer-effort-score-tracks
import-products-task
shipping-smart-defaults
shipping-setting-tour
homescreen
marketing
minified-js
mobile-app-banner
onboarding
onboarding-tasks
payment-gateway-suggestions
product-custom-fields
printful
remote-inbox-notifications
remote-free-extensions
shipping-label-banner
subscriptions
store-alerts
transient-notices
woo-mobile-welcome
wc-pay-promotion
wc-pay-welcome-page
launch-your-store
blockified-add-to-cart
disable-core-profiler-fallback

product-pre-publish-modal
settings
use-wp-horizon
beta-tester-slotfill-examples

Daily Cron: ✔ Next scheduled: 2025-04-15 10:25:00 +00:00
Options: ✔
Notes: 65
Onboarding: skipped

### Action Scheduler ###

Complete: 157
Oldest: 2025-04-09 10:26:39 +0000
Newest: 2025-04-15 07:50:42 +0000

Failed: 1
Oldest: 2025-04-09 10:26:47 +0000
Newest: 2025-04-09 10:26:47 +0000

Pending: 2
Oldest: 2025-04-15 07:58:26 +0000
Newest: 2025-04-15 10:47:44 +0000


Generated at: 2025-04-15 07:58:36 +00:00
```
</details>

plugins/woocommerce/client/blocks/packages/components/radio-control/style.scss
```diff
@@ -144,16 +149,34 @@
...
 
 	.wc-block-components-radio-control__secondary-label,
 	.wc-block-components-radio-control__secondary-description {
 		text-align: right;
-		min-width: 50%;
+		flex-basis: 50%;
+	}
+}
+
+// For column description stacking we need to apply some layout changes.
```
opr, line 162: Thanks a lot for adding these comments here, super helpful!

plugins/woocommerce/client/blocks/packages/components/radio-control/index.tsx
opr: Because this radio control is used by several other components. `row` is the default behaviour and we only want to apply `column` when it's used in the shipping methods block.

plugins/woocommerce/client/blocks/packages/components/radio-control/option-layout.tsx
opr: Since we're not applying any styling to `--row` classes, could we change this so we only add `wc-block-components-radio-control__description-group--column` if `descriptionStackingDirection` is `column`? We'd prevent outputting `wc-block-components-radio-control__description-group--row` then.

Though I'm wondering if we would prefer to keep this in case we _do_ want to add styling to `--row` classes in future.

plugins/woocommerce/client/blocks/packages/components/radio-control/index.tsx
straku: It's the current behavior, I wanted to retain backwards compatibility.

plugins/woocommerce/client/blocks/packages/components/radio-control/option-layout.tsx
straku: I agree in general, but in this case, I would leave it in just for the sake of readability (keeping the lines shorter). `clsx` is not a dependency that we can realistically ever get rid of, so it doesn't feel that bad.

opr (approved):
Thanks @straku this PR is great! I left only one comment that might need action, the one about `clsx` so please double check that before merging. I don't think it's blocking so approving.

Noting that I tested this in Firefox and Edge, I also tested on mobile and it all looks good. I tested with long words trying to make widows appear but none showed up 👌🏼

plugins/woocommerce/client/blocks/packages/components/radio-control/option-layout.tsx
straku: @opr The reply above was written before yours appeared (don't know why, probably yours were part of the code review flow, not an ad hoc comment?). I applied the change you suggested. If it's not needed now, it most likely shouldn't be there.

opr: I think it's because I added it in bulk as part of the review, sorry for any confusion.

I think the change below is wrong, it should be for `column` right? Since `row` is default and has no styling we only want to do this when it's in column mode.

opr (approved):
Thanks for making that change, tested and the text is showing up well in Firefox and Edge.

PR #57240, opened 2025-04-14
Move generateUniqueId to @woocommerce/utils package

### LINKED ISSUE #57156: Create a shared generateUniqueId() utility function
### Background
While working on [PR](https://github.com/woocommerce/woocommerce/pull/57091#discussion_r2035674974), we identified that there are multiple implementations of `generateUniqueId()` scattered throughout the codebase:

* [plugins/woocommerce/client/blocks/assets/js/base/context/event-emit/reducer.ts](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/base/context/event-emit/reducer.ts#L11-L13)
* [plugins/woocommerce/client/blocks/assets/js/blocks/attribute-filter/utils.ts](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/blocks/attribute-filter/utils.ts#L24-L26)
* [plugins/woocommerce/client/blocks/assets/js/blocks/rating-filter/utils.ts](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/blocks/rating-filter/utils.ts#L26-L28)
* [plugins/woocommerce/client/blocks/assets/js/blocks/stock-filter/utils.ts](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/blocks/stock-filter/utils.ts#L31-L33)

Each implements essentially the same functionality but in slightly different ways.

In our new [useApplyEditorStyles](https://github.com/woocommerce/woocommerce/blob/cdf59e82bcb8636133a94414bb6bd9f55b737e9e/plugins/woocommerce/client/blocks/assets/js/shared/hooks/use-theme-colors.ts#L17) hook, we currently require consumers to provide a `styleId` parameter:
- [Mini-Cart Block](https://github.com/woocommerce/woocommerce/blob/cdf59e82bcb8636133a94414bb6bd9f55b737e9e/plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/edit.tsx#L101)
- [Add to Cart with Options Variation Selector Attribute Options Block](https://github.com/woocommerce/woocommerce/blob/cdf59e82bcb8636133a94414bb6bd9f55b737e9e/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/edit.tsx#L72)

But ideally this could be auto-generated internally using a shared utility.

### Proposed Solution
Create a single, shared implementation of `generateUniqueId()` that can be used across the codebase. This would:

1. Reduce code duplication
2. Ensure consistent ID generation logic
3. Simplify usage of components/hooks that need unique IDs

### Implementation Details
1. Create a new utility function in an appropriate shared location (e.g., [plugins/woocommerce/client/blocks/assets/js/utils/](https://github.com/woocommerce/woocommerce/tree/trunk/plugins/woocommerce/client/blocks/assets/js/utils))
2. Update existing implementations to use this shared utility
3. Modify the `useApplyEditorStyles` hook to generate IDs internally rather than requiring them as parameters

amitraj2203:

Currently, there are four different implementations of `generateUniqueId()`:

* [plugins/woocommerce/client/blocks/assets/js/base/context/event-emit/reducer.ts](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/base/context/event-emit/reducer.ts#L11-L13)
* [plugins/woocommerce/client/blocks/assets/js/blocks/attribute-filter/utils.ts](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/blocks/attribute-filter/utils.ts#L24-L26)
* [plugins/woocommerce/client/blocks/assets/js/blocks/rating-filter/utils.ts](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/blocks/rating-filter/utils.ts#L26-L28)
* [plugins/woocommerce/client/blocks/assets/js/blocks/stock-filter/utils.ts](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/blocks/stock-filter/utils.ts#L31-L33)

This PR:

- Creates a new shared utility function in `plugins/woocommerce/client/blocks/assets/js/utils/get-unique-id.ts` to replace multiple implementations scattered throughout the codebase.

Closes #57156 

#### **Testing Filter Blocks**

1. Navigate to the **Product Catalog** template:  
   _Appearance > Editor > Templates > Product Catalog_.
2. Edit the template.
3. Switch to **Code editor** mode.
4. Copy and paste the following filter blocks, then return to **Visual editor** mode.

<details>
<summary>Filter Blocks</summary>

```html
<!-- wp:woocommerce/attribute-filter {"attributeId":2,"showCounts":true,"queryType":"and","displayStyle":"dropdown","heading":"Filter by Color"} -->
<div class="wp-block-woocommerce-attribute-filter is-loading"></div>
<!-- /wp:woocommerce/attribute-filter -->
```

```html
<!-- wp:woocommerce/stock-filter {"showCounts":true} -->
<div class="wp-block-woocommerce-stock-filter is-loading"></div>
<!-- /wp:woocommerce/stock-filter -->
```

```html
<!-- wp:woocommerce/rating-filter {"showCounts":true,"lock":{"remove":true}} -->
<div class="wp-block-woocommerce-rating-filter is-loading"></div>
<!-- /wp:woocommerce/rating-filter -->
```

</details>

5. Save the changes and go to the **Shop** page.

---

##### **Attribute Filter Block**
- The `generateUniqueId()` function is used to generate a unique key for the `FormTokenField` component, allowing it to properly remount when filters are reset.
- Verify the dropdown-style filter functions correctly.
- Ensure the attribute filter dropdown opens and closes properly.
- Select multiple attribute options and confirm that filtering works.
- Click the **Reset** button — the dropdown should clear completely and return to its initial state.

---

##### **Rating Filter Block**
- The `generateUniqueId()` function is used to generate a unique key for the `FormTokenField` component, allowing it to remount after filter changes.
- Select different star ratings and verify that products are filtered accordingly.
- Click the **Reset** button — the rating filter should fully reset.
- Products should update based on the removed filters.

---

##### **Stock Filter Block**
- The `generateUniqueId()` function is used to ensure that the `FormTokenField` component remounts properly when stock filters are reset.
- Select various stock statuses (e.g., In Stock, Out of Stock, On Backorder).
- Confirm that products are filtered according to the selected stock statuses.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/shared/hooks/use-theme-colors.ts
```diff
@@ -1,24 +1,32 @@
...
  * @param getStyleContent - Callback that receives editor colors and returns CSS to inject.
  */
 export const useApplyEditorStyles = (
-	styleId: string,
 	getStyleContent: ( colors: EditorColors ) => string
 ): void => {
+	// Generate a unique ID once when the hook is initialized.
+	const styleIdRef = useRef< string >(
+		`editor-style-${ generateUniqueId() }`
+	);
```
amitraj2203: Added the `editor-style-` prefix since a classname cannot start with a digit.

plugins/woocommerce/client/blocks/assets/js/base/context/event-emit/reducer.ts
```diff
@@ -8,18 +13,14 @@ import {
...
-
 export const actions = {
 	addEventCallback: (
 		eventType: string,
 		callback: ActionCallbackType,
 		priority = 10
 	): ActionType => {
 		return {
-			id: generateUniqueId(),
+			id: generateUniqueIdAsString(),
```
Aljullu: Given that `generateUniqueIdAsString()` is only used here, I wonder if we could simply use `generateUniqueId()` and transform it into a string here. What do you think? This way we need on less util function.

plugins/woocommerce/client/blocks/assets/js/shared/hooks/use-theme-colors.ts
Aljullu: It makes sense! :+1:

plugins/woocommerce/client/blocks/assets/js/shared/hooks/use-theme-colors.ts
```diff
@@ -55,23 +63,45 @@ export const useApplyEditorStyles = (
...
+		const styleContent = getStyleContent( {
+			editorBackgroundColor,
+			editorColor,
+		} );
+
+		// Create a signature for this style content.
+		const styleSignature = `${ editorBackgroundColor }-${ editorColor }-${ styleContent }`;
+
+		// Check if we've already injected a style with this signature.
+		if ( injectedStyles.has( styleSignature ) ) {
```
Aljullu: I'm not sure to understand why we need the `styleSignature`. Can't we use the `styleId` to identify whether this style was already added or not? (I might be missing something, though)

Aljullu (commented):
Thanks for working on this, @amitraj2203!

Overall, it's looking well, I left one comment about getting rid of `generateUniqueIdAsString()` (I think if it's only used once we don't need to create a util for it) and a question about `styleSignature` (which I'm not sure to understand). But besides that, everything looks good. Nice clean up! :clap:

plugins/woocommerce/client/blocks/assets/js/shared/hooks/use-theme-colors.ts
amitraj2203: The `styleSignature` is needed because the hook is being called multiple times for the same component. I observed that the same style was being injected three times with different IDs, which created duplicate styles in the DOM.
When our hook gets initialized multiple times (due to re-renders), each instance creates its own unique `styleId`. Without the signature check, each of these instances would add their own style element, even though the content is identical.

Example:
image

plugins/woocommerce/client/blocks/assets/js/base/context/event-emit/reducer.ts
amitraj2203: Sounds good, I've used it that way now.

plugins/woocommerce/client/blocks/assets/js/shared/hooks/use-theme-colors.ts
Aljullu: Ooh, I see what you mean. The simple action of selecting or unselecting the blocks causes a re-render which ultimately causes the id to be different, so the `<style>` element gets added/removed every time.

In that case, I'm wondering if we should keep passing the `id` from the consumer as we did until now. Maybe, instead of passing a hardcoded string ID with the block slug, we could pass the [`blockProps.id` here](https://github.com/woocommerce/woocommerce/blob/abfc0c5d1846212c30c6fad184e4272264650f5b/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/edit.tsx#L71-L80) so its unique to that block instance?

I know I'm contradicting with my previous comment where I suggested not setting the ID from the consumer. But I didn't think of this issue with the re-renders.

The main reason I'm suggesting to go back and pass the ID from the consumer is that this way we don't need to keep the `injectedStyles` Map and I think it will simplify the implementation. What do you think? Sorry again for the back and forth! :pray:

amitraj2203: @Aljullu I tried to do it as we were doing it until now, but if we pass the `blockProps.id` directly then it doesn't work - the styles aren't getting added (maybe because it's empty at the time the hook is called).

```tsx
useApplyEditorStyles(
    blockProps.id,
    ( { editorBackgroundColor, editorColor } ) => `
        :where(.wc-block-mini-cart__badge) {
            color: ${ editorBackgroundColor };
            background-color: ${ editorColor };
        }
    `
);
```

Also, if we try to do it like this with a prefix, we still get the same issue where multiple duplicate style tags are added:

```tsx
useApplyEditorStyles(
    `mini-cart-badge-${blockProps.id}`,
    ( { editorBackgroundColor, editorColor } ) => `
        :where( .wc-block-mini-cart__badge ) {
            color: ${ editorBackgroundColor };
            background-color: ${ editorColor };
        }
    `
);
```
image

plugins/woocommerce/client/blocks/assets/js/shared/hooks/use-theme-colors.ts
```diff
@@ -1,24 +1,32 @@
...
+// Static map to track injected styles by their signature.
+const injectedStyles = new Map< string, string >();
+
 /**
  * Hook to inject a <style> element in the Site Editor using theme background and foreground colors.
  *
- * @param styleId         - ID of the style tag to create and append.
  * @param getStyleContent - Callback that receives editor colors and returns CSS to inject.
  */
 export const useApplyEditorStyles = (
```
Aljullu: I would rename this hook to match the file name: `useThemeColors()`, how does it sound?

plugins/woocommerce/client/blocks/assets/js/shared/hooks/use-theme-colors.ts
Aljullu: Thanks for investigating this, @amitraj2203, I really appreciate it! :raised_hands: 

My understanding is that the issue we have is that for some reason the editor components are being unmounted + remounted many times and that causes their `blockProps.id` to be different every time. I see what you mean about many `<style>` tags being appended. However, in the current state of the PR, what happens is that the same `<style>` tag is being removed and appended many times to the DOM. So we don't end up with multiple `<style>` tags, but we are doing unnecessary operations.

I'm not sure whether there is a perfect solution for this, to be honest, as this seems to be a limitation from Gutenberg that we need to work around. But this is how I would proceed (please let me know how it sounds):

1. Instead of using the `blockProps.id` for the id, let's go back to using hardcoded string IDs like before.
2. When the components are unmounted, let's not remove the `<style>` element. I know that means those `<style>` will be lingering until the page is reloaded. But considering we only use it in two blocks, I think it's a worth tradeoff.

Something along these lines: https://github.com/woocommerce/woocommerce/commit/d1e61c4c5bff08adeab2c0b1c396e44018ad1970. What do you think?

amitraj2203: Done! I've updated the PR with the hardcoded IDs and no style removal approach you suggested.

plugins/woocommerce/client/blocks/assets/js/shared/hooks/use-theme-colors.ts
amitraj2203: Yup sounds good, changed the hook name.

plugins/woocommerce/client/blocks/assets/js/shared/hooks/use-theme-colors.ts
```diff
@@ -55,23 +55,35 @@ export const useApplyEditorStyles = (
...
+
+		// Generate the content for this style.
+		const styleContent = getStyleContent( {
+			editorBackgroundColor,
+			editorColor,
+		} );
+
+		// Create and inject the style element with the ref ID.
+		const styleElement = document.createElement( 'style' );
+		styleElement.id = styleId;
```
Aljullu: Sorry I missed that, but we should probably add a prefix or suffix to the ID; otherwise it's going to be too generic (ie: `#mini-cart`). What about something like `#mini-cart-editor-theme-colors` or something like that?

This will need to be updated everywhere were we use `styleId` (or maybe create a new variable so we can reuse it).

Aljullu (approved):
Thanks for the updates, @amitraj2203! This is looking great:

imatge

I left one last comment, but after that, this should be ready to go!

<details>

* Theme: Twenty Twenty-Five
* Store size: small
</details>

plugins/woocommerce/client/blocks/assets/js/shared/hooks/use-theme-colors.ts
amitraj2203: Done 😄

Aljullu (approved):
Thanks for the updates! Retested, and everything looks fine. :ok_hand:

Issue #6725, opened 2014-11-10
 Reset Password Notification for user not part of site


robertdall:
This is going to be a long issue but bare with me. I just want to get it all down. 

I found an issue with user (customer) accounts being created by WooCoomerce. 

But first here is my set up. 

**WordPress 4.0 Multisite **
**WooCommerce 2.2.8**

We are currently using Quickbooks via a [Quickbooks Connector Plugin](http://61extensions.com/shop/quickbooks-integration-woocommerce/) as our fulfilment of orders (shipping receipts etc) so we have the following settings setup on our system: 

### Turned Off:
* Processing order email
* Completed order email 

### Turned On:
* Reset password
* New Account 

### Registration Options:
* Enabled: Registration on the "Checkout" page
* Enabled: Registration on the "My Account" page 
* Enabled: Display returning customer login reminder on the "Checkout" page 
* Enabled: Automatically generate username from customer email 
* Enabled: Automatically generate customer password 

### User Accounts:

If the user account has been created but then delete from the WordPress by an Administrator. The user can still reset their password and still access their account in WooCommerce. But their is no record of the user in the WP admin. 

We also use the plugin Stream which tracks users and admin changes. Stream was able to track the user signing in and signing out and reseting a password, but then doing a search for that user in the WP admin provides not results. 

So when we deleting a user from the WP Admin it isn't really delete from WooCommerce and the user can still operate as if they have an account.

I dug deeper and did a search of the wp_users database table and found that another user was still in the database as a user but not if you search for the user in the WP Admin. 

User Screenshot showing no record of user:
![User Screen showing no record of user](https://cloud.githubusercontent.com/assets/1909677/4981894/88778b18-690d-11e4-9fc6-244ddd8d4393.png)
Stream plugin showing user account reset password after deletion
![Stream showing user account reset password after deletion](https://cloud.githubusercontent.com/assets/1909677/4981895/888125ce-690d-11e4-887f-811714bfe2fc.png)
MySQL showing user that hasn't been deleted.
![MySQL database user name](https://cloud.githubusercontent.com/assets/1909677/4981961/55c5ab36-690e-11e4-8212-84e2d0b20d3e.png)

PR #57242, opened 2025-04-14
Allow Password Reset for Users that are not members of the current blog of Multisite Network

martin-neumann-gurus:
Remove the check whether user is part of current blog in a Multisite network

When a user had an account that was not part of the current blog within the Multisite network, he was not able to use the password reset link. This was extremely awkward, because obviously he could not create a new account with their email either and was unable to place an order. I found that there was actually a check in the code that prevented that, and I removed it.

1. Set up a Multisite Network, install and activate WooCommerce in 2 blogs
2. In each of the WooCommerce installations go to WooCommerce -> Settings -> Accounts and disable guest checkout, enable allow customers to create account during checkout
3. Add a product on each WooCommerce site
4. Place an order that creates an account in the first network
5. Place an order in the second network and try the password reset link in the checkout
Before the fix you should get the error message: Invalid username or email.
After applying the fix, the password reset request should pass

I thoroughly tested the proposed fix with complete step by step debugging in a local ddev environment. (Standard WordPress repository on nginx)

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 
Allowing password reset in a Multisite Network even if the user is not part of the current blog.
</details>

<details>

#### Comment 

</details>


prettyboymp:
@martin-neumann-gurus, Thanks for the PR! I noticed that the snippet being removed here was originally added to address #6725, which reported that it was an issue to have a site allow password resets for subsites they weren't already associated with.  However, it's unclear why preventing users from other subsites on the same network from recovering their password (and thus registering and checking out on another subsite) is considered a problem.  So I agree with the change you are proposing.

One thing to consider: we might want to retain the current blocking of password recovery if none of the account creation options are enabled under WooCommerce → Settings → Accounts & Privacy. I’ll need to do some further review and testing before making a final recommendation.

prettyboymp (approved):
Approving this change.

Because this behavior deviates from the standard WordPress wp-login handling it just adds confusion since a user that isn't active on the current subsite could use the password recovery directly providd by wp-login.

martin-neumann-gurus:
> @martin-neumann-gurus, Thanks for the PR! I noticed that the snippet being removed here was originally added to address #6725, which reported that it was an issue to have a site allow password resets for subsites they weren't already associated with. However, it's unclear why preventing users from other subsites on the same network from recovering their password (and thus registering and checking out on another subsite) is considered a problem. So I agree with the change you are proposing.
> 
> One thing to consider: we might want to retain the current blocking of password recovery if none of the account creation options are enabled under WooCommerce → Settings → Accounts & Privacy. I’ll need to do some further review and testing before making a final recommendation.

Yes, I was looking at the issue #6725 and it seems they had a specific scenario that they banned a user and did not want to allow him to recover the account. Just I believe the majority of users would prefer that the user is able to complete their order. Plus it was inconsistent to add the user to the site when making a login, but blocking when he makes a password reset.

It might be consistent to disallow that if account creation is not permitted at the WooCommerece settings of the site, but if so, you would need to apply the same logic to the process_login() function (Starting line 981 in class-wc-form-handler.php). There it effectively adds the user to the blog if he was registered elsewhere on another blog and allows him to complete the login process.

PR #57253, opened 2025-04-14
[Experimental] Add the correct product quantities in the Add to Cart with Options block

### LINKED ISSUE #54416: Blockified Add to Cart with Options block: make it possible to add simple products to cart
* Blocked by https://github.com/woocommerce/woocommerce/issues/53010.

Once we have all the inner blocks of the blockified Add to Cart with Options block ready, we will need to make them actually work with the Interactivity API. This issue is focused exclusively on simple products, as they require fewer inner blocks than variable and grouped products.

As an outcome, it should be possible to add the blockified _Add to Cart with Options_ block in the Single Product template (or the Single Product block when a simple product is selected) and, from the frontend, add to cart without causing a page reload.

Out of scope of this issue:

* Other product types
* Extensibility support

---

Requirements:

* handle add to cart success
* handle add to cart error
* investigate how it works when there is a Product Collection block in the same template, if the buttons aren't synced, we should create a new issue

Aljullu:

This PR makes it so the correct quantities are added to cart when using the _Add to Cart with Options_ block.

Part of #54416.
Part of WOOPLUG-2986.

0. Install and activate the WooCommerce Beta Tester plugin.
1. Go to _Tools_ > _WCA Helper_ > _Features_ and enable `experimental-blocks` and `blockified-add-to-cart`.
2. Create a post or page, add the _Single Product_ block and update the _Add to Cart with Options_ block to the blockified version.
3. In the frontend, add several items of this product to the cart. Using the Mini-Cart block, verify the correct amount was added.
4. Also verify the Add to Cart Button shows the correct number of products in the cart.
5. Add more items to the cart. Verify they are added correctly and the totals are correct.
6. Reload the page and verify the Add to Cart Button still shows the correct number of items.

https://github.com/user-attachments/assets/9432daf2-41f4-4f41-901a-bf3c45481b39

### DIFF
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
@@ -70,20 +70,15 @@ const addToCartWithOptionsStore = store(
 					{ lock: universalLock }
 				);
 
-				const { productId } = getContext< Context >();
+				const { productId, quantity, variation } =
+					getContext< Context >();
 				const product = wooState.cart?.items.find(
 					( item ) => item.id === productId
 				);
 				const currentQuantity = product?.quantity || 0;
 
-				const {
-					productId: id,
-					quantity,
-					variation,
-				} = getContext< Context >();
-
 				yield actions.addCartItem( {
-					id,
+					id: productId,
 					quantity: currentQuantity + quantity,
 					variation,
 				} );
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
@@ -10,11 +10,18 @@ type Context = {
 	productId: number;
 	variation: CartVariationItem[];
 	quantity: number;
+	tempQuantity: number;
 };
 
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
```diff
@@ -63,13 +70,23 @@ const addToCartWithOptionsStore = store(
 					{ lock: universalLock }
 				);
 
+				const { productId } = getContext< Context >();
```
kmanijak: Below in lines 79-83 you access the context again, especially getting the `productId` is duplicated. What about moving those lines here so it's consolidated?

kmanijak (approved):
Super cool work! The synchronisation feels so good because it's bi-directional and manipulating products in cart, updates the button as well. It feels smooth, I love this change and direction it goes towards! 😍 

https://github.com/user-attachments/assets/5a01f52c-d03b-4d10-a485-7121d545f511

I left a minor comment but preapproving!

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
Aljullu: Ouch. :man_facepalming:  Good catch with that! Improved it in 0a987468fc04a8d48d92882ee3cc0ca4544e384c.

PR #57256, opened 2025-04-14
Dismiss wc-admin update note when updating via CLI

### LINKED ISSUE #29211: Database Update via wp-cli and Thank You notice shown
**Prerequisites (mark completed items with an [x]):**
- [x] I have have carried out troubleshooting steps and I believe I have found a bug.

**Describe the bug**
Greetings!
This is a follow-up issue from p9o2xV-1fS-p2 around performing WooCommerce database updates via wp cli. Currently, after one performs database updates via wp cli, a Store Alert message is still shown, thanking the user for applying the updates:

database-update-done

Which I believe makes sense looking at the [code behind the note](https://github.com/woocommerce/woocommerce/blob/65443d923bfab4e8c12286204a6ce99483147942/includes/admin/notes/class-wc-notes-run-db-update.php#L261), so this is working as designed, I'm just wondering if we should consider not showing the thank you notice if the update was performed via `wp wc update`.

Some more context, and why I think this is a good idea:

- Hosts, like WordPress.com, might use wp-cli to update the database on behalf of users. The alert then seems odd to the actual end users as they did not perform a db update.
- If a site operator updates via wp-cli too, the confirmation message in the UI seems redundant because the command output itself already messages the db update was a success

```
Found 2 updates (wc_update_500_fix_product_review_count, wc_update_500_db_version)
Updating database  100% [================================================================================================================] 0:00 / 0:00
Success: 2 update functions completed. Database version is 5.0.0
```
Curious to hear other thoughts. We could likely filter off the thank you message in the use case of Atomic sites, but just wondering if we should tweak this behavior in core instead.

**Steps to reproduce the bug (We need to be able to reproduce the bug in order to fix it.)**
Steps to reproduce the bug:
1. Install Woo 4.9
2. Update to 5.0, and note the prompt to update the database
3. Execute `wp wc update`
4. Verify the update happens, and then re-load a Woo page and see the alert above.

jorgeatorres:

When performing database updates via CLI it makes no sense to keep the wc-admin notice about the update displayed, as the CLI tool itself provides completion feedback.

Closes #29211.

1. Enable the following code either as a code snippet or as a .php file inside `wp-content/mu-plugins/`:
   ```php
   <?php
   add_filter( 'woocommerce_enable_auto_update_db', '__return_false' );
   add_filter( 'action_scheduler_allow_async_request_runner', '__return_false' );
   add_action( 'init', fn () => remove_action( ActionScheduler_QueueRunner::WP_CRON_HOOK, [ ActionScheduler_QueueRunner::instance(), 'run' ] ) );
   ```
  
   This ensures that automatic database updates (which are now on by default on `trunk`) are not enabled and also that Action Scheduler doesn't run any migration routines while you're performing your testing.
1. Run the following commands to simulate the initial state for an update and trigger the update routine:
   ```shell
   wp db query "DELETE FROM $(wp db prefix)wc_admin_notes WHERE name='wc-update-db-reminder'"
   wp transient delete wc_installing
   wp option set woocommerce_db_version 9.7.0
   wp option set woocommerce_version 9.7.0
   ```
1. Go to **WC > Home** and confirm that the admin notice about the update is visible:
   Screenshot 2025-04-21 at 15 29 27
   Do not click any button or dismiss the notice.
1. Go to **Plugins** and confirm that the old-style (no wc-admin) notice also appears:
   Screenshot 2025-04-21 at 15 29 23
   Again, do not click any button.
1. Perform the update via CLI:
   ```shell
   wp wc update
   ```
1. Go back to both **WC > Home** and **Plugins**. No update admin notice should be visible.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/includes/cli/class-wc-cli-update-command.php
```diff
@@ -84,6 +84,7 @@ public static function update() {
 		$progress->finish();
 
 		WC_Admin_Notices::remove_notice( 'update', true );
+		( new WC_Notes_Run_Db_Update() )->set_notice_actioned();
```
joshuatf, line 87: How do you feel about introducing a hook for `woocommerce_db_version_updated` so we can action the notice from directly within `WC_Notes_Run_Db_Update` and keep the logic consolidated?

We can technically do this as is now with the `update_option_woocommerce_db_version` hook as well.

joshuatf (commented):
Thanks for tackling this, Jorge!  Left an alternative place we could move this logic; let me know what you think 😄

plugins/woocommerce/includes/cli/class-wc-cli-update-command.php
jorgeatorres: Thanks for the review @joshuatf! The thing is we probably still want users to see the success message once the update has been completed in those cases where the admin started it manually. Without getting "too clever" in `WC_Notes_Run_Db_Update`, I'm not sure if we can detect whether the update was completed through the CLI or not.

That said, we _could_ add a hook specifically to the CLI update command and action the notices (from within `WC_Notes_Run_Db_Update`) by hooking onto this hook. What do you think?

jorgeatorres:
Hey @joshuatf! What are your thoughts on my suggestion above? This is a tiny thing, but I'd like to wrap it up.

plugins/woocommerce/includes/cli/class-wc-cli-update-command.php
joshuatf: I see, so the notes are not initialized for the CLI command, correct?

In that case, I think a DB update hook for the CLI makes sense.  I'd just like to keep individual note logic consolidated to respective classes and avoid `plugins/woocommerce/includes/cli/class-wc-cli-update-command.php` becoming a dumping ground for a bunch of business logic.

joshuatf (commented):
Hey Jorge!  Sorry for holding you up on this.  Responded to your suggestion above.  I don't want to overcomplicate this PR so if it turns out to be too much work to introduce that hook, I'm happy to approve as is.

plugins/woocommerce/includes/cli/class-wc-cli-update-command.php
jorgeatorres: I took another look and notes and other admin logic isn't run from WP-CLI, so even if we were to fire a hook, the note itself wouldn't even notice it unless we initialize it early during the process, but then it also means adding code to the CLI command itself (so we might as well just action it). The current approach is probably what makes the most sense unless we want to refactor the whole install process (which isn't necessarily a bad idea given there's duplication between the CLI command and `WC_Install` but maybe out of scope for this issue).

Apologies for the unnecessary back and forth @joshuatf! What do you think?

jorgeatorres:
> Hey Jorge! Sorry for holding you up on this. Responded to your suggestion above. I don't want to overcomplicate this PR so if it turns out to be too much work to introduce that hook, I'm happy to approve as is.

No worries. Thanks for the review @joshuatf! It seems my suggestion was based on hope and the current approach is the most straightforward one (unless we want to make bigger changes). I left a response above, but if you still feel we need to rework this, I'd be more than happy to make the updates.

plugins/woocommerce/includes/cli/class-wc-cli-update-command.php
joshuatf: I do think ideally this would be set up so that we can fire admin code from the CLI as the current state will lead to unexpected behavior with hooks.

But for the sake of this PR, let's keep things simple 😄

joshuatf (approved):
Tested and working well.

* ✅ Admin notices show as expected when DB update is required on both the wc-admin screens and non-wc screens.
* ✅ Admin notices are removed once the update occurs from CLI

**Before**
Screenshot 2025-05-26 at 6 00 44 AM
Screenshot 2025-05-26 at 6 00 38 AM


**After**
Screenshot 2025-05-26 at 6 00 22 AM
Screenshot 2025-05-26 at 6 00 15 AM

PR #57260, opened 2025-04-14
Change pill styles to use button elements on add to cart with options block

MaggieCabrera:

This PR changes the way the pills are styled. Instead of getting the colors from the background and text colors, we use the button elements as defined by the theme.json. Doing it this way allows for better control from the theme perspective (elements can be edited on the Global Styles pattern by the user too) instead of expecting the background color and text colors are the ones we want to use always. We are also moving some of the CSS defaults to block.json, we should allow for these settings to be able to be changed by the user. 

This has a few caveats:

- The theme should define these
- There are more than one style, we should allow for themers to style all 3. I wonder if woo should provide its own set of elements to style that are shared between blocks. 
- I wasn't able to make the disabled styles work on the frontend at all.

To do: 
- a fallback for when the theme doesn't have any button element styles
- allow for the user to change the colors in case the background of the section makes the default styles unsuitable

This is how it looks on the editor with the woo starter theme

Screenshot 2025-04-15 at 18 08 37

1. Check the single product template and click on the Add to cart block
2. Make sure that it's the Add to cart with options (the sidebar will let you upgrade)
3. Click on the toolbar of the block to see the Variable product option
4. The buttons should be inheriting from the theme values

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/block.json
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/block.json
@@ -19,7 +19,29 @@
 	"$schema": "https://schemas.wp.org/trunk/block.json",
 	"supports": {
 		"inserter": false,
-		"interactivity": true
+		"interactivity": true,
+		"__experimentalStyle": {
+			"elements": {
+				"button": {
+					"spacing": {
+						"padding": {
+							"top": "0.25em",
+							"bottom": "0.25em",
+							"left": "0.75em",
+							"right": "0.75em"
+						}
+					},
+					"typography": {
+						"fontSize": "var(--wp--preset--font-size--small)"
+					},
+					"border": {
+						"width": "1px",
+						"style": "solid",
+						"radius": "2px"
+					}
+				}
+			}
+		}
 	},
 	"usesContext": [
 		"woocommerce/attributeId",
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/block.json
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/block.json
@@ -33,6 +33,10 @@
 					},
 					"typography": {
 						"fontSize": "var(--wp--preset--font-size--small)"
+					},
+					"border": {
+						"width": "1px",
+						"style": "solid"
 					}
 				}
 			}
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/block.json
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/block.json
@@ -33,11 +33,6 @@
 					},
 					"typography": {
 						"fontSize": "var(--wp--preset--font-size--small)"
-					},
-					"border": {
-						"width": "1px",
-						"style": "solid",
-						"radius": "2px"
 					}
 				}
 			}
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/edit.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/edit.tsx
@@ -45,6 +45,7 @@ function Pills( {
 						{
 							'wc-block-add-to-cart-with-options-variation-selector-attribute-options__pill--selected':
 								index === 0,
+							'is-style-outline': index !== 0,
 							'wc-block-add-to-cart-with-options-variation-selector-attribute-options__pill--disabled':
 								option.disabled,
 						}
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/edit.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/edit.tsx
@@ -15,11 +15,6 @@ import { useCustomDataContext } from '@woocommerce/shared-context';
 import type { ProductResponseAttributeItem } from '@woocommerce/types';
 import clsx from 'clsx';
 
-/**
- * Internal dependencies
- */
-import { useThemeColors } from '../../../../shared/hooks/use-theme-colors';
-
 interface Attributes {
 	className?: string;
 	style?: 'pills' | 'dropdown';
@@ -67,18 +62,6 @@ export default function AttributeOptionsEdit(
 	const blockProps = useBlockProps( {
 		className,
 	} );
-
-	// Apply selected variation pill styles based on Site Editor's background and text colors.
-	/*useThemeColors(
-		'add-to-cart-with-options-variation-selector-attribute-options',
-		( { editorBackgroundColor, editorColor } ) => `
-			:where(.wc-block-add-to-cart-with-options-variation-selector-attribute-options__pill--selected) {
-				background-color: ${ editorColor };
-				color: ${ editorBackgroundColor };
-				border-color: ${ editorColor };
-			}
-		`
-	);*/
 
 	const { data: attribute } =
 		useCustomDataContext< ProductResponseAttributeItem >( 'attribute' );
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/edit.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/edit.tsx
@@ -41,7 +41,7 @@ function Pills( {
 				<li
 					key={ option.value }
 					className={ clsx(
-						'wc-block-add-to-cart-with-options-variation-selector-attribute-options__pill',
+						'wp-element-button wc-block-add-to-cart-with-options-variation-selector-attribute-options__pill',
 						{
 							'wc-block-add-to-cart-with-options-variation-selector-attribute-options__pill--selected':
 								index === 0,
@@ -68,7 +68,7 @@ export default function AttributeOptionsEdit(
 	} );
 
 	// Apply selected variation pill styles based on Site Editor's background and text colors.
-	useThemeColors(
+	/*useThemeColors(
 		'add-to-cart-with-options-variation-selector-attribute-options',
 		( { editorBackgroundColor, editorColor } ) => `
 			:where(.wc-block-add-to-cart-with-options-variation-selector-attribute-options__pill--selected) {
@@ -77,7 +77,7 @@ export default function AttributeOptionsEdit(
 				border-color: ${ editorColor };
 			}
 		`
-	);
+	);*/
 
 	const { data: attribute } =
 		useCustomDataContext< ProductResponseAttributeItem >( 'attribute' );
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/style.scss
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/style.scss
@@ -8,16 +8,11 @@
 }
 
 .wc-block-add-to-cart-with-options-variation-selector-attribute-options__pill {
-	border-width: 1px;
-	border-style: solid;
-	border-radius: 2px;
-	padding: 0.25em 0.75em;
-	font-size: var(--wp--preset--font-size--small);
 	cursor: pointer;
 	user-select: none;
 
 	&:hover {
-		background-color: color-mix(in srgb, var(--wp--preset--color--contrast) 10%, transparent);
+		/*background-color: color-mix(in srgb, var(--wp--preset--color--contrast) 10%, transparent);*/
 	}
 
 	&:focus {
@@ -27,7 +22,7 @@
 
 	&[aria-checked="true"] {
 		&:hover {
-			background-color: color-mix(in srgb, var(--wp--preset--color--contrast) 85%, transparent);
+			/*background-color: color-mix(in srgb, var(--wp--preset--color--contrast) 85%, transparent);*/
 		}
 	}
 
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/style.scss
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/style.scss
@@ -11,19 +11,9 @@
 	cursor: pointer;
 	user-select: none;
 
-	&:hover {
-		/*background-color: color-mix(in srgb, var(--wp--preset--color--contrast) 10%, transparent);*/
-	}
-
 	&:focus {
 		outline-color: var(--wp--preset--color--accent-4);
 		outline-offset: 2px;
-	}
-
-	&[aria-checked="true"] {
-		&:hover {
-			/*background-color: color-mix(in srgb, var(--wp--preset--color--contrast) 85%, transparent);*/
-		}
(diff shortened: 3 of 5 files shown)


plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/style.scss
```diff
@@ -7,35 +7,19 @@
...
 	cursor: pointer;
 	user-select: none;
 
-	&:hover {
-		background-color: color-mix(in srgb, var(--wp--preset--color--contrast) 10%, transparent);
+	&.is-style-outline {
+		color: inherit;
+		background-color: transparent;
+		border-color: inherit;
+		backdrop-filter: none;
```
MaggieCabrera, line 18: I think I need to change this in the woo starter theme instead of here, but I'll leave it for now for testing purposes

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/style.scss
```diff
@@ -7,35 +7,19 @@
...
-	border-width: 1px;
-	border-style: solid;
-	border-radius: 2px;
-	padding: 0.25em 0.75em;
-	font-size: var(--wp--preset--font-size--small);
+.wp-element-button.wc-block-add-to-cart-with-options-variation-selector-attribute-options__pill {
 	cursor: pointer;
 	user-select: none;
 
-	&:hover {
```
MaggieCabrera, line 19: We let the hover be handled by the wp-button-element. TO DO: the hover on the outlined buttons.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorAttributeOptions.php
```diff
@@ -150,13 +150,15 @@ protected function render_pills( $attributes, $content, $block ) {
 		$pills = '';
 		foreach ( $attribute_terms as $attribute_term ) {
 			$pills .= sprintf(
-				'<div %s>%s</div>',
+				'<li %s>%s</li>',
```
MaggieCabrera: To match the markup from the editor. I'm unsure which of the two is more semantically correct, to be honest

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/edit.tsx
```diff
@@ -41,10 +36,11 @@ function Pills( {
 				<li
 					key={ option.value }
 					className={ clsx(
-						'wc-block-add-to-cart-with-options-variation-selector-attribute-options__pill',
+						'wp-element-button wc-block-add-to-cart-with-options-variation-selector-attribute-options__pill',
 						{
 							'wc-block-add-to-cart-with-options-variation-selector-attribute-options__pill--selected':
 								index === 0,
+							'is-style-outline': index !== 0,
```
MaggieCabrera, line 43: I'm unsure if we should be using this class. This is used by core as a variation of the button block, not the element. We should probably have a Woo specific class instead.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/block.json
```diff
@@ -19,7 +19,29 @@
...
+							"right": "0.75em"
+						}
+					},
+					"typography": {
+						"fontSize": "var(--wp--preset--font-size--small)"
+					},
+					"border": {
+						"width": "1px",
+						"style": "solid",
+						"radius": "2px"
```
MaggieCabrera: This will override the theme defaults, is that what we want?

MaggieCabrera:
@Aljullu this is by no means done, but I would love if you had a minute in case you think this should be scrapped altogether, since it comes with a bunch of caveats. I still think it's worth having a discussion over.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/block.json
Aljullu: I think we probably want to override the padding and typography styles, but not the border ones, at least not the border radius.

I feel big padding or font sizes are easy to break the layout of the pills. But inheriting the correct border would be nice. How does it sound?

Aljullu (commented):
Thanks for looking into this, @MaggieCabrera! Using the `wp-element-button` is something that we discussed, but didn't get to try.

About the caveats:

> The theme should define these

You mean the theme should define `wp-element-button` styles? I think it should be fine considering this is a new block, and element styles have been released for some time in Gutenberg, so is not something entirely new.

> There are more than one style, we should allow for themers to style all 3. I wonder if woo should provide its own set of elements to style that are shared between blocks.

By "all 3" you mean active, non-active, and disabled? I don't think we need to allow customizing the styles of the three separately. In my mind users should provide the colors for the non-active style and all the other styles should derive their styles from the default one.

> I wasn't able to make the disabled styles work on the frontend at all.

Right, enabling/disabling the pills based on whether the variation exists is not implemented yet. Right now all pills are always enabled.

---

Btw, if we go this route, we will also need to remove the styles which are being set in the frontend:

https://github.com/woocommerce/woocommerce/blob/ec74ba48b93c0cdec4dedfb4b52fff9a65d547d8/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts#L30-L31

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/block.json
MaggieCabrera: I copied the CSS and moved it here, but it sounds good to me if we don't add more styling than necessary

MaggieCabrera:
> You mean the theme should define `wp-element-button` styles? I think it should be fine considering this is a new block, and element styles have been released for some time in Gutenberg, so is not something entirely new.
> 

That would be really good, but we should make it clear in the documentation. Then we can extend the same implementation to other blocks/forms using buttons



> Right, enabling/disabling the pills based on whether the variation exists is not implemented yet. Right now all pills are always enabled.
> 

Good to know!

> Btw, if we go this route, we will also need to remove the styles which are being set in the frontend:
> 
> https://github.com/woocommerce/woocommerce/blob/ec74ba48b93c0cdec4dedfb4b52fff9a65d547d8/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts#L30-L31

Good call, I missed that

ResponsiveFlow:
Inspired by [Surecart's design](https://stainless-spider-ni.zipwp.site/products/single-variable-product/), I added a few lines of CSS to make the pill style more modern and coherent
| Before | After |
|--------|--------|
| Screenshot 2025-05-01 at 6 50 14 PM | Screenshot 2025-05-01 at 6 50 00 PM | 


What do you think of this design as default?


.

MaggieCabrera:
> What do you think of this design as default?
> 
> .

I think we want to have default styles to be as minimal as possible. The theme already defines button styles via the button element, we rather want to use that for consistency

ResponsiveFlow:
> default styles to be as minimal as possible

Isn't my design more minimal than the current one?
Less heavy and marked borders, more space, more coherent with the quantity selector.

MaggieCabrera:
> > default styles to be as minimal as possible
> 
> Isn't my design more minimal than the current one? Less heavy and marked borders, more space, more coherent with the quantity selector.

I don't mean minimal in design, I mean that the block shouldn't be opinionated about syles, that's the theme's prerrogative

ResponsiveFlow:
> I mean that the block shouldn't be opinionated about syles

Understood, although with the current implementation, if the merchant chooses a theme which doesn't not apply styling to the pills (very likely at least in the first period after release), the merchant will have a vert poor experience with ugly pill selectors.

MaggieCabrera:
> > I mean that the block shouldn't be opinionated about syles
> 
> Understood, although with the current implementation, if the merchant chooses a theme which doesn't not apply styling to the pills (very likely at least in the first period after release), the merchant will have a vert poor experience with ugly pill selectors.

Ideally they would be using the new woo starter theme, which will fully support this. Button elements have been supported for a long while in WordPress and most block themes support theme, and this block is meant for block themes, not classic themes.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/style.scss
MaggieCabrera: I think the outline button hover is something we can go without on a v1 of this block, we should find a way to deal with the element variation and hover styling in core

MaggieCabrera:
Screenshot 2025-05-02 at 18 26 15

I updated this PR after rebasing with your latest changes @Aljullu. Because we don't really have access to the theme colors, and we are just applying them via classes, this solution is not as flexible as what the block was doing before in terms of dealing with the hover colors, but it does bring a more consistent experience with what core does. The woo starter theme right now needs some tweaking for the buttons, so if you want to test this I suggest you change the values ont heme.json for border radius and such.

Do you think this is something we should move ahead with?

ResponsiveFlow:
> new woo starter theme


Any indication regarding its availability?

MaggieCabrera:
> > new woo starter theme
> 
> Any indication regarding its availability?

I'm not sure of the official date, but soon!

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/VariationSelectorAttributeOptions.php
```diff
@@ -144,18 +144,20 @@ protected function render_pills( $attributes, $content, $block ) {
...
-						'data-wp-watch'               => 'callbacks.watchSelected',
-						'data-wp-on--click'           => 'actions.toggleSelected',
-						'data-wp-on--keydown'         => 'actions.handleKeyDown',
-						'data-wp-context'             => array(
+						'role'                                               => 'radio',
+						'class'                                              => 'wp-element-button wc-block-add-to-cart-with-options-variation-selector-attribute-options__pill is-style-outline',
+						'data-wp-bind--tabindex'                              => 'state.pillTabIndex',
+						'data-wp-bind--aria-checked'                          => 'state.isPillSelected',
+						'data-wp-bind--aria-disabled'                         => 'state.isPillDisabled',
+						'data-wp-class--wc-block-add-to-cart-with-options-variation-selector-attribute-options__pill--selected' => 'state.isPillSelected',
```
Aljullu, line 155: If we are settings this class for selected pills, maybe we should do something similar for disabled ones? This way in the CSS file we only use classes instead of `[aria]` selectors.

Btw, since https://github.com/woocommerce/woocommerce/pull/57537, this can be tested. :slightly_smiling_face:

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/edit.tsx
Aljullu: Right, that's a good concern. If we ship it like that, it's likely that themes will start relying on the pills having the `is-style-outline` class. I don't think that's an issue because if WC core were to change that, it would be a breaking change in core styles too.

But you have more context, so I'm fine with whatever you think is better. :slightly_smiling_face:

Aljullu (commented):
> I updated this PR after rebasing with your latest changes @Aljullu. Because we don't really have access to the theme colors, and we are just applying them via classes, this solution is not as flexible as what the block was doing before in terms of dealing with the hover colors, but it does bring a more consistent experience with what core does.

I wonder if for the hover styles we could play a bit with the opacity? I know that brings a11y challenges as it might affect contrast, but not having any hover style is also kind of an a11y limitation, so I don't think there is a perfect solution here.

Another option would be to add an `outline` like the focused state. But I feel it looks a bit too intrusive. :sweat_smile: 

> Do you think this is something we should move ahead with?

It works great, and code changes also look good to me. I'm just a bit worried about how "ok-ish" this approach is from a WordPress core perspective. I understand that using `wp-element-button` classes in extension elements is supported, but we are also overriding some of their styles, using the `is-style-outline` class, etc. Which I worry might be pushing the boundaries a bit.

We need to be extra-cautious because once we ship it with certain class names, it's likely themes will start using them, so we need to be mindful to ship it only if we think we won't need to change the approach in the future. As I mentioned in my previous comment, you probably have more context than me on how "safe" this approach is, so I would like to know what your thoughts are. :pray:

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/VariationSelectorAttributeOptions.php
MaggieCabrera: Yes, I rebased this PR and it works nicely on the frontend

MaggieCabrera: I will look into the disabled styles

MaggieCabrera:
> I wonder if for the hover styles we could play a bit with the opacity? I know that brings a11y challenges as it might affect contrast, but not having any hover style is also kind of an a11y limitation, so I don't think there is a perfect solution here.
> 

So we do have hover styles if the theme provides them for the regular button style (not the outline one). Woo starter theme doesn't YET, but we can fix that on the theme. I'm unsure if we should provide any styles if the theme doesn't provide hover styles. My instinct says no, but this is a strong opinion loosely held.


> 
> It works great, and code changes also look good to me. I'm just a bit worried about how "ok-ish" this approach is from a WordPress core perspective. I understand that using `wp-element-button` classes in extension elements is supported, but we are also overriding some of their styles, using the `is-style-outline` class, etc. Which I worry might be pushing the boundaries a bit.
> 

So, `is-style-outline` is already doing this for the core button block (the overriding, that is). My only concern is that the outline class is a class especific to the button block, not the button element, which means this implementation is not entirely correct. Ideally we would be able to have "block styles" for elements, but we don't, and the alternative is to do some CSS manipulation that has other pitfalls. I would like to hear any ideas from @mikachan or @scruffian but I will keep thinking about this.


> We need to be extra-cautious because once we ship it with certain class names, it's likely themes will start using them, so we need to be mindful to ship it only if we think we won't need to change the approach in the future. As I mentioned in my previous comment, you probably have more context than me on how "safe" this approach is, so I would like to know what your thoughts are. 🙏

The outline class is already in core and will keep having support in the future, which is why I chose it over a woo specific one, but my previous stated concerns still remain

Aljullu:
It makes sense to me, thanks @MaggieCabrera! Let's see if @mikachan or @scruffian have any strong thoughts, otherwise we can go ahead with the current approach. :slightly_smiling_face:

scruffian:
I think using `is-style-outline` is reasonable. I can anticipate that being added to core one day, but in the meantime, we can add whatever CSS we need to to make it work.

I think this approach is good and seems to be well aligned with core.

scruffian:
> I think we want to have default styles to be as minimal as possible.

I agree with this. Even though the theme styles should override the block styles, there could be cases where themes keep their rules pretty sparse and in these cases its better if the block is limited in what CSS it provides, so that these elements match the other buttons in the theme.

mikachan:
>  Ideally we would be able to have "block styles" for elements, but we don't, and the alternative is to do some CSS manipulation that has other pitfalls.

I agree, I think using `is-style-outline` is a fine workaround in this case and is aligned with Core.

nerrad:
Closing because this is a stale draft issue

*\*left by Biff (AI agent) on behalf of Darren*

Issue #56675, opened 2025-03-25
Allow to identify `POS` orders from the order detail page


albarin:

BurakParsAydin:
@albarin Just wanted to check in - if you or anyone else isn’t already planning to work on this issue, I’d be happy to take it on.

BurakParsAydin:
Hi @albarin FYI, I went ahead and opened a PR to address this task. Of course, happy to make any adjustments if needed - just let me know!

albarin:
> [@albarin](https://github.com/albarin) Just wanted to check in - if you or anyone else isn’t already planning to work on this issue, I’d be happy to take it on.

I think we are not planning to work on it yet, because we still need to decide how to identify the `POS` orders visually.
Adding @malinajirka, who may have some more info about it.

malinajirka:
Thanks for the ping! @jaclync will be taking over this effort again. Could you please follow up here @jaclync?

jaclync:
Hi @albarin and @BurakParsAydin, as Alba mentioned earlier we're still finalizing the UI (ref p1745375279135859/1742894475.967679-slack-C070SJRA8DP). We will post an update when the UI is confirmed.

From taking a look at your PR [https://github.com/woocommerce/woocommerce/pull/57432](https://github.com/woocommerce/woocommerce/pull/57432), it seems to be using the source type metadata instead of the `created_via` attribute that we're using to identify POS orders, similar to the addition of a filter woocommerce/woocommerce#57273.

BurakParsAydin:
@jaclync  Thanks for the update!

Just a quick note - since I'm contributing externally, I don’t have access to the Slack channel you referenced. If there's any chance you could share updates here as well, that would be super helpful.

Issue #56667, opened 2025-03-25
Identifying POS Orders


albarin:
Enable merchants to easily view and distinguish POS orders in the order list and order details.

For more context, see: pdfdoF-6xX-p2

samrat-biswas-git:
It looks like this has already been fixed, or it's no longer happening on trunk.

Here's what I found: Trunk now contains POS order attribution utilities and admin order-surface handling, including order attribution meta boxes and list-table support. The repo already has current POS order identification paths that cover the request.

I'm going to close this one. If you're still facing this issue on a current version, please reply with steps to reproduce and we'll happily take another look.

Fix/changelog reference: Merged PR https://github.com/woocommerce/woocommerce/pull/57432 / commit https://github.com/woocommerce/woocommerce/commit/091d0ac968

PR #57273, opened 2025-04-15
Add `Filter by sales channel` to the admin Orders page

### LINKED ISSUE #56673: Add filter by `created_via` to the `wp-admin/admin.php?page=wc-orders` page

albarin:

Closes #56673 

| Before | After |
| ------ | ----- |
| Screenshot 2025-04-15 at 15 43 00 | Screenshot 2025-04-17 at 13 36 11 |

1. Go to `WooCommerce > Settings > Advanced > Features` (`/wp-admin/admin.php?page=wc-settings&tab=advanced&section=features`) and ensure you have selected `WordPress posts storage (legacy)` in `Order data storage`.
2. Go to WooCommerce > Orders (`/wp-admin/admin.php?page=wc-orders`)
3. Check for the new `Filter by sales channel` dropdown.
4. Select one of the values, click `Filter`, and check that the orders are filtered by it. Test with different options.
5. Go to `WooCommerce > Settings > Advanced > Features` (`/wp-admin/admin.php?page=wc-settings&tab=advanced&section=features`) and ensure you have selected `High-performance order storage (recommended)` in `Order data storage`. Repeat steps 2 to 4.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Add "Filter by sales channel" to the admin Orders page.
</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
```diff
@@ -565,6 +567,47 @@ private function set_search_args(): void {
...
+	/**
+	 * Render the created_via filter dropdown.
+	 *
+	 * @return void
+	 */
+	public function created_via_filter() {
+		// phpcs:disable WordPress.Security.NonceVerification.Recommended
+		$current_created_via = isset( $_GET['_created_via'] ) ? sanitize_text_field( wp_unslash( $_GET['_created_via'] ) ) : '';
+
+		$created_via_options = array(
```
albarin, line 595: @malinajirka @joshuatf Should we have those two options for now? Should we make the `created_via_options` filterable? Any suggestions? 🤔 TY

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
```diff
@@ -565,6 +567,47 @@ private function set_search_args(): void {
...
+			''             => __( 'Filter by sales channel', 'woocommerce' ),
+			'store-api'    => __( 'Site Checkout', 'woocommerce' ),
```
malinajirka: ❓ ⚠️  Is this intentionally set to `store-api` or should it be `rest-api`? Right now, the 'store checkout' filter doesn't display any orders.

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
malinajirka: I think the two options might be enough for now. However, this makes me wonder whether `Site checkout` label isn't a bit misleading since it'll display even orders created via wp-admin 🤔 . Wdyt?

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
albarin: > ❓ ⚠️ Is this intentionally set to store-api or should it be rest-api? Right now, the 'store checkout' filter doesn't display any orders.

Is it possible that your store does not have orders created using the normal shop checkout? 🤔 

According to my tests, these are the possible values for `created_via` depending on how the order is made (might be more):

- store-api - Created by the customer in a normal purchase using checkout in the frontend.
- admin - Created manually in the orders admin page.
- rest-api - Created by making a request to `POST /orders`.
- pos-rest-api - Created with a POS request (once #57225 is merged)

@malinajirka @joshuatf Not sure if we should add all of these? Or what should we use as a label for the `rest-api` ones, since there could be multiple sources for them (mobile app, another plugin or custom integration, etc.).

| Value | Label |
|--------|--------|
| `store-api` | Site Checkout |
| `admin` | Admin ? |
| `rest-api` | ? |
| `pos-rest-api` | Point of Sale |

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
albarin: I think orders manually created in wp-admin have `created_via`=`admin`, see my other [comment](https://github.com/woocommerce/woocommerce/pull/57273/files#r2046542171) 🤔

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
malinajirka: > Is it possible that your store does not have orders created using the normal shop checkout? 🤔

Nope, there are plenty orders. I re-tested it after creating a fresh order and it still displays an empty list.
Screenshot 2025-04-16 at 13 19 03

> @malinajirka @joshuatf Not sure if we should add all of these? Or what should we use as a label for the rest-api ones, since there could be multiple sources for them (mobile app, another plugin or custom integration, etc.).

Perhaps we could use something generic like “External Integration” or “Connected via API”? I realize it's not the most user friendly, but at the same time not that confusing either. Wdyt?

malinajirka: I checked the value of the attribute for the newly created order and it's: `"created_via": "checkout"`.

albarin: @malinajirka @joshuatf Okay, just realized that `checkout` is set if you use the old checkout with shortcodes, and `store-api` is used for the blockified version of the checkout. How can we differentiate between those two in the label? 🤔

joshuatf: Ah, that's interesting. `store-api` is probably not very useful (or recognizable) information for merchants.  Assuming this information is helpful for merchants, we could use:
* `Legacy Checkout`
* `Blocks Checkout`

malinajirka: I can't think of a use-case in which a merchant would want to differentiate between these two. What about calling it `Site Checkout` and map both attributes to it?

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
```diff
@@ -565,6 +567,49 @@ private function set_search_args(): void {
...
+		$created_via_options = array(
+			''             => __( 'Filter by sales channel', 'woocommerce' ),
+			'admin'        => __( 'Admin', 'woocommerce' ),
+			'checkout'     => __( 'Legacy Checkout', 'woocommerce' ),
+			'store-api'    => __( 'Blocks Checkout', 'woocommerce' ),
+			'pos-rest-api' => __( 'Point of Sale', 'woocommerce' ),
+		);
+		?>
+
+		<select name="_created_via" id="filter-by-created-via">
```
malinajirka, line 603: 💡I noticed we have two types of filters at the moment:
1. Filters with default value - All dates/All order types -> the default text is black.
2. Filters with no default value - Filter by registered customers -> the default text is greyed out.

Screenshot 2025-04-17 at 8 19 25

Wdyt, should we change the color of the text to grey or perhaps re-name the filter to `All sales channels`?

malinajirka (approved):
Thanks @albarin! I've tested the behavior and it works as expected. I've left one idea, but other than that it looks good to me. Having said that, it should still be reviewed by someone with more PHP experience.

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
albarin: Good idea, just pushed the change!

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
albarin: Makes sense, pushed the change! Thanks!

albarin:
@malinajirka @joshuatf After some discussions in this PR, the filter options are:
```php
$created_via_options = array(
	''                   => __( 'All sales channels', 'woocommerce' ),
	'admin'              => __( 'Admin', 'woocommerce' ),
	'checkout,store-api' => __( 'Checkout', 'woocommerce' ),
	'pos-rest-api'       => __( 'Point of Sale', 'woocommerce' ),
);
```

Also, I realised it was only working when the `High-performance order storage` option was enabled. I've pushed a change to make it work always, and I've updated the testing steps.

Thanks for your help! 🙌

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
```diff
@@ -565,6 +567,48 @@ private function set_search_args(): void {
...
+	 */
+	private function set_created_via_args(): void {
+		// phpcs:disable WordPress.Security.NonceVerification.Recommended
+		$created_via = sanitize_text_field( wp_unslash( $_GET['_created_via'] ?? '' ) );
+
+		if ( empty( $created_via ) ) {
+			return;
+		}
+
+		$this->order_query_args['created_via'] = explode( ',', $created_via );
```
joshuatf: Might be a little bit cautious, but we could map over these and trim just to be safe:

```suggestion
		$this->order_query_args['created_via'] = array_map( 'trim', explode( ',', $created_via ) );
```

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
```diff
@@ -565,6 +567,48 @@ private function set_search_args(): void {
...
+	/**
+	 * Render the created_via filter dropdown.
+	 *
+	 * @return void
+	 */
+	public function created_via_filter() {
+		// phpcs:disable WordPress.Security.NonceVerification.Recommended
+		$current_created_via = isset( $_GET['_created_via'] ) ? sanitize_text_field( wp_unslash( $_GET['_created_via'] ) ) : '';
+
+		$created_via_options = array(
```
joshuatf, line 595: I was also wondering about a filter here, but I'm good holding off on this until it's requested.

If we do create a POS class at some point, it might make sense to move these options into a getter in that class, but we can old off on this if that class/feature is not set up yet.

plugins/woocommerce/tests/php/src/Internal/Admin/Orders/ListTableTest.php
```diff
@@ -180,4 +180,56 @@ private function get_months_count( \DateTime $start, \DateTime $end ): int {
 
 		return $months_from_years - $start_month_diff + $end_month;
 	}
+
+	/**
+	 * @testdox When filtering by created_via, only orders with that specific value should be shown.
+	 */
+	public function test_filtering_by_created_via_shows_only_matching_orders() {
```
joshuatf, line 187: Thanks for the tests ❤️

joshuatf (approved):
Tested with HPOS enabled and disabled and the filter works as expected!

Screenshot 2025-04-17 at 10 55 52 AM
Screenshot 2025-04-17 at 10 50 39 AM

Left one optional comment and a potential future update once we have a POS class enabled.

Is the plan to still make this a toggleable feature?  If so, we may need to move some of this logic or wrap in conditions once the feature is set up.

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
albarin: Yeah, I asked about that in a previous [comment](https://github.com/woocommerce/woocommerce/pull/57273#discussion_r2044590480). But I think you are right, we can hold off on this a bit 👍

plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php
albarin: Good idea, just pushed the change.

albarin:
> Is the plan to still make this a toggleable feature? If so, we may need to move some of this logic or wrap in conditions once the feature is set up.

@joshuatf From this comment p1744642687927519/1742894475.967679-slack-C070SJRA8DP I understood we don't need the feature toggle for this issue or any of the others under https://github.com/woocommerce/woocommerce/issues/56667, is that right @malinajirka? or did I misunderstand it? 🤔

malinajirka:
I believe the filter is useful even for non-POS merchants and the fact one of the values is `POS` imho doesn't justify hiding it for all non-POS users, therefore, I'd personally vote for showing the filter at all times.

Regarding other tasks in the list. The toggle is relevant only for UI tasks:
- [Add badge next to order status indicating the order is from POS](https://github.com/woocommerce/woocommerce/issues/56674) => this one will be just a badge, so there is nothing to hide for non-POS orders. If we decide to show some kind of non-POS-order badge, than I'd say it should be hidden behind the toggle.
- [Allow to identify POS orders from the order detail page](https://github.com/woocommerce/woocommerce/issues/56675) => similar to the above, if we decide to show something on all orders, I'd say we might want to hide it for non-POS orders, depends on the final solution.

Having said all that, @jaclync will be taking over this effort again, so I'll defer to her for the final decision.

jaclync:
Moved the discussion to p1745374585710389/1742894475.967679-slack-C070SJRA8DP for better collaboration with different teams.

PR #57274, opened 2025-04-15
Add to cart variations: Don't break word

scruffian:

This prevents words in the variations table from "breaking". The variations table is often loaded inside a columns block, which applies `word-break: break-word;` to its contents.

Ideally these variations would be in divs that reflow rather than a table, but that would be a very large change potentially breaking many stores. I think this small CSS fix is more pragmatic.

Closes 156811-ghe-Automattic/wpcom-themes

| Before | After |
| ------ | ----- |
| Screenshot 2025-04-15 at 12 12 52|Screenshot 2025-04-15 at 12 12 44|

1. Open a product with variations
2. Resize the browser window to a mobile view
3. Confirm that the text doesn't break onto multiple lines

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

<summary>Ensure product variation names don't get broken onto multiple lines.</summary>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/fix-variations-word-break
+++ b/plugins/woocommerce/changelog/fix-variations-word-break
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Keep variations labels on one line
--- a/plugins/woocommerce/client/legacy/css/woocommerce-blocktheme.scss
+++ b/plugins/woocommerce/client/legacy/css/woocommerce-blocktheme.scss
@@ -120,6 +120,12 @@
 				margin-bottom: 0;
 				vertical-align: middle;
 			}
+
+			.variations {
+				td, th {
+					word-break: normal;
+				}
+			}
 		}
 
 		.related.products {
--- a/plugins/woocommerce/client/legacy/css/woocommerce.scss
+++ b/plugins/woocommerce/client/legacy/css/woocommerce.scss
@@ -404,6 +404,7 @@ p.demo_store,
 					border: 0;
 					line-height: 2em;
 					vertical-align: top;
+					word-break: normal;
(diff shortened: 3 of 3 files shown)


MaggieCabrera:
This sounds reasonable to me

Aljullu (approved):
LGTM, thanks for this fix, @scruffian!

Getting the same results on my end, with TT5:

Before | After
--- | ---
imatge | imatge


<details>

* Theme: Twenty Twenty-Five
* Store size: small
</details>

---

Btw, to get the automations pass, you will need to add a changelog file by running: `pnpm --filter='@woocommerce/plugin-woocommerce' changelog add`.

Aljullu:
@scruffian as I was thinking about this again... Given that this issue is specific to block themes, I think it would be better if the new CSS declaration was in [woocommerce-blocktheme.scss](https://github.com/woocommerce/woocommerce/blob/e6c606a4edefa819b013511b8a7fb1bc16e0df2e/plugins/woocommerce/client/legacy/css/woocommerce-blocktheme.scss), which is a CSS file only loaded in block themes. What do you think?

scruffian:
> Given that this issue is specific to block themes

That makes sense. Should we move the other CSS which targets these selectors over as well (there's more in the same file)?

Aljullu:
> That makes sense. Should we move the other CSS which targets these selectors over as well (there's more in the same file)?

I don't think so, those elements are present in classic and block themes.

But my understanding is that the issue is only reproducible in block themes because we inherit `word-break` from `.wp-block-column`, so the fix is only required in block themes.

PR #57280, opened 2025-04-15
Add email block renderer for Quote core block

yuliyan:

* This PR adds an email block renderer for the Quote core block and adds it to the allowed blocks in the email editor.

Closes [WOOPLUG-3864](https://linear.app/a8c/issue/WOOPLUG-3864/core-blocks-quote).

| Screenshot |
| ------ |
|  sHVMpCmlKhxRXfXN  |

1. Enable the block-based email editor from WooCommerce → Settings → Advanced → Features → **Block Email Editor (alpha)**
2. Open an email in the editor WooCommerce → Settings → **Emails**
3. Confirm the Quote block is present in the blocks list.
4. Add a Quote block to the email content and apply some customizations from the sidebar.
5. Send a test email from the Preview dropdown and confirm the applied styling for the Quote block is displayed correctly in the email.
6. Confirm integration tests for the Quote block pass.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

<summary>Add an email block renderer for the Quote core block in the email block-based editor.</summary>

#### Comment 

</details>


lysyjan (changes requested):
Hey @yuliyan 👋 
I tested your PR and I found a couple of issues:
1. There is a missing line on the left side of the block. It's missing in the rendered HTML and also in the email editor. You can check how it should look like in the post editor. Here is an example in the second image below:
Screenshot 2025-04-17 at 16 59 25

---

Screenshot 2025-04-17 at 17 00 10

2. The citation is not rendered in the email:
Screenshot 2025-04-17 at 17 03 32

3. Alignment for the quote block does not work in the template mode:
Screenshot 2025-04-17 at 17 06 01

4. The selected alignment is not reflected in the rendered email.

5. There is a redundant margin in the email editor:
Screenshot 2025-04-17 at 17 11 35


Feel free to ping me if you need help addressing these issues or getting more details. I haven't tested the block across different email clients yet.

yuliyan:
Thank you for the detailed feedback, @lysyjan!

1. The border and padding styles for the Quote block in the post editor appear to be set in the theme. However, we don't seem to load the theme styles in the Email editor and thus if not set as an attribute, the border is missing in the preview and the rendered email markup. If we want to have a border by default, I suppose we can override the default values for the border attribute, but I don't think it'd be a good approach. What do you think?

2. I've added the citation content to the renderer. One thing that concerns me is that I'm not escaping the HTML, since the citation supports the typography controls (bold, italic, highlights, etc.) and the tags need to be preserved. I noticed we're handling it more or less the same way in other places (like the paragraph block, for example). Do you have any suggestions here?

3. I decided to remove the alignment controls in this case, since it seemed like this type of alignment (float) doesn't make much sense in the email markup context and wouldn't work as expected. Also, none of the other blocks seem to support this type of alignment. Let me know if I'm missing something.

4. Since I've removed the alignment controls for the Quote block, there's no need to add alignment support in the renderer. However, I noticed that the renderer was also missing logic to handle the Quote block's text alignment attribute, so I added that in. That said, there's another issue: while the citation aligns correctly, some child blocks, like paragraphs, set their own `text-align` attributes, which can override the alignment set by the Quote block. Do you think we should disable the text align controls as well, or see if it's possible to pass the text alignment value from the Quote block down to the children?

5. I've reset the inline margins for the `blockquote` element in the Email editor styles.

lysyjan:
> The border and padding styles for the Quote block in the post editor appear to be set in the theme. However, we don't seem to load the theme styles in the Email editor and thus if not set as an attribute, the border is missing in the preview and the rendered email markup. If we want to have a border by default, I suppose we can override the default values for the border attribute, but I don't think it'd be a good approach. What do you think?

@yuliyan I was investigating a possible approach with styles here, and I would extend [content-editor](https://github.com/woocommerce/woocommerce/blob/88f1c3c00c9bfc7c51f0b250623c4be761381d52/packages/php/email-editor/src/Engine/content-editor.css#L69-L91) styles. There are already some styles for the image block. In my opinion, the quote block renderer should also handle those styles for the correct look in emails.

I know it can be overwhelming, but we want to ensure the look in the editor and all supported email clients is almost identical.

lysyjan:
@yuliyan, I was testing how text alignment works in the post editor, and I would recommend the same. For example, when the text alignment is set to right, the default border is changed to right and when it is centered, there is no border.

I was also thinking about a possible conflict with the child blocks, and I would create a preprocessor for that. You can check how similar classes look in the PHP package and `Automattic\WooCommerce\EmailEditor\Engine\Renderer\ContentRenderer` namespace

yuliyan:
@lysyjan, thank you for the suggestions. I've [added the default border styles in the email editor](https://github.com/woocommerce/woocommerce/pull/57280/commits/8c153a52f6a8d2a89c25108d325365f0144b5e82) and the renderer and [created a new preprocessor to handle the text alignment](https://github.com/woocommerce/woocommerce/pull/57280/commits/a29ea8de129454bd9ad6ac9532958773fd7d428c) for the nested blocks. Could you take a look?

yuliyan:
@lysyjan, I've addressed the items that we discussed:

* Moved the default border styles to `Integration/Core/theme.json`.
* Created new post processor to handle properly setting the `border-style` declaration only where needed to properly render the expected borders across email clients.
* Removed the style variations for the Quote block (Plain, default) like we do for other blocks.
* Use `<blockquote>` for the wrapper, and `<cite>` for the citation content.

Could you take another look when you get the chance?

packages/php/email-editor/src/Integrations/Core/theme.json
```diff
@@ -1,27 +1,62 @@
...
+			},
+			"core/quote": {
+				"border": {
+					"style": "solid",
+					"width": "0 0 0 1px",
+					"color": "currentColor"
+				},
+				"spacing": {
+					"margin": {
+						"top": "0",
```
lysyjan: We don't want to reset the margin top and bottom because there is a default block spacing. This spacing can also be set in the layout settings. You can notice that in the current email editor, there is no vertical space between quote blocks, but when the email is rendered, there are spaces, which is correct.

packages/php/email-editor/tests/integration/Integrations/Core/Renderer/Blocks/Quote_Test.php
```diff
@@ -0,0 +1,153 @@
...
+					'top'    => array(
+						'color' => '#444444',
+						'width' => '4px',
+					),
+					'radius' => array(
+						'bottomLeft'  => '5px',
+						'bottomRight' => '10px',
+						'topLeft'     => '15px',
+						'topRight'    => '20px',
+					),
```
lysyjan, line 129: I am not sure about this test block configuration because the quote block does not allow configuration for separated border parts to be set up. It is always set for the whole block.

lysyjan: Also, when I tested your changes, I had some trouble with the border color. I think the reason is that the border color is stored at another level. From what I checked, the color is stored under the key `borderColor`.

packages/php/email-editor/src/Integrations/Core/Renderer/Blocks/class-quote.php
```diff
@@ -0,0 +1,134 @@
...
+		// Padding properties need to be added to the table cell.
+		$cell_styles = $this->get_styles_from_block(
+			array(
+				'spacing' => array( 'padding' => $block_attributes['style']['spacing']['padding'] ?? array() ),
+			)
+		)['declarations'];
+
+		return sprintf(
+			'<!--[if mso | IE]><table align="left" role="presentation" border="0" cellpadding="0" cellspacing="0" width="100%%" style="%1$s"><tr><td style="%2$s" width="100%%"><![endif]-->
+			<blockquote class="email-block-quote %3$s" style="%1$s %2$s">
```
lysyjan: Thank you for adding the blockquote tag here. Unfortunately, when I tested the latest changes, I had some issues with the Gmail mobile app on EmailOnAcid. It looks like styling this HTML tag is not possible, and the left and right borders are missing. I am afraid we need to avoid using this tag because of the Gmail app.
Screenshot 2025-05-06 at 18 51 56

packages/php/email-editor/src/Engine/Renderer/ContentRenderer/Postprocessors/class-border-style-postprocessor.php
```diff
@@ -0,0 +1,236 @@
...
+ * This file is part of the WooCommerce Email Editor package
+ *
+ * @package Automattic\WooCommerce\EmailEditor
+ */
+
+declare(strict_types = 1);
+namespace Automattic\WooCommerce\EmailEditor\Engine\Renderer\ContentRenderer\Postprocessors;
+
+/**
+ * Ensures every border with border-width > 0 has a border-style, defaulting to 'solid' if missing, and removes extra border-style properties.
```
lysyjan: Good job with this class. It is pretty complex. Please can you extend the class description about the purpose and add some examples?

lysyjan (commented):
The rendering is improved compared with the previous version. However, I still found some issues with EmailOnAcid. For example, Outlook clients are not well supported.

Here is one example:
Editor:
Screenshot 2025-05-06 at 18 52 13

Outlook:
Screenshot 2025-05-06 at 18 51 37

I also noticed a problem with the spacing between blocks that is mentioned in another comment.

packages/php/email-editor/src/Integrations/Core/theme.json
yuliyan: This was an unintentional change. I've removed the top and bottom margin resets.

packages/php/email-editor/src/Engine/Renderer/ContentRenderer/Postprocessors/class-border-style-postprocessor.php
yuliyan: Added a more detailed description about the issue that this post processor resolves.

packages/php/email-editor/tests/integration/Integrations/Core/Renderer/Blocks/Quote_Test.php
yuliyan: I might be misunderstanding your comment, but from what I've seen both border configuration formats are supported. When a uniform border style is used, the block editor generates attributes in this format:

```php
'border' => array(
	'width' => ...,
	'style' => ...,
	// etc.
)
```

For mixed border styles, the attributes are parsed into this structure:

```php
'border' => array(
	'bottom' => array(
		'color' => ...,
		'width' => ...,
		'style' => ...,
	),
	'left' => ...,
	'right' => ...,
	'top' => ...,
)
```

As for the border color - you are correct. In the case of having uniform border styles, it is being stored as a top level attribute with `borderColor` key, and when having mixed styles, it's stored individually as a child of the respective border side. I had it handled at some point, but it seems like I forgot to transfer it after switching to the post processor approach.

I've now re-added it and updated the unit test to cover the case.

packages/php/email-editor/src/Integrations/Core/Renderer/Blocks/class-quote.php
yuliyan: Apologies - I didn't realize it was broken. I updated the email source code using the `<blockquote>` tag on Email On Acid, but I was apparently looking at outdated screenshots and thought everything was working fine.

It turns out the `<blockquote>` tag also caused the border style issue you mentioned in your other comment. Reverting back to using tables seems to have resolved all of these issues.

Do you think we should also replace the `<cite>` tag with a table? From my testing, the `<cite>` tag appears to render well across email clients, but I wanted to double check.

lysyjan (approved):
Thank you for all your effort and patience.
The new quote block works nice now.

I tested it on my local with the latest WP 6.8.
I also used EmailOnAcid to check the output for different email clients. Here are some screenshots:
Screenshot 2025-05-07 at 17 47 57
Screenshot 2025-05-07 at 17 47 39
Screenshot 2025-05-07 at 17 47 04
Screenshot 2025-05-07 at 17 46 17

packages/php/email-editor/src/Integrations/Core/Renderer/Blocks/class-quote.php
lysyjan: I think the `<cite>` tag is fine. I checked, and there are no default styles for that.

PR #57315, opened 2025-04-16
Use iAPI helper to define `data-wp context` where possible

SantosGuillamot:

I'm using the `wp_interactivity_data_wp_context` helper to define this directive instead of `wp_json_encode`.

The modified blocks should have the `data-wp-context` attribute and keep working as expected:
- Add to cart with options.
- Add to cart with options variation selector.
- Add to cart button.
- Active filters.
- Attribute filter.
- Price filter.
- Rating filter.
- Product filters.
- Status filter.
- Product template.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
```diff
@@ -237,8 +235,9 @@ protected function render( $attributes, $content, $block ) {
 			remove_filter( 'render_block_context', array( $this, 'set_is_descendant_of_add_to_cart_with_options_context' ) );
 
 			$form_html = sprintf(
-				'<form %1$s>%2$s%3$s%4$s</form>',
```
dinhtungdu, line 283: To me, I still prefer defining the context data as the wrapper attributes for readability. Using `wp_interactivity_data_wp_context` this way works, but it makes this hard to read LOC harder to read, which in a way, goes against the purpose of refactoring, imho. Curious to see what others think :D

luisherranz: Perhaps we could take this opportunity to introduce the additional helper that doesn't include the attribute name and see how that one feels.

luisherranz (commented):
One thing that is clearly seen in this PR is that there isn't much consistency in how templating is done on Woo blocks.

Some use `sprintf`, others use `strtr`. Some include the attributes in `get_block_wrapper_attributes`, others add them directly to the HTML. And others even include [their own functions](https://github.com/woocommerce/woocommerce/blob/d0e2c0bc7badced522e6b0f4d9d271808b45fd38/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorAttributeOptions.php#L100-L119) to manage the attributes…

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
dinhtungdu: > the additional helper that doesn't include the attribute name

Yes, I was thinking about extracting the current help into a new one such as `wp_interactivity_prepare_data_wp_context` and refactor the `wp_interactivity_data_wp_context` using the new helper.

dinhtungdu:
There may be some special needs for Add to Cart with Options, but I agree we should standardize the template for Woo blocks.

dinhtungdu:
Created [WOOPLUG-4028](https://linear.app/a8c/issue/WOOPLUG-4028/review-and-standardize-woocommerce-blocks-templating) for tracking the work on this.

luisherranz:
> There may be some special needs for Add to Cart with Options

Hmm... interesting. May I know what the special needs are that make that block require [this helper](https://github.com/woocommerce/woocommerce/blob/d0e2c0bc7badced522e6b0f4d9d271808b45fd38/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorAttributeOptions.php#L100-L119), while the rest of the blocks don't? It seems like a fairly generic helper to me…

dinhtungdu:
> May I know what the special needs are that make that block require [this helper](https://github.com/woocommerce/woocommerce/blob/d0e2c0bc7badced522e6b0f4d9d271808b45fd38/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorAttributeOptions.php#L100-L119), while the rest of the blocks don't? It seems like a fairly generic helper to me…

I don't know why, either. @Aljullu works on that block, so he'll know better than me.

Aljullu (commented):
I smoke tested some blocks, including the _Add to Cart with Options_ block, and everything seems to be working as it used to on my end:

https://github.com/user-attachments/assets/5c2b4352-8e6b-46c8-b036-a632eff08c80

I will let the others do the final approval as I they have a better acknowledge of the Interactivity API, but no blockers from my side. :slightly_smiling_face: 

> Hmm... interesting. May I know what the special needs are that make that block require [this helper](https://github.com/woocommerce/woocommerce/blob/d0e2c0bc7badced522e6b0f4d9d271808b45fd38/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorAttributeOptions.php#L100-L119), while the rest of the blocks don't? It seems like a fairly generic helper to me…

I'm not 100% sure, to be honest. Apparently, it has a [special check for null values](https://github.com/woocommerce/woocommerce/blob/d0e2c0bc7badced522e6b0f4d9d271808b45fd38/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorAttributeOptions.php#L106-L108), which might come from [`get_default_selected_attribute()`](https://github.com/woocommerce/woocommerce/blob/d0e2c0bc7badced522e6b0f4d9d271808b45fd38/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsVariationSelectorAttributeOptions.php#L121-L135). However, after testing it, I don't think it's needed and we can remove it. `wp_interactivity_data_wp_context()` handles `null` values without problems.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
Aljullu: I wanted to `+1` to this. Specially in cases where we work with the `WP_HTML_Tag_Processor`, it would be much more convenient to have the attribute value instead of the formatted string of attribute name + value.

luisherranz: Feel free to add a helper to test it in the Woo code that can then be added to WordPress Core.

nerrad:
@dinhtungdu and @SantosGuillamot - is this PR still relevant? it looks like the [Linear issue](https://linear.app/a8c/issue/WOOPLUG-4028/review-and-standardize-woocommerce-blocks-templating) Tung created and referenced in here was automatically archived by Linear (before we turned off auto-archiving). Corresponding GitHub issue is #57445.

dinhtungdu:
IMO, we should close this and replace it with a smaller one, gradually targeting a smaller number of blocks instead. I'm closing this one but please feel free to reopen if you disagree.

dinhtungdu:
For the linked issue, I will keep it close for now, it's still valid but not something we should address this time.

PR #57330, opened 2025-04-16
Add subject and pre-header to email editor sidebar Settings panel

lysyjan:

- This PR adds new rich-text components to the email editor sidebar, where it is possible to configure the subject and preview text. 

Closes WOOPLUG-3680.

| Before | After |
| ------ | ----- |
|Screenshot 2025-04-16 at 17 11 04|Screenshot 2025-04-16 at 16 50 28|

1. Enable the block-based email editor from WooCommerce → Settings → Advanced → Features → Block Email Editor (alpha)
2. Open an email in the editor WooCommerce → Settings → Emails
3. Select an email and set the subject and preheader, and use the personalization tags there
4. Replicate sending for the selected email (because preview does not work now)
5. Check that the received email has the configured subject and preheader correctly

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/sidebar_settings.tsx
+++ b/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/sidebar_settings.tsx
@@ -42,7 +42,7 @@ const SidebarSettings = ( { RichTextWithButton } ) => {
 					...woocommerce_data,
 					[ name ]: value,
 				},
-			},
+			}
 		);
 	};
 
@@ -51,13 +51,36 @@ const SidebarSettings = ( { RichTextWithButton } ) => {
 	return (
 		<>
 			<br />
-			<RichTextWithButton
-				attributeName="subject"
-				attributeValue={ woocommerce_email_data.subject }
-				updateProperty={ updateWooMailProperty }
-				label={ __( 'Subject', 'woocommerce' ) }
-				placeholder={ woocommerce_email_data.default_subject }
-			/>
+			{ woocommerce_email_data.email_type ===
+			'customer_refunded_order' ? (
+				<>
+					<RichTextWithButton
+						attributeName="subject_full"
+						attributeValue={ woocommerce_email_data.subject_full }
+						updateProperty={ updateWooMailProperty }
+						label={ __( 'Full Refund Subject', 'woocommerce' ) }
+						placeholder={ woocommerce_email_data.default_subject }
+					/>
+					<br />
+					<RichTextWithButton
+						attributeName="subject_partial"
+						attributeValue={
+							woocommerce_email_data.subject_partial
+						}
+						updateProperty={ updateWooMailProperty }
+						label={ __( 'Partial Refund Subject', 'woocommerce' ) }
+						placeholder={ woocommerce_email_data.default_subject }
+					/>
+				</>
+			) : (
+				<RichTextWithButton
+					attributeName="subject"
+					attributeValue={ woocommerce_email_data.subject }
+					updateProperty={ updateWooMailProperty }
+					label={ __( 'Subject', 'woocommerce' ) }
+					placeholder={ woocommerce_email_data.default_subject }
+				/>
+			) }
 
 			<br />
 			<RichTextWithButton
--- a/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/sidebar_settings.tsx
+++ b/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/sidebar_settings.tsx
@@ -1,6 +1,8 @@
 /**
  * External dependencies
  */
+import { select, dispatch } from '@wordpress/data';
+import { store as coreDataStore, useEntityProp } from '@wordpress/core-data';
 import { addFilter } from '@wordpress/hooks';
 import { __ } from '@wordpress/i18n';
 
@@ -12,9 +14,32 @@ import { NAME_SPACE } from './constants';
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
 const SidebarSettings = ( { RichTextWithButton } ) => {
+	const [ woocommerce_email_data ] = useEntityProp(
+		'postType',
+		'woo_email',
+		'woocommerce_data'
+	);
+
 	const updateWooMailProperty = ( name: string, value: string ) => {
-		console.log(name);
-		console.log(value);
+		const editedPost = select( coreDataStore ).getEditedEntityRecord(
+			'postType',
+			'woo_email',
+			window.WooCommerceEmailEditor.current_post_id
+		);
+
+		// @ts-expect-error Property 'mailpoet_data' does not exist on type 'Updatable<Attachment<any>>'.
+		const woocommerce_data = editedPost?.woocommerce_data || {};
+		void dispatch( coreDataStore ).editEntityRecord(
+			'postType',
+			'woo_email',
+			window.WooCommerceEmailEditor.current_post_id,
+			{
+				woocommerce_data: {
+					...woocommerce_data,
+					[ name ]: value,
+				},
+			},
+		);
 	};
 
 	return (
@@ -22,7 +47,7 @@ const SidebarSettings = ( { RichTextWithButton } ) => {
 			<br />
 			<RichTextWithButton
 				attributeName="subject"
-				attributeValue="Subject"
+				attributeValue={ woocommerce_email_data.subject }
 				updateProperty={ updateWooMailProperty }
 				label={ __( 'Subject', 'woocommerce' ) }
 				placeholder={ __(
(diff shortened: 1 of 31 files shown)


plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/sidebar_settings.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { addFilter } from '@wordpress/hooks';
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import { NAME_SPACE } from './constants';
+
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
```
costasovo: Could you please replace this with @ts-expect-error? This will give us a notification when the error is fixed e.g., via updating a package, etc. Also, please provide a basic comment on why we are accepting the error.

plugins/woocommerce/src/Internal/EmailEditor/EmailApiController.php
```diff
@@ -24,9 +37,13 @@ public function get_email_data( $post_data ): array {
 		$email_type  = get_post_meta( $post_data['id'], Integration::WC_EMAIL_TYPE_ID_POST_META_KEY, true );
 		$post_option = get_option( "woocommerce_{$email_type}_settings" );
 
+		$email = $this->get_email_by_type( $email_type );
+
 		return array(
 			'subject' => $post_option['subject'] ?? null,
 			'heading' => $post_option['heading'] ?? null,
+			'default_subject' => $email->get_default_subject(),
+			'default_heading' => $email->get_default_heading(),
```
costasovo: Why do we need to pass these defaults as separate properties? Can't we just use them as a subject fallback directly?

```php
'subject' => $post_option['subject'] ?? $email->get_default_subject(),
```

plugins/woocommerce/includes/emails/class-wc-email-cancelled-order.php
```diff
@@ -97,6 +97,10 @@ public function trigger( $order_id, $order = false ) {
 				$this->placeholders['{order_billing_full_name}'] = $this->object->get_formatted_billing_full_name();
 			}
 
+			if ( $this->block_email_editor_enabled ) {
+				$this->personalizer->set_context( $this->prepare_context_data( $this->personalizer->get_context() ) );
+			}
+
```
costasovo: I think we should try to avoid modifying individual emails as much as possible. In this case we repeat the same two lines in every email type class. This will add a requirement that every plugin that adds email type will need to add these two lines.

Could we move it to `WC_Email::send` ? It receives the subject, so we could personalize it directly in the method or add a new filter there. I think it would be better to avoid changing every sinle WC_Email. What do you think?

plugins/woocommerce/includes/emails/class-wc-email.php
```diff
@@ -1429,6 +1451,29 @@ protected function get_option_or_transient( string $key, $empty_value = null ) {
...
+		if ( $this->object instanceof \WP_User ) {
+			$context['wp_user'] = $this->object;
+		} elseif ( $this->object instanceof WC_Order ) {
+			$context['wp_user'] = $this->object->get_user();
+		} else {
+			$context['wp_user'] = null;
+		}
+		$context['wc_email'] = $this;
+		return $context;
+	}
```
costasovo: I see that we already have this logic in 
https://github.com/woocommerce/woocommerce/blob/5e8830758700e6c5e3655adcd2e571ab8b945ab0/plugins/woocommerce/src/Internal/EmailEditor/BlockEmailRenderer.php#L134-L141

Can we somehow avoid duplicating the logic for creating the context? I think that in the future when we will implement the Woo Order blocks we may need to be able to pass such a context directly into block renderers. So having a service that could build or build and carry sending context would be helpful.

A brief draft I have in my mind could look like this:
```php
class SendContext {
  private $context = [];
  
  public function addContextFromWCEmail( WC_Email $email) {
     ...the code you have here
  }
  
  public function getContext() {
    return $this->context;
  }
}
```
Do you think something like that could work?

costasovo (commented):
@lysyjan Great job!

I found a few places I would like to discuss. Please see my other comments.
Note: I didn't test the feature locally in detail, because I would like to discuss the comments first.

plugins/woocommerce/src/Internal/EmailEditor/EmailApiController.php
lysyjan: @costasovo The default values are used as placeholders, not as default values. Do you think we should use another name in the response here?

costasovo: @lysyjan Placeholders, I see. How does it work during sending, when a user doesn't save any Subject and just keeps the placeholder? What subject would be used?

lysyjan: Sending uses the [WC_Email::get_subject()](https://github.com/woocommerce/woocommerce/blob/b9956f17234492c7272aed5f16684d0ca2062aa3/plugins/woocommerce/includes/emails/class-wc-email.php#L485-L505) method, which handles logic such as the use of the default subject if needed. I'd say the logic is the same as the old email setup page.

plugins/woocommerce/includes/emails/class-wc-email-cancelled-order.php
lysyjan: > Could we move it to WC_Email::send ? It receives the subject, so we could personalize it directly in the method or add a new filter there. I think it would be better to avoid changing every sinle WC_Email. What do you think?

I am not sure about this approach. Because the replacement would be used only in the case we send the email through this method. But what about a preview? For example, sending a preview is managed in the PHP email editor package through `wp_mail`. Another use case can be HTML preview.  I prefer keeping the setting context for personalization tags in the same place as it was for the old placeholders. Maybe it would be better to chat about it if you want.

plugins/woocommerce/includes/emails/class-wc-email.php
lysyjan: I was looking at this, and we can remove the setting context in the `BlockEmailRenderer` class because using this class and the main method is a part of the `get_content()` method. In my opinion, the context should be set before using `get_content()`.

lysyjan:
@costasovo I refactored the code related to setting context for personalization to reduce code repetition. There are three new commits. Please let me know what you think about this approach.

plugins/woocommerce/includes/emails/class-wc-email-cancelled-order.php
costasovo: I see that you've implemented a different approach and now we don't have to touch every single email 👍

plugins/woocommerce/includes/emails/class-wc-email.php
costasovo: Makes sense. It is good that we have that in one place. I think that in the future, we might need context in different places than in the personalizer (e.g., blocks that render an order table might need it as well), but we can address that later when we have more details.

plugins/woocommerce/includes/emails/class-wc-email.php
```diff
@@ -479,12 +486,37 @@ public function get_additional_content() {
...
+	 * This method ensures the Personalizer context is set correctly before emails are generated.
+	 * It skips context reconfiguration if it's already set.
+	 *
+	 * @return void
+	 */
+	protected function ensure_personalization_context_is_set(): void {
+		if ( $this->context_set ) {
+			// Context is already set, skip reconfiguration.
+			return;
+		}
```
costasovo: I think there might be an issue with this caching. Can we ensure we don't cache "empty" context?

The context data is set in trigger methods in classes extending WC_Email. So I have a concern that when somebody calls
`$email->getSubject()` before the trigger method is called, it may cause us to store an empty context, and we won't update it after the trigger is actually called.

costasovo (commented):
@lysyjan Thank you for the improvements. I like how the personalization is aligned with the original placeholder replacements.

I added one comment with a question about caching the context.

I did some basic testing locally and set subject and preheder for a couple of emails and email I received had proper subject and preheader.

I noticed a small issue with the subject, though. It seems that because we use Richtext it may contain some HTML.

**Here is how I set it:**
Screenshot 2025-04-29 at 13 33 54

**Here is what I received:**
Notice `<br>` tag in the subject.
Screenshot 2025-04-29 at 13 32 55
After I inserted the second personalization tag to subject via modal it added an extra line between the fris couple of words in the subject and newly inserted tag. Perhaps because I placed the cursor there. I'm not sure. It is probably not that important. The problem is that new lines were interpreted as <br> tags. Tags are not supported in subject in email clients so I think we should strip them.

lysyjan:
@costasovo Thank you for your ideas and testing. I added two more commits with tiny adjustments. I think it solves the mentioned issues.

costasovo (approved):
@lysyjan Great job!

Kudos for going with composition for the TransactionalEmailPersonalizer 👍. Great idea!

### Testing performed

I tested with the "Processing order" email which is sent to a shopper after they make an order.

I filled custom subject and preheader with personalization tags:

Screenshot 2025-05-02 at 12 47 04

and I can see I received the email with proper values in subject and preheader.

Screenshot 2025-05-02 at 12 46 21

I also tested with an empty subject and I see it uses the default subject:

Screenshot 2025-05-02 at 12 47 40

Issue #52835, opened 2024-11-14
Product Summary: Improve summary truncation with HTML API


kmanijak:
**Background:**

https://github.com/woocommerce/woocommerce/pull/48984 introduces improvements to Product Summary block, among others it preserves styling from Product Editor (essentially HTML). This doesn't work perfectly with truncation feature.

Method introduced in the above PR ([here](https://github.com/woocommerce/woocommerce/pull/48984/files#diff-e96fb009919c72732ff3382c3a9864180be8be430e8359cf0377c0e24515939cR112)) is far from perfection. It works well and reliably with the simple text but works much worse with the rich text (HTML) as it shortens out the text due to artifacts that occur from decoding HTML. It can be considered a good balance between the what we want and what we can. Unfortunately, ATM there's no good tool for robust solution:
- `DOMDocument` is a reasource-heavy tool and isn’t available on many hosting platforms (POC: https://github.com/woocommerce/woocommerce/pull/52769),
- HTML API doesn't support HTML manipulations just yet.

However, once HTML API will allow for HTML manipulation, the current `trim_keeping_html_formatting` method should be replaced with implementation based on it along with JS (Editor) equivalent implemented by @gigitux in this POC: https://github.com/woocommerce/woocommerce/pull/51783.

PR #57342, opened 2025-04-17
Variable product description in featured-product block fixed

### LINKED ISSUE #42302: Featured Product Block: does not show variable product description on backend side.
### Prerequisites

### Describe the bug

I tested featured product block and found the does not showing variable product description in featured product block backend side. Please check the below provided video recording for a better understanding of issue.

Video URL: https://share.cleanshot.com/G58ZWrJ4Rzfrsb0GyGfD

### Expected behavior

Need to showing variable product description on backend side as well. 

### Actual behavior

does not showing variable product description on backend side.

### Steps to reproduce

1. Create a new page and take the featured product block.
2. Next, Select variable product inner any option.
3. And check, description are not showing. Also check it with "Show description" toggle setting.

### WordPress Environment

WC Version: 7.8.2
WP Version: 6.2.2
Active Theme: Twenty Twenty-Three (twentytwentythree)
Active theme version:1.1

Server Environment

Web server: nginx/1.16.0
PHP Version: 7.3.5 (Supports 64bit values)
PHP memory limit: 256M
PHP Time Limit: 1200
PHP Max Input Vars: 4000
cURL Version: 7.54.0 LibreSSL/2.6.5
OpenSSL/3.0.2

### Isolating the problem

- [X] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/storefront/).

Kallyan01:
### Issue Description:

I have tested the featured product block and found it's not showing the variable product description in featured product block backend side/edit mode .

### Issue root cause:

- The block doesn’t show any description in the editor because the product's short description is empty. However, on the frontend, the description appears since there's a fallback logic that displays up to 400 words from the full product description when the short description is missing. [[Ref](https://github.com/woocommerce/woocommerce/blob/027bf00f291967608abbbd6408193c970dffdd2a/plugins/woocommerce/src/Blocks/BlockTypes/FeaturedProduct.php#L88)]

### Changes Proposed in this Pull Request:

If there's no short description available, but a full product description exists, then show the full description—up to 400 words similar to frontend logic.

Closes [#42302](https://github.com/woocommerce/woocommerce/issues/42302)

  | Before
image
  | After   
image

1. Create a new page and take the featured product block.
2. Next, Select a product with a long description.
3. Ensure that the description is trimmed.
4. Save the post.
5. Visit the saved post.
6. Ensure that the description is trimmed as on the editor side.
7. Smoke test block manually by different variable products both short and long description values present.

- Tested the block manually by different variable products both short and long description values present .

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


Kallyan01:
@gigitux @ralucaStan waiting for your review ! Thanks 🙌

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/utils.ts
```diff
@@ -91,3 +91,14 @@ export function dimRatioToClass( ratio: number ) {
...
 }
+
+/**
+ * Convert long texts into short preview format with trailing dots.
+ *
+ * @param {string} text   store the the long string.
+ * @param {number} length preview text's length.
+ * @return {string} The truncated preview text.
+ */
+export function getLongTextPreview( text: string, length: number ): string {
```
gigitux: This code looks okay, but in WooCommerce we already have the `trimWords` function:

https://github.com/Kallyan01/woocommerce/blob/585311cd260a97010fbaca2a692766ff9694fef4/plugins/woocommerce/client/blocks/assets/js/utils/trim-words.ts#L37-L58

Could you try to re-use it?

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
```diff
@@ -295,7 +296,13 @@ export const withFeaturedItem =
...
 										__html:
 											category?.description ||
-											product?.short_description,
+											product?.short_description ||
+											( product?.description?.length > 0
+												? getLongTextPreview(
+														product.description,
+														400
+												  )
+												: '' ),
```
gigitux: Consider adding a comment explaining why we’re returning a 400 here. Ideally, link to the relevant PHP logic for context.

Something like this could make sense:

```
// Returning 400 to match the validation logic in PHP: see https://github.com/Kallyan01/woocommerce/blob/585311cd260a97010fbaca2a692766ff9694fef4/plugins/woocommerce/src/Blocks/BlockTypes/FeaturedProduct.php#L88-L89
```

gigitux (changes requested):
Thanks for the PR! The feature works well. That said, we could simplify the code by reusing some existing functions already available in the monorepo.

gigitux (approved):
Thanks for working on this PR! With 37ae3c8a1a0018fc6aa2f4712b815e7e21439c84, I used `trimCharacters` instead of `trimWords` to ensure that the editor side preview matches with the frontend. Everything looks good to me! :shipit: 


### Before

| Editor | Frontend |
|--------|--------|
|image|image| 



### After

| Editor | Frontend |
|--------|--------|
|image| image |

Kallyan01:
Hi @gigitux,  

Apologies for the delay—I've been doing some testing on the local changes done by me where i've used `trimCharacters()` too so haven't asked for re-review after pushing the last [commit](https://github.com/woocommerce/woocommerce/pull/57342/commits/9ed3653dc6e0f6bcb179dbb889814ef4d08621fd) , i've identified an edge case and an issue with the `wc_trim_string()` function in PHP (reference: [[WooCommerce GitHub](https://github.com/woocommerce/woocommerce/blob/027bf00f291967608abbbd6408193c970dffdd2a/plugins/woocommerce/src/Blocks/BlockTypes/FeaturedProduct.php#L88)](https://github.com/woocommerce/woocommerce/blob/027bf00f291967608abbbd6408193c970dffdd2a/plugins/woocommerce/src/Blocks/BlockTypes/FeaturedProduct.php#L88

**Existing issues :**
- By default, the PHP function `wc_trim_string()` does not strip HTML tags from the input string, which leads to inaccurate length calculations—it's including the HTML tags in the character count.This results in the visible string being trimmed in editor and string with tags is being trimmed in frontend .
- In PHP, the trailing ellipsis (...) is included within the 400-character limit enforced by wc_trim_string(), whereas in JavaScript, the trimCharacters() it's not.

**Editor**
image

**Frontend**
image


I think it's better to sanitize the content using `wp_strip_all_tags()` before passing it into `wc_trim_string()` to ensure it trims based solely on the visible text ? and trip only 397 chars in JS so that it matches the 400 char trim logic.
Let me know your thoughts on handling these issues !

gigitux:
Thanks for pointing this out! It’s similar to [#52835](https://github.com/woocommerce/woocommerce/issues/52835).

The discrepancy comes from how the description length is computed: on JS side, the calculation excludes HTML tags, whereas in PHP, it includes them.

We should improve the trimming logic on the PHP side to align the behavior. It's important that we still preserve the correct HTML structure in the output.

gigitux:
I think that this PR is valuable because for simple cases, the editor side matches on the frontend side. Feel free to work on https://github.com/woocommerce/woocommerce/issues/52835, but it could be pretty complex to do.

PR #57344, opened 2025-04-17
[Blueprint] Enhance SQL execution safety

chihsuan:

Closes WOO6-10

The runSQL step is the most danger step in the blueprint import process, as it enables the execution of arbitrary SQL queries.

It's only accessible to users with the `manage_options` capability, but it's still a potential security risk if users import malicious blueprints without proper validation.

This PR enhances the security and reliability of SQL execution in the Blueprint's ImportRunSql class by implementing stricter validation and comprehensive test coverage.

### Key Changes

- Enhance SQL execution safety in the ImportRunSql class:
  - Add validation for SQL query types, only allow "INSERT", "UPDATE", "REPLACE INTO" queries
  - Implement protection against SQL injection patterns
  - Add safeguards for protected tables and user capabilities
  - Improve error handling and reporting
-  Refactor SQL query preparation in ExportWCTaxRates class for improved security
- Add comprehensive unit test coverage

### Test import

1. Enable blueprint feature `wp option set woocommerce_feature_blueprint_enabled yes`
2. Try importing an invalid SQL query; it should return an error  `wc blueprint import invalid-sql.json --show-messages=all`
3. Try importing a valid SQL query; it should return a success message  `wc blueprint import valid-sql.json --show-messages=all`
4. Act as a malicious user and try different malicious SQLs; it should return an error.

- [invalid-sql.json](https://github.com/user-attachments/files/19791393/invalid-sql.json)
- [valid-sql.json](https://github.com/user-attachments/files/19791394/valid-sql.json)

Screenshot 2025-04-17 at 16 42 07

### Test export tax

1. Create some tax rates
2. Run `wp wc blueprint export blueprint.json --steps=setWCTaxRates`
3. Confirm it generates the blueprint file with expected SQL statements

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


packages/php/blueprint/src/Importers/ImportRunSql.php
```diff
@@ -79,4 +146,173 @@ public function check_step_capabilities( $schema ): bool {
...
+	 * to hide malicious SQL commands or bypass security filters.
+	 *
+	 * @param string $sql_content The SQL query to check.
+	 * @return bool True if suspicious comments found, false otherwise.
+	 */
+	private function contains_suspicious_comments( string $sql_content ): bool {
+		// Quick check if there are any comments at all before running regex.
+		if ( strpos( $sql_content, '--' ) === false &&
+		strpos( $sql_content, '/*' ) === false &&
+		strpos( $sql_content, '#' ) === false ) {
```
louwie17: Some formatting issues here

packages/php/blueprint/src/Importers/ImportRunSql.php
```diff
@@ -79,4 +146,173 @@ public function check_step_capabilities( $schema ): bool {
...
+	}
+
+
+	/**
+	 * Check for common SQL injection patterns.
+	 *
+	 * @param string $sql_content The SQL query to check.
+	 * @return bool True if potential injection patterns found, false otherwise.
+	 */
+	private function contains_sql_injection_patterns( string $sql_content ): bool {
```
louwie17, line 229: Curious where you got these patterns from and some of the other checks in this file?

louwie17 (approved):
I think the fact we are already limiting this by user type goes a long way, this also tested well and code looks good, although curious where you got these checks from?

There are a couple other things though that I am wondering about ( we could do this as follow up ):
- From looking at the code this would currently break for any WordPress instances that make use of table prefixes right? which brings me to the second suggestion.
- I know this mimics the Playground step, but I wonder if we should change the structure of the run sql step to limit it and mimic what insert, update functions that wpdb support https://developer.wordpress.org/reference/classes/wpdb/#update-rows ( it would also allow us to better support table prefixes )

packages/php/blueprint/src/Importers/ImportRunSql.php
chihsuan: Reformatted in d9ad44f2045eed0921f9be733d34bc8bae90d347

packages/php/blueprint/src/Importers/ImportRunSql.php
chihsuan: I used AI tools (ChatGPT, Claude, Cursor) to identify potential security issues and gather suggestions for enhancing the code's security measures.

chihsuan:
> * From looking at the code this would currently break for any WordPress instances that make use of table prefixes right? which brings me to the second suggestion.

Yes, it would break for any WordPress instances that use table prefixes.

> * I know this mimics the Playground step, but I wonder if we should change the structure of the run sql step to limit it and mimic what insert, update functions that wpdb support https://developer.wordpress.org/reference/classes/wpdb/#update-rows ( it would also allow us to better support table prefixes )

This is a good idea! But I think using the same structure will help us use the same package across various projects. I'll check with the Playground team to see if they have any input. Thanks for the suggestion!

louwie17:
> > * From looking at the code this would currently break for any WordPress instances that make use of table prefixes right? which brings me to the second suggestion.
> 
> Yes, it would break for any WordPress instances that use table prefixes.

Gotcha, that does make the `runSql` step not very useful in this context, as a lot of WP/WC sites use prefixes. I guess we will gather that from feedback the only way to really fix that is to change to format, as we shouldn't solve that with regex.

> This is a good idea! But I think using the same structure will help us use the same package across various projects. I'll check with the Playground team to see if they have any input. Thanks for the suggestion!

Thanks! I think the generic `runSql` makes a lot of sense in the Playground instance, but a bit less so for existing sites. It be nice if Playground could support both and we only support the more defined version.

louwie17:
Btw feel free to merge this, I was just curious about the above questions, but they weren't blockers.

PR #57355, opened 2025-04-17
Properly account for custom order status

straku:

In the Store API, we were always setting the order status to "pending" before processing payment. It was happening regardless of whether some custom status was set through the available `woocommerce_store_api_checkout_order_processed` hook. This behavior is different from shortcode, breaking expectations for advanced usages through custom plugins as described in the original report.

Closes [WOOPLUG-1743](https://linear.app/a8c/issue/WOOPLUG-1743/block-checkout-process-order-without-payment-sets-order-as-pending)

As far as I am able to tell, this bug has been there since the inception of the Store API.

### Screen recording

This screen recording shows that when you check out using shortcode and blocks checkout, the finalized order is in custom status called "Ready for Pickup". The custom status is added through a snippet that you can inspect in the testing steps. On `trunk` both of these flows would end up in "Completed" status, regardless of whether order status was set to a custom one through `woocommerce_store_api_checkout_order_processed` hook.

https://github.com/user-attachments/assets/f914ee18-c74e-404d-841f-d76069d8a3a2

Since this change is touching flows for orders that either require, but also not require payments, we need to test both of them. I suggest starting with simpler regression tests and then moving to a custom order status setup to repeat initial steps.

1. Add item to cart and go through the blocks checkout for an order that doesn't require payment. The order doesn't require payment if its cost is equal to 0. You can achieve it by having a free product (and a free shipping option). Or you can create a coupon with a 100% discount.
2. Make sure there are no errors in the flow and when you check the order details it is in the completed state.
3. Repeat step 1 but with an order that requires payment. Going with bank transfer or cash on delivery should be enough for testing this flow. 
4. Make sure there are no errors in the flow and when you check the order details it is in the completed state.
5. Enable a snippet that adds support for custom order status. You can use [Code Snippets](https://wordpress.org/plugins/code-snippets/) plugin for this purpose.

```
// 1. Register new status in WooCommerce/WordPress
function add_ready_for_pickup( $order_statuses ) {
    $order_statuses['wc-ready_for_pickup'] = array(
        'label'                     => 'Ready for Pickup',
        'public'                    => false,
        'exclude_from_search'       => false,
        'show_in_admin_all_list'    => true,
        'show_in_admin_status_list' => true,
        'label_count'               => _n_noop( 'Ready for Pickup (%s)', 'Ready for Pickup (%s)' )
    );
    return $order_statuses;
}
add_filter( 'woocommerce_register_shop_order_post_statuses', 'add_ready_for_pickup' );

// 2. Add it to WooCommerce's status list
function add_ready_for_pickup_to_order_statuses( $order_statuses ) {
    $order_statuses['wc-ready_for_pickup'] = 'Ready for Pickup';
    return $order_statuses;
}
add_filter( 'wc_order_statuses', 'add_ready_for_pickup_to_order_statuses' );

// 3. Set order to this status after checkout
function set_order_to_ready_for_pickup_store_api( $order ) {
    $order->set_status( 'ready_for_pickup' );
    $order->save();
}
add_action( 'woocommerce_store_api_checkout_order_processed', 'set_order_to_ready_for_pickup_store_api', 10, 1);
```

6. Repeat steps 1-4. But this time around, the end status of both orders should be set to "Ready for Pickup".

Steps outlined above, but extended to the shortcode checkout to make sure the flows are the same.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


opr (commented):
Hey @straku thanks for this PR - I've just been digging deeper into this because I noticed an error while trying to pay with Stripe.

It shows this:

image

Digging deeper, it led me to https://github.com/woocommerce/woocommerce/blob/0b7aaa15a3a1724e11aeabdf1411d6b36cc22381/plugins/woocommerce/includes/class-wc-order.php#L1780-L1790

Because the status is not one of the allowed ones for payment, the order is marked as not needing payment, but Stripe still tries to pay.

Adding this snippet fixes it:

```php
add_filter( 'woocommerce_valid_order_statuses_for_payment', 'add_ready_for_pickup_to_valid_payment_statuses' );
function add_ready_for_pickup_to_valid_payment_statuses( $statuses ) {
	$statuses[] = 'ready_for_pickup';
	return $statuses;
}
```

however I'm wondering if this indicates an issue with how/when we change status, and the `woocommerce_store_api_checkout_order_processed` hook. I'm thinking we should add a similar hook for _after_ the payment has been processed, what do you think? I'll also cc @mikejolley on this one for thoughts on adding a new hook after payment processing

mikejolley:
> In the Store API, we were always setting the order status to "pending" before processing payment. It was happening regardless of whether some custom status was set through the available woocommerce_store_api_checkout_order_processed hook. This behavior is different from shortcode, breaking expectations for advanced usages through custom plugins as described in the original report.

Do you mean the `woocommerce_default_order_status` hook? 

Note that `pending` is unpaid, and default status (pending) is the unpaid status. Not the same as the `processing` status after payment.

I don't understand the test case you've added here:

```
Order status should not change after payment complete.
```

The order status 100% should change, to `processing` `on-hold` or `completed`, which the payment gateway controls. Some extensions rely on transition hooks such as `pending_to_processing` to implement logic such as email alerts.

straku:
> I'm thinking we should add a similar hook for after the payment has been processed, what do you think?

@opr I think it makes sense. Just verifying whether I understand the flow correctly:
- we would allow users to set a custom status before payment
- payment would understand that this is a valid status and proceed through the default states `processing`, `on-hold` and `completed` (the one mentioned in @mikejolley comment)
- after payment, we would allow users to set a custom status again

--- 

Some answers to @mikejolley comment below.

> Do you mean the `woocommerce_default_order_status` hook?

No, I understand that this is a hook that allows you to set a default status for an order, but we've focused on the ones mentioned in the original bug report, which are:
- `woocommerce_checkout_order_processed` in shortcode
- `woocommerce_store_api_checkout_order_processed` in blocks checkout

> The order status 100% should change.

We've tested shortcode flow for orders that require (and not) payments, and the order status remained set to a custom one after processing payment. My personal testing might've been flawed as I didn't set up Stripe yet (will try doing it today), but I wonder how it worked for @opr after using `woocommerce_valid_order_statuses_for_payment` to unblock the flow. Can you share what was the end status after the payment?

opr:
> but I wonder how it worked for @opr after using woocommerce_valid_order_statuses_for_payment to unblock the flow. Can you share what was the end status after the payment?

After I added '`ready_for_pickup`' to the `woocommerce_valid_order_statuses_for_payment` the end status was `ready_for_pickup` in the order admin.

opr:
> The order status 100% should change, to processing on-hold or completed, which the payment gateway controls

Yep, but then it's OK to transition away from that after it's been set right? The flow should be like this imo:

For orders requiring payment:
`checkout-draft` -> `pending` -> `whatever the gateway sets` -> `at this point, extensions should be free to modify the status further`

For orders not requiring payment:

`checkout-draft` -> `pending` -> `at this point, extensions should be free to modify the status further`.

We don't have a place where extensions can do this.

opr:
👋🏼 @straku I've just been chatting with Mike and I think that the proposed fix here could be simplified.

If we switch the order status to pending just _before_ the `woocommerce_store_api_checkout_order_processed` action fires, it would still ensure orders switch to pending (which is currently happening in `process_payment` and `process_without_payment`), so hoisting it up one level would ensure the order is in the same status unless modified after the action.

I think this is fine, because modifying the status and expecting other unrelated gateways to still work with it is not supported, right @mikejolley ?

straku:
It feels better from a "semantic" point of view, and would address the original issue for orders that do not require a payment. But if I understand correctly, it wouldn't resolve the issue that you've encountered? The result would be aligned with what is proposed in this PR, unless I am missing something 🤔

opr:
> But if I understand correctly, it wouldn't resolve the issue that you've encountered? The result would be aligned with what is proposed in this PR, unless I am missing something 🤔

I think that's correct, but I also think the use-case of changing the status should be limited to when specific gateways are used. The extension author shouldn't change the status if Stripe is being used to pay for the order, for example.

straku:
@opr I did more digging today. If we move setting the status to the place you suggested it won't have an opportunity to ever go into `process_payment` flow (without usage of additional filters) as `needs_payment` is checking whether the current order status is valid. The ones considered valid can be changed by a filter, but initially are either `pending` or `failed`.

To summarize the flow:
1. Set to `pending` (the change you suggested)
2. Hook fires allowing to set a custom status, let's assume we set a custom status `ready-for-pickup`
3. `needs_payment` is checked, it returns `false` as `ready-for-pickup` is not a valid status
4. Since the above is `false` we go to `process_without_payment` flow (we will always end up here as long as custom status will be set in 2.)

What do you think about it?

opr:
> @opr I did more digging today. If we move setting the status to the place you suggested it won't have an opportunity to ever go into `process_payment` flow (without usage of additional filters) as `needs_payment` is checking whether the current order status is valid. The ones considered valid can be changed by a filter, but initially are either `pending` or `failed`.
> 
> To summarize the flow:
> 
> 1. Set to `pending` (the change you suggested)
> 2. Hook fires allowing to set a custom status, let's assume we set a custom status `ready-for-pickup`
> 3. `needs_payment` is checked, it returns `false` as `ready-for-pickup` is not a valid status
> 4. Since the above is `false` we go to `process_without_payment` flow (we will always end up here as long as custom status will be set in 2.)
> 
> What do you think about it?

Good digging, but I still think it's OK, given:

> 3. `needs_payment` is checked, it returns `false` as `ready-for-pickup` is not a valid status

I think if someone extending this endpoint wants to update the status and _still_ require payment, they should also update the valid statuses that'll require it.

@webdados can you help us a bit with your use-case for changing the order status? Does it match up with what we're saying here?

webdados:
@opr are you sure you meant me? Can't find any issue opened by me regarding this (but I might be missing something)

opr:
> @opr are you sure you meant me? Can't find any issue opened by me regarding this (but I might be missing something)

Yep! https://github.com/woocommerce/woocommerce/issues/42748 it was this one

webdados:
Got it. It's from 2023 and that's why I couldn't remember, and I found a workaround in the meantime.

Use case: This is for a plugin that sets the order in a custom status "wc-swoa-for-approval" (needs approval, so that the admin can deal with it and approve it or cancel). "Pending" or "On hold" does not do the job for us.

For the classic checkout we just do this:
```
/**
 * Maybe set the order as needing approval after the classic checkout
 *
 * @param int    $order_id The order ID.
 * @param array  $posted_data Posted data from checkout.
 * @param object $order WC_Order.
 */
public function maybe_set_order_as_need_approval_classic( $order_id, $posted_data, $order ) {
	if ( $this->needs_approval( $order ) ) {
		// Make sure all emails are instantiated.
		\WC_Emails::instance();
		// Update order status.
		$order->update_status( 'wc-swoa-for-approval', __( 'Order needs approval', 'simple-woocommerce-order-approval' ) . ' classic' );
	}
}

add_action( 'woocommerce_checkout_order_processed', array( $this, 'maybe_set_order_as_need_approval_classic' ), 10, 3 );
```

For the blocks checkout we were trying to use the equivalent hook `woocommerce_store_api_checkout_order_processed` without success.

We found a workaround tho. We use that hook to check if we need to change to our custom status or not and then hook into `woocommerce_pre_payment_complete` to "really" do it:

```
/**
 * Maybe set the order as needing approval after the blocks checkout
 * Not working because of this: https://github.com/woocommerce/woocommerce/issues/42748
 * Workaround: use woocommerce_pre_payment_complete to change the status and then woocommerce_valid_order_statuses_for_payment_complete to remove pending because the object is not actually changed
 *
 * @param object $order WC_Order.
 */
public function maybe_set_order_as_need_approval_blocks( $order ) {
	if ( $this->needs_approval( $order ) ) {
		// Make sure all emails are instantiated.
		\WC_Emails::instance();
		// Update order status.
		// $order->update_status( 'wc-swoa-for-approval', __( 'Order needs approval', 'simple-woocommerce-order-approval' ) ); // Not working because of this: https://github.com/woocommerce/woocommerce/issues/42748
		// Workaround: Update later as the blocks checkout will put it as pending
		add_action( 'woocommerce_pre_payment_complete', array( $this, 'really_set_order_as_need_approval_blocks' ) );
	}
}

add_action( 'woocommerce_store_api_checkout_order_processed', array( $this, 'maybe_set_order_as_need_approval_blocks' ) );

/**
 * Really set the order as needing approval after the blocks checkout
 *
 * @param int $order_id The order ID.
 */
public function really_set_order_as_need_approval_blocks( $order_id ) {
	$order = wc_get_order( $order_id );
	$order->update_status( 'wc-swoa-for-approval', __( 'Order needs approval', 'simple-woocommerce-order-approval' ) . ' blocks' );
	// Now filter woocommerce_valid_order_statuses_for_payment_complete to remove pending (because we changed the status on the database but not on the order object)
	add_filter( 'woocommerce_valid_order_statuses_for_payment_complete', array( $this, 'really_set_order_as_need_approval_blocks_remove_pending' ) );
	// Just in case, remove the action
	remove_action( 'woocommerce_pre_payment_complete', array( $this, 'really_set_order_as_need_approval_blocks' ) );
}

/**
 * Temporarily remove pending as a valid status for payment complete, so that the order stays in our custom status - blocks checkout
 *
 * @param array $statuses The valid statuses for payment complete.
 */
public function really_set_order_as_need_approval_blocks_remove_pending( $statuses ) {
    $key = array_search( 'pending', $statuses, true );
    if ( $key ) { 
        unset( $statuses[ $key ] );
    }
    // Just in case, remove the filter
    remove_filter( 'woocommerce_valid_order_statuses_for_payment_complete', array( $this, 'really_set_order_as_need_approval_blocks_remove_pending' ) );
    return $statuses;
}
```

It works, it's just not very elegant.

It would be less of a hack to be able to do it directly on `woocommerce_store_api_checkout_order_processed` (blocks) as we do on `woocommerce_checkout_order_processed` (classic), and that's why we never closed the issue.

webdados:
Also, to answer your question:

> Does it match up with what we're saying  here?

Yup! Exactly. It will go to `process_without_payment` because that's exactly what we need.

And I agree 100% with:

> I think if someone extending this endpoint wants to update the status and still require payment, they should also update the valid statuses that'll require it.

Actually, as you can see on my code above, I had to remove "pending" from the valid statuses for payment because I do update_status on the order, which will update on the database, but not on the current object being processed by WooCommerce. Maybe with this change this hack will not be needed anymore.

opr:
Thanks a lot for the explanation of the use case @webdados that helps a lot, I think! 🙇🏼

opr (approved):
Thanks @straku this is testing well for me, and I think we're all OK with the implications of changing the status without updating the valid statuses for payment.

Let's ship it 🚢 !

PR #57361, opened 2025-04-17
Ensure errors notices clear after valid requests on checkout

### LINKED ISSUE #54884: Errors should clear after valid PUT request on checkout
### Prerequisites

### Describe the bug

In #53553 we are persisting additional fields. If they are invalid, the error is displayed, but is not cleared once the shopper enters a valid value. Scrolling to the error could also be improved. As you can see in the video, even though the field was in view, the page scrolled. So a nice to have would be to only scroll if the field (or the parent of the error) is not in view. 

https://github.com/user-attachments/assets/8cfecbd2-800d-4290-8e43-526fb6de59e7

### Expected behavior

I would expect the error to clear after a successful PUT request.

### Actual behavior

The error should clear after a successful PUT request

### Steps to reproduce

1. Add an additional field with validation (e.g. an email address)
2. Go to Checkout block
3. Enter an invalid state in the field
4. Make sure the error is displayed
5. Enter a valid state in the field
6. The error is still visible.

### WordPress Environment

WP 6.7
WC 9.7

### Isolating the problem

tjcafferkey:

Closes https://github.com/woocommerce/woocommerce/issues/54884
Closes [WOOPLUG-3075](https://linear.app/a8c/issue/WOOPLUG-3075/errors-should-clear-after-valid-put-request-on-checkout)

| Before | After |
| ------ | ----- |
|        |       |

1. Ensure you're using [additional fields](https://github.com/woocommerce/additional-checkout-fields-tester)
2. Update this plugin to add another email field to the existing Alternative Email one (e.g Backup Email), or follow the steps in the below snippet for updates you can make to `additional-checkout-fields-tester.php`. This is required to check that the notices are grouped together, rather than replaced by the latest notice.
3. Add an invalid Alternative Email and blur the input, observe error notice appear
4. Add an invalid Backup Email (assuming this is the input you added) and blur the input, observe two notice errors
6. Fix each one individually and check that the notice disappears.

```php
// 1. add into custom_fields_tester_register_custom_checkout_fields()
	array(
		'id'                => 'plugin-namespace/backup-email',
		'label'             => 'Backup Email',
		'required'          => true,
		'validate_callback' => function ( $field_value ) {
			if ( ! is_email( $field_value ) ) {
				return new \WP_Error( 'invalid_backup_email', 'Please ensure your backup email matches the correct format.' );
			}
		},
	)
);

// 2. update woocommerce_validate_additional_field action
add_action(
	'woocommerce_validate_additional_field',
	function ( \WP_Error $errors, $field_key, $field_value ) {
		if ( 'plugin-namespace/gov-id' === $field_key ) {
			$match = preg_match( '/[A-Z0-9]{5}/', $field_value );
			if ( 0 === $match || false === $match ) {
				$errors->add( 'invalid_gov_id', 'Please ensure your government ID matches the correct format.' );
			}
		}
		if ( 'plugin-namespace/alt-email' === $field_key ) {
			if ( ! is_email( $field_value ) ) {
				$errors->add( 'invalid_alt_email', 'Please ensure your alternative email matches the correct format.' );
			}
		}

		if ( 'plugin-namespace/backup-email' === $field_key ) {
			if ( ! is_email( $field_value ) ) {
				$errors->add( 'invalid_backup_email', 'Please ensure your backup email matches the correct format.' );
			}
		}
	},
	10,
	3
);
```

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/base/utils/create-notice.ts
```diff
@@ -84,3 +84,33 @@ export const removeNoticesWithContext = ( context: string ) => {
 		removeNotice( notice.id, context );
 	} );
 };
+
+/**
+ * Remove a notice by its ID.
+ *
```
opr: Need to update the description of this I think. `id` is actually the _field_ ID, we are using that to find notices that _start_ with the field ID.

It also looks like `removeNoticeForField` is never called with context, I think we should remove that argument from the function, rename this function to `removeNoticesForField` (implying more than 1 notice may be removed) and remove the check that removes a notice in a specific context.

plugins/woocommerce/client/blocks/assets/js/data/checkout/utils.ts
```diff
@@ -283,3 +285,22 @@ export const validateAdditionalFields = (
...
+export const clearFieldErrorNotices = ( data: CheckoutPutData ) => {
+	if ( ! isObject( data ) ) {
+		return;
+	}
+
+	// Check if additional fields were updated successfully
+	if ( data?.additional_fields ) {
+		// Get the fields names that we should remove notices for.
+		// If a field is in the success response we can assume it has no errors.
+		const noticeFields = Object.keys( data.additional_fields );
```
opr, line 301: Do we want to harden this a bit and check it's actually an object? We have `isObject` type guards. If it makes it through the initial `if` then I think `Object.keys` would never error, but then we may foreach over data we don't want, for example _if_ `additional_fields` comes back as an array for some reason, we would iterate over each key in the array and applying `removeNoticeForField` with 0..n as the argument.

plugins/woocommerce/client/blocks/assets/js/data/utils/process-error-response.ts
```diff
@@ -78,7 +78,7 @@ export const getErrorDetails = (
 									message: decodeEntities(
 										additionalError.message
 									),
-									data,
```
opr, line 81: There are checks above (lines 67 - 70) to ensure `additionalError` has a `message` and `code` prop, do we want to include `data` in those checks?

plugins/woocommerce/client/blocks/tests/e2e/tests/checkout/additional-fields.shopper.block_theme.spec.ts
```diff
@@ -566,6 +570,36 @@ test.describe( 'Shopper → Additional Checkout Fields', () => {
 				}
 			);
 
+			await checkoutPageObject.page
+				.getByLabel( 'Alternative Email' )
+				.fill( 'test@' );
+
+			// Blur the field to trigger the validation.
+			await checkoutPageObject.page.evaluate(
+				'document.activeElement.blur()'
```
opr: why not

```js
await checkoutPageObject.page
	.getByLabel( 'Alternative Email' )
	.blur();
```

opr (approved):
Thanks @tjcafferkey this is testing well, I tested multiple different scenarios moving both fields from valid to invalid at the same time, moving both from invalid to valid, and then one valid and one invalid, the notices showed, and were removed, as expected.

Couple of comments suggesting improvements to the code, but overall I think they are not blocking, you can make the changes and merge.

plugins/woocommerce/client/blocks/assets/js/data/utils/process-error-response.ts
tjcafferkey: According to the type definition of `ApiParamError` and `ApiErrorResponse` it defines `data` as an optional param so I don't think we should add the check you're suggesting because we are anticipating that it can proceed (and not early return which is what would happen with your change) without `data`. We also add defensive checks before operating on `data` further down the stack.

plugins/woocommerce/client/blocks/tests/e2e/tests/checkout/additional-fields.shopper.block_theme.spec.ts
tjcafferkey: Honestly, didn't know it had that method available.

tjcafferkey: https://github.com/woocommerce/woocommerce/pull/57361/commits/8c19bfef13bc0c0955a47bf0ab3acb0c91bdac75

plugins/woocommerce/client/blocks/assets/js/data/checkout/utils.ts
tjcafferkey: https://github.com/woocommerce/woocommerce/pull/57361/commits/d01a718e218b7fc355e01ff2c9b67a8029e57f9d

plugins/woocommerce/client/blocks/assets/js/base/utils/create-notice.ts
tjcafferkey: I've implemented some of this feedback here https://github.com/woocommerce/woocommerce/pull/57361/commits/4d7987205b9f1039fadc7867a013bbe56be69f25 however I did leave the `context` as an optional param since then it mirrors the `removeNotice` signature and is preferred usage if you do know the context ahead of time.

PR #57370, opened 2025-04-17
Add a setting to WC -> Settings -> General to enable address autocomplete

opr:
This PR is based on: https://github.com/woocommerce/woocommerce/pull/56927

- Adds a setting to WC -> Settings -> General to enable address autocomplete
- If more than one provider is enabled, also register a select box to select the preferred one
- Add unit tests to cover changing settings and getting preferred provider.

Closes WOOPLUG-3614

| Before | After |
| ------ | ----- |
| image | image |

https://www.loom.com/share/0a7eb2885c6648a48fbec4ce10f2abc7

**Run `composer i` before testing**

1. Go to WC -> Settings -> General, observe the `Enable predictive address search` setting. It should be disabled with some helper text saying: `Suggest full addresses for customer as they type. To use this feature, you need to install an address provider such as WooPayments.`
2. Using [Code Snippets](https://wordpress.org/plugins/code-snippets/), add a PHP Snippet and enter:

```php
class WP_Provider extends WC_Address_Provider {
	public $id;
	public $name;
	
		$this->id   = 'wp-provider';
		$this->name = 'WordPress provider';
	}
};


add_filter( 'woocommerce_address_providers', function( $providers ) {
	$providers[] = 'WP_Provider';
	return $providers;
} );
```

3. Refresh the settings page and ensure the checkbox is enabled and can be checked and unchecked. Save the settings page with it checked and refresh the page. Ensure it stays checked on reload. Do the same with it unchecked.
5. Add another provider using the same snippet above, add a new class and return it in the filter like so:

```php
class WP_Provider extends WC_Address_Provider {
	public $id;
	public $name;
	
		$this->id   = 'wp-provider';
		$this->name = 'WordPress provider';
	}
};
class WC_Provider extends WC_Address_Provider {
	public $id;
	public $name;
	
		$this->id   = 'woo-provider';
		$this->name = 'Woo provider';
	}
};


add_filter( 'woocommerce_address_providers', function( $providers ) {
	$providers[] = 'WP_Provider';
	$providers[] = 'WC_Provider';
	return $providers;
} );
```

6. Refresh the setting page and check the `Enable predictive address search` box.
7. Ensure a `Preferred address autocomplete provider` dropdown menu appears with both providers listed.
8. Select one and save the page. Refresh the page and ensure the same one remains selected.
9. Try again with a different provider, ensure the correct one stays selected after a reload.
10. Uncheck the `Enable predictive address search` and ensure the dropdown menu is hidden.
11. Re-check the checkbox and ensure the dropdown menu reappears.

- Unit tests written to cover different settings scenarios
- Manual testing carried out as per instructions, see screen recording for steps taken.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


tjcafferkey (approved):
This tests well, and the code LGTM to me as well.

plugins/woocommerce/includes/admin/settings/class-wc-settings-general.php
```diff
@@ -46,6 +49,59 @@ protected function get_settings_for_default_section() {
 			$currency_code_options[ $code ] = $name . ' (' . get_woocommerce_currency_symbol( $code ) . ') — ' . esc_html( $code );
 		}
 
+		$enable_autocomplete_setting             = array();
+		$autocomplete_preferred_provider_setting = array();
+		$autocomplete_desc_tip                   = __( 'Suggest full addresses for customer as they type.', 'woocommerce' );
+
+		if ( Features::is_enabled( 'experimental-blocks' ) ) {
+			// This is in a try because getting the class from the container may fail if the class is not available.
```
ralucaStan, line 57: what would be a case where the class would not be available?

plugins/woocommerce/includes/admin/settings/class-wc-settings-general.php
```diff
@@ -188,6 +244,10 @@ protected function get_settings_for_default_section() {
 					),
 				),
 
+				$enable_autocomplete_setting,
```
ralucaStan: As someone without context, looking at `enable_autocomplete_setting` the name is vague. How do you feel about being explicit and using `address_autocomplete` instead of `autocomplete` in both the variables.

plugins/woocommerce/includes/admin/settings/class-wc-settings-general.php
```diff
@@ -298,6 +358,14 @@ protected function get_settings_for_default_section() {
 				),
 			);
 
+		// Remove any empty items from settings array.
+		// e.g. The preferred autocomplete provider setting would be empty if <=1 providers are registered.
+		$settings = array_filter(
```
ralucaStan, line 364: Wouldn't it be more performant to remove the empty value for `autocomplete_preferred_provider_setting`? Also, I'm not sure if there is an impact on other settings here being account as empty.

plugins/woocommerce/src/Internal/AddressProvider/AddressProviderController.php
```diff
@@ -117,6 +117,22 @@ public function get_registered_providers(): array {
 			$providers[] = $provider_instance;
 		}
 
+			// Check if there's a preferred provider set in the settings.
+		$preferred_provider_id = get_option( 'woocommerce_address_autocomplete_provider', '' );
```
ralucaStan: I'd wrap this whole block in a if block `! empty( $providers )`:
- execute only of there are providers set up
- keep preferred_provider_id in the if scope
```
if (! empty( $providers )) {
$preferred_provider_id = get_option( 'woocommerce_address_autocomplete_provider', '' );
if ( ! empty( $preferred_provider_id ) ) { ...}

}
```

ralucaStan: I see you have a `get_preferred_provider` function. Could you reuse it here and test against the empty string to executed your logic when you have a providers? The function already checks for empty providers etc

ralucaStan (changes requested):
I tested the PR with the testing steps provided and added some feedback and some questions to gain more context. Thank you Thomas, feel free to assign me to feature PRs related to the feature.

NickGreen:
Love this!

Any chance you can clarify what this feature will mean for the end user/store manager? Would this enable Autocomplete at Woo checkout without extra custom development? Or is it perhaps just a way for developers to hook into block checkout to add their own autocomplete solution?

opr:
Hey @NickGreen thanks for asking! This single PR is part of a larger project that will allow third party plugins to register an "Address Provider" with WooCommerce. This "Address Provider" would do tasks such as looking up address autocomplete matches and returning them to be displayed in the checkout, and looking up map locations to show on the order confirmation screen.

This feature is a work in progress and is only available on experimental builds of WooCommerce (i.e. running locally in dev mode). As the project progresses you can expect documentation and feature highlights to be showcase the feature.

ResponsiveFlow:
> allow third party plugins


So Woo doesn't plan to have something by default, maybe with WooPayments?

opr:
> > allow third party plugins
> 
> So Woo doesn't plan to have something by default, maybe with WooPayments?

Indeed the plan is to provide this functionality through WooPayments to begin with. The API will be public anyway so I expect many other plugins to implement address providers

plugins/woocommerce/includes/admin/settings/class-wc-settings-general.php
opr: I'm not able to think of one off the top of my head, but `wc_get_container()->get` throws so it's better to catch that, rather than letting it error.

plugins/woocommerce/includes/admin/settings/class-wc-settings-general.php
opr: I updated all variable names here to be more explicit

plugins/woocommerce/includes/admin/settings/class-wc-settings-general.php
opr: Yes, it would be more performant, but: This is an indexed array, so we can't remove a single item unless we know its index, we can specifically remove the item at index _n_ if it's empty, but that is very risky and fragile code, if someone else updates the code in future to add a new setting before index _n_ that would no longer work.

> Also, I'm not sure if there is an impact on other settings here being account as empty.

No other settings are empty, we can see the other items are added to the array unconditionally. I also don't think it'd be a problem to remove them in future if they were added conditionally and therefore became empty.

An alternative approach would be to only insert the new settings if not empty, and insert them specifically after the `woocommerce_default_customer_address` key, but that would fall into a similar problem making the code harder to modify in future.

In my opinion this is still the most readable, and most robust way of handling this.

plugins/woocommerce/src/Internal/AddressProvider/AddressProviderController.php
opr: > I see you have a `get_preferred_provider` function. Could you reuse it here and test against the empty string to executed your logic when you have a providers? The function already checks for empty providers etc

`get_preferred_provider` calls `get_registered_providers` so it would be an infinite loop.

ralucaStan: My thoughts around the two functions are that
- there is some code duplication: 
     - ` get_option('woocommerce_address_autocomplete_provider', '')` is called in both functions. Could we store this somewhere?
     - we check for empty( $providers ) in both places
- the `get_registered_providers` does a couple of things; the sorting based on the preferred provider logic is one of its responsibilities. How do you feel about extracting that in a separate function?
- could we work around the  infinite loop by passing the providers otherwise? This would enable us to have a dedicated function to get the  preferred provider logic, which we could use for the sorting logic.

ralucaStan (commented):
asked a couple of questions around the preferred provider logic. let me know what you think

opr:
@ralucaStan thanks for the feedback. I've removed the cache because now the filter only runs on `init` so caching is not necessary. I also get the option for preferred provider once on init and store it in a class variable.

I've made a getter function to get the registered providers, as well as simplifying the getter for preferred provider. I do think this approach is better, caching and comparing address providers is an overengineered solution now i've thought about it in more detail.

ralucaStan (approved):
Approving! 
I've followed the testing steps after the changes to the AddressProviderController class and things work as expected. 
Thank you for your patience, Thomas!

PR #57376, opened 2025-04-17
Use PHP to check for theme template parts in Add to Cart with Options block

### LINKED ISSUE #57081: Blockified Add to Cart with Options: avoid 404 request to determine if the theme has specific template part
Highlighted by @dinhtungdu in [https://github.com/woocommerce/woocommerce/pull/57069#pullrequestreview-2748589765](https://github.com/woocommerce/woocommerce/pull/57069#pullrequestreview-2748589765).

In the Blockified Add to Cart with Options block, we make a request to know if the theme has a specific template part, this way we know if we need to display the theme template part or the default WooCommerce template part:

[https://github.com/woocommerce/woocommerce/blob/56202577f719bc7bd09366f5e4510edb9a178558/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/edit/edit-template-part.tsx#L91-L109](https://github.com/woocommerce/woocommerce/blob/56202577f719bc7bd09366f5e4510edb9a178558/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/edit/edit-template-part.tsx#L91-L109)

This issue is about exploring better approaches for this that don't require making a request that might return 404. One option would be to check this on the PHP side and pass it with the Asset Data Registry so it can be accessed from the editor.

amitraj2203:

- This PR addresses the issue of unnecessary 404 requests made by the Blockified `Add to Cart with Options` block when determining if a theme has specific template parts. 
- Instead of making requests client side that might return 404s, this PR implements a server-side check using PHP and passes template part paths through the `Asset Data Registry` so they can be accessed from the editor.

Closes #57081 

1. Install and activate the `WooCommerce Beta Tester plugin`
2. Go to *WooCommerce Admin Test Helper > Features* and enable `experimental-blocks` and `blockified-add-to-cart`
3. Go to *Appearance > Editor > Templates > Single Product*, select the `Add to Cart with Options` block and update it to the blockified one
4. Open dev tools and check the network tab - Look for any network requests to `/wp-json/wp/v2/template-parts/` endpoints or any 404 responses that might indicate template part checking. Verify none exist.
5. Switch between product types (Simple, Variable, External, Grouped) while making some edits and confirm no template part check requests are made
6. Add a file to your theme under `parts/simple-product-add-to-cart-with-options.html` with the following content:
   ```html
   <!-- wp:paragraph -->
   <p>This is a Simple Product template from the theme.</p>
   <!-- /wp:paragraph -->
   <!-- wp:woocommerce/product-button {"textAlign":"center","isDescendentOfQueryLoop":true,"fontSize":"small"} /-->
   ```
7. Go to _Appearance > Editor > Templates > Single Product_ again
8. Verify when you select the Simple product type, the template from the theme is loaded, so the `This is a Simple Product template from the theme.` paragraph block is visible.
9. Verify again by opening dev tools and checking the network tab that there are no requests being sent to check if a product template part exists, especially when you select the Simple product type
10. Confirm that the theme's template part is correctly used for the Simple product type and the default template is used for other product types

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


kmanijak (commented):
Thanks for your effort @amitraj2203! Solution looks robust!

It tests well:
- no extra calls resulting in 404 are made which is a nice improvement
- when there's theme's template part overriding the default template part, it's fetched correctly.

I see you also took into account @Aljullu's [comment](https://github.com/woocommerce/woocommerce/issues/57081#issuecomment-2812323254) to pass actual template part's paths instead of booleans. For now they're not utilised and wanted to confirm @Aljullu you suggested it as a preparation for future work, not a part of this PR, right? 🙌 You have more context on it so wanted to double check.

Once confirmed by Albert, I'm happy to approve.

prettyboymp (approved):
From a performance standpoint, these changes look great.  Thank you!  I'm good with these changes as long as  @Aljullu agrees with the direction it is taking for front end.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/edit/edit-template-part.tsx
```diff
@@ -88,33 +89,18 @@ export const AddToCartWithOptionsEditTemplatePart = ( {
...
+				productType in themeTemplatePartPaths &&
+				themeTemplatePartPaths[ productType ] !== null;
 
 			return {
-				templatePartId: themeTemplatePartIsMissing
-					? wooCommerceTemplatePartId
-					: themeTemplatePartId,
+				templatePartId: themeTemplatePartExists
+					? themeTemplatePartId
+					: wooCommerceTemplatePartId,
```
Aljullu: My bad in [my comment](https://github.com/woocommerce/woocommerce/issues/57081#issuecomment-2812323254) I said "template part path", I should have said "template part id". (I got mixed up with a [different PR](https://github.com/woocommerce/woocommerce/pull/56799) :man_facepalming: )

So I wonder if we can pass the ids for all template parts from the PHP file, that would include the WooCommerce core and the theme ones. So in this file we would only need to access it and could remove the `themeTemplatePartExists` check.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions.php
```diff
@@ -31,6 +31,23 @@ protected function enqueue_data( array $attributes = [] ) {
 		parent::enqueue_data( $attributes );
 		$this->asset_data_registry->add( 'isBlockifiedAddToCart', Features::is_enabled( 'blockified-add-to-cart' ) );
 		$this->asset_data_registry->add( 'productTypes', wc_get_product_types() );
+		$this->asset_data_registry->add( 'themeTemplatePartPaths', $this->get_theme_template_part_paths() );
```
Aljullu: Besides changing `paths` for `ids` in this variable, I would also mention the block name. Something like:

```suggestion
		$this->asset_data_registry->add( 'addToCartWithOptionsTemplatePartIds', $this->get_theme_template_part_paths() );
```

Aljullu (commented):
Thanks for working on this, @amitraj2203! Overall, the approach looks great! As @kmanijak mentioned, I think we should use the variable we pass from the backend directly in the editor component. We might need to pass the template parts IDs instead of the paths. I left a comment inline: :point_down:

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/edit/edit-template-part.tsx
amitraj2203: Thanks for the review @Aljullu, I tried to do it that way now, can you please check?

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/edit/edit-template-part.tsx
```diff
@@ -68,14 +69,14 @@ export const AddToCartWithOptionsEditTemplatePart = ( {
...
+
 	const { templatePartId } = useSelect(
 		( select ) => {
-			const {
-				getCurrentTheme,
-				getEditedEntityRecord,
-				hasFinishedResolution,
-			} = select( coreStore );
-
+			const { getCurrentTheme } = select( coreStore );
```
Aljullu: Given that we aren't using the `currentTheme` variable anymore, I think we can remove it. In fact, maybe we can even remove the whole `useSelect()` and simple do:

```TSX
const templatePartId = addToCartWithOptionsTemplatePartIds?.[ productType ];
```

Aljullu (commented):
Thanks for the updates, @amitraj2203! Overall, it's looking good on my end. I just left one comment about some additional clean up that I think we can do now.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/edit/edit-template-part.tsx
amitraj2203: Thanks for the feedback, I've done it now.

Aljullu (approved):
Thanks for the update, @amitraj2203! I just re-tested again and confirming there is no longer a 404 request and the theme template part show up as expected:

Editor | Frontend
--- | ---
Captura de pantalla de 2025-04-22 15-18-03 | Captura de pantalla de 2025-04-22 15-18-22

<details>

* Theme: Twenty Twenty-Five
* Store size: small
</details>

PR #57384, opened 2025-04-18
[Blueprint] Implement logging for export and import operations

chihsuan:

Closes WOOPLUG-3946

This PR adds logging for the Blueprint package to improve traceability and debugging capabilities during export and import operations. This aids in traceability and incident response.

The following things are logged:

1. User ID - Who is doing the export/import
2. Step names - Which steps are being exported/imported
3. Step classes - Which classes are being used to export/import the steps (which is useful to know the detail actions taken)
4. success or error - Whether the export/import was successful or not

### Key Changes

- Add new `Logger` class for centralized logging
- Add detailed logging for export operations
- Add detailed logging for import operations

1. pnpm install
2. Enable the Blueprint feature `wp option set woocommerce_feature_blueprint_enabled yes`
3. Go to `/wp-admin/admin.php?page=wc-settings&tab=advanced&section=blueprint` and export a blueprint
4. Go to WooCommerce > Status > Logs and check the logs for the Blueprint logs
5. Confirm that it logs the export steps
6. Import the blueprint
7. Confirm that it logs the import steps

Screenshot 2025-04-18 at 14 48 41
Screenshot 2025-04-18 at 14 47 42

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


packages/php/blueprint/src/ImportStep.php
```diff
@@ -80,33 +82,80 @@ function ( $key, $importer ) {
...
+			$result->add_error( 'Unable to find an importer' );
+			return false;
+		}
+
+		$importer = $this->indexed_importers[ $this->step_definition->step ];
+		// Validate importer is a step processor before processing.
 		if ( ! $importer instanceof StepProcessor ) {
-			$result->add_warn( "Importer {$this->step_definition->step} is not a valid step processor" );
-			return $result;
+			$result->add_error( 'Incorrect importer type' );
```
louwie17, line 122: Any specific reason for changing this and the one above to `error` instead of `warn`?

packages/php/blueprint/src/ImportStep.php
```diff
@@ -80,33 +82,80 @@ function ( $key, $importer ) {
...
+	}
+
+	/**
+	 * Check if the step can be imported.
+	 *
+	 * @param StepProcessorResult $result The result object to add messages to.
+	 *
+	 * @return bool True if the step can be imported, false otherwise.
+	 */
+	protected function can_import( &$result ) {
```
louwie17, line 112: Should we log these errors as well, if the user can't import for example?

packages/php/blueprint/src/ExportSchema.php
```diff
@@ -93,26 +96,57 @@ function ( $exporter ) {
...
+		$exporters_classes = array();
 		// Make sure the user has the required capabilities to export the steps.
 		foreach ( $exporters as $exporter ) {
 			if ( ! $exporter->check_step_capabilities() ) {
 				return new WP_Error( 'wooblueprint_insufficient_permissions', 'Insufficient permissions to export for step: ' . $exporter->get_step_name() );
 			}
+
+			$step_name           = $exporter instanceof HasAlias ? $exporter->get_alias() : $exporter->get_step_name();
+			$export_steps[]      = $step_name;
+			$exporters_classes[] = get_class( $exporter );
```
louwie17: I think we should move this stuff to the `Logger` class.
For example, we could have a `$logger->start_export` function that takes the list of `$exporters`. This would avoid us having to explicitly getting the name and class name here, and instead move that to the Logger class.
We could have similar functions for `export_step_success` and `start_import` etc...

louwie17 (commented):
This tested quite well, thanks! I did notice if the user doesn't have permissions it doesn't log anything.
I also left several inline comments, let me know what you think about them?
Lastly, I kinda wonder if we should do the same thing for the export as we did for the import, where we do log when each step is done, instead of a single log line for all of the Export steps.

packages/php/blueprint/src/ImportStep.php
chihsuan: I realized that using `warn` would be problematic for this case because it wouldn't be visible in the UI, and the import would be considered successful, but it wouldn't import anything.

packages/php/blueprint/src/ExportSchema.php
chihsuan: Sounds good! I'll move this to the Logger class.

packages/php/blueprint/src/ImportStep.php
chihsuan: I think additional error logging is unnecessary, as users receive error messages in UI or CLI with `--show-messages` when imports fail, which should be a better experience. Logging is the primary focus of security-related events. Let me know if you don't see any errors when permissions are not allowed. 🙏

chihsuan:
> Lastly, I kinda wonder if we should do the same thing for the export as we did for the import, where we do log when each step is done, instead of a single log line for all of the Export steps.

I had the same thought about logging individual export steps. However, I found that logging each export step created excessive verbosity in the logs, making them harder to read. Unlike imports, which import data at each step, exports only complete when all steps succeed. Therefore, a single log entry at the end should be sufficient. Let me know if you think otherwise. 🙏

packages/php/blueprint/src/ExportSchema.php
chihsuan: Updated ExportSchema and ImportStep to use new logging methods in the Logger class in 4d0c0784e68ecc9454860abbf07ceac546ddab19

packages/php/blueprint/src/ImportStep.php
chihsuan: Ah, there is a known issue that errors are not shown correctly. It's fixed in https://github.com/woocommerce/woocommerce/pull/57336.

packages/php/blueprint/src/ExportSchema.php
```diff
@@ -100,19 +102,24 @@ function ( $exporter ) {
...
 
-			if ( is_array( $step ) ) {
-				foreach ( $step as $_step ) {
-					$schema['steps'][] = $_step->get_json_array();
-				}
-			} else {
-				$schema['steps'][] = $step->get_json_array();
+			} catch ( \Throwable $e ) {
+				$step_name = $exporter instanceof HasAlias ? $exporter->get_alias() : $exporter->get_step_name();
+				$logger->export_step_failed( $step_name, $e );
```
louwie17, line 116: May be nice to just pass the `$exporter` to the `step_failed` and also move the `$step_name` logic to the Logger.
This is small though, so not a blocker to merge.

louwie17 (approved):
Nice work on this and thanks for the updates!
This tested well and code looks good! I left one tiny comment, but not a blocker to merge.

Issue #54730, opened 2025-01-21
[Product Details Block] Only render item when relevant content is available


gigitux:
The "Additional Information" and "Reviews" sections within the Product Details block should only render when relevant content is available. Currently, patterns do not support conditional logic, but a block implementation can address this limitation.

#### Technical Considerations
The following logic from WooCommerce determines whether these sections should render:

```php
if ( ( $product instanceof WC_Product ) && ( $product->has_attributes() || apply_filters( 'wc_product_enable_dimensions_display', $product->has_weight() || $product->has_dimensions() ) ) ) { 
	$tabs['additional_information'] = array( 
		'title'    => __( 'Additional information', 'woocommerce' ), 
		'priority' => 20, 
		'callback' => 'woocommerce_product_additional_information_tab', 
	); 
} 

// Reviews tab - shows comments. 
if ( comments_open() ) { 
	$tabs['reviews'] = array( 
		/* translators: %s: reviews count */ 
		'title'    => sprintf( __( 'Reviews (%d)', 'woocommerce' ), $product->get_review_count() ), 
		'priority' => 30, 
		'callback' => 'comments_template', 
	); 
}
```

#### Solution
1. Implement logic within the Product Details block to evaluate the following conditions:
   - Does the product have attributes, weight, or dimensions?
   - Are reviews enabled and is there review content?

2. Based on these evaluations:
   - Hide the Accordion containing "Additional Information" or "Reviews" blocks if they do not produce content.

3. Inform merchants within the editor UI when certain sections may not be rendered due to the absence of content.

#### Editor Behavior
- Display a notice or placeholder message in the editor for missing content.

PR #57387, opened 2025-04-18
Product Specifications: Add settings to control the visibility of weight/dimensions/attribute

### LINKED ISSUE #57111: Add conditional options for displaying weight, dimensions, and attributes
Lourens shared on [**New Block: Product Attributes block to display weight, dimensions, and attributes**](https://github.com/woocommerce/woocommerce/pull/56621/files#top)**:**

> We should add an attribute or multiple attributes to configure the display of dimensions and weight.
>
> In my mind a user should be able to display either weight, dimensions, or attributes by themselves.
>
> Display them all together if they exist, or mix-match.
>
> For example I could easily see a scenario where a user wants to display the dimensions separately from the attributes, they may actually create an accordion just for dimensions and weight.
>
> Displaying them always together doesn't allow for this.
>
> Making it configurable, allows users to re-use this block in multiple places, but display either attributes in one place and dimensions and weight in another.

dinhtungdu:

This PR adds conditional display options for weight, dimensions, and attributes in the Product Specifications block. Users can now selectively show/hide these elements through the block's settings panel.

Closes #57111.

| Before | After |
| ------ | ----- |
|     image   |    image   |

1. Edit the Single Product template.
2. Add the "Blockified Product Details" block to the template.
3. Select the "Product Specifications" block and open block settings.
4. Verify that there are three new toggle controls under "Display Settings":
    - Show Weight
    - Show Dimensions
    - Show Attributes
5. Test each toggle control:
    - Toggle "Show Weight" off and verify the weight information disappears
    - Toggle "Show Dimensions" off and verify the dimensions information disappears
    - Toggle "Show Attributes" off and verify all product attributes disappear
6. Toggle multiple controls and verify combinations work correctly.
7. Save the page and view it on the frontend.
8. Verify that the display settings persist and work correctly on the frontend.
9. Note that when all settings are toggled off, the block is empty but the Additional Information accordion items are still visible. This is a known issue and belongs to https://github.com/woocommerce/woocommerce/issues/54730 scope.

### DIFF
--- a/plugins/woocommerce/changelog/wooplug-3844-add-conditional-options-for-displaying-weight-dimensions-and
+++ b/plugins/woocommerce/changelog/wooplug-3844-add-conditional-options-for-displaying-weight-dimensions-and
@@ -1,4 +1,4 @@
 Significance: patch
 Type: add
 
-Add settings to control the visibility of weight/dimensions/attribute.
+Product Specifications: Add settings to control the visibility of weight/dimensions/attribute.
--- a/plugins/woocommerce/changelog/wooplug-3844-add-conditional-options-for-displaying-weight-dimensions-and
+++ b/plugins/woocommerce/changelog/wooplug-3844-add-conditional-options-for-displaying-weight-dimensions-and
@@ -1,0 +1,4 @@
+Significance: patch
+Type: add
+
+Add settings to control the visibility of weight/dimensions/attribute.
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/product-specifications/block.json
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/product-specifications/block.json
@@ -14,10 +14,6 @@
 		"core/post-template"
 	],
 	"attributes": {
-		"fontSize": {
-			"type": "string",
-			"default": "medium"
-		},
 		"showWeight": {
 			"type": "boolean",
 			"default": true
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/product-specifications/block.json
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/product-specifications/block.json
@@ -17,6 +17,18 @@
 		"fontSize": {
 			"type": "string",
 			"default": "medium"
+		},
+		"showWeight": {
+			"type": "boolean",
+			"default": true
+		},
+		"showDimensions": {
+			"type": "boolean",
+			"default": true
+		},
+		"showAttributes": {
+			"type": "boolean",
+			"default": true
 		}
 	},
 	"usesContext": [ "postId", "postType" ],
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/product-specifications/edit.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/product-specifications/edit.tsx
@@ -1,11 +1,12 @@
 /**
  * External dependencies
  */
-import { useBlockProps } from '@wordpress/block-editor';
+import { useBlockProps, InspectorControls } from '@wordpress/block-editor';
 import { __ } from '@wordpress/i18n';
 import { useQueryLoopProductContextValidation } from '@woocommerce/base-hooks';
 import { useSelect } from '@wordpress/data';
 import { optionsStore, Product, productsStore } from '@woocommerce/data';
+import { PanelBody, ToggleControl } from '@wordpress/components';
 
 /**
  * Internal dependencies
@@ -39,7 +40,10 @@ const getFormattedDimensions = (
 const Edit = ( {
 	context: { postId, postType },
 	clientId,
+	attributes,
+	setAttributes,
 }: ProductSpecificationsEditProps ) => {
+	const { showWeight, showDimensions, showAttributes } = attributes;
 	const blockProps = useBlockProps( {
 		className: 'wc-block-product-specifications',
 	} );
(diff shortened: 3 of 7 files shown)


plugins/woocommerce/client/blocks/assets/js/blocks/product-specifications/edit.tsx
```diff
@@ -150,27 +154,72 @@ const Edit = ( {
...
+	if ( ! showDimensions ) {
+		productData.dimensions.value = '';
+	}
+	if ( ! showAttributes ) {
+		Object.entries( productData ).forEach( ( [ key, data ] ) => {
+			if ( key !== 'weight' && key !== 'dimensions' ) {
+				data.value = '';
+			}
+		} );
+	}
```
kmanijak: What about preventing assigning the values if they're supposed be an empty string? We could skip some logic there instead of creating values and then clearing them.

plugins/woocommerce/client/blocks/assets/js/blocks/product-specifications/types.ts
```diff
@@ -7,4 +7,12 @@ type Context = {
 	context: { postId: string; postType: string };
 };
 
-export type ProductSpecificationsEditProps = BlockEditProps< object > & Context;
+interface ProductSpecificationsAttributes {
+	fontSize?: string;
```
kmanijak: I don't think `fontSize` should be listed as an attribute. I see it's mentioned in the `block.json` but it's coming from `useBlockProps` thanks to `supports` declaration. And that means it's not part of `ProductSpecificationsEditProps`.

plugins/woocommerce/client/blocks/tests/js/jest.config.json
```diff
@@ -9,6 +9,7 @@
 	"moduleDirectories": [ "node_modules" ],
 	"moduleNameMapper": {
 		"@wordpress/private-apis": "<rootDir>/node_modules/@wordpress/private-apis",
+		"@wordpress/core-data/build/(.*)$": "<rootDir>/node_modules/@wordpress/core-data/build/$1",
```
kmanijak, line 12: What this line does? 🤔

plugins/woocommerce/client/blocks/assets/js/blocks/product-specifications/block.json
```diff
@@ -17,6 +17,18 @@
 		"fontSize": {
 			"type": "string",
 			"default": "medium"
```
kmanijak: As mentioned below, `fontSize` should be removed from here as it's not block's attribute, it's only injected via `useBlockProps` thanks to `supports`declaration.

plugins/woocommerce/changelog/wooplug-3844-add-conditional-options-for-displaying-weight-dimensions-and
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: add
+
+Add settings to control the visibility of weight/dimensions/attribute.
```
kmanijak: Worth adding some context. Changelog is WooCommerce-wide so it might be hard to get where was it added 😄 

```suggestion
Product Specifications: Add settings to control the visibility of weight/dimensions/attribute.
```

plugins/woocommerce/client/blocks/assets/js/blocks/product-specifications/test/block.ts
```diff
@@ -0,0 +1,122 @@
...
+				} )
+			);
+
+			// Verify all sections are visible again
+			expect( block.getByText( /Weight/i ) ).toBeInTheDocument();
+			expect( block.getByText( /Dimensions/i ) ).toBeInTheDocument();
+			expect( block.getByText( /Test Attribute/i ) ).toBeInTheDocument();
+		} );
+	} );
+} );
```
kmanijak, line 122: Cool integration tests! <3

kmanijak (changes requested):
It tests well!

I played around with multiple combinations of show/hide particular elements and they behave as expected in the editor and frontend.

I left couple of comments regarding code here and there.

plugins/woocommerce/client/blocks/assets/js/blocks/product-specifications/edit.tsx
dinhtungdu: I did it to avoid nested conditions and repeated check. For example, if we check for the setting early, we end up with something like this, which is harder to read IMO. I can see that clearing value at the end also can create some confusion, so I'm open to change this logic for a more straightforward one.

```ts
	const productData: Record< string, { label: string; value: string } > = {};

	if ( showWeight ) {
		productData.weight = {
			label: __( 'Weight', 'woocommerce' ),
			value: '',
		};
	}

	if ( showDimensions ) {
		productData.dimensions = {
			label: __( 'Dimensions', 'woocommerce' ),
			value: '',
		};
	}

	if ( isSpecificProductContext ) {
		if ( showWeight && product?.weight ) {
			productData.weight.value = `${ product.weight } ${ weightUnit }`;
		}
		if ( showDimensions && product?.dimensions ) {
			productData.dimensions.value = getFormattedDimensions(
				product.dimensions,
				dimensionUnit
			);
		}
		if ( showAttributes && product?.attributes ) {
			product.attributes.forEach( ( attribute ) => {
				productData[ attribute.name.toLowerCase() ] = {
					label: attribute.name,
					value: attribute.options.join( ', ' ),
				};
			} );
		}
	} else {
		if ( showWeight ) {
			productData.weight.value = `10 ${ weightUnit }`;
		}
		if ( showDimensions ) {
			productData.dimensions.value = `10 × 10 × 10 ${ dimensionUnit }`;
		}
		if ( showAttributes ) {
			productData.test_attribute = {
				label: __( 'Test Attribute', 'woocommerce' ),
				value: __( 'First, Second, Third', 'woocommerce' ),
			};
		}
	}
```

plugins/woocommerce/client/blocks/assets/js/blocks/product-specifications/types.ts
dinhtungdu: Yep, good catch, we should remove `fontSize`.

plugins/woocommerce/client/blocks/tests/js/jest.config.json
dinhtungdu: It's added to support this import:
https://github.com/woocommerce/woocommerce/blob/146d95902fa9b131f5970e0ddb1ee6b36c1d7aa6/packages/js/data/src/setting-options/actions.ts#L10

plugins/woocommerce/changelog/wooplug-3844-add-conditional-options-for-displaying-weight-dimensions-and
dinhtungdu: Right, we need more context. Thank you!

plugins/woocommerce/client/blocks/assets/js/blocks/product-specifications/edit.tsx
kmanijak: I see what you mean but I'd rather gather the whole data assignment to something like this:

```ts

	const productData: Record< string, { label: string; value: string } > = {};

	if ( showWeight ) {
		productData.weight.label = __( 'Weight', 'woocommerce' );
		productData.weight.value = isSpecificProductContext && product.weight
			? `${ product.weight } ${ weightUnit }`
			: `10 ${ weightUnit }`;
	}

	if ( showDimensions ) {
		productData.weight.label = __( 'Dimensions', 'woocommerce' );
		productData.weight.value = isSpecificProductContext && product.dimensions
			? getFormattedDimensions(
					product.dimensions,
					dimensionUnit
			  )
			: `10 × 10 × 10 ${ dimensionUnit }`;
	}

	if ( showAttributes ) {
                if ( isSpecificProductContext && product.attributes ) {
			product.attributes.forEach( ( attribute ) => {
				productData[ attribute.name.toLowerCase() ] = {
					label: attribute.name,
					value: attribute.options.join( ', ' ),
				};
			} );
		} else {
			productData.test_attribute = {
				label: __( 'Test Attribute', 'woocommerce' ),
				value: __( 'First, Second, Third', 'woocommerce' ),
			};
		}
	}
```

I don't know if you agree but in that case no logic is run unnecessarily so `showX` determines if we even trigger any logic related to attribute and then, there's no data reassignment. 

I'm not precious about this change but for me it's much easier to read such code when you start from "global" checks (`showATTRIBUTE`) and go deeper to decide specifics `isSpecificProductContext`.

Let me know what you think! 🙌

dinhtungdu: Yeah, this is easier to read! I will update the PR following your suggestion. Thank you!

dinhtungdu: @kmanijak I ended up with a slightly different version in [2525c65](https://github.com/woocommerce/woocommerce/pull/57387/commits/2525c65b5f9d34ace7f5fc8a33f401c6733dbf52) due to the requirement to hide the item when the product context is available but no corresponding spec is set for the product. For example adding the block to the Single Product block, but the selected product doesn't have any weight/dimensions/attribute set. We still need three levels nested condition, but it still easier to read than my original version.

kmanijak: Ah, I see! That makes sense! But anyway, appreciate the responsiveness and willingness to adjust it 🙇

plugins/woocommerce/client/blocks/assets/js/blocks/product-specifications/edit.tsx
```diff
@@ -117,60 +121,113 @@ const Edit = ( {
...
+				? `${ product.weight } ${ weightUnit }`
+				: '';
+		} else {
+			productData.weight.value = `10 ${ weightUnit }`;
+		}
+
+		productData.weight.value =
+			isSpecificProductContext && product.weight
+				? `${ product.weight } ${ weightUnit }`
+				: `10 ${ weightUnit }`;
```
kmanijak: Seems like leftover and duplication of lines 132-138 (without the case of empty string when there's no product data)

```suggestion
```

kmanijak (approved):
Great job, thanks for addressing all my comments!

I tested:
- multiple combinations of on/off toggles
- on multiple products with and without weight/dimensions/custom attributes
- within Single Product block and Single Product Template

There's just one case in which default `10 ${ weightUnit }` is displayed for specific product when there's no weight defined but it's due to code leftover (see comment). Otherwise, it's all good so please remove the leftover and I'm pre-approving! 🚀

PR #23687, opened 2019-05-13
Delete guest cart session if user logs in, solves #23686

tatusah:
### All Submissions:

* [x] Have you followed the [WooCommerce Contributing guideline](https://github.com/woocommerce/woocommerce/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This code changes the session handler behaviour so that if a guest user who has products in cart later logs in the cart content is moved to the now logged in user user session and the guest session is deleted after login. Previously in woocommerce_sessions table would exist both the guest session and the logged in session. This would cause confusion as the same cart items would show up twice in calculations where one would check cart content from different sessions.

Closes #23686 .

### How to test the changes in this Pull Request:

1. Add products to cart while not logged in to user account. Check woocommerce_sessions table which shows the guest user session
2. Now log in to the user account
3. View the woocommerce_sessions table to see that now the guest session is removed and only logged in user session exists. Cart content has carried over correctly. Without this fix there would exist both the guest and logged in session.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully run tests with your changes locally?

### Changelog entry

Changes the session handler behavior to delete the guest session after successful login.

PR #57391, opened 2025-04-18
Avoid attempting to delete a non-existent session while logged in.

prettyboymp:

This adds a check to `WC_Session_Handler::save_data()` to verify that `old_session_key` isn't empty before attempting to delete the old session.

(For Bug Fixes) Bug introduced in PR #23687 .

PR #23687 updated `WC_Session_Handler::save_data()` to delete the previous guest session after a user logs in and their session is saved under their user account. However, the logic to delete the old session was added to `::save_data()` instead of `::init_session_cookie()`, where the transition from guest to logged-in user is actually detected.

Since `::save_data()` is hooked to the `shutdown` action which is called at the end of every request, this results in unnecessary calls to `WC_Session_Handler::delete_session( '' )` (with an empty string for $old_session_key). As a result, an unneeded delete query is executed on every shutdown, even when there is no valid session to delete.

While I would prefer to remove this handling from within `::save_data()` and put it where we detect the transition from guest to logged in user, we are unable to do so without potentially breaking backward compatibility.  Some extensions, including WooCommerce Payments, extend and override the session handler and save_data method altering its behavior.  WooCommerce Payments goes even further so that `::delete_session( $customer_id )` ignores the passed in `$customer_id` and uses the current session_id which would result in the wrong session instance getting deleted.

1.  Setup: Because `WC_Session_Handler::save_data()` runs late on shutdown, Query Monitory won't display the queries included by it.  So adding error logging to `WC_Session_Handler::delete_session()` is the easiest way:
  ```php
  error_log( "deleting session: '{$customer_id}'");
  ```
2. As a non-logged in user, go to the store front and add an item to your cart in order to start a session.
3. Now, log in as a user.  When logging in, you should see the error log entry for `deleting session: '{some value}'`.
4. Browse other pages, you shouldn't see any more attempts to delete a session and none of the error log entries should be `deleting session: ''`, with an empty string.


plugins/woocommerce/includes/abstracts/abstract-wc-session.php
```diff
@@ -19,7 +19,7 @@ abstract class WC_Session {
 	/**
 	 * Customer ID.
 	 *
-	 * @var int $_customer_id Customer ID.
+	 * @var string $_customer_id Customer ID.
```
prettyboymp, line 22: I went ahead and updated this documentation as all the usage I've seen is always a string - though it is a numeric string when the user is logged in.

plugins/woocommerce/includes/class-wc-session-handler.php
```diff
@@ -363,9 +363,11 @@ private function get_cache_prefix() {
 	/**
 	 * Save data and delete guest session.
 	 *
-	 * @param int $old_session_key session ID before user logs in.
+	 * @param string|mixed $old_session_key Optional session ID prior to user log-in.  If $old_session_key is not tied
+	 *                                      to a user, the session will be deleted tied to that key will be deleted with the
```
albarin: I think there's some duplication in this comment, it says "will be deleted" twice.

albarin (approved):
✅ Confirmed I can see the error log after logging in.
✅ Confirmed I don't see any more logs after browsing other pages.

Just left a message about one of the comments, but pre-approving!

PR #57392, opened 2025-04-19
Fix a typo in class-wc-post-types.php

szepeviktor:
Noticed a misspelling while researching product-editor-template-system feature.


jorgeatorres (approved):
LGTM. Thanks @szepeviktor!

PR #57428, opened 2025-04-19
Add new database tables for fulfillments

### LINKED ISSUE #57356: [Order Fulfillments] Create DB tables for fulfillment and its meta
**Issue:** We need to store the fulfillment data to the local database to be able to display the fulfillments correctly on Woo core UI, and to be able to act correctly.

**Proposed solution:** The proposed DB schema can be found in [https://wooorderstatusp2.wordpress.com/2025/04/16/tsd-woocommerce-fulfillment-entity/#fulfillment-data](https://wooorderstatusp2.wordpress.com/2025/04/16/tsd-woocommerce-fulfillment-entity/#fulfillment-data)

Also a good read for having DB schema changes while developing on a feature branch: [https://protonpower.wordpress.com/2025/02/24/best-approach-for-adding-db-changes-to-core/](https://protonpower.wordpress.com/2025/02/24/best-approach-for-adding-db-changes-to-core/)

tpaksu:

Closes #57356 .

#### Schema check
1. Open the changes of this PR
2. Validate the SQL code added against common practice.

#### Manual trigger
1. Checkout this branch
3. Go to WooCommerce > Status > Tools
4. Click "Verify base database tables" > "Verify tables" action,  
5. Open your DB client and check if the tables were created.

#### Auto trigger
1. Delete the tables created by the previous test.
2. Go to plugins page.
3. Disable any plugins dependent on WooCommerce
4. Disable WooCommerce and re-enable it back
5. See that the tables are created again

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 
Add new database tables for storing the order fulfillment records.
</details>


plugins/woocommerce/includes/class-wc-install.php
```diff
@@ -1987,7 +1987,27 @@ private static function get_schema() {
 	PRIMARY KEY (category_tree_id,category_id)
 ) $collate;
 $hpos_table_schema;
-		";
+CREATE TABLE {$wpdb->prefix}wc_order_fulfillments (
+	fulfillment_id bigint(20) unsigned NOT NULL AUTO_INCREMENT,
+	entity_type varchar(255) NOT NULL,
+	entity_id bigint(20) unsigned NOT NULL,
+	date_created datetime NOT NULL default '0000-00-00 00:00:00',
```
prettyboymp: The specs show columns for updated date defaulting to CURRENT_TIMESTAMP and status, is that still planned?

tpaksu: Thanks for catching this, I overlooked that part while designing the schema. I'll update the table.

tpaksu: Actually, I thought a bit about it and I don't see a way that the data on the fulfillment table (not the meta table) will get updated once created. We can keep the `date_created` name on the field, but default it to CURRENT_TIMESTAMP. 

On the meta table, do we want to track only the updated date? Or would it be better for us to keep a history table for all changes?

prettyboymp: Won't a fulfillment transition through statuses and be updated along the way?

tpaksu: The status updates will be held on the metadata table. The `entity type` and `entity ID` fields will stay the same, for example for WC Core, they will be `WC_Order` and the order ID respectively.

prettyboymp: What is the reasoning behind having `status` stored as meta?  It seems like it would be a required property for every entry and more likely than not need to be queried against on occasion.

tpaksu: It doesn't have a concrete reason, but it's done like that for making it more flexible. 

Context: After some conversation about supporting custom fulfillment statuses, the current fulfillment status started to have of two components, a string based `status` meta, which holds the fulfillment status, which can be extended via plugins to include statuses other than `fulfilled` and `unfulfilled`, and a `fulfilled` boolean value to indicate if a custom fulfillment status will resolve to `fulfilled`or not. For example, a plugin may extend the fulfillment status to include `backorder` and `delivered` and those will resolve to fulfilled being `true` and `false` respectively. So I moved it to the meta table in case it has more components that should be taken into account when calculating a fulfillment status.

ismaeldcom (approved):
Testing instructions work as expected and the schema matches the one in the TSD except for the `status` column which is already mentioned in the other review comment.

plugins/woocommerce/includes/class-wc-install.php
prettyboymp: I would lean towards including it in the main fulfillments table.  If there are any attributes that are likely to be a part of every fulfillment record like the status, putting it in the main table as opposed to meta is preferred.  This is especially so if we may need to filter by that attribute in the future.

We may even need to consider whether the calculated fulfillment state (fulfilled/unfulfilled) is needed as its own column in the future, but I would hold off on that until we better understand the use cases and query performance without it.

tpaksu: Makes sense. I will move `status` and `is_fulfilled` metas as new columns on the `fulfillments` table.

tpaksu: Addressed in [9b76e9d](https://github.com/woocommerce/woocommerce/pull/57428/commits/9b76e9d45112947fa4637a9496564337df14f157).

prettyboymp (approved):
👍

PR #57429, opened 2025-04-19
Add order fulfillments data store

### LINKED ISSUE #57357: [Order Fulfillments] Create data store for fulfillment tables
**Issue:** We need a data repository for accessing the created tables for the fulfillment project.

Proposed solution: Introduce a Fulfillments Data Store class to handle all possible actions outlined in [https://wooorderstatusp2.wordpress.com/2025/04/16/tsd-woocommerce-fulfillment-entity/#fulfillment-data-store-methods](https://wooorderstatusp2.wordpress.com/2025/04/16/tsd-woocommerce-fulfillment-entity/#fulfillment-data-store-methods)

tpaksu:

Closes #57357  .

New code should make sense in terms of adding a new data store, and automated tests should pass.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 
Add order fulfillments data store to interact with the new fulfillment tables.
</details>


plugins/woocommerce/src/Internal/DataStores/Fulfillments/FulfillmentsDataStore.php
```diff
@@ -0,0 +1,401 @@
...
+	 * @param Fulfillment $data The fulfillment object to create.
+	 *
+	 * @return void
+	 */
+	public function create( &$data ) {
+		$this->clear_error();
+
+		// Validate the fulfillment data.
+		if ( ! $data->get_entity_type() ) {
+			$this->set_error( new WP_Error( 'invalid_entity_type', __( 'Invalid entity type.', 'woocommerce' ) ) );
```
prettyboymp: I'm a little concerned that we're introducing a different pattern here for this datastore where a client needs to check for errors explicitly after write commands.   It is probably safest to have an exception be thrown so execution doesn't continue with the assumption that everything went okay.

plugins/woocommerce/src/Internal/Fulfillments/Fulfillment.php
```diff
@@ -0,0 +1,200 @@
...
+	 */
+	public function get_id() {
+		return $this->fulfillment_id;
+	}
+	/**
+	 * Set the fulfillment ID.
+	 *
+	 * @param int $id Fulfillment ID.
+	 */
+	public function set_id( $id ) {
```
prettyboymp: Can we specify the parameter types and return types in code where possible (meaning the type constraint is valid in PHP 7.4)?

plugins/woocommerce/src/Internal/Fulfillments/Fulfillment.php
```diff
@@ -0,0 +1,200 @@
...
+		} elseif ( is_string( $data ) && ! empty( $data ) ) {
+			$this->set_id( absint( $data ) );
+		} elseif ( is_object( $data ) && isset( $data->id ) ) {
+			$this->set_id( absint( $data->id ) );
+		} else {
+			$this->set_object_read( true );
+		}
+
+		// Load the items array.
+		$this->data_store = new FulfillmentsDataStore();
```
prettyboymp: We'll want to be sure we use a shared datastore between entities so we don't end up with every entity pointing to a new datastore instance.

plugins/woocommerce/src/Internal/DataStores/Fulfillments/FulfillmentsDataStore.php
tpaksu: Yes, I wasn't sure about that either, when extending the data class, I saw that the docblocks show as return bool, return array, etc. but those return values weren't used in the data store implementation. Throwing an exception will be a better solution to stop the execution.

plugins/woocommerce/src/Internal/Fulfillments/Fulfillment.php
tpaksu: Got it. I'll use DI to get the instance, just saw it on the other data object code.

tpaksu: Addressed in [6f4db67](https://github.com/woocommerce/woocommerce/pull/57429/commits/6f4db670a3eb2edff79903047b3713c133ec368d).

plugins/woocommerce/src/Internal/DataStores/Fulfillments/FulfillmentsDataStore.php
tpaksu: Addressed in [0e98a2c](https://github.com/woocommerce/woocommerce/pull/57429/commits/0e98a2cd980f305f0321bdef4877c52ad559ba8e).

plugins/woocommerce/src/Internal/Fulfillments/Fulfillment.php
tpaksu: Addressed in [ec7c946](https://github.com/woocommerce/woocommerce/pull/57429/commits/ec7c946bd84cb1abcca349682de6e46361971fa0).

plugins/woocommerce/src/Internal/DataStores/Fulfillments/FulfillmentsDataStore.php
```diff
@@ -0,0 +1,375 @@
...
+		}
+		foreach ( $data->get_items() as $item ) {
+			if ( ! isset( $item['item_id'] ) || ! isset( $item['qty'] ) ) {
+				throw new \Exception( esc_html__( 'Invalid item.', 'woocommerce' ) );
+			}
+		}
+
+		// Set fulfillment properties.
+		$data->set_date_updated( new \DateTime( current_time( 'mysql' ) ) );
+		$data->set_date_deleted( null );
```
prettyboymp: Since `Fulfillment::date_deleted` defaults to `null`, can we remove this?  That way if there is a valid reason for a new Fulfillment to be created in a deleted state, e.g. imports/migrations, the client doesn't have to save twice.

plugins/woocommerce/src/Internal/DataStores/Fulfillments/FulfillmentsDataStore.php
```diff
@@ -0,0 +1,375 @@
...
+
+		// Save the fulfillment to the database.
+		global $wpdb;
+		$data_id = $wpdb->insert(
```
prettyboymp: Setting the result of the insert to `$data_id` may be confusing since $wpdb returns the number of rows inserted here.

plugins/woocommerce/src/Internal/DataStores/Fulfillments/FulfillmentsDataStore.php
```diff
@@ -0,0 +1,375 @@
...
+	 * @return void
+	 *
+	 * @throws \Exception If the fulfillment data can't be read.
+	 */
+	public function read( &$data ): void {
+		// Read the fulfillment from the database.
+		global $wpdb;
+
+		$data_id          = $data->get_id();
+		$fulfillment_data = $wpdb->get_row(
```
prettyboymp, line 104: This is okay for an initial pass, but we'll probably want to make sure we implement a `read_multiple()` or similar interface for loading multiple Fulfillments at once so we don't have to run a query for each one being displayed/returned.

plugins/woocommerce/src/Internal/DataStores/Fulfillments/FulfillmentsDataStore.php
```diff
@@ -0,0 +1,375 @@
...
+			array( '%d' )
+		);
+
+		// Check for errors.
+		if ( $wpdb->last_error ) {
+			throw new \Exception( esc_html__( 'Failed to delete fulfillment.', 'woocommerce' ) );
+		}
+
+		// Delete the metadata for the fulfillment.
+		$wpdb->delete(
```
prettyboymp: In #57428 the meta table has a `date_deleted` column.  Is there a reason to hard delete meta if we re only soft deleting the fulfillment?

plugins/woocommerce/src/Internal/Fulfillments/Fulfillment.php
```diff
@@ -0,0 +1,202 @@
...
+ *
+ * @since 9.9.0
+ */
+class Fulfillment extends \WC_Data {
+	/**
+	 * Fulfillment ID.
+	 *
+	 * @var int
+	 */
+	protected int $fulfillment_id = 0;
```
prettyboymp: If we're extending `\WC_Data` all of the attributes need to be stored in `$data` instead of object properties. It is what allows tracking of changes while a WC_Data object is in a changed state prior to saving back to the DB.  The blended approach where some of the properties are duplicated, like the fulfillment_id and items, it will likely cause bugs down the line if we don't have confidence in which is the canonical source for the value.

prettyboymp (changes requested):

plugins/woocommerce/src/Internal/DataStores/Fulfillments/FulfillmentsDataStore.php
tpaksu: We will want to keep the history of the order fulfillments with soft deleting records, but I think we may not want to keep the old metas. By keeping the deleted fulfillments on the table, we can see the history of order fulfillments by looking at the timeline, and their last state of their metadata, but I believe we will only will need the last state of the fulfillment meta. But if you have some experience about why keeping history of metas is important, I'll appreciate if you share it, so I can add a date_deleted column to the meta table.

plugins/woocommerce/src/Internal/Fulfillments/Fulfillment.php
tpaksu: Yes, I realized it while building the REST API, as the current state of the data store doesn't allow me to query the DB with entity type and id. I will add a new interface that will contain the REST method <-> database bridging functions and extend the data store with their implementations.

plugins/woocommerce/src/Internal/DataStores/Fulfillments/FulfillmentsDataStore.php
tpaksu: Yes, I realized it while building the REST API, as the current state of the data store doesn't allow me to query the DB with entity type and id. I will add a new interface that will contain the REST method <-> database bridging functions and extend the data store with their implementations.

plugins/woocommerce/src/Internal/DataStores/Fulfillments/FulfillmentsDataStore.php
tpaksu: Okay, maybe we should not allow users to delete metadata at all. We can only allow creating and updating a metadata, so we will be keeping the last state of the metadata when the fulfillment is deleted. How does it sound?

plugins/woocommerce/src/Internal/Fulfillments/Fulfillment.php
tpaksu: Addressed in https://github.com/woocommerce/woocommerce/pull/57429/commits/f07fcfde8335623fd46fd7adfdb1628421eaccf7.

plugins/woocommerce/src/Internal/DataStores/Fulfillments/FulfillmentsDataStore.php
tpaksu: Addressed in [10d619f](https://github.com/woocommerce/woocommerce/pull/57429/commits/10d619f0ae1879dc45f153db06dd8a2d6b1b64f6)

plugins/woocommerce/src/Internal/DataStores/Fulfillments/FulfillmentsDataStore.php
tpaksu: Addressed in [10d619f](https://github.com/woocommerce/woocommerce/pull/57429/commits/10d619f0ae1879dc45f153db06dd8a2d6b1b64f6)

plugins/woocommerce/src/Internal/DataStores/Fulfillments/FulfillmentsDataStore.php
tpaksu: Addressed in [10d619f](https://github.com/woocommerce/woocommerce/pull/57429/commits/10d619f0ae1879dc45f153db06dd8a2d6b1b64f6).

prettyboymp (approved):
Thanks for the updates.  🚢

PR #57437, opened 2025-04-21
Fix password-protected product layout in blockified templates shows unnecessary product title and product excerpt

### LINKED ISSUE #50391: Password-protected product layout in blockified templates shows unnecessary product title and product excerpt
In password-protected products, the Single Product template looks very different between the classic and the blockified templates, which seems incorrect.

Classic template | Blockified template
--- | ---
imatge | imatge

## Steps to reproduce

0. Make sure you are using the blockified _Single Product_ template. You can go to Appearance > Site Editor > Templates > Single Product and, if you see the _Classic Template Placeholder_ block, click on _Transform into blocks_.
1. Create or edit a product changing its visibility to _Password protected_.
2. Visit the product on the frontend and notice the weird layout:

imatge

Expected result: product title and excerpt shouldn't be displayed.

AhmarZaidi:

This PR checks for `core/post-title` and `core/post-excerpt` block and remove it's parent `core/column` block in the array reduce function while replacing the `single_product_template_blocks`.

The reason that we need to remove the parent `core/column` is that the `serialize_blocks()` gives error notices due to empty `core/column` block.

Closes #50391 

| Before | After |
| ------ | ----- |
| image | image |

Comparing Classic Template and Blockified Template after fix:
| Classic Template (twenty-twnety-four) | Blockified Template (twenty-twnety-five) |
| ------ | ----- |
| image | image |

0. Make sure you are using the blockified Single Product template. You can go to Appearance > Site Editor > Templates > Single Product and, if you see the Classic Template Placeholder block, click on Transform into blocks.
1. Create or edit a product changing its visibility to Password protected.
2. Visit the product on the frontend and notice that the layout is similar to the classic template (product title and excerpt removed)

image

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Blocks/Templates/SingleProductTemplate.php
```diff
@@ -142,9 +142,24 @@ private static function replace_first_single_product_template_block_with_passwor
 		// We want to replace the first single product template block with the password form. We also want to remove all other single product template blocks.
 		// This array doesn't contains all the blocks. For example, it missing the breadcrumbs blocks: it doesn't make sense replace the breadcrumbs with the password form.
 		$single_product_template_blocks = array( 'woocommerce/product-image-gallery', 'woocommerce/product-details', 'woocommerce/add-to-cart-form', 'woocommerce/product-meta', 'woocommerce/product-rating', 'woocommerce/product-price', 'woocommerce/related-products' );
```
Aljullu: Btw, no need to do in this PR, but while testing your PR I noticed there are some blocks that are not hidden in password-protected products. For example, _Related products_ (which is a _Product Collection_ block) is visible when it shouldn't:

imatge

In case you want to give it a try, I think we will need to add these extra blocks to the list:

* `woocommerce/add-to-cart-with-options`
* `woocommerce/product-gallery`
* `woocommerce/blockified-product-details`
* `woocommerce/product-collection`

As mentioned, not a blocker as it's not directly related to this task, but I noticed some blocks were missing from the list while reviewing your PR so I thought I would mention it anyway. :slightly_smiling_face:

Aljullu (commented):
Thanks for working on this, @AhmarZaidi!

> The reason that we need to remove the parent `core/column` is that the `serialize_blocks()` gives error notices due to empty `core/column` block.

Good finding. Apparently, that's an issue in the current implementation too. If the _Product Title_ and _Product Summary_ blocks are removed, we also end up with an empty column that causes some warnings to appear in the frontend.

I tested it and I could verify your implementation fixes that. But I worry the implemented solution might be a bit brittle. For example, if I add a _Spacer_ block before the _Product Title_ block, the warnings appear again:

imatge

I wonder if a more reliable approach would be to remove the columns with no children right before this return:

https://github.com/woocommerce/woocommerce/blob/ce0ab84e71d2683140b063507ea8038ffd16c345/plugins/woocommerce/src/Blocks/Templates/SingleProductTemplate.php#L206-L211

Something along these lines, maybe:

```PHP
if ( count( $new_inner_blocks ) === 0 ) {
	return array(
		'blocks'              => $carry['blocks'],
		'html_block'          => null,
		'removed'             => true,
		'is_already_replaced' => $carry['is_already_replaced'],
	);
}

return array(
	'blocks'              => array_merge( $carry['blocks'], array( $block ) ),
	'html_block'          => null,
	'removed'             => false,
	'is_already_replaced' => $carry['is_already_replaced'],
);
```

AhmarZaidi:
Hey @Aljullu, Thanks for the review.

I've updated the logic according to the suggestions. Removing the empty inner blocks seems to be a better approach.

I've also added the missing woocommerce blocks to remove list.

> In case you want to give it a try, I think we will need to add these extra blocks to the list:
woocommerce/add-to-cart-with-options
woocommerce/product-gallery
woocommerce/blockified-product-details
woocommerce/product-collection

Aljullu (commented):
Thanks for the updates, @AhmarZaidi! I noticed the _Related Products_ block is still visible on the page:

imatge

I noticed in the tempalte it's in fact a pattern and not a _Product Collection_ block as I thought:

https://github.com/woocommerce/woocommerce/blob/a5cc5f3f90fcd4176daafffa9f3f32a00c77449f/plugins/woocommerce/templates/templates/blockified/single-product.html#L49

I wonder if we should add an extra check for this case. So when we check if `$block['blockName']` is in `$single_product_template_blocks` [here](https://github.com/woocommerce/woocommerce/blob/a5cc5f3f90fcd4176daafffa9f3f32a00c77449f/plugins/woocommerce/src/Blocks/Templates/SingleProductTemplate.php#L148), we could add an extra check if the block is a pattern and the `slug` is `woocommerce-blocks/related-products`. What do you think? I guess that way the _Related Products_ will be correctly hidden.

AhmarZaidi:
@Aljullu Good catch, missed it while testing.

The fix you suggested works great and hides the related products pattern.
> add an extra check if the block is a pattern and the slug is woocommerce-blocks/related-products

Updated that in the PR: https://github.com/woocommerce/woocommerce/pull/57437/commits/6a68521892473cd9915d1baecd28d2f6b02ba4a7

Aljullu (commented):
Thanks for the updates, @AhmarZaidi! This looks perfect now:

Before | After
--- | ---
imatge | imatge

<details>

* Theme: Twenty Twenty-Five
* Store size: small
</details>

Apparently there is one test that needs to be updated before being able to merge:

https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/tests/php/src/Blocks/Templates/SingleProductTemplateTests.php#L308

If you need any help running the tests locally, let me know. :+1: The documentation can be found [here](https://github.com/woocommerce/woocommerce/blob/d905fc71e125b45d7bf2b7a700e0c6aaa9cb0c4f/plugins/woocommerce/tests/README.md#set-up-test-environment).

Aljullu (approved):
Thanks for updating the test, @AhmarZaidi! LGTM :rocket: 

This improvement will ship in WooCommerce 9.9.

PR #56986, opened 2025-04-02
Add a setting to WC -> Settings -> General to enable address autocomplete

opr:
This PR is based on: https://github.com/woocommerce/woocommerce/pull/56927

### Changes proposed in this Pull Request:

- Adds a setting to WC -> Settings -> General to enable address autocomplete
- If more than one provider is enabled, also register a select box to select the preferred one
- Add unit tests to cover changing settings and getting preferred provider.

Closes WOOPLUG-3614

### Screenshots or screen recordings:

| Before | After |
| ------ | ----- |
| image | image |

https://www.loom.com/share/0a7eb2885c6648a48fbec4ce10f2abc7

### How to test the changes in this Pull Request:

Using the [WooCommerce Testing Instructions Guide](https://github.com/woocommerce/woocommerce/wiki/Writing-high-quality-testing-instructions), include your detailed testing instructions:

1. Add `define( 'WC_EXPERIMENTAL_ADDRESS_AUTOCOMPLETE', true );` to your wp-config.php file
2. Go to WC -> Settings -> General, observe the `Enable predictive address search` setting. It should be disabled with some helper text saying: `Suggest full addresses for customer as they type. To use this feature, you need to install an address provider such as WooPayments.`
3. Using [Code Snippets](https://wordpress.org/plugins/code-snippets/), add a PHP Snippet and enter:

```php
__experimental_woocommerce_register_address_autocomplete_provider( 'woo-test', 'Woo test provider');
```

4. Refresh the settings page and ensure the checkbox is enabled and can be checked and unchecked. Save the settings page with it checked and refresh the page. Ensure it stays checked on reload. Do the same with it unchecked.
5. Add another provider using the same snippet above, change the ID and name, e.g.

```php
__experimental_woocommerce_register_address_autocomplete_provider( 'wp-test', 'WordPress test provider');
```

6. Refresh the setting page and check the `Enable predictive address search` box.
7. Ensure a `Preferred address autocomplete provider` dropdown menu appears with both providers listed.
8. Select one and save the page. Refresh the page and ensure the same one remains selected.
9. Try again with a different provider, ensure the correct one stays selected after a reload.
10. Uncheck the `Enable predictive address search` and ensure the dropdown menu is hidden.
11. Re-check the checkbox and ensure the dropdown menu reappears.

### Testing that has already taken place:

- Unit tests written to cover different settings scenarios
- Manual testing carried out as per instructions, see screen recording for steps taken.

### Changelog entry

[...]

PR #57441, opened 2025-04-21
Create address autocomplete registration JS function

opr:

This PR is based on `56715-implement-a-way-for-developers-to-register-an-address-autocomplete-provider-in-php` - https://github.com/woocommerce/woocommerce/pull/56986

Closes WOOPLUG-3616
Closes WOOPLUG-3660

This change introduces an address autocomplete provider registration function for the classic checkout. It also updates the Jest config for blocks to allow legacy JS to be tested.

The JS script is enqueued only when there are valid providers registered, but it is _registered_ unconditionally. Therefore any scripts that list it as a dependency won't crash if there are no PHP providers registered.

**I opted to include the legacy tests in the Blocks infrastructure because it seemed inappropriate to spin up a whole new workflow for testing legacy JS. There is no prior art for legacy JS testing.** 

**Key Features:**
- JavaScript provider registration that validates against server-registered providers  
- Address search functionality (triggers after 3+ characters)
- Suggestion display with match highlighting (max 5 suggestions)
- Address field population on selection with proper `Select2` handling for state/country fields
- Re-checks eligible providers when changing country
- Input sanitization on results from server
- Conditional script/style loading (only when providers exist)
- Legacy JS testing infrastructure integrated with WC Blocks Jest setup

**Technical Implementation:**
- Frontend script registered unconditionally but enqueued only with valid providers
- Provider data passed via `wc_checkout_params` with sanitized output
- Debounced search (100ms) to prevent excessive API calls
- Browser autofill disabled during autocomplete interaction, lastpass/1password also attempt to disable (in some cases it comes back)
- Global state namespaced under `window.wc.addressAutocomplete` to avoid conflicts

image

#### Prerequisites:
- Set up a test site with WooCommerce installed
- Ensure the experimental blocks feature is enabled
- Use Chrome/Firefox with Developer Tools open

#### Part 1: Basic Provider Registration

1. Get the mock provider plugins here:
- [mock-address-provider.zip](https://github.com/user-attachments/files/20726099/mock-address-provider.zip)
- [mock-address-provider-germany.zip](https://github.com/user-attachments/files/20726300/mock-address-provider-germany.zip)

- Install to `wp-content/plugins/mock-address-provider` and `wp-content/plugins/mock-address-provider-germany`
- Enter its directory, `npm i` and `npm run start`
- Activate the plugins, then go to WooCommerce -> Settings -> General and select `Enable predictive address search`. 
- Ensure you see both "Mock address provider" and "Mock address provider" germany in the `Preferred address autocomplete provider` option.

2. Go to the shortcode checkout page and ensure:
   - No JavaScript errors appear in the console
   - The `wc-address-autocomplete` script is loaded (check Network tab)
   - The `wc-address-autocomplete.css` stylesheet is loaded

3. In the browser console, verify the providers are available:
```js
console.log(window.wc_checkout_params.address_providers);
// Should show: [{id: "wp-provider", name: "WordPress provider"}, {id: "woo-provider", name: "Woo provider"}]
```

#### Part 2: JavaScript Provider Registration

4. Register a valid provider in the console (this is done via the Mock address provider plugin)
5. Ensure you get no JS errors when loading the page.

```

6. Try entering the following in the JS console:
```js
const invalidProvider = {
    id: 'invalid-provider',
    canSearch: () => true,
    search: () => Promise.resolve([]),
    select: () => Promise.resolve({})
};

window.wc.addressAutocomplete.registerAddressAutocompleteProvider(invalidProvider);
// Expected: Error "Provider invalid-provider not registered on server", returns false
```

8. Test various invalid provider configurations:
```js
// Missing required properties
window.wc.addressAutocomplete.registerAddressAutocompleteProvider({id: 'woo-provider'});
// Expected: Error "Address provider must have a canSearch function"

window.wc.addressAutocomplete.registerAddressAutocompleteProvider(null);
// Expected: Error "Address provider must be a valid object"

window.wc.addressAutocomplete.registerAddressAutocompleteProvider({
    id: 123, // Wrong type
    canSearch: () => true,
    search: () => Promise.resolve([]),
    select: () => Promise.resolve({})
});
// Expected: Error "Address provider must have a valid ID"
```

#### Part 3: Preferred Provider Settings

9. Go to **WooCommerce → Settings → General** and look for the "Address autocomplete provider" setting
10. Select "Mock address provider Germany" as the preferred provider and save
11. Refresh the checkout page and run:
```js
console.log(window.wc_checkout_params.address_providers);
// Expected: "mock-address-provider-germany" should be first in the array
```

#### Part 4: Address Autocomplete Functionality

13. Test address search:
    - Set country to "United States (US)"
    - Type "Main" in the billing address field
    - Expected: After 3 characters, suggestions appear with different areas highlighted. The locations will not be filtered down to match the input now as it is just a mock provider for testing!
    - Expected: Search icon appears next to the input field

14. Test suggestion selection:
    - Click on a suggestion
    - Expected: All address fields are populated (address, city, state, postcode)
    - Expected: Country remains "US"
    - Expected: Suggestions disappear

15. Test keyboard navigation:
    - Type in address field again
    - Use arrow keys to navigate suggestions
    - Press Enter to select
    - Press Escape to close suggestions

#### Part 5: Country Change Behavior

16. Change country to "Canada":
    - Trigger autocomplete suggestions by typing into address 1.
    - Pick a US address
    - Ensure the fields update correctly including State.

17. Deactivate the "Mock address provider" (disable the plugin). In the "Mock Address Provider Germany" plugin:
	- Modify the `canSearch` function to:
```js
		canSearch: ( country ) => {
			return country === 'DE';
		},
```
18. Change country to USA
19. Type into the address 1 field, ensure no suggestions show as the provider can't search.
20. Change to Germany and repeat, ensure suggestions show and they can be selected and populate the form.
21. Reactivate the "Mock Address Provider" plugin again. Go to WC -> Settings -> General and set "Mock Address Provider Germany" as the preferred provider.
22. Change country back to USA, ensure you see the non-Germany suggestions when typing in address 1.
25. For US addresses, verify state field:
    - After selecting an address, state dropdown should show correct state

#### Part 7: Security Testing

26. Test input sanitization:
```js
// Try to inject malicious content
const input = document.getElementById('billing_address_1');
input.value = '<script>alert("XSS")</script>';
// Expected: No alert appears, suggestions work normally
```

27. Test malformed provider responses - edit the `src/index.js` file of the preferred provider plugin.
```js
const maliciousProvider = {
    id: 'wp-provider',
    canSearch: () => true,
    search: () => '<script>alert("XSS")</script>', // Not an array
    select: () => Promise.resolve({})
};
// Type in address field
// Expected: Error "Invalid suggestions response - not an array"
```

#### Part 8: Edge Cases

28. Test rapid country changes:
    - Quickly switch between countries while typing
    - Expected: No errors, suggestions update correctly

29. Test concurrent searches:
    - Type quickly to trigger multiple searches
    - Expected: Only latest results shown, no errors

30. Remove the PHP snippet and refresh:
    - Expected: `wc-address-autocomplete` script is NOT loaded
    - Expected: Checkout works normally without autocomplete

#### Part 9: Accessibility

31. Test with screen reader:
    - Suggestions should be announced
    - Arrow key navigation should announce selected items
    - ARIA attributes should be properly set

32. Test without JavaScript:
    - Disable JavaScript and refresh
    - Expected: Checkout still works normally

### Expected Results Summary:
- ✅ Providers can only be registered if they exist server-side
- ✅ Preferred provider setting is respected
- ✅ Address autocomplete works with proper country filtering
- ✅ Select2 fields are properly handled
- ✅ Security measures prevent XSS attacks
- ✅ System gracefully handles errors and edge cases
- ✅ Feature can be completely disabled by removing providers

Implemented legacy JS tests for this.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

## Summary by CodeRabbit

- **New Features**
  - Introduced address autocomplete functionality for WooCommerce checkout forms, supporting billing and shipping addresses.
  - Added a suggestions dropdown with keyboard and mouse navigation, including accessibility enhancements.
  - Enabled dynamic provider registration for address autocomplete, with support for multiple providers.
  - New styles for the address autocomplete dropdown and icon.
  - Conditional loading of address autocomplete scripts and styles based on an experimental feature flag.

- **Bug Fixes**
  - Improved error handling and validation during provider registration.

- **Tests**
  - Added comprehensive tests for address autocomplete provider registration and suggestions UI.

- **Chores**
  - Updated test configuration to include legacy JavaScript files.
  - Documented changes in the WooCommerce plugin changelog.

### DIFF
--- a/plugins/woocommerce/client/legacy/css/address-autocomplete.scss
+++ b/plugins/woocommerce/client/legacy/css/address-autocomplete.scss
@@ -1,0 +1,70 @@
+.woocommerce-address-suggestions {
+	position: absolute;
+	z-index: 1000;
+	width: 100%;
+	max-width: 100%;
+	margin-top: 4px;
+	background: #fff;
+	border: 1px solid #ddd;
+	border-radius: 4px;
+	box-shadow: 0 2px 4px rgba(0, 0, 0, 0.1);
+	box-sizing: border-box;
+	display: none;
+
+	.suggestions-list {
+		list-style: none;
+		margin: 0;
+		padding: 0;
+		max-height: 200px;
+		overflow-y: auto;
+
+		li {
+			padding: 8px 12px;
+			font-size: 14px;
+			cursor: pointer;
+			text-overflow: ellipsis;
+			overflow: hidden;
+			white-space: nowrap;
+
+			&:last-child {
+				border-bottom: none;
+			}
+
+			&:hover,
+			&.active {
+				background-color: #0073aa;
+				color: #fff;
+			}
+
+			&:focus {
+				outline: none;
+				background-color: #0073aa;
+				color: #fff;
+			}
+		}
+	}
+}
+
+// Ensure the input container has relative positioning
+.woocommerce-input-wrapper:has(#billing_address_1, #shipping_address_1) {
+	position: relative;
+}
+
+// Search icon styles
+.address-search-icon {
+	position: absolute;
+	inset-inline-end: 16px;
+	top: 50%;
+	transform: translateY(-50%);
+	pointer-events: none;
+	width: 16px;
+	height: 16px;
+	line-height: 16px;
+	z-index: 1;
+
+	svg {
+		fill: none;
+		stroke: var(--wc-form-color-text, #444);
+		stroke-width: 1;
+	}
+}
--- a/plugins/woocommerce/client/legacy/css/address-autocomplete.scss
+++ b/plugins/woocommerce/client/legacy/css/address-autocomplete.scss
@@ -48,6 +48,7 @@
 // Ensure the input container has relative positioning
 .woocommerce-input-wrapper:has(#billing_address_1, #shipping_address_1) {
 	position: relative;
+	display: block;
 }
 
 // Search icon styles
--- a/plugins/woocommerce/client/legacy/css/address-autocomplete.scss
+++ b/plugins/woocommerce/client/legacy/css/address-autocomplete.scss
@@ -72,6 +72,8 @@
 			height: 16px;
 			line-height: 16px;
 			z-index: 1;
+
+			// Mask needed to ensure the SVG has the correct color.
 			background-color: var(--wc-form-color-text, #444);
 			mask: url('data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHZpZXdCb3g9IjAgMCAxNCAxNCIgZm9jdXNhYmxlPSJmYWxzZSIgYXJpYS1oaWRkZW49InRydWUiPgogIDxjaXJjbGUgY3g9IjYiIGN5PSI2IiByPSI0IiBmaWxsPSJub25lIiBzdHJva2U9ImJsYWNrIiBzdHJva2Utd2lkdGg9IjEuNSI+PC9jaXJjbGU+CiAgPHBhdGggc3Ryb2tlLWxpbmVjYXA9InJvdW5kIiBzdHJva2UtbGluZWpvaW49InJvdW5kIiBzdHJva2U9ImJsYWNrIiBzdHJva2Utd2lkdGg9IjEuNSIgZD0ibTkuMjUgOS4yNSAyLjUgMi41Ij48L3BhdGg+Cjwvc3ZnPg==') no-repeat center;
 			mask-size: contain;
--- a/plugins/woocommerce/client/legacy/css/address-autocomplete.scss
+++ b/plugins/woocommerce/client/legacy/css/address-autocomplete.scss
@@ -72,12 +72,13 @@
 			height: 16px;
 			line-height: 16px;
 			z-index: 1;
+			background-color: var(--wc-form-color-text, #444);
+			mask: url('data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHZpZXdCb3g9IjAgMCAxNCAxNCIgZm9jdXNhYmxlPSJmYWxzZSIgYXJpYS1oaWRkZW49InRydWUiPgogIDxjaXJjbGUgY3g9IjYiIGN5PSI2IiByPSI0IiBmaWxsPSJub25lIiBzdHJva2U9ImJsYWNrIiBzdHJva2Utd2lkdGg9IjEuNSI+PC9jaXJjbGU+CiAgPHBhdGggc3Ryb2tlLWxpbmVjYXA9InJvdW5kIiBzdHJva2UtbGluZWpvaW49InJvdW5kIiBzdHJva2U9ImJsYWNrIiBzdHJva2Utd2lkdGg9IjEuNSIgZD0ibTkuMjUgOS4yNSAyLjUgMi41Ij48L3BhdGg+Cjwvc3ZnPg==') no-repeat center;
+			mask-size: contain;
 
-			svg {
-				fill: none;
-				stroke: var(--wc-form-color-text, #444);
-				stroke-width: 1;
-			}
+			// Safari support.
+			-webkit-mask: url('data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHZpZXdCb3g9IjAgMCAxNCAxNCIgZm9jdXNhYmxlPSJmYWxzZSIgYXJpYS1oaWRkZW49InRydWUiPgogIDxjaXJjbGUgY3g9IjYiIGN5PSI2IiByPSI0IiBmaWxsPSJub25lIiBzdHJva2U9ImJsYWNrIiBzdHJva2Utd2lkdGg9IjEuNSI+PC9jaXJjbGU+CiAgPHBhdGggc3Ryb2tlLWxpbmVjYXA9InJvdW5kIiBzdHJva2UtbGluZWpvaW49InJvdW5kIiBzdHJva2U9ImJsYWNrIiBzdHJva2Utd2lkdGg9IjEuNSIgZD0ibTkuMjUgOS4yNSAyLjUgMi41Ij48L3BhdGg+Cjwvc3ZnPg==') no-repeat center;
+			-webkit-mask-size: contain;
 		}
 	}
 }
--- a/plugins/woocommerce/client/legacy/css/address-autocomplete.scss
+++ b/plugins/woocommerce/client/legacy/css/address-autocomplete.scss
@@ -49,23 +49,35 @@
 .woocommerce-input-wrapper:has(#billing_address_1, #shipping_address_1) {
 	position: relative;
 	display: block;
-}
 
-// Search icon styles
-.address-search-icon {
-	position: absolute;
-	inset-inline-end: 16px;
-	top: 50%;
-	transform: translateY(-50%);
-	pointer-events: none;
-	width: 16px;
-	height: 16px;
-	line-height: 16px;
-	z-index: 1;
+	// Default to showing nothing if autocomplete is not available.
+	.address-search-icon {
+		display: none;
+	}
 
-	svg {
-		fill: none;
-		stroke: var(--wc-form-color-text, #444);
-		stroke-width: 1;
+	// Search icon styles
+	&.autocomplete-available {
+		input#billing_address_1, input#shipping_address_1 {
+			padding-right: calc(1.1rem + 16px); // Adjust padding for icon
+		}
+
+		.address-search-icon {
+			display: block;
+			position: absolute;
+			inset-inline-end: 16px;
+			top: 50%;
+			transform: translateY(-50%);
+			pointer-events: none;
+			width: 16px;
+			height: 16px;
+			line-height: 16px;
+			z-index: 1;
+
+			svg {
+				fill: none;
+				stroke: var(--wc-form-color-text, #444);
+				stroke-width: 1;
+			}
+		}
 	}
 }
--- a/plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
+++ b/plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
@@ -38,7 +38,11 @@ function registerAddressAutocompleteProvider( provider ) {
 
 	// Check if provider is registered on server
 	var serverProviders = [];
-	if ( window && window.wc_checkout_params && window.wc_checkout_params.address_providers ) {
+	if (
+		window &&
+		window.wc_checkout_params &&
+		window.wc_checkout_params.address_providers
+	) {
 		serverProviders = window.wc_checkout_params.address_providers;
 	}
 
--- a/plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
+++ b/plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
@@ -48,7 +48,12 @@ function registerAddressAutocompleteProvider( provider ) {
 	}
 
 	var isRegistered = serverProviders.some( function ( serverProvider ) {
-		return serverProvider.id === provider.id;
+		return (
+			serverProvider &&
+			typeof serverProvider === 'object' &&
+			typeof serverProvider.id === 'string' &&
+			serverProvider.id === provider.id
+		);
 	} );
 	if ( ! isRegistered ) {
 		console.error(
--- a/plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
+++ b/plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
@@ -1,0 +1,55 @@
+/**
+ * Simple address provider registration for WooCommerce checkout
+ */
+var wooAddressProviders = {};
+
+/**
+ * Register an address autocomplete provider
+ *
+ * @param {Object} provider The provider object
+ */
+function registerAddressAutocompleteProvider( provider ) {
+	// Check required properties
+	if ( ! provider.id || typeof provider.id !== 'string' ) {
+		console.error( 'Address provider must have a valid ID' );
+		return false;
+	}
+
+	if ( typeof provider.canSearch !== 'function' ) {
+		console.error( 'Address provider must have a canSearch function' );
+		return false;
+	}
+
+	if ( typeof provider.search !== 'function' ) {
+		console.error( 'Address provider must have a search function' );
+		return false;
+	}
+
+	if ( typeof provider.select !== 'function' ) {
+		console.error( 'Address provider must have a select function' );
+		return false;
+	}
+
+	// Check if provider is registered on server
+	var serverProviders = window.wc_checkout_params.address_providers || [];
+	console.log( JSON.stringify( serverProviders ) );
+	var isRegistered = serverProviders.some( function ( serverProvider ) {
+		return serverProvider.id === provider.id;
+	} );
+	if ( ! isRegistered ) {
+		console.error(
+			'Provider ' + provider.id + ' not registered on server'
+		);
+		return false;
+	}
+
+	// Add provider to registry
+	wooAddressProviders[ provider.id ] = provider;
+	return true;
+}
+
+// Make functions available globally
+window.wc = window.wc || {};
+window.wc.addressAutocomplete = {
+	registerAddressAutocompleteProvider: registerAddressAutocompleteProvider,
+};
--- a/plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
+++ b/plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
@@ -57,7 +57,8 @@ function registerAddressAutocompleteProvider( provider ) {
 		return false;
 	}
 
-	// Add provider to registry
+	// Freeze and add provider to registry.
+	Object.freeze( provider );
 	wooAddressProviders[ provider.id ] = provider;
 	return true;
 }
--- a/plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
+++ b/plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
@@ -10,66 +10,66 @@ var wooAddressProviders = {};
  * @return {boolean} Whether the registration was successful
  */
 function registerAddressAutocompleteProvider( provider ) {
-	// Check required properties
-	if ( ! provider || typeof provider !== 'object' ) {
-		console.error( 'Address provider must be a valid object' );
+	try {
+		// Check required properties
+		if ( ! provider || typeof provider !== 'object' ) {
+			throw new Error( 'Address provider must be a valid object' );
+		}
+
+		if ( ! provider.id || typeof provider.id !== 'string' ) {
+			throw new Error( 'Address provider must have a valid ID' );
+		}
+
+		if ( typeof provider.canSearch !== 'function' ) {
+			throw new Error(
+				'Address provider must have a canSearch function'
+			);
+		}
+
+		if ( typeof provider.search !== 'function' ) {
+			throw new Error( 'Address provider must have a search function' );
+		}
+
+		if ( typeof provider.select !== 'function' ) {
+			throw new Error( 'Address provider must have a select function' );
+		}
+
+		// Check if provider is registered on server
+		var serverProviders = [];
+		if (
+			window &&
+			window.wc_checkout_params &&
+			window.wc_checkout_params.address_providers
+		) {
+			serverProviders = window.wc_checkout_params.address_providers;
+		}
+
+		if ( ! Array.isArray( serverProviders ) ) {
+			throw new Error( 'Server providers configuration is invalid' );
+		}
+
+		var isRegistered = serverProviders.some( function ( serverProvider ) {
+			return (
+				serverProvider &&
+				typeof serverProvider === 'object' &&
+				typeof serverProvider.id === 'string' &&
+				serverProvider.id === provider.id
+			);
+		} );
+		if ( ! isRegistered ) {
+			throw new Error(
+				'Provider ' + provider.id + ' not registered on server'
+			);
+		}
+
+		// Freeze and add provider to registry.
+		Object.freeze( provider );
+		wooAddressProviders[ provider.id ] = provider;
+		return true;
+	} catch ( error ) {
+		console.error( 'Error registering address provider:', error.message );
 		return false;
 	}
-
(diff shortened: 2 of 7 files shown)


plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
```diff
@@ -0,0 +1,69 @@
...
+ */
+var wooAddressProviders = {};
+
+/**
+ * Register an address autocomplete provider
+ *
+ * @param {Object} provider The provider object
+ * @return {boolean} Whether the registration was successful
+ */
+function registerAddressAutocompleteProvider( provider ) {
```
tjcafferkey, line 18: Might be good to also use a `try/catch` here to handle unexpected errors

plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
```diff
@@ -0,0 +1,69 @@
...
+		serverProviders = window.wc_checkout_params.address_providers;
+	}
+
+	if ( ! Array.isArray( serverProviders ) ) {
+		console.error( 'Server providers configuration is invalid' );
+		return false;
+	}
+
+	var isRegistered = serverProviders.some( function ( serverProvider ) {
+		return serverProvider.id === provider.id;
```
tjcafferkey: Even though we might be `serverProviders` is an array of objects we should probably put some guards in since its in the client and can be modified.

plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
```diff
@@ -0,0 +1,69 @@
...
+	} );
+	if ( ! isRegistered ) {
+		console.error(
+			'Provider ' + provider.id + ' not registered on server'
+		);
+		return false;
+	}
+
+	// Add provider to registry
+	wooAddressProviders[ provider.id ] = provider;
```
tjcafferkey: Shall we use `Object.freeze` here?

opr: Yeah it might be a good idea to prevent modification later. Since we don't have any nested objects here it would work well.

plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
opr: [ad639c6](https://github.com/woocommerce/woocommerce/pull/57441/commits/ad639c63595eac26b23fbfb3b9a5c3235733000a)

plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
opr: [ec95c9f](https://github.com/woocommerce/woocommerce/pull/57441/commits/ec95c9fe0917f8dd01ea36c1632a81e1703181d5)

plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
opr: [5411fe0](https://github.com/woocommerce/woocommerce/pull/57441/commits/5411fe02038ad36dbf0f63945ed45d146fe90469)

plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
```diff
@@ -0,0 +1,751 @@
...
+			// Validate address type.
+			if ( ! validateAddressType( type ) ) {
+				console.error( 'Invalid address type:', type );
+				return;
+			}
+
+			// Sanitize input value.
+			const sanitizedInput = sanitizeInput( inputValue );
+			if ( sanitizedInput !== inputValue ) {
+				console.warn( 'Input was sanitized for security' );
```
opr, line 754: Remove

plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
```diff
@@ -0,0 +1,751 @@
...
+			}
+
+			// Create new AbortController for this search.
+			searchControllers[ type ] = new AbortController();
+			const controller = searchControllers[ type ];
+
+			try {
+				const filteredSuggestions = await activeAddressProvider[
+					type
+				].search( type, sanitizedInput );
```
opr: pass country instead of type

plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
```diff
@@ -0,0 +1,751 @@
...
+			if (
+				typeof addressId !== 'string' ||
+				addressId.length === 0 ||
+				addressId.length > 100
+			) {
+				console.error( 'Invalid address ID for selection:', addressId );
+				return;
+			}
+
+			const addressInput = addressInputs[ type ][ 'address_1' ];
```
opr: re-select nodes in case they entered/left since caching

plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
```diff
@@ -0,0 +1,679 @@
...
+						'</svg>';
+					addressInput.parentNode.appendChild( searchIcon );
+				}
+
+				addressInputs[ type ] = {};
+				addressInputs[ type ][ 'address_1' ] = addressInput;
+				addressInputs[ type ][ 'city' ] = cityInput;
+				addressInputs[ type ][ 'country' ] = countryInput;
+				addressInputs[ type ][ 'postcode' ] = postcodeInput;
+
```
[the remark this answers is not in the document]
opr: This is only happens on first load of the IIFE, so it's not a risk to drop the object entry

> Potential notice when no providers are available

Providers cannot be registered with empty string as their ID.

plugins/woocommerce/client/legacy/js/frontend/test/address-autocomplete.js
```diff
@@ -0,0 +1,280 @@
...
+describe( 'Address Autocomplete Provider Registration', () => {
+	beforeEach( () => {
+		// Reset the window object and providers before each test
+		global.window = {};
+		global.window.wc_checkout_params = {
+			address_providers: [
+				{ id: 'test-provider', name: 'Test provider' },
+				{ id: 'wc-payments', name: 'WooCommerce Payments' },
+			],
+		};
```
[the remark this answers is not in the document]
opr: Done in b1cb04740222e8419e836b897fd0fc56d215b124

plugins/woocommerce/client/legacy/js/frontend/test/address-autocomplete.js
```diff
@@ -0,0 +1,281 @@
...
+			},
+		} );
+		global.console = {
+			error: jest.fn(),
+		};
+
+		// Reset the module before each test
+		jest.resetModules();
+		require( '../address-autocomplete' );
+	} );
```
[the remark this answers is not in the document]
opr: Last review you told me nuking the jsdom window wasn't good.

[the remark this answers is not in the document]
opr: I don't think removing the wc namespace is particularly important here. I will skip the suggestion! Thanks.

plugins/woocommerce/client/legacy/css/address-autocomplete.scss
```diff
@@ -0,0 +1,70 @@
...
+	height: 16px;
+	line-height: 16px;
+	z-index: 1;
+
+	svg {
+		fill: none;
+		stroke: var(--wc-form-color-text, #444);
+		stroke-width: 1;
+	}
+}
```
senadir: Should this be scoped to the input?

plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
```diff
@@ -0,0 +1,679 @@
...
+						'viewBox="0 0 14 14" ' +
+						'focusable="false" ' +
+						'aria-hidden="true">' +
+						'<circle cx="6" cy="6" r="4"></circle>' +
+						'<path stroke-linecap="round" ' +
+						'stroke-linejoin="round" ' +
+						'd="m9.25 9.25 2.5 2.5"></path>' +
+						'</svg>';
+					addressInput.parentNode.appendChild( searchIcon );
+				}
```
senadir: 2 Things:
- Let's use a CSS image for this (can be just this as base64 or something?)
- This is still visible even when we didn't have any JS providers active.

plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
```diff
@@ -0,0 +1,679 @@
...
+
+		/**
+		 * Sanitize HTML for display by removing any HTML tags.
+		 *
+		 * @param html
+		 * @return {string|string}
+		 */
+		function sanitizeForDisplay( html ) {
+			const doc = document.implementation.createHTMLDocument( '' );
+			doc.body.innerHTML = html;
```
straku, line 738: It might be problematic from security point of view? Any user input will be parsed and executed here

plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
```diff
@@ -0,0 +1,679 @@
...
+			}
+
+			if ( addressData.country ) {
+				setFieldValue(
+					addressInputs[ type ][ 'country' ],
+					addressData.country
+				);
+			}
+
+			setTimeout( function () {
```
straku: Do we need to clear the timeout?

plugins/woocommerce/client/legacy/js/frontend/test/address-autocomplete.js
```diff
@@ -0,0 +1,285 @@
+/**
+ * @jest-environment jsdom
+ */
+
+describe( 'Address Autocomplete Provider Registration', () => {
```
straku, line 5: I can see that the registration is covered thoroughly. Do we have plans to cover the component logic itself?

straku:
Notes from my testing.

> Expected: Search icon appears next to the input field

It shows, but for longer input values it overlays the value:

Screenshot 2025-06-17 at 14 40 53

> Expected: Country remains "US"

The country is being set to the country suggested in the autocomplete, not the one that I've initially inputted. This happens consistently.

---

We should hide the suggestions when moving focus to another input field without selecting a suggestion (so if you click TAB while suggestions are displayed). See screen recording:

https://github.com/user-attachments/assets/2a0bfe46-ffc3-4c21-9d08-07f4af79d6c6

---

I am not sure how I feel about input fields staying populated with the previously selected values in the autocomplete when I explicitly change country to a different one. I think I would expect them to reset. Although I acknowledge this is an edge case as behavior like this won't happen often.

--- 

I know we're working with test data now, but letting you know that for Germany mock provider, I am not getting county/state filled when using any of the suggestions.

---

> Debounced search (100ms) to prevent excessive API calls

I am getting ajax calls for every keystroke 🤔 

---

In voiceover, once the suggestions are shown they completely take over the focus. Changes to the input state are no longer announced, instead on every input update I am being informed of the top suggestion. I think this happens because we automatically focus on the top suggestion? Compare it to the way Ariakit does it for Combobox https://ariakit.org/components/combobox, you need to explicitly move with arrow keys to the first suggestion before it gets announced.

plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
straku: I tested it extensively and can't trip it, so I guess I was wrong!

straku:
@opr Finished my review, I feel like we're in a good spot to merge this, but I left enough comments that I will wait a bit for you to acknowledge them. Can you rerequest a review afterward?

senadir:
> I am getting ajax calls for every keystroke 🤔

I think the AI get confused by the timeout that's only meant for first render

plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
opr: Good idea, thanks. Added it

plugins/woocommerce/client/legacy/css/address-autocomplete.scss
opr: Done, what do you think of the changes?

plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
```diff
@@ -0,0 +1,708 @@
...
+				suggestionsContainer.style.marginTop =
+					addressInputs[ type ][ 'address_1' ].offsetHeight + 'px';
+				addressInput.setAttribute( 'aria-expanded', 'true' );
+				addressInput.setAttribute(
+					'aria-owns',
+					`address_suggestions_${ type }_list`
+				);
+				suggestionsList.id = `address_suggestions_${ type }_list`;
+				setActiveSuggestion( type, 0 );
+			} catch ( error ) {
```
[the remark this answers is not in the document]
senadir: Seems valid @opr ?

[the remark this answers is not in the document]
senadir: Example of Shopify

https://github.com/user-attachments/assets/c6244df1-79f3-462c-a7ef-798f115df0bf

senadir (commented):
Looking good, giving it another review today

plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
opr: > Let's use a CSS image for this (can be just this as base64 or something?)

Done in 4dfe78f0428b6f16c08384d53b489fe0ab74cfb1

> This is still visible even when we didn't have any JS providers active.

I just disabled both of mine and it's not showing. Can you re-test?

opr:
@coderabbitai review

opr:
@coderabbitai
> **Actionable comments posted: 0**
> 
> ♻️ Duplicate comments (4)
> 🧹 Nitpick comments (7)
> > plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js (6)> `46-47`: **Apply optional chaining for cleaner null checks.**
> > > The current nested property access can be simplified using optional chaining as suggested by static analysis.
> > > ```diff
> > > -		if (
> > > -			window &&
> > > -			window.wc_checkout_params &&
> > > -			Array.isArray( window.wc_checkout_params.address_providers ) &&
> > > -			window.wc_checkout_params.address_providers.length > 0
> > > -		) {
> > > +		if ( window?.wc_checkout_params?.address_providers?.length > 0 ) {
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > `87-87`: **Remove redundant strict mode directive.**
> > > The 'use strict' directive is unnecessary in modern JavaScript modules as they are automatically in strict mode.
> > > ```diff
> > > -	'use strict';
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > `98-100`: **Apply optional chaining for safer property access.**
> > > Simplify the nested property access pattern with optional chaining.
> > > ```diff
> > > -		const serverProviders =
> > > -			( window &&
> > > -				window.wc_checkout_params &&
> > > -				window.wc_checkout_params.address_providers ) ||
> > > -			[];
> > > +		const serverProviders = window?.wc_checkout_params?.address_providers || [];
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > `245-245`: **Apply optional chaining for jQuery check.**
> > > Simplify the nested property access for jQuery and Select2 checking.
> > > ```diff
> > > -				if ( window.jQuery && window.jQuery( countryInput ).select2 ) {
> > > +				if ( window.jQuery?.( countryInput )?.select2 ) {
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > `498-501`: **Apply optional chaining for jQuery hasClass check.**
> > > Simplify the nested jQuery property access.
> > > ```diff
> > > -				if (
> > > -					window.jQuery &&
> > > -					window
> > > -						.jQuery( input )
> > > -						.hasClass( 'select2-hidden-accessible' )
> > > -				) {
> > > +				if ( window.jQuery?.( input )?.hasClass?.( 'select2-hidden-accessible' ) ) {
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > `629-629`: **Remove unnecessary variable aliasing.**
> > > The aliasing of `this` to `inputElement` is unnecessary since arrow functions inherit `this` from their enclosing scope.
> > > ```diff
> > >  				addressInput.addEventListener( 'input', function () {
> > >  					clearTimeout( inputTimeout );
> > > -					const inputElement = this;
> > >  					inputTimeout = setTimeout( () => {
> > > -						if ( document.activeElement === inputElement ) {
> > > +						if ( document.activeElement === this ) {
> > >  							displaySuggestions(
> > > -								inputElement.value,
> > > +								this.value,
> > >  								countryInput.value,
> > >  								type
> > >  							);
> > > ```
> > 
> > 
> > 
> > plugins/woocommerce/client/legacy/js/frontend/test/address-autocomplete.js (1)> `7-7`: **Consider performance impact of delete operator usage.**
> > > Static analysis suggests avoiding the delete operator due to potential performance impact. However, in test environments, this is typically acceptable for ensuring clean state.
> > > If performance becomes a concern, you could use assignment instead:
> > > ```diff
> > > -		delete global.window.wc;
> > > +		global.window.wc = undefined;
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > However, `delete` is commonly used in test setup for ensuring clean slate between tests, so the current approach is acceptable.
> 
> 📜 Review details



Optional chaining doesn't work in this environment it causes errors, we have to use "old" JS

opr:
@straku @senadir I'd appreciate another test/review on the changes I've made, specifically looking at autocomplete functionality, it should disable when entering the third character and re-enable when removing down to less than 3 characters.

straku:
Retesting some of the issues that I've encountered in my initial review.

> It shows, but for longer input values it overlays the value

✅ This is no longer happening 

---

> We should hide the suggestions when moving focus to another input field without selecting a suggestion (so if you click TAB while suggestions are displayed).

✅ Fixed

---

> Debounced search (100ms) to prevent excessive API calls

❓ I think it's fixed, but I don't understand clearly how it works under the hood. See the calls that happened after I quickly typed in three letters:

Screenshot 2025-06-25 at 15 07 16

---

⚠️ Something weird is happening with VoiceOver, when I pick a suggested item with keyboard navigation the focus is being transferred somewhere outside the form. I think the main browser window is getting focused. This is not happening with VoiceOver off.

plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
```diff
@@ -0,0 +1,1249 @@
...
+		if ( typeof provider.search !== 'function' ) {
+			throw new Error( 'Address provider must have a search function' );
+		}
+
+		if ( typeof provider.select !== 'function' ) {
+			throw new Error( 'Address provider must have a select function' );
+		}
+
+		// Check if provider is registered on server.
+		var serverProviders = [];
```
straku:
```suggestion
		let serverProviders = [];
```
I don't think you are relying on hoisting here, so `let` should be safe. I wouldn't suggest it normally, but I can see you're using `const` further down in the implementation, so probably better to align on modern pattern.

plugins/woocommerce/client/legacy/js/frontend/address-autocomplete.js
```diff
@@ -0,0 +1,1249 @@
...
+			window.wc_checkout_params.address_providers.length > 0
+		) {
+			serverProviders = window.wc_checkout_params.address_providers;
+		}
+
+		if ( ! Array.isArray( serverProviders ) ) {
+			throw new Error( 'Server providers configuration is invalid' );
+		}
+
+		var isRegistered = serverProviders.some( function ( serverProvider ) {
```
straku:
```suggestion
		let isRegistered = serverProviders.some( function ( serverProvider ) {
```

straku:
I am not getting any bolding on matched parts and I think it should happen based on the implementation 🤔 @opr Is this related to the mocked provider or is matching done fully client-side?

Screenshot 2025-06-26 at 10 52 10

EDIT: I know it should come from the provider. Any way for me to test it?

straku:
Did a bit more testing with VoiceOver. Results below.

> In voiceover, once the suggestions are shown they completely take over the focus. Changes to the input state are no longer announced, instead on every input update I am being informed of the top suggestion.

This is still happening. Once I am on suggestion, changes to the input state are not announced, instead on every keystroke the currently selected suggestion is announced.

--- 

While using VoiceOver, there's no way to detect that the list of suggestions is visible. The only thing that is announced when I type in a few characters (I can see the suggestions) is this:

Screenshot 2025-06-26 at 11 07 18

Maybe we should consider using combobox role instead of a classic input?

plugins/woocommerce/client/legacy/js/frontend/test/address-autocomplete.js
```diff
@@ -0,0 +1,1337 @@
...
+			expect( suggestionsList.innerHTML ).toBe( '' );
+			expect( mockProvider.search ).not.toHaveBeenCalled();
+		} );
+
+		test( 'should hide suggestions when input goes from 3+ characters to less than 3', async () => {
+			// First show suggestions with 3+ characters
+			billingAddressInput.value = '123';
+			billingAddressInput.focus();
+			billingAddressInput.dispatchEvent( new Event( 'input' ) );
+			await new Promise( ( resolve ) => setTimeout( resolve, 150 ) );
```
straku, line 643: I wonder whether we could use testing library user-event in this environment?

opr:
> I am not getting any bolding on matched parts and I think it should happen based on the implementation 🤔 @opr Is this related to the mocked provider or is matching done fully client-side?
> 
> Screenshot 2025-06-26 at 10 52 10
> 
> EDIT: I know it should come from the provider. Any way for me to test it?

It's actually there, just difficult to see, I guess we should increase the weight 🤔 

This is how it looks in Safari 
image

opr:
Closing in favour of https://github.com/woocommerce/woocommerce/pull/59189

Issue #56558, opened 2025-03-20
Blockified Add to Cart with Options block: move blocks files in `/src/Blocks/BlockTypes` into a subfolder folder


Aljullu:
Currently, we have nine blocks which are specific to the blockified _Add to Cart with Options_ block:

https://github.com/woocommerce/woocommerce/tree/537650180b902897980b67387943198f788d36b1/plugins/woocommerce/src/Blocks/BlockTypes

I suggest we create a folder similar to `Accordion` and `OrderConfirmation` so we have all the _Add to Cart with Options_ files in the same subfolder.

Aljullu:
While working on this, we might evaluate creating a Utils class to encapsulate this check that we have in several of the blocks: `ProductType::SIMPLE === $product_type && ( ! $product->is_in_stock() || ! $product->is_purchasable() )`.

amitraj2203:
Hi @Aljullu, since [#57442](https://github.com/woocommerce/woocommerce/pull/57442) introduces the `AddtoCartWithOptions` folder along with a `Utils` class, I wanted to check if I can take up this issue and start working on it—assuming you haven’t already started.

Aljullu:
Sure, thanks! I assigned it to you, @amitraj2203.

Aljullu:
@amitraj2203 just a quick note: in #57442 I noticed the folder is called `AddtoCartWithOptions`. As you work on this issue, do you think we could rename it to `AddToCartWithOptions` (with capital `T` in `To`)? This way, it's easier to distinguish between words and we keep it consistent with the file names.

amitraj2203:
Hi @aljullu, apart from moving the files into a subfolder, I have made these additional improvements:

1. Created an abstract base class named `AbstractAddToCartWithOptionsBlock.php` in the new subfolder and moved all the methods and common functions there.
2. Renamed the other files to remove the `AddToCartWithOptions` prefix, as they're already in a dedicated folder.
3. Fixed script issues:
   - Removed unused `viewScriptModule` from `add-to-cart-with-options-grouped-product-selector` and `add-to-cart-with-options-grouped-product-selector-item-cta` block JSON files.
   - Added the missing `style` file for `add-to-cart-with-options-grouped-product-selector-item` block.
4. In `Utils` class I have kept only methods that don’t need state from the block classes.
5. Removed `wp_enqueue_script_module( $this->get_full_block_name() );` from the `quantity-selector` block.
[...]

PR #57442, opened 2025-04-21
[Experimental] Blockified Update Quantity Selector and Grouped Product Selector to use Stepper mode only

### LINKED ISSUE #56686: Blockified Add to Cart with Options block: Quantity Selector and Grouped Product Selector Item CTA blocks should only have Stepper mode
The Quantity Selector block currently supports _Input_ and _Stepper_ mode. The Grouped Product Selector Item CTA block doesn't have that option and it always renders the _Input_ (see https://github.com/woocommerce/woocommerce/issues/56287).

The issue is about updating the two blocks so they always render the _Stepper_ mode. We won't be allowing the _Input_ mode.

amitraj2203:

This PR updates both the `Quantity Selector` block and the `Grouped Product Selector Item CTA` block to exclusively use Stepper mode for quantity inputs. The Input mode option has been removed completely from both blocks.

##### Shared:
- Introduced `components/quantity-stepper/index.tsx` – a shared component.
- Moved shared logic into `add-to-cart-with-options` > `frontend.ts`, `style.scss`, and `editor.scss`.
- Created `Utils.php` in `BlockTypes/AddtoCartWithOptions/` for shared PHP logic.
- Removed unused frontend scripts, classes, attributes, and styles.

##### Quantity Selector Block:
- Removed `quantitySelectorStyle` attribute and all related logic.
- Updated rendering and styles to only use the stepper layout.
- Replaced duplicated scripts with centralized utilities.

##### Grouped Product Selector Item CTA Block:
- Refactored to reuse `<QuantityStepper />`.
- Cleaned up redundant CSS and JS.
- Removed frontend script and replaced it with shared logic.

Closes #56686 

| Block                | Before | After |
| --------------| ------ | ----- |
| Quantity Selector (Editor)              | before-editor-quantity-selector | after-editor-quantity-selector |
| Quantity Selector (Frontend)              | before-frontend-quantity-selector | after-frontend-quantity-selector |
| Grouped Product Selector Item CTA (Editor)              | before-editor-grouped-product-slector | after-editor-grouped-product-slector |
| Grouped Product Selector Item CTA (Frontend)              | before-frontend-grouped-product-slector | after-frontend-grouped-product-slector |

1. Install and activate the `WooCommerce Beta Tester` plugin.
2. Go to *Tools > Helper > Features* and enable the `experimental-blocks` and `blockified-add-to-cart` options.
3. Go to `Appearance > Editor > Templates` and select the `Single Product` template.
4. Select the `Add to cart` block, then click on the *Upgrade to the blockified Add to Cart with Options block* link in the inspector panel.
5. With the new `Add to Cart with Options` block selected, click on the `Switch product type` button in the toolbar and select either `Single Product` or `Variable Product`.
6. Now select the `Quantity Selector (Experimental)` block and verify that:
   - The toggle for changing the input mode has been removed
   - The stepper mode (with + and - buttons) appears by default
7. To test the `Grouped Product Selector Item CTA (Experimental)` block:
   - Select the `Add to Cart with Options` block
   - Click on the `Switch product type` button in the toolbar
   - Select `Grouped Product`
   - Verify that the quantity selector now shows in stepper mode with + and - buttons
8. To verify the frontend appearance:
   - Go to `Products > All Products` from the admin dashboard
   - View any simple product to test the `Quantity Selector (Experimental)` block
   - View any grouped product to test the `Grouped Product Selector Item CTA (Experimental)` block
   - Confirm both blocks display with the stepper interface instead of the input style
9. Test the functionality by clicking the + and - buttons to make sure they properly increment and decrement the quantity values.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-cta/editor.scss
```diff
@@ -0,0 +1,12 @@
+.wc-block-add-to-cart-with-options-grouped-product-selector-item-cta {
+	.quantity .qty {
+		// WooCommerce core styles: https://github.com/woocommerce/woocommerce/blob/c9f62609155825cd817976c7611b9b0415e90f2f/plugins/woocommerce/client/legacy/css/woocommerce.scss/#L111-L112
+		width: 3.631em;
+		text-align: center;
+	}
+
+	.wc-block-components-quantity-selector {
+		margin: 0 4px 0 0;
```
kmanijak: What is this right margin for? It creates this gap that looks buggy:

image

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-cta/editor.scss
```diff
@@ -0,0 +1,12 @@
...
+		width: unset !important;
```
kmanijak: Also, I see `width: unset;` is applied in `styles.scss` so I think it's a duplication.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-cta/editor.scss
```diff
@@ -0,0 +1,12 @@
+.wc-block-add-to-cart-with-options-grouped-product-selector-item-cta {
+	.quantity .qty {
+		// WooCommerce core styles: https://github.com/woocommerce/woocommerce/blob/c9f62609155825cd817976c7611b9b0415e90f2f/plugins/woocommerce/client/legacy/css/woocommerce.scss/#L111-L112
+		width: 3.631em;
+		text-align: center;
+	}
```
kmanijak: These styles are applied in editor anyway, is there a need to repeat them here?

Screenshot 2025-04-22 at 14 24 23

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/quantity-selector/block.json
```diff
@@ -1,15 +1,8 @@
 {
 	"name": "woocommerce/add-to-cart-with-options-quantity-selector",
 	"title": "Quantity Selector (Experimental)",
-	"description": "Display an input field to select the number of products to add to cart.",
+	"description": "Display an input field with buttons to select the number of products to add to cart.",
```
kmanijak: Not a strong opinion but maybe we could call it stepper here as well? WDYT?

```suggestion
	"description": "Display an input field with stepper to select the number of products to add to cart.",
```

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-cta/frontend.ts
```diff
@@ -2,8 +2,97 @@
...
+	'I acknowledge that using a private store means my plugin will inevitably break on the next store release.';
+
+const getInputElementFromEvent = (
+	event: HTMLElementEvent< HTMLButtonElement >
+) => {
+	const target = event.target as HTMLButtonElement;
+
+	const inputElement = target.parentElement?.querySelector(
+		'.input-text.qty.text'
+	) as HTMLInputElement | null | undefined;
```
kmanijak: AFAIK it either returns element or `null`:

```suggestion
	) as HTMLInputElement | null;
```

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsQuantitySelector.php
```diff
@@ -169,19 +147,17 @@ protected function render( $attributes, $content, $block ) {
...
-		$parsed_attributes  = $this->parse_attributes( $attributes );
-		$product_html       = $is_stepper_style ? $this->add_stepper_classes( $product_html ) : $product_html;
 		$classes_and_styles = StyleAttributesUtils::get_classes_and_styles_by_attributes( $attributes, array(), array( 'extra_classes' ) );
 
 		$classes = implode(
 			' ',
 			array_filter(
 				array(
-					'wp-block-add-to-cart-with-options-quantity-selector wc-block-add-to-cart-with-options__quantity-selector',
+					'wp-block-add-to-cart-with-options-quantity-selector wc-block-add-to-cart-with-options__quantity-selector wc-block-add-to-cart-with-options__quantity-selector--stepper',
```
Aljullu: I think we can remove the `wc-block-add-to-cart-with-options__quantity-selector--stepper` class now. It's not used anywhere. :slightly_smiling_face:

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/quantity-selector/edit.tsx
```diff
@@ -1,62 +1,16 @@
...
-		quantitySelectorStyle === QuantitySelectorStyle.Input
-			? 'wc-block-add-to-cart-with-options__quantity-selector--input'
-			: 'wc-block-add-to-cart-with-options__quantity-selector--stepper';
+import { Disabled } from '@wordpress/components';
 
+const AddToCartWithOptionsQuantitySelectorEdit = () => {
 	const blockProps = useBlockProps( {
-		className: `wc-block-add-to-cart-with-options__quantity-selector ${ quantitySelectorStyleClass }`,
+		className:
+			'wc-block-add-to-cart-with-options__quantity-selector wc-block-add-to-cart-with-options__quantity-selector--stepper',
```
Aljullu: Same here, given that we aren't using the `wc-block-add-to-cart-with-options__quantity-selector--stepper` for styling, let's remove it. `wc-block-add-to-cart-with-options__quantity-selector` should be enough.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsQuantitySelector.php
```diff
@@ -54,7 +34,7 @@ protected function enqueue_assets( $attributes, $content, $block ) {
 	 *
 	 * @param string $product_html Quantity input HTML.
 	 * @param string $product_name Product name.
-	 * @return stringa Quantity input HTML with increment and decrement buttons.
+	 * @return string Quantity input HTML with increment and decrement buttons.
```
Aljullu: Good catch! :smile:

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-cta/frontend.ts
```diff
@@ -2,8 +2,97 @@
...
+		'woocommerce/add-to-cart-with-options',
+		{},
+		{ lock: universalLock }
+	);
 
 store( 'woocommerce/add-to-cart-with-options', {
-	state: {},
-	actions: {},
+	actions: {
+		addQuantity: ( event: HTMLElementEvent< HTMLButtonElement > ) => {
```
kmanijak: I suggest action names as `increase/decreaseQuantity` instead of `add/removeQuantity`.

Aljullu (commented):
Great job here, @amitraj2203! :clap: 

Overall it's looking good, but answering your comment on the issue:

> Additionally, I feel that some of the logic is duplicated between the `Quantity Selector` and `Grouped Product Selector Item CTA` blocks and could potentially be combined. Let me know your thoughts on this.

Right, I had the same feeling as I was reviewing the PR. :smile: I think it would be nice to handle it as part of this PR, so we try to avoid duplicated logic before merging.

For the PHP side I think we might need to create a Utils file similar to the [one from the Product Collection block](https://github.com/woocommerce/woocommerce/blob/7edde48eaf24e719ade0c1fcb1446d0aaf06c019/plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/Utils.php), so we can use the same methods in different blocks.

For the JS code, I think we could move it directly into [the _Add to Cart with Options_ frontend.ts file](https://github.com/woocommerce/woocommerce/blob/7edde48eaf24e719ade0c1fcb1446d0aaf06c019/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts). I have not tried it, but I think having the `addQuantity()` and `removeQuantity()` actions there will make it possible for any block to use them.

For the CSS code, I think we could do something similar. Moving it to the parent block [`style.scss`](https://github.com/woocommerce/woocommerce/blob/7edde48eaf24e719ade0c1fcb1446d0aaf06c019/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/edit.scss) and [`editor.scss`](https://github.com/woocommerce/woocommerce/blob/7edde48eaf24e719ade0c1fcb1446d0aaf06c019/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/edit.scss) files, so the styles are always loaded.

As mentioned, I didn't test it, so if you have a better idea of how to avoid duplication, I'm open to it. :slightly_smiling_face:

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-cta/frontend.ts
kmanijak, line 1: I can't help but notice we're duplicating `plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/quantity-selector/frontend.ts`.

It clearly shows that there should be a way to unify and share lots of stuff between these two blocks or reuse one of them. I think`quantity-selector` is more atomic and should be reused in here. That would probably apply to `edit.tsx` as well as the markup is the same.

With the current approach any change to the quantity input requires duplicated work (and knowledge that is has to be updated in two places).

kmanijak (dismissed):
Thanks for the contribution! I tested the contribution and it works fine however when reviewing the code and couldn't help but notice significant code duplication  (iAPI code, markup, styles) between `grouped-product-selector/product-item-cta` and `quantity-selector`. Looks like the only difference is `grouped-product-selector/product-item-cta` can render:
- checkbox
- link
- or Quantity Selector.

Having said that, I paused the review and would like to suggest unifying the duplication and either extracting the code to share between those two or reusing Quantity Selector code in `grouped-product-selector/product-item-cta`.

What do you think about it @amitraj2203?

Also, @Aljullu - apologies for the ping, but wanted to confirm if you agree with the above. Maybe, considering the block roadmap and plans the code duplication would be acceptable and can be addressed later on? If so, I'd resume the review but otherwise I think most of this code should be reused. 🙌

Aljullu:
> Also, @Aljullu - apologies for the ping, but wanted to confirm if you agree with the above. Maybe, considering the block roadmap and plans the code duplication would be acceptable and can be addressed later on? If so, I'd resume the review but otherwise I think most of this code should be reused. 🙌 

Thanks for the review, @kmanijak! Yes, I definitely agree, and it's something @amitraj2203 also highlighted in the issue (https://github.com/woocommerce/woocommerce/issues/56686#issuecomment-2820333366). I agree that we should solve the duplication in this PR before merging. :slightly_smiling_face:

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-cta/frontend.ts
amitraj2203: @kmanijak @Aljullu I'm thinking of refactoring the structure like this—please let me know if it looks good to you or not:
```
plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/
├── components/                               # Shared component directory
│   └── quantity-stepper/                     # New component directory
│       └── index.ts                          # Main component
├── grouped-product-selector/
│   └── product-item-cta/
│       ├── edit.tsx                          # Will use the <QuantityStepper/> component
│       ├── frontend.ts                       # Will be removed
│       └── ...
├── quantity-selector/
│   ├── edit.tsx                              # Will use the <QuantityStepper/> component
│   ├── frontend.ts                           # Will be removed
│   └── ...
├── frontend.ts                               # Move the addQuantity and removeQuantity actions (which will be renamed to increaseQuantity/decreaseQuantity) here
├── style.scss                                # Common styles from both blocks will be moved here
└── editor.scss                               # Common styles from both blocks will be moved here

```

- The `<QuantityStepper/>` component will [specifically represents the stepper-style quantity input control](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/quantity-selector/edit.tsx#L121-L146)
- On the PHP side, as mentioned by @Aljullu, I will create a new directory in [BlockTypes](https://github.com/woocommerce/woocommerce/tree/trunk/plugins/woocommerce/src/Blocks/BlockTypes) named `AddtoCartWithOptions`, and create a class `Utils.php` which will consolidate functions from both the `Quantity Selector` and `Grouped Product Selector Item CTA` blocks.
- Later on, we can move all the block files into the `AddtoCartWithOptions` subfolder, as outlined in [issue](https://github.com/woocommerce/woocommerce/issues/56558)

Aljullu: It sounds like a solid plan, thanks @amitraj2203! :100:

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-cta/editor.scss
amitraj2203: Not sure 🤔 it was there in the `variation-selector` block and I picked it from there but now I've removed it as it was redundant.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-cta/editor.scss
amitraj2203: Yup, removed it.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-cta/editor.scss
amitraj2203: The last one is coming from the `add-to-cart-form` block.

amitraj2203:
Hi @Aljullu, I'm not sure why [these](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/quantity-selector/edit.tsx#L102-L110) styles are applied here but it's affecting the styles in the post editor. For now I've kept it in the new `<QuantityStepper/>` component as it is but let me know if it needs to me removed.
post-editor-quantity-selector

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-cta/frontend.ts
amitraj2203: I've resolved all the feedbacks and done with the refactoring, let me know if something need to be changed. Thanks!

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/style.scss
Aljullu, line 1: I would like if we can wrap all the styles in this file with the two blocks parent classes, just to make sure they don't leak into other blocks:

```CSS
.wc-block-add-to-cart-with-options__quantity-selector,
.wc-block-add-to-cart-with-options-grouped-product-selector-item-cta {
	/* ... all the styles here */
}
```

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsGroupedProductSelectorItemCTA.php
```diff
@@ -106,19 +96,16 @@ private function get_checkbox_markup( $product ) {
...
-		if ( ! empty( $post_id ) ) {
-			$product = wc_get_product( $post_id );
-		}
+		$product_data = Utils::get_product_from_context( $block, $previous_product );
+		$product      = $product_data['product'];
+		$markup       = '';
 
-		$markup = '';
+		if ( $product_data['is_valid'] ) {
+			wp_enqueue_script_module( $this->get_full_block_name() );
```
Aljullu: I think this is no longer needed because the Grouped Product Selector Item CTA block no longer has a frontend script, no?

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsGroupedProductSelectorItemCTA.php
```diff
@@ -136,4 +123,15 @@ protected function render( $attributes, $content, $block ): string {
...
+	/**
+	 * Enqueue frontend script
+	 *
+	 * @param array    $attributes Block attributes.
+	 * @param string   $content Rendered block output.
+	 * @param WP_Block $block Block instance.
+	 */
+	protected function enqueue_scripts( $attributes = array(), $content = '', $block = null ) {
+		wp_enqueue_script_module( 'woocommerce/add-to-cart-with-options-quantity-selector' );
+	}
```
Aljullu: Same here. :slightly_smiling_face:

plugins/woocommerce/src/Blocks/BlockTypes/AddtoCartWithOptions/Utils.php
```diff
@@ -0,0 +1,128 @@
...
+		$post_id = isset( $block->context['postId'] ) ? $block->context['postId'] : '';
+		$product = null;
+
+		if ( ! empty( $post_id ) ) {
+			$product = wc_get_product( $post_id );
+		}
+
+		if ( ! $product instanceof \WC_Product ) {
+			$product = $previous_product;
+		}
```
Aljullu: Given that this method can be called from anywhere, it might be good to check that `$previous_product` exists and is an instance of `WC_Product`.

Aljullu (commented):
This is looking great, @amitraj2203, good job!

One more thing that we can clean up now is the assignment of the `quantitySelectorStyle` attribute when downgrading from the blockified _Add to Cart with Options_ block to the previous version:

https://github.com/woocommerce/woocommerce/blob/8e83a294cd08ad0f1f81dca1c257bac1257b3527/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/components/downgrade-notice/index.tsx#L36-L38

It was checking if it was _Input_ or _Stepper_ mode, but now it's no longer necessary. :slightly_smiling_face: 

Besides that, I left a few minor comments inline, but overall looks good. :ok_hand: 

> Hi @Aljullu, I'm not sure why [these](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/quantity-selector/edit.tsx#L102-L110) styles are applied here but it's affecting the styles in the post editor. For now I've kept it in the new `<QuantityStepper/>` component as it is but let me know if it needs to me removed.

Apparently that was introduced in this PR: https://github.com/woocommerce/woocommerce/pull/51662. I just tested, and I can't reproduce the issue anymore in WP 6.8, so I would remove that code.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptionsGroupedProductSelectorItemCTA.php
amitraj2203: Oops I missed that 😅, removed it now.

plugins/woocommerce/src/Blocks/BlockTypes/AddtoCartWithOptions/Utils.php
amitraj2203: Done

plugins/woocommerce/src/Blocks/BlockTypes/AddtoCartWithOptions/Utils.php
```diff
@@ -0,0 +1,128 @@
...
+		if ( ! $product instanceof \WC_Product && $previous_product instanceof \WC_Product ) {
+			$product = $previous_product;
+		}
+
+		$is_valid_product = $product instanceof \WC_Product;
+
+		return array(
+			'product'  => $product,
+			'is_valid' => $is_valid_product,
+		);
```
Aljullu: I would prefer if this method only returned `$product`. If the product isn't valid, we can simply return `null`. Something like this:

```suggestion
		return $product instanceof \WC_Product ? $product : null;
```

You will need to update the functions where this method is consumed. But IMO it's better to keep the signature of the method simpler.

Aljullu (commented):
This looks great, @amitraj2203! I noticed there are some linting issues, will you be able to fix them?

Besides that, I left one last inline comment, but everything else looks good and is testing as expected, thanks!

plugins/woocommerce/src/Blocks/BlockTypes/AddtoCartWithOptions/Utils.php
amitraj2203: Done and fixed the linting error.

Aljullu (commented):
Everything looks great on my end now, thanks for all the cleaning up involved in this PR, @amitraj2203! :raised_hands: 

Apparently, there is still one Linting issue. Will you be able to look into it? Thanks!

<details>

* Theme: Twenty Twenty-Five
* Store size: small
</details>

kmanijak (commented):
@amitraj2203, I'll leave the reviews with @Aljullu who has been overlooking this daily but I just came here to say how amazing work you've done! It's amazing to see PRs adding features when cutting down lines of code! 🙇 Thanks for your contributions and responsiveness to comments! 🙌

Aljullu (approved):
> @amitraj2203, I'll leave the reviews with @Aljullu who has been overlooking this daily but I just came here to say how amazing work you've done! It's amazing to see PRs adding features when cutting down lines of code! 🙇 Thanks for your contributions and responsiveness to comments! 🙌 

100% agree with this. :raised_hands:

Aljullu:
@kmanijak we will need you to withdraw your review requesting changes in order to be able to merge. :pray:

PR #57446, opened 2025-04-22
Mini-Cart Contents block: Fix Mini-Cart Shopping Button color styling

### LINKED ISSUE #56074: Mini-Cart Contents block: Mini-Cart Shopping Button background color isn't set properly
### Prerequisites

### Describe the bug

When setting the background color of the _Mini-Cart Shopping Button_ block, it change the color outside of the button instead of the button background.

### Expected behavior

When changing the background color of the _Mini-Cart Shopping Button_ block, the background of the button should change, instead of the background outside of the button. Changing the text color should also be applied properly.

### Actual behavior

When changing the background color of the _Mini-Cart Shopping Button_ block, the background of the button doesn't change, but the background outside of the button does change. Changing the text color does nothing.

### Steps to reproduce

1. Go to Appearance > Editor > Patterns > Mini-Cart.
2. Open the List View and select the _Empty Mini-Cart view_ block.
3. Select the _Mini-Cart Shopping Button_ and change the text color and background color.
4. Notice the text color doesn't change, while the background color is set outside of the button, not inside it.

### WordPress Environment

<details>

* WordPress 6.7.2 local dev environment
* Plugins: WooCommerce, WooCommerce Beta Tester
* Theme: Twenty Twenty-Four
* Store creation: February 2025
* Store size: small

</details>

### Isolating the problem

amitraj2203:

This PR fixes an issue with the `Mini-Cart Shopping Button` block where color settings (background and text colors) were being applied to the outer container div instead of the button itself.

Implemented the fix by:

1. Adding `__experimentalSkipSerialization: true` to the color support in block.json
2. Using the `__experimentalUseColorProps` hook to properly extract color styles
3. Applying these color props directly to the button element instead of the container
4. Adding proper style handling in the frontend block component to ensure colors appear correctly

Closes #56074 

| Block                | Before | After |
| --------------| ------ | ----- |
| Mini-Cart Shopping Button (Editor)              | before-editor-mini-cart-shoppting-button | after-editor-mini-cart-shoppting-button |
| Mini-Cart Shopping Button (Frontend)              | after-frontend-mini-cart-shoppting-button | before-frontend-mini-cart-shoppting-button |

1. Go to _Appearance > Editor > Patterns > Mini-Cart_.
2. Open the `Document Overview` and select the `Empty Mini-Cart view` > `Mini-Cart Shopping button` block.
3. Change the text color and background color from the `Inspector Controls`
4. Notice the text and background color is now changing and being applied to the button itself.
5. Test with both preset colors and custom colors from the color picker
6. To test in frontend, open any page which has the Mini-Cart block and click on the block _(Make sure the cart is empty)_
7. Verify that the same styles applied in the editor are also appearing correctly on the frontend.
> [!NOTE]
> If you've previously applied colors to this block, first clear all color settings by clicking the "Clear" button in the color picker panel. This is necessary because this PR changes where colors are applied (moving them from the outer container to the button itself).

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/mini-cart-contents/inner-blocks/mini-cart-shopping-button-block/block.tsx
```diff
@@ -14,24 +14,53 @@ import { getVariant } from '../utils';
...
+	const buttonStyles: Record< string, string > = {};
+
+	const parsedStyle =
+		typeof style === 'string' ? JSON.parse( style ) : style || {};
+	if ( parsedStyle?.color?.text ) {
+		buttonStyles.color = parsedStyle.color.text;
+	}
+	if ( parsedStyle?.color?.background ) {
+		buttonStyles.backgroundColor = parsedStyle.color.background;
+	}
```
tjcafferkey: We can simplify this by making it more consistent with other blocks by using the `useStyleProps` hook found [here](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/base/hooks/use-style-props.ts). I imagine it will look like this...

```jsx
	const styleProps = useStyleProps( {
		style,
		textColor,
		backgroundColor,
	} );

	return (
		<div className="wp-block-button has-text-align-center">
			<Button
				className={ clsx(
					className,
					styleProps.className,
					'wc-block-mini-cart__shopping-button'
				) }
				style={ styleProps.style }
				variant={ getVariant( className, 'contained' ) }
				href={ SHOP_URL }
			>
				{ startShoppingButtonLabel || defaultStartShoppingButtonLabel }
			</Button>
		</div>
	);

**Review by tjcafferkey — CHANGES_REQUESTED** [review]
Thanks for this! I can confirm this does fix the issue but I have left a comment to refactor some of the code so its simplified and more consistent with other blocks.

One thing to mention though, is that as far as I can tell the user doesn't need to clear and reapply the changes as per your note. This fix will automatically apply those outer container styles to the inner button once it has been released.

This means that if people relied on the 'bug' in the past then visually their empty cart button will automatically change once updating. Since this is in the empty mini cart, and fixes a bug I don't think we need to consider backcompat in this case as its not considered critical and can be easily fixed via the editor.

Once my suggestion has been implemented I think we can move forward. Thanks again :)

**Comment by amitraj2203** [comment]
> One thing to mention though, is that as far as I can tell the user doesn't need to clear and reapply the changes as per your note. This fix will automatically apply those outer container styles to the inner button once it has been released.

@tjcafferkey Thanks for the review! However, I tested the behavior again on my local setup, and **I still need to clear all previously applied color settings before using this fix**, otherwise the classes applied to the outer `div` don’t get removed.

## Here are the steps where I encountered this issue:

1. On the `trunk` branch, I opened _Appearance > Editor > Patterns > Mini-Cart_.
2. Opened the `Document Overview` and selected `Empty Mini-Cart view` > `Mini-Cart Shopping button` block.
3. Changed the text color and background color using the `Inspector Controls` (I'm using TT5 theme — I set the background color to `Accent-3` and the text color to `Accent-1`, which looked something like this):
   image
4. The background color was applied to the outer `div`, and the text color wasn’t applied anywhere.
5. Then I checked out this branch: `amitraj2203:fix/mini-cart-shopping-button-colors`, ran `pnpm build`, and reopened the `Mini-Cart Shopping button` block.
6. Now the background color is correctly applied to the button itself, and the text color is also reflected. However, the background color still remains on the outer `div`.
7. If I change the background color now, it updates correctly on the button (meaning it’s applied to the button), but the old background color still remains on the outer `div`. This might be happening because the class changes now affect only the button, while the outer `div` still retains the old classes that were set when I was on the `trunk` branch.

   The below screen recording shows the behavior after I applied the colors in the block while on the `trunk` branch, then checked out to the `amitraj2203:fix/mini-cart-shopping-button-colors` branch, ran `pnpm build`, and refreshed the editor.

   https://github.com/user-attachments/assets/26b23b76-5ae2-4aa5-bb27-5151d46e275b

Please let me know if I’m missing something!

File: plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/mini-cart-contents/inner-blocks/mini-cart-shopping-button-block/block.tsx (hunk 1 above)
Comment (amitraj2203) [reply plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/mini-cart-contents/inner-blocks/mini-cart-shopping-button-block/block.tsx]: I have implemented it that way now.

File: plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/mini-cart-contents/inner-blocks/mini-cart-shopping-button-block/block.tsx [hunk 2]
```diff
@@ -14,24 +15,37 @@ import { getVariant } from '../utils';
...
 const Block = ( {
 	className,
 	startShoppingButtonLabel,
+	style,
+	textColor,
+	backgroundColor,
 }: MiniCartShoppingButtonBlockProps ): JSX.Element | null => {
 	if ( ! SHOP_URL ) {
 		return null;
 	}
```
tjcafferkey: We will need to move this below the `useStyleProps` hook so that we don't get the linting error of hooks conditionally running.

tjcafferkey (changes requested):
I understand what you mean now thanks for the detailed explanation @amitraj2203. Given this change doesn't impact the _frontend_ of existing stores, and can be easily resolved in the editor I think that it's fine and a good fix so thank you for opening a PR.

I've added one comment for change as its being flagged by the CI Linter then I think we're good to merge.

Again, thanks for submitting this 🙌🏻

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/mini-cart-contents/inner-blocks/mini-cart-shopping-button-block/block.tsx
amitraj2203: Thanks for the review. I have resolved the lint error now 🙂

tjcafferkey: Running CI again :)

tjcafferkey (approved):
This change LGTM. Thank you for addressing my feedback. Approving this, should be good to merge once CI checks pass successfully

PR #57460, opened 2025-04-22
Fix radio label and description no taking all of available space

straku:

A small fix for a bug introduced in https://github.com/woocommerce/woocommerce/pull/57229.

When a flex container holds another element with flex display we need to stretch the latter to fill available space as it won't do it by default.

UPDATE: This can be only reproduced with Stripe payment being available in the list. The wrap doesn't happen for "direct bank transfer" and "cash on delivery" in the list.

I also moved when we apply a default value for `descriptionStackingDirection` prop to address a TypeScript error in `RadioControlAccordion`.

| Before | After |
| ------ | ----- |
| Screenshot 2025-04-22 at 16 20 25 | Screenshot 2025-04-22 at 15 27 23 |

1. You need a store with the following payment methods defined:
    - "cash on delivery" or "direct bank transfer" (ideally both)
    - any additional payment that is registered with a custom label (for example: Stripe payments)
3. In this store, add some items to the cart and proceed to block checkout
4. Verify that the list of payment options is readable and no text wrapping happens

I followed the above steps and also steps defined in the original PR that caused the issue.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

There's already a changelog for the original fix, this change is part of the original fix.

</details>


mikejolley (approved):
Thanks, I've confirmed this fixes the issue when Stripe is active.

I tested: 

1. Styling with Stripe active
2. Styling without Stripe
3. Mobile views

Looks good to go 👍🏻

PR #57468, opened 2025-04-22
Introducing payment gateway features constants

wjrosa:

This PR introduces constants for all the payment gateway features' values. The idea is to make it easier to reference them, avoid typos, make the code stricter, make it easier to find status usages, centralize them, improve documentation, and many other reasons that should help developers create code related to them.

- Checkout and build this branch on your test environment (`dev/introducing-payment-gateway-features-constants`)
    - Store setup
    - Product creation
    - Common checkout
    - Shortcode checkout
    - My account page navigation
    - Order listing/details

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Introduces and implements payment gateway features' constants in the WooCommerce plugin codebase.

</details>

<details>

#### Comment 

</details>


barryhughes:
Looks good, though I'll leave to Vlad and Flux to review, since I'm focused on other things right now. One passing thought is, should we be explicit that the class constants (in the 'enum' class) are public, like we would do for other class members such as methods and properties?

```php
class Foo {
    public const BAR = 'baz';
}
```

wjrosa:
> Looks good, though I'll leave to Vlad and Flux to review, since I'm focused on other things right now. One passing thought is, should we be explicit that the class constants (in the 'enum' class) are public, like we would do for other class members such as methods and properties?
> 
> ```
> class Foo {
>     public const BAR = 'baz';
> }
> ```

Thanks for looking into this, Barry! I have added them as reviewers here. About your suggestion, I think it makes total sense. Constant visibility is supported since PHP 7.1. I have added them in https://github.com/woocommerce/woocommerce/pull/57468/commits/a605290be137c6e6342a66a561ddd6d5685cfdec. And I will create another PR to update the existing classes.

plugins/woocommerce/src/Enums/PaymentGatewayFeatures.php
```diff
@@ -0,0 +1,136 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace Automattic\WooCommerce\Enums;
+
+/**
+ * Enum class for all the payment gateway feature's values.
+ */
+final class PaymentGatewayFeatures {
```
kalessil: Noticed we miss some constants: `$gateway->supports( 'pay_button' )`, can we cover it as well, please?

kalessil (changes requested):
Looks good overall: noticed a constant we missed - will do some smoke testing once we cover that.

PS: thanks for adding more enums!

vladolaru (commented):
Thank you for working on this, @wjrosa! These kind of changes push our product quality higher.

Apart for the missed feature @kalessil mentioned, I couldn't find other misses. The smoke testing that I've done (the list you provided), didn't reveal any issues.

plugins/woocommerce/src/Enums/PaymentGatewayFeatures.php
wjrosa: Oops, missed this one! Added in https://github.com/woocommerce/woocommerce/pull/57468/commits/683db24b3e1d2deb4245fc711a3079e798bfba88

kalessil (approved):
LGTM!

Since @vladolaru did heavy lifting already with testing, I only tested manual plugin update from zip-file :heavy_check_mark:

mohsinoffline:
Is there a specific reason for changing the 'add_payment_method' feature to 'add_payment_methods' while implementing the constants to be used for payment gateway features? Looks like both Stripe and WooPayments plugins still use 'add_payment_method' so I think this might be unintentional?

wjrosa:
That was not intentional. Thank you for flagging the issue, and I apologize for the bug!

@szepeviktor I appreciate the contribution. I will review your PR now 👀

PR #57469, opened 2025-04-23
Implement BlocksSharedState for shared WooCommerce block config/state.

samueljseay:

The goal is to move interactive blocks away from using the old `wcSettings` global and move to the simpler and more efficient `wp_interactivity_config`. 

To support the old shared configs from `wcSettings` we introduce `BlocksSharedState`. It's purpose is to initialize the necessary shared settings and state.

We move ProductFilters away from using `wcSettings` and have it initialize the shared config it needs and also the block specific config it needs via `wp_interactivity_config`.

We also move ProductButton to initializing the cart state via this class.

N/A

Smoke test the product filter blocks, particularly take note that adjusting them does not cause any browser exceptions and that they work correctly as before.

Also pay close attention to the price product filter that it works as before.

To do that:

1. Add a post or page or edit a template
2. Add product collection block
3. Within that block add product filters block
4. Save
5. Smoke test front end

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Add BlocksSharedState trait to store shared WooCommerce settings with wp_interactivity_config and wp_interactivity_state.

</details>

<details>

#### Comment 

</details>


luisherranz:
Re-sharing [this comment](https://github.com/woocommerce/woocommerce/pull/56888#discussion_r2055661159) here:

Perhaps we could combine this with the same API that we need to create when multiple blocks need to add the same state:

https://github.com/woocommerce/woocommerce/blob/381d06e0cd5368938c5f61c0e28ba0a763fb8ec7/plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php#L293-L309

In a way, state and config should work in a very similar way on the server.

samueljseay:
@luisherranz could you clarify what you mean a little bit more? 

Originally I thought you're asking for this to be implemented more like the code example you gave but now I re-read a few times I think you're suggesting we should put the code you gave alongside the code we wrote here. Is that right? 

Do you have a vision for what that API would look like or just a feeling that we need to consolidate it?

luisherranz:
> Do you have a vision for what that API would look like or just a feeling that we need to consolidate it?

A feeling that we need to consolidate it 🙂

This is going to be an API that will be exposed to the WooCommerce blocks, but also to the extenders, and it would be confusing if the API were very different for cases where it handles "wp_interactivity_config" than for cases where it handles "wp_interactivity_state". The fewer concepts extenders have to learn, the better.

The code example I've provided is a specific case of a similar need but for "wp_interactivity_state".

luisherranz:
To clarify, I'm not proposing here to replace the direct use of `wp_interactivity_config` or `wp_interactivity_state` for individual cases, but rather when multiple blocks have to repeatedly add the same config/state, as in the cases of the global wc settings that were being added by default, or the case of adding the cart to the state that I shared.

samueljseay:
@luisherranz yep that makes perfect sense to me, I understand we don't want to write a redundant wrapper around the wp_interactivity methods, but just make it easy to initialize this frequently used shared state.

I think it could be as simple as:

1. Rename the class to better capture that we're setting up both shared state and config
2. Add methods that would setup the shared state you linked above. Explicitly call them where we need the state in a block, as we do with the config.
3. Remove the old code you linked above.

Wdyt? There will likely only be a handful of cases where we need this kind of shared state to be setup outside of the blocks themselves so I don't anticipate this class growing much beyond that.

luisherranz:
Yes, that makes sense.

Would that class be public to extenders right now, or is it a class that they can't (or are not supposed to) use, and in the future, once we open up the extensibility, the methods of this class will be mapped to public functions? It's not something we have to decide exactly now, but I ask to have at least an initial idea of how things would be for extenders.

samueljseay:
@luisherranz Its not impossible for them to use it, we cannot stop that, but I would discourage them from using it (if only because we haven't really thought very hard about extensibility or beneficial public APIs here).

> once we open up the extensibility, the methods of this class will be mapped to public functions? It's not something we have to decide exactly now, but I ask to have at least an initial idea of how things would be for extenders.

That sounds reasonable to me yes.

plugins/woocommerce/src/Blocks/Utils/BlocksSharedState.php
```diff
@@ -0,0 +1,152 @@
...
+ * Intended to be used as a singleton.
+ */
+class BlocksSharedState {
+
+	/**
+	 * The consent statement for using private APIs of this class.
+	 *
+	 * @var string
+	 */
+	private $consent_statement = 'I acknowledge that using private APIs means my theme or plugin will inevitably break in the next version of WooCommerce';
```
samueljseay: @luisherranz I've taken the same approach as private APIs from Gutenberg, and required passing a consent string to use either of these methods. This should allow us not to worry about deprecating these methods or their side effects in future versions.

I don't expect 3PDs to use these methods but this does actively discourage them from doing so.

I've also consolidated the cart state here as well.

plugins/woocommerce/src/Blocks/BlockTypes/ProductQuery.php
```diff
@@ -180,9 +182,11 @@ public function update_query( $pre_render, $parsed_block ) {
...
 			// Indicate to interactivity powered components that this block is on the page
 			// and needs refresh to update data.
-			$this->asset_data_registry->add(
-				'needsRefreshForInteractivityAPI',
-				true
+			wp_interactivity_config(
+				'woocommerce',
+				[
+					'needsRefreshForInteractivityAPI' => true,
+				]
```
luisherranz, line 189:
```suggestion
				array( 'needsRefreshForInteractivityAPI' => true )
```

luisherranz (approved):
I like the changes so fee feel to merge, but what do you think about using more explicit methods to initialize just part of the config or the state? After all, we are not going to want to initialize everything every time, but most of the time we will be more selective, aren't we?

Something like:

- `blocks_shared_data->initialize_currency_data`
- `blocks_shared_data->initialize_cart`

Any thoughts on what the final API for the extenders could be? Maybe something like `woo_interactivity_initialize_cart`?

plugins/woocommerce/src/Blocks/BlockTypes/MiniCart.php
```diff
@@ -77,9 +78,10 @@ class MiniCart extends AbstractBlock {
 	 * @param AssetApi            $asset_api Instance of the asset API.
 	 * @param AssetDataRegistry   $asset_data_registry Instance of the asset data registry.
 	 * @param IntegrationRegistry $integration_registry Instance of the integration registry.
+	 * @param BlocksSharedState   $blocks_shared_state Instance of the blocks shared state class.
 	 */
-	public function __construct( AssetApi $asset_api, AssetDataRegistry $asset_data_registry, IntegrationRegistry $integration_registry ) {
-		parent::__construct( $asset_api, $asset_data_registry, $integration_registry, $this->block_name );
+	public function __construct( AssetApi $asset_api, AssetDataRegistry $asset_data_registry, IntegrationRegistry $integration_registry, BlocksSharedState $blocks_shared_state ) {
+		parent::__construct( $asset_api, $asset_data_registry, $integration_registry, $blocks_shared_state, $this->block_name );
```
dinhtungdu: The construction is doing nothing, why do we need it in the first place? 🤔

plugins/woocommerce/src/Blocks/BlockTypes/ProductQuery.php
dinhtungdu: `DisallowShortArraySyntax` is disabled for `src/Blocks`. We currently don't enforce any types.

plugins/woocommerce/src/Blocks/BlockTypesController.php
```diff
@@ -114,7 +124,7 @@ public function register_blocks() {
 		foreach ( $block_types as $block_type ) {
 			$block_type_class = __NAMESPACE__ . '\\BlockTypes\\' . $block_type;
 
-			new $block_type_class( $this->asset_api, $this->asset_data_registry, new IntegrationRegistry() );
+			new $block_type_class( $this->asset_api, $this->asset_data_registry, new IntegrationRegistry(), $this->blocks_shared_state );
```
dinhtungdu: This is not a suggestion as we're following the convention and it's okay.
Personally I want to just get all dependencies directly from the container instead of passing them as BlockTypesController deps just to pass them to block initialization.

dinhtungdu (commented):
Looking at the diff and the number of unrelated files (but had to be updated because we're changing class signature), I think we should consider using a trait before going with DI. I know I advocated for DI, but I'm wary about changing the block class signature for just a portion of blocks. Traits are more testable than static methods (eg, PHPUnit provides `getMockForTrait` method), makes it an attractive choice in this case to me.

plugins/woocommerce/src/Blocks/BlockTypesController.php
samueljseay: Yes I agree fwiw

plugins/woocommerce/src/Blocks/BlockTypes/MiniCart.php
samueljseay: We don't, I should have deleted this. Looks like it was added by accident in the past or maybe some other code was removed and the override was forgotten.

samueljseay: This doesn't matter now. I could remove the constructor here, but just feels too unrelated to the PR.

plugins/woocommerce/src/Blocks/BlockTypesController.php
samueljseay: Not relevant since we moved to trait

samueljseay:
> After all, we are not going to want to initialize everything every time, but most of the time we will be more selective, aren't we?

yes, for config I implemented it more like the original global vars where it's pretty much expected to always be setup and available, but I think we can rethink that in future.

> Something like:
> blocks_shared_data->initialize_currency_data
> blocks_shared_data->initialize_cart
> Any thoughts on what the final API for the extenders could be? Maybe something like woo_interactivity_initialize_cart

I'll have a think about how we can make this more granular. How about `woocommerce_register_cart_interactivity_state`? I think having state in the name feels a bit more descriptive. I'm also not sure.. if this is a trait, maybe we should drop the `woocommerce` part, if designed to use within a block perhaps `register_cart_interactivity_state` could be sufficient?

luisherranz:
> within a block perhaps register_cart_interactivity_state could be sufficient?

Yeah. Maybe drop `_state` because it could also do other stuff, like enqueue a JS file, add config…

> maybe we should drop the woocommerce part

Don't plugins need a namespace for global functions? Like `wc_register_cart_interactivity`?

We can keep `register` and `interactivity` as an indication so that it is consistent in the future:

- `(wc_)register_cart_interactivity`
- `(wc_)register_checkout_interactivity`

Perhaps we could also get closer to the Core syntax (`wp_interactivity_`) and change the order: `wc_interactivity_register_cart`

Anyway, it's okay to have a slight idea of how this could be exposed in the future, but there's no need to decide anything now.

samueljseay:
> Don't plugins need a namespace for global functions? Like wc_register_cart_interactivity?

@luisherranz If we expose this as a global function for extenders then you're right, but if we expose it as a trait within the AbstractBlock class or some future implementation of the AbstractBlock class I think the prefix is not necessary, I guess I was getting focussed on 3rd party extenders using the trait or using this from within a Block class we implemented :) yes if it's a global function we should prefix it.

> Yeah. Maybe drop _state because it could also do other stuff, like enqueue a JS file, add config…

🤔 ok yeah that's a good point. Even for some granular setup of cart we may want to enqueue some JS 👍🏻 

> We can keep register and interactivity as an indication so that it is consistent in the future...

`register_cart_interactivity` sounds good. I'll use this in the trait and it can be an indication of future naming as you say 👍🏻

plugins/woocommerce/src/Blocks/Utils/BlocksSharedState.php
```diff
@@ -0,0 +1,152 @@
...
+	/**
+	 * Check that the consent statement was passed.
+	 *
+	 * @param string $consent_statement - The consent statement string.
+	 * @return true
+	 * @throws \InvalidArgumentException - If the statement does not match the class consent statement string.
+	 */
+	private static function check_consent( $consent_statement ) {
+		if ( $consent_statement !== self::$consent_statement ) {
+			throw new \InvalidArgumentException( 'This method cannot be called without consenting the API may change.' );
```
dinhtungdu: We already imported `InvalidArgumentException` above.
```suggestion
			throw new InvalidArgumentException( 'This method cannot be called without consenting the API may change.' );
```

plugins/woocommerce/src/Blocks/BlockTypes/ProductQuery.php
```diff
@@ -180,9 +182,11 @@ public function update_query( $pre_render, $parsed_block ) {
 		if ( self::is_woocommerce_variation( $parsed_block ) ) {
 			// Indicate to interactivity powered components that this block is on the page
 			// and needs refresh to update data.
-			$this->asset_data_registry->add(
-				'needsRefreshForInteractivityAPI',
-				true
+			wp_interactivity_config(
+				'woocommerce',
+				[
+					'needsRefreshForInteractivityAPI' => true,
```
dinhtungdu, line 188: ClassicTemplate block also registers this setting and need to be replaced.

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/frontend.ts
```diff
@@ -190,12 +191,11 @@ const productFiltersStore = {
...
 
-			const isBlockTheme = getSetting( 'isBlockTheme' );
-			const isProductArchive = getSetting( 'isProductArchive' );
-			const needsRefreshForInteractivityAPI = getSetting(
-				'needsRefreshForInteractivityAPI',
-				false
-			);
+			const sharedSettings = getConfig( 'woocommerce' );
+			const isBlockTheme = sharedSettings?.isBlockTheme || false;
+			const isProductArchive = sharedSettings?.isProductArchive || false;
```
dinhtungdu: `isProductArchive` doesn't belong to shared setting but `woocommerce/product-filters` only.

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilters.php
```diff
@@ -35,10 +39,14 @@ protected function enqueue_data( array $attributes = array() ) {
...
-		$this->asset_data_registry->add( 'isBlockTheme', wp_is_block_theme() );
-		$this->asset_data_registry->add( 'isProductArchive', is_shop() || is_product_taxonomy() );
-		$this->asset_data_registry->add( 'isSiteEditor', 'site-editor.php' === $pagenow );
-		$this->asset_data_registry->add( 'isWidgetEditor', 'widgets.php' === $pagenow || 'customize.php' === $pagenow );
+		$this->initialize_shared_config( 'I acknowledge that using private APIs means my theme or plugin will inevitably break in the next version of WooCommerce' );
+
+		wp_interactivity_config(
+			$this->get_full_block_name(),
+			[
+				'isProductArchive' => is_shop() || is_product_taxonomy(),
```
dinhtungdu: Can we take this chance and also add `is_search() && 'product' === get_post_type()` to the mix? I'm fine with doing it separately too, as this is unrelated.

dinhtungdu (changes requested):
Great work! Thank you for all the updates. This is looking much better now. There are some small issue with filter block then this is ready to go.

plugins/woocommerce/src/Blocks/BlockTypes/ProductQuery.php
samueljseay: Thanks for spotting that 🙇

samueljseay: Resolved in [a8504e9](https://github.com/woocommerce/woocommerce/pull/57469/commits/a8504e91c4561b75edf36ddf4fc9bf1ca93e93f6)

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/frontend.ts
samueljseay: Whoops, good spotting.

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilters.php
samueljseay: Easy enough to enhance this here. Will add.

samueljseay: @dinhtungdu should it be `&&` or `||`?

ie this:

`'isProductArchive' => ( is_shop() || is_product_taxonomy() ) && is_search() && 'product' === get_post_type(),`

or

`'isProductArchive' => is_shop() || is_product_taxonomy() || ( is_search() && 'product' === get_post_type() ),`

dinhtungdu: The latter :)

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/frontend.ts
samueljseay: Resolved in https://github.com/woocommerce/woocommerce/pull/57469/commits/0e589f9657e14d690599c34ca9505fe04c940700

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilters.php
samueljseay: Haha in retrospect thats probably obvious. Fixed in https://github.com/woocommerce/woocommerce/pull/57469/commits/a4943ec916994f38211b1dfdcaf05a371ae311c0

samueljseay:
I've done some follow up manual testing including checking that `isProductArchive` is present on the correct settings object in front-end. I think this should be ready for re-review now.

dinhtungdu (approved):
Thanks for the update! I tested Product Collection and Product Filters against classic and block themes, everything works as expected. This PR looks is ready to go from my end. :shipit: 


#### 🖼️ Screenshots/ screen recordings

image
image

PR #57487, opened 2025-04-23
Fix: enable filter blocks in the Widgets editor

dinhtungdu:

This PR enable new filter blocks in the Widgets editor so classic themes can use the new filter blocks.

Fixes WOOPLUG-NEW

Note that there is a styling issue in the Widgets editor that will be fixed in [WOOPLUG-4051: Product Filters: review the usage of `:where` selector.](https://linear.app/a8c/issue/WOOPLUG-4051/product-filters-review-the-usage-of-where-selector).

1. Active a classic theme like Storefront.
2. Go to Appearance > Widgets.
3. Verify that you can search and add Product Filters block to widget areas.
4. Smoke test and verify that the block works as expected in the editor.
5. Go to the shop page and verify that the block also functions as expected: you can set/unset a filter, set/unset multiple filters, clear the selected filters, the product grid is updated when filters applied. A new page will be loaded everytime filters change.

### DIFF
--- a/plugins/woocommerce/src/Blocks/BlockTypesController.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypesController.php
@@ -371,9 +371,7 @@ final class BlockTypesController {
 	 */
 	protected function get_widget_area_block_types() {
 		return array(
-			'ActiveFilters',
 			'AllReviews',
-			'AttributeFilter',
 			'Breadcrumbs',
 			'CartLink',
 			'CatalogSorting',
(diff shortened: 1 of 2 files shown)


kmanijak (approved):
I tested the new filters on:
- Storefront
- TT1
and in both cases Product Filters are available in the Widgets inserter and they work correctly (I used multiple mix of filters).

---

I found some weird checkbox artefact in TT1 but I don't consider it a blocker at all especially with this mention:

> Note that there is a styling issue in the Widgets editor that will be fixed in [WOOPLUG-4051: Product Filters: review the usage of :where selector.](https://linear.app/a8c/issue/WOOPLUG-4051/product-filters-review-the-usage-of-where-selector).

Editor | FE
-- | --
image | image

plugins/woocommerce/src/Blocks/BlockTypesController.php
```diff
@@ -401,6 +401,17 @@ protected function get_widget_area_block_types() {
...
+			'ProductFilterStatus',
+			'ProductFilterPrice',
+			'ProductFilterPriceSlider',
+			'ProductFilterAttribute',
+			'ProductFilterRating',
+			'ProductFilterActive',
+			'ProductFilterRemovableChips',
+			'ProductFilterClearButton',
+			'ProductFilterCheckboxList',
+			'ProductFilterChips',
```
kmanijak, line 398: Ah! I just spotted this AFTER review, please move the `ProductFilters*` above line 394 which says:

> // Below product grids are hidden from inserter however they could have been used in widgets.
> // Keep them for backward compatibility.

This was used specifically to separate and highlight product grid blocks within the list.

dinhtungdu: Right, I updated the list to make it clearer in https://github.com/woocommerce/woocommerce/pull/57487/commits/fa9fe8ca38959f5248d5b93d5f4ba5ceff2935b9. Let me know what you think.

kmanijak: Nice! I like additional section with legacy filters! 💪 Ship! 🚢

PR #57490, opened 2025-04-23
Added Featured product & category image background warning notice

### LINKED ISSUE #56137: [Enhancement]: Featured Product and Featured Category: background setting visibility

The background setting appears to function correctly, but it is only visible when the product or category does not have a featured image. This behavior is unclear and may cause confusion.

1. Create a new post.
2. Add the Featured Product and Featured Category blocks.
3. For both blocks, select an item that has a **featured image** set.
4. Via block settings, update the background color.
5. If the image covers all the block, the background color set isn't visible.

https://github.com/user-attachments/assets/b3c46963-47ef-48d9-855d-ff1d7513dc93

Kudos to @Aljullu [for sharing the video](https://github.com/woocommerce/woocommerce/issues/42343#issuecomment-2700532889)

_No response_

### Additional context

_No response_

Kallyan01:

### Changes Proposed in this Pull Request:

- Added transparency detection logic.
- Added dimension check incase image size is smaller than parent container size.
- Added a dismissable notice incase it's detected the background color can't be visible with the image.

#### Issue Description:

The background color setting appears to function correctly, but it is only visible when the product or category does not have a featured image , transparent image or have image which dosen't cover the full height / width of the parent.

#### Issue root cause:

- There is no transprency & dimension  detection logic present for the product / category image to check the background color visibility. So whenever some product image has non-transparent image and background color set this bug appeares.

Closes #56137 

### Before

https://github.com/user-attachments/assets/2fcf2c84-b214-4817-8eb1-dfbc3a4d94b3

### After

https://github.com/user-attachments/assets/5cccf029-21de-4627-97de-0e06ad983881

1. Create a new post.
2. Add the Featured Product and Featured Category blocks.
3. For both blocks, select an item that has a featured image set.
4. Via block settings, update the background color.
5. If the image covers all the block, check if the background color set is visible.
6. Try toggling the Image Fit and Repeated Background options .

1. Transparent Image 
- Small image -> No Notice 
- Covered image -> No Notice 
- Repeated image -> No Notice 

2. Non-Transparent Image
- Small image -> No Notice 
- Covered image -> Notice Present
- Repeated image -> Notice Present

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
```diff
@@ -114,17 +116,75 @@ export const withFeaturedItem =
...
+						'.featured-product-parent'
+					) as HTMLElement | null;
+
+				if ( ! element ) return;
+
+				observer.observe( element );
+			}
+
+			return () => observer.disconnect();
+		}, [ isLoading ] );
```
gigitux: Is there a particular reason for this logic? Can't we compute the height and width inside the useEffect?

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
```diff
@@ -114,17 +116,75 @@ export const withFeaturedItem =
...
+
+		useEffect( () => {
+			if (
+				! backgroundColorVisibility &&
+				backgroundColor &&
+				! isLoading
+			) {
+				dispatch( 'core/notices' ).createNotice(
+					'warning',
```
gigitux: I think that it is better if we inject the notice in the InspectorControls. We do for other Woo Blocks. For instance, you can check this:

https://github.com/Kallyan01/woocommerce/blob/bc4f027a79cf2e48867419f82aa0eac18b13af73/plugins/woocommerce/client/blocks/assets/js/blocks/cart-checkout-shared/block-settings/express-payment-settings.tsx#L180-L190



You can wrap the `Notice` component inside the `InspectorControls` component.

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/use-background-image.ts
```diff
@@ -23,21 +24,41 @@ interface BackgroundProps {
...
 	const [ backgroundImageSrc, setBackgroundImageSrc ] = useState( '' );
+	const [ backgroundColorVisibility, setBackgroundColorVisibility ] =
+		useState( false );
+	const [
+		featuredProductParentDivDimensions,
+		setFeaturedProductParentDivDimensions,
+	] = useState< BgImageDimensions >( { height: 0, width: 0 } );
+	const [ isImageBgTransparent, setIsImageBgTransparent ] = useState( false );
+	const [ originalImgDimension, setOriginalImgDimension ] =
+		useState< BgImageDimensions >( { height: 0, width: 0 } );
```
gigitux, line 51: Can we compute these value at runtime? Why do we need to store them in a state?

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/use-background-image.ts
```diff
@@ -23,21 +24,41 @@ interface BackgroundProps {
...
+	blockAttributes,
 }: BackgroundProps ): BackgroundImage {
 	const [ backgroundImageId, setBackgroundImageId ] = useState( 0 );
 	const [ backgroundImageSrc, setBackgroundImageSrc ] = useState( '' );
+	const [ backgroundColorVisibility, setBackgroundColorVisibility ] =
+		useState( false );
+	const [
+		featuredProductParentDivDimensions,
+		setFeaturedProductParentDivDimensions,
+	] = useState< BgImageDimensions >( { height: 0, width: 0 } );
```
gigitux: Can we compute these value at runtime? Why do we need to store them in a state?

gigitux (changes requested):
Thanks for the working on this PR! It is a great work but I left few questions regarding not necessary state. I think that we can get rid of those. Also, I think that we can move the Notice in the `InspectorControls` as other blocks do.

Also, I noticed that when the color is picked from the color picker, the entire logic doesn't run:


https://github.com/user-attachments/assets/2bc13974-5951-4fd2-8197-33ffcc83ee48

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
Kallyan01: @gigitux tried that but i think we can't show notice component under style tab in block inspector ! Could you confirm ?
image

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
Kallyan01: There is a block resize functionality which won't work if we try to access height width with normal approach , we need to used event listeners for that then , so i found this browser-api [Ref](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) more useful and efficient and implemented it here .

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/use-background-image.ts
Kallyan01: image

The dimensions are being calculated by instantiating a new Image() object,
but doing this on runtime just to get the size may not be efficient can cause performance issue .

So i just stored it insdie a state . Let me know your thought on this if i'm wrong !

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/use-background-image.ts
Kallyan01: No ! as discussed above there might be problem where we try to resize the block element and also this state helps the useEffect to recheck the notice insertion logic whenever dimension changes detected .

https://github.com/Kallyan01/woocommerce/blob/bc4f027a79cf2e48867419f82aa0eac18b13af73/plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/use-background-image.ts#L127

Kallyan01:
@gigitux Thanks for the feedbacks !
I've fixed the colorpicker issue you encountered and replied to your comments . Let me know thoughts on those replies. Once the open discussions ends i'll push all the required changes and request for re-review .

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
gigitux: > There is a block resize functionality which won't work if we try to access height width with normal approach , we need to used event listeners for that then , so i found this browser-api [Ref](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) more useful and efficient and implemented it here .

Not sure that I'm following but when the resize happens, the attributes are updated and this will trigger this useEffect:

https://github.com/Kallyan01/woocommerce/blob/bc4f027a79cf2e48867419f82aa0eac18b13af73/plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/use-background-image.ts#L127-L166

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/use-background-image.ts
gigitux: Oh gotcha! It makes sense, but we could use `useRef` to avoid a not necessary re-render: does it make sense?

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/use-background-image.ts
```diff
@@ -23,21 +24,41 @@ interface BackgroundProps {
...
+	const [ isImageBgTransparent, setIsImageBgTransparent ] = useState( false );
```
gigitux: We could use `useRef` for this as well.

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
gigitux: Based on my idea, I think that this `useEffect` should have as dependency only `blockAttributes`. Potentially, we could unify both `useEffect` in one, ensuring that we run `img.onload` logic only when the `backgroundImageSrc` is updated. 

You could cache the `backgroundImageSrc` previous value using the `usePrevious` hook. Does it make sense?

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
gigitux: I don't have a lot of context, but it looks like that it is possible: https://developer.wordpress.org/news/2023/06/using-block-inspector-sidebar-groups/

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
```diff
@@ -114,17 +116,75 @@ export const withFeaturedItem =
...
+					width: entries[ 0 ].contentRect.width,
+				} );
+			} );
+
+			if ( isLoading === false ) {
+				const element = document
+					?.getElementsByTagName( 'iframe' )[ 0 ]
+					?.contentWindow?.document.querySelector(
+						'.featured-product-parent'
+					) as HTMLElement | null;
```
gigitux: Also, if possible, we should avoid using the query selector but, instead, we should use `ref` to control DOM elements: https://react.dev/learn/manipulating-the-dom-with-refs

gigitux:
> @gigitux Thanks for the feedbacks ! I've fixed the colorpicker issue you encountered and replied to your comments . Let me know thoughts on those replies. Once the open discussions ends i'll push all the required changes and request for re-review .

Thanks for your replies! I left my thoughts! Please let me know what you think about it! Thanks 🙇

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
Kallyan01: @gigitux didn't get this ! are you talking about img.onload logic in user-background-image hook ?

gigitux: > are you talking about img.onload logic in user-background-image hook ?

Yes!

Kallyan01:
@j111q @gigitux able to show the warning notice in the control-panel -> style tab .

could you confirm the warning message which need to be shown ! bit confused in these as the messages shared by @j111q earlier was not mentioning if the image is opaque or not .
```bash
You've set a background color in the Featured Product block. Note that if there is an opaque image, the background color cannot be seen.
You've set a background color in the Featured Category block. Note that if there is an opaque image, the background color cannot be seen.
```
but actually we know the image bg color compatibility status so could we say differently this and also as the notice placement has been changed so , we can exclude the `Featured Product` / `Featured Category` mentions on the notices message.

```
You've set background color to an opaque image, the background color cannot be seen.
```
image

gigitux:
> ```
> You've set background color to an opaque image, the background color cannot be seen.
> ```

I think that it makes sense 👍

Kallyan01:
Hello @gigitux sorry for the delay was busy with some other work . Please do check the changes i've made . Thanks !

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
```diff
@@ -114,16 +118,88 @@ export const withFeaturedItem =
...
+			// Checks if bg-image is not transparent and repeated all-over parent div or covers available parent div space.
+			if (
+				! isImageBgTransparent &&
+				( isRepeated || imageFit === 'cover' )
+			) {
+				isBgVisible = false;
+			}
+
+			if ( bgColorVisibility !== isBgVisible ) {
+				setAttributes( { bgColorVisibility: isBgVisible } );
```
gigitux: Is there a particular reason why we need to use `setAttributes`? Could it be a classic React state?

gigitux (changes requested):
Thanks for addressing the feedback! 🙇 Great work!  I left a comment, but I noticed that currently the “notice” continues to be visible even if the image background is removed:


https://github.com/user-attachments/assets/d7dbf6d0-d4da-4c4e-99ab-aab19ab51f88

I will take another review once this bug is fixed.

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
Kallyan01: @gigitux few reasons for which i haven't used useState here 

- Passing local state values between different files ( `/with-featured-items.tsx` to `/inspector-controls.tsx` ) would have added unnecessary complexity.
- You previously encouraged minimizing the use of useState where it's not strictly necessary.
- The current approach keeps things cleaner and more straightforward. Since we’re updating the attribute directly after detecting a change, skipping useState eliminates an extra step and avoids redundant state management.

Kallyan01: Added the fix (Attached the testing videos for both featured-product and featured-category block ): 

https://github.com/user-attachments/assets/b82f51cf-47c6-4fd1-9da5-bf167b6115aa

Kallyan01: https://github.com/user-attachments/assets/ae77c581-0fc0-41fb-b7bf-afa063c40d6a

Kallyan01:
@gigitux have you had the chnace to test the new changes ?

Konamiman:
I've removed myself as reviewer since this is a blocks related change, which I'm not familiar with, and @gigitux is already taking care of reviewing and suggesting changes.

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
gigitux: > @gigitux few reasons for which i haven't used useState here
> 
> * Passing local state values between different files ( `/with-featured-items.tsx` to `/inspector-controls.tsx` ) would have added unnecessary complexity.
> * You previously encouraged minimizing the use of useState where it's not strictly necessary.
> * The current approach keeps things cleaner and more straightforward. Since we’re updating the attribute directly after detecting a change, skipping useState eliminates an extra step and avoids redundant state management.

Under the hood, `setAttributes` updates the React state. The problem is that we're currently using attributes solely to represent state within the editor. However, attributes should also be structured to reflect the desired output on the frontend.

Kallyan01:
Hello @gigitux 
Have you checked the changes ? Any suggestion which needs to be addressed ?

gigitux:
> Hello @gigitux Have you checked the changes ? Any suggestion which needs to be addressed ?

Sorry if I wasn't clear, but I think that we should not use attributes https://github.com/woocommerce/woocommerce/pull/57490#discussion_r2106977703. Could you refactor it?

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
Kallyan01: @gigitux Sorry , It's still not clear to me why we can't use setAttribute here and whats the problem ?
- We need to share the state between two files `inspector-controls.tsx` and `with-featured-item.tsx` to make it work 
- We want to store the `bgColorVisibility` status for each block 
- We can't use global state cause this feature is specific to a single block

All requirements can be satisfied by the setAttribute , useState can't do them ! mainly storing the value in DB and sharing the state value between two files as they don't have parent child relation .

Kallyan01: @gigitux if you have time could you check this once , once the requirments are clear will start implement the fix if required .

gigitux: Sorry for the delay! `attributes` should be used to represent different configurations or UI states of the block. In 
this case, we're using them to store data that's strictly related to the editor, which might not be appropriate.


You can pass the computed value as props here:

image


It will be present in the `withInspectorControls` file:

image


I hope that it makes sense.

Kallyan01: @gigitux Thanks for the hint , I have added necessary changes please do review when you have some time ! 
I hope all the feedback points are addressed with this fix .
Let me know if you find anything else to improve .

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
```diff
@@ -114,16 +117,88 @@ export const withFeaturedItem =
...
+				setBgColorVisibility( isBgVisible );
+			}
+		}, [
+			parentContainerDimension,
+			originalImgDimension,
+			isRepeated,
+			imageFit,
+			backgroundImageSrc,
+			isImageBgTransparent,
+		] );
```
gigitux: I think that we can get rid of the state for the `bgColorVisibility` variable.

We can extract the visibility logic into a pure function: 

```ts
const calculateBackgroundVisibility = ({
    isImageBgTransparent,
    originalImgDimension,
    parentContainerDimension,
    isRepeated,
    imageFit
}) => {
    if (isImageBgTransparent) {
        return true;
    }

    if (
        originalImgDimension.height < parentContainerDimension.height ||
        originalImgDimension.width < parentContainerDimension.width
    ) {
        return true;
    }

    if (
        !isImageBgTransparent &&
        originalImgDimension.height >= parentContainerDimension.height &&
        originalImgDimension.width >= parentContainerDimension.width
    ) {
        return false;
    }

    if (!isImageBgTransparent && (isRepeated || imageFit === 'cover')) {
        return false;
    }

    return true; // default case
};
```

After that, we can use `useMemo` to cache the result of a calculation

```
const isBgVisible = useMemo(
    () => calculateBackgroundVisibility({
        isImageBgTransparent,
        originalImgDimension,
        parentContainerDimension,
        isRepeated,
        imageFit
    }),
    [
        isImageBgTransparent,
        originalImgDimension,
        parentContainerDimension,
        isRepeated,
        imageFit
    ]
);
```

gigitux: What do you think?

gigitux (changes requested):
Thanks for working on this! I think that we're near to merging this PR! I left a comment to code-wise improvements. Also, I noticed that we should disable the check when `repeat background` or `cover` image fit are enabled:


https://github.com/user-attachments/assets/33d5e583-3a7f-4d32-8fa7-b93823698bab

Does it make sense?

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
Kallyan01: Yes ! We could do this , it's a good point which we can improve . I'll do this asap .

Kallyan01: @gigitux About the video you have shared in this thread , instead of removing the logic, could we modify the logic to show a different message like "The image settings is set to repeat/cover , the background color cannot be seen."

gigitux: @Kallyan01 It makes sense to me! 

Let's create a dedicated message for both scenarios. I would suggest this:

### For background cover

```
You’ve set a background color behind an image set to cover, the background color cannot be seen.
```

### For background repeat

```
You’ve set a background color behind an image set to repeat, the background color cannot be seen.
```

Kallyan01:
@gigitux the changes had been implemented and requested re-review, looking forward for your feedback , Thanks !

Kallyan01:
@gigitux the changes have been implemented and requested re-review, looking forward for your feedback , Thanks !

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
```diff
@@ -114,16 +127,140 @@ export const withFeaturedItem =
...
+					reason: null,
+				};
+			}
+
+			// default case
+			return {
+				value: true,
+				reason: null,
+			};
+		};
```
gigitux: Let's move this function to a `utils` file. Also, given that we update the return of this function, I would suggest renaming it. What do you think with: `getBackgroundColorVisibilityStatus`?

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
```diff
@@ -114,16 +127,140 @@ export const withFeaturedItem =
...
+			}
+
+			// Checks if original-bg-image is smaller than the parent container's available space.
+			if (
+				originalImgDimension.height < parentContainerDimension.height ||
+				originalImgDimension.width < parentContainerDimension.width
+			) {
+				return {
+					value: true,
+					reason: null,
```
gigitux: Let's use `message` instead of `reason`.

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
```diff
@@ -114,16 +127,140 @@ export const withFeaturedItem =
...
+				};
+			}
+
+			// default case
+			return {
+				value: true,
+				reason: null,
```
gigitux: Same here

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
```diff
@@ -114,16 +127,140 @@ export const withFeaturedItem =
...
+						value: false,
+						message: __(
+							'You’ve set a background color behind an image set to repeat, the background color cannot be seen.',
+							'woocommerce'
+						),
+					};
+				}
+
+				return {
+					value: false,
```
gigitux: Let's make this boolean more explicit, renaming it with `isBackgroundVisible`.

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
```diff
@@ -114,16 +127,140 @@ export const withFeaturedItem =
...
+			}
+
+			// default case
+			return {
+				value: true,
+				reason: null,
+			};
+		};
+
+		const isBgVisible = useMemo(
```
gigitux: Let's rename the variable name. What do you think about `backgroundColorVisibilityStatus`?

gigitux (changes requested):
Thanks for your effort! I left some suggestions regarding the code readability. I hope that they make sense. I will test the PR after you will address them! 🙇 


Also, there are some lint errors, can you fix them? Thanks!

Kallyan01:
Apology for mistakenly requesting for re-review !

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
Kallyan01: Yes ! Works too .

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
Kallyan01: Works !

Kallyan01:
@gigitux Addressed the suggested changes and also fixed some eslint errors . Tested the functionality , all looks good to me , let me know if you found any bug while testing . Thanks !

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/utils.ts
```diff
@@ -91,3 +97,89 @@ export function dimRatioToClass( ratio: number ) {
...
+ * based on the image's transparency, repetition, fit, and container size.
+ *
+ * @param isImageBgTransparent - Whether the image has transparent pixels.
+ * @param originalImgDimension - The original dimensions of the background image.
+ * @param parentContainerDimension - The dimensions of the container holding the image.
+ * @param isRepeated - Whether the image is set to repeat.
+ * @param imageFit - How the image is fitted (e.g., 'cover' or 'none').
+ * @return An object indicating whether the background color is visible
+ *          and the notice message.
+ */
```
gigitux: There are some lint errors regarding this JSDOC. I think that we can remove it given that there are type definitions.

gigitux (changes requested):
Thanks for the update! Linting is still failing, could you fix the last errors? Thanks!

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/inspector-controls.tsx
```diff
@@ -63,6 +64,10 @@ interface WithInspectorControlsRequiredProps< T > {
 	attributes: InspectorControlsRequiredAttributes &
 		EditorBlock< T >[ 'attributes' ];
 	setAttributes: InspectorControlsProps[ 'setAttributes' ];
+	isBgVisible: {
+		isBackgroundVisible: boolean;
+		message?: string | null;
```
gigitux: Let's make it required

```suggestion
		message: string | null;
```

Otherwise, `message` could be `string | null | undefined` and create not necessary complexity

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
```diff
@@ -329,6 +389,7 @@ export const withFeaturedItem =
 				<Component
 					{ ...props }
 					backgroundImageSize={ backgroundImageSize }
+					isBgVisible={ backgroundColorVisibilityStatus }
```
gigitux: Let's rename `isBgVisible` to `backgroundColorVisibilityStatus` for consistency.

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/with-featured-item.tsx
```diff
@@ -338,6 +399,7 @@ export const withFeaturedItem =
...
```
gigitux: Let's rename `isBgVisible` to `backgroundColorVisibilityStatus` for consistency.

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/inspector-controls.tsx
```diff
@@ -326,6 +364,10 @@ export const withInspectorControls =
 					setAttributes={ setAttributes }
 					setGradient={ setGradient }
 					showDesc={ showDesc }
+					bgColorVisibility={ isBgVisible }
```
gigitux: Let's rename `bgColorVisibility` to `backgroundColorVisibilityStatus` for consistency.

gigitux (changes requested):
Thanks for working on this feature. Amazing work! 🚀  I tested the PR and everything works as expected:


#### No Media Settings

https://github.com/user-attachments/assets/4b676e73-2742-453c-8538-e1597d672e8b

#### Cover set to true

https://github.com/user-attachments/assets/4e132f1c-4e8f-47f6-b408-dbed21c91d7a

#### Repeated background set to true


https://github.com/user-attachments/assets/d345bd73-ce5d-42c7-9ad2-573da031d1ea

#### Image bigger than container


https://github.com/user-attachments/assets/069e4163-5079-4466-943a-fa8af57665c6

I did a final review and I noticed other code-wise improvements. Could you take a look? Thanks! 🙇

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/inspector-controls.tsx
```diff
@@ -104,168 +111,196 @@ export const InspectorControls = ( {
...
-							) }
-						</PanelBody>
+						</PanelColorGradientSettings>
+					</>
+				) }
+			</GutenbergInspectorControls>
+			<GutenbergInspectorControls group="color">
+				{ backgroundColorVisibilityStatus &&
+					backgroundColorVisibilityStatus.isBackgroundVisible === false &&
+					backgroundColor && (
```
gigitux, line 290: I would add another check to ensure that the `message` is defined to avoid rendering the notice with an empty message.

```
backgroundColorVisibilityStatus.isBackgroundVisible === false && backgroundColorVisibilityStatus.message && ...
```


What do you think?

gigitux (approved):
Thanks for your effort! I'm going to pre-approve your PR, but before merging it, can you address the last feedback? Thanks! 🙇

gigitux:
Also, I noticed that there are some lint errors. Can you fix them?

plugins/woocommerce/client/blocks/assets/js/blocks/featured-items/inspector-controls.tsx
Kallyan01: @gigitux As the message isn’t coming from any third-party source (APIs, user input etc) — we have only three predefined messages. Since it's quite predictable that the message would be one of the defined ones for false cases, I believe keeping the current logic as-is shouldn’t cause any issues. Let me know your thoughts.

gigitux: Yes, but I’d still stay defensive in case any mistakes are introduced during future refactors and slip through during the code review phase. Does it make sense?

Kallyan01: Yes ! ok i'm adding the check .

Kallyan01:
@gigitux The changes have been added , Let me know if anything else needs improvement . 
Thank you so much for your continuous feedback and patience throughout — it’s been a great learning experience for me!

gigitux:
Thank you again for your contribution! Let's merge it! 🚀

PR #57491, opened 2025-04-23
Add Cart/Checkout Redirect Tests

### LINKED ISSUE #57453: Add Cart/Checkout Redirect Tests
There is no test coverage for redirects between cart and checkout pages. We should add appropriate tests to ensure the customer gets redirected to the correct page depending on the state of the cart.

Noticed in [https://github.com/woocommerce/woocommerce/pull/57172#issuecomment-2797217338](https://github.com/woocommerce/woocommerce/pull/57172#issuecomment-2797217338)

BurakParsAydin:

Closes woocommerce/woocommerce#57453  .

1. Run the following command in the WC root `pnpm --filter=@woocommerce/plugin-woocommerce test:unit:env -- --filter=WC_Shortcodes_Test`
2. You should see `OK (15 tests, 27 assertions)`

### Changelog entry

#### Significance

- [ ] Patch
- [X] Minor
- [ ] Major

#### Type

- [X] Add - Adds functionality

#### Message

</details>

#### Comment

Add test to ensure no redirect occurs when both \[woocommerce_cart\] and \[woocommerce_checkout\] shortcodes are used

</details>


kalessil:
@tjcafferkey: routing this review to Rubik as it blocks related.

plugins/woocommerce/tests/php/includes/class-wc-shortcodes-test.php
```diff
@@ -476,4 +476,30 @@ public function test_product_page_shortcode_protected_product() {
...
+	public function test_combined_cart_and_checkout_shortcodes_do_not_redirect() {
+		// Simulate being on the cart page (as if [woocommerce_cart] is present).
+		add_filter( 'woocommerce_is_cart', '__return_true' );
+
+		// Track if a redirect is triggered.
+		$redirect_url = null;
+		add_filter( 'wp_redirect', function ( $location ) use ( &$redirect_url ) {
+			$redirect_url = $location;
+			return $location;
+		}, 10 );
```
tjcafferkey: Lets refactor this slightly so we can clean up this filter after the test has run:

* Move `$redirect_url` to a variable on the class.
* Move the anonymous function into a helper method on this class also.

I imagine it looking something like this:

```php
/**
* Helper method to track redirects.
*
* @param string $location The redirect location.
* @return string The unchanged redirect location.
*/
public function track_redirect( $location ) {
	$this->redirect_url = $location;
	return $location;
}
```

Then we can reference it when adding the filter and also removing it:

```php
add_filter( 'wp_redirect', array( $this, 'track_redirect' ), 10, 1 );
remove_filter( 'wp_redirect', array( $this, 'track_redirect' ), 10 );
```

plugins/woocommerce/tests/php/includes/class-wc-shortcodes-test.php
```diff
@@ -476,4 +476,30 @@ public function test_product_page_shortcode_protected_product() {
 
 		$this->assertStringContainsString( 'This content is password protected', $product_page );
 	}
+
+	/**
+	 * Ensure that when both [woocommerce_cart] and [woocommerce_checkout] shortcodes
+	 * are rendered together, no redirect to the cart page occurs.
+	 */
+	public function test_combined_cart_and_checkout_shortcodes_do_not_redirect() {
```
tjcafferkey, line 502: It looks like there was some linting issues were reported in the function that were refactoring anyway, so I'm not sure whether you have it setup correctly in your IDE but might be worth doing it to avoid further linting failures.

tjcafferkey (changes requested):
This looks almost done @BurakParsAydin and I really appreciate you submitting this test. I have just left a couple comments to refactor but overall its looking great!

BurakParsAydin:
@tjcafferkey Thanks so much for the detailed guidance and examples! It also seems that the linting issue disappeared after adjusting the code as you suggested. 
Please let me know if there’s anything else I should improve!

plugins/woocommerce/tests/php/includes/class-wc-shortcodes-test.php
```diff
@@ -476,4 +476,42 @@ public function test_product_page_shortcode_protected_product() {
...
 	}
+
+	/**
+	 * Helper method to track redirects.
+	 *
+	 * @param string $location The redirect location.
+	 * @return string The unchanged redirect location.
+	 */
+	public function track_redirect( $location ) {
+		$this->redirect_url = $location;
```
tjcafferkey, line 494: Last thing: We just need to add this variable to the top of the main class alongside `$user_administrator` and `$user_contributor` like so:

```php
	/**
	 * The redirect URL used in tests.
	 *
	 * @var string
	 */
	protected static $redirect_url;
```

tjcafferkey (changes requested):
Thanks for addressing that feedback @BurakParsAydin! I think you just missed one thing, then we're ready to merge!

BurakParsAydin:
@tjcafferkey Thank you! I've added it

tjcafferkey:
Something I've just tested is that when we revert this PR https://github.com/woocommerce/woocommerce/pull/57172 your test still passes, it's something I really should have checked earlier in the review process.

I the the issue is we are testing at a lower-level and by doing so bypassing some of the functionality that we're really supposed to be testing e.g. the redirect.

Let me know your own assessment of this @BurakParsAydin because if we come to the same conclusion we can look to move this to an E2E test instead which I think is where we can catch this bug.

BurakParsAydin:
@tjcafferkey Thanks for the clarification - I’ve now reverted #57172 locally and confirmed that my current test still passes, so it looks like it’s not effectively catching the original issue as you outlined. 

I’ll look into improving the test further, possibly by triggering `template_redirect` or simulating a more complete request context. That said, I’m not yet certain if it’s fully achievable at the unit test level due to how the redirect logic is tied to WordPress's page lifecycle.

I’ll give it a try and report back with what I find, and if it turns out not to be feasible, we can reassess the best path forward.

--- 

Apart note,  I’m on holiday for the next few days and will be back on Monday. If the timing works on your end, I’ll continue looking into the test improvement early next week.

tjcafferkey:
>I’m not yet certain if it’s fully achievable at the unit test level due to how the redirect logic is tied to WordPress's page lifecycle.

> I’ll give it a try and report back with what I find, and if it turns out not to be feasible, we can reassess the best path forward.

I'm not sure either so I'm totally up for learning together on this one. It might also be good to refer to the [WordPress repo](https://github.com/WordPress/wordpress-develop) for examples on this.

>Apart note, I’m on holiday for the next few days and will be back on Monday. If the timing works on your end, I’ll continue looking into the test improvement early next week.

That is fine with me. Appreciate the heads up. I hope you enjoy your holiday and we can pick this up when you're back.

PS: I am also available in the Woo community Slack channel should you need to reach out for more direct conversation on this too when you come to it :)

BurakParsAydin:
@tjcafferkey Quick update - I spent quite a bit of time trying to simulate this redirect logic in a unit test, but unfortunately couldn’t get it to behave reliably. I explored several approaches, but none managed to consistently trigger or block the redirect within PHPUnit.

As a side note, my knowledge of WooCommerce internals is still a bit limited, so there may be solutions or test patterns I haven’t come across yet. At this point, I’m not quite sure how best to move forward - is there anything else you'd suggest trying?

tjcafferkey:
@BurakParsAydin this might be a candidate for an E2E test if we are unable to cover it via a unit test. I'd explore that option, you should be able to find existing tests around this area [here](https://github.com/woocommerce/woocommerce/tree/trunk/plugins/woocommerce/tests/e2e-pw)

BurakParsAydin:
@tjcafferkey Thanks for the suggestion. I’ll explore the existing E2E tests. 
Please let me know if, after your review, you think it’s worth pursuing. I’ll be ready to pick it up if needed.

nerrad:
Stale and not relevant.

PR #57495, opened 2025-04-24
Hide 'View Cart' button when the cart page is not defined

### LINKED ISSUE #56563: Cart link shown when adding to the cart on store with no Cart page
### Prerequisites

### Describe the bug

A common use case I've seen is that people disable the cart page, and show the mini cart drawer to get to the checkout step.

The Mini-cart block handles this case and hides the cart button

Image

There are 2 cases where this link is not hidden:

1. The Shop page shows a link for the cart under the button. 

Image
 
2. Adding to cart from a product page shows a link to the cart 

Image

### Expected behavior

Don't show a cart link if there is no cart page defined

### Actual behavior

The link is shown, and the user is redirect to the homepage.

### Steps to reproduce

Disable a cart page for a store. 
Add products to the cart from the archive and product detail pages
Notice the link appearing
Image

### WordPress Environment

n/a

### Isolating the problem

gigitux:

This PR hides the `View Cart` button when the cart page is not defined. I created `has_cart_page` util to avoid replicating the same check across the codebase. I added a unit test to ensure that the behavior is the expected one.

> [!IMPORTANT]
> I used the "View Cart" string across the PHP codebase to ensure that I updated all the cases, but in case you think that I missed some use cases, please let me know.

Closes https://github.com/woocommerce/woocommerce/issues/56563 (WOOPLUG-3535) .

## Block Theme

1. Visit `/wp-admin/admin.php?page=wc-settings&tab=advanced`.
2. Remove the Cart Page:
image
3. Install and enable a block theme (like TT5).
5. Visit a Single Product page
6. Add the product to the cart.
7. Ensure that the notice on top doesn't have the `View Cart` link.

| Before | After |
|--------|--------|
|image| image|

8. Edit the product and enable the `Limit purchases to 1 item per order` option.
9. Re-add the product.
10. Ensure that the notice on top doesn't have the `View Cart` link.

| Before | After |
|--------|--------|
| image| image| 

11. Visit the `/shop` page.
12. Add a product to the cart.
13. Ensure that the `view cart` link isn't rendered.

| Before | After |
|--------|--------|
|image | image | 

## Classic Theme

1. Visit `/wp-admin/admin.php?page=wc-settings&tab=advanced`.
2. Remove the Cart Page:
image
3. Install and enable Storefront (or another classic theme)
5. Visit a Single Product page
6. Add the product to the cart.
7. Ensure that the notice on top doesn't have the `View Cart` link.

| Before | After |
|--------|--------|
| image |image | 

8. Edit the product and enable the `Limit purchases to 1 item per order` option.
9. Re-add the product.
10. Ensure that the notice on top doesn't have the `View Cart` link.

| Before | After |
|--------|--------|
| image | image| 

> [!WARNING]
> The /shop page in the Classic theme wasn’t updated because, with the current UI/UX, removing the “View Cart” link leaves users without clear feedback that a product was added. Happy to implement it if you think it’s worth doing.

https://github.com/user-attachments/assets/bc8f2a95-7d56-40be-847f-9d3845978510

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Hide 'View Cart' button when the cart page is not defined.
</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/57495-wooplug-3535-cart-link-shown-when-adding-to-the-cart-on-store-with-no
+++ b/plugins/woocommerce/changelog/57495-wooplug-3535-cart-link-shown-when-adding-to-the-cart-on-store-with-no
@@ -1,0 +1,4 @@
+Significance: minor
+Type: fix
+
+Hide 'View Cart' button when the cart page is not defined.
--- a/plugins/woocommerce/includes/class-wc-cart.php
+++ b/plugins/woocommerce/includes/class-wc-cart.php
@@ -9,6 +9,7 @@
  * @version 2.1.0
  */
 
+use Automattic\WooCommerce\Blocks\Utils\CartCheckoutUtils;
 use Automattic\WooCommerce\Enums\ProductStatus;
 use Automattic\WooCommerce\Enums\ProductType;
 use Automattic\WooCommerce\Utilities\DiscountsUtil;
@@ -1193,7 +1194,13 @@ class WC_Cart extends WC_Legacy_Cart {
 					$message         = apply_filters( 'woocommerce_cart_product_cannot_add_another_message', $message, $product_data );
 					$wp_button_class = wc_wp_theme_get_element_class_name( 'button' ) ? ' ' . wc_wp_theme_get_element_class_name( 'button' ) : '';
 
-					throw new Exception( sprintf( '%s <a href="%s" class="button wc-forward%s">%s</a>', $message, esc_url( wc_get_cart_url() ), esc_attr( $wp_button_class ), __( 'View cart', 'woocommerce' ) ) );
+					if ( ! CartCheckoutUtils::has_cart_page() ) {
+						$message = sprintf( '%s', esc_html( $message ) );
+					} else {
+						$message = sprintf( '%s <a href="%s" class="button wc-forward%s">%s</a>', $message, esc_url( wc_get_cart_url() ), esc_attr( $wp_button_class ), __( 'View cart', 'woocommerce' ) );
+					}
+
+					throw new Exception( $message );
 				}
 			}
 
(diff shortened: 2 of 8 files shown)


kmanijak (approved):
This is a nice improvement! 

Tested on:
- TT5
- Storefront

✅ On Single Product (with and without product limit) and on product catalog.

The only issues I found is the one you actually mentioned:

> The /shop page in the Classic theme wasn’t updated because, with the current UI/UX, removing the “View Cart” link leaves users without clear feedback that a product was added. Happy to implement it if you think it’s worth doing.

I agree we should not remove it and having a feedback is more important. I'm wondering though if we could display something like `Added!` instead of `View cart`. What do you think? But I don't think it's a blocker because anyway this PR is an improvement.

gigitux:
> I agree we should not remove it and having a feedback is more important. I'm wondering though if we could display something like `Added!` instead of `View cart`. What do you think? But I don't think it's a blocker because anyway this PR is an improvement.

My main concern is that this UI/UX change could impact many stores and could have unexpected side effects.

kmanijak:
> My main concern is that this UI/UX change could impact many stores and could have unexpected side effects.

But current behavior is kind of a bug. "View cart" appears but it's incorrect. I just mean changing copy in case there's no cart, otherwise stick with "View cart". What unexpected side effects you're worried about?

gigitux:
> > My main concern is that this UI/UX change could impact many stores and could have unexpected side effects.
> 
> But current behavior is kind of a bug. "View cart" appears but it's incorrect. I just mean changing copy in case there's no cart, otherwise stick with "View cart". What unexpected side effects you're worried about?

- The string `Added` is not included in our translation set, so some stores may display it untranslated until it's added.
- Some third-party plugins might use custom JavaScript to mimic similar behavior. We should investigate how common this is and how to avoid breaking their implementations.
- The current script relies on the `woocommerce_add_to_cart_redirect` hook, which is [deprecated](https://github.com/woocommerce/woocommerce/blob/128f2a23bcb53db5d20cdf49e270cc96feb23f49/plugins/woocommerce/includes/class-wc-deprecated-filter-hooks.php#L27-L28):

  https://github.com/woocommerce/woocommerce/blob/128f2a23bcb53db5d20cdf49e270cc96feb23f49/plugins/woocommerce/includes/class-wc-frontend-scripts.php#L595-L602

Some stores may not define a cart page, but instead redirect users elsewhere. In these cases, the `View Cart` button should still be displayed. 

I'm aware that all these points can be fixed, but given that we want to invest in block themes, I'm wondering if it makes sense to spend time on these: what do you think?

kmanijak:
@gigitux appreciate the response! 😄 I get it know and agree with you!

> I'm aware that all these points can be fixed, but given that we want to invest in block themes, I'm wondering if it makes sense to spend time on these: what do you think?

No, not worth it, let's ship it as-is!

PR #57500, opened 2025-04-24
[Feat] Allow exporting specific products and its variants by IDs from the product list

naman03malhotra:
**Description:**

This PR introduces the ability for users to export a specific selection of products directly from the main Products list table. Previously, product export was limited to all products or filtering by product type and category. This change enhances user workflow by allowing targeted exports. The dynamic text for the export button is now translatable, and the page load for selected exports is protected by a nonce.

**Problem:**
Users lacked an efficient way to export only a handpicked set of products. If they needed a CSV for just a few specific items, they would have to export a much larger set (e.g., by category) and then manually filter the resulting CSV, or use other more cumbersome methods. Additionally, loading the export page with pre-selected product IDs via URL parameters lacked CSRF protection.

**Solution:**
This PR implements the following to enable selected product export:

1.  **Product List UI Enhancements (`woocommerce_admin.js`, `class-wc-admin-assets.php`):**
    *   When checkboxes are selected on the "Products > All Products" screen, the main "Export" button dynamically updates.
    *   The button text changes to reflect the number of selected items (e.g., "Export 5 selected"). This text is now translatable due to the addition of the `export_selected_products` localized string.
    *   The button's link is appended with a `product_ids` query parameter (comma-separated list of selected product IDs) and an `_wpnonce` query parameter (generated from the `export_selected_products_nonce` localized nonce) for security.

2.  **Exporter Page Enhancements (`html-admin-page-product-export.php`, `admin.scss`):**
    *   **Nonce Verification:** Before processing `$_GET['product_ids']`, `check_admin_referer('export-selected-products')` is called to validate the nonce passed in the URL, enhancing security.
    *   When accessed with valid `product_ids` and nonce in the URL:
        *   An informational notice is displayed confirming which products are targeted for export (e.g., "You are exporting the 5 products you selected..."). This notice includes a `// translators:` comment for placeholder clarification.
        *   A "Clear selection" link is provided in the notice to revert to exporting all products.
        *   The main description text is updated to reflect that selected products will be exported.
        *   A hidden input field `product_ids` is added to the form, carrying the selected IDs for the AJAX export process.
        *   The "Product types" and "Product category" filter fields are hidden (styles moved from inline to `admin.scss`), as they are not applicable when specific IDs are being exported.
    *   A minor CSS adjustment in `admin.scss` increases the `max-width` of the exporter wrapper slightly.

3.  **Backend Export Logic Updates:**
    *   **AJAX Handling (`wc-product-export.js`, `class-wc-admin-exporters.php`):**
        *   The `export_product_ids` (from the hidden field) are now included in the AJAX request when generating the CSV. This AJAX request remains protected by its own existing nonce (`wc_product_export_params.export_nonce`).
        *   The AJAX handler in `WC_Admin_Exporters` retrieves these IDs, sanitizes them (ensuring they are absolute integers), and passes them to the exporter instance.
    *   **CSV Exporter Core (`class-wc-product-csv-exporter.php`):**
        *   A new method `set_product_ids_to_export()` is added to store the sanitized list of product IDs.
        *   The `prepare_data_to_export()` method is updated to prioritize these specific IDs:
            * It uses the IDs set via `set_product_ids_to_export()`.
            * If specific IDs are present, the product query (`wc_get_products`) is modified to fetch only those products using the `include` argument, bypassing type and category filters.
            * The logic for fetching variations of variable products is adjusted to ensure variations are included if their parent product is selected either by ID or by a category filter (when not exporting specific IDs).

**Key Changes by File:**

*   `plugins/woocommerce/changelog/exp-product-export-by-id`: Added changelog entry.
*   `plugins/woocommerce/client/legacy/js/admin/woocommerce_admin.js`: Implemented product selection detection on the All Products page. Dynamically updated the "Export" button's text (using localized string) and link (appending `product_ids` and `_wpnonce`). Cleaned up an unused variable and addressed line length.
*   `plugins/woocommerce/client/legacy/js/admin/wc-product-export.js`: Modified AJAX call to include `export_product_ids` if present in the form.
*   `plugins/woocommerce/includes/admin/class-wc-admin-assets.php`:
    *   Added the `export_selected_products` string to the localized parameters, making the dynamic export button text translatable.
    *   Added the `export_selected_products_nonce` to the localized parameters for securing the GET request to the export page.
    *   Corrected typographic quotes/apostrophes in existing localized strings.
*   `plugins/woocommerce/includes/admin/views/html-admin-page-product-export.php`:
    *   Added `check_admin_referer('export-selected-products')` to validate the nonce from the URL.
    *   Checks for `product_ids` in GET parameters.
    *   Displays an informational notice about selected products and a "clear selection" link (with `// translators:` comment).
    *   Adds a hidden input field for `product_ids`.
    *   Adjusts descriptive text based on whether specific products are being exported.
    *   Removed inline styles for hiding filters (moved to `admin.scss`).
*   `plugins/woocommerce/includes/admin/class-wc-admin-exporters.php`: In `do_ajax_export()`, added logic to retrieve, sanitize, and set `export_product_ids` on the exporter instance.
*   `plugins/woocommerce/includes/export/class-wc-product-csv-exporter.php`:
    *   Added `protected $product_ids_to_export` property.
    *   Added public method `set_product_ids_to_export(array $product_ids)` (now with `@since 9.9.0`).
*   `plugins/woocommerce/client/legacy/css/admin.scss`: Minor increase to `max-width` of `.woocommerce-exporter-wrapper`. Added styles to hide type/category filters on the exporter page when specific IDs are being exported (moved from inline styles).

**How to Test:**

1.  Navigate to "Products > All Products".
2.  Select one or more products using the checkboxes in the list table.
3.  **Observe:** The "Export" button at the top of the page should change its text to "Export X selected" (where X is the number of products) and its URL should now include `&product_ids=...` and `&_wpnonce=...`. Verify this text is translatable if you switch WordPress languages.
4.  Click the updated "Export" button.
    *   **Verify:** The page loads correctly due to the valid nonce.
5.  On the "Export Products" page:
    *   **Verify:** An informational notice appears stating "You are exporting the X product(s) you selected... Clear selection...".
    *   **Verify:** The "Which product types should be exported?" and "Which product category should be exported?" filter rows are hidden.
    *   **Verify:** The descriptive text indicates that selected products will be exported.
6.  Click "Generate CSV".
7.  **Verify:** The downloaded CSV file contains only the products (and their variations, if applicable) that you selected in step 2.
8. Experiment with exporting different types of products.
9.  **Test Clearing Selection:**
    *   On the "Export Products" page (when specific IDs are pre-selected), click the "Clear selection" link in the notice.
    *   **Verify:** The page reloads without the `product_ids` and `_wpnonce` in the URL, the notice disappears, and the product type/category filters become visible again.
10.  **Test Standard Export:**
    *   Navigate directly to the export page (or clear selection) and export all products or by type/category as usual. Verify this functionality remains unchanged.
11. **Test "Select All" Checkboxes:** Ensure "select all" functionality on the product list correctly updates the export button and exports all visible selected products.
12. **Test Nonce Failure (Manual URL Tampering):**
    *   Copy the generated URL from step 3 (with `product_ids` and `_wpnonce`).
    *   Try to load the URL after slightly changing the `_wpnonce` value or removing it.
    *   **Verify:** WordPress should show an error page ("The link you followed has expired.") indicating a failed nonce check.

**Security Considerations:**

*   **GET Request CSRF Protection:** The link to the export page with pre-selected IDs is now protected by a WordPress nonce (`_wpnonce` parameter validated by `check_admin_referer('export-selected-products')`).
*   **POST/AJAX Request CSRF Protection:** The actual export process (AJAX call) remains protected by its existing nonce (`wc_product_export_params.export_nonce`).
*   **Data Sanitization:** Product IDs passed via URL (`$_GET`) and POST (`$_POST`, hidden field) are rigorously sanitized using `sanitize_text_field`, `explode`, `array_map('absint', ...)`, and `array_filter(...)` at multiple stages.
*   **Output Encoding:** Output of IDs into HTML attributes is escaped using `esc_attr()`.
*   **Capability Checks:** The feature relies on standard WordPress capability checks for accessing product listing and export functionalities.


plugins/woocommerce/client/legacy/css/admin.scss
```diff
@@ -8256,7 +8256,7 @@ table.bar_chart {
 .woocommerce-exporter-wrapper,
 .woocommerce-importer-wrapper {
 	text-align: center;
-	max-width: 700px;
+	max-width: 710px;
```
naman03malhotra, line 8267: The PR is not finished yet, it needs a self review and some code refactor.

plugins/woocommerce/includes/export/class-wc-product-csv-exporter.php
```diff
@@ -183,25 +199,57 @@ public function prepare_data_to_export() {
...
 		$variable_products = array();
 
 		foreach ( $products->products as $product ) {
-			// Check if the category is set, this means we need to fetch variations separately as they are not tied to a category.
-			if ( ! empty( $args['category'] ) && $product->is_type( ProductType::VARIABLE ) ) {
-				$variable_products[] = $product->get_id();
+			// Check if the product is variable and if either the include or category filter is active.
+			// This is to ensure that product variations are only included if they are being selectively exported or if they are part of a category.
+			if ( $product->is_type( ProductType::VARIABLE ) && ( ! empty( $args['include'] ) || ! empty( $args['category'] ) ) ) {
+				if ( ! in_array( $product->get_id(), $variable_products, true ) ) {
```
jimjasson: That's smart! 

Worth noting that the `$products->products` array is generated by `wc_get_products`, which by default excludes duplicate products. However, the return value of `get_products` is filterable (see: `woocommerce_product_object_query`), so it makes sense to double-check here in case someone is filtering the returned products and is adding duplicate entries. 

That said, is there any reason why a nested `if` is needed here instead of a single `if`?

plugins/woocommerce/includes/export/class-wc-product-csv-exporter.php
```diff
@@ -183,25 +199,57 @@ public function prepare_data_to_export() {
...
+		$ids_for_include = array();
+
+		/**
+		 * Filter to allow exporting only specific product IDs.
+		 * Takes precedence over IDs set via set_product_ids_to_export().
+		 *
+		 * @since 9.9.0
+		 * @param array $product_ids Array of product IDs to export. Default empty array.
+		 */
+		$hook_provided_ids = apply_filters( 'woocommerce_product_export_specific_product_ids', array() );
```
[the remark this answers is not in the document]
jimjasson: This is valid feedback.

jimjasson (commented):
Hay @naman03malhotra, 

Thank you for implementing this feature! 

You have done an exceptional job documenting the proposed changes -- this made the PR review way faster and the code changes more easy to follow! 

I tested exporting:
- Simple Products, 
- Variable Products, 
- Simple/Variable Subscription Products, 
- Product Bundles, 
- Composite Products and;
- Gift Cards.

All products were exported successfully, similarly to exporting them via the standard Product Exporter. The only difference is that for Variable Products that are selected via the UI, variations are exported automatically, which makes sense as there isn't a way to select individual variations. 

I also tested the filter you provided and confirmed that it takes precedence over the products selected via the UI. 

I tried passing invalid values to the filter, like: 

```
function my_filter_product_export_ids( $product_ids ) {
	// Replace 15, 28, and 101 with the actual IDs you need.
	return array( -1, 'test', 0 );
}

add_filter( 'woocommerce_product_export_specific_product_ids', 'my_filter_product_export_ids' );
```

and confirmed that nothing got exported. Nice! 

The only limitation I found with this feature is that it can only export products from 1 page -- it is not possible to select products from different pages of the Products list table. The maximum amount of products per page is `999`, so this looks like the limit of this feature. I am raising this in case it is helpful for the feature's documentation. 

I left one comment around a specific part of the code and confirmed that the co-pilot's findings were valid. 

Let's have a look at those, before fully approving.

plugins/woocommerce/client/legacy/css/admin.scss
PanosSynetos: Why are the 10px necessary?

plugins/woocommerce/client/legacy/js/admin/wc-product-export.js
```diff
@@ -46,21 +46,24 @@
 			export_meta      = $( '#woocommerce-exporter-meta:checked' ).length ? 1: 0,
 			export_types     = $( '.woocommerce-exporter-types' ).val(),
 			export_category  = $( '.woocommerce-exporter-category' ).val();
+			// Check for specific product IDs from the hidden input.
+			var export_product_ids = $this.$form.find('input[name="product_ids"]').val() || '';
```
PanosSynetos: Let's follow the same approach as above, in one statement.

```
			export_category    = $( '.woocommerce-exporter-category' ).val(),
			export_product_ids = $this.$form.find('input[name="product_ids"]').val() || '';
```

plugins/woocommerce/client/legacy/js/admin/woocommerce_admin.js
```diff
@@ -791,4 +791,49 @@
 		wc_order_lock.init();
 	} );
 
+	// Function to handle selected product export
+	$( function () {
+		const $exportButton = $( 'a.page-title-action[href*="page=product_exporter"]');
+		// added defensive check to ensure the export button is present.
```
PanosSynetos: [nitpick] Just a `Bail out early.` would be enough as a comment

plugins/woocommerce/client/legacy/js/admin/woocommerce_admin.js
```diff
@@ -791,4 +791,49 @@
...
+		if ( !$exportButton.length ) {
+			return;
+		}
+
+		const originalExportHref = $exportButton.attr( 'href' );
+		const originalExportText = $exportButton.text();
+
+		// Use event delegation on the form containing the list table
```
PanosSynetos: [nitpick] Add `.` at the end of all comments - I think that PHPCS will complain if they are missing.

plugins/woocommerce/client/legacy/js/admin/woocommerce_admin.js
```diff
@@ -791,4 +791,49 @@
...
+					)
+						.map( function () {
+							return $( this ).val();
+						} )
+						.get(); // .get() converts the jQuery object to a standard array
+
+					// Update Export button
+					if ( selectedProductIds.length > 0 ) {
+						let newHref = originalExportHref + '&product_ids=' + selectedProductIds.join(',');
+						newHref += '&_wpnonce=' + woocommerce_admin.nonces.export_selected_products_nonce;
```
PanosSynetos: Instead of concatenations with `+` we could use `searchParams.set` , which is used elsewhere in the repository.
```
const url = new URL('https://www.google.com/');
url.searchParams.set('q', 'hello, world!');
```

plugins/woocommerce/client/legacy/js/admin/woocommerce_admin.js
```diff
@@ -791,4 +791,49 @@
...
+						// Construct the text with the count
+						const count = selectedProductIds.length;
+						// Use the localized string if available, otherwise fallback to a hardcoded string
+						const buttonText = woocommerce_admin.strings.export_selected_products
+							? woocommerce_admin.strings.export_selected_products.replace( '%d', count )
+							: 'Export ' + count + ' selected';
```
PanosSynetos: Is there a reason why the localized string is not available? If it's not translated, then the English version will be used anyway.

plugins/woocommerce/includes/admin/class-wc-admin-assets.php
```diff
@@ -234,11 +234,14 @@ public function admin_scripts() {
...
-						'export_products' => __( 'Export', 'woocommerce' ),
+						'import_products'          => __( 'Import', 'woocommerce' ),
+						'export_products'          => __( 'Export', 'woocommerce' ),
+						// translators: %d: number of selected products.
+						'export_selected_products' => __( 'Export %d selected', 'woocommerce' ),
 					),
 					'nonces'                            => array(
-						'gateway_toggle' => current_user_can( 'manage_woocommerce' ) ? wp_create_nonce( 'woocommerce-toggle-payment-gateway-enabled' ) : null,
+						'gateway_toggle'                 => current_user_can( 'manage_woocommerce' ) ? wp_create_nonce( 'woocommerce-toggle-payment-gateway-enabled' ) : null,
+						'export_selected_products_nonce' => wp_create_nonce( 'export-selected-products' ),
```
PanosSynetos: Similar to `'export_products' => current_user_can( 'export' )` below,  let's check if `current_user_can( 'export' )`

plugins/woocommerce/includes/admin/class-wc-admin-exporters.php
```diff
@@ -158,6 +158,16 @@ public function do_ajax_product_export() {
 			$exporter->set_product_category_to_export( wp_unslash( array_values( $_POST['export_category'] ) ) ); // WPCS: input var ok, sanitization ok.
 		}
 
+		// Set specific product IDs if provided.
+		if ( ! empty( $_POST['export_product_ids'] ) ) { // WPCS: input var ok.
+			$ids_raw     = explode( ',', sanitize_text_field( wp_unslash( $_POST['export_product_ids'] ) ) ); // WPCS: input var ok, sanitization ok.
+			$product_ids = array_filter( array_map( 'absint', $ids_raw ) );
```
PanosSynetos: `set_product_ids_to_export` is also doing the array_filter/absint . I don't think it's necessary to do it here too.

plugins/woocommerce/includes/admin/views/html-admin-page-product-export.php
```diff
@@ -12,16 +12,68 @@
 wp_enqueue_script( 'wc-product-export' );
 
 $exporter = new WC_Product_CSV_Exporter();
+
+$product_ids_to_export = array();
+$is_exporting_specific = false;
```
PanosSynetos: [nitpick] I'd rename this variable to something like `is_exporting_product_ids` - `specific` is a newly introduced "word/terminology" and can be confusing

plugins/woocommerce/includes/admin/views/html-admin-page-product-export.php
```diff
@@ -12,16 +12,68 @@
...
+
+if ( ! empty( $_GET['product_ids'] ) ) {
+	check_admin_referer( 'export-selected-products' );
+
+	$ids_raw               = explode( ',', sanitize_text_field( wp_unslash( $_GET['product_ids'] ) ) );
+	$product_ids_to_export = array_filter( array_map( 'absint', $ids_raw ) );
+
+	if ( ! empty( $product_ids_to_export ) ) {
```
PanosSynetos: Let's convert this to an inline if, for simplicity

plugins/woocommerce/includes/admin/views/html-admin-page-product-export.php
```diff
@@ -12,16 +12,68 @@
...
 	<h1><?php esc_html_e( 'Export Products', 'woocommerce' ); ?></h1>
 
+	<?php
+	if ( $is_exporting_specific ) {
+		$clear_url = remove_query_arg( 'product_ids' );
+		$count     = count( $product_ids_to_export );
+		$notice    = sprintf(
+			// translators: %1$d: Number of products, %2$s: URL to clear selection.
+			_n(
+				'You are exporting the %1$d product you selected in the previous step. <a href="%2$s">Clear selection</a> to export all products.',
```
PanosSynetos: Pinging @jimjasson for a copy review here. At least the singular one could be 
```
You are exporting the product you selected in the previous step.
-or-
You are about to export one product. Clear selection to export all products.
```

Jason, suggestions? :)

plugins/woocommerce/includes/export/class-wc-product-csv-exporter.php
```diff
@@ -56,6 +56,13 @@ class WC_Product_CSV_Exporter extends WC_CSV_Batch_Exporter {
 	 */
```
PanosSynetos, line 56: Being good citizens, we can change this to `array` :)

plugins/woocommerce/includes/export/class-wc-product-csv-exporter.php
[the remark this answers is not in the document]
PanosSynetos: I like that you think in an extensible way, but the filter here seems redundant.
Further down, there's already a filter to modify the args
```
		$products = wc_get_products( apply_filters( "woocommerce_product_export_{$this->export_type}_query_args", $args ) );
```

I like that you protect the ids, one final time, so it could be done like

```
		$args = apply_filters( "woocommerce_product_export_{$this->export_type}_query_args", $args );
        // If $args['include'] then array_map/absint before we pass it to wc_get_products
		$products = wc_get_products( $args );
```

plugins/woocommerce/includes/export/class-wc-product-csv-exporter.php
PanosSynetos: I'd also add here, that it's a less expensive operation to check for the `! empty( $args['include'] ) || ! empty( $args['category'] )` first and then `$product->is_type( ProductType::VARIABLE )`

PanosSynetos (changes requested):
Hey @naman03malhotra  - Grear work here! Great PR description and very helpful to get the context of the changes and dive in. Respect!

I've tested various scenarios and tried to break the selective exporter and I couldn't. That's awesome.

I tested by exporting
- Bundles
- Variable Products
- Variable Subscriptions
- Grouped Products
- Simple Products

I exported up to the maximum of 999 products, and worked well. 

Like @jimjasson , I also used the `woocommerce_product_export_specific_product_ids` filter, and tried to break it by passing invalid values, and it went well - great defensive coding.

I'd suggest adding some E2E tests for the selective export functionality, especially if there are tests for the exporter.

I'm sharing some more specific comments inline - let's discuss if you think otherwise.

plugins/woocommerce/includes/admin/views/html-admin-page-product-export.php
jimjasson: > You are about to export 1 product. To export all products, clear your selection.

plugins/woocommerce/includes/export/class-wc-product-csv-exporter.php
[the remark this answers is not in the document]
PanosSynetos: @xristos3490 what are your thoughts on the above? Initially I thought that an additional filter is not needed, but then we discussed with @naman03malhotra and he mentioned that the `woocommerce_product_export_` dynamic filter was not identified by one of our HVMs.

Now I believe that it makes sense to introduce the new, more explicit filter, as Naman did in this PR, but wanted to hear your thoughts too.

Thank you!

plugins/woocommerce/includes/export/class-wc-product-csv-exporter.php
naman03malhotra: Changed as part of https://github.com/woocommerce/woocommerce/pull/57500/commits/f4a8bf5df0fcad7e4db62403c36b615f21ce5749

plugins/woocommerce/includes/export/class-wc-product-csv-exporter.php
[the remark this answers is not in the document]
naman03malhotra: >I like that you protect the ids, one final time, so it could be done like

Made changes as part of https://github.com/woocommerce/woocommerce/pull/57500/commits/f4a8bf5df0fcad7e4db62403c36b615f21ce5749

plugins/woocommerce/includes/export/class-wc-product-csv-exporter.php
naman03malhotra: Changed as part of https://github.com/woocommerce/woocommerce/pull/57500/commits/f4a8bf5df0fcad7e4db62403c36b615f21ce5749

plugins/woocommerce/includes/admin/views/html-admin-page-product-export.php
```diff
@@ -12,16 +12,66 @@
...
+$product_ids_to_export    = array();
+$is_exporting_product_ids = false;
+
+if ( ! empty( $_GET['product_ids'] ) ) {
+	check_admin_referer( 'export-selected-products' );
+
+	$ids_raw               = explode( ',', sanitize_text_field( wp_unslash( $_GET['product_ids'] ) ) );
+	$product_ids_to_export = array_filter( array_map( 'absint', $ids_raw ) );
+
+	$is_exporting_product_ids = ! empty( $product_ids_to_export ) ? true : false;
```
jimjasson, line 24: Let's fix the `=` alignment here

plugins/woocommerce/includes/admin/views/html-admin-page-product-export.php
```diff
@@ -12,16 +12,66 @@
...
+}
 ?>
+
```
jimjasson: Let's remove this line to avoid having it printed in the HTML.

jimjasson (approved):
Thanks for making these changes @naman03malhotra! 

I left two small comments about the new changes you did. 

I re-tested the export flow and confirmed that it still works as expected. Let's update the PR description to remove the mention of the snippet, given that this hook has since been removed.

plugins/woocommerce/includes/admin/views/html-admin-page-product-export.php
naman03malhotra: https://github.com/woocommerce/woocommerce/pull/57500/commits/675177433a0c379591165d52254dccb7d6199b4e

plugins/woocommerce/includes/admin/views/html-admin-page-product-export.php
naman03malhotra: https://github.com/woocommerce/woocommerce/pull/57500/commits/675177433a0c379591165d52254dccb7d6199b4e

PanosSynetos (approved):
I've tested all the changes, re-run some imports and we're good to go.

I also tested the snippet you provided, and works as expected

```php
function my_filter_product_export_by_ids( $args ) {
    $my_product_ids_to_export = array( 10, 20, 30 ); // Define your list of product IDs
 
    // Set the 'include' argument to your specific list.
    // This tells wc_get_products to fetch only these products.
    $args['include'] = $my_product_ids_to_export;
    return $args;
}
 
add_filter( 'woocommerce_product_export_product_query_args', 'my_filter_product_export_by_ids' );
```

PR #55305, opened 2025-02-10
Add brands support to Store API

senadir:
This PR is part of modernizing Store API and filling missing gaps, it is cherry-picked from https://github.com/woocommerce/woocommerce/pull/54287

In this PR, I add support for brands, including filtering products by brand, and getting brands by slug and ID.

### How to test the changes in this Pull Request:

Using the [WooCommerce Testing Instructions Guide](https://github.com/woocommerce/woocommerce/wiki/Writing-high-quality-testing-instructions), include your detailed testing instructions:

1. In a WooCommerce store, add 2-3 brands, one of them named `apple` and with a slug of `apple`
2. Add them to some products.
3. Using a REST client, hit `YOUR_SITE.example/wp-json/wc/store/products/brands` make sure you're getting brands back.
4. Hit `YOUR_SITE.example/wp-json/wc/store/products/brands/apple` make sure you get one brand back.
5. Filter some products by brand `YOUR_SITE.example/wp-json/wc/store/products?brand=apple` and make sure you're getting only those products.

### Changelog entry

-   [x] Automatically create a changelog entry from the details below.

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Details</summary>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

-   [ ] Fix - Fixes an existing bug
-   [ ] Add - Adds functionality
-   [ ] Update - Update existing functionality
-   [ ] Dev - Development related task
-   [ ] Tweak - A minor adjustment to the codebase
-   [ ] Performance - Address performance issues
-   [x] Enhancement - Improvement to existing functionality

#### Message 
Add WooCommerce brands to Store API response and filter products by brand ID and slug.
</details>

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

PR #57507, opened 2025-04-24
fix: regression in #55305

Tofandel:
Fixes Bug introduced in PR #55305 .

| Before | After |
| ------ | ----- |
|    PHP Warning:  Trying to access array offset on int in plugins/woocommerce/src/StoreApi/Utilities/ProductQuery.php on line 128 image   |    No warning  and image  |

Before this PR passing an id like ?category=1 to the request would not return any result anymore because the check would use `slug` instead of `term_id` which would never match anything

It now correctly handles both arrays or simple fields

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 
Fix warning in ProductQuery where no result would be returned if passing a term id
</details>

### DIFF
--- a/plugins/woocommerce/changelog/57507-patch-7
+++ b/plugins/woocommerce/changelog/57507-patch-7
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+Comment: Fix warning in ProductQuery where no result would be returned if passing a term id
+
--- a/plugins/woocommerce/src/StoreApi/Utilities/ProductQuery.php
+++ b/plugins/woocommerce/src/StoreApi/Utilities/ProductQuery.php
@@ -123,7 +123,7 @@ class ProductQuery {
 		// Set tax_query for each passed arg.
 		foreach ( $taxonomies as $taxonomy => $key ) {
 			if ( ! empty( $request[ $key ] ) ) {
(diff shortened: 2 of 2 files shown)


gigitux:
Thanks for opening this PR! To help the reviewers and testers, could you write some testing instructions?

mikejolley:
@Tofandel I too would like more instructions on reproduction steps. I did try `store.local/wp-json/wc/store/v1/products?category=16` but the endpoint worked and there were no notices. Tested PHP 7.x and 8.x—same results. Could you explain what you're doing to see the notices?

mikejolley:
I've looked into this again and cannot see a problem with the existing logic. The routes sanitize taxonomy args using `wp_parse_list` which produces a numerically indexed array, therefore if I were to GET:

```
store.local/wp-json/wc/store/v1/products?category=16
```

This would provide the `ProductQuery` class this value:

```php
array(1){
	[0]=>string(2)"16"
}
```

The existing logic handles this correctly:

```php
$type = is_numeric( $request[ $key ][0] ) ? 'term_id' : 'slug';
```

This would check if `"16"` is numeric and use it as a term_id. This will always be an array of term_id or slug. 

I think I will close this PR since we haven't heard back how to replicate your issue. If you want to provide more details that would be great and we can reconsider.

Thanks again

PR #57514, opened 2025-04-24
Local Pickup labels are inconsistent on confirmation page

### LINKED ISSUE #57506: Local Pickup labels are inconsistent on confirmation page
## **Describe the bug**

When choosing pickup on checkout, the confirmation page shows that collection was used, but the display can be incorrect or lack details depending on local pickup settings.

1. If a cost is entered for local pickup, the cost is missing in the display
2. If the pickup location does not have a complete address, pickup details are missing in the display

## **To Reproduce**

* Settings > Shipping > Local Pickup
* Enter cost
* Create pickup locations with/without addresses
* Place order and view confirmation

## **Expected behavior**

Prices should still be visible regardless of pickup method, and details such as the collection address, name, and details should be visible if set by the merchant.

BurakParsAydin:

Closes #57506  .

| Before  |

|     
Screenshot from 2025-04-24 23-49-38

Screenshot from 2025-04-24 23-41-38
|  
|   After |
Screenshot from 2025-04-24 23-43-34
Screenshot from 2025-04-24 23-43-45

|  

1. Settings > Shipping > Local Pickup
2. Enter cost
3. Create pickup locations with/without addresses
4. Place order and view confirmation

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 
Ensure local pickup cost is displayed on the order confirmation page when a fee is configured. 
Show pickup location name and merchant-provided details even if the pickup address is incomplete or missing.

</details>

### DIFF
--- a/plugins/woocommerce/src/Blocks/Shipping/ShippingController.php
+++ b/plugins/woocommerce/src/Blocks/Shipping/ShippingController.php
@@ -129,16 +129,40 @@ class ShippingController {
 		$details         = $shipping_method->get_meta( 'pickup_details' );
 		$location        = $shipping_method->get_meta( 'pickup_location' );
 		$address         = $shipping_method->get_meta( 'pickup_address' );
+		$cost            = $shipping_method->get_total();
 
-		if ( ! $address ) {
+		$section = '';
+
+		if ( $location ) {
+			$section .= sprintf(
+				// Translators: %s location name.
+				__( 'Collection from <strong>%s</strong>:', 'woocommerce' ),
+				$location
+			);
+		}
+
+		if ( $address ) {
+			$section .= '<br/><address>' . nl2br( esc_html( str_replace( ',', ', ', $address ) ) ) . '</address>';
+		}
+
+		if ( $details ) {
+			$section .= '<br/>' . wp_kses_post( $details );
+		}
+
+		if ( $cost > 0 ) {
+			$section .= '<br/>' . sprintf(
+				// Translators: %s is the formatted price.
+				__( 'Pickup cost: %s', 'woocommerce' ),
+				wc_price( $cost, array( 'currency' => $order->get_currency() ) )
+			);
+		}
+
+		// If nothing is available, return original.
+		if ( empty( $section ) ) {
 			return $return_value;
 		}
 
-		return sprintf(
-			// Translators: %s location name.
-			__( 'Collection from <strong>%s</strong>:', 'woocommerce' ),
-			$location
-		) . '<br/><address>' . str_replace( ',', ',<br/>', $address ) . '</address><br/>' . $details;
+		return $section;
 	}
 
 	/**
--- a/plugins/woocommerce/src/Blocks/Shipping/ShippingController.php
+++ b/plugins/woocommerce/src/Blocks/Shipping/ShippingController.php
@@ -150,10 +150,41 @@ class ShippingController {
 		}
 
 		if ( $cost > 0 ) {
+			$tax_display = get_option( 'woocommerce_tax_display_cart' );
+			$tax = $shipping_method->get_total_tax();
+
+			// Format cost with tax handling
+			if ( 'excl' === $tax_display ) {
+				// Show pickup cost excluding tax
+				$formatted_cost = wc_price( $cost, array( 'currency' => $order->get_currency() ) );
+				if ( (float) $tax > 0 && $order->get_prices_include_tax() ) {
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/src/Blocks/Shipping/ShippingController.php
```diff
@@ -129,16 +129,40 @@ public function show_local_pickup_details( $return_value, $order ) {
 		$details         = $shipping_method->get_meta( 'pickup_details' );
 		$location        = $shipping_method->get_meta( 'pickup_location' );
 		$address         = $shipping_method->get_meta( 'pickup_address' );
+		$cost            = $shipping_method->get_total();
 
-		if ( ! $address ) {
+		$section = '';
```
mikejolley: This might be cleaner if we avoid adding `<br/>` to the beginning of the strings below and instead add each item to an array. Then we can implode with `<br/>` later and return the result.

plugins/woocommerce/src/Blocks/Shipping/ShippingController.php
```diff
@@ -129,16 +129,40 @@ public function show_local_pickup_details( $return_value, $order ) {
...
+		if ( $location ) {
+			$section .= sprintf(
+				// Translators: %s location name.
+				__( 'Collection from <strong>%s</strong>:', 'woocommerce' ),
+				$location
+			);
+		}
+
+		if ( $address ) {
+			$section .= '<br/><address>' . nl2br( esc_html( str_replace( ',', ', ', $address ) ) ) . '</address>';
```
mikejolley: We end up with an extra line break after this because `<address>` is block level. Do you think it would work better if we wrapped each item with `<p>` or as a list instead of using line breaks?

plugins/woocommerce/src/Blocks/Shipping/ShippingController.php
```diff
@@ -129,16 +129,40 @@ public function show_local_pickup_details( $return_value, $order ) {
...
+
+		if ( $details ) {
+			$section .= '<br/>' . wp_kses_post( $details );
+		}
+
+		if ( $cost > 0 ) {
+			$section .= '<br/>' . sprintf(
+				// Translators: %s is the formatted price.
+				__( 'Pickup cost: %s', 'woocommerce' ),
+				wc_price( $cost, array( 'currency' => $order->get_currency() ) )
```
mikejolley: This doesn't deal with prices inclusive or exclusive of tax like the `get_shipping_to_display` method does. What if we used the original `$return_value` instead of regenerating it? Either that or we need to account for the logic in the `get_shipping_to_display` method.

mikejolley (changes requested):
Thanks for looking into this. I think there are a few things we'd want to take care of here before accepting, the main one being the cost. 

See here: https://github.com/woocommerce/woocommerce/blob/2afc098512d2f6782339ca0b34c9320c70acdb57/plugins/woocommerce/includes/abstracts/abstract-wc-order.php#L2247-L2268

There is handling to support display of taxes which we would need to keep if possible.

What do you think?

BurakParsAydin:
@mikejolley Thanks for the feedback! I’ve addressed all but one point. I didn’t get a chance to look into it today, but I’ll follow up and do a full test once I’m back from a short holiday on Monday - hope that’s okay!

mikejolley:
Great, thanks @BurakParsAydin

BurakParsAydin:
@mikejolley I've made the necessary adjustments based on your suggestions. Let me know if anything else stands out or needs a second look. Appreciate your time reviewing this 🙏

mikejolley (approved):
Thanks for making those adjustments. I ran through a few scenarios and it seems to be testing well.

Screenshot 2025-05-08 at 11 45 54
Screenshot 2025-05-08 at 11 47 49
Screenshot 2025-05-08 at 11 48 33
Screenshot 2025-05-08 at 11 49 27

I also confirmed that data is escaped on render later on via `render_order_details_table_totals`.

I've just allowed the CI to run, so as long as the tests pass this can be merged. Thanks again 👍🏻

BurakParsAydin:
@mikejolley Thanks for the thorough review and confirmation earlier! The previous CI failures were related to PHP linting - these should now be resolved with the latest commit I’ve just pushed.

Let me know if there's anything else you'd like me to adjust.

mikejolley:
Thanks, I'm merging now 👍🏻

PR #57516, opened 2025-04-25
Product Details: only render non-empty accordion items

dinhtungdu:

This PR improves the Product Details block by hiding empty accordion items when they have no relevant content to display. Accordion Item can have inner blocks but those blocks can return empty string or empty tags.

| Before | After |
| ------ | ----- |
| image |   image    | 

1. Add `Blockified Product Details` block to the Single Product template.
2. View a product that doesn't have any weight/dimensions/attribute.
3. Doesn't see the Additional Information item.
4. View another product that has weight/dimensions/attribute.
5. See the Additional Information item is visible and clicking on the header expands the item's content.
6. Clear the description of a product.
7. On the frontend, see the Description tab is hidden.

### DIFF
--- a/plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
@@ -2,6 +2,9 @@
 declare(strict_types=1);
 namespace Automattic\WooCommerce\Blocks\BlockTypes;
 
+use WP_Block;
+use WP_HTML_Tag_Processor;
+
 /**
  * BlockifiedProductDetails class.
  */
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
```diff
@@ -32,6 +35,67 @@ protected function get_block_type_style() {
...
+			},
+			$accordion_items
+		);
+
+		$p = new WP_HTML_Tag_Processor( $content );
+
+		$counter = 0;
+		while ( $p->next_tag( array( 'class_name' => 'wp-block-woocommerce-accordion-item' ) ) ) {
+			if ( ! $accordion_items_visibility[ $counter ] ) {
+				$p->set_attribute( 'style', 'display:none;' );
```
Aljullu, line 76: Should we also set the [`hidden` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/hidden) for a11y reasons?

plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
```diff
@@ -32,6 +35,67 @@ protected function get_block_type_style() {
...
+			return $content;
+		}
+
+		$accordion_items_visibility = array_map(
+			function ( $item ) use ( $block ) {
+				$content_block          = end( $item['innerBlocks'] );
+				$rendered_content_block = ( new WP_Block( $content_block, $block->context ) )->render();
+				$p                      = new WP_HTML_Tag_Processor( $rendered_content_block );
+
+				return $p->next_tag( 'img' ) || $p->next_tag( 'video' ) || $p->next_tag( 'iframe' ) || ! empty( wp_strip_all_tags( $rendered_content_block, true ) );
```
Aljullu: Out of curiosity, how were ``, `<video>` and `<iframe>` tags selected? Wondering if we should include others like `<figure>` or `<meter>`, for example. :thinking:

Aljullu (approved):
Tested and confirming it works as described, the empty accordion elements aren't rendered:

Before | After
--- | ---
imatge | imatge

I left a couple of questions/suggestions inline, but they aren't blockers, so pre-approving.

<details>

* Theme: Twenty Twenty-Five
* Store size: small
</details>

plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
dinhtungdu: The reason is there isn't any text inside those selected tags, but they can still have content. `<figure>` always wraps some elements, so we don't need to check for it. `<meter>` is a good addition, TIL. I will add it to the list.

plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
dinhtungdu: Good point! Sure!

PR #57519, opened 2025-04-25
Product Gallery: optimise block

kmanijak:

This PR introduces number of changes to Product Gallery block in order to optimise for initial loading as well as for number and size of images we're loading. This includes also other optimisations on different level):

1. Previously we didn't utilise SSR with iAPI. Image data was put into context, then it was manipulated in the client-side script before the actual rendering was happening. Initial reason for it was performance (to slice data and fetch only initial images). But the initial loading was hurt drastically. This was changed so we rely on SSR by putting the image data into context and apply optimisations elsewhere.
2. iAPI context has been simplified significantly. Previously we kept all the image data in context (which gets rendered in the markup and looks messy). Data was reduced to just an array of image IDs (check table below for comparison with 20 images.
3. As image data in context was a single source of truth: thumbnails, large image and dialog were getting the same FULL size images. I extended the method preparing image data with `size` and:
    - thumbnails fetch images for size `woocommerce_thumbnail` forcing the lowest srcset resolution
    - large image fetches images for size `woocommerce_single`
    - dialog fetches images for size `full`
4. I switched from rendering `<template>` element with `wp-each` method. Since image data is static, there's no point in doing that. I switched to regular PHP markup which allows us for more flexibility, for example I can attached `is-active` class to the first Thumbnail to avoid flickering when initial state is calculated
6. Minor CSS improvements
7. Last but not least, I'm happy the codebase decreased by 200 LOC

In addition, I made some cleanup in tests by simplifying some methods finding blocks. I also changed a way the `active` large image is detected. Previously it was through the `class` which was attached to the active image. It was bad for two reasons:
1. This was the only reason for this class to exist and I don't think implementation should be adjusted for tests but tests to the implementation
2. Class was attached to the image without guarantee the image was actually displayed as chosen
So in this case it got a bit more complex (file: plugins/woocommerce/client/blocks/tests/e2e/tests/product-gallery/product-gallery.page.ts) as we need to manually find what image is displayed. Playwright doesn't provide a locator that would allow recognising displayed element in overflow container. Using `:visible` doesn't solve it as technically all of the images are visible from Playwright POV (neither is `:hidden`) hence the manual implementation. 

Closes WOOPLUG-3798

Case | Before | After
--- | ------ | -----
Video (no throttling)    |  https://github.com/user-attachments/assets/c9ed222d-014d-4d2c-bbea-296356a3adeb  | https://github.com/user-attachments/assets/ef8e593b-f990-4b78-88ac-b21339889e82
Video (Slow 4G) |https://github.com/user-attachments/assets/2aa53a3d-e6b9-456c-ab23-d9c1abf650a2  | https://github.com/user-attachments/assets/be817f9e-9421-4e25-a8fa-723c7b60b862
Performance metrics | image |  image
Lighthouse report (desktop) | image | image
Lighthouse report (mobile) |image | image
Transferred resources | image | image
Context in markup | image | image 

#### Smoke testing

0. Create product with multiple images, 15+
1. Go to Editor > Single Product template
3. Visit the block and smoke test it:
    - change images via thumbnails
    - change images via arrows
    - open dialog
    - zoom on hover
    - visit variable product and make sure image changes along with changing variables

#### Performance test

1. Compare loading time on the same Single Product page on `trunk` vs this branch. This branch should be noticeably faster
2. Run Lighthouse check on `trunk` vs this branch. This branch should be noticeably faster

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryThumbnails.php
```diff
@@ -92,25 +91,23 @@ class="wc-block-product-gallery-thumbnails <?php echo esc_attr( $classes_and_sty
...
-							data-wp-bind--srcset="context.image.srcset"
-							data-wp-bind--sizes="context.image.sizes"
+							class="wc-block-product-gallery-thumbnails__thumbnail__image <?php echo 0 === $index ? 'is-active' : ''; ?>"
+							data-image-id="<?php echo esc_attr( $image['id'] ); ?>"
+							src="<?php echo esc_attr( $image['src'] ); ?>"
+							srcset="<?php echo esc_attr( $image['srcset'] ); ?>"
+							sizes="<?php echo esc_attr( $image['sizes'] ); ?>"
 							data-wp-on--click="actions.selectCurrentImage"
 							data-wp-on--keydown="actions.onThumbnailKeyDown"
+							data-wp-watch="callbacks.toggleActiveImageAtrributes"
```
gigitux: There is a typo: `toggleActiveImageAtrributes`

plugins/woocommerce/client/blocks/tests/e2e/tests/product-gallery/inner-blocks/product-gallery-large-image/product-gallery-large-image.block_theme.spec.ts
```diff
@@ -327,17 +298,12 @@ test.describe( `${ blockData.name }`, () => {
 
 			// Timeout is needed to allow the image animation to finish.
 			// eslint-disable-next-line playwright/no-wait-for-timeout, no-restricted-syntax
-			await page.waitForTimeout( 1000 );
+			await page.waitForTimeout( 400 );
```
gigitux: Could we try to use `expect.toPass` (https://playwright.dev/docs/test-assertions#expecttopass) instead of this timeout?

plugins/woocommerce/client/blocks/tests/e2e/tests/product-gallery/product-gallery.block_theme.spec.ts
```diff
@@ -169,11 +153,12 @@ test.describe( `${ blockData.name }`, () => {
...
 				.click();
 
-			const newVisibleLargeImageId = await getVisibleLargeImageId(
-				await pageObject.getMainImageBlock( {
-					page: 'frontend',
-				} )
-			);
+			// Wait for the transition to complete
+			// eslint-disable-next-line playwright/no-wait-for-timeout, no-restricted-syntax
+			await page.waitForTimeout( 400 );
```
gigitux: Same

plugins/woocommerce/tests/php/src/Blocks/Utils/ProductGalleryUtilsTest.php
gigitux, line 1: Not for this PR, but it would be great to add a test case for products without images.

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryLargeImage.php
```diff
@@ -122,83 +117,42 @@ function ( $carry, $key ) use ( $directives ) {
...
+				<?php foreach ( $image_data as $image ) : ?>
+					<li class="wc-block-product-gallery-large-image__wrapper">
 						<img
 							class="<?php echo esc_attr( $base_classes ); ?>"
-							data-wp-bind--src="context.image.src"
-							data-wp-bind--srcset="context.image.srcset"
-							data-wp-bind--sizes="context.image.sizes"
-							data-wp-bind--data-image-id="context.image.id"
-							data-wp-bind--tabindex="context.image.tabIndex"
+							src="<?php echo esc_attr( $image['src'] ); ?>"
```
gigitux, line 129: To increase the LCP metric, what do you think if we add the `fetchpriority` for the first image and lazy load the rest of the images?

```php
<?php if ( 0 === $index ) : ?>
	fetchpriority="high"
<?php else : ?>
	fetchpriority="low"
	loading="lazy"
<?php endif; ?>
```


| Before | After |
|--------|--------|
|image  | image|

gigitux (approved):
Awesome work! 🚀 

I left a few comments, but besides the typo, nothing is a blocker. I think that the `fetchpriority` approach is easy to implement, and it looks like that it is suggested as well (https://web.dev/articles/fetch-priority), but I'm fine if you prefer not to implement in this PR :D 


Also, I noticed two issues, but as above, feel free to address in dedicated PRs:

## Flickering Thumbnails

There is a small flickering effect regarding the Thumbnails:


https://github.com/user-attachments/assets/50adcc0c-7ad0-4f5d-9b37-5da73a9abe79

## Disabled button on the editor side

I don’t think the left button in the editor should appear **disabled** . I was confusing, since clicking it still focuses the `Next/Previous Buttons` block.

https://github.com/user-attachments/assets/26f2c7e6-b8cd-457e-9a50-5312f787e2cd

plugins/woocommerce/tests/php/src/Blocks/Utils/ProductGalleryUtilsTest.php
kmanijak: Good idea! Didn't think of it 😅

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryThumbnails.php
kmanijak: Fixed: https://github.com/woocommerce/woocommerce/pull/57519/commits/3632e53a9db4e226eb31205df8c6dca7cbced4e8

plugins/woocommerce/client/blocks/tests/e2e/tests/product-gallery/inner-blocks/product-gallery-large-image/product-gallery-large-image.block_theme.spec.ts
kmanijak: Good idea. I thought timeout is acceptable here as we know for sure the transition is happening but actually `expect.toPass` is more universal! Done https://github.com/woocommerce/woocommerce/pull/57519/commits/226cddd88264ed066aa224b451261a7aeef324a7

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryLargeImage.php
kmanijak: Amazing idea! Done https://github.com/woocommerce/woocommerce/pull/57519/commits/c2b21d030cfd0c30adecb18e8df8815b1ab28660

kmanijak:
@gigitux, thanks a lot for a review and thorough testing!

> Flickering Thumbnails
> There is a small flickering effect regarding the Thumbnails:

Yeah, I couldn't figure out how to get rid of this. We control the gradient overflow by iAPI context. To remove the flickering we'd have to know the initial state if the gradient appears on the bottom or on the right (if anywhere at all). So at the beginning the values are `false`. Then [`on-init` the overflow is calculated](https://github.com/woocommerce/woocommerce/blob/08894a09bb88646bda313cf826fcda6a058202a6/plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryThumbnails.php#L93) and initial state is set with the delay.

There was the same problem with initial active thumbnail which I'm solving with `<?php echo 0 === $index ? 'is-active' : ''; ?>` because I know for sure the first thumbnail will be active. I can't know during rendering if overflow appears and where 😞 I could potentially guess 🤔 But I'd leave it for another iteration TBH unless you have some idea? 🤔 

> Disabled button on the editor side
> I don’t think the left button in the editor should appear disabled . I was confusing, since clicking it still focuses the Next/Previous Buttons block.

I see what you mean. So I think it should be disabled to showcase how it would look like but I don't think it should be interactive. Basically there should not be a different cursor or right button should not be focusable. Rather than that, clicking on both arrows should just focus on block. I'll address it separately 🙌

gigitux:
> But I'd leave it for another iteration TBH unless you have some idea?

I'm fine to fix this in a dedicated PR 👍 

> Basically there should not be a different cursor or right button should not be focusable.

> I see what you mean. So I think it should be disabled to showcase how it would look like but I don't think it should be interactive. Basically there should not be a different cursor or right button should not be focusable. Rather than that, clicking on both arrows should just focus on block. I'll address it separately 🙌

🚀 

Thanks for addressing the feedback 🙇

PR #57526, opened 2025-04-25
Fix blocks registration on test env

### LINKED ISSUE #54812: [Unit Test] Block Registration fallback to legacy method in unit test environment due to missing `block.json` access
## Description  
In the unit test environment, block registration falls back to the legacy method instead of using `register_block_type_from_metadata`.  

The root cause is that [`get_block_metadata_path`](https://github.com/woocommerce/woocommerce/blob/9a95cbbf3f36df09396fdd4c667b0f1283c5911c/plugins/woocommerce/src/Blocks/BlockTypes/AbstractBlock.php#L222-L223) returns `false` because `block.json` files are inaccessible in the PHP unit test environment.  

## Steps to Reproduce  
1. Add a log after this logic:

https://github.com/woocommerce/woocommerce/blob/9a95cbbf3f36df09396fdd4c667b0f1283c5911c/plugins/woocommerce/src/Blocks/BlockTypes/AbstractBlock.php#L254-L262

2. Run unit tests that involve block registration. [For instance: BlockifiedProductDetails.](https://github.com/woocommerce/woocommerce/blob/9a95cbbf3f36df09396fdd4c667b0f1283c5911c/plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php#L8)
3. Observe that your log is printend, so the fallback to the legacy registration method.  

## Expected Behavior  
Block registration should use `register_block_type_from_metadata`.

This is essential because blocks relying on `attributes`, `supports`, and `uses_context` will lack the necessary data. 

For instance, for the `attributes` object, the default value will be undefined.

Without proper metadata registration, functions like `get_block_type_attributes`, `get_block_type_supports`, and `get_block_type_uses_context` are not defined for new blocks, leading to incomplete block functionality.  

https://github.com/woocommerce/woocommerce/blob/9a95cbbf3f36df09396fdd4c667b0f1283c5911c/plugins/woocommerce/src/Blocks/BlockTypes/AbstractBlock.php#L265-L272

gigitux:

This PR ensure that `block.json` files of blocks are available when the unit test suite runs. Currently, `block.json` files are copied/pasted into the build folder thanks to this webpack configuration:

https://github.com/woocommerce/woocommerce/blob/a9b6620fc3ec4249781226aa6f068a716e108b66/plugins/woocommerce/client/blocks/bin/webpack-configs.js#L229-L255 

On the CI, when unit test suite runs, the WooCommerce plugin isn't built and for this reason, `block.json` files aren't available and this check fails:

https://github.com/woocommerce/woocommerce/blob/a9b6620fc3ec4249781226aa6f068a716e108b66/plugins/woocommerce/src/Blocks/BlockTypes/AbstractBlock.php#L273-L280

This causes bugs that I mentioned in #54812.

In this PR, I introduced a bash script that mimics the `CopyWebpackPlugin` behavior. In this way, we can avoid building the project, but make `block.json` files available and ensure that the blocks are registered correctly.

To ensure that everything works as expected, I added a unit test for the Catalog Sorting block. According to the `block.json`, by default, the block should have the font sets to small:

https://github.com/woocommerce/woocommerce/blob/a9b6620fc3ec4249781226aa6f068a716e108b66/plugins/woocommerce/client/blocks/assets/js/blocks/catalog-sorting/block.json#L19-L28

In this PR, the unit test passes because the block is registered via `block.json` (so, the default value is set). In #57525 the test fails due to the absence of the block.json.

Closes #54812 (WOOPLUG-3061) .

1. Ensure that unit test jobs pass. 

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Fix blocks registration on test env.

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/bin/copy-blocks-json.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# Create target directory if it doesn't exist
+mkdir -p "$TARGET_DIR"
+
+# Find all block.json files
+find plugins/woocommerce/client/blocks/assets/js -name "block.json" | while read file; do
+    # Read the block name from the JSON file
+    block_name=$(cat "$file" | grep -o '"name": "[^"]*"' | cut -d'"' -f4 | cut -d'/' -f2)
+
+    # Check if it's a parent block by looking for "parent" field
```
kmanijak:
```suggestion
    # Check if it's an inner block by looking for "parent" field
```

plugins/woocommerce/client/blocks/bin/copy-blocks-json.sh
```diff
@@ -0,0 +1,44 @@
...
+    if grep -q '"parent":' "$file"; then
```
kmanijak: If it needs to mimic the CopyWebpackPlugin behavior, I think we should also include the second condition there which is `! genericBlocks[ blockName ]`. Otherwise these generic blocks won't have `block.json` in the right spot and problem won't be addressed for them, no? 🤔 

```
if (
	metadata.parent &&
	! genericBlocks[ blockName ]
)
```

`genericBlocks` are [listed here](https://github.com/woocommerce/woocommerce/blob/2afc098512d2f6782339ca0b34c9320c70acdb57/plugins/woocommerce/client/blocks/bin/webpack-entries.js#L212).

kmanijak (commented):
I left one comment regarding `genericBlocks` condition. If you don't think it's needed, I'm happy to approve!

plugins/woocommerce/client/blocks/bin/copy-blocks-json.sh
gigitux: Nice catch! I fixed with 3c1fe64fbdac3ded43fded07f6b610a749bebae2

kmanijak (approved):
Nice one! I tested locally and `block.json` files are copied correctly to the right places:
- blocks with no `parent` to `plugins/woocommerce/assets/client/blocks`
- blocks with `parent` but "generic" ones to `plugins/woocommerce/assets/client/blocks`
- other blocks with `parent` to `plugins/woocommerce/assets/client/blocks/inner-blocks`

👏 🚀

PR #57537, opened 2025-04-25
[Experimental] Blockified Add to Cart with Options block: disable Variation Selector options conditionally

### LINKED ISSUE #57535: Blockified Add to Cart with Options block: disable Variation Selector options based on current selected ones
In the existing Add to Cart form, the available options in the variation selector depend on the currently selected ones, so we don't allow shoppers to select invalid combinations of attributes.<br><br>We should do the same in the blockified Add to Cart with Options block. That includes both views: dropdowns and pills.

Aljullu:

Closes #57535.
Closes WOOPLUG-4073.

This PR makes it so pills and dropdown items are disabled in the Variation Selector block depending on the other selected attributes.

3. Create a variable product. Try to get complex scenarios with the possible variations, like making combinations where one variation allows "any" while another variation only allows one term, terms which are only used by one variation, but that variation has no price, or whatever you can think of. For example, this is how I set it up:

Variations | Variation with no price
--- | ---
imatge | imatge

4. Visit the product on the frontend. Verify you can select all valid variations, while you can't select any invalid combination of attributes.
5. Try using the keyboard to select some of the attributes, verify the invalid ones can't be selected.

https://github.com/user-attachments/assets/f2a93117-6900-4e65-825d-d4f7b9f49095

6. Go to the _Single Product_ template and change the display style to _Dropdown_. Repeat step 4.

### DIFF
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/VariationSelectorItemTemplate.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/VariationSelectorItemTemplate.php
@@ -98,7 +98,12 @@ class VariationSelectorItemTemplate extends AbstractBlock {
 	 *
 	 * @param string $attribute_name Product Attribute Name.
 	 * @param array  $attribute_terms Product Attribute Terms.
-	 * @return srtring
+	 * @return array[] Array of term data with structure:
+	 *                 [
+	 *                     'label'      => (string) Display label for the term.
+	 *                     'value'      => (string) Internal value/slug for the term.
+	 *                     'isSelected' => (bool)   Whether this term is the default selection.
+	 *                 ]
 	 */
 	protected function get_terms( $attribute_name, $attribute_terms ) {
 		global $product;
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/VariationSelectorItemTemplate.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/VariationSelectorItemTemplate.php
@@ -39,7 +39,7 @@ class VariationSelectorItemTemplate extends AbstractBlock {
 		$product_attributes = $product->get_variation_attributes();
 
 		foreach ( $product_attributes as $product_attribute_name => $product_attribute_terms ) {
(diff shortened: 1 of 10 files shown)


plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/VariationSelectorItemTemplate.php
```diff
@@ -50,13 +50,30 @@ protected function render( $attributes, $content, $block ): string {
...
 	 * @return string Row HTML
 	 */
-	private function get_product_row( $product_attribute_name, $product_attribute_terms, $attributes, $block ): string {
-		$attribute_name  = $product_attribute_name;
-		$attribute_terms = $this->get_terms( $product_attribute_name, $product_attribute_terms );
+	private function get_product_row( $product_attribute_name, $product_attribute_terms, $block ): string {
+		global $product;
+
+		$attribute_name     = $product_attribute_name;
+		$attribute_terms    = $this->get_terms( $attribute_name, $product_attribute_terms );
```
gigitux, line 60: Not related to this PR, but I guess that the PHPDoc of `get_terms` function is not correct (see return type - it is `srtring`): https://github.com/woocommerce/woocommerce/blob/b10c01e48fd5ecece0dec0a6a7e5ad31fb86deb7/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/VariationSelectorItemTemplate.php#L96-L103

gigitux (approved):
Except for the comment regarding the `get_terms` function, this PR LGTM and works as expected.

I tested this PR against this product configuration:

image





| Pills layout | Dropdown layout |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/0ffe179d-a883-4644-b0be-cb8ef9b00a7c" /> | <video src="https://github.com/user-attachments/assets/a9a56a5b-b24b-4c4d-b2de-f851a1b95e06" /> |

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/VariationSelectorItemTemplate.php
Aljullu: Ups, good catch. I totally missed it! Should be fixed in: c417e5821171ce010981f8345d72939d22889c8a.

Issue #57353, opened 2025-04-17
Order Fulfillment Entity Project


tpaksu:
PRD: [https://wooorderstatusp2.wordpress.com/2024/07/03/prd-woocommerce-shipping-fulfillments-entity/](https://wooorderstatusp2.wordpress.com/2024/07/03/prd-woocommerce-shipping-fulfillments-entity/)<br>TSD: [https://wooorderstatusp2.wordpress.com/2025/04/16/tsd-woocommerce-fulfillment-entity](https://wooorderstatusp2.wordpress.com/2025/04/16/tsd-woocommerce-fulfillment-entity)

This epic issue serves as an epic/container which groups together several sub-issues for tracking and managing progress of various tasks related to adding order fulfillment entities to enable merchants create and track order fulfillments on their backend. 

> [!Important]
> There shouldn't be any direct edits to this branch. 

> [!Note]
> All linked PR's that ended up being merged here have reviews and approvals on their own PR and branch.

A shared branch `feature/57353-order-fulfillments-entity` has been created and will serve as a feature-branch, meaning all subsequent PR's should be merged into this branch first, allowing us test and release everything together.

PR #57549, opened 2025-04-28
Add `AbstractFulfillmentManager` to manage fulfillments of an entity

### LINKED ISSUE #57358: [Order Fulfillments] Create WC_Fulfillable trait and connect to WC_Order
**Issue:** We need a bridge between WC_Order and fulfillments to be able to add fulfillment support to orders initially, in a way that other objects with items can utilize the same trait to add fulfillments support to them.

Proposed solution: Add the WC_Fulfillable trait outlined in [https://wooorderstatusp2.wordpress.com/2025/04/16/tsd-woocommerce-fulfillment-entity/#wc-fulfillable-trait](https://wooorderstatusp2.wordpress.com/2025/04/16/tsd-woocommerce-fulfillment-entity/#wc-fulfillable-trait) and attach to WC_Order object.

tpaksu:

This PR connects WC_Order with the fulfillment sub-system using a `FulfillableEntityTrait` trait. 

The trait requires four abstract methods to be implemented on the base class: 
- `get_fulfillment_id` needs to return the ID property of the entity
- `get_fulfillment_type` needs to return an identifier for the object
- `get_fulfillment_status` needs to return the stored status of the overall entity fulfillment
- `set_fulfillment_status` needs to save the overall fulfillment status handled by the fulfillment system to a place that it likes, that will eventually be returned back via `get_fulfillment_status` method.

The trait will add these functionality to the base object:
- `get_fulfillments()` will return the attached fulfillments of the entity.
- `add_fulfillment( Fulfillment $fulfillment )` adds a new fulfillment to the entity.
- `update_fulfillment( Fulfillment $fulfillment )` updates a single fulfillment to have new values.
- `delete_fulfillment( Fulfillment $fulfillment )` removes a single fulfillment from the entity.
- `fulfill( Fulfillment $fulfillment )` is a shorthand for setting the fulfillment as fulfilled. Notifications are not added yet to this function, and will be handled on a separate PR. 

Closes #57358 .

Part of #57353

> [!Note]
> You need to run `composer dumpautoload` in `plugins/woocommerce` folder to get the classes registered to composer autoloader.

The code should make sense, automated tests should pass.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

Add FulfillableEntityTrait to WC_Order to create a bridge with the fulfillment functionality.

</details>

### DIFF
--- a/plugins/woocommerce/src/Internal/Fulfillments/AbstractFulfillmentManager.php
+++ b/plugins/woocommerce/src/Internal/Fulfillments/AbstractFulfillmentManager.php
@@ -30,22 +30,6 @@ abstract class AbstractFulfillmentManager {
 	abstract public function get_fulfillable_type(): string;
 
 	/**
-	 * Fulfillable status, which is the status of the object with fulfillments. For example, `WC_Order::get_status()` is used for order fulfillments.
-	 *
-	 * @param string $status Fulfillable status.
-	 *
-	 * @var string
-	 */
-	abstract public function set_fulfillment_status( string $status ): void;
-
-	/**
-	 * Fulfillable status, which is the status of the object with fulfillments. For example, `WC_Order::get_status()` is used for order fulfillments.
-	 *
-	 * @var string
-	 */
-	abstract public function get_fulfillment_status(): string;
-
-	/**
 	 * Get fulfillments of an entity.
 	 *
 	 * @return Array<Fulfillment>
@@ -67,7 +51,6 @@ abstract class AbstractFulfillmentManager {
 		wc_get_container()
 			->get( FulfillmentsDataStore::class )
 			->create( $fulfillment );
-		$this->update_order_fulfillment_status();
 	}
 
 	/**
@@ -81,7 +64,6 @@ abstract class AbstractFulfillmentManager {
 		wc_get_container()
 			->get( FulfillmentsDataStore::class )
 			->update( $fulfillment );
-		$this->update_order_fulfillment_status();
 	}
 
 	/**
@@ -95,7 +77,6 @@ abstract class AbstractFulfillmentManager {
 		wc_get_container()
 			->get( FulfillmentsDataStore::class )
 			->delete( $fulfillment );
-		$this->update_order_fulfillment_status();
 	}
 
 	/**
@@ -110,16 +91,13 @@ abstract class AbstractFulfillmentManager {
 		$fulfillment->set_status( $status );
 		$fulfillment->set_is_fulfilled( true );
 		$fulfillment->save();
-		$this->update_order_fulfillment_status();
 	}
 
+
 	/**
-	 * Update the fulfillment status of the entity based on the fulfillments.
-	 *
-	 * This method checks if all fulfillments are fulfilled, if some fulfillments are fulfilled, or if no fulfillments are present.
-	 * It sets the fulfillment status of the entity accordingly.
+	 * Get the fulfillment status of the entity. This acts like a computed property.
 	 */
-	private function update_order_fulfillment_status(): void {
+	public function get_fulfillment_status(): string {
 		$fulfillments     = $this->get_fulfillments();
 		$has_fulfillments = ! empty( $fulfillments );
 		$all_fulfilled    = true;
@@ -135,14 +113,14 @@ abstract class AbstractFulfillmentManager {
 			}
 
 			if ( $all_fulfilled ) {
-				$this->set_fulfillment_status( 'fulfilled' );
+				return 'fulfilled';
 			} elseif ( $some_fulfilled ) {
-				$this->set_fulfillment_status( 'partially_fulfilled' );
+				return 'partially_fulfilled';
 			} else {
-				$this->set_fulfillment_status( 'unfulfilled' );
+				return 'unfulfilled';
 			}
 		} else {
-			$this->set_fulfillment_status( 'no_fulfillments' );
+			return 'no_fulfillments';
 		}
 	}
 }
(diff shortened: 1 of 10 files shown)


prettyboymp (changes requested):
Hi @tpaksu, thank you for your work on this!

I think we should avoid introducing a trait on the Order class to manage related fulfillments. Instead, it would be better to work toward creating a separate API for this functionality.

The main concern is that once these methods become part of the Order’s public interface, we would need to support them in their current form indefinitely — not just for Order, but for any other entity that might adopt the trait in the future. This can make maintenance and evolution of the API much harder over time.

Keeping fulfillment management in a separate, dedicated API would allow the Order class to remain more focused on managing order data itself, and it would provide a more cohesive and intentional interface for handling fulfillments.

tpaksu:
Thanks for your review @prettyboymp, it makes sense to use this in a separate API instead of  adding as a trait in `WC_Order`. I thought making an order to be `fulfillable` while opening the possibility to make 3rd party entities also utilize this trait to make their objects with items also fulfillable easily will fit perfectly with the definition of a trait, and built it that way. I couldn't foresee any possible maintenance issues, like also needing to show them in StoreAPI when an order is requested, etc. I will convert it into a separate class that's isolated. 

Can you suggest me a good name for this new class?

tpaksu:
Moved on with `AbstractFulfillmentManager` to let other entities work with fulfillments, and `OrderFulfillmentManager` extending that, for `WC_Order` based fulfillments.

allie500 (approved):
Thanks for working on this Fulfillment Manager, Taha. The code looks good. `AbstractFulfillmentManager` provides two abstract public methods to be implemented by a class that extends it. In this case `OrderFulfillmentManager` handles implementation of them. 

`AbstractFulfillmentManager` class also handles implementation of public CRUD methods for creating, getting, updating and deleting fulfillments. It also contains helper methods to fulfill a fulfillment and get a fulfillment's status.

`OrderFulfillmentManagerTest` is comprehensive and covers all our base scenarios. I ran the tests on my local and they all passed.

```bash
pnpm --filter=@woocommerce/plugin-woocommerce test:unit:env -- --filter=OrderFulfillmentManagerTest  

> @woocommerce/plugin-woocommerce@9.9.0 test:unit:env /Users/allie/repos/woocommerce/woocommerce/plugins/woocommerce
> pnpm test:php:env "--" "--filter=OrderFulfillmentManagerTest"


> @woocommerce/plugin-woocommerce@9.9.0 test:php:env /Users/allie/repos/woocommerce/woocommerce/plugins/woocommerce
> wp-env run --env-cwd='wp-content/plugins/woocommerce' tests-cli vendor/bin/phpunit -c phpunit.xml --verbose "--" "--filter=OrderFulfillmentManagerTest"

ℹ Starting 'vendor/bin/phpunit -c phpunit.xml --verbose --filter=OrderFulfillmentManagerTest' on the tests-cli container. 

Installing...
Running as single site... To run multisite, use -c tests/phpunit/multisite.xml
Installing WooCommerce...
Not running ajax tests. To execute these, use --group ajax.
Not running ms-files tests. To execute these, use --group ms-files.
Not running external-http tests. To execute these, use --group external-http.
PHPUnit 9.6.20 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.4.33
Configuration: phpunit.xml

......                                                              6 / 6 (100%)

Time: 00:00.688, Memory: 143.00 MB
```

There are no screenshots or other analysis to share at this time as there is no user interface and the `OrderFulfillmentManager` is not integrated into the wider WooCommerce plugin as this time.

I believe this is good to merge, but let's wait to see what @prettyboymp thinks of the latest changes.

plugins/woocommerce/src/Internal/Fulfillments/AbstractFulfillmentManager.php
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	/**
+	 * Create a new fulfillment on the entity.
+	 *
+	 * @param Fulfillment $fulfillment Fulfillment object.
+	 *
+	 * @return void
+	 */
+	public function add_fulfillment( Fulfillment $fulfillment ): void {
```
prettyboymp, line 50: If the client already has an instance of a `Fulfillment` can't they just call `$fulfillment->save()`?

plugins/woocommerce/src/Internal/Fulfillments/AbstractFulfillmentManager.php
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	/**
+	 * Delete a fulfillment from the entity.
+	 *
+	 * @param Fulfillment $fulfillment Fulfillment object.
+	 *
+	 * @return void
+	 */
+	public function delete_fulfillment( Fulfillment $fulfillment ): void {
```
prettyboymp, line 76: Similar situation here.  The caller can just use `$fulfillment->delete()`.

prettyboymp (changes requested):
Hi @tpaksu, can you help me understand the purpose of a *FulfillmentManager?  I'm not sure I understand the problem it solves or how some of these methods would be used instead acting directly on the Fulfillment instance.

tpaksu:
Closing this PR as we can return back when it becomes a hard requirement for the feature. Currently it was thought for limiting the users access to the fulfillment data when they only have an entity that supports fulfillments, but using the data store for this purpose will be sufficient, as the plugin authors also will have access to the DB functions and filters to reach whatever they want.

PR #57557, opened 2025-04-28
Ensure login-form is rendered on in Classic Checkout for guests

### LINKED ISSUE #57513: 'Log in' link not doing anything on legacy checkout
### Prerequisites

### Describe the bug

Under certain configuration (see below), if a user tries to purchase something but their email address is already registered, they receive a banner that says "An account is already registered with your email address. [Please log in](#)". Clicking the "log in" link should open some sort of login screen but instead does nothing.

This seems to be because the [login form template](https://github.com/woocommerce/woocommerce/blob/ece5ba3ff7c381c30a73c42e634595ea72f1490d/plugins/woocommerce/templates/global/form-login.php#L4) does not appear to be rendered in the page markup, and [the legacy checkout JS does not appear to be enqueued](https://github.com/woocommerce/woocommerce/blob/ece5ba3ff7c381c30a73c42e634595ea72f1490d/plugins/woocommerce/client/legacy/js/frontend/checkout.js#L844-L852).

### Expected behavior

I expect the login form template to appear somewhere when the user clicks the "Please log in" link, or a redirect to the My Account login screen, or the link to not appear (which is how block-based checkout handles it)

Image

### Actual behavior

Nothing happens when the link is clicked.

### Steps to reproduce

1. You'll want to use legacy checkout (either the shortcode or the legacy checkout block) and these account settings (guest checkout and log in during checkout disabled):
Image

2. Attempt to purchase a product with an email address that is already registered on the site. Observe you get this banner and clicking the link does nothing:
Image

### WordPress Environment

I was able to reproduce on: 
- the latest WP version
- twentytwentyfive theme
- latest WooCommerce version 
- latest WooCommerce Stripe Gateway version 
- **no other plugins active**.

### Isolating the problem

tjcafferkey:

For guests the login form isn't rendered in the checkout markup, so when users click the login link that appears in the validation error nothing happens.

This PR renders login template, and scrolls the user to it once its in view.

Closes [WOOPLUG-4066](https://linear.app/a8c/issue/WOOPLUG-4066/log-in-link-not-doing-anything-on-legacy-checkout)
Closes https://github.com/woocommerce/woocommerce/issues/57513

CleanShot 2025-04-28 at 11 30 44

1. Enable the Classic Checkout instead of the Block Checkout
2. Enable account creation during checkout in the settings: `/wp-admin/admin.php?page=wc-settings&tab=account`
3. Create a customer account if you haven't already got one, but once created make sure you log out.
4. Add a product to the basket and go to checkout. Before placing order tick "Create an account"
5. Ensure the correct validation appears saying: "An account is already registered with your email address. Please log in"
6. Click the "Please log in" link, make sure a login form appears above the checkout and that you get scrolled to it.
7. Login to your customer account and place your order successfully.
8. Finally: Ensure CI passes which includes a PHP test that checks the form is present for guest accounts.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


mikejolley (changes requested):
There are some issues here, and some testing instructions that could use more detail, namely on the settings required to cause the issue to occur. To reproduce the issue you need to turn off `Enable log-in during checkout`.

With `Enable log-in during checkout` on, the bug is not present. However, this PR introduces a regression because then you get two forms:

Screenshot 2025-04-28 at 14 26 18

I'm wondering about the approach here. By rendering the form to fix the link, we're going against the setting `Enable log-in during checkout`. So my question would be which approach is better:

1. Forcing the form to render even if `Enable log-in during checkout` is off (this PR)
2. Removing the link or updating the link to go to the WP Login page/Account page

Related, do we even want to leave the options like this allowing the merchant to break the experience? Is `Enable log-in during checkout` option actually beneficial in existing?

plugins/woocommerce/client/legacy/js/frontend/checkout.js
```diff
@@ -846,7 +846,21 @@ jQuery( function( $ ) {
 			$( document.body ).on( 'click', 'a.showlogin', this.show_login_form );
 		},
 		show_login_form: function() {
-			$( 'form.login, form.woocommerce-form--login' ).slideToggle();
+			var $form = $( 'form.login, form.woocommerce-form--login' );
```
tjcafferkey, line 1126: The setting `woocommerce_enable_checkout_login_reminder` renders the form under the notice so its always presumed to be in view.

However with the `woocommerce_checkout_registration_enabled` setting the form isn't below the notice since we get the notice error back as a response from the server so I scroll it into view.

tjcafferkey: I _could_:

1. Look into rendering the one form for both scenarios under all the notices
2. Refactor so this form can be rendered separately from the notice. Maintaining the existing position for the login reminder, whilst also rendering it under the registration error separately.

Both options feel like possible improvements if warranted but this feels like it fixes the issue with minimal changes.

tjcafferkey:
Thanks @mikejolley I wasn't aware of the other setting that triggered the additional notice. I've consolidated the logic for both in the same file but mentioned some tradeoffs.

mikejolley (changes requested):
The duplication is fixed now ✅ So the main issue is fixed. The scroll works too.

I do have to be annoying and point out some usability issues however. If you think these are all better suited for a followup I'm happy to approve, but (1) is directly related to the change.

1. I noticed an inconsistency with the coupon form. `Click here to enter your code` toggles the form open and closed. `Click here to login` toggles it open and scrolls it into view. I can see that having a "click here to _show_" link doens't really make full sense to toggle, but seeing as the forms are in such close proximity I feel consistency is important. Ideally both would open the form and there would be some other way to close them (X).

2. If you try to login and get the password wrong, its not possible to reopen the form without first placing the order again.

3. If you login successfully (yay!) your form is cleared (boo!). I realise this one is probably out of scope but it could work better if the logic was part of the checkout form submission, if thats even possible.

Let us know what you think. We can solve the prettier formatting of the JS file before merge as well.

tjcafferkey:
@mikejolley I've rerequested review after fixing feedback points 1 & 2. I do feel like 3 is out of scope too, but worth following up with in another PR at some point.

plugins/woocommerce/templates/checkout/form-login.php
```diff
@@ -12,25 +12,44 @@
  *
  * @see https://woocommerce.com/document/template-structure/
  * @package WooCommerce\Templates
- * @version 3.8.0
+ * @version 9.9.0
```
mikejolley: This will need a higher bump because 9.9 was already tagged.

plugins/woocommerce/templates/checkout/form-login.php
```diff
@@ -12,25 +12,44 @@
...
+	<?php
+endif;
+
+if ( $registration_at_checkout || $login_reminder_at_checkout ) :
+
+	$hide_form = true;
+	// Always show the form after a login attempt.
+	if ( isset( $_SERVER['REQUEST_METHOD'] ) && 'POST' === $_SERVER['REQUEST_METHOD'] && isset( $_POST['login'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Missing
+		$hide_form = false;
+	}
```
mikejolley: I think you can simplify this quite a bit i.e. 

```suggestion
	// Always show the form after a login attempt.
	$hide_form = ! isset( $_POST['login'] ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
```

I'm not sure the other logic adds much value here. Might be more readable if you flip the logic too (_show_form_).

plugins/woocommerce/client/legacy/js/frontend/checkout.js
```diff
@@ -846,7 +846,28 @@ jQuery( function( $ ) {
...
+			if ($form.is(':visible')) {
```
mikejolley: I know it will increase diff but don't forget to run prettier on this file to fix the spacing issues.

plugins/woocommerce/templates/checkout/form-login.php
```diff
@@ -12,25 +12,44 @@
...
  * @see https://woocommerce.com/document/template-structure/
  * @package WooCommerce\Templates
- * @version 3.8.0
+ * @version 9.9.0
  */
 
 defined( 'ABSPATH' ) || exit;
 
-if ( is_user_logged_in() || 'no' === get_option( 'woocommerce_enable_checkout_login_reminder' ) ) {
+$registration_at_checkout   = WC_Checkout::instance()->is_registration_enabled();
```
mikejolley, line 20: I do prefer this, but this template file is provided `$checkout` in `woocommerce_checkout_login_form`. Just making a note, no need to change it.

mikejolley (approved):
Thanks @tjcafferkey that seems to be working better. I've left some non-blocking feedback and will pre-approve.

mikejolley:
Actually the template bump is blocking but the rest are not!

PR #57566, opened 2025-04-28
Expose an action to block checkout in checkout data store

### LINKED ISSUE #57509: Expose an action to block checkout in checkout data store
There's already a very explicit (and potentially error-prone) way to achieve that, by using actions that allow to increment or decrement pending "calculations" within the checkout. Example:

```
import { useDispatch } from '@wordpress/data';
import { checkoutStore } from '@woocommerce/block-data';

const { __internalIncrementCalculating, __internalDecrementCalculating } = useDispatch( checkoutStore );

__internalIncrementCalculating();

await someFunctionalityThatRequireCheckoutToBeBlocked();

__internalDecrementCalculating();
```

We would like to mitigate a potential risk of using increment without accompanying decrement action, which can result in a critical problem for the checkout flow.

Our initial idea is to provide an action that would accept an async function (or put differently—a function that returns a promise). We would manage increment/decrement on our side based on the promise result (whether it was resolved or rejected), we could also add a safeguard in form of a timeout if the process takes too long.

Pseudocode to illustrate the idea (provided by @thomas.roberts):

```
// Core function provided to extension developers
function blockCheckoutForOperation( asyncOperation ) {
  // Signal checkout is blocked
  incrementCalculating();
  
  return asyncOperation()
    .then(result => {
      // Unblock checkout on success
      decrementCalculating();
      return result;
    })
    .catch(error => {
      // Unblock checkout on error
      decrementCalculating();
      throw error; // Rethrow to allow extension to handle
    });
}

// Usage example for extension developers
blockCheckoutForOperation(async () => {
  // Third-party API call or any async operation
  const response = await thirdPartyApiCall();
  return response.data;
});
```

straku:

Exposing a new action (it's really a thunk) called `disableCheckoutFor` that allows us and plugin developers to disable the proceed to checkout button for the duration of some async operation. This is a replacement for usage of `__internalIncrementCalculating` and `__internalDecrementCalculating` by 3rd party developers, providing a less error-prone API that works seamlessly with promises/async functions.

For the name, I got inspired by the `waitFor` API from the testing library, but I am open to different suggestions.

Closes #57509

We're going to simulate 3rd party usage by running some console scripts for testing purposes.

1. Checkout this branch in your local environment and open a page with the blocks cart.
2. Open the console in your devtools of choice and paste the following script:

```js
let wait = ( timeout ) => new Promise( ( resolve ) => setTimeout( resolve, timeout ) );
wp.data.dispatch(wc.wcBlocksData.checkoutStore).disableCheckoutFor(async () => {
    await wait(10_000); // you can pick a different timeout
});
```

3. After the script is pasted, hover over the proceed to checkout button and verify that it is disabled for the amount of time specified in the script. I think it is safe to rely on manual counting, but if you want to whip out your stopwatch—go for it!
4. Let's refresh the site and check whether our new method handles errors correctly. Paste the following script:

```js
let wait = ( timeout = 0 ) => new Promise( ( resolve ) => setTimeout( resolve, timeout ) );
wp.data.dispatch(wc.wcBlocksData.checkoutStore).disableCheckoutFor(async () => {
    await wait(5_000); // you can pick a different timeout
    throw new Error('testing error handling');
});
```

5. After the script is pasted, hover over the proceed to checkout button and verify that it is disabled for the amount of time specified in the script. After the timeout has passed, verify that the error was logged to the console as well.

Above testing steps + modifying a block to perform similar set of steps locally +  integration tests in JavaScript test suite present.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>
<details>

#### Comment 

</details>


ralucaStan:
@straku I find it interesting offering an action to disable checkout. I'm curious what are the use case behind it. 

I've been having some discussions about the `isCalculating` value, which is currently used to block placing orders or proceeding to checkout.  pca54o-9mS-p2#comment-7717 (this link might not work on Linear).

This is important for your PR because we were discussing removing the mechanism based on `__internalIncrementCalculating` and `__internalDecrementCalculating`  for getting the `isCalculating` value. We'd prefer a more deterministic approach to avoid the vagueness that a single boolean value brings around what is blocking checkout. More details in my link, but essentially we want to keep track of all the states/flags that need to block the checkout, and use those to block interactions or trigger loading states.

`disableCheckoutFor` could still be one of the things blocking checkout:
```js
 isCalculating = checkoutDiabledFor || isLoadingRates || etc
```
 but moving forward, we'd be removing `__internalStartCalculation` from other actions that block checkout; I don't see a problem with renaming it and keeping it only for `disableCheckoutFor`.

Like I've said, I'm curious about the use cases. Let's keep in touch.

straku:
@ralucaStan Thank you for making me aware of this discussion. The reasoning behind this change is not to affect the current internal mechanism of tracking pending operations, but to stop users from relying on the current mechanism that can be error-prone. It stemmed from [this discussion](https://github.com/woocommerce/woocommerce-shipstation/pull/291#issuecomment-2815188710).

When looking at it from this perspective, it is a step forward towards the potential changes that you would like to introduce. `disableCheckoutFor` could be an external API available to plugin developers, whereas the internal actions could be adjusted to our new needs.

It might be worthwhile though to think a lit more about the proposed internal actions: `__internalStartCalculation` and `__internalFinishCalculation` feels like a better abstraction than increment/decrement, but maybe we should allow it to take an argument (even if it wouldn't do anything just yet) specifying the type of calculation/operation that we're tracking (e.g. "applying-coupon", "shipping-rate-being-selected", etc.).

I am open to other suggestions as well!

ralucaStan:
Thank you so much for the context!

> disableCheckoutFor could be an external API available to plugin developers, whereas the internal actions could be adjusted to our new needs.

100%.

> It might be worthwhile though to think a lit more about the proposed internal actions: __internalStartCalculation and __internalFinishCalculation feels like a better abstraction than increment/decrement, but maybe we should allow it to take an argument (even if it wouldn't do anything just yet) specifying the type of calculation/operation that we're tracking (e.g. "applying-coupon", "shipping-rate-being-selected", etc.).

I like the abstractisation you introduce, and we're most likely going to keep if for the `disableCheckoutFor` action.

With the current status quo, my concerns related to this PR are that:
- we're using the same mechanism as the one for blocking actions we control
- we're going to remove start/finish calculations for blocking actions we control

My suggestion would be to not replace/reuse the current internal functions, but create new dedicated ones just for your use case. This wouldn't need a deprecation of the current internal methods and it would enable us to easily remove the legacy calculation function.

I'm also thinking: would it be helpful for 3PD to provide their extension name when blocking the checkout? I think that would be really helpful for debugging purposes.

straku:
> This wouldn't need a deprecation of the current internal methods and it would enable us to easily remove the legacy calculation function.

The deprecation of the current internal methods is not here because functionality introduced in this PR requires it. It's here because we've shared this internal methods as a potential solution to similar problems with 3PD (@opr I got this impression from the initial discussion, if I am wrong I would appreciate a correction).

@ralucaStan If you're sure we don't need to worry about removing those internal methods then we can remove deprecations that we've introduced here and rely on existing internal actions. This way any move to a more granular pending operations  tracking (as discussed in the shared post) would have to be accounted for in `disableCheckoutFor`, but not blocked by it's presence in any way.

opr:
> The deprecation of the current internal methods is not here because functionality introduced in this PR requires it. It's here because we've shared this internal methods as a potential solution to similar problems with 3PD (@opr I got this impression from the initial discussion, if I am wrong I would appreciate a correction).

Correct! The fact that if 3pd code errors and doesn't decrement the counter correctly, the checkout is stuck in a state where the shopper cannot do anything to fix it.

> @ralucaStan If you're sure we don't need to worry about removing those internal methods then we can remove deprecations that we've introduced here and rely on existing internal actions. This way any move to a more granular pending operations tracking (as discussed in the shared post) would have to be accounted for in `disableCheckoutFor`, but not blocked by it's presence in any way.

From what I understand we are adding the deprecation messages for the benefit of 3pds who are using them, and letting them know there is a better option. Raluca I don't follow your explanation of why leaving them in place without the deprecation message is OK?

plugins/woocommerce/client/blocks/assets/js/data/checkout/actions.ts
```diff
@@ -71,21 +71,49 @@ export const __internalSetHasError = ( hasError = true ) => ( {
...
+ * shipping, etc. Increases the `calculatingCount` which tracks ongoing
+ * calculations. A `calculatingCount` of 0 means nothing is being updated.
  */
-export const __internalDecrementCalculating = () => ( {
+export const __internalFinishCalculation = () => ( {
 	type: types.DECREMENT_CALCULATING,
 } );
 
+/**
+ * @deprecated Use disableCheckoutFor thunk instead
```
tjcafferkey, line 92: I think its common practice to also include the WC version we deprecated this function too:

`@deprecated since 9.9.0: Use disableCheckoutFor thunk instead` for example.

plugins/woocommerce/client/blocks/assets/js/data/checkout/actions.ts
```diff
@@ -71,21 +71,49 @@ export const __internalSetHasError = ( hasError = true ) => ( {
...
+		alternative: 'disableCheckoutFor',
+		plugin: 'WooCommerce',
+	} );
+	return {
+		type: types.INCREMENT_CALCULATING,
+	};
+};
+
+/**
+ * @deprecated Use disableCheckoutFor thunk instead
```
tjcafferkey, line 106: Same as https://github.com/woocommerce/woocommerce/pull/57566/files#r2069998256

plugins/woocommerce/client/blocks/docs/internal-developers/block-client-apis/checkout/checkout-api.md
```diff
@@ -65,14 +65,22 @@ The following actions can be dispatched from the Checkout data store:
...
--   `__internalDecrementCalculating()`: Decrement `state.calculatingCount`
+-   `__internalStartCalculation`: Increment `state.calculatingCount`
+-   `__internalFinishCalculation`: Decrement `state.calculatingCount`
 -   `__internalSetCustomerId( id: number )`: Set `state.customerId` to `id`
 -   `__internalSetUseShippingAsBilling( useShippingAsBilling: boolean )`: Set `state.useShippingAsBilling` to `useShippingAsBilling`
 -   `__internalSetShouldCreateAccount( shouldCreateAccount: boolean )`: Set `state.shouldCreateAccount` to `shouldCreateAccount`
 -   `__internalSetOrderNotes( orderNotes: string )`: Set `state.orderNotes` to `orderNotes`
 -   `setExtensionData( namespace: string, extensionData: Record< string, unknown > )`: Set `state.extensionData` to `extensionData`
 
+##### Deprecated actions
```
tjcafferkey, line 77: Same as https://github.com/woocommerce/woocommerce/pull/57566/files#r2069998256

Lets add the version we deprecated these into the docs too.

tomando97:
Tu cres q vale la pena todo esto

ralucaStan:
> The deprecation of the current internal methods is not here because functionality introduced in this PR requires it. It's here because we've shared this internal methods as a potential solution to similar problems with 3PD (@opr I got this impression from the initial discussion, if I am wrong I would appreciate a correction).

> From what I understand we are adding the deprecation messages for the benefit of 3pds who are using them, and letting them know there is a better option. Raluca I don't follow your explanation of why leaving them in place without the deprecation message is OK?

I didn't have this understanding when I added my comments. Makes sense. Eventually, the `__internalIncrementCalculating` and `__internalDecrementCalculating` will be deprecated when we remove them from under the `isCalculating` flag. I was worried about the duplicated purpose and thought we might want a different API if we also want 3PD to provide a namespace when they block the checkout. 

@straku @opr we should include some docs about the new `disableCheckoutFor`  and explain its purpose + provide usage examples (before/after examples)

opr:
> we should include some docs about the new disableCheckoutFor and explain its purpose + provide usage examples (before/after examples)

💯 agree, we should ensure examples are included.

plugins/woocommerce/client/blocks/assets/js/data/checkout/actions.ts
straku: Addressed in https://github.com/woocommerce/woocommerce/pull/57566/commits/a211f2995a9328bfecfa61339536cc925e3be2df, I also did the same to `__internalSetExtensionData` after verifying that it targets `9.9.0` too.

plugins/woocommerce/client/blocks/assets/js/data/checkout/actions.ts
straku: https://github.com/woocommerce/woocommerce/pull/57566#discussion_r2073659519

plugins/woocommerce/client/blocks/docs/internal-developers/block-client-apis/checkout/checkout-api.md
straku: https://github.com/woocommerce/woocommerce/pull/57566#discussion_r2073659519

straku:
I've added a version to deprecation notices and updated the docs. FYI @tjcafferkey @opr @ralucaStan

plugins/woocommerce/client/blocks/docs/internal-developers/block-client-apis/checkout/checkout-api.md
```diff
@@ -65,13 +65,24 @@ The following actions can be dispatched from the Checkout data store:
...
--   `__internalIncrementCalculating()`: Increment `state.calculatingCount`
--   `__internalDecrementCalculating()`: Decrement `state.calculatingCount`
+-   `__internalStartCalculation`: Increment `state.calculatingCount`
+-   `__internalFinishCalculation`: Decrement `state.calculatingCount`
 -   `__internalSetCustomerId( id: number )`: Set `state.customerId` to `id`
 -   `__internalSetUseShippingAsBilling( useShippingAsBilling: boolean )`: Set `state.useShippingAsBilling` to `useShippingAsBilling`
 -   `__internalSetShouldCreateAccount( shouldCreateAccount: boolean )`: Set `state.shouldCreateAccount` to `shouldCreateAccount`
 -   `__internalSetOrderNotes( orderNotes: string )`: Set `state.orderNotes` to `orderNotes`
 -   `setExtensionData( namespace: string, extensionData: Record< string, unknown > )`: Set `state.extensionData` to `extensionData`
+-   `disableCheckoutFor( asyncFunc: () => Promise< unknown > )`: A thunk that allows to disable checkout button until promise returned by `asyncFunc` gets resolved. Relies on the same internal state as the one powering the tracking of pending calculations.
```
ralucaStan, line 75: > Relies on the same internal state as the one powering the tracking of pending calculations.

Not sure if we should mention this, as we're going to remove the state for pending calculations in the future.

straku: I didn't mention it in the developer docs (top level `/docs` directory), but since these are more targeted at internal developers and discuss internals in detail, then I figured it is a worthy addition. If we decide to remove the state tracking pending calculations these docs will need an overhaul.

If you have other suggestions, I am happy to apply them.

docs/cart-and-checkout-blocks/checkout-payment-methods/checkout-flow-and-events.md
```diff
@@ -69,6 +69,22 @@ The following are booleans exposed via the checkout provider that are independen
 
 **isCalculating:** This is true when the total is being re-calculated for the order. There are numerous things that might trigger a recalculation of the total: coupons being added or removed, shipping rates updated, shipping rate selected etc. This flag consolidates all activity that might be occurring (including requests to the server that potentially affect calculation of totals). So instead of having to check each of those individual states you can reliably just check if this boolean is true (calculating) or false (not calculating).
```
ralucaStan: > **isCalculating:** This is true when the total is being re-calculated for the order. [...]

I'd update this part to say: This is true when the total is being re-calculated for the order **or when a plugin is intentionally disabling the Proceed to checkout using the `disableCheckoutFor` action**.

and then have 2 subsections
- Total recalculations 
- Blocking checkout with `disableCheckoutFor`

docs/cart-and-checkout-blocks/checkout-payment-methods/checkout-flow-and-events.md
```diff
@@ -69,6 +69,22 @@ The following are booleans exposed via the checkout provider that are independen
...
 
+You can also programmatically control this state using the `disableCheckoutFor` thunk. This allows you to disable the checkout button while an asynchronous operation completes:
```
ralucaStan: Also, worth noting that `isCalculating` isn't only blocking the Place order button.
It also blocks the Proceed to Checkout button from the cart block.
It's also showing a loading state for the Express Payments methods

Maybe we can specify before this section what `isCalculating` currently impacts.



```
What `isCalculating` affects:
	•	Disables the “Place Order” button in the checkout block
	•	Disables the “Proceed to Checkout” button in the cart block
	•	Shows a loading state for Express Payment methods while calculations are pending
```

And here we can only mention 
> You can also programmatically control `isCalculating` using the `disableCheckoutFor` thunk.

ralucaStan (changes requested):
Great PR! It tests well for me. I stress tested this with multiple Promises and it works as expected.

I suggested some docs improvements, which also include details not relevant to this PR, just to make docs clear about the impact of isCalculating.

docs/cart-and-checkout-blocks/checkout-payment-methods/checkout-flow-and-events.md
straku: I've pushed updates to the docs. Can you take a look?

docs/cart-and-checkout-blocks/checkout-payment-methods/checkout-flow-and-events.md
straku: This should be hopefully addressed with the latest changes to the docs.

docs/cart-and-checkout-blocks/checkout-payment-methods/checkout-flow-and-events.md
ralucaStan: Looks good. Thank you for improving this section. I'll go ahead and approve.

ralucaStan (approved):

opr (approved):
Thanks @straku this is looking good now! Not sure what the next steps are with docs, but I think you're in comms with Pi? I will approve as it's working well according to my testing. I also tested with multiple promises, some of which threw errors, others didn't. This still blocked checkout until the last function was complete, and then unblocked.

@straku you will need to update the unit tests to expect the updated deprecation messages as a result of adding the version.

straku:
@ralucaStan @opr Removed all changes to the top level `docs` directory and moved them to https://github.com/woocommerce/woocommerce/pull/57850 targeting temporary branch with docusaurus.

PR #57567, opened 2025-04-28
Woo AI: Add a deprecation notice for users using the feature

emdashcodes:
See pdtR3Z-4bB-p2 #comment-4419
See `AI-133/sunset-woo-ai`
See https://github.com/woocommerce/woocommerce/pull/57075

> I think showing a banner to users who have used Woo AI features explaining the deprecation would be enough.

This PR adds a deprecation banner to the Woo AI plugin if the user interacts with any of the features. This is to help us display a deprecation notice for WooCommerce on Atomic users for a ~2 week period, but avoid showing the notice to everyone else since the usage of this feature is lower.

This PR and the whole plugin will later be removed in https://github.com/woocommerce/woocommerce/pull/57075.

I'll work with Happiness for the actual message we want here.

https://github.com/user-attachments/assets/1b257fb9-a8b4-4f19-bbfb-50bdd63d992e

https://github.com/user-attachments/assets/e468efbd-37c0-474e-bc88-023cda57bb1e

* If you tested the PR previously, go to `https://{SITE_URL}/wp-admin/post-new.php?post_type=product&reset_woo_ai_notice=true` to reset the option
* Open the network console and filter by admin-ajax
* Go to the product editor and use one of the features: Title generation, category generation, background removal, or description generation
* See that a notice about the feature being deprecated shows up
* Click the dismiss button, and verify that an admin-ajax request is sent
* Interacting with another feature after dismissing should not result in another notice

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

Adds deprecation notice for Woo AI interactions

</details>

<details>

#### Comment 

</details>


gigitux:
Thanks for opening this PR! I tested it and works as expected:


https://github.com/user-attachments/assets/cbd38bfb-3fba-4bd9-a567-be3d70d82e9b

I have two questions regarding UI/UX.
- Should we remove the snack bar about leaving feedback? 
- Also, isn't it better if we render the notice every time that a Woo AI feature is used?

peterfabian:
> isn't it better if we render the notice every time that a Woo AI feature is used?

+1 on this one. I think it's too easy to miss it after hitting publish, as the whole screen gets reloaded, and merchants already expect a notice at the top appearing at that point (saying the product was published).

emdashcodes:
@gigitux @peterfabian Thank you for looking!

I pushed some changes:

Updated the UX to show the notice after interacting with the feature. Notices can still be permanently dismissed, so we don't spam users with them.

Removed the feedback snackbar from the features that were using it.

Moved to the notice handling to the react app.

--

Testing instructions have been updated in the main PR.

## Demo

https://github.com/user-attachments/assets/313846bd-2866-4633-924e-d7d049e12a2a

plugins/woo-ai/includes/class-woo-ai-deprecation-notice.php
```diff
@@ -0,0 +1,88 @@
...
+		// Register AJAX actions
+		add_action( 'wp_ajax_woo_ai_dismiss_deprecation', array( $this, 'dismiss_notice' ) );
+
+		// Check for reset query parameter for testing purposes
+		// TODO: Remove this before merging
+		if ( isset( $_GET['reset_woo_ai_notice'] ) && current_user_can( 'manage_options' ) ) {
+			delete_option( 'woo_ai_deprecation_dismissed' );
+			$current_url = remove_query_arg( 'reset_woo_ai_notice', $_SERVER['REQUEST_URI'] );
+			wp_safe_redirect( $current_url );
+			exit;
```
gigitux: Should we remove this code, right?

gigitux (commented):
Thanks for updating this PR. I tested the PR and it LGTM: 

https://github.com/user-attachments/assets/55adb79b-27b0-42ef-b2b4-6e5a3cbdd115

> Updated the UX to show the notice after interacting with the feature. Notices can still be permanently dismissed, so we don't spam users with them.

I'm not a UX expert, but I believe the notice should still be shown even if the user has previously dismissed it.  
I'd suggest handling the `isDismissed` logic on the client side, so the notice disappears after dismissal but reappears upon page refresh.

This approach also helps us avoid storing data in the database that will no longer be used after removing the Woo AI plugin from atomic env.

plugins/woo-ai/includes/class-woo-ai-deprecation-notice.php
emdashcodes: Removed! I was leaving it there temporarily for testing.

emdashcodes:
> I'm not a UX expert, but I believe the notice should still be shown even if the user has previously dismissed it.
I'd suggest handling the isDismissed logic on the client side, so the notice disappears after dismissal but reappears upon page refresh.

> This approach also helps us avoid storing data in the database that will no longer be used after removing the Woo AI plugin from atomic env.

Fair. I was trying to avoid spamming the user with notices, but considering the usage of these features is pretty low anyway, I think that makes sense. 

I removed the server side logic from this.

gigitux (approved):
Thanks for addressing the feedback! 🙇 Everything works as expected!


https://github.com/user-attachments/assets/9276aaa6-feaa-4c66-bd8f-c428e26597de

PR #57586, opened 2025-04-29
Update Product Image block to support inner blocks

### LINKED ISSUE #55996: Spike: Blockify the Product Image Block's On-Sale Badge

sunyatasattva:

Implements inner blocks support in the Product Image block, allowing for more flexible layouts.

Refactors rendering logic to accommodate new inner blocks structure.

Backwards-compatibility is ensured by providing a deprecation mechanism and making sure the “On Sale Badge” is still displayed on the front-end if the block includes the legacy attributes and has not been migrated yet.

Closes #55996.

| Before | After |
| ------ | ----- |
|  Screenshot 2025-05-07 at 19 24 19 <br/> Screenshot 2025-05-07 at 19 24 07 |  Screenshot 2025-05-07 at 19 14 20 |

#### 1. Using the block from scratch inside the “Product Collection” 

1. Add the “Product Collection” block in any page.
2. Select the “Product Image” inside the “Product Template” block and add the “On-Sale Badge” block.
3. Style the inner block at will, including trying to justify the block using the layout controls on the “Product Image” inspector controls and toolbar to make sure it adheres to them both in the editor and in the front-end.

#### 2. Using the block from scratch inside the “Single Product” 

1. Add the “Single Product” block in any page.
2. Select a product on sale.
3. Select the “Product Image” inside the block and add the “On-Sale Badge” block.
4. Style the inner block at will, including trying to justify the block using the layout controls on the “Product Image” inspector controls and toolbar to make sure it adheres to them both in the editor and in the front-end.

#### 3. Using other inner blocks

1. Repeat instructions above, but try adding other inner blocks, such as Paragraphs.
2. Check that they work both in the editor and in the front-end.

#### 4. Migrate an old version of the block

1. Prepare a page with a version of the block before this PR.
2. Make sure the “Show on sale badge” inspector control is enabled.
3. Play around with the sale badge alignment settings.
4. Check out this branch.
5. Go to the front-end, make sure everything still works seamlessly.
6. Go to the editor, make sure you can migrate the block to the new settings. If the original block had the “Show on sale badge” setting enabled, the “On-Sale badge” should be available automatically as an inner block.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

Enhanced the Product Image block to act as a container for inner blocks, allowing for more flexible layouts.

</details>


plugins/woocommerce/src/Blocks/BlockTypes/ProductImage.php
```diff
@@ -219,19 +233,21 @@ protected function render( $attributes, $content, $block ) {
 		);
 
 		if ( $product ) {
+			$inner_content = $this->render_anchor(
+				$product,
+				$this->render_on_sale_badge( $product, $parsed_attributes ),
```
kmanijak, line 226: While I can add the On Sale Badge in the Editor, it doesn't render on the frontend for some reason 🤔 (tried in Single Product block)

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.json
```diff
@@ -5,7 +5,7 @@
 	"category": "woocommerce-product-elements",
 	"attributes": {
 		"showProductLink": { "type": "boolean", "default": true },
-		"showSaleBadge": { "type": "boolean", "default": true },
+		"showSaleBadge": { "type": "boolean", "default": false },
```
kmanijak: I think I'd consider removing the attribute completely and use `migrate` function from [Deprecation API](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-deprecation/) to map the flag to the On Sale Badge inner blog, of `true`.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.json
```diff
@@ -24,6 +24,7 @@
 			"clientNavigation": true
 		},
 		"html": false,
+		"layout": true,
```
kmanijak: AFAIK `layout: true` is not enough to display [layout controls](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-supports/#layout).

You should provide `default` in order to set the layout `flex`. That could simplify `style.scss` changes and manual `atteibutes['layout']` mapping as it would be handled natively by Core.

kmanijak (commented):
I left some comments regarding the code but the good thing is I was able to insert On Sale Badge in Editor! 🚀 I'll continue checking the approach on this and another PR tomorrow! 💪

plugins/woocommerce/src/Blocks/BlockTypes/ProductImage.php
sunyatasattva: Interesting, thanks for testing. I'll check it out. It renders on the Product Catalogue → Product Collection for me. Can you maybe verify that it renders there for you as well?

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.json
sunyatasattva: Yes, sorry, this PR was not really ready for review 😅 , using the Deprecation API was the next step.

sunyatasattva:
@kmanijak I set the PR as Ready for Review. I still have some doubts about `supports: layout`, however. I seemed to need to implement a bunch of manual CSS and even manual classes in PHP. I thought that things like `StyleAttributesUtils::get_classes_and_styles_by_attributes()` and `get_block_wrapper_attributes()` would help generate those, but they don't, as it seems, though perhaps I am missing something.

Also, I had some trouble making the justification work in the editor and the front-end. It now should work correctly, but I think there are unnecessary hacks, though I can't find what.

Please note that I also fixed some other problems, such as the fact that `isDescendentOfSingleProductBlock` was actually never implemented correctly, hence why you didn't see the badge in the Single Product Block. It should work now.

dinhtungdu:
>  I still have some doubts about supports: layout, however. I seemed to need to implement a bunch of manual CSS and even manual classes in PHP. I thought that things like StyleAttributesUtils::get_classes_and_styles_by_attributes() and get_block_wrapper_attributes() would help generate those, but they don't, as it seems, though perhaps I am missing something.

Looks like the layout style is generated as expected, but it's only for the wrapper element:
image

sunyatasattva:
We had a review with @kmanijak and we agreed that `supports: layout` is not required, and most of the use-cases can be covered with just inner blocks alignment. More complex use-cases can be done by using layout inner blocks, if required.

I'd say this is flexible enough.

Screenshot 2025-05-08 at 18 33 04

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.json
```diff
@@ -33,6 +31,7 @@
 			"__experimentalSkipSerialization": true
 		},
 		"spacing": {
+			"blockGap": true,
```
kmanijak: I don't think that does anything. Is it leftover or what's the use case for it?

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/deprecated.tsx
```diff
@@ -0,0 +1,68 @@
...
+import clsx from 'clsx';
+import { createBlock } from '@wordpress/blocks';
+
+/**
+ * Internal dependencies
+ */
+import metadata from './block.json';
+import { BlockAttributesV1 } from './types';
+
+// In v2, we removed the `showSaleBadge` attribute and converted it to an inner block.
```
kmanijak: Nitpick
```suggestion
// In v2, we removed the `showSaleBadge` and `saleBadgeAlign` attributes and converted badge to an inner block.
```

kmanijak (commented):
I realised we should also include On Sale badge by default in multiple places:
- [Product Collection](https://github.com/woocommerce/woocommerce/blob/ad38a7cbb3bc39a61906d6f37ee0a954a5b1ab8c/plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/constants.ts#L126)
- [Product Query - even though it's hidden from inserter, let's do it for clarity](https://github.com/woocommerce/woocommerce/blob/ad38a7cbb3bc39a61906d6f37ee0a954a5b1ab8c/plugins/woocommerce/client/blocks/assets/js/blocks/product-query/constants.ts#L96)
- [Related Products - as above](https://github.com/woocommerce/woocommerce/blob/d3f09fff5d363debe7041ad4f3419d6d6c74eaad/plugins/woocommerce/client/blocks/assets/js/blocks/product-query/variations/related-products.tsx#L73)
- in patterns (`plugins/woocommerce/patterns/*`)
- and most important in blockified templates (`plugins/woocommerce/templates/templates/blockified/*`)

Except for it, I realised:
1. We cannot rely on `attributes.isDescendantOfAllProducts` as it doesn’t exist. I was mistaken, it’s a `prop` passed to Block not `attribute` so it’s not available on save or edit.
2. `isEligible` logic doesn’t work in All Products and migration doesn’t fire. This is due to `attributes` propagation in that block (which I don’t understand 100%) but we receive block in a state it’s defined in [this const](https://github.com/woocommerce/woocommerce/blob/d4bbb38d0a8c4d6e329a597a692bdc26324e6c45/plugins/woocommerce/client/blocks/assets/js/blocks/products/base-utils.js#L16), basically with single attribute imageSizing. I forgot but I actually run the migration for summary if… [attributes is empty](https://github.com/woocommerce/woocommerce/blob/d4bbb38d0a8c4d6e329a597a692bdc26324e6c45/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/summary/index.ts#L31)! Because that’s what distinguishes All Products from other blocks :see_no_evil:
3. All Products is client-side rendered meaning there’s no saved markup in DB and basically visiting frontend means running `block.jsx` without running `deprecated`. In other words, when we apply changes to `block.jsx` it’s directly served to the frontend in All products after update. So the “old” behaviour has to be mimicked anyway for backwards compatibility. I [even described that to myself when working on Summary block](https://github.com/woocommerce/woocommerce/blob/d4bbb38d0a8c4d6e329a597a692bdc26324e6c45/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/summary/block.tsx#L70C1-L77C30).

Given the above I'm working on a PR branched out from this branch to cover the All Products backwards compatibility.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.json
sunyatasattva: This is a remnant from when I enabled the layout. Good catch, needs to go.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/deprecated.tsx
sunyatasattva: Considering we might want to still provide these attributes, I'm gonna hold off committing this suggestion.

sunyatasattva:
@kmanijak merged #57889 and resolved merge conflicts caused by [this PR](https://github.com/woocommerce/woocommerce/pull/57914/). What's missing from merging this?

kmanijak (commented):
I tested the changes, especially migration when switching from `trunk` to this branch:

- ✅ All Products
    - in pristine state
    - On Sale Badge explicitly on
    -  On Sale Badge explicitly off
- ✅ Product Collection (I consider it a representatifve case for other blocks like Single Products)
    - in pristine state
    - On Sale Badge explicitly on
    -  On Sale Badge explicitly off 

❌ However, there's a problem when inserting a NEW Product Collection block on this branch:
1. Create a post
2. Insert PC block
3. Change the On Sale Badge layout to left
4. Save and go to frontend
**Expected:** On Sale badge is on the left
**Actual:** There are two On Sale badges 

This happens as in this case migration _doesn't_ run meaning frontend gets:
- an inner block
- `showSaleBadge` is `true` as a default attribute

I must've missed this case eventually [in helper PR](https://github.com/woocommerce/woocommerce/pull/57889) focusing on migration cases.

Options:
- What we could do is adding a check in `plugins/woocommerce/src/Blocks/BlockTypes/ProductImage.php`: if On Sale badge is among inner blocks, ignore  `showSaleBadge`. However it has huge downside, if someone insert Product Collection and remove On Sale badge block... then badge will appear on the frontend based on `showSaleBadge` being `true`.

- I'm thinking more and more that we're trying to hack the system by keeping  `showSaleBadge` as default `true` and it has implications 😞 .

I'll try to think of another alternative.

kmanijak (commented):
I was following these scenarios:

Tested scenarios:

1. Migration (blocks added on `trunk` and switching to this branch):
  - All products
    - ✅  Pristine
    - ✅  Explicit "Show sale badge" ON
    - ✅  Explicit "Show sale badge" OFF
  - Product Collection
    - ✅  Pristine
    - ✅  Explicit "Show sale badge" ON
    - ⚠️  Explicit "Show sale badge" OFF
    - **Expected:** render 
  - Single Product
    - ⚠️ Pristine - adds badge on migration
    - ✅  Explicit "Show sale badge" ON
    - ⚠️  Explicit "Show sale badge" OFF - adds badge on migration
  - **Expected** for the above:
    - Preserve the previous state
    - Badge must work in frontend before entering editor
    - Badge is migrated from hardcoded to inner block upon entering the editor (except for All Products)

2. Newly added block (blocks added on this branch)
  - Product Collection
    - ✅ default template
    - ⚠️ removing the badge
    -  ✅ reinserting Product Image
  - Single Product
    - ✅ default template
    - ⚠️ removing the badge
    -  ✅ reinserting Product Image
  - Expected for the above
    - Product Image is inserted with On Sale badge as inner block
    - "Hardcoded" On Sale badge is not rendered on the frontend

Product Image didn't respect REMOVING Sale Badge. Basically, when the inner content was empty, the default On Sale Badge was applied. In other words - On Sale badge was re-added upon removal. I fixed it in https://github.com/woocommerce/woocommerce/pull/57586/commits/0899a38da7be4f81c5fa33b935b4d7d0b966147c and will make another (hopefully final 🤞 ) round of testing tomorrow.

kmanijak (approved):
I pushed one more change https://github.com/woocommerce/woocommerce/pull/57586/commits/18893744b47dbeb288ed30a57fe33d94ba4c3df8 that adds Sale badge explicitly in Single Product block. The `wasBlockJustInserted` method I added within Product Image (https://github.com/woocommerce/woocommerce/pull/57586/commits/0899a38da7be4f81c5fa33b935b4d7d0b966147c) is `false` when we're adding set of blocks like Single Product hence we need to explicitly pass inner blocks.

Final testing:

1. Migration (blocks added on `trunk` and switching to this branch):
  - All products
    - ✅  Pristine
    - ✅  Explicit "Show sale badge" ON
    - ✅  Explicit "Show sale badge" OFF
  - Product Collection
    - ✅  Pristine
    - ✅  Explicit "Show sale badge" ON
    - ✅  Explicit "Show sale badge" OFF
  - Single Product
    - ✅  Pristine
    - ✅  Explicit "Show sale badge" ON
    - ✅  Explicit "Show sale badge" OFF
  - **Expected** for the above:
    - Preserve the previous state
    - Badge must work in frontend before entering editor
    - Badge is migrated from hardcoded to inner block upon entering the editor (except for All Products)

2. Newly added block (blocks added on this branch)
  - Product Collection
    - ✅ default template
    - ✅ removing the badge
    - ✅ reinserting Product Image
  - Single Product
    - ✅ default template
    - ✅ removing the badge
    - ✅ reinserting Product Image
  - Expected for the above
    - Product Image is inserted with On Sale badge as inner block
    - "Hardcoded" On Sale badge is not rendered on the frontend

@sunyatasattva , I consider this PR ready to ship! But please feel free to make another round of testing before merging as we both contributed to the PR 🙌  Also, amazing work and thanks for productive collaboration!

🚢 🚢 🚢

PR #57590, opened 2025-04-29
Allow Product Image block to accept a direct Image Src

### LINKED ISSUE #57585: Product Image: Add optional attribute to Product Image block for direct image source
This issue proposes adding a new optional attribute, tentatively named `imageSrc`, to the `woocommerce/product-image block`. This attribute would allow developers to pass an image URL directly to the block, bypassing the default behavior of fetching the image based on the current product context.

**Motivation:**

Currently, the `woocommerce/product-image` block relies on the product context to determine which image to display. While this works well for its primary use case, it limits the block's reusability.

We want to use the Product Image block as a component within other blocks, such as the Product Gallery block. In such scenarios, the parent block would already have the specific image URLs for gallery items and should be able to pass them directly to child Product Image blocks.

sunyatasattva:
Introduces the ability for the Product Image block to display a specific image from a product's gallery, not just the featured image. This is achieved by consuming an `imageId` from the block context.

-   [x] I have followed the [WooCommerce Contributing Guidelines](https://github.com/woocommerce/woocommerce/blob/trunk/.github/CONTRIBUTING.md) and the [WordPress Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/).
-   [x] I have checked to ensure there aren't other open [Pull Requests](https://github.com/woocommerce/woocommerce/pulls) for the same update/change.
-   [x] I have reviewed my code for [security best practices](https://developer.wordpress.org/apis/security/).

This PR enhances the Product Image block to allow it to display a specific image from a product's gallery, rather than just the featured image. This significantly increases the reusability and flexibility of the block in creating diverse product layouts.

This is achieved by allowing the block to consume an `imageId` from the block context, which can be provided by parent blocks like 'Single Product'.

- If a valid `imageId` is provided in the context and it corresponds to an image within the product's gallery, that specific image will be displayed.
- If no `imageId` is provided, or if the ID is invalid, the block gracefully falls back to displaying the product's default featured image, ensuring full backward compatibility.

Closes #57585 

**Note:** To test this functionality, you'll first need to simulate a block providing the `imageId` in the context. You can do this by temporarily modifying the **Single Product** block.

1.  In `plugins/woocommerce/client/blocks/assets/js/blocks/single-product/block.json`:
    *   Add an `imageId` attribute to the `attributes` object:
        ```json
        "imageId": {
        	"type": "number"
        }
        ```
    *   Add a new context provider to the `providesContext` object:
        ```json
        "imageId": "imageId"
        ```
2.  In your WordPress admin, create a product that has a featured image and at least one other image in its gallery. Note the ID of the gallery image.
3.  Add a **Single Product** block to a new page and select the product you just created.
4.  Switch to the Code editor for the page.
5.  Add the `imageId` attribute to its JSON.

Now, proceed with the following test cases:

1.  **Default Behavior (Featured Image)**
    *   Add a Single Product block to the page.
    *   Save and view the page on the frontend and in the editor.
    *   **Expected**: The product's main featured image is displayed.

2.  **Test Case 2: Invalid Image ID**
    *   Set the `imageId` attribute in the Code editor to an ID that exists but does not belong to the product.
    *   Save and view the page.
    *   **Expected**: The product's featured image is displayed as a fallback.

3.  **Test Case 3: Specific Gallery Image**
    *   Set the `imageId` attribute in the Code editor to the valid ID of the gallery image.
    *   Save and view the page.
    *   **Expected**: The specific gallery image is displayed instead of the featured image.

4.  **Test Case 4: Regression Testing**
    *   On a different page, add a **Product Collection** and **All Producs** block.
    *   Verify that the product images within the collection still display correctly (they should all show their respective featured images).
    *   Check other blocks using the Product Image block to ensure they are unaffected.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message
Enhance Product Image block to display a specific product gallery image via context.

</details>


plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.json
```diff
@@ -5,7 +5,7 @@
 	"category": "woocommerce-product-elements",
 	"attributes": {
 		"showProductLink": { "type": "boolean", "default": true },
-		"showSaleBadge": { "type": "boolean", "default": true },
+		"showSaleBadge": { "type": "boolean", "default": false },
```
kmanijak: I assume it's leftover, we should not change it for backwards compatibility reasons

```suggestion
		"showSaleBadge": { "type": "boolean", "default": true },
```

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.json
```diff
@@ -48,7 +50,7 @@
 		"woocommerce/product-template",
 		"core/post-template"
 	],
-	"usesContext": [ "query", "queryId", "postId" ],
+	"usesContext": [ "imageId", "postId", "query", "queryId" ],
```
kmanijak, line 51: In testing steps you mention:

> Add a new context provider to the providesContext object:
> `"woocommerce/product-image/imageId": "imageId"`

so either testing steps should be updated or update `usesContext`:

```suggestion
	"usesContext": [ "woocommerce/product-image/imageId", "postId", "query", "queryId" ],
```

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.json
```diff
@@ -33,6 +34,7 @@
 			"__experimentalSkipSerialization": true
 		},
 		"spacing": {
+			"blockGap": true,
```
kmanijak: I assume it's leftover. Changing it does nothing.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.json
```diff
@@ -24,6 +24,7 @@
 			"clientNavigation": true
 		},
 		"html": false,
+		"layout": true,
```
kmanijak: This may be leftover or kept on purpose. It adds `Inner blocks use content width` option but please confirm if you wanted to add it or it should be removed.

kmanijak (changes requested):
Thanks for working on it! It looks pretty good and promising!

I tested scenarios you provided and it's testing well. In addition I checked the case providing imageId that doesn't exist (it imitates situation in which image was removed):

Editor | Frontend
-- | --
In Editor it fallbacks to featured image | In frontend nothing is displayed
image | image

---

Once this is PR is finished we can start integrating Product Image in Product Gallery:
- insert Product Image as inner block of Large Image and rename Large Image into Slider or similar
- make sure we iterate on images and render Product Image per image
- attach the iAPI directives for zoom on hover, opening full page gallery on click etc. ([source](https://github.com/woocommerce/woocommerce/blob/3a287d6b597dbc39a50635ffcc3c5358d481ab2b/plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryLargeImage.php#L123-L148)). Most likely we'll do it via HTML API
- make sure there's no a11y and UI regressions

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.json
sunyatasattva: What do you think? The [docs](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-context/#providing-block-context) recommend to include a namespace, which I think makes sense. However, none of our contexts include a namespace, so that's kinda inconsistent…

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.json
sunyatasattva: Sorry, merge conflicts were harder than I thought and some of these things happened during them.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.json
sunyatasattva: Same as above.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.json
kmanijak: Most of them don't but some of them do:
- [Add to Cart](https://github.com/woocommerce/woocommerce/blob/91fbd4f72587ac8ff28cf61038a3ec4b07bc6e22/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/block.json#L14)
- [Product Reviews](https://github.com/woocommerce/woocommerce/blob/2b22912d1ff83bc3ec94ebf457907909cba5a6ee/plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/reviews-pagination/block.json#L27)

TBH most context we provide is pretty unique and most likely we will avoid name clashes (e.g. `fullScreenOnClick` in Product Gallery) while some are less specific. `imageId` seems more common BUT... I **wouldn't** namespace it. I think it has potential to become more universal context similar to `query` or `queryId` so basically Product Image could be compatible with other blocks displaying images. In other words, I image Product Image / Image / Cover could be at some point used inside Product Gallery (or elsewhere) in order to display specific content provided via `imageId` context.

kmanijak (approved):
@sunyatasattva, I fixed failing unit tests, so here some explanation of my changes:

[This short-circuit](https://github.com/woocommerce/woocommerce/pull/57590/commits/79ef0cff4db3505e965c614b89edd38070d8b86c#diff-ea35afa38f5a5ecf63272f427676909e9f695e8b08403d676a85d07da5957051L151) broke the block's logic and unit tests were failing. When no images were provided, only placeholder was rendered and not the anchor (ignoring `showProductLink` completely). So I updated this logic. To be fair, this block would definitely benefit from some refactoring but I don't think it's time for it.

Additionally, not related to this PR, I made minor a11y improvement by removing `alt` from placeholder image, decorative images are supposed to have `alt` empty and then they have `presentation` role (hence the test update). Also, I fixed some lint issues.

I reviewed the code and did another round of testing and I feel it's ready to ship! 🚢

PR #57591, opened 2025-04-29
[WC Payments NOX in-context flows] Harden backend api logic for greater coherence

vladolaru:

We make the following changes to the backend API:
1. We introduce a global onboarding lock (via a WP option) to prevent concurrent onboarding actions from taking place (including read actions that are not idempotent)
2. We introduce stricter, standardized, and layered onboarding action checks (from the WooPayments extension needing to be active down to an action associated with a certain step requiring that step's requirements to be met).
3. We introduce dedicated exceptions that the business logic layer will use to communicate with the REST API controller layer. These allow for more granular and informative API responses, including more deliberate HTTP status codes.

Closes WOOPLUG-4102

1. The code changes should make sense and align with the overall API design.
2. The PHPUnit tests should pass.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/src/Internal/Admin/Settings/PaymentProviders/WooPayments/WooPaymentsService.php
+++ b/plugins/woocommerce/src/Internal/Admin/Settings/PaymentProviders/WooPayments/WooPaymentsService.php
@@ -730,6 +730,9 @@ class WooPaymentsService {
 		// For sanity, make sure the payment methods step is marked as completed.
 		// This is to avoid the user being prompted to set up payment methods again.
 		$this->set_onboarding_step_completed( self::ONBOARDING_STEP_PAYMENT_METHODS, $location );
+		// For sanity, make sure the test account step is marked as completed.
+		// After disabling a test account, the user should be prompted to set up a live account.
+		$this->set_onboarding_step_completed( self::ONBOARDING_STEP_TEST_ACCOUNT, $location );
 
 		return $response;
 	}
--- a/plugins/woocommerce/src/Internal/Admin/Settings/PaymentProviders/WooPayments/WooPaymentsService.php
+++ b/plugins/woocommerce/src/Internal/Admin/Settings/PaymentProviders/WooPayments/WooPaymentsService.php
@@ -4,10 +4,13 @@ declare( strict_types=1 );
 namespace Automattic\WooCommerce\Internal\Admin\Settings\PaymentProviders\WooPayments;
 
 use Automattic\Jetpack\Connection\Manager as WPCOM_Connection_Manager;
+use Automattic\WooCommerce\Internal\Admin\Settings\Exceptions\ApiArgumentException;
+use Automattic\WooCommerce\Internal\Admin\Settings\Exceptions\ApiException;
 use Automattic\WooCommerce\Internal\Admin\Settings\PaymentProviders;
 use Automattic\WooCommerce\Internal\Admin\Settings\Utils;
 use Automattic\WooCommerce\Proxies\LegacyProxy;
 use Exception;
+use WP_Http;
 
 defined( 'ABSPATH' ) || exit;
 /**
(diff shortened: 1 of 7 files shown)


plugins/woocommerce/src/Internal/Admin/Settings/PaymentProviders/WooPayments/WooPaymentsService.php
```diff
@@ -661,6 +734,98 @@ public function disable_test_account( string $location, string $from = '', strin
...
+	 * Lock the onboarding.
+	 *
+	 * This will save a flag in the database to indicate that onboarding is locked.
+	 * This is used to prevent certain onboarding actions to happen while others have not finished.
+	 * This is especially important for actions that modify the account (initializing it, deleting it, etc.)
+	 * These actions tend to be longer-running and we want to have backstops in place to prevent race conditions.
+	 *
+	 * @return void
+	 */
+	private function set_onboarding_lock(): void {
```
oaratovskyi, line 899: I feel like this logic belongs to something like `OnboardingLockManager` but I would not insist. If we don't plan to change it's implementation to redis or similar things, I guess we could omit create additional classes and leave it like that. Though, ideally, I would separate it with `NOX_ONBOARDING_LOCKED_KEY` and clear, is_locked functions.

plugins/woocommerce/src/Internal/Admin/Settings/PaymentProviders/WooPayments/WooPaymentsService.php
```diff
@@ -533,12 +542,24 @@ public function get_onboarding_kyc_session( string $location, array $self_assess
 			)
 		);
 
+		// Unlock the onboarding after the API call finished.
+		$this->clear_onboarding_lock();
```
oaratovskyi, line 578: just to be sure - are you confident that `$this->proxy->call_static` wouldn't end up with an exception that will end immediately and make this lock permanent? 
I assume that we'll have the response object in 100% cases, either WP_Error or some array, but better safe than sorry. WDYT?

oaratovskyi (commented):
Works well and the changes look good to me. Left one question before approval
___
#### Review checklist
| Parameter  | Result |
| Approach  | ✅ All good |
| Implementation  | ⚠️ Potential issue, asked for clarification |
| Automated tests  | ✅ Excellent |
| Naming  | ✅ All good |
| Code comments  | ✅ Great |

#### Scenarios tested
| Kicked new in-context and saw 409 Not acceptable when creating test account  | ✅ Functionality worked correctly |  image |

plugins/woocommerce/src/Internal/Admin/Settings/PaymentProviders/WooPayments/WooPaymentsService.php
vladolaru: Would you expand on what you mean by "change its implementation to redis or similar things"?
Also, I need further clarification about the following: "I would separate it with NOX_ONBOARDING_LOCKED_KEY and clear, is_locked functions." I don't follow.

plugins/woocommerce/src/Internal/Admin/Settings/PaymentProviders/WooPayments/WooPaymentsService.php
vladolaru: Good call, Oleks. It is unlikely that exceptions will be thrown, but it is better to be safe than sorry, especially with the danger of ending up with the onboarding locked, as you rightfully pointed out. I pushed changes to make it extra extra defensive.

plugins/woocommerce/src/Internal/Admin/Settings/PaymentProviders/WooPayments/WooPaymentsService.php
```diff
@@ -632,35 +741,160 @@ public function reset_onboarding( string $from = '', string $source = '' ): arra
...
 
+	/**
+	 * Check if an onboarding action should be allowed to be processed.
+	 *
+	 * @return void
+	 * @throws ApiException If the extension is not active or onboarding is locked.
+	 */
+	private function check_if_onboarding_action_is_acceptable() {
+		// If the WooPayments plugin is not active, we can't do anything.
+		if ( ! $this->is_extension_active() ) {
```
elazzabi, line 819: Should we also check the version of WooPayments here? Or are we leaving that to the UI-only?

EDIT: I mean, given the NOX is only available starting +9.3.0

elazzabi (approved):
The code changes make sense. I added a small question but it isn't a blocker

plugins/woocommerce/src/Internal/Admin/Settings/PaymentProviders/WooPayments/WooPaymentsService.php
vladolaru: Good call, Ahmed! I don't see a reason not to enforce it on the backend. I made [the changes](https://github.com/woocommerce/woocommerce/pull/57591/commits/a993c823756baf05b494a98305304a6a967d5250).

plugins/woocommerce/src/Internal/Admin/Settings/PaymentProviders/WooPayments/WooPaymentsService.php
oaratovskyi: Sorry for the confusion. I meant extracting set_onboarding_lock, get_onboarding_lock, is_onboarding_lock methods and NOX_ONBOARDING_LOCKED_KEY constant into `OnboardingLockManager`. If we decide to change the implementation of the lock (not database, but something else, Memcache/Redis for example - but I went too much there, sorry). But for using this outside of the WooPaymentsService, it will be beneficial. Anyway, if we need to use it elsewhere, we'll extract it to the class methods (now private methods of WooPaymentsService). Closing this.

vladolaru: Thanks for clearing that up, Oleks. For now, we will keep it simple.

PR #57594, opened 2025-04-29
Don't validate hidden postcodes in Checkout block

senadir:

This PR fixes a bug in which hidden (but previously required) postcode would prevent Checkout from being placed. By default, this bug doesn't affect any country, only via modifying Checkout via `woocommerce_get_country_locale` would a site be affected.

Closes WOOPLUG-4095

(For Bug Fixes) Bug introduced in PR https://github.com/woocommerce/woocommerce/pull/55254 but made visible in https://github.com/woocommerce/woocommerce/pull/57128

0. Add the following snippet:
```php
add_filter(
	'woocommerce_get_country_locale',
	function ( array $locale ): array {
		$locale['RO']['postcode'] = array(
			'required' => false,
			'hidden'   => true,
		);
		return $locale;
	}
);

```
2. On Checkout, select Romania.
3. Fill out your address, make sure you can place an address.
4. Select a country with a postcode visible and required.
5. Make sure you can't place an order without filling a correct postcode.
6. Select a country with no postcode (e.g UAE), make sure you can place an order.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Don't run postcode validation for hidden postcodes.
</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/use-form-validation.ts
+++ b/plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/use-form-validation.ts
@@ -148,7 +148,8 @@ export const useFormValidation = (
 		if (
 			hasSchemaRules( field, 'validation' ) && // Schema validation only run for fields with validation rules.
 			! field.hidden && // And visible
-			( field.required || field.key in values ) // And is required or has a optional with a value (or both).
+			// @ts-expect-error
+			( field.required || values[ field.key ] ) // And is required or has a optional with a value (or both).
 		) {
 			acc[ field.key ] = field.rules.validation;
 		}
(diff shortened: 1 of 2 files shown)


plugins/woocommerce-blocks/assets/js/base/components/cart-checkout/form/use-form-validation.ts
```diff
@@ -148,7 +148,8 @@ export const useFormValidation = (
 		if (
 			hasSchemaRules( field, 'validation' ) && // Schema validation only run for fields with validation rules.
 			! field.hidden && // And visible
-			( field.required || field.key in values ) // And is required or has a optional with a value (or both).
+			// @ts-expect-error
+			( field.required || values[ field.key ] ) // And is required or has a optional with a value (or both).
```
senadir, line 152: TS would complain here that field.key is not part of values, I will dig into this in a follow up TS clean up.

tjcafferkey (approved):
This tests well and code LGTM - thanks for the explanation/follow-up comment around the expect-error

tjcafferkey:
@senadir lets also follow up with some test coverage as priority I think once this is wrapped up.

PR #57611, opened 2025-04-30
Incubate interactivity API fork in WooCommerce (behind feature flag)

samueljseay:

When you have the feature flag enabled this PR will replace the bundled interactivity API with a transpiled version of the interactivity API contained within the WooCommerce fork of Gutenberg, allowing us to experiment with potential solutions to problems (such as exporting more of Preact), that would not be possible right now without them being done already upstream in Gutenberg.

This fixes the version of the fork to the branch `interactivity-api-001` to ensure a consistent version is installed each time.

To do this we:

* Install the interactivity API via pnpm's feature that allows installing dependencies from a git sub directory
* Have a separate Webpack build for the interactivity API installed via github urls
* Deregister the bundled interactivity API scripts to ensure that the importmap will point to our version
* Register and enqueue our version of the interactivity API
* Add a feature flag that only enables the runtime if the flag is enabled, it is enabled by default in development

I've opted not to re-export anything in our own package (ie `@woocommerce/interactivity`) because this will reduce any transition work needed in existing blocks and also if we want to go back to the bundled version we won't need to do any extra effort.

N/A

For now, the best we have is:

1. Existing e2e tests covering the interactive blocks
2. Smoke test/manual test the interactive blocks

To test that the fork built version is indeed being loaded you can:

1. Load a page with interactive blocks (such as product collection and product filters)
2. Search the HTML for `importmap` using dev tools, see that `@wordpress/interactivity` points to the one built in woocommerce rather than the default bundled version
3. Also check network to see that only the `interactivity.js` and `interactivity-router.js` built by us are loaded.

You can also test that when the feature flag is disabled the core runtime is registered and used.

1. Install woocommerce-beta-tester plugin which is available in the monorepo.
2. From the features section of that plugin enable `experimental-iapi-runtime` toggle
3. Follow steps from before and check the import map points to the core interactivity api modules rather than our built ones.

N/A

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Behind a feature flag de-register the bundled `@wordpress/interactivity` and `interactivity-router` modules and register/bundle our own within WooCommerce.
</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/tsconfig.json
```diff
@@ -4,12 +4,11 @@
 		"./assets/js/**/*",
 		"./packages/checkout/**/*",
 		"./packages/components/**/*",
-		"./packages/interactivity-components/**/*",
```
samueljseay, line 7: Just some clean up, spotted this which is no longer used.

plugins/woocommerce/client/blocks/babel.config.js
```diff
@@ -11,14 +11,7 @@ module.exports = {
 		},
 	},
 	presets: [
-		[
-			'@wordpress/babel-preset-default',
-			{
-				targets: {
-					node: 'current',
```
samueljseay, line 18: I don't know why this was here tbh. I could remove this change, because I don't think we're relying on this config right now, but I also don't think this should be here at all.

plugins/woocommerce/client/blocks/package.json
```diff
@@ -264,6 +264,7 @@
 		"@emotion/styled": "^11.11.0",
 		"@preact/signals": "^1.3.0",
 		"@woocommerce/tracks": "workspace:*",
+		"@wordpress/a11y": "4.22.0",
```
samueljseay, line 268: Needed this as a dependency otherwise pnpm would doggedly try to resolve the dependency from relative path regardless of what I did in `readPackage`

plugins/woocommerce/client/blocks/bin/webpack-config-interactive-blocks.js
```diff
@@ -77,6 +77,21 @@ module.exports = {
...
+									'@wordpress/babel-preset-default'
+								),
+							],
+						},
+					},
+				],
+			},
 			...moduleConfig.module.rules.filter(
 				( rule ) =>
 					! rule.test.test( '.css' ) &&
```
luisherranz: I wonder why is this rule needed now, and if it is really needed, why not remove the filter for JS/TS that comes by default?

.pnpmfile.cjs
```diff
@@ -159,5 +192,25 @@ function afterAllResolved( lockfile, context ) {
...
+			// We resolve @wordpress/interactivity and @wordpress/interactivity-router via pnpm's git sub dir feature
+			// and as a result need to resolve some of their dependencies also via sub dir.
+			if ( pkg.name === '@wordpress/interactivity-router' ) {
+				if (
+					pkg.dependencies &&
+					pkg.dependencies[ '@wordpress/a11y' ] &&
+					pkg.dependencies[ '@wordpress/interactivity' ]
+				) {
+					//  Use the latest from npm
+					pkg.dependencies[ '@wordpress/a11y' ] = '4.22.0';
```
luisherranz: I guess we have to keep this synchronized with the version of the `package.json`. Perhaps instead of hardcoding this version, it could be imported directly from the `package.json` itself?

plugins/woocommerce/src/Blocks/AssetsController.php
```diff
@@ -57,6 +57,21 @@ public function register_script_modules() {
...
 
+		$interactivity_api_asset_data = $this->api->get_asset_data(
+			$this->api->get_block_asset_build_path( 'interactivity-api-assets', 'php' )
+		);
+
+		// Dequeue the WordPress interactivity script modules and ensure it's removed from the import map.
+		wp_dequeue_script_module( '@wordpress/interactivity' );
+		wp_deregister_script_module( '@wordpress/interactivity' );
+		wp_dequeue_script_module( '@wordpress/interactivity-router' );
+		wp_deregister_script_module( '@wordpress/interactivity-router' );
```
luisherranz: There shouldn't be a need to dequeue, since these modules shouldn't be enqueued.

Aside from that, I'm seeing at what point it registers the WordPress modules, and it seems to be in the `wp_default_scripts` action.

https://github.com/WordPress/wordpress-develop/blob/0aa29fd973213146d7b7875d17d3729990eb66a6/src/wp-includes/default-filters.php#L587

So, since the action `wp_default_scripts` runs after `init`, and this WooCommerce function (`register_script_modules`) is running on `init`, it seems to me that what is happening here is that WooCommerce is registering the interactivity modules first, and when Core tries to register them, they are simply [not registered again](https://github.com/WordPress/wordpress-develop/blob/0aa29fd973213146d7b7875d17d3729990eb66a6/src/wp-includes/class-wp-script-modules.php#L76).

What Gutenberg is doing is directly deleting the WordPress action and adding its own. But they can do that because they replace all the Core packages and we only want to replace the interactivity ones.

https://github.com/WordPress/gutenberg/blob/11adb45dcf87204a17604c7a5d67bb3d36b4508e/lib/client-assets.php#L660C39-L661

So, there seem to be two options: 

1. Register the modules before WordPress Core (and Gutenberg). It would be something similar to what is happening now.
2. Wait for Core/Gutenberg to register the modules, then unregister them, and register the WooCommerce ones. 

The disadvantage of Option 1 is that if Gutenberg someday decides, for whatever reason, to change its approach and unregister the modules instead of deleting the WordPress action, it would actually end up unregistering the WooCommerce modules.

For Option 2, I wonder if the best way to do it would be to also use the `wp_default_scripts` action, but with a lower priority so it runs after the Core/Gutenberg one.

Thoughts?

cc: @darerodz

plugins/woocommerce/client/blocks/bin/webpack-config-interactivity.js
```diff
@@ -0,0 +1,72 @@
...
+ */
+const { sharedOptimizationConfig } = require( './webpack-shared-config' );
+
+// Config to build and incubate the interactivity API within WooCommerce.
+module.exports = {
+	entry: {
+		'@wordpress/interactivity': path.resolve(
+			__dirname,
+			'..',
+			'node_modules/@wordpress/interactivity/src/index.ts'
```
luisherranz, line 21: We would still need to add the entry point that contains `preact/debug` (`debug.ts`), but we can do it later if its really needed.

luisherranz (changes requested):
Great work, Sam. It looks fine to me, although I think we should rethink the way new script modules are being unregistered/registered. I've cc'ed @DAreRodz to sync that work.

plugins/woocommerce/client/blocks/bin/webpack-config-interactive-blocks.js
samueljseay: Ah this is left in by mistake, I'll clean this up

plugins/woocommerce/src/Blocks/AssetsController.php
DAreRodz: I guess Gutenberg would not change the current approach of deleting the WordPress action. The second approach still looks safer, though. 🤔

As you mentioned, we could add a hook to the `wp_default_scripts` action with a lower priority (e.g., 20) that unregisters Core's or Gutenberg's iAPI modules and registers the WooCommerce ones.

Regarding the plugin we plan to release for Full-Page Navigation, we would have to decide which plugin would take precedence:  WooCommerce or the Full-Page Navigation one.

luisherranz: Oh yeah, sorry, I forgot to tell you. That plugin has to autodeactivate (or just bail off) if it detects that WooCommerce is activated. So yes, WooCommerce should always take precedence.

samueljseay: > As you mentioned, we could add a hook to the wp_default_scripts action with a lower priority (e.g., 20) that unregisters Core's or Gutenberg's iAPI modules and registers the WooCommerce ones.

This sounds good to me, but I will wait to hear what the general consensus is about merging this PR first.

luisherranz (changes requested):
Oh, I almost forgot. Before merging, we have to create a new branch in the fork that includes some sort of versioning, so that the interactivity API code does not automatically synchronize with the changes in the fork every time someone does an npm install. 

Something like `interactivity-api-001`.

plugins/woocommerce/client/blocks/bin/webpack-config-interactivity.js
```diff
@@ -0,0 +1,72 @@
...
+		path: path.resolve( __dirname, '../build/' ),
+		asyncChunks: false,
+		chunkFormat: 'module',
+		environment: { module: true },
+		module: true,
+	},
+	resolve: {
+		extensions: [ '.js', '.ts', '.tsx' ],
+	},
+	plugins: [
```
DAreRodz, line 50: In Gutenberg, we're using `DefinePlugin` to assign values to certain global variables.

https://github.com/woocommerce/gutenberg/blob/62037f3d0d0c31dae74f5c76a01105bde91afaf4/tools/webpack/shared.js#L65-L76

The Interactivity API, in particular, uses `globalThis.SCRIPT_DEBUG` to remove certain logs from the bundle. I think we should do the same here.

DAreRodz: Confirmed. 🕵️  I can see this in the bundled code:

```
if(globalThis.SCRIPT_DEBUG&&!t)throw Error("Cannot call `getContext()` when there is no scope. If you are using an async function, please consider using a generator instead. If you are using some sort of async callbacks, like `setTimeout`, please wrap the callback with `withScope(callback)`.");
```

luisherranz: Sure, go ahead 🙂

samueljseay: Sounds good :)

samueljseay: https://github.com/woocommerce/woocommerce/pull/57611/commits/fb5603e033966fafd48a091ca94565c7260affc8

plugins/woocommerce/src/Blocks/AssetsController.php
samueljseay: So as discussed we've decided to put the runtime behind a feature flag and I've also taken option 2 here to register the scripts: https://github.com/woocommerce/woocommerce/pull/57611/commits/c746904d883c933633851a375cfd7e39730e52d0

.pnpmfile.cjs
samueljseay: Yep good call. I utilize the `@woocommerce/block-libary` installed version if possible here: https://github.com/woocommerce/woocommerce/pull/57611/commits/7878cfdc17685f4b7e20e3a32a94037729076d1c

plugins/woocommerce/client/blocks/bin/webpack-config-interactivity.js
samueljseay: Yep, I'm thinking for now to just leave this. We can add later if needed as you say.

samueljseay:
@dinhtungdu Pulled you in as my favorite blocks reviewer 🤣  you're welcome.

plugins/woocommerce/src/Blocks/AssetsController.php
```diff
@@ -44,33 +45,40 @@ protected function init() { // phpcs:ignore WooCommerce.Functions.InternalInject
...
+
+		if ( Features::is_enabled( 'experimental-iapi-runtime' ) ) {
+			// Run after the WordPress iAPI runtime has been registered by setting a lower priority.
+			add_filter( 'wp_default_scripts', array( $this, 'deregister_core_iapi_runtime' ), 20 );
+		}
 	}
 
 	/**
-	 * Register script modules.
+	 * De-register the iAPI runtime registered by WordPress core/Gutenberg, wllowing us to register our own version of the iAPI runtime.
```
luisherranz:
```suggestion
	 * Re-registers the iAPI runtime registered by WordPress Core/Gutenberg, allowing WooCommerce to register its own version of the iAPI runtime.
```

plugins/woocommerce/src/Blocks/AssetsController.php
```diff
@@ -44,33 +45,40 @@ protected function init() { // phpcs:ignore WooCommerce.Functions.InternalInject
...
 	 */
-	public function register_script_modules() {
-		// Right now we only have one script modules build for supported interactivity API powered block front-ends.
-		// We generate a combined asset file for that via DependencyExtractionWebpackPlugin to make registration more
-		// efficient.
-		$asset_data = $this->api->get_asset_data(
-			$this->api->get_block_asset_build_path( 'interactivity-blocks-frontend-assets', 'php' )
-		);
+	public function deregister_core_iapi_runtime() {
```
luisherranz: Perhaps this function should be called reregister, since it is not only unregistering the interactivity API from Core, but also registering the WooCommerce interactivity API?

```suggestion
	public function reregister_core_iapi_runtime() {
```

luisherranz (approved):
LGTM! Thanks Sam! 🎉

plugins/woocommerce/src/Blocks/AssetsController.php
samueljseay: good suggestion, thanks https://github.com/woocommerce/woocommerce/pull/57611/commits/951e5fc6b875086f10cd3530fd3ac6ed2aef7ce2

plugins/woocommerce/src/Blocks/AssetsController.php
```diff
@@ -44,6 +45,28 @@ protected function init() { // phpcs:ignore WooCommerce.Functions.InternalInject
...
+		wp_deregister_script_module( '@wordpress/interactivity-router' );
+
+		$interactivity_api_asset_data = $this->api->get_asset_data(
+			$this->api->get_block_asset_build_path( 'interactivity-api-assets', 'php' )
+		);
+
+		foreach ( $interactivity_api_asset_data as $handle => $data ) {
+			$handle_without_js = str_replace( '.js', '', $handle );
+			wp_register_script_module( $handle_without_js, plugins_url( $this->api->get_block_asset_build_path( $handle_without_js ), dirname( __DIR__ ) ), $data['dependencies'], $data['version'] );
+		}
```
dinhtungdu: Can we be more defensive here? By only deregistering after ensuring the `$interactivity_api_asset_data` is valid.

dinhtungdu (changes requested):
LGTM! Tested well with Product Collection and Product Filters against both current and new runtime.

> Rebuild the app

Can we mention WC Beta Testers instead in the testing instructions to enable/disable the new runtime? I can see that we need to build the app to update the features array, but this gives me a false sense that we need to rebuild the app when switching the runtime.

plugins/woocommerce/src/Blocks/AssetsController.php
samueljseay: `get_asset_data` already validates the file exists, if it doesn't it returns an empty array and there is no-op.

So to be a bit more defensive here, I'll check if the handles match the packages during the loop and only de-register then. Thanks its a good suggestion :)

samueljseay: Resolved here: https://github.com/woocommerce/woocommerce/pull/57611/commits/ae3731fcacd5418a82d19ae10bd2fcbb904f4c86

Note I could have just deregistered every `handle_without_js` but just to be extra careful I'm explicitly only deregistering handles that match the core iapi runtimes.

samueljseay:
Thanks for the great review @dinhtungdu I believe I've addressed all your concerns.

dinhtungdu (approved):
Thanks for addressing my feedback! This is LGTM :shipit:

PR #57624, opened 2025-04-30
[Blueprint] Fix Missing Settings in Blueprint Export

chihsuan:

Closes WOOPLUG-4106

Several WooCommerce settings are not being exported in Blueprint, including Tax, Emails, and Integrations.

- Some basic Tax settings are not exported because we missed to extend the `ExportWCSettings` class to export the basic settings.
- The individual Emails and Integrations settings are not exported because they are registered as wc setting page instances but created by the wc setting page API class.

To ensure complete settings export functionality, I modified the `SettingOptions` class to retrieve settings using the `WC_REST_Setting_Options_Controller` class rather than iterating through `WC_Admin_Settings::get_settings_pages()` and also changed the exporters to correctly export the settings.

1. Go to `WooCommerce > Settings > General` and enable taxes
2. Go to `WooCommerce > Settings > Tax` and change the tax settings
3. Go to `WooCommerce > Settings > Emails` and change the email/Email notifications settings
4. Run `wp wc blueprint export blueprint-tax.json --user=<admin-username> --steps=setWCSettingsTax` and confirm that all tax settings are exported
5. Run `wp wc blueprint export blueprint-emails.json --user=<admin-username> --steps=setWCSettingsEmails` and confirm that all email settings are exported
6. Run `wp wc blueprint export blueprint-integrations.json --user=<admin-username> --steps=setWCSettingsIntegrations` and confirm that all integration settings are exported
7. Optionally, try to export other settings pages and confirm that the blueprint file contains the expected settings

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Admin/Features/Blueprint/SettingOptions.php
```diff
@@ -4,162 +4,89 @@
...
 
 	/**
-	 * Array of WC_Settings_Page objects.
+	 * Ignore setting types.
 	 *
-	 * @var WC_Settings_Page[]
+	 * @var array
 	 */
-	private array $setting_pages;
+	private $ignore_setting_types = array( 'title', 'sectionend', 'slotfill_placeholder' );
```
louwie17: I don't think this is relevant anymore right given we make use of the settings controller now.
It looks like any of the above items aren't in the settings type list: https://github.com/woocommerce/woocommerce/blob/0a9c0049b300f2f42dd74899485e2eb1278a5708/plugins/woocommerce/includes/rest-api/Controllers/Version3/class-wc-rest-setting-options-controller.php#L236

plugins/woocommerce/src/Admin/Features/Blueprint/SettingOptions.php
```diff
@@ -4,162 +4,89 @@
...
+	 * Get a settings page by ID.
 	 *
-	 * @param array $page_info The page info.
+	 * @deprecated 9.9.0 This method is no longer used.
+	 * @return null
+	 */
+	public function get_page() {
+		wc_deprecated_function( __METHOD__, '9.9.0' );
+		return null;
+	}
```
louwie17: How come we are deprecating this and the `get_page_info`? given this feature hasn't been officially released yet?
I would say, just remove them, unless your aware of places where it is already being extended?

louwie17 (changes requested):
Nice call on making use of `WC_REST_Setting_Options_Controller`, that simplifies things quite a bit!
I left a couple comments, let me know what you think. Otherwise it did test well :)

plugins/woocommerce/src/Admin/Features/Blueprint/SettingOptions.php
chihsuan: The API endpoint's list of valid setting types appears to be incomplete, as it omits several supported type:

- email_image_url
- email_font_family  
- relative_date_selector

 (We probably need to update the API endpoint.)

Also, we have to ignore types here because we use `get_group_settings` instead of the API endpoint (get_items). The `get_group_settings` method returns all settings without filtering out the invalid types.

We need to use `get_group_settings` because the API endpoint does not include the `option_key` required to retrieve database values.

plugins/woocommerce/src/Admin/Features/Blueprint/SettingOptions.php
chihsuan: I was thinking about removing them too but I marked them as deprecated since the class was introduced in 9.8.0. But your point is valid, it should be safe to remove them since the feature is not released yet and don't see any reason why it's being used by 3rd parties. I'll remove them directly.

chihsuan: Removed them in https://github.com/woocommerce/woocommerce/pull/57624/commits/643b53eb6b1a70e4c4f35339e5df773285878325

louwie17 (approved):
Thanks for the removal of the deprecated functions and explaining the filter. All is good in that case :)

PR #57673, opened 2025-05-01
Fix various console warnings from useSelect usage on cart/checkout

mikejolley:

Fixes warnings in our useSelect usage. Context: https://make.wordpress.org/core/2025/03/12/data-a-helpful-performance-warning-for-developers-in-the-useselect-hook/

Should be covered by existing CI, touches validation, addresses, and payment statuses.

1. CI should pass.
2. Smoke test cart/checkout to place order and validate fields.
3. Keep console open and ensure there are no warnings relating to useSelect.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/blocks/assets/js/base/context/hooks/cart/use-store-cart-coupons.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/base/context/hooks/cart/use-store-cart-coupons.ts
@@ -28,22 +28,15 @@ export const useStoreCartCoupons = ( context = '' ): StoreCartCoupon => {
 	const { createNotice } = useDispatch( 'core/notices' );
 	const { setValidationErrors } = useDispatch( validationStore );
 
-	const {
-		isApplyingCoupon,
-		isRemovingCoupon,
-	}: Pick< StoreCartCoupon, 'isApplyingCoupon' | 'isRemovingCoupon' > =
-		useSelect( ( select ) => {
-			const store = select( cartStore );
-
-			return {
-				isApplyingCoupon: store.isApplyingCoupon(),
-				isRemovingCoupon: store.isRemovingCoupon(),
-			};
-		} );
+	const [ isApplyingCoupon, isRemovingCoupon ] = useSelect( ( select ) => {
+		const store = select( cartStore );
+		return [ store.isApplyingCoupon(), store.isRemovingCoupon() ];
+	}, [] );
 
 	const { applyCoupon, removeCoupon } = useDispatch( cartStore );
-	const orderId = useSelect( ( select ) =>
-		select( checkoutStore ).getOrderId()
+	const orderId = useSelect(
+		( select ) => select( checkoutStore ).getOrderId(),
+		[]
 	);
 
 	// Return cart, checkout or generic error message.
--- a/plugins/woocommerce/client/blocks/assets/js/base/context/hooks/cart/use-store-cart-coupons.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/base/context/hooks/cart/use-store-cart-coupons.ts
@@ -2,14 +2,18 @@
  * External dependencies
  */
 import { __, sprintf } from '@wordpress/i18n';
-import { useDispatch, useSelect } from '@wordpress/data';
+import { dispatch, useDispatch, useSelect } from '@wordpress/data';
+import { useCallback } from '@wordpress/element';
 import {
 	cartStore,
 	validationStore,
 	checkoutStore,
 } from '@woocommerce/block-data';
 import { decodeEntities } from '@wordpress/html-entities';
-import type { StoreCartCoupon, ApiErrorResponse } from '@woocommerce/types';
+import type {
+	StoreCartCoupon,
+	CouponApiErrorResponse,
+} from '@woocommerce/types';
 import { applyCheckoutFilter } from '@woocommerce/blocks-checkout';
 
 /**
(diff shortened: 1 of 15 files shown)


plugins/woocommerce/client/blocks/assets/js/base/context/hooks/payment-methods/use-payment-method-interface.ts
```diff
@@ -56,61 +56,74 @@ export const usePaymentMethodInterface = (): PaymentMethodInterface => {
...
 			};
-		} );
-	const { paymentStatus, activePaymentMethod, shouldSavePayment } = useSelect(
-		( select ) => {
-			const store = select( paymentStore );
+		}, [] );
 
-			return {
-				// The paymentStatus is exposed to third parties via the payment method interface so the API must not be changed
-				paymentStatus: {
```
mikejolley, line 66: This object caused the problem because it changes with every re-render. Fix was simply to move each status to top level and construct the object later on.

plugins/woocommerce/client/blocks/assets/js/base/context/hooks/use-validation.ts
```diff
@@ -18,34 +18,32 @@ export const useValidation = (): ValidationData => {
...
-	const { hasValidationErrors, getValidationError } = useSelect(
+	const { hasValidationErrors, getValidationErrorFromStore } = useSelect(
 		( mapSelect ) => {
 			const store = mapSelect( validationStore );
 			return {
 				hasValidationErrors: store.hasValidationErrors(),
-				getValidationError: ( validationErrorId: string ) =>
-					store.getValidationError(
-						`${ prefix }-${ validationErrorId }`
-					),
```
mikejolley, line 29: This wasn't stable so I've moved it out of the useSelect to a useMemo below.

opr (changes requested):
There is still a warning when checking out with coupons, I think it's coming from https://github.com/woocommerce/woocommerce/blob/99f04562c084808e02515a526664fdc0c77fa688/plugins/woocommerce/client/blocks/assets/js/base/context/hooks/cart/use-store-cart-coupons.ts#L35 This is visible on Cart and Checkout.

Besides that one, I couldn't get any warnings to show, I checked out with Stripe and COD, and I also tested functionality where errors in shipping/billing addresses for logged in customers would expand the address forms, which still worked fine.

plugins/woocommerce/client/blocks/assets/js/data/cart/selectors.ts
```diff
@@ -26,17 +27,21 @@ export const getCartData = ( state: CartState ): Cart => {
...
-): {
-	shippingAddress: ShippingAddress;
-	billingAddress: BillingAddress;
-} => {
-	return {
-		shippingAddress: state.cartData.shippingAddress,
-		billingAddress: state.cartData.billingAddress,
-	};
-};
+export const getCustomerData = createSelector(
```
opr, line 30: Why is this the only selector updated to `createSelector` what is the benefit compared to leaving it as-is?

It looks like most selectors just select a single item directly from state, but `getCartItem`, `isItemPendingDelete`, and `isItemPendingQuantity` do some work before returning a value.

mikejolley: I think its because the selector returns a new object `{}`. Every time the selector is used that object will change. Create Selector handles memoization. 

> createSelector memoizes the selector function, so it only recalculates the result if the input state changes. This means:
> If you call getCustomerData with the same state object (or with state where the relevant parts haven't changed), it returns the cached result instead of recomputing.
> This is useful for performance, especially in React, because it prevents unnecessary re-renders and recalculations when the data hasn't changed.
> It also helps with referential equality: if the output is the same, the returned object is the same reference, which is important for shallow comparison optimizations in React components.
> Restated: createSelector wraps your selector so it only recomputes when its inputs change, making your app faster and reducing unnecessary work. This is especially important for selectors that return objects or arrays, since new references can trigger re-renders even if the data is the same.
> If you want to see the implementation, it's from @wordpress/data (which is similar to reselect's createSelector). It tracks the last arguments and result, and only calls your function if the inputs are different.
> Sources:
> https://developer.wordpress.org/block-editor/reference-guides/data/data-selectors/#createselector
> https://github.com/reduxjs/reselect
> https://github.com/WordPress/gutenberg/blob/trunk/packages/data/src/create-selector.js

opr: Right but the others only return booleans or objects that already exist in state. Got it ✋🏼

mikejolley:
@opr I've pushed a fix for coupons and another I found when editing checkout relating to payment methods and lack of createSelector. There is one I could not resolve which I think may be coming from `createNotice`—when you remove a coupon or add a coupon there is one console warning. I will leave this one as I hit a wall debugging it.

plugins/woocommerce/client/blocks/assets/js/base/context/hooks/cart/use-store-cart-coupons.ts
```diff
@@ -24,116 +28,122 @@ import { useStoreCart } from './use-store-cart';
...
+		( select ) => select( cartStore ).isApplyingCoupon(),
+		[]
+	);
+	const isRemovingCoupon = useSelect(
+		( select ) => select( cartStore ).isRemovingCoupon(),
+		[]
+	);
+	const orderId = useSelect(
+		( select ) => select( checkoutStore ).getOrderId(),
+		[]
```
opr: Why did you split these out into individual `useSelect`s? I see originally we had two, one for the cart store and one for the checkout store, but why can't we combine them and do it like this? I tested it and didn't see any warnings.

```js
const { isApplyingCoupon, isRemovingCoupon, orderId } = useSelect(
	( select ) => ( {
		isApplyingCoupon: select( cartStore ).isApplyingCoupon(),
		isRemovingCoupon: select( cartStore ).isRemovingCoupon(),
		orderId: select( checkoutStore ).getOrderId(),
	} ),
	[]
);
```

plugins/woocommerce/client/blocks/assets/js/base/context/hooks/cart/use-store-cart.ts
```diff
@@ -122,122 +121,113 @@ export const defaultCartData: StoreCart = {
...
+			[ shouldSelect ]
+		);
+
+	const cartFees = useMemo( () => {
+		return cartData.fees.length > 0
+			? cartData.fees.map( ( fee: CartResponseFeeItem ) =>
+					decodeValues( fee )
+			  )
+			: EMPTY_CART_FEES;
+	}, [ JSON.stringify( cartData.fees ) ] );
```
opr: You shouldn't do this,  `JSON.stringify` is expensive and would run on each render, negates the point of `useMemo`.

Also `JSON.stringify` creates a new string on each render, then `useMemo` compares it to the old one, it won't match because they are _referentially_ not equal, the memoized function will therefore recompute.

I think just using `cartData.fees` might be OK here, I think it is the same object unless the state is changed in the data store, so if this hook reruns but data store hasn't changed it would not recompute.

plugins/woocommerce/client/blocks/assets/js/base/context/hooks/cart/use-store-cart.ts
```diff
@@ -122,122 +121,113 @@ export const defaultCartData: StoreCart = {
...
+					cartData: store.getCartData(),
+					cartErrors: store.getCartErrors(),
+					cartTotals: store.getCartTotals(),
+					cartIsLoading:
+						// @ts-expect-error `hasFinishedResolution` is not typed in @wordpress/data yet.
+						! store.hasFinishedResolution( 'getCartData' ),
+					isLoadingRates: store.isCustomerDataUpdating(),
+				};
+			},
+			[ shouldSelect ]
```
opr: Is this actually doing anything useful here? As I understand it, this would only recreate the selector function if `shouldSelect` changes, I think it's fine to keep the deps array empty here since `shouldSelect` doesn't affect the outcome of the function.

opr (changes requested):
Thanks for looking at the additional warnings I found, I left some comments. I think it might be better to create a follow up issue for the `createNotice` warning.

plugins/woocommerce/client/blocks/assets/js/base/context/hooks/cart/use-store-cart.ts
mikejolley: Hmm I thought we'd used this method before but its not in the codebase. https://medium.com/@ivanreshetar/enhanced-strategies-for-using-objects-and-arrays-as-dependencies-in-react-hooks-43bbfa4f20d2

`cartData.fees` is an object so I think it will have issues. I'll have to test and see if its running on every render.

plugins/woocommerce/client/blocks/assets/js/base/context/hooks/cart/use-store-cart-coupons.ts
mikejolley: https://github.com/woocommerce/woocommerce/pull/57673/commits/73922186231e5bd1a9a8ee9e45aaaf7066371695

plugins/woocommerce/client/blocks/assets/js/base/context/hooks/cart/use-store-cart.ts
mikejolley: https://github.com/woocommerce/woocommerce/pull/57673/commits/ddae31f21f3a13a65fabbeabec7d06092b1144bc

plugins/woocommerce/client/blocks/assets/js/base/context/hooks/cart/use-store-cart.ts
mikejolley: https://github.com/woocommerce/woocommerce/pull/57673/commits/a25d487e8089f13b7ab1da1d9e29df2836967700

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-shipping-address-block/customer-address.tsx
```diff
@@ -27,24 +27,28 @@ const CustomerAddress = () => {
...
+	const invalidProps = useMemo( () => {
+		return Object.keys( shippingAddress )
+			.filter( ( key ) => {
+				return (
+					getValidationErrorSelector( 'shipping_' + key ) !==
+					undefined
+				);
+			} )
+			.filter( Boolean );
+	}, [ JSON.stringify( shippingAddress ), getValidationErrorSelector ] );
```
opr: Tested with and without stringify, did not notice any difference in amount of times the hook was called. Also tested with this snippet:

```js
const shippingAddressRef = useRef( shippingAddress );

useEffect( () => {
	console.log( 'Previous address:', shippingAddressRef );
	console.log( 'Current address:', shippingAddress );
	console.log(
		'Same reference?',
		shippingAddressRef.current === shippingAddress
	);

	// Update the ref after logging
	shippingAddressRef.current = shippingAddress;
}, [ shippingAddress ] );
```

and it's the same ref. The ref would not change unless the shipping address values change, in which case a recomputation of the memo is expected and OK.

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-billing-address-block/customer-address.tsx
```diff
@@ -27,25 +27,28 @@ const CustomerAddress = () => {
...
+	const invalidProps = useMemo( () => {
+		return Object.keys( billingAddress )
+			.filter( ( key ) => {
+				return (
+					key !== 'email' &&
+					getValidationErrorSelector( 'billing_' + key ) !== undefined
+				);
+			} )
+			.filter( Boolean );
+	}, [ JSON.stringify( billingAddress ), getValidationErrorSelector ] );
```
opr: See comment on shipping address, think this is fine to use the object directly.

opr (approved):
We discussed the changes to `useStoreCart` and the fact that the hook now returns default data early if `shouldSelect` is false. We agreed this is fine, and the only difference between the new code and the old is that the ref containing store data will no longer be updated to contain default data before returning. This is fine, if the data from the store is different to what's in the ref, the ref would be updated with that and return the new data anyway.

I tested the cart/checkout in front end and editor, including adding/removing coupons, changing quantities, checkout with different payment methods. Thanks for working on this Mike.

PR #53466, opened 2024-12-04
Remove shipping if there is no address, always allow hide shipping costs option

samueljseay:
Closes #52335 

With some adjustments:

1. We used to disable "hide shipping costs until an address is entered" when local pickup was enabled. We now allow it to be enabled but we explain the behaviour (`Local pickup rates will display in the Checkout block, even without an address`). @opr and I agreed this experience felt clearer and covered all cases well, but it would be good to get some feedback about this cc @pmcpinto 

2. If an address has not been entered and the user has hide shipping costs enabled we do not display the *Shipping* options

Added by @opr 

- Changes the shipping calculator logic to display "Enter an address to check delivery options" if only a partial address has been entered (e.g. when the shopper has a default location)
- We previously hid the "Ship" selector when:
  - Shipping is not required
  - Shipping calculations have not been completed
  - No shipping rates are available
  - The item is not collectible
  - Local pickup is not enabled
  - No shipping methods are currently available
- however this would hide it even if rates _might_ become available later, after a full address had been entered. Instead I changed the logic to force the control to show if the following is true:
  - Shipping is enabled, and
  - The cart needs shipping, and
  - All packages in the cart support collectible, and
  - Local pickup is enabled, and
  - Shipping methods exist on the site
- Updated shipping calculator so it doesn't show a "No delivery options available for XXX, XX" message unless there's a full address. I figured the shopper could be negatively affected by such a message if they haven't even filled their address in and they're already seeing a message saying there are no shipping methods. Because of this, I updated the tests in `plugins/woocommerce-blocks/tests/e2e/tests/cart/cart-checkout-block-shipping.block_theme.spec.ts`.
- Added a `has_full_shipping_address` method to `ShippingController` - this uses `WC()->countries->get_country_locale` to determine whether the state, city, country, postcode (based on locale requirement) is filled before stripping out "shipping" methods (and leaving pickup methods intact).
  - Needed because we have to determine whether to remove rates from the response if shipping requires a full address.
  

### How to test the changes in this Pull Request:

Smoke test a few scenarios to ensure no regressions. For the shopper part it is best to open an incognito window to clear the address state that would be remembered between scenarios.

#### Scenario 1
1. Have local pickup disabled
2. Have hide shipping costs until address entered **enabled**
3. As shopper go through cart and checkout and ensure that shipping costs are not shown on cart page / checkout page until address entered

#### Scenario 2

1.  Have local pickup disabled
2. Have hide shipping costs until address entered **disabled**
[...]

PR #57674, opened 2025-05-01
Prevent checkout without paying for shipping when hiding core fields

opr:

- This PR semi reverts https://github.com/woocommerce/woocommerce/pull/55804
- Adds a new `$cart_context` property to the WC cart, this lets us specify routes through legacy code depending on the source of the `WC_Cart` object. In WC Blocks, the context is set to `store-api` in other contexts it is set to `shortcode`.
- We use this to determine which code path to take in `has_full_shipping_address` - we need to support certain hooks in shortcode, but not in WC Blocks (we use locale only in WC Blocks, whereas in Shortcode we use `woocommerce_shipping_fields` because the list of required shipping fields is retrieved from `get_address_fields()`
- Expands on the show_shipping test case to test filtering functionality around this.

Closes WOOPLUG-4037

(For Bug Fixes) Bug introduced in PR https://github.com/woocommerce/woocommerce/pull/55804 and https://github.com/woocommerce/woocommerce/pull/53466

[Note, in the below videos, I have made postcode optional using the ThemeHigh checkout field editor plugin](https://en-gb.wordpress.org/plugins/woo-checkout-field-editor-pro/) which we believe was a very common plugin installed on sites experiencing this issue.

| Before | After |
| ------ | ----- |
| https://www.loom.com/share/ed626d5a8e1c4f88b7b49154aebd69d9 | https://www.loom.com/share/67efeafb77e54af48f5ded0e8bab328b |

#### Happy path testing 1

This happy path makes the following assumption:
 - Extensions modifying the address fields through `woocommerce_shipping_fields` and `woocommerce_billing_fields`  _without_ updating locale correctly would still make the frontend fields optional on the JS side. (This is what happens in the checkout field editor plugin this was tested against, there is some JS that removes the "required" classes from non-required fields).

1. Install [ThemeHigh checkout field editor plugin](https://en-gb.wordpress.org/plugins/woo-checkout-field-editor-pro/)  
2. Go to the settings (WC -> Checkout Fields) and make the billing and shipping postcode optional (i.e. not required)
image

3. Set up shipping zones and ensure methods are set up too.
4. Go to WC -> Settings -> Shipping -> Shipping Settings. Enable "Hide shipping costs until an address is entered"
5. Add an item to your cart and go to the shortcode checkout.
6. Fill in your details, except postcode. Ensure shipping rates show up for the correct country.
7. Ensure you can check out and it charges the correct amount of shipping.

#### Developer only testing/happy path

This test covers only using locale to update the checkout fields. 

1. Disable the checkout field editor plugin from test 1.
2. Add the following snippet.

```php
// Make postcode optional in checkout fields.
add_filter( 'woocommerce_get_country_locale', 'make_postcode_optional' );
function make_postcode_optional( $locales ) {
	foreach( $locales as $country => $locale ) {
		$locales[$country]['postcode']['required'] = false;
	}
    return $locales;
}

add_filter( 'woocommerce_get_country_locale_default', 'make_postcode_optional_default' );

function make_postcode_optional_default( $locale ) {
	$locale['postcode']['required'] = false;
	$locale['postcode']['hidden'] = false;
    return $locale;
}
```
3. Go to WC -> Settings -> Shipping -> Shipping Settings. Enable "Hide shipping costs until an address is entered"
4. Add an item to your cart and go to the shortcode checkout.
5. Fill in your details, except postcode. Ensure shipping rates show up for the correct country.
6. Ensure you can check out and it charges the correct amount of shipping.

#### Other testing

- Test with "Hide shipping costs until an address is entered" disabled - rates should appear as soon as a country is selected.
- Test with virtual products.
- Test with all shipping methods disabled (checkout should not be possible)
- Test with NO shipping methods at all (all deleted), checkout should be possible.
- Test in the blocks checkout with various different zones, different local pickup setting, and "Hide shipping costs until an address is entered" on and off.

All of the above was tested already, Mike, Nadir, and I spent sync time together working through the problem before settling on the approach here of separating blocks and shortcode logic.

### Other notes

The below was part of the original testing instructions, however we removed the code responsible for blocking orders without posted shipping. The risk of blocking legit orders was too great and we would prefer to allow orders to be placed in this edge case without shipping, rather than blocking them altogether.

Note that this edge case occurs even as far back as WC 7.0, I didn't test earlier versions, but this clearly isn't a common use case pattern. This edge case would occur when editing the checkout fields incorrectly, and _not_ updating the fields on the front-end to not be required. When fields are marked as required, and without values, the new address values are not sent to the server.

For a merchant to face this, they would have to be content with an "optional" field still being presented as required on the front end.

##### Edge case 1

This edge case occurs when plugins "do it wrong" by updating the address fields through `woocommerce_shipping_fields` and `woocommerce_billing_fields`  _without_ updating locale correctly and without removing the classes that style a field as required on the front-end.

The postcode field will be made optional by the snippet below, but it will still have classes that make it required on the front end. Because it is still "required" the partial address updates will not be pushed to the server, meaning shipping rates will not update.

I have put a mechanism in place to stop orders going through with no shipping, but it doesn't really give shoppers a way to remedy it, unless they realise they need to enter a postcode.

Given this is an edge case, and a result of plugins "doing it wrong" I don't think it is worth spending too much time trying to work around this.

1. Disable the checkout field editor plugin from test 1.
2. Add the following snippet
```php
add_filter( 'woocommerce_checkout_fields', function( $fields ) {
	$fields['billing']['billing_postcode']['required'] = false;
	$fields['shipping']['shipping_postcode']['required'] = false;
	return $fields;
} );
add_filter( 'woocommerce_shipping_fields', function( $fields ) {
	$fields['shipping_postcode']['required'] = false;
	return $fields;
} );
add_filter( 'woocommerce_billing_fields', function( $fields ) {
	$fields['billing_postcode']['required'] = false;
	return $fields;
} );
```

3. Go to WC -> Settings -> Shipping -> Shipping Settings. Enable "Hide shipping costs until an address is entered"
4. Add an item to the cart and go to the shortcode checkout.
5. Fill in all fields except postcode.
7. Notice that, despite postcode looking like it is required, you can still check out without it.
8. Ensure that an error appears preventing checkout.
9. Reload the page.
10. Ensure rates appear and you can check out.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/fix-shortcode-shipping-release-branch
+++ b/plugins/woocommerce/changelog/fix-shortcode-shipping-release-branch
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Prevent shipping rates remaining hidden when checkout fields filtered to be optional are not filled
--- a/plugins/woocommerce/changelog/fix-shortcode-shipping-release-branch
+++ b/plugins/woocommerce/changelog/fix-shortcode-shipping-release-branch
@@ -1,4 +1,4 @@
 Significance: patch
 Type: fix
 
-Prevent shipping rates remaining hidden when checkout fields filtered to be optional are not filled
+Prevent shipping rates remaining hidden when checkout fields filtered to be optional are not filled on classic checkout.
--- a/plugins/woocommerce/includes/class-wc-cart.php
+++ b/plugins/woocommerce/includes/class-wc-cart.php
@@ -26,9 +26,9 @@ require_once WC_ABSPATH . 'includes/class-wc-cart-session.php';
 class WC_Cart extends WC_Legacy_Cart {
 
 	/**
-	 * Cart context.
+	 * Cart context, used to determine if the cart is being used in a StoreAPI or shortcode context.
 	 *
-	 * @var string
+	 * @var string shortcode|store-api
 	 */
 	public $cart_context = 'shortcode';
 
--- a/plugins/woocommerce/includes/class-wc-cart.php
+++ b/plugins/woocommerce/includes/class-wc-cart.php
@@ -26,6 +26,13 @@ require_once WC_ABSPATH . 'includes/class-wc-cart-session.php';
 class WC_Cart extends WC_Legacy_Cart {
 
 	/**
+	 * Cart context.
+	 *
+	 * @var string
+	 */
+	public $cart_context = 'shortcode';
+
+	/**
 	 * Contains an array of cart items.
 	 *
 	 * @var array
--- a/plugins/woocommerce/includes/class-wc-cart.php
+++ b/plugins/woocommerce/includes/class-wc-cart.php
@@ -1653,7 +1653,9 @@ class WC_Cart extends WC_Legacy_Cart {
 				 */
 				$state_enabled  = apply_filters( 'woocommerce_shipping_calculator_enable_state', true );
 				$state_required = isset( $country_fields['shipping_state'] ) && $country_fields['shipping_state']['required'];
-				if ( $state_enabled && $state_required && ! $this->get_customer()->get_shipping_state() && isset( $checkout_fields['shipping']['shipping_state'] ) ) {
+				// Takes care of late unsetting of checkout fields via hooks (woocommerce_checkout_fields, woocommerce_shipping_fields).
+				$checkout_state_field_exists = isset( $checkout_fields['shipping']['shipping_state'] );
+				if ( $state_enabled && $state_required && ! $this->get_customer()->get_shipping_state() && $checkout_state_field_exists ) {
 					return false;
 				}
 				/**
@@ -1666,7 +1668,9 @@ class WC_Cart extends WC_Legacy_Cart {
 				 */
 				$postcode_enabled  = apply_filters( 'woocommerce_shipping_calculator_enable_postcode', true );
 				$postcode_required = isset( $country_fields['shipping_postcode'] ) && $country_fields['shipping_postcode']['required'];
-				if ( $postcode_enabled && $postcode_required && '' === $this->get_customer()->get_shipping_postcode() && isset( $checkout_fields['shipping']['shipping_postcode'] ) ) {
+				// Takes care of late unsetting of checkout fields via hooks (woocommerce_checkout_fields, woocommerce_shipping_fields).
+				$checkout_postcode_field_exists = isset( $checkout_fields['shipping']['shipping_postcode'] );
+				if ( $postcode_enabled && $postcode_required && '' === $this->get_customer()->get_shipping_postcode() && $checkout_postcode_field_exists ) {
 					return false;
 				}
 			}
--- a/plugins/woocommerce/includes/class-wc-cart.php
+++ b/plugins/woocommerce/includes/class-wc-cart.php
@@ -1628,10 +1628,44 @@ class WC_Cart extends WC_Legacy_Cart {
 		}
 
 		if ( 'yes' === get_option( 'woocommerce_shipping_cost_requires_address' ) ) {
-			$customer = $this->get_customer();
+			if ( 'store-api' === $this->cart_context ) {
+				$customer = $this->get_customer();
 
-			if ( ! $customer instanceof \WC_Customer || ! $customer->has_full_shipping_address() ) {
-				return false;
+				if ( ! $customer instanceof \WC_Customer || ! $customer->has_full_shipping_address() ) {
+					return false;
+				}
+			} else {
+				$country = $this->get_customer()->get_shipping_country();
+				if ( ! $country ) {
+					return false;
+				}
+				$country_fields = WC()->countries->get_address_fields( $country, 'shipping_' );
+				/**
+				 * Filter to not require shipping state for shipping calculation, even if it is required at checkout.
+				 * This can be used to allow shipping calculations to be done without a state.
+				 *
+				 * @since 8.4.0
+				 *
+				 * @param bool $show_state Whether to use the state field. Default true.
+				 */
+				$state_enabled  = apply_filters( 'woocommerce_shipping_calculator_enable_state', true );
+				$state_required = isset( $country_fields['shipping_state'] ) && $country_fields['shipping_state']['required'];
+				if ( $state_enabled && $state_required && ! $this->get_customer()->get_shipping_state() ) {
+					return false;
+				}
+				/**
+				 * Filter to not require shipping postcode for shipping calculation, even if it is required at checkout.
+				 * This can be used to allow shipping calculations to be done without a postcode.
+				 *
+				 * @since 8.4.0
+				 *
+				 * @param bool $show_postcode Whether to use the postcode field. Default true.
+				 */
+				$postcode_enabled  = apply_filters( 'woocommerce_shipping_calculator_enable_postcode', true );
+				$postcode_required = isset( $country_fields['shipping_postcode'] ) && $country_fields['shipping_postcode']['required'];
+				if ( $postcode_enabled && $postcode_required && '' === $this->get_customer()->get_shipping_postcode() ) {
+					return false;
+				}
 			}
 		}
 
--- a/plugins/woocommerce/includes/class-wc-cart.php
+++ b/plugins/woocommerce/includes/class-wc-cart.php
@@ -31,7 +31,7 @@ class WC_Cart extends WC_Legacy_Cart {
 	 *
 	 * @var string shortcode|store-api
 	 */
-	public $_cart_context = 'shortcode';
+	public $cart_context = 'shortcode';
 
 	/**
 	 * Contains an array of cart items.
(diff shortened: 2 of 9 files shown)


plugins/woocommerce/includes/class-wc-cart.php
```diff
@@ -25,6 +25,13 @@
  */
 class WC_Cart extends WC_Legacy_Cart {
 
+	/**
+	 * Cart context, used to determine if the cart is being used in a StoreAPI or shortcode context.
+	 *
+	 * @var string shortcode|store-api
+	 */
+	public $cart_context = 'shortcode';
```
opr, line 34:
```suggestion
	public $_cart_context = 'shortcode';
```

plugins/woocommerce/includes/class-wc-checkout.php
```diff
@@ -835,6 +835,20 @@ public function get_posted_data() {
 	 * @param  WP_Error $errors Validation error.
 	 */
 	protected function validate_posted_data( &$data, &$errors ) {
+
```
opr: Move to validate_checkout

plugins/woocommerce/includes/class-wc-checkout.php
```diff
@@ -835,6 +835,20 @@ public function get_posted_data() {
...
+		// Verify shipping was posted if needed.
+		if ( WC()->cart->needs_shipping() && WC()->cart->show_shipping() ) {
+			if ( empty( $data['shipping_method'] ) ) {
```
opr: Merge ifs, remove show_shipping

plugins/woocommerce/tests/php/includes/class-wc-cart-test.php
```diff
@@ -232,8 +232,99 @@ public function test_show_shipping() {
...
+		WC()->cart->get_customer()->set_shipping_city( '' );
+		$this->assertTrue( WC()->cart->show_shipping() );
+		WC()->cart->get_customer()->set_shipping_postcode( '12345' );
+		WC()->cart->get_customer()->set_shipping_city( 'San Francisco' );
+		$this->assertTrue( WC()->cart->show_shipping() );
+
+		remove_filter( 'woocommerce_checkout_fields', 'remove_unwanted_fields_from_checkout_page' );
+
 		update_option( 'woocommerce_shipping_cost_requires_address', $default_shipping_cost_requires_address );
+		add_filter( 'woocommerce_get_country_locale', 'make_locale_postcode_optional' );
```
opr: Remove

plugins/woocommerce/tests/php/includes/class-wc-checkout-test.php
```diff
@@ -70,6 +70,68 @@ function () {
 		);
 	}
 
+	/**
+	 * @testdox 'validate_posted_data' adds errors for non-posted shipping method.
```
opr: update

plugins/woocommerce/includes/class-wc-cart.php
opr: prefixing with _ is deprecated

plugins/woocommerce/changelog/fix-shortcode-shipping-release-branch
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Prevent shipping rates remaining hidden when checkout fields filtered to be optional are not filled
```
mikejolley:
```suggestion
Prevent shipping rates remaining hidden when checkout fields filtered to be optional are not filled on classic checkout.
```

plugins/woocommerce/includes/class-wc-cart.php
```diff
@@ -1621,10 +1629,50 @@ public function show_shipping() {
...
+				 *
+				 * @param bool $show_postcode Whether to use the postcode field. Default true.
+				 */
+				$postcode_enabled  = apply_filters( 'woocommerce_shipping_calculator_enable_postcode', true );
+				$postcode_required = isset( $country_fields['shipping_postcode'] ) && $country_fields['shipping_postcode']['required'];
+				// Takes care of late unsetting of checkout fields via hooks (woocommerce_checkout_fields, woocommerce_shipping_fields).
+				$checkout_postcode_field_exists = isset( $checkout_fields['shipping']['shipping_postcode'] );
+				if ( $postcode_enabled && $postcode_required && '' === $this->get_customer()->get_shipping_postcode() && $checkout_postcode_field_exists ) {
+					return false;
+				}
```
mikejolley, line 1675: 👍🏻

plugins/woocommerce/includes/class-wc-customer.php
```diff
@@ -273,55 +273,26 @@ public function has_shipping_address() {
...
+	 * This method uses the current country's locale to determine if a field is required, or falls back to the default
+	 * locale if there's no country-specific setting for that field.
+	 *
+	 * This method is only used internally by StoreAPI, and not by the classic/shortcode checkout.
 	 *
 	 * @since 9.8.0
-	 * @return bool Whether the customer has a full shipping address (address_1, city, state, postcode, country).
-	 * Only required fields are checked.
+	 * @return bool Whether the customer has a full shipping address (city, state, postcode, country).
+	 * Only required (according to locale) fields are checked.
```
mikejolley:
```suggestion
	 * Only required fields are checked based on the country locale.
```

mikejolley (approved):
- Was the pnpm lock change intentional? You can use `pnpm install --frozen-lockfile` to leave it alone when building from the release branch.
- This PR needs syncing with the release branch

**Developer only testing/happy path**
- Step 3 requires local pickup to be disabled otherwise you can't toggle the setting.
- Tested the snippets and some variations of settings including forcing to ship to billing address
- Edge case 1 no longer needed (order goes through)
- Tested all shipping methods disabled and confirmed "No shipping method has been selected. Please double check your address, or contact us if you need any help." with and without edge case snippets in place.
- As above with methods enabled in everywhere zone, but none in active zone
- Tested successful checkout via Blocks with snippets in place

I think this does its job and ensures the `has_full_shipping_address` logic is confined to blocks, while restoring shortcode to how it was previously. Its sad we need to include both paths separately, and that its so easy for someone to break the experience through hooks so it might be good to log a future task to deprecate and/or consolidate some of the logic but for now I'm happy for this to proceed. 

Added non-blocking feedback so approving.

opr:
> Was the pnpm lock change intentional? You can use pnpm install --frozen-lockfile to leave it alone when building from the release branch.

No, removed, thanks for catching.

> This PR needs syncing with the release branch

Thanks, done!

> Step 3 requires local pickup to be disabled otherwise you can't toggle the setting.

Updated

> Edge case 1 no longer needed (order goes through)

Moved to "other notes"

> it might be good to log a future task to deprecate and/or consolidate some of the logic but for now I'm happy for this to proceed.

I'll add this as a follow up.

jorgeatorres:
@opr: CI should be passing now if you rebase. If this doesn't need additional reviews could you please merge (after you rebase)? Or if this is indeed waiting for another approval, can we try to get it by EOD today? Thanks!

opr:
@jorgeatorres thanks, rebased. I have asked Tom to review.

plugins/woocommerce/includes/class-wc-cart.php
```diff
@@ -1621,10 +1629,50 @@ public function show_shipping() {
...
+				 *
+				 * @since 8.4.0
+				 *
+				 * @param bool $show_state Whether to use the state field. Default true.
+				 */
+				$state_enabled  = apply_filters( 'woocommerce_shipping_calculator_enable_state', true );
+				$state_required = isset( $country_fields['shipping_state'] ) && $country_fields['shipping_state']['required'];
+				// Takes care of late unsetting of checkout fields via hooks (woocommerce_checkout_fields, woocommerce_shipping_fields).
+				$checkout_state_field_exists = isset( $checkout_fields['shipping']['shipping_state'] );
+				if ( $state_enabled && $state_required && ! $this->get_customer()->get_shipping_state() && $checkout_state_field_exists ) {
```
tjcafferkey, line 1658: Just noting here that we check:

`! $this->get_customer()->get_shipping_state()`

But down where we check `get_shipping_postcode()` we do a different check:

`'' === $this->get_customer()->get_shipping_postcode()`

Not saying that there's anything wrong with it but I noticed it because it feels inconsistent.

tjcafferkey (approved):
I've followed the testing steps and it does work as expected. I'm aware this is a fragile area, and my experience in it is limited so it's difficult to be in the mindset of "_how might this break_". That said, looks like @mikejolley has approved and gone beyond the testing instructions which is good, @senadir has also tested it and if you're also confident in this change then I am happy.

plugins/woocommerce/includes/class-wc-cart.php
opr: That is because a postcode of `0` would pass. PR introducing this: https://github.com/woocommerce/woocommerce/pull/51623

senadir (approved):
I have not retested this beyond our initial review but the changes has been minimal. Thank you for working on this Thomas

PR #57675, opened 2025-05-01
Fix race conditions when quickly switching between delivery options

### LINKED ISSUE #57542: Incorrect rate when toggling between shipping and local pickup quickly
## **Describe the bug**

When toggling between shipping and local pickup quickly the incorrect shipping rate can be selected.

## **To Reproduce**

1. Go to WC -> Settings -> Shipping -> Local pickup.
2. Enable Local Pickup and set up a location.
3. Ensure you have valid shipping methods set up.
4. Add an item to your cart and go to the Checkout block.
5. Switch between Local Pickup and Shipping quickly and notice that sometimes the incorrect rate is chosen, e.g. "Flat rate" will show while Local Pickup is selected.

## **Expected behavior**

The correct rate should be shown

## **Screenshots**

[Local pickup bug](https://www.loom.com/share/3967a1f782cd4e51b1c0d54ecd0cc7db?sid=6e74d9af-4459-448d-9c5d-ba56800ffa53)

## Environment

Please fill out any of the below that applies

**Operating System/Device**

- [ ] Windows
- [ ] MacOS
- [ ] Linux
- [ ] iPhone
- [ ] Samsung
- [ ] Other (add in additional context field)

**Version:** 

**Browser**

- [ ] Desktop Edge
- [ ] Desktop Chrome/Chromium
- [ ] Desktop Safari
- [ ] Mobile Edge
- [ ] Mobile Chrome/Chromium
- [ ] Mobile Safari
- [ ] Other (add in additional context field)

**Version:** 

## WordPress Environment

<details><summary>System Status report</summary>

</details>

straku:

Closes #57542

Every time the new delivery option is selected, we're doing a server call through `selectShippingRate` thunk present in the cart data store. The problem is, even though we properly abort the async operation within `selectShippingRate` thunk (see [this abort controller](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/data/cart/thunks.ts#L418)), we don't call it consistently every time we switch the tab between delivery and pickup options. 

My fix resolves around:
- calling it consistently on component mount and letting thunk handle conflicts with the mentioned abort controller
- make sure it is consistently called every time a new delivery option is picked from the list (not changing tabs, but picking some specific option from the list)

I also simplified the `LocalPickupSelect` component API slightly to make it easier to follow. Instead of passing functions that are leaking implementation details (`setSelectedOption`, `onSelectRate`) I've opted it to a generic `onChange` handler.

1. Go to WC → Settings → Shipping → Local pickup.
2. Enable Local Pickup and set up a location.
3. Ensure you have valid shipping methods set up.
4. Add an item to your cart and go to the Checkout block.
5. Switch between Local Pickup and Shipping quickly and notice that sometimes the incorrect rate is chosen. Some additional notes here:
    - suggest watching the video attached to the issue before proceeding with testing
    - what you need to do essentially is be on the "ship" tab and then quickly switch to "pickup" and back to "ship" again (or pickup → ship → pickup), without the fix this would cause the pickup rate being selected even though you are on the ship tab after performing these steps
    - with "Slow 4G" network throttling it was easier to reproduce the bug, so you might opt it for testing with some throttling as well
    - you can inspect the network tab and verify whether the API calls for selecting shipping rate are properly cancelled

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/shipping-rates-control-package/package-rates.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/shipping-rates-control-package/package-rates.tsx
@@ -34,13 +34,13 @@ const PackageRates = ( {
 	disabled = false,
 	highlightChecked = false,
 }: PackageRates ): JSX.Element => {
-	const selectedRateId = selectedRate?.rate_id || '';
+	const selectedRateId = selectedRate?.rate_id;
 	const previousSelectedRateId = usePrevious( selectedRateId );
 
 	// Store selected rate ID in local state so shipping rates changes are shown in the UI instantly.
-	const [ selectedOption, setSelectedOption ] = useState(
-		selectedRateId ?? ''
-	);
+	const [ selectedOption, setSelectedOption ] = useState<
+		string | undefined
+	>( selectedRateId ?? rates[ 0 ]?.rate_id );
 
 	// Update the selected option if cart state changes in the data store.
 	useEffect( () => {
@@ -54,14 +54,13 @@ const PackageRates = ( {
 	}, [ selectedRateId, selectedOption, previousSelectedRateId ] );
 
 	// Update on mount, we do it every time to:
-	// - set the initial value if selectedOption not set
+	// - sync the initial value with the server
 	// - or reset pending request to change shipping rate that might be coming
 	//   from other components (e.g. local pickup), selectShippingRate thunk in
 	//   the cart store properly handles aborting the previous request if needed
 	useEffect( () => {
-		if ( rates.length > 0 ) {
-			setSelectedOption( selectedOption || rates[ 0 ].rate_id );
-			onSelectRate( selectedOption || rates[ 0 ].rate_id );
+		if ( selectedOption ) {
+			onSelectRate( selectedOption );
 		}
 		// We want this to run on mount only, beware of updating it as it may cause
 		// shipping rate selection to end up in inifite loop
@@ -81,7 +80,7 @@ const PackageRates = ( {
 			} }
 			highlightChecked={ highlightChecked }
 			disabled={ disabled }
-			selected={ selectedOption }
+			selected={ selectedOption ?? '' }
 			options={ rates.map( renderOption ) }
 			descriptionStackingDirection="column"
 		/>
--- a/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/shipping-rates-control-package/package-rates.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/shipping-rates-control-package/package-rates.tsx
@@ -42,17 +42,6 @@ const PackageRates = ( {
 		string | undefined
 	>( selectedRateId ?? rates[ 0 ]?.rate_id );
 
-	// Update the selected option if cart state changes in the data store.
-	useEffect( () => {
-		if (
-			selectedRateId &&
-			selectedRateId !== previousSelectedRateId &&
-			selectedRateId !== selectedOption
-		) {
-			setSelectedOption( selectedRateId );
-		}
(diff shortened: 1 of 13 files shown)


plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/shipping-rates-control-package/package-rates.tsx
```diff
@@ -53,13 +53,19 @@ const PackageRates = ( {
 		}
 	}, [ selectedRateId, selectedOption, previousSelectedRateId ] );
```
straku: Apply something similar to the other component to account for changes in data store.

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/shipping-rates-control-package/package-rates.tsx
```diff
@@ -67,6 +56,17 @@ const PackageRates = ( {
...
+	useEffect( () => {
+		if (
+			selectedRateId &&
+			selectedRateId !== previousSelectedRateId &&
+			selectedRateId !== selectedOption
+		) {
+			setSelectedOption( selectedRateId );
+		}
+	}, [ selectedRateId, selectedOption, previousSelectedRateId ] );
+
```
straku, line 69: Moved it for the sake of clarity. It works better in my head to have `useEffect` that runs only on mount to be above this one.

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-pickup-options-block/block.tsx
```diff
@@ -175,6 +175,16 @@ const Block = () => {
...
+	// Update the selected option if cart state changes in the data store.
+	useEffect( () => {
+		const selectedRate = pickupLocations.find( ( rate ) => rate.selected );
+		const selectedRateId = selectedRate?.rate_id;
+
+		if ( selectedRateId && selectedRateId !== selectedOption ) {
+			setSelectedOption( selectedRateId );
+		}
+	}, [ pickupLocations, selectedOption ] );
+
```
straku, line 187: Not tested yet

straku:
@opr It took a while, but the tests are there 😌

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-pickup-options-block/test/block.tsx
```diff
@@ -0,0 +1,216 @@
...
+			selectShippingRate: jest.fn(),
+			isSelectingRate: false,
+			shippingRates: [ testPackageData ],
+		};
+	} );
+
+	render( <CheckoutPickupOptionsBlock /> );
+
+	// Find the radio buttons directly using screen queries
+	// The baseElement option allows us to search within portals
```
opr: Should this `baseElement` comment be here, doesn't look like we're using that

opr (approved):
Thanks @straku I tested this pretty extensively with different network throttle speeds and it worked each time! Appreciate you picking this up and the tests look great too 👏🏼  Just one comment about a comment but this can be merged.

Thanks!

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-pickup-options-block/test/block.tsx
straku: It's a leftover from my pairing session with Claude (it wasn't too helpful in this case), let me remove it. Thanks for spotting!

PR #57683, opened 2025-05-02
Improve the spacing in the header area

oaratovskyi:

Improve the spacing in the header area that was found in the design QA.

Closes WOOPLUG-4111.

1. Spin up new JN sire from this branch
2. Visit Settings -> Emails. Check that spacing around Email notifications header was fixed.
image
3. Navigate to Settings -> Shipping -> Add Zone. Check that spacing around Add Zone / Edit zone was fixed.
image

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/legacy/css/admin.scss
```diff
@@ -4160,6 +4160,7 @@ table.wc_shipping {
 }
 
 .wc-shipping-zone-settings {
+	margin-top: 8px !important;
```
elazzabi: I usually try to avoid `!important` so just flagging it here if there is a better way to do it, if not, then we can go with it.

elazzabi (approved):
Changes look good, I left a non-blocking comment

plugins/woocommerce/client/legacy/css/admin.scss
dmallory42: I had the same feeling, usually we can get around it by being more specific when targeting the selector or there's some other rule somewhere which is overriding it

mordeth: I suggest using the `$gap-smaller` instead.

plugins/woocommerce/client/legacy/css/admin.scss
```diff
@@ -5876,6 +5877,9 @@ img.help_tip {
 	.wc-settings-row-woocommerce_custom_orders_table_data_sync_enabled td.forminp {
 		padding-top: 0;
 	}
+	#email_notification_settings-description {
+		margin-top: 16px;
```
mordeth, line 5883:
```suggestion
		margin-top: $gap;
```

oaratovskyi: Unfortunately, it's not supported in this file. The app crashes when I try to build it.

plugins/woocommerce/client/legacy/css/admin.scss
oaratovskyi: I managed to avoid `!important`, thanks for noticing! 
using gap vars in this file seems to be not possible, it's not building with it. Maybe that's why no similar vars like this exist

Issue #56288, opened 2025-03-11
Blockified Add to Cart with Options block: make product context available when editing the template parts


Aljullu:
Issue highlighted by @mdperez86.

When editing the _Add to Cart with Options_ template parts directly in the Site Editor, the product context doesn't match the product type being edited, so some of the inner blocks aren't rendered.

## Steps to reproduce

0. Install and activate the WooCommerce Beta Tester plugin
0. Go to _WooCommerce Admin Test Helper_ > _Features_ and enable `experimental-blocks` and `blockified-add-to-cart`.
1. Go to Appearance > Editor > Patterns > Add to Cart with Options.
2. Edit the _Grouped Product Add to Cart with Options_ template part, for example.
3. Notice the blocks specific to grouped products are not rendered in the editor. That's because the product context is not a grouped product.

![Image](https://github.com/user-attachments/assets/d5bdfc1e-138c-41e9-b6ff-d65f6522f249)

### Expected result

Blocks specific to grouped products should be visible. Like the Grouped Product Selector block.

PR #57690, opened 2025-05-02
Blockified Add to Cart with Options: Add grouped product context

mikachan:

This adds fallback product context to the Blockified Add to Cart with Options block so that the inner blocks render correctly when no product context is provided, such as being opened in the Editor in the Grouped Product Add to Cart with Options template part.

The product context is fetched from an existing grouped product, and if one doesn't exist, any other 3 products are used in the block.

Part of https://github.com/woocommerce/woocommerce/issues/56288.

| Before | After |
| ------ | ----- |
| image | image |

0. Install and activate the WooCommerce Beta Tester plugin
1. Go to Appearance > Editor > Patterns > Add to Cart with Options.
2. Edit the _Grouped Product Add to Cart with Options_ template part.
3. Make sure the blocks specific to grouped products are rendered in the Editor if there are grouped products available. If no grouped products are available, a warning message should show instead.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Add grouped product context to Blockified Add to Cart with Options when block is rendered in the Editor outside of a product.

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/edit.tsx
```diff
@@ -19,21 +24,49 @@ export default function AddToCartWithOptionsGroupedProductSelectorEdit(
 	props: BlockEditProps< Attributes >
 ) {
 	const { className } = props.attributes;
-	const { current: currentProductType } = useProductTypeSelector();
 	const { product } = useProductDataContext();
-	const productType =
-		product.id === 0 ? currentProductType?.slug : product.type;
-
```
mikachan, line 26: `currentProductType` was only being used when `product.id` was `0`, which is the same logic that's used to render the fallback context. I believe `currentProductType` will always be `simple` when no product context is provided, so I decided to remove it.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/edit.tsx
```diff
@@ -19,21 +24,49 @@ export default function AddToCartWithOptionsGroupedProductSelectorEdit(
...
+	// Fetch an existing grouped product so template can be edited.
+	useEffect( () => {
+		if ( product.id !== 0 && product.type === 'grouped' ) {
+			setGroupedProduct( [ product as ProductResponseItem ] );
+		}
+		if ( groupedProduct.length === 0 && product.id === 0 ) {
+			resolveSelect( productsStore )
+				.getProducts( {
+					type: 'grouped',
+					per_page: 3,
```
Aljullu: I'm not sure to understand this `per_page: 3`, we only one product here, no?

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/edit.tsx
```diff
@@ -19,21 +24,49 @@ export default function AddToCartWithOptionsGroupedProductSelectorEdit(
...
+			<div { ...innerBlocksProps }>
+				<NoticeBanner status="warning">
+					{ __(
+						'No grouped products were found. Please create a grouped product first.',
+						'woocommerce'
+					) }
+				</NoticeBanner>
+			</div>
+		);
 	}
```
Aljullu: If no Grouped Products are found, I think it would be better to fall back to displaying 3 random products in the _Grouped Product Selector_ block, what do you think?

Aljullu (commented):
Thanks for investigating this issue, @mikachan! Really nice to see the _Grouped Product Selector_ block being populated when editing the template part. :slightly_smiling_face: 

> I'm not sure if this is the best approach, perhaps we should provide demo/example content instead of fetching real product data. Would appreciate any other thoughts on how to allow these blocks to render without a product 🙇 

Right... Unfortunately, we don't have clear guidelines on this, but they way I would do is:

1. When we know the parent product, try to get real child products or show a notice if there are no children.
2. When we don't know the parent product, use sample data. That could be existing products in the store or simply hard-coded sample data.

How does it sound?

> Closes https://github.com/woocommerce/woocommerce/issues/56288.

Heads-up that we will need to solve the same issue for variable products. I'm happy to work on it in different PRs to keep them smaller. But just mentioning because #56288 is the generic issue for all product types, so this PR won't solve it completely. :smile: 

---

I noticed that clicking on the inner blocks is a bit buggy in this PR. You can reproduce by creating a new post or page, adding the _Single Product_ block, and selecting a grouped product:


https://github.com/user-attachments/assets/66020a87-e80e-45ec-b963-2a1bb8a920ca

I also left a couple of comments inline: :point_down:

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/edit.tsx
mikachan: Yes I think we only want 1, but I noticed we already used `per_page: 3` here for the template, so I was copying that logic: https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-template/edit.tsx#L97. But it might be best to just specify 1!

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/edit.tsx
mikachan: Yes that sounds good, I'll try that. I think that's nicer than a warning message. Thanks!

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/edit.tsx
mikachan: Ah I think the query for 3 products is for the child products, not the parent. I've updated this to `per_page: 1` in https://github.com/woocommerce/woocommerce/pull/57690/commits/6c9d3f70e36b6f065cca7b60a4ccba171d2c7759.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/edit.tsx
mikachan: I've removed the `NoticeBanner` and instead changed this to fallback to displaying 3 products when a grouped product isn't available: https://github.com/woocommerce/woocommerce/pull/57690/commits/f0c2fed8835fe47c1d4c06736c67d5a8662dab7f

mikachan:
> I noticed that clicking on the inner blocks is a bit buggy in this PR. You can reproduce by creating a new post or page, adding the Single Product block, and selecting a grouped product:

It looks like this was because I'd used `getProducts` rather than `getEntityRecords` to fetch the grouped products. I think this was returning the inner blocks for the product info incorrectly. This seems to be fixed in https://github.com/woocommerce/woocommerce/pull/57690/commits/184292cb1fd26741041e188870514e9162677389.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/edit.tsx
```diff
@@ -19,20 +22,50 @@ export default function AddToCartWithOptionsGroupedProductSelectorEdit(
...
+							} )
+							.then( ( fetchedProduct ) => {
+								if ( fetchedProduct.length > 0 ) {
+									setGroupedProduct( fetchedProduct );
+								}
+							} );
+					}
+				} );
+		}
+	}, [ groupedProduct, product ] );
```
Aljullu: I'm not sure to understand what is `groupedProduct` used for? Apparently it's only used in the `if` below, but it's not clear to me what's the benefit of the change. :thinking:

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-template/edit.tsx
```diff
@@ -87,32 +88,67 @@ export default function ProductItemTemplateEdit(
...
-			const { getEntityRecords } = select( coreStore );
+	useEffect( () => {
+		const fetchGroupedProducts = async (
+			productContext: ProductResponseItem[]
+		) => {
+			if ( ! productContext || productContext.length === 0 ) {
+				return;
+			}
+
+			let query: Record< string, unknown > = { per_page: 3 };
```
Aljullu: I know this was like that in `trunk`, but wondering if we could make this `per_page` match `productContext[ 0 ].grouped_products` when it exists? Maybe setting a limit like 10 or 30 so it doesn't impact the editor's performance too much.

Aljullu (commented):
Thanks for the updates, @mikachan! Overall this is looking good and testing as described. :ok_hand: I only left a couple of questions inline.

> It looks like this was because I'd used `getProducts` rather than `getEntityRecords` to fetch the grouped products. I think this was returning the inner blocks for the product info incorrectly. This seems to be fixed in https://github.com/woocommerce/woocommerce/pull/57690/commits/184292cb1fd26741041e188870514e9162677389.

Confirming that's fixed, thanks!

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/edit.tsx
mikachan: I added this `groupedProduct` to match the product context in `ProductItemTemplateEdit`. Both of these are used to set the context based on whether an appropriate grouped product has been provided, and if not, look for an existing grouped product or fallback to any other type of product.

I thought it was best to keep these contexts the same, as if there is no `groupedProduct`, the inner blocks will not be rendered. 

Let me know if you think there's a better way to do this.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-template/edit.tsx
mikachan: Nice, good shout. I've set `per_page` to the number of grouped products found in the provided context, if grouped products are found, otherwise it still defaults to 3: https://github.com/woocommerce/woocommerce/pull/57690/commits/3c88f5534d558da8c29ea59b7cc69848c42ae1d3. Does this sound better, or would it be better to increase the fallback to something like 10?

Aljullu: Thanks! I think it should be fine. We can iterate on it later if we receive reports of performance issues on grouped products with many children.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/edit.tsx
Aljullu: It makes sense, but what are the benefits of doing so instead of simply rendering the wrapping `<div>` here and letting the inner blocks (in this case `ProductItemTemplateEdit`) handle whether they should render or not?

I don't have a strong opinion, but I feel that would allow simplifying the logic a lot? :thinking:

mikachan: Oh I see, yes that would be much simpler! I've removed all the product context from the selector in https://github.com/woocommerce/woocommerce/pull/57690/commits/c18e4bdbaa94b2d60c20a89b56ab953a69bf472e. I guess I was trying to maintain the existing logic but with additional product context, but yes, I don't think this is necessary 😄

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-template/edit.tsx
```diff
@@ -87,32 +88,68 @@ export default function ProductItemTemplateEdit(
...
+		null
+	);
 
-	const { products } = useSelect(
-		( select ) => {
-			const postType = 'product';
-			const { getEntityRecords } = select( coreStore );
+	useEffect( () => {
+		const fetchGroupedProducts = async (
+			productContext: ProductResponseItem[]
```
Aljullu: What do you think about this function taking `groupedProductIds` as the parameter instead of `productContext`? I find it a bit confusing right now because `productContext` might contain the grouped product or the child products, and we handle it differently by looking at `productContext.length`.

If we change it, we could even rename the function to `fetchChildProducts()` or something along those lines so it's clearer it's retrieving the child products and not the parent grouped product.

Aljullu (commented):
Thanks for the updates, @mikachan, this is working perfectly! I just left one suggestion about changing the `fetchGroupedProducts()` function signature.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/edit.tsx
Aljullu: :100:

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-template/edit.tsx
mikachan: Great idea, thank you! I've renamed this function to `fetchChildProducts`, and it now takes an array of product IDs, which means they can be passed directly to the query. This looks much simpler 🎉 Let me know if I've missed anything with the latest changes.

mikachan:
With the recent refactor, I noticed that `getProducts` seems to be working fine when used instead of `getEntityRecords` to fetch the child products. Previously, this seemed to be causing the bug mentioned in [this comment](https://github.com/woocommerce/woocommerce/pull/57690#pullrequestreview-2812299197), but I'm no longer able to reproduce this when editing a Single Product block for a grouped product.

@Aljullu, if you see the same, I think it'd be better to stick with `getProducts` (as in the change in https://github.com/woocommerce/woocommerce/pull/57690/commits/839f97f3b53ef9dcdf2745fa5dad610fcc820fda). If you're still seeing issues, I'll swap back to `getEntityRecords`.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-template/edit.tsx
```diff
@@ -87,32 +87,53 @@ export default function ProductItemTemplateEdit(
...
+			}
 
-			let query: Record< string, unknown > = {
-				per_page: 3,
-			};
+			resolveSelect( productsStore )
+				.getProducts( {
+					include: groupedProductIds,
+					per_page: groupedProductIds.length,
+				} )
```
Aljullu: We could query only for the `id` here to save some bytes. :slightly_smiling_face: 

```suggestion
				.getProducts( {
					include: groupedProductIds,
					per_page: groupedProductIds.length,
					_fields: [ 'id' ],
				} )
```

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-template/edit.tsx
```diff
@@ -87,32 +87,53 @@ export default function ProductItemTemplateEdit(
...
+					.getProducts( { type: 'grouped', per_page: 1 } )
+					.then( ( groupedProduct ) => {
+						if ( groupedProduct.length > 0 ) {
+							fetchChildProducts(
+								groupedProduct[ 0 ].grouped_products
+							);
+						} else {
+							// If there are no grouped products, query for any three other products.
+							resolveSelect( productsStore )
+								.getProducts( { per_page: 3 } )
```
Aljullu: Same here:

```suggestion
								.getProducts( {
									per_page: 3,
									_fields: [ 'id' ],
								} )
```

Aljullu (approved):
Great work, @mikachan! This looks good to me. I left a couple of inline suggestions, but pre-approving anyway.

Confirming the Grouped Product Selector block displays properly when editing the template part:

imatge


<details>

* Theme: Twenty Twenty-Five
* Store size: small
</details>

Aljullu:
> With the recent refactor, I noticed that `getProducts` seems to be working fine when used instead of `getEntityRecords` to fetch the child products. Previously, this seemed to be causing the bug mentioned in [this comment](https://github.com/woocommerce/woocommerce/pull/57690#pullrequestreview-2812299197), but I'm no longer able to reproduce this when editing a Single Product block for a grouped product.
> 
> @Aljullu, if you see the same, I think it'd be better to stick with `getProducts` (as in the change in https://github.com/woocommerce/woocommerce/pull/57690/commits/839f97f3b53ef9dcdf2745fa5dad610fcc820fda). If you're still seeing issues, I'll swap back to `getEntityRecords`.

I wasn't able to reproduce the issue anymore, so it looks good to me as-is. Thanks for checking!

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-template/edit.tsx
mikachan: Nice, thanks!

PR #57693, opened 2025-05-02
Show all validation errors in Checkout even if email is invalid

senadir:

in 9.5, we added inline errors that are read from the notice returned in Checkout (https://github.com/woocommerce/woocommerce/pull/49094). However, in Checkout, if you submit an invalid email, validation will get shortcut early and only one error is shown. In this PR, I allow validation to continue and ignore that error, this is because the same field and value will get validated in the next function call.

I also skip a redundant error of "Please enter an address" when other errors are present. 

Closes WOOPLUG-4086

| Before | After |
| ------ | ----- |
|    image |   image |

1. In shortcode Checkout, with all fields empty (including email and country), submit the order.
2. You should see inline errors for all fields.
3. Insert an invalid value for email like "invalid.com" and submit
4. Notice that all fields still have errors with inline error.
5. Enable shipping, but don't have a country selected, submit the order.
6. Notice there is no "Please enter an address to continue".
7. Provide a country, you should not see that error still.
8. Fill all fields, make sure you can submit the order.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Validate all fields and don't bail early on email.

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/57693-fix-show-all-errors-on-checkout
+++ b/plugins/woocommerce/changelog/57693-fix-show-all-errors-on-checkout
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Validate all fields and don't bail early on email.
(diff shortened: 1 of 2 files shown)


mikejolley (approved):
New logic makes sense. I think this works much better. I ran through the steps with a combination of fields populated and spotted no issues.

Do we need test coverage for this flow? Pre-approving.

senadir:
I don't think so, this is a minor UX bug that I don't think deserve bloating our E2E further.

Issue #52311, opened 2024-10-24
Autoloader manifest not updated during WooCommerce upgrade


joshuatf:
### Prerequisites

- [x] I have carried out troubleshooting steps and I believe I have found a bug.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

In https://github.com/woocommerce/woocommerce/issues/51368 a fatal error occurred due to a file that was removed.  This was most likely due to the older WooCommerce version's manifest still being loaded during upgrade while the removed file was no longer present in the newer version being upgraded to.

Even if a file was marked as internal, moving any file could potentially trigger this issue going forward.  

### Expected behavior

Files should be removable without triggering fatal errors.

### Actual behavior

Removing or renaming a file could trigger a fatal error on upgrade.

### Steps to reproduce

1. Build the `trunk` of the current repo
2. Install that version of WooCommerce on your site
3. Remove any autoloaded file under the `WooCommerce/src` directory
4. Remove any references to that file/class so that no error should be thrown
5. Bump the WooCommerce version in `woocommerce.php`
6. Build the plugin
7. Install the new version on your site using the Plugins->Add new->Upload option
8. See the fatal error for the missing class

### WordPress Environment

n/a

### Isolating the problem

- [x] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [x] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/products/storefront/).
- [x] I can reproduce this bug consistently using the steps above.

barryhughes:
Unsure what the best focus label is; it feels like perhaps a Best Practices item for us to document, though possibly some supporting tooling/linting could be added also.

pmcpinto:
It looks like this might be covered by the following work: https://github.com/woocommerce/woocommerce/pull/65361. So, I'll label this as in progress.

PR #57695, opened 2025-05-02
Refactor Create Account Service Class and replace with core customer new account email

### LINKED ISSUE #56766: [Technical Debt] Refactor Create Account Service Class
_[This issue was split out of this spike looking into areas where blocks override core functionality.](https://github.com/woocommerce/woocommerce/issues/55856)_

Ref: https://github.com/woocommerce/woocommerce/blob/c807d2d068125af1607afd22e54c0b13f7f83bc1/plugins/woocommerce/src/Blocks/Domain/Services/CreateAccount.php#L10

When using block checkout, we have an override in place for the new order email which sends a password reset link for new accounts instead of showing a password inline.

We should review the new account email in core and either:

- Remove this service class and handle this logic in the core new account email class
- Make all new account emails behave the same way

mikejolley:

The blocks feature plugin implemented a new email template for customers which included the "set password" link and never showed a plain text password. Since then, the core email is pretty much identical and takes the same approach. This means we can remove the blocks logic and just use the core email.

This addresses some technical debt in the plugin.

Closes #56766

1. Enable delayed account creation
2. Place an order and create a new account.
3. Check that the new account email arrives and contains a working "set password" link

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Blocks/Domain/Services/CreateAccount.php
straku, line 1: Would you mind explaining why we also need to leave public methods of this class in place? I've noticed you're not doing the same for `CustomerNewAccount` and would like to learn the difference.

straku (approved):
Tests well for me, I received a correctly formatted email with working set password link. Asked one question, mostly for my own education purposes.

plugins/woocommerce/src/Blocks/Domain/Services/CreateAccount.php
mikejolley: I left these here to prevent if causing fatal error if someone (I could find no usage) is calling these methods. 

`CustomerNewAccount` extends `WC_Email` which defines all the public methods so there was no reason to keep them around.

I think once https://github.com/woocommerce/woocommerce/issues/52311 is solved we can remove these classes without issue.

straku: Thanks for explaining.

PR #57705, opened 2025-05-04
Fix order statuses array check in OrderAwareControllerTrait

nerrad:

This PR fixes an issue where the `woocommerce_actionable_order_statuses` option
could be set to a non-array value, causing errors in analytics endpoints that
rely on actionable order statuses. The change ensures the value is always an
array, preventing such issues.

Closes #WOOPLUG-4124

1. Set the `woocommerce_actionable_order_statuses` option in the database to a
   non-array value (e.g., a string or integer).
2. Access any analytics endpoint or feature that uses order statuses.
3. Confirm that no errors occur and that the feature works as expected.
4. Restore the option to a valid array and confirm normal operation.

- Manual testing with both valid and invalid option values.
- Verified analytics endpoints do not throw errors with non-array values.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

Fix: Ensure `woocommerce_actionable_order_statuses` is always an array to
prevent errors in analytics endpoints.

</details>

### DIFF
--- a/plugins/woocommerce/changelog/57705-fix-WOOPLUG-4124-order-statuses-array-check
+++ b/plugins/woocommerce/changelog/57705-fix-WOOPLUG-4124-order-statuses-array-check
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+Comment: Fix: Ensure `woocommerce_actionable_order_statuses` is always an array to prevent errors in analytics endpoints.
+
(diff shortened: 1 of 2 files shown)


nerrad:
Had to close and reopen because of the workflows getting in a borked state 😕

louwie17 (approved):
Thanks for this @nerrad, this works for me!
There are a couple other places where the option is being used, but it is handled gracefully so this fix is fine. In terms of the other places an example is the Orders activity panel ( see screenshot ) or when updating the Analytics settings ( although a user can still successfully update the settings, which overwrites the invalid value ).
Screenshot 2025-05-06 at 14 46 58

PR #57713, opened 2025-05-05
Remove the `!important` label from the display rule on the Add to Cart > Quantity selector element

jimjasson:

This PR removes the `!Important` label added to the CSS rule that targets `.wp-block-woocommerce-add-to-cart-form .variations_button .quantity, .wp-block-woocommerce-add-to-cart-form form.cart .quantity` and was added via this PR: https://github.com/woocommerce/woocommerce/pull/57570.

The `important` label is not needed and it conflicts with CSS rules added by other extensions or custom code that hide this element. 

1. The main goal is to ensure that no visual regressions happen to the `quantity` after removing the `!important` label.
2. Set your theme to Storefront, or any other non-blockified theme.
3. Create a new blank page.
4. Add the Single Product block. 
5. Ensure that the quantity selector shows up similarly to [here](https://github.com/woocommerce/woocommerce/pull/57570#pullrequestreview-2802701833).
6. View the product -- ensure that in the frontend the quantity selector shows up similarly to [here](https://github.com/woocommerce/woocommerce/pull/57570#pullrequestreview-2802701833).
7. Back to the editor, click on the Single Product block and select the 'Stepper' Quantity selection option. 
8. Ensure that the quantity selector shows up similarly to [here](https://github.com/woocommerce/woocommerce/pull/57570#pullrequestreview-2806100716).
9. View the product -- ensure that the quantity selector shows up as in the editor.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 
This is a fix for a feature that hasn't been released yet. 

</details>

### DIFF
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/product-elements/add-to-cart-form/style.scss
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/product-elements/add-to-cart-form/style.scss
@@ -28,7 +28,7 @@
 		display: grid;
 
 		.quantity {
-			display: flex !important;
+			display: flex;
 			align-items: stretch;
 		}
 
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/product-elements/add-to-cart-form/style.scss
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/product-elements/add-to-cart-form/style.scss
@@ -28,7 +28,7 @@
 		display: grid;
 
 		.quantity {
-			display: flex;
+			display: inline-flex;
 			align-items: stretch;
 		}
 
@@ -103,7 +103,6 @@ div.wc-block-add-to-cart-form.wc-block-add-to-cart-form--stepper {
 	}
 
 	.wc-block-components-quantity-selector {
-		display: inline-flex;
 
 		.input-text {
 			font-size: var(--wp--preset--font-size--small);
--- a/plugins/woocommerce/client/legacy/css/woocommerce-blocktheme.scss
+++ b/plugins/woocommerce/client/legacy/css/woocommerce-blocktheme.scss
@@ -102,7 +102,6 @@
 
 		form.cart {
 			div.quantity {
-				display: inline-block;
 				float: none; // Remove float set by WC core.
 				vertical-align: middle;
 


Aljullu (commented):
Thanks for picking this up, @jimjasson!

We should have probably done that in the first PR, but I think we will need to remove this line:

https://github.com/woocommerce/woocommerce/blob/08894a09bb88646bda313cf826fcda6a058202a6/plugins/woocommerce/client/legacy/css/woocommerce-blocktheme.scss#L105

Otherwise this is how a simple product looks like in a block theme:

`trunk` | This branch
--- | ---
imatge | imatge

I suspect that's also why `!important` was added in the first place.

---

Composite products are correctly fixed, though:

Before | After
--- | ---
imatge | imatge

And classic themes keep working fine as well:

imatge

So I think it's just a matter of removing that CSS line I mentioned above.

jimjasson:
Thanks for catching that @Aljullu! I thought that the original PR only targeted non-block themes -- after the change you proposed, I see the quantity selector working correctly for both block and non-block themes.

plugins/woocommerce/client/blocks/assets/js/blocks/product-elements/add-to-cart-form/style.scss
```diff
@@ -28,7 +28,7 @@
 		display: grid;
 
 		.quantity {
-			display: flex !important;
+			display: flex;
```
Aljullu: I just noticed this might need to be `inline-flex` instead of `flex`. Otherwise this is how Composite Products look like in block themes:

`flex` | `inline-flex`
--- | ---
imatge | imatge

I tried it and other products would look good as well:

Simple products | Simple products (classic theme)
--- | ---
imatge | imatge

If we do that, I think we might also be able to remove this line:

https://github.com/woocommerce/woocommerce/blob/08894a09bb88646bda313cf826fcda6a058202a6/plugins/woocommerce/client/blocks/assets/js/blocks/product-elements/add-to-cart-form/style.scss#L73

As `inline-flex` would already be the inherited `display`.

Aljullu (commented):
Thanks for the update! I left one extra comment below: :point_down:

plugins/woocommerce/client/blocks/assets/js/blocks/product-elements/add-to-cart-form/style.scss
jimjasson: Fixed that @Aljullu, but I noticed that there are some issues when using the Single Product block in a custom page with Bundles/Composites that also persist in `trunk`, like the quantity amount not being visible inside the quantity selector and the quantity selector being higher than the button:

Markup on 2025-05-05 at 13:38:46


Markup on 2025-05-05 at 13:39:00

These all seem to be coming from this rule: 

Markup on 2025-05-05 at 13:45:43

as disabling it fixes the issue: 

Markup on 2025-05-05 at 13:46:12

but perhaps this is something that should be addressed separately.

Aljullu: Good catch! I'm going to create a separate PR for that.

Aljullu (approved):
LGTM, thanks for the updates.

<details>

* Plugins: WooCommerce Beta Tester, Composite Products, Product Bundles
* Theme: Twenty Twenty-Five, Storefront
* Store size: small
</details>

plugins/woocommerce/client/blocks/assets/js/blocks/product-elements/add-to-cart-form/style.scss
Aljullu: Issue created: https://github.com/woocommerce/woocommerce/issues/57721.

PR #57723, opened 2025-05-05
Fix Add to Cart Form block quantity input size in some extensions

### LINKED ISSUE #57721: Add to Cart quantity input is too small in some extensions like Product Bundles and Composite Products
Issue found by @jason.kytros (see https://github.com/woocommerce/woocommerce/pull/57713#discussion_r2073223725)

> I noticed that there are some issues when using the Single Product block in a custom page with Bundles/Composites that also persist in `trunk`, like the quantity amount not being visible inside the quantity selector and the quantity selector being higher than the button:
> 
> Markup on 2025-05-05 at 13:38:46
> 
> Markup on 2025-05-05 at 13:39:00
> 
> These all seem to be coming from this rule: 
> 
> Markup on 2025-05-05 at 13:45:43
> 
> as disabling it fixes the issue: 
> 
> Markup on 2025-05-05 at 13:46:12
> 
> but perhaps this is something that should be addressed separately.

Aljullu:

Closes https://github.com/woocommerce/woocommerce/issues/57721.
Closes WOOPLUG-4150.

Simple product

Storefront (before) | Storefront (after)
--- | ---
Image | Image

TT5 (before) | TT5 (after)
--- | ---
Image | Image

Simple product in Single Product block

Storefront (before) | Storefront (after)
--- | ---
Image | Image

TT5 (before) | TT5 (after)
--- | ---
Image | Image

Product Bundle

Storefront (before) | Storefront (after)
--- | ---
Image | Image

TT5 (before) | TT5 (after)
--- | ---
Image | Image

Product Bundle in Single Product block

Storefront (before) | Storefront (after)
--- | ---
Image | Image

TT5 (before) | TT5 (after)
--- | ---
Image | Image

1. Create a Composite Product and Product Bundle in your store.
2. In a classic theme, visit their template. Verify the quantity inputs are visible and have enough width.
3. Create a post, add two _Single Product_ blocks and select the products you created in step 1.
4. Visit the post in the frontend and verify the quantity inputs are visible and have enough width.
5. Switch to a block theme and visit the post and templates again. Verify the sizing of the input looks good and there are no regressions.

### DIFF
--- a/plugins/woocommerce/changelog/fix-57721-small-inputs-in-extensions
+++ b/plugins/woocommerce/changelog/fix-57721-small-inputs-in-extensions
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix Add to Cart Form block quantity input size in some extensions
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/product-elements/add-to-cart-form/style.scss
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/product-elements/add-to-cart-form/style.scss
@@ -60,6 +60,7 @@
 		margin-right: 4px;
 		// WooCommerce core styles: https://github.com/woocommerce/woocommerce/blob/c9f62609155825cd817976c7611b9b0415e90f2f/plugins/woocommerce/client/legacy/css/woocommerce.scss/#L111-L112
 		.qty {
+			box-sizing: content-box;
 			width: 3.631em;
 			text-align: center;
 		}
@@ -138,10 +139,6 @@ div.wc-block-add-to-cart-form.wc-block-add-to-cart-form--stepper {
 		.wc-block-components-quantity-selector {
 			width: 107px;
 		}
-	}
-	.input-text.qty.text {
-		// WooCommerce core styles: https://github.com/woocommerce/woocommerce/blob/c9f62609155825cd817976c7611b9b0415e90f2f/plugins/woocommerce/client/legacy/css/woocommerce.scss/#L111-L112
-		width: 3.631em;
 	}
 }
 


jimjasson:
Hey @Aljullu, 

Thanks for looking into this! 

Although the issue with the quantity value being invisible in the single product block when Storefront is used is resolved, I noticed that there are some alignment issues between the quantity selector and the Add to Cart button in the block.

See some examples: 

(1) A Product Bundle/Composite Product with Storefront + standard Add to Cart button:

Markup on 2025-05-06 at 11:45:42

The quantity selector is bigger than the button. 

(2) A Product Bundle/Composite Product with Storefront + Add to Cart button with options:

Markup on 2025-05-06 at 11:47:02

This looks good! ✅

(3) A Product Bundle/Composite Product with TT5 + standard Add to Cart button:

Markup on 2025-05-06 at 11:46:23

The quantity selector is smaller than the button. 

(4) A Product Bundle/Composite Product with Storefront + Add to Cart button with options:

Markup on 2025-05-06 at 11:46:33

The quantity selector is smaller than the button.

Aljullu:
@jimjasson thanks for taking a look! You're right, I noticed that too, but I don't think we can easily fix that from WooCommerce core. In core product types, this is fixed via having a grid in `.wp-block-woocommerce-add-to-cart-form form.cart` that makes the last two elements vertically aligned and the same height. However, Composite Products, for example, modify the markup and wrap the quantity input and Add to Cart button into a couple of divs, so the WC core default CSS can't align them:

Simple Product | Variable Product | Composite Product
--- | --- | ---
imatge | imatge | imatge

So I think the ideal solution would be to either move the quantity input and Add to Cart button up to be direct descendants of `form.cart` (which I guess it's kinda tricky :sweat_smile: ) or implementing custom CSS from the extension side.

I might be missing something, though. What do you think?

jimjasson (approved):
Hey @Aljullu, 

Thank you for the additional explanations! 

I tested with Simple Products, Variable Products, Product Bundles, Composite Products and Gift Cards both using Storefront and TT5 and both using the Single Product block in a standalone page and in the single product page and confirmed that:
- the quantity value is now visible inside the selector for custom product types and;
- the quantity selector is aligned with the Add to Cart button for core product types. For non-core types, there is an alignment issue, but as we discussed [here](https://github.com/woocommerce/woocommerce/pull/57723#issuecomment-2853897417), it likely needs to be fixed on the plugin side of things. 

One issue I noticed while testing this PR, but is not related to the changes is how Variable Products show up in the Single Product block:

TT5:

Markup on 2025-05-07 at 15:12:21

Storefront:

Markup on 2025-05-07 at 15:13:04

Up to you if you want to fix this in this PR or we should open an issue for later.

Aljullu:
@jimjasson thanks for reviewing!

> One issue I noticed while testing this PR, but is not related to the changes is how Variable Products show up in the Single Product block

That's weird, I can't reproduce:

Editor | Frontend
--- | ---
imatge | imatge

I just tested in TT5 and TT4, in Firefox and Chrome, and with products with one and several attributes, and it seems to be properly aligned in all of them. Tested on `trunk` with WP 6.8.1 and Gutenberg disabled. Did you do something special? I can keep an eye on this, in case I can reproduce somewhere else, but so far I wasn't able. :thinking: 

> Up to you if you want to fix this in this PR or we should open an issue for later.

I will go ahead and merge this PR as-is and if we can consistently reproduce the other issue, we can work on it in a follow-up. :slightly_smiling_face:

PR #57732, opened 2025-05-05
Add base user interface for order fulfillments

tpaksu:

This PR includes the fulfillment drawer UI as the base for the fulfillments editor. This adds a `FulfillmentsRenderer.php` file, which is included with the `OrderFulfillmentsServiceProvider` that adds 3 new columns to the orders list: 

- **Fulfillment Status:** Will display the combined status of the fulfillment(s) of the order
- **Shipping Provider:** Will display the shipment provider(s) for the order
- **Shipment Tracking:** Will display the tracking number(s) for the order

And an edit icon 🖊️ next to the fulfillment status badge, which opens the fulfillment drawer.

Closes WOOPLUG-3976

https://github.com/user-attachments/assets/56925923-79f4-4373-8fbe-026ff222058d

1. Checkout this branch
2. Change the directory to `plugins/woocommerce`
3. Run `composer dump-autoload` just in case the autoloader doesn't pick the new file
4. Run `pnpm run build` to build the assets
5. Open the orders list page, make sure you have some orders to list
6. See that the three new columns are displayed
7. See that there are toggles for the new columns that show and hide the columns
8. When you click to the pen icon in the fulfillment status column, see that a drawer opens from the right side, matching the design in Figma
9. Assert that there is no overflow issues for the order table on both when the drawer is visible, and hidden
10. Assert that when you click the pen icon on another order, the drawer content changes, reflecting the selected order number on its title
11. Assert that the selected order row is highlighted correctly, with a shade of blue, and a left blue border
12. Assert that the close button on the fulfillment drawer hides the drawer, and clears the order selection on the list.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

Add the drawer component for displaying order fulfillments

</details>


plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/style.scss
ismaeldcom, line 1: A few suggestions for the style:
- We should use color variables to ensure support for custom themes.
- I think the pen icon should have the same color as the eye one, only the Modern Admin theme uses that tone of blue.
- The pen icon doesn't align properly on smaller screens.
- Not sure if there is a mobile design, but I think the drawer should have a 100% width on smaller screens.
- The drawer scrolls with the page. The activity drawer works as expected because it's within the header.

Screenshot 2025-05-07 at 11 40 19

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/index.tsx
ismaeldcom, line 1: DOM manipulation within a React component always looks odd, but there is no alternative due to the page being PHP-based.

It's better to create a single listener on `document` rather than one per trigger, although with only 20 per page, it is not that important.

You can also move the `is-selected` logic to a new method to be used both by the `useLayourEffect` and `onClose`.

Apart from that, if we're super picky, the order number gets removed before the drawer gets closed, so it's visible while it does the close transition 🕵️ 😄

ismaeldcom (commented):
This is taking shape already! I left some comments that need to be addressed, but the code looks good in general and works as expected.

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/style.scss
tpaksu: Thanks for the detailed review @ismaeldcom! Greatly appreciated!

> We should use color variables to ensure support for custom themes.

I tried to search for the colors and icons in the current environment with no luck, so I had two choices, one is using the values on Figma, and second, lose more time. I chose the prior one :) The design was initially for WP Calypso `DataView`s , and the colors/icons were using their standards, but now it's for `WP_List_Table`, and the colors and icons don't exist on this environment. 

> I think the pen icon should have the same color as the eye one, only the Modern Admin theme uses that tone of blue.

Yep, I agree. Addressed. 

> The pen icon doesn't align properly on smaller screens.

I had already included the fix on #57758, but moved it to this one too. Thanks for flagging :) 

> Not sure if there is a mobile design, but I think the drawer should have a 100% width on smaller screens.

There's no mobile design yet, and I'll try to ask for one when I'm near the end of development. I'll handle all in a last issue. 

> The drawer scrolls with the page. The activity drawer works as expected because it's within the header.

Yep, fixed it on the latter PR, but moved it into this one for visibility.

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/index.tsx
tpaksu: > DOM manipulation within a React component always looks odd, but there is no alternative due to the page being PHP-based.

True. I didn't have much chance than this solution. 

> It's better to create a single listener on document rather than one per trigger, although with only 20 per page, it is not that important.

> You can also move the is-selected logic to a new method to be used both by the useLayourEffect and onClose.

Done.

> Apart from that, if we're super picky, the order number gets removed before the drawer gets closed, so it's visible while it does the close transition 🕵️ 😄

Also done. The close effect was bound to the existence of an order ID, so separating the logic into two different states fixed it (needed to set the transition delay to 5secs to debug it though 😬)

All on the first and this comment are done in [7a799e2](https://github.com/woocommerce/woocommerce/pull/57732/commits/7a799e2be5903fa0ac4180999263235a0491c943).

plugins/woocommerce/src/Internal/Fulfillments/FulfillmentsRenderer.php
```diff
@@ -117,6 +118,7 @@ private function render_fulfillment_status_column_row_data( WC_Order $order, arr
 				<path d='M11.8333 2.83301L9.33329 0.333008L2.24996 7.41634L1.41663 10.7497L4.74996 9.91634L11.8333 2.83301ZM5.99996 12.4163H0.166626V13.6663H5.99996V12.4163Z' fill='#3858E9'/>
```
ismaeldcom, line 118: The inline `fill` in the `path` element will take precedence over any style from a stylesheet.

ismaeldcom (approved):
Oh! I wasn't aware of the existence of the next PR, thanks bringing the fixes here and addressing all the comments.

There is only a tiny nitpick, apart from that, it looks good to me!

plugins/woocommerce/src/Internal/Fulfillments/FulfillmentsRenderer.php
tpaksu: Thanks for pointing that out! I'll remove the `fill` property so it can use the one from CSS.

tpaksu: Or, I'll do the change on the next PR if it's ok.

ismaeldcom: Sure!

PR #57751, opened 2025-05-06
[Experimental] Reviews Pagination Client Navigation

dinhtungdu:

This PR adds client-side navigation to the reviews pagination block. Clicking on pagination links will no longer trigger a full page reload.

Changes include:
- Added a new frontend script to handle client-side navigation
- Updated block.json to support interactivity and register the view script
- Modified PHP block classes to enable interactive attributes
- Added interactive data attributes to pagination links

https://github.com/user-attachments/assets/4f074b3a-440f-48c2-bcef-2b4bfbd87d81

1. Add the `Blockified Product Details` block to the Single Product template.
2. Enable paginated reviews in Settings > Discussion > Break comments into pages.
3. Go to a product with multiple reviews (or create enough reviews to trigger pagination). You can also reduce the `Top level comments per page` setting to avoid creating too many reviews.
4. Open the reviews section.
5. Click on the pagination links (next page, previous page, or page numbers).
6. Verify that the page transitions smoothly without a full page reload.
7. Verify that the reviews content updates correctly.
8. Verify that browser back/forward buttons work as expected.
9. Verify that the page scrolls back to the top of the comment list.

### DIFF
--- a/plugins/woocommerce/src/Blocks/BlockTypes/Reviews/ProductReviews.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/Reviews/ProductReviews.php
@@ -33,7 +33,7 @@ class ProductReviews extends AbstractBlock {
 		}
 
 		$p = new \WP_HTML_Tag_Processor( $content );
-		$p->next_tag( 'div' );
+		$p->next_tag();
 		$p->set_attribute( 'data-wp-interactive', $this->get_full_block_name() );
 		$p->set_attribute( 'data-wp-router-region', $this->get_full_block_name() );
 
--- a/plugins/woocommerce/src/Blocks/BlockTypes/Reviews/ProductReviews.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/Reviews/ProductReviews.php
@@ -3,37 +3,20 @@
 namespace Automattic\WooCommerce\Blocks\BlockTypes\Reviews;
 
 use Automattic\WooCommerce\Blocks\BlockTypes\AbstractBlock;
+use Automattic\WooCommerce\Blocks\BlockTypes\EnableBlockJsonAssetsTrait;
 
 /**
  * ProductReviews class.
  */
 class ProductReviews extends AbstractBlock {
+	use EnableBlockJsonAssetsTrait;
+
 	/**
 	 * Block name.
 	 *
 	 * @var string
 	 */
 	protected $block_name = 'blockified-product-reviews';
-
-	/**
-	 * Get the frontend style handle for this block type.
-	 *
-	 * @return string[]|null
-	 */
-	protected function get_block_type_style() {
-		return null;
-	}
-
-	/**
-	 * Get the frontend script handle for this block type.
-	 *
-	 * @see $this->register_block_type()
-	 * @param string $key Data to get, or default to everything.
-	 * @return array|string|null
-	 */
-	protected function get_block_type_script( $key = null ) {
-		return null;
-	}
 
 	/**
 	 * Render the block.
@@ -48,6 +31,12 @@ class ProductReviews extends AbstractBlock {
 		if ( ! comments_open() ) {
 			return '';
 		}
-		return $content;
+
+		$p = new \WP_HTML_Tag_Processor( $content );
+		$p->next_tag( 'div' );
+		$p->set_attribute( 'data-wp-interactive', $this->get_full_block_name() );
+		$p->set_attribute( 'data-wp-router-region', $this->get_full_block_name() );
+
+		return $p->get_updated_html();
 	}
 }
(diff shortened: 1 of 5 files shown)


plugins/woocommerce/src/Blocks/BlockTypes/Reviews/ProductReviewsPagination.php
```diff
@@ -32,12 +32,22 @@ protected function render( $attributes, $content, $block ) {
...
+			array(
+				'class'               => $classes,
+				'data-wp-interactive' => 'woocommerce/blockified-product-reviews',
+			)
+		);
+
+		$p = new \WP_HTML_Tag_Processor( $content );
+		while ( $p->next_tag( 'a' ) ) {
+			$p->set_attribute( 'data-wp-on--click', 'actions.navigate' );
+		}
```
Aljullu, line 45: What's the benefit of modifying the anchors here instead of doing so in `ProductReviewsPaginationNumbers`?

plugins/woocommerce/src/Blocks/BlockTypes/Reviews/ProductReviews.php
```diff
@@ -48,6 +31,12 @@ protected function render( $attributes, $content, $block ) {
 		if ( ! comments_open() ) {
 			return '';
 		}
-		return $content;
+
+		$p = new \WP_HTML_Tag_Processor( $content );
+		$p->next_tag( 'div' );
+		$p->set_attribute( 'data-wp-interactive', $this->get_full_block_name() );
+		$p->set_attribute( 'data-wp-router-region', $this->get_full_block_name() );
```
Aljullu: Not a strong opinion (and I'm not even sure if it would work), but I feel it would be safer to add these attributes in a wrapper `<div>` than modifying the block's content. While it has the drawback that it adds an extra `<div>`, it makes it resilient to any changes in the block's markup.

Aljullu (commented):
Nice improvement, it works really great!

https://github.com/user-attachments/assets/e94f4373-de17-4724-91b9-33a982a7f6b2

I left two inline questions, but nothing blocking.

<details>

* Theme: Twenty Twenty-Five
* Store size: small
</details>

plugins/woocommerce/src/Blocks/BlockTypes/Reviews/ProductReviewsPagination.php
dinhtungdu: Because we also have next/prev link in `ProductReviewsPaginationNext` and `ProductReviewsPaginationPrevious` classes.

plugins/woocommerce/src/Blocks/BlockTypes/Reviews/ProductReviews.php
dinhtungdu: It will work, but we don't have to be extra cautious here IMO. The markup of the Product Reviews block is the contract that needs to be held; it's only the wrapper `div` in this case.

By injecting the directives into the wrapper div, we can match the markup between the editor and the frontend, making styling easier. If we want to change the markup of the block, the logic here, which is a part of the block rendering, should be updated as well.

The markup might be changed for sure, but it's not likely to change here, and we're injecting the directive into the wrapper `div` element, which is even less likely to change.

Aljullu (approved):
Thanks for the extra explanations, @dinhtungdu! I still have some mixed feelings about modifying the `$content` markup during `render()`. But given that these blocks are fully owned by WooCommerce, it's probably safe. LGTM

dinhtungdu:
@Aljullu, I feel like this is a good topic to discuss further. I avoid adding an extra wrapper because it will make styling harder (remember the issue with the extra wrapper for notices in the Product Collection block?). IMO, the output of the block should follow what is defined in https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/save.tsx, and we can rely on that. That markup can be modified by plugins through the PHP filter, but the markup should still be respected, or styling or interactivity can break; it's the extender's responsibility IMO.

Thinking more about it, maybe we can make this more tolerant by injecting the directive to the first wrapper tag, not only `div`, but any element, so if 3PDs change the wrapper to `section` or `aside`, the injection still works as expected.

dinhtungdu:
I updated the PR to inject directives into the first opening tag, making the block more resilient. As we have a `tagName` attribute (no inspector control has been implemented, yet), this allows the block to have a different wrapper element than just `div`.

dinhtungdu:
Merging this PR to unblock WOOPLUG-4163.

Aljullu:
> I updated the PR to inject directives into the first opening tag, making the block more resilient. As we have a `tagName` attribute (no inspector control has been implemented, yet), this allows the block to have a different wrapper element than just `div`.

Perfect, thanks!

PR #57756, opened 2025-05-06
[dev] Refactor workflow to highlight REST API changes based on file changes

adimoldovan:

Closes WOOPLUG-4145

Updated the workflow that sends notifications highlighting REST Api changes to use file changes instead of a label.

- renamed the workflow
- used dorny/paths-filter to detect changed files, to be consistent with what we use in ci.yml
- simplified the notification content to only include a PR title and link and the milestone.

- Check the file paths defined in rest filter. Do they make sense, are there others you would include?

- No matching files changed - notification skipped: [run](https://github.com/woocommerce/woocommerce/actions/runs/14858193707/job/41716396659?pr=57756)
- Matching files - notification sent: 
	- [run](https://github.com/woocommerce/woocommerce/actions/runs/14858331363/job/41716825392?pr=57756#step:2:28)
	- notification with milestone: p1746531420604319-slack-C02180M7JFM
	- notification without milestone: p1746531195102129-slack-C02180M7JFM
- Closed un-merged PR did not trigger a notification: [run](https://github.com/woocommerce/woocommerce/actions/runs/14858724262?pr=57756)


.github/workflows/pr-highlight-rest-api-changes.yml
```diff
@@ -0,0 +1,41 @@
+name: Highlight Rest API Changes
+
+on:
+  pull_request:
+    types: [synchronize, closed]
```
adimoldovan: Revert before merge
```suggestion
    types: [closed]
```

.github/workflows/pr-highlight-rest-api-changes.yml
```diff
@@ -0,0 +1,41 @@
...
+
+concurrency:
+  group: highlight-rest-api-changes-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  send-notification-for-rest-api-changes:
+    name: Send Notification for Rest API Changes
+    # if: github.event.pull_request.merged == true
```
adimoldovan: Revert before merge
```suggestion
    if: github.event.pull_request.merged == true
```

Konamiman:
There's a bunch of controllers inheriting from [RestApiControllerBase](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Internal/RestApiControllerBase.php) that have all class names ending with `RestController`. So maybe we could add the following one: `'plugins/woocommerce/src/**/*RestController'` (I'm not sure if that's the correct syntax, what's the difference between `*` and `**`?)

adimoldovan:
> There's a bunch of controllers inheriting from [RestApiControllerBase](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Internal/RestApiControllerBase.php?rgh-link-date=2025-05-08T08%3A49%3A42Z) that have all class names ending with `RestController`. So maybe we could add the following one: `'plugins/woocommerce/src/**/*RestController'` (I'm not sure if that's the correct syntax, what's the difference between `*` and `**`?)

Thanks for that @Konamiman! Added with [10b23b8](https://github.com/woocommerce/woocommerce/pull/57756/commits/10b23b864eddee9d05a4022e39a3da13d75bf91e)

Konamiman (approved):

PR #57768, opened 2025-05-06
Move recipient settings to the email editor sidebar

lysyjan:

Closes WOOPLUG-3527.

This PR adds new input fields to the email settings panel that could already be set in the old email settings.
Those fields are:
- Recipient - displayed only when the recipient can be set. If not, a text mentions that the email is sent to the customer.
- CC - displayed when the toggle is enabled
- BCC - displayed when the toggle is enabled

| Before | After |
| ------ | ----- |
|Screenshot 2025-05-07 at 9 14 40|Screenshot 2025-05-07 at 9 11 29|
||Screenshot 2025-05-07 at 9 12 00|

1. Make sure the Block Email Editor feature is enabled in WooCommerce > Settings > Advanced > Features.
2. Go to WooCommerce > Settings > Emails.
3. Edit one of the emails and update values in the recipient fields in the settings sidebar.
4. Make sure that changes are stored.
5. Sent email and verify that the new values are used.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/sidebar_settings.tsx
+++ b/plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/sidebar_settings.tsx
@@ -3,9 +3,16 @@
  */
 import { select, dispatch } from '@wordpress/data';
 import { store as coreDataStore, useEntityProp } from '@wordpress/core-data';
+import {
+	BaseControl,
+	PanelRow,
+	TextControl,
+	ToggleControl,
+} from '@wordpress/components';
 import { addFilter } from '@wordpress/hooks';
 import { __ } from '@wordpress/i18n';
 import clsx from 'clsx';
+import { useState } from '@wordpress/element';
 
 /**
  * Internal dependencies
@@ -24,7 +31,13 @@ const SidebarSettings = ( { RichTextWithButton } ) => {
 		'woocommerce_data'
 	);
 
-	const updateWooMailProperty = ( name: string, value: string ) => {
+	// Initialize toggle control state
+	const [ addBCC, setAddBCC ] = useState(
+		woocommerce_email_data?.bcc || false
+	);
+	const [ addCC, setAddCC ] = useState( woocommerce_email_data?.cc || false );
+
+	const updateWooMailProperty = ( name: string, value: string | boolean ) => {
 		const editedPost = select( coreDataStore ).getEditedEntityRecord(
 			'postType',
 			'woo_email',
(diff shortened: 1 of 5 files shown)


plugins/woocommerce/src/Internal/EmailEditor/EmailApiController.php
```diff
@@ -47,17 +47,36 @@ final public function init(): void {
 	 * @return array - The email data.
 	 */
 	public function get_email_data( $post_data ): array {
-		$email_type  = $this->post_manager->get_email_type_from_post_id( $post_data['id'] );
-		$post_option = get_option( "woocommerce_{$email_type}_settings" );
-		$email       = $this->get_email_by_type( $email_type );
+		$email_type = $this->post_manager->get_email_type_from_post_id( $post_data['id'] );
+		$email      = $this->get_email_by_type( $email_type );
 
+		// When the email type is not found, it means that the email type is not supported.
```
pavel-mailpoet, line 54: I don't know if this is the best option. The merchant shouldn't be on that page if the email type is not supported. Would it be better to return 404 or another HTTP error instead?

plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/sidebar_settings.tsx
```diff
@@ -108,6 +121,104 @@ const SidebarSettings = ( { RichTextWithButton } ) => {
...
+					<BaseControl __nextHasNoMarginBottom>
+						<TextControl
+							__nextHasNoMarginBottom
+							__next40pxDefaultSize
+							value={ woocommerce_email_data?.cc || '' }
+							onChange={ ( value ) => {
+								updateWooMailProperty( 'cc', value );
+							} }
+							help={ __(
+								'Separate with commas to add multiple email addresses.',
```
pavel-mailpoet: I know this is coming from the design, but should we explain what CC is instead? Instead of explaining how to add more? Same for BCC

plugins/woocommerce/src/Internal/EmailEditor/EmailApiController.php
```diff
@@ -148,7 +148,7 @@ public function get_email_data_schema(): array {
 	 * @param string $id - The email ID.
 	 * @return \WC_Email|null - The email object or null if not found.
 	 */
-	private function get_email_by_type( string $id ): ?WC_Email {
+	private function get_email_by_type( ?string $id ): ?WC_Email {
```
pavel-mailpoet, line 158: Here the commit message explain what but not why and I am not sure why do we need this change.

pavel-mailpoet (commented):
I have a couple of comments here, but still need to test the functionality.

pavel-mailpoet:
Screenshot 2025-05-07 at 12 07 50
Screenshot 2025-05-07 at 12 08 11


When I add someone to CC or BCC, I don't see the changes on the email listings. Should we reflect the changes in there? Or at least create a follow-up ticket?

pavel-mailpoet:
I also noticed that if I change the values, the "Save" button is not active. Is that intentional?

plugins/woocommerce/src/Internal/EmailEditor/EmailApiController.php
lysyjan: This commit was created before our retrospective today. The change was made to ensure consistency in the state when the email type is not found. Because the post manager can return the null value, this is a consequence of handling this state.

plugins/woocommerce/src/Internal/EmailEditor/EmailApiController.php
lysyjan: That sounds reasonable. I haven't tried this approach in the WC plugin. I am going to take a look

plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/sidebar_settings.tsx
lysyjan: We can change it as we want. This message was added based on the recipient field. We can remove it or change it now or later. Or we can ask Shiva now or during some feedback iteration.

lysyjan:
> When I add someone to CC or BCC, I don't see the changes on the email listings. Should we reflect the changes in there? Or at least create a follow-up ticket?

The CC or BCC field is not visible directly, but the listing has a hover when there is more than one recipient.

> I also noticed that if I change the values, the "Save" button is not active. Is that intentional?

The behavior is identical to another edit action in the email editor. You need to use "Save draft" and then you can click on "Save email" which returns you to the listing page.

plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/sidebar_settings.tsx
pavel-mailpoet: Let's change it to:
"Add recipients who will receive a copy of the email. Separate multiple addresses with commas."
"Add recipients who will receive a hidden copy of the email. Separate multiple addresses with commas."

plugins/woocommerce/src/Internal/EmailEditor/EmailApiController.php
lysyjan: @pavel-mailpoet We cannot return 404 here because we use the `register_rest_field` method. When there is a registered expected schema, the output has to match it for each response. Returning empty values is the best option here.

pavel-mailpoet (approved):

PR #57775, opened 2025-05-06
Migrate block email editor to the wordpress/editor Editor component

costasovo:

This PR is a step towards migrating the block email editor to WordPress's Block Editor. In this PR, we migrate from the custom email editor to the Editor component, which is used in the site editor and the post editor, and covers a lot of functionality we replicated (e.g., whole UI around canvas) or missed in the custom editor implementation.

We do this migration to improve the alignment with core's editors and to reduce the codebase we need to maintain and update to keep up with the changes in the core.

Closes WOOPLUG-3951

| Before | After |
| ------ | ----- |
|    Screenshot 2025-05-06 at 17 17 42  |  Screenshot 2025-05-07 at 9 21 46 |

#### Note for reviewer:
Sorry for the big PR. I tried my best to split changes into small commits and explain changes in the commit description.

1. Enable the Block Email Editor in Woo > Settings > Advanced > Features
2. Go to Woo > Settings > Emails and try editing emails. 
3. Test also the Edit template mode (Woo > Settings > Emails > Edit template)
4. To verify that we can create a new email with the Select template modal, you need to enable native listing for Woo Email CPT
add following to
```php
'public'       => false,
'show_ui'      => true,
'show_in_menu' => true,
'show_in_rest' => true,
```
to 

https://github.com/woocommerce/woocommerce/blob/1817abf09affb7e33fb0c4d68cdecf6a810536c4/plugins/woocommerce/src/Internal/EmailEditor/Integration.php#L122-L139			

Overall, please try to compare it with the version before the refactor and look for regressions.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


triple0t (commented):
Incredible work. Well done, @costasovo 

Feedback from my first look at the UI. I haven't dived into the codebase and commits yet.

1. The Preferences button is not working. Clicking on it does not do anything
Screenshot 2025-05-08 at 12 40 44 PM
   * We don't have it on the old editor. So maybe fix it, or we hide/remove it.

2. Border: We only need to support `px`. The value can currently be saved in px, %, em, rem, etc
Screenshot 2025-05-08 at 12 46 26 PM
Screenshot 2025-05-08 at 12 45 33 PM
Screenshot 2025-05-08 at 12 53 49 PM


3. Colors: we only need to support `hex`.
Screenshot 2025-05-08 at 12 47 21 PM


4. Saving email shows the post updated in the Snackbar.
Screenshot 2025-05-08 at 12 48 26 PM
Screenshot 2025-05-08 at 12 48 42 PM


5. Why "Send preview", not "Send test email" or "Send email preview"?
Screenshot 2025-05-08 at 2 32 28 PM


6. There's a problem with the ListView:  
Comparing: 

7. Do we want to allow folks to edit the Email Template description?
Screenshot 2025-05-08 at 1 25 15 PM


8. ~Saving an email does not redirect back to the email listing page~ I needed to click the button twice. 
Save email button needing to be clicked twice: What if we add support for this https://developer.wordpress.org/block-editor/reference-guides/slotfills/main-dashboard-button/ in which we can add an icon to close or return back to a page?


**What went well.**
I can see all available blocks
Screenshot 2025-05-08 at 1 17 31 PM

The Interface works as expected
Screenshot 2025-05-08 at 1 37 22 PM

The new email template selector works
Screenshot 2025-05-08 at 2 36 31 PM

packages/php/email-editor/src/Engine/content-editor.css
```diff
@@ -114,3 +114,41 @@ ol.has-background {
...
+	margin-block-start: 0;
+}
+
+/*
+ * Hide the post title.
+ * When user disables the template-lock mode we don't want to show the post title.
+ */
+.editor-visual-editor__post-title-wrapper {
+	display: none;
+}
```
triple0t, line 133: Can you share a screenshot of what the style is fixing?
When I disabled the style, I didn't notice any difference.

packages/php/email-editor/src/Engine/content-editor.css
```diff
@@ -63,7 +63,7 @@
  * For the WYSIWYG experience we don't want to display any margins between blocks in the editor
  */
 .wp-block {
-	clear: both; // for ensuring that floated elements (images) are cleared
+	clear: both;
```
triple0t, line 66: Thanks for fixing this

packages/js/email-editor/src/hooks/use-email-styles.ts
```diff
@@ -150,11 +150,11 @@ export const useEmailStyles = (): EmailStylesData => {
...
 	// This is email level styling stored in post meta.
-	const styles = useMemo(
-		() =>
-			cleanupUserStyles( shortenWpPresetVariables( userTheme?.styles ) ),
-		[ userTheme ]
-	);
+	const styles = useMemo( () => {
+		return userTheme
+			? cleanupUserStyles( shortenWpPresetVariables( userTheme?.styles ) )
+			: {};
```
triple0t: To prevent a lot of unnecessary renders, I think we should provide a constant Empty Object or Empty Array for use in components.

We did something similar [here](https://github.com/woocommerce/woocommerce/blob/038bfbee7e16b98903c0634af70af60fdfb74c05/packages/js/email-editor/src/hooks/use-content-validation.ts#L19-L21), and [here](https://github.com/woocommerce/woocommerce/blob/99f04562c084808e02515a526664fdc0c77fa688/plugins/woocommerce/client/blocks/assets/js/data/collections/selectors.js#L10) is an example use in Woo Blocks.

packages/js/email-editor/src/hacks/publish-save.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { useContentValidation } from '../hooks';
+
+import { editorCurrentPostType } from '../store';
+
+type NextButtonSlotPropType = {
+	children: React.ReactNode;
+};
+
+function NextPublishSlot( { children }: NextButtonSlotPropType ) {
+	const [ sendButtonPortalEl ] = useState( document.createElement( 'div' ) );
```
triple0t: The decision to use `useState` here instead of `useRef` to store the Node Element is quite interesting. May I ask why?

packages/js/email-editor/src/hacks/publish-save.tsx
```diff
@@ -0,0 +1,111 @@
...
+		}
+	}, [ sendButtonPortalEl ] );
+
+	return createPortal( <>{ children }</>, sendButtonPortalEl );
+}
+
+export function PublishSave() {
+	const { validateContent, isInvalid } = useContentValidation();
+	const observerRef = useRef< MutationObserver | null >( null );
+	const { dirtyEntityRecords } = useEntitiesSavedStatesIsDirty();
```
triple0t: This is pretty incredible, but looking at it from a performance point of view, it is difficult to accept.
The component will render and re-render for every unrelated change in the editor canvas because of  `useEntitiesSavedStatesIsDirty`, `useContentValidation`, and `MutationObserver`

Since we need to replace the `Save` button when the post is saved, why not subscribe only to that notification?

Options
1. Listen for doAction [hook](https://github.com/WordPress/gutenberg/blob/bc28417b3a95c1ab8fb8f5c67a98dbf31a6cf3ec/packages/editor/src/store/actions.js#L279-L283)
2. Other selectors: didPostSaveRequestSucceed
3. If none of that works, we could use a [custom](https://github.com/WordPress/gutenberg/blob/7e13df08782341db799138e943b56d037cd7fe4e/packages/editor/src/components/editor-interface/index.js#L51) save button. Skipping the whole hack and replacement stuff.

packages/js/email-editor/src/hacks/publish-save.tsx
```diff
@@ -0,0 +1,111 @@
...
+			observerRef.current.disconnect();
+		}
+		observerRef.current = new MutationObserver( () => {
+			toggleElementVisible( publishButton, displayOriginalPublishButton );
+		} );
+		observerRef.current.observe( publishButton, {
+			attributes: true,
+			childList: true,
+			subtree: false,
+		} );
```
triple0t: Let's wrap this block of code in an if statement.
We're going to have a bunch of errors here if the Node element can not be found.

packages/js/email-editor/src/components/styles-sidebar/styles-sidebar.tsx
```diff
@@ -3,19 +3,15 @@
...
+import { PluginSidebar, PluginSidebarMoreMenuItem } from '@wordpress/editor';
 import { useSelect } from '@wordpress/data';
-import { ComplementaryArea } from '@wordpress/interface';
-import { ComponentProps } from 'react';
 import { styles } from '@wordpress/icons';
-import {
-	__experimentalNavigatorProvider as NavigatorProvider,
-	__experimentalNavigatorScreen as NavigatorScreen,
-} from '@wordpress/components';
+import { Navigator } from '@wordpress/components';
```
triple0t: Thank you for migrating to the stable component.

packages/js/email-editor/src/components/sidebar/template-settings-panel.tsx
```diff
@@ -20,10 +22,14 @@ export function TemplateSettingsPanel() {
...
+		<PluginDocumentSettingPanel
+			name="template-settings-panel"
+			title={ __( 'Settings', 'woocommerce' ) }
+			className="woocommerce-email-editor__settings-panel"
+		>
 			{ templateSections.map( ( section ) => (
 				<div key={ section.id }>{ section.render() }</div>
 			) ) }
-		</>
+		</PluginDocumentSettingPanel>
```
triple0t: I think at this point, we should wrap this component in a custom error boundary so that errors that occur from third-party integration do not totally break the whole email editor interface.

packages/js/email-editor/src/components/sidebar/style.scss
```diff
@@ -8,17 +8,3 @@
...
-		box-shadow: none;
-		outline: none;
-	}
-
-	// Because we use complementary area component from the WP interface package the is visible an unwanted close icon.
-	// This style hides it.
-	.interface-complementary-area-header__small {
-		display: none;
-	}
-}
```
triple0t, line 24: We can remove this as well. https://github.com/woocommerce/woocommerce/blob/173bae563b0e0741f3c1cb63760696a7715be2a1/packages/js/email-editor/src/components/styles-sidebar/style.scss#L55-L59

packages/js/email-editor/src/store/actions.ts
```diff
@@ -1,139 +1,31 @@
 /**
  * External dependencies
  */
-import { dispatch, select } from '@wordpress/data';
-import { store as interfaceStore } from '@wordpress/interface';
```
triple0t, line 5: Since we don't need `@wordpress/interface` anymore, can we remove these
* https://github.com/woocommerce/woocommerce/blob/173bae563b0e0741f3c1cb63760696a7715be2a1/packages/js/email-editor/src/types/index.ts#L7
* https://github.com/woocommerce/woocommerce/blob/173bae563b0e0741f3c1cb63760696a7715be2a1/packages/js/email-editor/src/types/index.ts#L110
* https://github.com/woocommerce/woocommerce/blob/173bae563b0e0741f3c1cb63760696a7715be2a1/packages/js/email-editor/src/types/wordpress-modules.ts#L48-L67

triple0t (commented):
Well done, @costasovo 

Thanks for implementing this.

Some notes

1. The Global styles are not loading (and not applied) on the Editor for non-admin users. Preview works fine. This works with the former Editor.
2. We need to fix type errors `pnpm --filter='@woocommerce/email-editor' check-types`
3. Question: Since we are using the Editor now, do we still need the `SlotFillProvider` component?

packages/php/email-editor/src/Engine/content-editor.css
costasovo: The CSS is effective when user disables "Show template" in the preview dropdown.

Screenshot 2025-05-12 at 16 39 08

Without the CSS, when the "Show template" is disabled, there is a huge editable email Title displayed above the email content.

Screenshot 2025-05-12 at 16 43 50


We can't easily remove the "Show template" toggle from the preview options and IMO it may make sense for some users who want to focus purely on content also helps to educate that there is a template + content concept so I think it is ok to keep it. I hide the title from the canvas because, unlike a web post we don't render the title in the email content.

packages/js/email-editor/src/hacks/publish-save.tsx
costasovo: Oh! Nice catch! No reason. It is a mistake or a typo, probably created when I was exploring possible solutions. It seemed to work because we actually never changed that state. Updated in https://github.com/woocommerce/woocommerce/pull/57775/commits/ad3d3342c05437da97d65cb6c4f77df091741125

packages/js/email-editor/src/hacks/publish-save.tsx
costasovo: Good catch! Unfortunately, we can't use Option 1 or 2, because we need to change the button also when the user types. I wanted to avoid the custom button, because we would need to implement much more.
In the end, I used two boolean selectors from the editor store and moved the validation hook to a subcomponent.
So, currently, there are no re-renders when typing in Canvas. Done in https://github.com/woocommerce/woocommerce/pull/57775/commits/ad3d3342c05437da97d65cb6c4f77df091741125

packages/js/email-editor/src/hacks/publish-save.tsx
costasovo: Done in https://github.com/woocommerce/woocommerce/pull/57775/commits/ad3d3342c05437da97d65cb6c4f77df091741125

packages/js/email-editor/src/components/sidebar/template-settings-panel.tsx
costasovo: Added in https://github.com/woocommerce/woocommerce/pull/57775/commits/a47a9728b23a30e19f8819d40ae55d0ec21f7435

costasovo:
@triple0t Thank you for the thorough review! Much appreciated.

> 1. The Preferences button is not working. Clicking on it does not do anything

Fixed in https://github.com/woocommerce/woocommerce/pull/57775/commits/334ca2f0546ff9993fd24610920e64c4cbf60b3b

> 2. Border: We only need to support px. The value can currently be saved in px, %, em, rem, etc

This problem is not new. It is already present in the current trunk. I created an extra issue for that https://github.com/woocommerce/woocommerce/issues/57957

> 3. Colors: we only need to support hex.

I found that even when a user uses RGB to configure the color we render hex in the renderer so IMO it doesn't need fixing. Only problem is opacity, but we already have a warning saying "Border display may vary or be unsupported in some email clients." so it is IMO covered.

> 4. Saving email shows the post updated in the Snackbar.

Addressed in https://github.com/woocommerce/woocommerce/pull/57775/commits/7d6fc4b76b9f38c650bc8e3a6802bc703fb8dd71

> 5. Why "Send preview", not "Send test email" or "Send email preview"?

Good catch. It was kind of random text. Updated in https://github.com/woocommerce/woocommerce/pull/57775/commits/7d6fc4b76b9f38c650bc8e3a6802bc703fb8dd71

> 6 There's a problem with the ListView:

I found the lags in opening ListView when a block is selected are related to `template-lock` mode. I was able to replicate that also in the core's post editor. I didn't pursue that further. 

> 7 Do we want to allow folks to edit the Email Template description?

It is a feature we get "Out of the box". It is not helpful for Woo Integration, as we have only one template. But it also does no harm. In general, I think it is useful as users might want to add some additional info for their templates. 

> 8 I needed to click the button twice.
Save email button needing to be clicked twice: What if we add support for this https://developer.wordpress.org/block-editor/reference-guides/slotfills/main-dashboard-button/, in which we can add an icon to close or return to a page?

Currently, you still need to click Save and then Save email if there are some changes. This will be addressed in https://github.com/woocommerce/woocommerce/issues/57959

> The Global styles are not loading (and not applied) on the Editor for non-admin users. Preview works fine. This works with the former Editor.

Fixed in https://github.com/woocommerce/woocommerce/pull/57775/commits/c97f87cbe7a37cfcf8250fec89c999e8e8555c8d

> We need to fix type errors pnpm --filter='@woocommerce/email-editor' check-types

Fixed in https://github.com/woocommerce/woocommerce/pull/57775/commits/a0b30102fc977b8e907aa37a670c1a464dd09aa3

Note: I fixed most of the warnings by adding ts-expect-error. I think we need to update our local WP packages to something newer than the current wp-6.6. I tried to do that, but it seems to require more time, so I'll prepare a ticket for that.

> Question: Since we are using the Editor now, do we still need the SlotFillProvider component?

I actually copied that from [edit-post's Layout](https://github.com/WordPress/gutenberg/blob/918d152f25248a53e3fedd1ec3af2eadeb038e9d/packages/edit-post/src/components/layout/index.js#L584). I think it is needed for modals and command menu.

I hope I didn't miss any of your remarks. I also added a few small fixes. New commits start from https://github.com/woocommerce/woocommerce/pull/57775/commits/334ca2f0546ff9993fd24610920e64c4cbf60b3b

packages/js/email-editor/src/hooks/use-email-styles.ts
costasovo: Implemented in https://github.com/woocommerce/woocommerce/pull/57775/commits/ad3d3342c05437da97d65cb6c4f77df091741125

triple0t (approved):
Looks good to me. Thank you for implementing the change.
Please feel free to merge when ready.

PR #57789, opened 2025-05-06
Update PHP test to match "Password-protected" text changes in WP nightly

opr:

Updates a PHP test to match text changes made in WP Core.

This commit: https://github.com/WordPress/wordpress-develop/commit/0f5314739c80b9cdf22e13b272b69c7c9ea5df28

changed the wording on password-protected posts from `This content is password protected` to `This content is password-protected` and the test in Woo which looks for that string has been updated.

Fixes WOOPLUG-4184

1. Ensure CI passes

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/tests/php/includes/class-wc-shortcodes-test.php
+++ b/plugins/woocommerce/tests/php/includes/class-wc-shortcodes-test.php
@@ -463,13 +463,7 @@ class WC_Shortcodes_Test extends WC_Unit_Test_Case {
 			)
 		);
 
-		// If the current WP version is greater than 6.9-alpha-60222 the password prompt will be different.
-		// It was changed in https://core.trac.wordpress.org/changeset/60222.
-		$expected = version_compare( get_bloginfo( 'version' ), '6.9-alpha-60222', '>=' )
-			? 'This content is password-protected'
-			: 'This content is password protected';
-
-		$this->assertStringContainsString( $expected, $product_page );
+		$this->assertMatchesRegularExpression( '/This content is password[- ]protected/', $product_page );
 
 		wp_set_current_user( self::$user_contributor );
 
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/tests/php/includes/class-wc-shortcodes-test.php
```diff
@@ -474,6 +480,6 @@ public function test_product_page_shortcode_protected_product() {
 			)
 		);
 
-		$this->assertStringContainsString( 'This content is password protected', $product_page );
+		$this->assertStringContainsString( $expected, $product_page );
```
mikejolley: This is completely fine but I wonder if instead of relying on specific versioning we should instead cover both possibilities? e.g. through a short regex. Copiliot response below.

--

Yes, it would be possible to achieve the same result using a regular expression (regex) that matches both possible strings. Here's an example:

```php
$this->assertMatchesRegularExpression( '/This content is password[- ]protected/', $product_page );
```

### Explanation:
- The regex `/This content is password[- ]protected/` accounts for both variations of the string:
  - `This content is password-protected`
  - `This content is password protected`
- `[- ]` is a character set that matches either a hyphen (`-`) or a space (` `) between `password` and `protected`.

### Benefits of Using Regex:
- It simplifies the logic by removing the need to check the WordPress version with `version_compare`.
- It ensures the test will work regardless of which variation of the string is present.

### Drawback:
- Using a regex may slightly reduce the readability of the test compared to explicitly defining the expected string based on the version check. It also assumes no other variations of the string will occur in future versions.

If clarity and maintainability are the priority, the current version-based approach might be preferable. However, if simplicity is desired, the regex approach is a valid alternative.

mikejolley (commented):
This looks fine however, what do you think about handling both strings without versioning?

plugins/woocommerce/tests/php/includes/class-wc-shortcodes-test.php
opr: Yeah, I agree. The _exact matching_ on the string isn't the important part of this test, checking that the content is password protected is. Using regex would make it simpler. I'll change.

mikejolley (approved):
Thanks thats better, if CI is passing (looks to be), :shipit:

PR #57797, opened 2025-05-07
[Blueprint] Prevent import interruption when site state changes from Coming Soon to Live

chihsuan:

Previously, if a Blueprint import was started while the site was in Coming Soon mode, but the import process itself changed the site to Live (as part of the imported settings), subsequent import steps would fail due to the site state change ([Add security restriction to only allow blueprint imports in Coming Soon mode #57382](https://github.com/woocommerce/woocommerce/pull/57382)). This resulted in incomplete imports.

This PR fixes the issue by introducing a session token mechanism: when an import starts, a session token is generated and returned to the client. All subsequent import step requests must include this session token, allowing the import to continue even if the site becomes Live mid-process. (New imports are still blocked on live sites, but ongoing sessions are allowed to complete.)

1. Set the site to Coming Soon mode.
2. Go to `/wp-admin/admin.php?page=wc-settings&tab=advanced&section=blueprint`
3. Import  [woo-blueprint.json](https://github.com/user-attachments/files/20077369/woo-blueprint.json)
4. Confirm the import is successful.
5. Reload the page and confirm the import is disabled.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Fix Blueprint imports interrupted if the site state changed from Coming Soon to Live

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/src/Admin/Features/Blueprint/RestApi.php
+++ b/plugins/woocommerce/src/Admin/Features/Blueprint/RestApi.php
@@ -262,7 +262,7 @@ class RestApi {
 	 *
 	 * @param \WP_REST_Request $request The request object.
 	 *
-	 * @return array
+	 * @return \WP_REST_Response|array
 	 */
 	public function import_step( \WP_REST_Request $request ) {
 		$session_token = $request->get_header( 'X-Blueprint-Import-Session' );
@@ -311,11 +311,14 @@ class RestApi {
 		$step_importer   = new ImportStep( $step_definition );
 		$result          = $step_importer->import();
 
-		return array(
-			'success'      => $result->is_success(),
-			'messages'     => $result->get_messages(),
-			'sessionToken' => $session_token,
+		$response = new \WP_REST_Response(
+			array(
+				'success'  => $result->is_success(),
+				'messages' => $result->get_messages(),
+			)
 		);
+		$response->header( 'X-Blueprint-Import-Session', $session_token );
+		return $response;
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


plugins/woocommerce/src/Admin/Features/Blueprint/RestApi.php
```diff
@@ -300,18 +312,24 @@ public function import_step( \WP_REST_Request $request ) {
 		$result          = $step_importer->import();
 
 		return array(
-			'success'  => $result->is_success(),
-			'messages' => $result->get_messages(),
+			'success'      => $result->is_success(),
+			'messages'     => $result->get_messages(),
+			'sessionToken' => $session_token,
```
louwie17: I don't think its common practice to pass the `sessionToken` as part of the network request body, we should be including this in the response header instead.
I believe we did have to do something like this:
```
$response = new \WP_REST_Response(
	array(
		'success'  => $result->is_success(),
		'messages' => $result->get_messages(),
   )
);
$response->header( 'X-Blueprint-Import-Session', $session_token );
return $response;
```

louwie17 (changes requested):
This tested well! I just left one comment around returning the session token, we should do this through the response header instead of the payload.

plugins/woocommerce/src/Admin/Features/Blueprint/RestApi.php
chihsuan: Thanks for the suggestion! I've updated the code to include the sessionToken in the response header in c5b0004af65266850c33aa9d3a6f853fcad83d5b

louwie17 (approved):
Thanks for the updates, this looks good, nice work!!

PR #57809, opened 2025-05-07
Add schema validation for Patterns Toolkit in PTKClient and update PTKPatternsStore to utilize validation

gigitux:

This PR adds schema validation for Patterns Toolkit response in `PTKClient`to validate the payload. Also, we do the same validation when we get the patterns from the transient. 

Closes # .

| Before | After |
| ------ | ----- |
|        |       |

1. Checkout this PR https://github.com/woocommerce/woocommerce/pull/57815 that included this patch + malformed payload.
2. Visit your store. Ensure that you don't see any fatal.
3. Remove:

```
|| ! $this->ptk_client->is_valid_schema( $patterns )
```

from [this check ](https://github.com/woocommerce/woocommerce/blob/ecf31e0881f3f4952a655ac0f2fa27e2ea452b09/plugins/woocommerce/src/Blocks/Patterns/PTKPatternsStore.php#L121-L122)
4. See fatal

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Add schema validation for Patterns Toolkit in PTKClient and update PTKPatternsStore to utilize validation.

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Blocks/Patterns/PTKPatternsStore.php
```diff
@@ -115,7 +115,7 @@ public function get_patterns() {
 		$patterns = get_transient( self::TRANSIENT_NAME );
 
 		// Only if the transient is not set, we schedule fetching the patterns from the PTK.
-		if ( false === $patterns ) {
+		if ( false === $patterns || ! $this->ptk_client->is_valid_ptk_patterns( $patterns ) ) {
 			$this->schedule_fetch_patterns();
```
gigitux, line 119: This not cause any loop because we check the `last_fetch_patterns_request` transient:

https://github.com/woocommerce/woocommerce/blob/ecf31e0881f3f4952a655ac0f2fa27e2ea452b09/plugins/woocommerce/src/Blocks/Patterns/PTKPatternsStore.php#L95-L107

We clean up it every hour.

plugins/woocommerce/src/Blocks/Patterns/PTKClient.php
```diff
@@ -66,4 +127,16 @@ public function fetch_patterns( array $options = array() ) {
...
+	/**
+	 * Validate the patterns toolkit patterns.
+	 *
+	 * @param array $patterns The patterns to validate.
+	 * @return bool
+	 */
+	public function is_valid_ptk_patterns( array $patterns ) {
+		$is_pattern_payload_valid = rest_validate_value_from_schema( $patterns, $this->schema );
+
+		return ! is_wp_error( $is_pattern_payload_valid );
```
gigitux: Technically, this class is named `PTKClient`, so it would be reasonable to expect only HTTP client-related code here. However, since this feature is in maintenance mode, it might not be worth creating a dedicated class at this point.

plugins/woocommerce/src/Blocks/Patterns/PTKClient.php
```diff
@@ -14,6 +14,68 @@ class PTKClient {
...
+					],
+					'name'       => [
+						'type' => 'string',
+					],
+					'html'       => [
+						'type' => 'string',
+					],
+					'categories' => [
+						'type'                 => 'object',
+						'additionalProperties' => [
```
gigitux, line 55: https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/#:~:text=%2C%0A)%3B-,additionalProperties,-Perhaps%20unintuitively%2C%20by

plugins/woocommerce/tests/php/src/Blocks/Patterns/PTKPatternsStoreTest.php
gigitux, line 1: Even for mock data, we should use mock with expected structure.

plugins/woocommerce/tests/php/src/Blocks/Patterns/PTKPatternsStoreTest.php
```diff
@@ -277,21 +406,4 @@ public function test_fetch_patterns_should_filter_out_the_patterns_with_dependen
...
-	 */
-	private function assertErrorResponse( $response, $expected_error_message ) {
-		$this->assertInstanceOf( WP_Error::class, $response );
-
-		$error_code = $response->get_error_code();
-		$this->assertEquals( 'patterns_store_error', $error_code );
-
-		$error_message = $response->get_error_message();
-		$this->assertEquals( $expected_error_message, $error_message );
-	}
```
gigitux, line 296: Not used

plugins/woocommerce/src/Blocks/Patterns/PTKClient.php
```diff
@@ -14,6 +14,71 @@ class PTKClient {
...
+					],
+					'dependencies' => [
+						'type'  => 'array',
+						'items' => [
+							'type' => 'string',
+						],
+					],
+				],
+			],
+			'required' => [
```
senadir: WP schema validator handles required prop differnelty, it must be defined at the prop level not on a separate one

> Version 3 Syntax
While WordPress mostly follows JSON Schema Version 4, one way that it doesn’t is with the syntax for defining required properties. The predominant way is using JSON Schema Version 3 syntax by adding the required keyword to each property’s definition.
https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/

gigitux: Oh nice catch! I fixed it with fed56eca1f77c5b5d394a950901ff3aedf2aad81

plugins/woocommerce/src/Blocks/Patterns/PTKClient.php
```diff
@@ -14,6 +14,67 @@ class PTKClient {
...
+							'type'       => 'object',
+							'properties' => [
+								'slug'        => [
+									'type'     => 'string',
+									'required' => true,
+								],
+								'title'       => [
+									'type'     => 'string',
+									'required' => true,
+								],
```
Aljullu, line 65: Should we enforce a [`minLength`](https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/#strings) for `slug` and `title` to make sure they are not empty strings?

plugins/woocommerce/src/Blocks/Patterns/PTKClient.php
```diff
@@ -14,6 +14,67 @@ class PTKClient {
...
+						'required' => true,
+					],
+					'name'       => [
+						'type'     => 'string',
+						'required' => true,
+					],
+					'html'       => [
+						'type'     => 'string',
+						'required' => true,
+					],
```
Aljullu, line 52: Similarly here, should we enforce a `minLength` just in case?

plugins/woocommerce/tests/php/src/Blocks/Patterns/PTKClientTest.php
```diff
@@ -147,6 +147,96 @@ function () {
...
+                                }
+                            }
+                        },
+						{
+                            "ID": 14870,
+                            "site_id": 174455321,
+                            "title": "Review: A quote with scattered images",
+                            "name": "review-a-quote-with-scattered-images",
+                            "html": "<!-- /wp:spacer -->",
+                            "categories": null,
```
Aljullu: In the test below, we are already testing that `fetch_patterns` returns an error when `categories` is null. My understanding is that in this test we are trying to verify that it also fails when a category is defined but `slug` or `title` are not strings (L167-168), am I right? In that case, I would define this `categories` property, so we make sure this only fails because of the missing `slug` and `title` in the `testimonials` category. Does it make sense?

plugins/woocommerce/tests/php/src/Blocks/Patterns/PTKClientTest.php
```diff
@@ -147,6 +147,96 @@ function () {
...
+                            "categories": {
+                                "testimonials": {
+                                    "slug": null,
+                                    "title": null,
+                                    "description": null
+                                }
+                            }
+                        },
+						{
+                            "ID": 14870,
```
Aljullu: Should the `ID` of the two patterns be a different number?

Aljullu (commented):
Thanks for working on this, @gigitux! I left a few small suggestions, but overall it's looking good on my end. I tested clearing the transients and the patterns are correctly set when the data is valid.

plugins/woocommerce/tests/php/src/Blocks/Patterns/PTKClientTest.php
gigitux: It makes sense! I updated it with 66041700cd9fa296f6bd5b571b691f44d4a7be14

plugins/woocommerce/tests/php/src/Blocks/Patterns/PTKClientTest.php
gigitux: It does! Fixed with bc579217d9556cfc1395c2fe1efced69cfb494ff

plugins/woocommerce/src/Blocks/Patterns/PTKClient.php
gigitux: I don't think that we should do it. The important thing is that the type is a string.

plugins/woocommerce/src/Blocks/Patterns/PTKClient.php
gigitux: Same here

Aljullu (approved):
Not a strong opinion on `minLength`, so it looks good to me. Thanks for improving this, @gigitux! :raised_hands: 

imatge

<details>

* Plugins: WooCommerce Beta Tester, Transients Manager
* Theme: Twenty Twenty-Five
* Store size: small
</details>

gigitux:
@senadir @tjcafferkey do you want to take another look? Do you think that we should open a CFE for WooCommerce 9.9 release? I think that we can postpone it.

gigitux:
> @senadir @tjcafferkey do you want to take another look? Do you think that we should open a CFE for WooCommerce 9.9 release? I think that we can postpone it.

@jorgeatorres, sorry... I forgot to tag you. Do you have some opinion on this?

jorgeatorres:
> @jorgeatorres, sorry... I forgot to tag you. Do you have some opinion on this?

No worries @gigitux, it's not really up to me 😺. I think the fix we already have in place should be good enough unless something incredibly strange happens again. It's better than what we had running for a long time. If that were the only criteria, I'd say this doesn't warrant a CFE as it instead introduces various changes too close to the release, but I'll obviously defer to you all.

gigitux:
I'm going to merge this PR given that it will go in WooCommerce 10.0 release. We will have time to revert/improve it if some defects are found.

Issue #57725, opened 2025-05-05
Blockified Add to Cart with Options block: show error notice when adding to cart fails


Aljullu:
When adding a product to cart using the new *Add to Cart with Options* block, we need to make it so we display an error notice if the action fails. Ideally, we should use the same approach as the *Product Collection* block.

luisherranz:
Related discussion: 
- https://github.com/woocommerce/woocommerce/discussions/55731

PR #57816, opened 2025-05-07
[Experimental] Display a not-allowed cursor when trying to Add to Cart an invalid variation

### LINKED ISSUE #57724: Blockified Add to Cart with Options block: make Add to Cart button disabled until a valid variation has been selected
In classic themes, the *Add to Cart* button is disabled until a valid variation has been selected:

We should do the same in the new blockified *Add to Cart with Options* block.

Aljullu:

Closes #57724.
Closes WOOPLUG-4151.

This PR makes it so we know when the selected attributes in the _Variation Selector_ block match a valid variation. This way, we can change the cursor of the Add to Cart button to `not-allowed`. While my initial plan was to completely disable the Add to Cart button like in the classic templates, after researching it, it turns out that's not good for accessibility reasons, so it's better to still allow focusing on the button and clicking it.

Note: displaying errors will be handled in https://github.com/woocommerce/woocommerce/issues/57725.

#### Preparation

#### Testing the PR

1. Create a variable product where some attribute combinations are allowed, and some others aren't.
2. Visit the product template in the frontend. Verify by default you see a `not allowed` cursor when hovering the add to cart button.
3. Select some attributes. Verify when they match a valid variation, the cursor on the button is no longer `not allowed`.

https://github.com/user-attachments/assets/75c4d1b5-5db2-4b83-a1fc-5a3b9b6ee108

### DIFF
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/AddToCartWithOptions.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/AddToCartWithOptions.php
@@ -167,26 +167,30 @@ class AddToCartWithOptions extends AbstractBlock {
 			$default_quantity = apply_filters( 'woocommerce_add_to_cart_quantity', 1, $product->get_id() );
 
 			$context = array(
-				'productId' => $product->get_id(),
-				'quantity'  => $default_quantity,
+				'productId'   => $product->get_id(),
+				'productType' => $product->get_type(),
+				'quantity'    => $default_quantity,
 			);
 
-			if ( $product instanceof \WC_Product && $product->is_type( 'variable' ) ) {
+			if ( $product->is_type( 'variable' ) ) {
 				$context['variation']           = array();
 				$context['availableVariations'] = $product->get_available_variations();
 			}
 
+			$wrapper_attributes = array(
+				'data-wp-interactive'       => 'woocommerce/add-to-cart-with-options',
+				'data-wp-context'           => wp_json_encode(
+					$context,
+					JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_AMP
+				),
+				'data-wp-on--submit'        => 'actions.handleSubmit',
+				'data-wp-class--is-invalid' => '!state.isFormValid',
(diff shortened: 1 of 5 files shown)


plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/AddToCartWithOptions.php
```diff
@@ -167,25 +167,27 @@ protected function render( $attributes, $content, $block ) {
...
 			$context = array(
-				'productId' => $product->get_id(),
-				'quantity'  => $default_quantity,
+				'productId'   => $product->get_id(),
+				'productType' => $product->get_type(),
+				'quantity'    => $default_quantity,
 			);
 
-			if ( $product instanceof \WC_Product && $product->is_type( 'variable' ) ) {
+			if ( $product->is_type( 'variable' ) ) {
```
Aljullu, line 175: We already checked that `$product` is an instance of `WC_Product` here:

https://github.com/woocommerce/woocommerce/blob/331801fadcb5de5cb6c94bc6ae53c5ca33fb5d69/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/AddToCartWithOptions.php#L103

So there was no need to repeat the check.

gigitux:
Code-wise the PR LGTM, but I think that we should disable the button if the form isn't valid:


https://github.com/user-attachments/assets/64bc3944-0bf3-476b-9ec0-2e70b37e156a

Aljullu:
Thanks for taking a look @gigitux!

> I think that we should disable the button if the form isn't valid

How do you feel if we take the decision after #57725 has been implemented? I agree that right now it looks broken, but once we have inline notices, the experience should be better. My initial inclination was also to disable the button, but as I was implementing it, I stumbled upon [this article](https://adamsilver.io/blog/the-problem-with-disabled-buttons-and-what-to-do-instead/) and some others, that suggest it's better not to disable the submit button. ChatGPT and Baymard gave me similar answers.

gigitux (approved):
> How do you feel if we take the decision after https://github.com/woocommerce/woocommerce/issues/57725 has been implemented?

All good. If you're aware of this, I'm happy to approve this PR! Thanks! 🚀

PR #57838, opened 2025-05-07
Take in account refunds when calculating Cost of Goods Sold for an order

Konamiman:

This pull request adds support for Cost of Goods Sold to orders with refunds:

- The value calculated by `calculate_cogs_total_value` for an order will have the combined cost of the refunded items substracted.
- The cost value for an order will be recalculated when a refund is created with `wc_create_refund`.
- The REST API endpoint that returns the refunds of an order (`/wp-json/wc/v3/orders/<order id>/refunds`, note that it's for v3 only) now includes the cost value for the refund (the one substracted from the order cost) as well as the cost of each line item of the refund.
- The "cost" column in the order details page in admin will include the cost associated to the refunded items in each line.
- Additionally, hovering the "cost" column will display a tooltip with the cost per unit for the product in the line. The tooltip can be customized by using the new `woocommerce_order_item_cogs_per_item_tooltip` filter.

image

image

1. Enable the Cost of Goods Sold feature (WooCommerce - Settings - Advanced - Features).

2. Create a few products with a non-zero value for the Cost of Goods field (see https://github.com/woocommerce/woocommerce/pull/54399).

3. In the admin area create an order with the above products, about 10 units of each so that you can create multiple refunds.

4. Issue a few partial refunds. Verify that the "Cost" column now displays the individual cost of the refunded item times the number of refunded items.

5.  Also verify that after creating each refund the total cost value of the order (displayed in the summary area under the order items as "Cost total") updates (gets decreased) accordingly.

6. Hover the mouse pointer in the "Cost" column of one of the line items with refunds and verify that the tooltip appears with the correct value.

7. You can test the new filter with the following snippet:

```php
add_filter( 'woocommerce_order_item_cogs_per_item_tooltip', function($tooltip_text, $cost_per_item, $formatted_cost_per_item, $order_item ) {
	return "Product " . $order_item->get_product_id() . ", cost per item $cost_per_item (formatted: $formatted_cost_per_item)";
}, 10, 4);
```

8. Issue a `GET` request to the following REST API endpoint: `/wp-json/wc/v3/orders/<order id>/refunds`. Verify that there's a `"cost_of_goods_sold"` field with the correct value for each refund, and also for each product line item within each refund.

9. Issue a `OPTIONS` request to the same endpoint, verify that the `"cost_of_goods_sold"` fields for both the refunds and their line items get a proper schema entry.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


prettyboymp (approved):
Tested well.  I didn't see anything specific that I have concerns with.  

Looks like you'll need to apply an ignore comment to the lint rule.  Preapproving.

Konamiman:
@prettyboymp Thank you. Instead of adding a linting ignore for that `$object` argument I renamed it to `$data_object` which is supposedly the right thing to do ™️ .

PR #57847, opened 2025-05-08
PTK Pattern: improve guard

gigitux:

This PR improves the guard implemented with #57759.

The malformed data was structured in this way:

```json
"categories": {
  "": {
    "slug": null,
    "title": null,
    "description": null
  }
}  
```

Instead, the correct structure is:

```json
"categories": {
  "category1": {
    "slug": "string",
    "title": "string",
    "description": "string"
    }, 
  "category2": {
    "slug": "string",
    "title": "string",
    "description": "string"
  },
 }
```

For this reason, we need to check for the presence of the categories object and ensure that each category has the "title" and "slug" defined. These two values are used in the codebase:

https://github.com/woocommerce/woocommerce/blob/088e63b775d0bab1d6fad00d94c81ed77404f122/plugins/woocommerce/src/Blocks/Patterns/PTKPatternsStore.php#L244-L262

https://github.com/woocommerce/woocommerce/blob/088e63b775d0bab1d6fad00d94c81ed77404f122/plugins/woocommerce/src/Blocks/BlockPatterns.php#L270-L283

This is just a better patch, but the right implemented will be implemented with the #57809

1. [Install `Transients Manager` plugin ](https://wordpress.org/plugins/transients-manager/).
2. Visit `wp-admin/tools.php?page=transients-manager`.
3. Click on `Delete All`.
4. Visit `wp-admin/admin.php?page=wc-settings&tab=advanced&section=woocommerce_com`
5. Enable tracking.
6. Save.
7. Visit `wp-admin/admin.php?page=wc-admin`.
8. Click on `Customize Your Store.
9. Click on `Use the store designer`.
10. Click on `Design Your Homepeage`.
11. Click on `Intro`.
12. Expect to see more than 4 patterns:
https://github.com/user-attachments/assets/c4bc2dae-4852-457e-9f0a-326c640284e6

13. Replace [this line](https://github.com/woocommerce/woocommerce/blob/72863ce6d4c2bdf735d868790184df2431193932/plugins/woocommerce/src/Blocks/BlockPatterns.php#L221) with:

```php
		$patterns = json_decode(
			"[\n\t{\n\t\t\"ID\": 450,\n\t\t\"site_id\": 231214916,\n\t\t\"title\": \"Intro: Heading with image and two columns above\",\n\t\t\"name\": \"intro-heading-with-image-and-two-columns-above-copy\",\n\t\t\"description\": \"\",\n\t\t\"ai_description\": \"\",\n\t\t\"html\": \"<!-- wp:group {\\\"metadata\\\":{\\\"name\\\":\\\"Intro\\\"},\\\"align\\\":\\\"full\\\",\\\"style\\\":{\\\"spacing\\\":{\\\"padding\\\":{\\\"top\\\":\\\"calc( 0.5 * var(\\\\u002d\\\\u002dwp\\\\u002d\\\\u002dstyle\\\\u002d\\\\u002droot\\\\u002d\\\\u002dpadding-right, var(\\\\u002d\\\\u002dwp\\\\u002d\\\\u002dcustom\\\\u002d\\\\u002dgap\\\\u002d\\\\u002dhorizontal)))\\\",\\\"bottom\\\":\\\"calc( 0.5 * var(\\\\u002d\\\\u002dwp\\\\u002d\\\\u002dstyle\\\\u002d\\\\u002droot\\\\u002d\\\\u002dpadding-right, var(\\\\u002d\\\\u002dwp\\\\u002d\\\\u002dcustom\\\\u002d\\\\u002dgap\\\\u002d\\\\u002dhorizontal)))\\\",\\\"left\\\":\\\"var(\\\\u002d\\\\u002dwp\\\\u002d\\\\u002dstyle\\\\u002d\\\\u002droot\\\\u002d\\\\u002dpadding-left, var(\\\\u002d\\\\u002dwp\\\\u002d\\\\u002dcustom\\\\u002d\\\\u002dgap\\\\u002d\\\\u002dhorizontal))\\\",\\\"right\\\":\\\"var(\\\\u002d\\\\u002dwp\\\\u002d\\\\u002dstyle\\\\u002d\\\\u002droot\\\\u002d\\\\u002dpadding-right, var(\\\\u002d\\\\u002dwp\\\\u002d\\\\u002dcustom\\\\u002d\\\\u002dgap\\\\u002d\\\\u002dhorizontal))\\\"},\\\"margin\\\":{\\\"top\\\":\\\"0\\\",\\\"bottom\\\":\\\"0\\\"}}},\\\"layout\\\":{\\\"type\\\":\\\"constrained\\\",\\\"justifyContent\\\":\\\"center\\\"}} -->\\n<div class=\\\"wp-block-group alignfull\\\" style=\\\"margin-top:0;margin-bottom:0;padding-top:calc( 0.5 * var(--wp--style--root--padding-right, var(--wp--custom--gap--horizontal)));padding-right:var(--wp--style--root--padding-right, var(--wp--custom--gap--horizontal));padding-bottom:calc( 0.5 * var(--wp--style--root--padding-right, var(--wp--custom--gap--horizontal)));padding-left:var(--wp--style--root--padding-left, var(--wp--custom--gap--horizontal))\\\"><!-- wp:spacer {\\\"height\\\":\\\"calc( 0.25 * var(\\\\u002d\\\\u002dwp\\\\u002d\\\\u002dstyle\\\\u002d\\\\u002droot\\\\u002d\\\\u002dpadding-right, var(\\\\u002d\\\\u002dwp\\\\u002d\\\\u002dcustom\\\\u002d\\\\u002dgap\\\\u002d\\\\u002dhorizontal)))\\\"} -->\\n<div style=\\\"height:calc( 0.25 * var(--wp--style--root--padding-right, var(--wp--custom--gap--horizontal)))\\\" aria-hidden=\\\"true\\\" class=\\\"wp-block-spacer\\\"></div>\\n<!-- /wp:spacer -->\\n\\n<!-- wp:group {\\\"align\\\":\\\"wide\\\",\\\"layout\\\":{\\\"type\\\":\\\"default\\\"}} -->\\n<div class=\\\"wp-block-group alignwide\\\"><!-- wp:columns {\\\"verticalAlignment\\\":\\\"top\\\",\\\"align\\\":\\\"wide\\\"} -->\\n<div class=\\\"wp-block-columns alignwide are-vertically-aligned-top\\\"><!-- wp:column {\\\"verticalAlignment\\\":\\\"top\\\",\\\"width\\\":\\\"\\\",\\\"layout\\\":{\\\"type\\\":\\\"constrained\\\",\\\"justifyContent\\\":\\\"left\\\",\\\"contentSize\\\":\\\"480px\\\"}} -->\\n<div class=\\\"wp-block-column is-vertically-aligned-top\\\"><!-- wp:heading -->\\n<h2 class=\\\"wp-block-heading\\\">Stay dry in style</h2>\\n<!-- /wp:heading --></div>\\n<!-- /wp:column -->\\n\\n<!-- wp:column {\\\"verticalAlignment\\\":\\\"top\\\",\\\"width\\\":\\\"40%\\\"} -->\\n<div class=\\\"wp-block-column is-vertically-aligned-top\\\" style=\\\"flex-basis:40%\\\"><!-- wp:paragraph -->\\n<p>Designed for both fashion and function, our rain jackets offers sleek styles and waterproof protection to keep you dry in any downpour.</p>\\n<!-- /wp:paragraph -->\\n\\n<!-- wp:buttons -->\\n<div class=\\\"wp-block-buttons\\\"><!-- wp:button --><div class=\\\"wp-block-button\\\"><a class=\\\"wp-block-button__link wp-element-button\\\">Discover collection</a></div><!-- /wp:button --></div>\\n<!-- /wp:buttons --></div>\\n<!-- /wp:column --></div>\\n<!-- /wp:columns -->\\n\\n<!-- wp:spacer {\\\"height\\\":\\\"calc( 0.25 * var(\\\\u002d\\\\u002dwp\\\\u002d\\\\u002dstyle\\\\u002d\\\\u002droot\\\\u002d\\\\u002dpadding-right, var(\\\\u002d\\\\u002dwp\\\\u002d\\\\u002dcustom\\\\u002d\\\\u002dgap\\\\u002d\\\\u002dhorizontal)))\\\"} -->\\n<div style=\\\"height:calc( 0.25 * var(--wp--style--root--padding-right, var(--wp--custom--gap--horizontal)))\\\" aria-hidden=\\\"true\\\" class=\\\"wp-block-spacer\\\"></div>\\n<!-- /wp:spacer -->\\n\\n<!-- wp:image {\\\"id\\\":75,\\\"aspectRatio\\\":\\\"16/9\\\",\\\"scale\\\":\\\"cover\\\",\\\"sizeSlug\\\":\\\"full\\\",\\\"linkDestination\\\":\\\"none\\\"} -->\\n<figure class=\\\"wp-block-image size-full\\\"><img src=\\\"https://wooblockpatterns.wpcomstaging.com/wp-content/uploads/2024/06/man-person-winter-photography-statue-coat.png\\\" alt=\\\"\\\" class=\\\"wp-image-75\\\" style=\\\"aspect-ratio:16/9;object-fit:cover\\\"/></figure>\\n<!-- /wp:image --></div>\\n<!-- /wp:group -->\\n\\n<!-- wp:spacer {\\\"height\\\":\\\"calc( 0.25 * var(\\\\u002d\\\\u002dwp\\\\u002d\\\\u002dstyle\\\\u002d\\\\u002droot\\\\u002d\\\\u002dpadding-right, var(\\\\u002d\\\\u002dwp\\\\u002d\\\\u002dcustom\\\\u002d\\\\u002dgap\\\\u002d\\\\u002dhorizontal)))\\\"} -->\\n<div style=\\\"height:calc( 0.25 * var(--wp--style--root--padding-right, var(--wp--custom--gap--horizontal)))\\\" aria-hidden=\\\"true\\\" class=\\\"wp-block-spacer\\\"></div>\\n<!-- /wp:spacer --></div>\\n<!-- /wp:group -->\",\n\t\t\"disallow_after\": [],\n\t\t\"dependencies\": [],\n\t\t\"categories\": {\n\t\t\t\"\": {\n\t\t\t\t\"slug\": null,\n\t\t\t\t\"title\": null,\n\t\t\t\t\"description\": null\n\t\t\t}\n\t\t},\n\t\t\"virtual_theme_categories\": [],\n\t\t\"tags\": [],\n\t\t\"pattern_meta\": [],\n\t\t\"source_url\": \"https://wooblockpatterns.wpcomstaging.com/2024/07/10/intro-heading-with-image-and-two-columns-above-copy/\",\n\t\t\"modified_date\": \"2025-05-06 14:29:32\",\n\t\t\"preview_data\": [],\n\t\t\"post_type\": \"wp_block\",\n\t\t\"is_included_in_big_sky\": 0\n\t}]",
			true
		);
```

15. Refresh WordPress. 
16. Ensure that you don't see any fatal. 
17. Click on `Design Your Homepeage`.
18. Click on `Intro`.
19. You should see only 4 patterns.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

PTK Pattern: improve guard

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/src/Blocks/BlockPatterns.php
+++ b/plugins/woocommerce/src/Blocks/BlockPatterns.php
@@ -261,7 +261,7 @@ class BlockPatterns {
 				$pattern['categories'] = array_map(
 					function ( $category ) {
 						foreach ( self::CATEGORIES_PREFIXES as $prefix ) {
-							if ( strpos( $category['title'] ?? '', $prefix ) !== false ) {
+							if ( strpos( $category['title'], $prefix ) !== false ) {
 								$parsed_category   = str_replace( $prefix, '', $category['title'] );
 								$parsed_category   = str_replace( '_', ' ', $parsed_category );
 								$category['title'] = ucfirst( $parsed_category );
--- a/plugins/woocommerce/src/Blocks/BlockPatterns.php
+++ b/plugins/woocommerce/src/Blocks/BlockPatterns.php
@@ -242,8 +242,10 @@ class BlockPatterns {
 	 * @param array $patterns The patterns to parse.
 	 * @return array The parsed patterns.
 	 */
-	private function parse_categories( array $patterns ) {
-
+	private function parse_categories( $patterns ) {
+		if ( ! is_array( $patterns ) ) {
+			return array();
+		}
 		return array_map(
 			function ( $pattern ) {
 				if ( ! isset( $pattern['categories'] ) ) {
--- a/plugins/woocommerce/src/Blocks/BlockPatterns.php
+++ b/plugins/woocommerce/src/Blocks/BlockPatterns.php
@@ -243,16 +243,25 @@ class BlockPatterns {
 	 * @return array The parsed patterns.
 	 */
 	private function parse_categories( array $patterns ) {
-		if ( ! isset( $patterns['categories'] ) || ! is_array( $patterns['categories'] ) ) {
-			return array();
-		}
 
 		return array_map(
 			function ( $pattern ) {
+				if ( ! isset( $pattern['categories'] ) ) {
+					$pattern['categories'] = array();
+				}
+
+				$values = array_values( $pattern['categories'] );
+
+				foreach ( $values as $value ) {
+					if ( ! isset( $value['title'] ) || ! isset( $value['slug'] ) ) {
+						$pattern['categories'] = array();
+					}
+				}
+
 				$pattern['categories'] = array_map(
 					function ( $category ) {
 						foreach ( self::CATEGORIES_PREFIXES as $prefix ) {
-							if ( strpos( $category['title'], $prefix ) !== false ) {
+							if ( strpos( $category['title'] ?? '', $prefix ) !== false ) {
 								$parsed_category   = str_replace( $prefix, '', $category['title'] );
 								$parsed_category   = str_replace( '_', ' ', $parsed_category );
 								$category['title'] = ucfirst( $parsed_category );
--- a/plugins/woocommerce/src/Blocks/BlockPatterns.php
+++ b/plugins/woocommerce/src/Blocks/BlockPatterns.php
@@ -210,7 +210,7 @@ class BlockPatterns {
 		$has_scheduled_action = function_exists( 'as_has_scheduled_action' ) ? 'as_has_scheduled_action' : 'as_next_scheduled_action';
 
 		$patterns = $this->ptk_patterns_store->get_patterns();
-		if ( empty( $patterns ) ) {
+		if ( empty( $patterns ) || ! is_array( $patterns ) ) {
 			// Only log once per day by using a transient.
 			$transient_key = 'wc_ptk_pattern_store_warning';
 			// By only logging when patterns are empty and no fetch is scheduled,
--- a/plugins/woocommerce/changelog/57847-fix-implement-better-patch-cache
+++ b/plugins/woocommerce/changelog/57847-fix-implement-better-patch-cache
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+PTK Pattern: improve guard
--- a/plugins/woocommerce/tests/e2e-pw/tests/customize-store/assembler/homepage.spec.js
+++ b/plugins/woocommerce/tests/e2e-pw/tests/customize-store/assembler/homepage.spec.js
@@ -326,7 +326,7 @@ test.describe( 'Homepage tracking banner', { tag: tags.NOT_E2E }, () => {
 		).toBeVisible();
 	} );
 
-	test.skip(
+	test(
 		'Should not show the "Want more patterns?" banner when tracking is allowed',
 		{ tag: tags.SKIP_ON_PRESSABLE },
 		async ( { baseURL, pageObject } ) => {


plugins/woocommerce/src/Blocks/BlockPatterns.php
```diff
@@ -243,16 +243,25 @@ public function register_ptk_patterns() {
...
 		return array_map(
 			function ( $pattern ) {
+				if ( ! isset( $pattern['categories'] ) ) {
+					$pattern['categories'] = array();
+				}
+
+				$values = array_values( $pattern['categories'] );
+
+				foreach ( $values as $value ) {
+					if ( ! isset( $value['title'] ) || ! isset( $value['slug'] ) ) {
```
Aljullu, line 258: Not a big deal, but why doing the `title` and `slug` check here instead of inside the `array_map()`?

Aljullu (approved):
Confirming patterns are correctly loading now. I repeated the testing steps in `trunk` and this branch and this is how CYS looks:

`trunk` | This branch
--- | ---
imatge | imatge

Also, there is no fatal when data is invalid. I left one minor question, but not a blocker.


<details>

* Plugins: WooCommerce Beta Tester, Transients Manager
* Theme: Twenty Twenty-Five
* Store size: small
</details>

plugins/woocommerce/src/Blocks/BlockPatterns.php
gigitux: In the `register_block_pattern` function, we do some data manipulation on the pattern data (including categories)

https://github.com/woocommerce/woocommerce/blob/bae1debc9ab6f1cf91427736e7cbaa921c621ec0/plugins/woocommerce/src/Blocks/Patterns/PatternRegistry.php#L98-L121

https://github.com/woocommerce/woocommerce/blob/bae1debc9ab6f1cf91427736e7cbaa921c621ec0/plugins/woocommerce/src/Blocks/Patterns/PatternRegistry.php#L158-L173



To keep this PR minimal, I preferred to set an empty array, instead to repeating the same check across the codebase: does it make sense?

plugins/woocommerce/src/Blocks/BlockPatterns.php
```diff
@@ -243,16 +243,25 @@ public function register_ptk_patterns() {
...
+						$pattern['categories'] = array();
+					}
+				}
+
 				$pattern['categories'] = array_map(
 					function ( $category ) {
 						foreach ( self::CATEGORIES_PREFIXES as $prefix ) {
-							if ( strpos( $category['title'], $prefix ) !== false ) {
+							if ( strpos( $category['title'] ?? '', $prefix ) !== false ) {
```
gigitux: @Aljullu This is a leftover that I can remove.

gigitux: Removed 3ec02ef

plugins/woocommerce/src/Blocks/BlockPatterns.php
Aljullu: Yes, it's fine. I think it should be possible to avoid iterating twice over the same array (`$pattern['categories']`). But it's definitely not important, let's keep it simple so we can ship the fix.

Aljullu (approved):
> Removed https://github.com/woocommerce/woocommerce/commit/3ec02ef900d8f7eabadee94886c406916bd1e564

Rested just in case and it looks good. :+1:

jorgeatorres:
Can we add a quick check in `get_patterns()` when getting the transient to at least make sure the transient is an array before we try to process it?
I know in theory the transient shouldn't be any other thing and that proper validation (in 9.9) will address this case, but I'd feel better if something as simple as changing the transient to a string didn't produce a fatal.

For example `wp transient set ptk_patterns 'foobar'` can brick a site. wdyt @gigitux?

gigitux:
> Can we add a quick check in `get_patterns()` when getting the transient to at least make sure the transient is an array before we try to process it? I know in theory the transient shouldn't be any other thing and that proper validation (in 9.9) will address this case, but I'd feel better if something as simple as changing the transient to a string didn't produce a fatal.
> 
> For example `wp transient set ptk_patterns 'foobar'` can brick a site. wdyt @gigitux?

It makes sense! I added it with 9ede59178ef20424cce9bfa1b353430867deab4b

plugins/woocommerce/src/Blocks/BlockPatterns.php
```diff
@@ -243,12 +243,21 @@ public function register_ptk_patterns() {
 	 * @return array The parsed patterns.
 	 */
 	private function parse_categories( array $patterns ) {
```
tjcafferkey: Even though we check before its usage that `$patterns` is an array, do you think we should add more type checking in here prior to processing?

```php
if ( ! is_array( $patterns ) || empty( $patterns ) ) {
    return array();
}
```

gigitux: `parse_categories` is used only in this class and invoked after the guard that we added. So I think that we're safe.

tjcafferkey: I personally think its good practice even if we have type checked the value before passing it to the method to also type check within the method just incase some of this code gets refactored in future and this trips us up again.

plugins/woocommerce/src/Blocks/BlockPatterns.php
```diff
@@ -243,12 +243,21 @@ public function register_ptk_patterns() {
...
-		if ( ! isset( $patterns['categories'] ) || ! is_array( $patterns['categories'] ) ) {
-			return array();
-		}
 
 		return array_map(
 			function ( $pattern ) {
```
tjcafferkey, line 250: Again I'd type check `$pattern` is an array here before operating

gigitux:
> I personally think its good practice even if we have type checked the value before passing it to the method to also type check within the method just incase some of this code gets refactored in future and this trips us up again.

@tjcafferkey I’m happy to add the check, though I think there’s room for discussion to adopt this approach as general pattern. In my view, repeating type checks within internal methods might lead to unnecessary duplication. A more scalable pattern could be to ensure that all external inputs (e.g., from DBs or APIs) are validated at the boundaries using a schema (see https://github.com/woocommerce/woocommerce/pull/57809)

plugins/woocommerce/src/Blocks/BlockPatterns.php
gigitux: Also, I would refactor this line:

```
private function parse_categories( array $patterns ) {
```

to:
 ```
private function parse_categories( $patterns ) {
```

If for some reason `$patterns` isn't an array, it would trigger a fatal error - WDYT?

tjcafferkey:
>A more scalable pattern could be to ensure that all external inputs (e.g., from DBs or APIs) are validated at the boundaries using a schema (see https://github.com/woocommerce/woocommerce/pull/57809)

I agree

>I’m happy to add the check, though I think there’s room for discussion to adopt this approach as general pattern. In my view, repeating type checks within internal methods might lead to unnecessary duplication.

I can see that pov for sure, but given we don't yet have that other pattern in place for this release I think it's a worthy trade-off, and covers us for any refactors.

plugins/woocommerce/src/Blocks/BlockPatterns.php
gigitux: @tjcafferkey Done with 39e8cf8

tjcafferkey (approved):

senadir (approved):
Thank you for the changes Luigi, the rest left good feedback, from my part, this is working well, getting CYS to load was a challenge but did get it work. 

I think the best guarding approach should be via a schema check that you have planned for 9.9

PR #57848, opened 2025-05-08
[Woo POS] Add POS settings for store details and display them in POS completed order HTML and plaintext email templates

jaclync:

Closes WOOMOB-227

This pull request introduces a new settings tab for Point of Sale (POS) functionality in WooCommerce, along with updates to HTML/plaintext email templates to include POS-specific store details in POS completed order emails. After this PR is approved, similar changes will be applied to the POS refunded order emails separately in WOOMOB-464. Block email templates will be updated separately in WOOMOB-460 as the way to show additional data is different.

### POS Settings:

* Added a new settings tab, `WC_Settings_Point_Of_Sale`, to manage POS store details such as store name, address, phone number, email, and refund policy. These settings are accessible when the `point-of-sale` feature flag is enabled.

### Email Template Updates:

* Updated the subject and heading of POS completed order emails to include the store name from the POS settings.
* Enhanced email content (both HTML and plaintext) to display POS store details (e.g., name, email, phone, address, refund policy) when available. Some copy in the templates were also updated to target POS customers as well.

POS settings:

Default | All filled in
-- | --
 | Screenshot 2025-05-08 at 5 23 09 PM

Completed order email:

Above order summary | Below order summary
-- | --
Screenshot 2025-05-08 at 5 15 56 PM | Screenshot 2025-05-08 at 5 15 31 PM

Prerequisites: the store has a way to see email logs (like with [WP Mail Logging plugin](https://wordpress.org/plugins/wp-mail-logging/)) or send emails.

- Run PR branch in development environment (e.g. `pnpm run --filter="@woocommerce/plugin-woocommerce" watch:build`), or test with Woo Beta Tester plugin and enable the `point-of-sale` admin feature flag (example p1746734640520419/1746734508.133719-slack-C070SJRA8DP)
- Go to WC settings --> there should be a new section "Point of Sale"
- Go to Point of Sale settings tab --> it should have 5 fields: store name, address, phone number, email, refund & returns policy. Store name, address, and email should have default values from the store WP/WC settings
- Make an API request to create a POS order ("created_via": "pos-rest-api") with a billing email and metadata for the new fields --> the request should succeed with an order ID in the response

<details>
<summary>Click to see example request body</summary>

```json
{
    "status": "completed",
    "currency": "USD",
    "billing": {
        "first_name": "",
        "last_name": "",
        "company": "",
        "address_1": "",
        "address_2": "",
        "city": "",
        "state": "",
        "postcode": "",
        "country": "",
        "email": "test@email.co",
        "phone": ""
    },
    "meta_data": [
        {
            "key": "_wc_order_attribution_source_type",
            "id": 0,
            "value": "mobile_app"
        },
        {
            "key": "_cash_change_amount",
            "value": "12.16"
        },
        {
            "key": "_charge_id",
            "value": "auth2j9wq8"
        }
    ],
    "line_items": [
        {
            "id": 0,
            "product_id": 40,
            "quantity": 2
        },
        {
            "id": 0,
            "product_id": 41,
            "quantity": 1
        }
    ],
    "created_via": "pos-rest-api"
}
```

</details>

- With the ID of the created order, make a request to send POS completed order email `POST /wc/v3/orders/$orderId/actions/send_email` with request body `{"template_id": "customer_pos_completed_order"}` --> it should succeed with a message like "Email template &quot;POS completed order&quot; sent to ...". In email logs or if the store can send email, check the email that was just sent. The email should look as in the screenshots. Below the order summary, there should be a section with heading that shows the store name from POS settings, and email/phone/address below when available in POS settings. If refund & returns policy is set in POS settings, there should be a section below the store details with a heading and the policy text
- Repeat the previous step after adding/removing/updating details in POS settings

- I tested that the POS settings tab does not appear in core environment.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


albarin:
@jaclync I've noticed something in my first test: if you enable the `point-of-sale` feature and then follow the testing steps (without saving the form on `wp-admin/admin.php?page=wc-settings&tab=point-of-sale`, just using the default), then the email is sent without any store name on the subject and nothings shows after the order summary.
After saving, though, the default fields show on the email:
| Before saving | After saving |
|--------|--------|
| Screenshot 2025-05-09 at 11 57 16 | Screenshot 2025-05-09 at 12 00 11 | 

I will keep going with the review and check the code. Thanks!

jaclync:
> I've noticed something in my first test: if you enable the `point-of-sale` feature and then follow the testing steps (without saving the form on `wp-admin/admin.php?page=wc-settings&tab=point-of-sale`, just using the default), then the email is sent without any store name on the subject and nothings shows after the order summary.
> After saving, though, the default fields show on the email:

Thanks for testing and noticing this issue! I think it's from the options not being initialized before the first settings save and the same default value needs to be provided to the template as fallback. I'll add a test case for this issue and update the code, will comment again when the PR is ready for another look.

jaclync:
@albarin @joshheald 👋 Just updated the PR to fix the default store name issue, there are 2 other POS settings (address and email) with a default value in POS settings but not in the email content before the first save, I felt like these two don't require a default value on the receipts unless the merchant has confirmed them. I'm open to other thoughts, though! Also added some test cases for the email changes.

albarin:
Thanks @jaclync! The update fixed the issue with the empty store name in the email subject, but I'm still not seeing the default values of store name, email, and physical address below the order summary _unless I save the form_ 🤔 
Screenshot 2025-05-12 at 11 45 48

jaclync:
> The update fixed the issue with the empty store name in the email subject, but I'm still not seeing the default values of store name, email, and physical address below the order summary _unless I save the form_ 🤔

Apology for not testing the full flow earlier, it looks like the options in settings are set in the database only after the first save. Therefore, all fields with a default value in settings that are used in the email templates need to also have the same default value for consistency. I updated the PR in https://github.com/woocommerce/woocommerce/pull/57848/commits/550d2d301eb11bfc071ef3670413799a5f141f25 to create a default settings helper class to provide the default value for 3 POS settings fields for the POS settings and POS completed order email, please feel free to suggest other approaches for sharing the default values (or not sharing them). In a future PR, the default values will be used in the POS refunded email in WOOMOB-464 as well.

For testing before the first save, I deleted the options `woocommerce_pos_*` from the `wp_options` table.

\ | settings | email
-- | -- | --
Before save | Screenshot 2025-05-13 at 9 03 20 AM | Screenshot 2025-05-13 at 9 03 57 AM
After save | Screenshot 2025-05-13 at 9 02 40 AM | Screenshot 2025-05-13 at 9 01 51 AM

Ready for another pass! 🙏

albarin (approved):
Looks good!

✅ Confirmed the email contains the default values when the form has not been saved before.
✅ Confirmed the email shows all values after saving.

jaclync:
@joshheald 👋 would you like to take a look? If you have some suggestions on the copy, I can update them when applying similar changes to the refunded emails in a future PR.

plugins/woocommerce/templates/emails/customer-pos-completed-order.php
```diff
@@ -31,24 +31,19 @@
...
 ?>
 </p>
-<?php if ( $email_improvements_enabled ) : ?>
-	<p><?php esc_html_e( 'We’ve successfully processed your order, and it’s on its way to you.', 'woocommerce' ); ?></p>
-	<p><?php esc_html_e( 'Here’s a reminder of what you’ve ordered:', 'woocommerce' ); ?></p>
-<?php else : ?>
-	<p><?php esc_html_e( 'We have finished processing your order.', 'woocommerce' ); ?></p>
-<?php endif; ?>
-<?php echo $email_improvements_enabled ? '</div>' : ''; ?>
+<p><?php esc_html_e( 'Here’s a reminder of what you’ve ordered:', 'woocommerce' ); ?></p>
```
joshheald:
```suggestion
<p><?php esc_html_e( 'Here’s a reminder of what you’ve bought:', 'woocommerce' ); ?></p>
```

"Ordered" feels strange to me in a POS context, when looking from the shopper's perspective – it sounds as though they have placed an order they expect to be fulfilled in future, when most likely they've got their purchases in hand already.

plugins/woocommerce/includes/emails/class-wc-email-customer-pos-completed-order.php
```diff
@@ -282,6 +279,61 @@ public function add_to_valid_template_classes( $valid_template_classes, $order )
...
+		/**
+		 * Get the refund and returns policy from POS settings.
+		 *
+		 * @return string
+		 */
+		private function get_pos_refund_returns_policy() {
+			return $this->format_string(
+				get_option( 'woocommerce_pos_refund_returns_policy' )
+			);
+		}
```
joshheald, line 336: Some/all of these details will need to be different for future location support.

This might just overcomplicate the work we're trying to do now, but what do you think about building a foundation for location awareness in to these functions now?

The settings screen could be limited to only ever work on the "primary" location, but if we store the data keyed to a location, perhaps that would be worthwhile to make the future work easier.

plugins/woocommerce/includes/emails/class-wc-email-customer-pos-completed-order.php
```diff
@@ -282,6 +279,61 @@ public function add_to_valid_template_classes( $valid_template_classes, $order )
...
+				get_option( 'woocommerce_pos_store_phone' )
+			);
+		}
+
+		/**
+		 * Get the store address from POS settings.
+		 *
+		 * @return string
+		 */
+		private function get_pos_store_address() {
```
joshheald, line 321: For address specifically, should we collect this with more line-level granularity like the one in the main Woo settings is done?

In particular, I think we should do it if tax calculations need the address to be broken up by line – we'll need POS location tax support quite soon.

plugins/woocommerce/src/Internal/Settings/PointOfSaleDefaultSettings.php
```diff
@@ -0,0 +1,62 @@
...
+				array(
+					'address_1' => WC()->countries->get_base_address(),
+					'address_2' => WC()->countries->get_base_address_2(),
+					'city'      => WC()->countries->get_base_city(),
+					'state'     => WC()->countries->get_base_state(),
+					'postcode'  => WC()->countries->get_base_postcode(),
+					'country'   => WC()->countries->get_base_country(),
+				),
+				"\n"
+			)
```
joshheald, line 59: I didn't see this working, because I didn't have a base address set when I started testing and I can't figure out how to reset the pos settings to defaults, but it's good to see it's here.

joshheald (approved):
@jaclync Nice work! I tested with various types of input and the email looks good. Thanks for noting that it's not got block email compatibility too.

I've made some suggestions inline, some small, some a little bigger. Let me know if you want to chat through some of the locations stuff I've mentioned.

plugins/woocommerce/includes/emails/class-wc-email-customer-pos-completed-order.php
jaclync: Thanks for raising the potential multi-location support in the future. This change involves more testing and I'd like to give it more thoughts (like maybe storing it as an array in case the location key name changes), I will look into this separately today and post a discussion about the solution. It's still the beginning of the release cycle (code freeze was this Monday), I believe we can get all the work in for release 10.0.0.

plugins/woocommerce/includes/emails/class-wc-email-customer-pos-completed-order.php
jaclync: Good point on potentially using the address here for other aspects of POS like tax calculations, originally I thought this address is just for display on the receipts and one single field allows the merchant to tweak how they want the address to appear on the receipts (like removing the country, condensing it into 1-2 lines, etc.). Since the changes to support individual address fields like in WC Settings > General aren't trivial (validation, country/state support, auto-completion support etc.), I will also leave them to a separate PR after more testing.

plugins/woocommerce/src/Internal/Settings/PointOfSaleDefaultSettings.php
jaclync: All the fields are saved to the `wp_options` table on the first save of the form, including the ones without user input set to the default value or empty string. If the default from the store base address is empty, then it remains empty even after the base address is filled in later.

plugins/woocommerce/templates/emails/customer-pos-completed-order.php
jaclync: Nice catch, updated in https://github.com/woocommerce/woocommerce/pull/57848/commits/67e2540005cdf1644b23fb1dd0ab8e02135b5a1c:

Screenshot 2025-05-15 at 9 26 30 AM

jaclync:
> I've made some suggestions inline, some small, some a little bigger. Let me know if you want to chat through some of the locations stuff I've mentioned.

Thanks for the review @joshheald! Since the PR is also approved on the Flux side, I just pushed an update to change the copy, will merge the PR shortly and look into the bigger changes separately in WOOPRD-518.

In the future, I will wait for your review before requesting review from the Flux team as the mobile team has more context about the future product plans. WDYT?

Issue #57087, opened 2025-04-08
The AJAX request in the admin interface is causing the server to freeze.


gidhublan:
### Prerequisites

- [x] I have carried out troubleshooting steps and I believe I have found a bug.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

On the WooCommerce orders and coupons pages, we are seeing the following AJAX request; however, the request does not complete and remains in a pending state for a long time. These requests accumulate over time, eventually causing the system to freeze.

https://www.mywebsite.com/wp-json/wc-analytics/products/count-low-in-stock?status=publish&page=1&per_page=1&_locale=user

### Expected behavior

Responding to the Ajax request with a 200 status code.

### Actual behavior

The related Ajax request remains in a waiting state and the system freezes.

### Steps to reproduce

You can easily create these requests on the Orders and Coupon creation page.

### WordPress Environment

<details><summary>System Status Report</summary>

```
### WordPress Environment ###

WordPress address (URL): [Redacted]
Site address (URL): [Redacted]
WC Version: 9.7.1
Legacy REST API Package Version: Eski REST API eklentisi bu sitede kurulu değil.
Action Scheduler Version: ✔ 3.9.0
Log Directory Writable: ✔
WP Version: 6.7.2
WP Multisite: –
WP Memory Limit: 16 GB
WP Debug Mode: –
WP Cron: –
Language: tr_TR
External object cache: ✔

### Server Environment ###

Server Info: Apache/2
PHP Version: 8.3.11
PHP Post Max Size: 1 GB
PHP Time Limit: 3000
PHP Max Input Vars: 100000
cURL Version: 7.76.1
OpenSSL/3.0.7

SUHOSIN Installed: –
MySQL Version: 10.4.32-MariaDB-log
Max Upload Size: 1 GB
Default Timezone is UTC: ✔
fsockopen/cURL: ✔
SoapClient: ✔
DOMDocument: ✔
GZip: ✔
Multibyte String: ✔
Remote Post: ✔
Remote Get: ✔

### Database ###

[REDACTED]

### Post Type Counts ###

attachment: 40908
custom_css: 2
customize_changeset: 2
elementor_library: 3
fusion_element: 12
fusion_form: 1
fusion_icons: 1
fusion_tb_layout: 4
fusion_tb_section: 13
fusion_template: 1
la_block: 2
la_portfolio: 5
la_team_member: 7
la_testimonial: 1
[...]

PR #57852, opened 2025-05-08
Split product stock count for custom stock into two queries to improve performance

louwie17:

This PR is the first item to improve the performance of the stock count request which gets triggered on any Woo Admin page, as we use it to add a notification to the **Activity** panel icon.

The endpoint runs one of two queries, depending if there are any products with a custom "Low stock threshold" defined. 
The query that gets run when their are "Low stock threshold" defined is inefficient and as shown in the original issue can be quite slow on a site with a very large amount of products: https://github.com/woocommerce/woocommerce/issues/57087#issuecomment-2848172801 

Splitting this query up seemed to greatly reduce the time when testing.

A follow up would be to implement caching. 

Partly addresses: #57087

| Before | After |
| ------ | ----- |
|        |       |

1. Create a store finish the onboarding and create several products including variations.
2. Enable the "Enable stock management" under **WooCommerce > Settings > Products > Inventory**
3. Go to a couple of the products and update the **Quantity** ( under Inventory tab ) to below the store wide threshold ( default is 2 I believe )
4. Go to **WooCommerce > Home** and hide your task list ( if it is still displayed ) 
5. Once task list is hidden it should show a "Stock" accordion, the number should be correct with the number of products that have a stock below the store wide threshold.
6. Now go to a couple products and variations and set the **Low stock threshold** in the **Inventory** tab, to something different to that of the store wide threshold. ( you can mix/match ).
7. Make sure the **Quantity** of the product is below the **Low stock threshold**
8. Go to **WooCommerce > Home** and make sure the count is correctly reflected.

**Bonus points**
Install and activate this plugin: [low-in-stock-debug.zip](https://github.com/user-attachments/files/20101314/low-in-stock-debug.zip)

1. Also install the WC Smooth Generator -> https://github.com/woocommerce/wc-smooth-generator
2. Create a large number of variable products over a couple thousand ( this may take a bit ) -> `wp wc generate products 2000 --type=variable`
3. Install the **Query Monitor** plugin and go to **Tools > Low In Stock Debug**
4. Look at the **Full Query with Meta** time and the time of the two combined queries below ( you can confirm with **Query Monitor** ( the time should be quite a bit faster ( less than half ) than the **Full Query with Meta** and the results should be the same ).

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/wooplug-3834-the-ajax-request-in-the-admin-interface-is-causing-the
+++ b/plugins/woocommerce/changelog/wooplug-3834-the-ajax-request-in-the-admin-interface-is-causing-the
@@ -1,0 +1,4 @@
+Significance: minor
+Type: performance
+
+Split low stock count query for custom stock into two queries to improve performance.
--- a/plugins/woocommerce/src/Admin/API/ProductsLowInStock.php
+++ b/plugins/woocommerce/src/Admin/API/ProductsLowInStock.php
@@ -74,7 +74,7 @@ final class ProductsLowInStock extends \WC_REST_Products_Controller {
 		$low_stock_threshold = absint( max( get_option( 'woocommerce_notify_low_stock_amount' ), 1 ) );
 
 		$sidewide_stock_threshold_only = $this->is_using_sitewide_stock_threshold_only();
-		$total_results = 0;
+		$total_results                 = 0;
 		if ( $sidewide_stock_threshold_only ) {
 			$count_query_string  = $this->get_count_query( $sidewide_stock_threshold_only );
 			$count_query_results = $wpdb->get_results(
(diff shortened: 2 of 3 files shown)


chihsuan (approved):
Nice improvement! 👍  I've tested the low stock flows and confirmed the results are consistent with the previous implementation. 

I can verify that splitting the stock count query into two has noticeably improved performance, especially on larger stores. Code looks clean and the logic is sound. 

---

> Once task list is hidden it should show a "Stock" accordion, the number should be correct with the number of products that have a stock below the store wide threshold.

Not related to this PR. I found that we need to create at least one order to see the stock panel.

https://github.com/woocommerce/woocommerce/blob/0363af32305d30d4f8b92b0a61dcd78a08519b47/plugins/woocommerce/client/admin/client/homescreen/activity-panel/panels.js#L57

The stock panel remains hidden when there are no orders, yet the activity panel displays an inventory notice. Clicking this notification redirects to the home page where the stock panel continues to be hidden.


Screenshot 2025-05-09 at 12 14 30

louwie17:
> Not related to this PR. I found that we need to create at least one order to see the stock panel.
> The stock panel remains hidden when there are no orders, yet the activity panel displays an inventory notice. Clicking this notification redirects to the home page where the stock panel continues to be hidden.

Thanks for calling that out, that is not ideal!
Honestly I was thinking we may want users to configure the Activity tab and be able to say when they want to be notified ( as I am sure not every store cares about inventory or orders as part of this view ).

PR #57863, opened 2025-05-08
[e2e tests] Improve page-load checks with more robust locators and error handling

adimoldovan:

Closes WOOPLUG-4228
More context: p1746708185690189/1746705044.685409-slack-C07414X4N

Update locators for the menu items in the page-loads spec to match the exact text.
Using the `exact: true` option on the locator is not an option because it would break for menu items that have a badge with a counter. This is the reason why I used a Regex, which accounts for a full text match with an optional number at the end of the string.

Other unrelated changes:
- cleaned up the setup by removing the customer creation which is not necessary
- extracted the setup and teardown from the pages loop, they were running for each page and they only need to run once
- added some error handling for easier debugging. If any of the api calls in the setup or teardown fail the errors were not helpful at all to figure out which one failed

```
pnpm env:start
pnpm test:e2e page-loads.spec.js
```


plugins/woocommerce/tests/e2e-pw/tests/basic/page-loads.spec.js
```diff
@@ -216,35 +223,59 @@ for ( const currentPage of wcPages ) {
...
+						e.data ? JSON.stringify( e.data ) : ''
+					}`
+				);
+				throw e;
 			} );
-
-		// create customer
-		await restApi
-			.post( `${ WC_API_PATH }/customers`, customer )
-			.then( ( r ) => ( customer.id = r.data.id ) );
```
alopezari, line 224: Out of curiosity, were we creating a customer that wasn't used in the test?

plugins/woocommerce/tests/e2e-pw/tests/basic/page-loads.spec.js
```diff
@@ -216,35 +223,59 @@ for ( const currentPage of wcPages ) {
...
+			// needs a Regexp on link name to match exact text and also match the possible counter
+			// E.g. should match "Orders 3" or "Orders", but should not match "Quick Orders"
 			await page
-				.locator(
-					`li.wp-menu-open > ul.wp-submenu > li a:has-text("${ currentPage.subpages[ i ].name }")`
-				)
+				.locator( 'li.wp-menu-open > ul.wp-submenu' )
+				.getByRole( 'link', {
+					name: new RegExp(
+						`^${ currentPage.subpages[ i ].name }( \\d+)?$`
```
alopezari, line 276: This might apply to every item in the WP-Admin menu, which means we might want to refactor existing tests to follow this approach or keep it in mind for new tests. Do you think it would be helpful to extract this code to a utils function we can reuse?

Something like:
```
function getPageName() {
    return new RegExp( `^${ currentPage.subpages[ i ].name }( \\d+)?$` );
}
```

alopezari (commented):
Thanks for working on this @adimoldovan!

I've reviewed the code and added a couple of questions. Will test it now 👍

plugins/woocommerce/tests/e2e-pw/tests/basic/page-loads.spec.js
adimoldovan: Yes.

plugins/woocommerce/tests/e2e-pw/tests/basic/page-loads.spec.js
adimoldovan: I don't know of any other test using this, so we're good. Usually navigation is done with direct link to the page.

alopezari: Ah cool!

alopezari (approved):
Finished testing it - LGTM!

I've conducted the following tests:
### Ensure tests work with current environment configuration
* Set up the environment with `pnpm env:start`.
* Ran the page loads test with `pnpm test:e2e page-loads.spec.js`.
* Ensured the test passed.

### Ensure tests work when two sections of the WP Admin menu contain the same word
* Set up the environment with `pnpm env:start`.
* Installed a plugin ([woocommerce-quick-orders.zip](https://github.com/user-attachments/files/20120609/woocommerce-quick-orders.zip)) that adds a new item to the WP Admin menu named `Quick Orders` under the `WooCommerce` item.
* Ran the page loads test with `pnpm test:e2e page-loads.spec.js`.
* Ensured the test passed.

PR #57865, opened 2025-05-08
[Experimental] Review Template: remove threading

dinhtungdu:

This PR cleans up the Review Template block to remove the logic handling thread because review doesn't support threading. This PR also remove the nested placeholder review for preview in the Site Editor.

Closes WOOPLUG-4205

| Before | After |
| ------ | ----- |
| image | image |

1. Add the Blockified Product Details block to the Single Product template.
2. Open the Reviews accordion item.
3. See the review list is flat, don't see the thread.

### DIFF
--- a/plugins/woocommerce/changelog/wooplug-4205-reduce-thread-depth-to-one-as-review-doesnt-support-replies
+++ b/plugins/woocommerce/changelog/wooplug-4205-reduce-thread-depth-to-one-as-review-doesnt-support-replies
@@ -1,0 +1,5 @@
+Significance: patch
+Type: update
+Comment: Expermental Review Template block: remove review thread logic.
+
+
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-template/edit.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-template/edit.tsx
@@ -27,7 +27,6 @@ import { useCommentQueryArgs, useCommentTree } from './hooks';
 
 interface Comment {
 	commentId: number;
-	children?: Comment[];
 }
 
 interface ReviewTemplateInnerBlocksProps {
@@ -51,8 +50,6 @@ type ReviewTemplateAttributes = {
 
 const TEMPLATE = [
 	[ 'core/avatar' ],
-	[ 'core/comment-reply-link' ],
-	[ 'core/comment-edit-link' ],
 	[ 'woocommerce/product-review-author-name' ],
 	[ 'woocommerce/product-review-date' ],
 	[ 'woocommerce/product-review-content' ],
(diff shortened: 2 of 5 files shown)


plugins/woocommerce/changelog/wooplug-4205-reduce-thread-depth-to-one-as-review-doesnt-support-replies
```diff
@@ -0,0 +1,5 @@
+Significance: patch
+Type: update
+Comment: Expermental Review Template block: remove review thread logic.
```
Aljullu, line 3:
```suggestion
Comment: Experimental Review Template block: remove review thread logic.
```

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/template.ts
```diff
@@ -72,12 +72,6 @@ const TEMPLATE: InnerBlockTemplate[] = [
 											fontSize: 'small',
 										},
 									],
-									[
-										'core/comment-edit-link',
-										{
-											fontSize: 'small',
-										},
-									],
```
Aljullu, line 80: Is this change related to the others?

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-template/edit.tsx
```diff
@@ -61,58 +58,28 @@ const TEMPLATE = [
 interface ReviewSettings {
 	perPage: number;
 	pageComments: boolean;
-	threadComments: boolean;
-	threadCommentsDepth: number;
 }
 
 const getCommentsPlaceholder = ( {
```
Aljullu, line 63: I think this can be now simplified to:

```TSX
const getCommentsPlaceholder = ( {
	perPage,
	pageComments,
}: ReviewSettings ): Comment[] => {
	const numberOfComments = pageComments ? Math.min( perPage, 3 ) : 3;

	return Array.from( { length: numberOfComments }, ( _, i ) => ( {
		commentId: -( i + 1 ),
	} ) );
};
```

I'm not sure if it's much clearer, but wanted to mention it just in case, as it took me a while to understand the original code. :sweat_smile:

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-template/hooks.tsx
```diff
@@ -153,39 +152,22 @@ export const useCommentQueryArgs = ( { postId }: { postId: number } ) => {
 };
 
 /**
- * Generate a tree structure of comment IDs from a list of comment entities. The
- * children of each comment are obtained from `_embedded`.
+ * Generate a tree structure of comment IDs from a list of comment entities.
+ * For product reviews, this is a flat list as reviews don't support threading.
  */
 export const useCommentTree = (
```
Aljullu: Not a big deal, but what do you think about renaming this hook to something like `useCommentList()` or even `useReviews()`? If we are no longer rendering a tree, it might be good to avoid all mentions of it? :thinking:

Aljullu (approved):
This is testing well and nice clean up of the code. :clap: 

I left some inline suggestions below. None is a blocker, so feel free to disregard, but wanted to mention them just in case. :slightly_smiling_face:

plugins/woocommerce/client/blocks/assets/js/blocks/product-reviews/inner-blocks/review-template/edit.tsx
dinhtungdu: Right!!!

Issue #57388, opened 2025-04-18
[Enhancement]: Add scope attributes to Cart and Cart Totals table headers for improved accessibility.


mahfuzurwp:
### Describe the solution you'd like

While auditing a site for accessibility, I noticed that WooCommerce’s cart-related tables contain <th> elements that are missing the scope="col" attribute. This attribute is important for accessibility, as it helps screen readers and assistive technologies associate header cells with their corresponding data cells.

**_I’d be happy to submit a pull request to implement this improvement if it would be helpful._**

**Affected templates:**

1. Cart Table (templates/cart/cart.php)

Example (current):
`<th class="product-name"><?php esc_html_e( 'Product', 'woocommerce' ); ?></th>`

Suggested (with scope):
`<th scope="col" class="product-name"><?php esc_html_e( 'Product', 'woocommerce' ); ?></th>`

2. Cart Totals Table (templates/cart/cart-totals.php)
Example (current):
`<th><?php esc_html_e( 'Subtotal', 'woocommerce' ); ?></th>`

Suggested:
`<th scope="row"><?php esc_html_e( 'Subtotal', 'woocommerce' ); ?></th>`

### Additional context

WCAG Reference:
	•	[H63: Using the scope attribute to associate header cells and data cells in data tables](https://www.w3.org/TR/WCAG20-TECHS/H63.html)

gigitux:
👋 @amberhinds, can you check this issue and share your thoughts? Thanks! 🙇

amberhinds:
Hi, thanks for the tag @gigitux.

Having a scope on table headers is important for row headers or if the table has both row and table headers. 

The [current cart table](https://github.com/woocommerce/woocommerce/blob/ad5cc8d5e8f010d670e6c80bc75db3dc57ce0f55/plugins/woocommerce/templates/cart/cart.php#L4) only has column headers and no row headers. In this instance, the screen reader already reads out the column header when navigating via arrow keys between columns. Here's a video that shows that:

https://github.com/user-attachments/assets/ac454338-894c-4c68-b582-e09ae3336eb2

So, while adding `scope="col"` to the th tags will not hurt anything, in this instance it is not necessary.

amberhinds:
Sorry, I just realized I posted the wrong video for the mentioned code. Here's a video of the cart shortcode table, which is what @mahfuzurwp was referencing in the cart.php template. The column headers are also read out on this table when navigating via the arrow keys between columns.

[...]

PR #57883, opened 2025-05-09
Improve cart table accessibility with scoped headers

mahfuzurwp:
- Added `scope="col"` to the Product, Price, Quantity, and Subtotal header cells in the `<thead>`
- Kept Remove and Thumbnail columns without scope attributes since they're not data headers
- Changed the product name cell in each table row from `<td>` to `<th scope="row">` to act as a row header
- This improves screen reader support when navigating both across rows and down columns

This PR addresses the accessibility concern raised in [#57388]

(https://github.com/woocommerce/woocommerce/issues/57388). The cart table was missing proper row and column headers, which made it difficult for screen reader users to interpret.

By adding scope="col" to appropriate column headers and changing the product name cells in each row to <th scope="row">, the table now follows accessible data table best practices.

Using the [WooCommerce Testing Instructions Guide](https://github.com/woocommerce/woocommerce/wiki/Writing-high-quality-testing-instructions):
	1.	Navigate to a cart page rendered via the [woocommerce_cart] shortcode.
	2.	Use a screen reader (e.g. VoiceOver on macOS or NVDA on Windows).
	3.	Move:
		•	Down a column (e.g. all prices): the column header should be read.
	•	Across a row (e.g. from price to quantity): the product name should be read at each cell.

- 	Manual screen reader testing with VoiceOver on macOS
- 	Verified that the table layout and frontend rendering remain unchanged
- 	Ensured compatibility with existing WooCommerce templates and hooks

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 
Improve cart table accessibility by properly defining column and row headers using scope attributes.
</details>

<details>

#### Comment 
</details>

### DIFF
--- a/plugins/woocommerce/changelog/57883-trunk
+++ b/plugins/woocommerce/changelog/57883-trunk
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+Improve cart table accessibility by properly defining column and row headers using scope attributes.
--- a/plugins/woocommerce/templates/cart/cart.php
+++ b/plugins/woocommerce/templates/cart/cart.php
@@ -12,7 +12,7 @@
  *
  * @see     https://woocommerce.com/document/template-structure/
  * @package WooCommerce\Templates
- * @version 7.9.0
+ * @version 10.0.0
  */
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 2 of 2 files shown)


opr (approved):
Hi @mahfuzurwp thanks for working on this. According to the testing instructions, it reads as if the heading should be read every time you move across or down.

> - Down a column (e.g. all prices): the column header should be read.
> - Across a row (e.g. from price to quantity): the product name should be read at each cell.

When I test, this doesn't happen each time I move down or across. I think this is expected, it should only read the heading once when entering the column/row right?

Please see the video attached to see what happens for me.

https://www.loom.com/share/4e4e35f97c6d45efa8e073c925442784

Approving because I think this is OK and expected, but double checking before proceeding with the merge. Thanks.

opr (changes requested):
Ah, I noticed the template version wasn't updated. You'll need to update `@version 7.9.0` to `@version 10.0.0`

mahfuzurwp:
> I think this is expected, it should only read the heading once when entering the column/row right?

Yes, I think it's expected! And I have also updated the template version.

opr (approved):
Thanks! Then this looks good. I think the template job is failing because this PR is created from a fork.

PR #57886, opened 2025-05-09
Consider implementing a checks to harden  woocommerce_checkout_fields

### LINKED ISSUE #57811: Consider implementing a checks to harden  `woocommerce_checkout_fields` use
This is a followup to [https://github.com/woocommerce/woocommerce/issues/57729](https://github.com/woocommerce/woocommerce/issues/57729)

Consider implementing a check to see if the `woocommerce_checkout_fields` `woocommerce_shipping_fields` and `woocommerce_billing_fields` are used, but `woocommerce_get_country_locale` or `woocommerce_get_country_locale_default` or `woocommerce_default_address_fields` isn't. If so, log a warning explaining that the locale or default address fields must also be updated to ensure the checkout form won't display in a state where orders can be placed without correct customer information.

BurakParsAydin:

Closes #57811 .

1 .Put the following into functions.php of your live theme:
```
add_filter( 'woocommerce_checkout_fields', function( $fields ) {
    $fields['billing']['billing_first_name']['placeholder'] = 'Custom Placeholder';
    return $fields;
} );
```
2. Go to Dashboard → Home 
Or Settings → General

3. Make sure you're logged in as an admin (with manage_woocommerce capability).

4. You should now see your admin warning.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/includes/admin/class-wc-admin.php
+++ b/plugins/woocommerce/includes/admin/class-wc-admin.php
@@ -337,7 +337,12 @@ class WC_Admin {
 		$uses_locale_filters = $uses_default_address_fields || $uses_get_country_locale || $uses_get_country_locale_default;
 
 		if ( $uses_field_filters && ! $uses_locale_filters ) {
-			$dismiss_url = add_query_arg( array( 'dismiss_wc_locale_notice' => '1', '_wc_notice_nonce' => wp_create_nonce( 'dismiss_wc_locale_notice' ) ) );
+			$dismiss_url = add_query_arg(
+				array(
+					'dismiss_wc_locale_notice' => '1',
+					'_wc_notice_nonce'         => wp_create_nonce( 'dismiss_wc_locale_notice' ),
+				)
+			);
 
 			echo '<div class="notice notice-warning is-dismissible">';
 			echo '<p><strong>' . esc_html__( 'WooCommerce Notice for Developers:', 'woocommerce' ) . '</strong> ';
--- a/plugins/woocommerce/includes/admin/class-wc-admin.php
+++ b/plugins/woocommerce/includes/admin/class-wc-admin.php
@@ -26,7 +26,7 @@ class WC_Admin {
 	 */
 	public function __construct() {
 		add_action( 'init', array( $this, 'includes' ) );
-		add_action( 'init', array( $this, 'wc_log_missing_locale_filters' ) );
+		add_action( 'wp_loaded', array( $this, 'wc_log_missing_locale_filters' ) );
 		add_action( 'current_screen', array( $this, 'conditional_includes' ) );
 		add_action( 'admin_init', array( $this, 'buffer' ), 1 );
 		add_action( 'admin_init', array( $this, 'preview_emails' ) );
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/includes/admin/class-wc-admin.php
```diff
@@ -289,6 +291,57 @@ public function preview_email_editor_dummy_content() {
...
+
+		$uses_default_address_fields     = has_filter( 'woocommerce_default_address_fields' );
+		$uses_get_country_locale         = has_filter( 'woocommerce_get_country_locale' );
+		$uses_get_country_locale_default = has_filter( 'woocommerce_get_country_locale_default' );
+
+		$uses_locale_filters = $uses_default_address_fields || $uses_get_country_locale || $uses_get_country_locale_default;
+
+		if ( $uses_field_filters && ! $uses_locale_filters ) {
+			echo '<div class="notice notice-warning">';
+			echo '<p><strong>WooCommerce Notice for Developers:</strong> One or more active plugins are customizing checkout, billing, or shipping fields without using required locale filters. This may result in incomplete or misconfigured address fields for some countries. Please ensure the plugins also hook into <code>woocommerce_default_address_fields</code>, <code>woocommerce_get_country_locale</code>, or <code>woocommerce_get_country_locale_default</code>.</p>';
```
opr: Could we ensure this string is translatable? 🙏🏼 Some examples:

- https://github.com/woocommerce/woocommerce/blob/8098c35588b71fd3a8b6a82e7690b96fc0d85a82/plugins/woo-ai/includes/views/html-admin-missing-jetpack.php#L1
- https://github.com/woocommerce/woocommerce/blob/cdd7b596bfcc8a07da9c96d34c15ede8fde64779/plugins/woocommerce/src/Blocks/Domain/Bootstrap.php#L187

I don't think we need the `WooCommerce Notice for Developers:` part.

I also considered if we should allow this notice to be dismissed and hidden, but I think this is important enough to keep nagging.

plugins/woocommerce/includes/admin/class-wc-admin.php
```diff
@@ -289,6 +291,57 @@ public function preview_email_editor_dummy_content() {
...
+
+		$uses_default_address_fields     = has_filter( 'woocommerce_default_address_fields' );
+		$uses_get_country_locale         = has_filter( 'woocommerce_get_country_locale' );
+		$uses_get_country_locale_default = has_filter( 'woocommerce_get_country_locale_default' );
+
+		$uses_locale_filters = $uses_default_address_fields || $uses_get_country_locale || $uses_get_country_locale_default;
+
+		if ( $uses_field_filters && ! $uses_locale_filters && function_exists( 'wc_get_logger' ) ) {
+			wc_get_logger()->warning(
+				'Detected use of checkout/billing/shipping field filters without corresponding locale filters. This may result in incomplete or misconfigured address fields at checkout. Encourage plugin developers to also use woocommerce_default_address_fields or woocommerce_get_country_locale / woocommerce_get_country_locale_default filters.',
```
opr, line 313: I suggested that the other message is updated to be translatable, but noting that this one is fine to keep in English since it only goes into the logs.

opr (changes requested):
Hi @BurakParsAydin thanks for this, I think the check needs to happen on the front-end too, for example if a plugin only adds a `woocommerce_checkout_fields` when `! is_admin()` then this won't trigger.

I think what you've done so far is great, and would catch plugins that add those filters when in admin, but there is an opportunity to also at least log in the wc_logger when plugins are doing this on the front end.

I think it would be great if we could also include a link to https://developer.woocommerce.com/2025/05/13/developer-advisory-update-locale-with-checkout-field-changes/ in the message which gives more information,

BurakParsAydin:
@opr Thank you for the review. I’ve implemented the requested changes as suggested and will wait for your feedback.

opr (commented):
@BurakParsAydin thanks for working on this, I realise we use the `woocommerce_get_country_locale_default` filter internally so I think this check won't work.

I also don't think that checking **only** `woocommerce_get_country_locale` is good either because a plugin may not use that (it might just use the default) so showing a warning for this case would be a false positive. 

I think it may be best to close this PR and review the issue (https://github.com/woocommerce/woocommerce/issues/57811) to discuss solutions.

Thanks a lot for your efforts here so far though, I really appreciate it 🙏🏼

BurakParsAydin:
@opr Thanks a lot for the clarification and feedback. I'll go ahead and close this PR for now and revisit the issue to help figure out a better approach. Happy to continue the discussion there and help find a solution that works well.
Appreciate your time and review!

PR #57888, opened 2025-05-09
[Blueprint] Only show WordPress.org themes

### LINKED ISSUE #57845: [Blueprint] Only show WordPress.org themes
> I am asked to select which plugins to include in the exported JSON file, but then only [WordPress.org](http://WordPress.org) plugins are included in the exported JSON file. Perhaps remove the rest of the plugins from the area where the user selects which plugins to export?
>
> pfwcF5-1Cm-p2#comment-1484

Similarly, we should only show WordPress.org themes in the blueprint export section

BurakParsAydin:

Closes #57845  .

### Before
Screenshot from 2025-05-12 22-49-02

### After
Screenshot from 2025-05-13 14-32-32

1. Install a few themes that are hosted on WordPress.org (e.g. "Twenty Twenty-Four"). 
2. Next, install a few themes that are not hosted on WordPress.org. You can use my testing themes for this purpose. Feel free to download them:
[MockPress.zip](https://github.com/user-attachments/files/20187304/MockPress.zip)
[CodeLite.zip](https://github.com/user-attachments/files/20187270/CodeLite.zip)
[PlainStarter.zip](https://github.com/user-attachments/files/20187271/PlainStarter.zip)
3. Go WooCommerce > Settings > Advanced . Check "Enable blueprint to import and export settings in bulk" option and save the changes. 
4. Now navigate to WooCommerce > Settings > Advanced > Blueprint (Beta) 
5. At this point, you should only see the themes that are hosted on WordPress.org

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


chihsuan (commented):
Appreciate your contribution! @BurakParsAydin 

Let me clarify: We need to exclude themes that are not hosted on wordpress.org from appearing in the Blueprint export section (`/wp-admin/admin.php?page=wc-settings&tab=advanced&section=blueprint`).


Screenshot 2025-05-12 at 11 25 44


That means we should to modify the `get_themes_for_export_group` method instead of `ExportInstallThemeSteps` class. The `ExportInstallThemeSteps` already has the logic to exclude non-WordPress.org themes by checking if `download_link` is available.

https://github.com/woocommerce/woocommerce/blob/33d3a7be645c9ef1e564e3b05ad042610b195523/plugins/woocommerce/src/Admin/Features/Blueprint/Init.php#L152-L175
Also, using the `ThemeURI` field to check if the theme is hosted on wordpress.org is not sufficient because some themes are hosted on wordpress.org but do not have a `ThemeURI` field. We should use the `download_link` field instead or other reliable source to check if the theme is hosted on wordpress.org.

For reference, here is the PR that addresses the plugins issue: https://github.com/woocommerce/woocommerce/pull/57807

Let me know if you have any questions. 🙂

BurakParsAydin:
@chihsuan Thanks for your help! 🙂 I've updated the logic to ensure we only return themes that are hosted on WordPress.org. I’ve also adjusted my initial comment above to reflect the latest changes.

Regarding the PR: I needed to loop through each installed theme and call the wp_themes_api() function individually using the theme_information action. As far as I know, unlike the plugins API, the themes API doesn't support querying multiple slugs at once - it only accepts a single 'slug' per request. Passing multiple slugs results in a “themes API invalid” error.

Let me know if you'd like this result to be cached for performance, or if any further adjustments are needed!

Also, do you think I should add a transient and write a unit test for this, similar to what was done in the plugin PR you previously shared?

chihsuan:
Thanks for the update! @BurakParsAydin 

> As far as I know, unlike the plugins API, the themes API doesn't support querying multiple slugs at once - it only accepts a single 'slug' per request. Passing multiple slugs results in a “themes API invalid” error.

I think it should be possible to query multiple slugs at once.  🤔 Here is the API URL: `https://api.wordpress.org/themes/info/1.2/?action=theme_information&request[slugs][]=twentytwentyone&request[slugs][]=twentytwentytwo`. As you can see, it returns the information for the two themes. Could you please check if this is the case?

> Let me know if you'd like this result to be cached for performance, or if any further adjustments are needed!

Yeah, I think we should cache the result for performance. You can test a site that has multiple themes installed and go to the blueprint page. We don't want to affect the page load time too much for the users.

BurakParsAydin:
@chihsuan _Thanks for getting back to me! 
First off, I’m still getting the hang of the codebase, so a few of my earlier contributions might’ve missed the mark. Really appreciate your understanding and patience while I’m ramping up._


I've been trying to find an effective way to check whether a theme is hosted on WordPress.org without sending multiple requests. However, it seems that plugins_api() and themes_api() behave a bit differently.

The plugins_api() function returns information for all valid slugs and doesn’t throw an error if one of the plugins isn't hosted on WordPress.org—it simply skips the invalid ones and returns the rest. In contrast, themes_api() fails and returns an error if any of the provided slugs aren't hosted on WordPress.org, even if some of them are valid.

You can see what I mean in the URLs below:
For theme: https://api.wordpress.org/themes/info/1.2/?action=theme_information&request[slugs][]=twentytwentyone&request[slugs][]=twentytwentytwo[slugs][]=test-theme
For plugin: https://api.wordpress.org/plugins/info/1.2/?action=plugin_information&request[slugs][]=woocommerce&request[slugs][]=testing

Please let me know if I'm missing something or if there's a better approach.

chihsuan:
> The plugins_api() function returns information for all valid slugs and doesn’t throw an error if one of the plugins isn't hosted on WordPress.org—it simply skips the invalid ones and returns the rest. In contrast, themes_api() fails and returns an error if any of the provided slugs aren't hosted on WordPress.org, even if some of them are valid.


Thanks for the clarification. @BurakParsAydin Indeed, the theme seems to behave differently, making this a challenging issue. 🤔  I think we will need to investigate alternative approaches. Sending multiple requests to verify a theme's presence on WordPress.org is impractical, as it would result in a large number of requests, leading to performance issues.

BurakParsAydin:
@chihsuan I've been trying to find a solution for this, but I feel a bit stuck at this point.

@mikejolley If you have any insights on this issue, I'd really appreciate your input. You can check my latest comment above for more context, but to summarize: is there a way to use `themes_api()` to check whether a theme is hosted on WordPress.org in a single request, similar to how `plugins_api()` handles it?

The challenge I'm facing is that, unlike `plugins_api()`, the `themes_api()` function returns an error if any of the requested themes are not hosted on WordPress.org—even if the others are valid. This makes it tricky to handle multiple slugs in a single request, since a single invalid slug breaks the entire response.

chihsuan:
> The challenge I'm facing is that, unlike `plugins_api()`, the `themes_api()` function returns an error if any of the requested themes are not hosted on WordPress.org—even if the others are valid. This makes it challenging to handle multiple slugs in a single request, as a single invalid slug can break the entire response.

@BurakParsAydin I'm revisiting this PR and feeling uncertain about the optimal approach. I have an idea that we could try to filter out non-WordPress.org themes in the frontend code to avoid blocking the initial page load. However, I'm worried this might cause UI delays - we'd probably need to test this approach to assess its impact.

If you're keen to give it a try, that would be great. Otherwise, I think we could close this PR for now until we have a better solution. Thanks for your time and effort!

BurakParsAydin:
@chihsuan Thanks for the reply! I'm eager to work on this. Before I dive in, could you give me a hint about this?:

> we could try to filter out non-WordPress.org themes in the frontend code

chihsuan:
@BurakParsAydin I think we could try filtering the list of themes in JavaScript

The theme list is currently being fetched from `window.wcSettings?.admin?.blueprint_step_groups`. We could call the API asynchronously/or in parallel to check if the theme is available on `WordPress.org` by using useState + useEffect hooks.`

https://github.com/woocommerce/woocommerce/blob/3003db4b2c7d2c2a035bae055eeeecbd6bafa438/plugins/woocommerce/client/admin/client/blueprint/settings/slotfill.js#L40-L52

BurakParsAydin:
@chihsuan I revisited the whole flow to ensure everything works as expected.

Initially, it seemed like `wp_themes_api()` wasn't returning the data we needed, but after digging deeper, I realized the issue was with how the theme slugs were being handled in the filtering logic.

That’s now corrected — and with the revised implementation, server-side filtering works cleanly. We’re correctly including only WordPress.org themes based on the presence of a `download_link` in the API response.

So no need for additional filtering on the frontend. I'll be waiting for your feedback. 

**_Edit_**: CodeRabbit suggested four different points, but some of them seem inconsistent. Should I just apply the one about the transient?

chihsuan:
> CodeRabbit suggested four different points, but some of them seem inconsistent. Should I just apply the one about the transient?

Yes, we should ensure that this checking doesn't affect the page load time too much, so caching is a good idea.

BurakParsAydin:
@chihsuan FYI, I've applied the transient. Waiting for your feedback.

plugins/woocommerce/src/Admin/Features/Blueprint/Init.php
```diff
@@ -313,4 +325,52 @@ function ( $plugin ) use ( $api_response ) {
...
+		foreach ( $all_themes as $key => $theme ) {
+			$theme_slugs[] = strtolower( $key );
+		}
+
+		$api_response = $this->wp_themes_api(
+			'theme_information',
+			array(
+				'fields' => array(
+					'downloadlink' => true,
+				),
```
chihsuan, line 370: It seems that `$api_response` not only contains the download link but also other details about the theme. Could you please check if we could exclude other fields from the API response to reduce the response size?

Screenshot 2025-07-04 at 16 33 14

chihsuan (commented):
Thanks @BurakParsAydin, nice work! 

I confirmed that non-wordpress.org themes are not showing up in the list. 👍 

Just a minor comment on the response fields, other than that, everything looks good.

BurakParsAydin:
@chihsuan I trimmed the unused data. Some additional data is still included, but I don’t think there’s a way to disable it. The overall payload is small, though.
Thank you :slightly_smiling_face:

chihsuan:
Hey @BurakParsAydin, it seems that PHP unit tests are failing for the new changes. Could you take a look?

```
Time: 05:32.088, Memory: 813.50 MB

There was 1 error:

1) Automattic\WooCommerce\Tests\Admin\Features\Blueprint\InitTest::test_get_themes_for_export_group
TypeError: strtolower(): Argument #1 ($string) must be of type string, int given

/var/www/html/wp-content/plugins/woocommerce/src/Admin/Features/Blueprint/Init.php:347
/var/www/html/wp-content/plugins/woocommerce/src/Admin/Features/Blueprint/Init.php:165
/var/www/html/wp-content/plugins/woocommerce/tests/php/src/Admin/Features/Blueprint/InitTest.php:90
```

BurakParsAydin:
@chihsuan FYI, I added an if-statement check. Feel free to review it please.

chihsuan (approved):
Thanks, LGTM

PR #57889, opened 2025-05-09
Product Image accepts On Sale badge as inner block (backward compatible update)

kmanijak:

This PR is a suplement to https://github.com/woocommerce/woocommerce/pull/57586, which makes sure the changes are backwards compatible. Changes:
- Making sure Product Image keeps rendering Sale badge in All Products 
- Making sure Product Image keeps rendering Sale badge in other cases but assuring migration happens
- Move to native alignment setting for On Sale Badge

| Before | After |
| ------ | ----- |
|        |       |

#### Backwards compatibility

> [!IMPORTANT]  
> Do all following actions on `trunk` or use last stable WooCommerce version. 

**All Products - pristine** 
1. Create new post and insert All Products block (paste [this markup](https://gist.github.com/gigitux/5cb489fa10a960354e0f1baf797e1c7f) in code editor)
2. Keep the block pristine and save
3. Keep a tab with editor and frontend opened

**All Products - edited** 
1. Create new post and insert All Products block (paste [this markup](https://gist.github.com/gigitux/5cb489fa10a960354e0f1baf797e1c7f) in code editor)
2. Enter edit mode of All Products
3. Focus on Product Image and explicitly enable "Show sale badge"
4. Click DONE and save 
5. Keep a tab with editor and frontend opened

**Product Collection** 
1. Create new post and insert
  - On Sale Products and keep it pristine
  - On Sale Products and focus on Product Image, disable and enable again "Show sale badge" setting
  - On Sale Products and focus on Product Image, disable "Show sale badge" setting
2. Save
3. Keep a tab with editor and frontend opened

Now, switch to this branch or this version if you test a ZIP

**All Products - pristine** 
1. Refresh editor and frontend tabs
2. **Expected:** make sure "Sale" badge is rendered for on sale products

**All Products - edited** 
1. Refresh editor and frontend tabs
2. **Expected:** make sure "Sale" badge is rendered for on sale products

**Product Collection - backward compatible frontend pre-migration** 
1. Refresh FRONTEND tab first:
2. **Expected:** make sure "Sale" badge is rendered for on sale products in first two cases and not the last one (where it was badge was explicitly disabled)

**Product Collection - migration** 
1. Refresh EDITOR tab
2. **Expected:** make sure "Sale" badge is rendered for on sale products in first two cases with a correct alignment. It's not rendered in the last one (where it was badge was explicitly disabled)
3. Focus on Product Image in first two cases and verify On Sale Badge is a separate inner block in Product Image

image

4. Save

**Product Collection - backward compatible frontend post-migration** 
1. Refresh FRONTEND tab first:
2. **Expected:** make sure "Sale" badge is rendered for on sale products in first two cases and not the last one (where it was badge was explicitly disabled)

#### Adding Product Collection now

1. Create new post
2. Add On Sale Products
3. Focus on Product Image and verify On Sale Badge is a separate inner block in Product Image

#### Default templates

1. Go to Product Catalog
2. Remove everything except for Header and Footer
3. Insert Product Grid (Classic) block
4. Focus on it and click "Tranform Into Blocks"
5. Focus on Product Image and verify On Sale Badge is a separate inner block in Product Image
6. Reset template to original state
7. Focus on Product Image and verify On Sale Badge is a separate inner block in Product Image

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.tsx
```diff
@@ -101,6 +102,28 @@ const Image = ( {
 type Props = BlockAttributes &
 	HTMLAttributes< HTMLDivElement > & { style?: Record< string, unknown > };
 
+// props.product is not listed in the BlockAttributes explicitly,
+// but it is implicitly passed from the All Products block.
+// This is what distinguishes this block from the other usage of the Product Image component.
+const displayLegacySaleBadge = ( props: Props ) => {
```
sunyatasattva: In this case I think we should still type `product`, maybe in a type called: `LegacyProps`. The props object should be aware of anything that _could_ be passed to it.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.tsx
```diff
@@ -101,6 +102,28 @@ const Image = ( {
...
+	if ( isInAllProducts ) {
+		// If the All Products block is pristine, it doesn't have a showSaleBadge attribute
+		// but it was `true` by default.
+		const trueByDefault = props.showSaleBadge === undefined;
+
+		// If the All Products block was edited, it will have a showSaleBadge attribute
+		// that we should respect.
+		const trueAfterEdit = props.showSaleBadge === true;
+
+		return trueByDefault || trueAfterEdit;
```
sunyatasattva: Nitpick and probably a matter of style preference, but you could return earlier if it was `trueByDefault`. As a reader, I'm just _ever-so-slightly_ confused by why we check if it's `trueAfterEdit` and I needed that extra second to understand it was just for style preference.

Not a big deal, but think about it.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/deprecated.tsx
```diff
@@ -1,66 +1,51 @@
...
+	isEligible: ( { showSaleBadge }: BlockAttributesV1 ) => {
+		// If the block is pristine, it doesn't have a showSaleBadge attribute
+		// but it is `true` by default.
+		const trueByDefault = showSaleBadge === undefined;
+
+		// If the block was edited, it will have a showSaleBadge attribute
+		// that we should respect.
+		const trueAfterEdit = showSaleBadge === true;
+
+		return trueByDefault || trueAfterEdit;
```
sunyatasattva: Another nitpick, but this code is repeated and I know it's only twice, but maybe at this point making it into a small function by itself could improve readability.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/deprecated.tsx
```diff
@@ -1,66 +1,51 @@
...
-		if ( ! showSaleBadge ) {
-			return [ rest ];
-		}
+		const { showSaleBadge, saleBadgeAlign } = attributes;
 
-		let justifyContent = 'flex-end';
-		if ( saleBadgeAlign === 'left' ) {
-			justifyContent = 'flex-start';
-		} else if ( saleBadgeAlign === 'center' ) {
-			justifyContent = 'center';
```
sunyatasattva, line 48: Mmm, I guess this is actually a problem I hadn't thought about. We are not keeping backwards compatibility with the `saleBadgeAlign` on migration since we migrated to block align. It's a small problem, the users can fix it immediately, but still… not the ideal experience.

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/constants.ts
```diff
@@ -127,6 +127,14 @@ export const INNER_BLOCKS_PRODUCT_TEMPLATE: InnerBlockTemplate = [
...
 				imageSizing: ImageSizing.THUMBNAIL,
 			},
+			[
+				[
+					'woocommerce/product-sale-badge',
+					{
+						align: 'right',
+					},
+				],
+			],
```
sunyatasattva, line 137: Good catch!

plugins/woocommerce/src/Blocks/BlockTypes/ProductImage.php
```diff
@@ -72,18 +72,15 @@ private function render_on_sale_badge( $product, $attributes ) {
 			return '';
 		}
 
-		$font_size = StyleAttributesUtils::get_font_size_class_and_style( $attributes );
+		$classes_and_styles = StyleAttributesUtils::get_classes_and_styles_by_attributes( $attributes );
```
sunyatasattva: Can you tell me more about this change?

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.tsx
kmanijak: Interesting. I prefer this style due to my thought evolution. I'm starting with:

```
return props.showSaleBadge === undefined || props.showSaleBadge === true;
```

but it's hard to understand so I extract constants to give them meaningful name and eventually read it as a sentence:

```
// This function returns true if it's true by default OR true after edits 
return trueByDefault || trueAfterEdit;
```

I'm more keen to go back to this version (so the second condition won't be fired):

```
return props.showSaleBadge === undefined || props.showSaleBadge === true;
```

rather than:

```
const trueByDefault = props.showSaleBadge === undefined;

if ( trueByDefault ) {
    return true;
}

const trueAfterEdit = props.showSaleBadge === true;

return trueAfterEdit;
```

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/deprecated.tsx
kmanijak: Ah right! I added it much later an missed the duplication! 👍

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.tsx
kmanijak: Actually I refactored it to

```
	// If the block is pristine, it doesn't have a showSaleBadge attribute
	// but it is supposed to be `true` by default.
	if ( showSaleBadge === undefined ) {
		return true;
	}

	// If the block was edited, it will have a showSaleBadge attribute
	// that we should respect.
	return showSaleBadge;
```

I guess it's a nice middle ground 😄

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/deprecated.tsx
kmanijak: We do, [I pass it here](https://github.com/woocommerce/woocommerce/pull/57889/files#diff-69e1c96271f9381541b4f08cab28247de165de7ef03c8d7cfc7064910b5b1a92R45-R48) but to the On Sale Badge directly (`align` -> `align`).

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.tsx
kmanijak: Done https://github.com/woocommerce/woocommerce/pull/57889/commits/f151536b1e085fa6ab1dd9fccee86c7afcf8f21d

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.tsx
kmanijak: Done https://github.com/woocommerce/woocommerce/pull/57889/commits/d172eef438ac47d24d7a6b5a15e736b340241f52

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/deprecated.tsx
kmanijak: Done https://github.com/woocommerce/woocommerce/pull/57889/commits/d172eef438ac47d24d7a6b5a15e736b340241f52

plugins/woocommerce/src/Blocks/BlockTypes/ProductImage.php
kmanijak: Ah, I had a draft with explanation what I did here. And then I realised what I did was a mistake 😅 

Long story short I wanted to utilise a "supports" alignment but it's obviously not yet there pre-migration. So I brought back the previous mechanism of aligning badge including bringing back CSS I thought would be redundant: https://github.com/woocommerce/woocommerce/pull/57889/commits/88c12271de3997f9d4eb642f865b8e30a9d8c4b6.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.tsx
sunyatasattva: Mmm… interesting way to type it, I find it a bit noisy, why did you intersect two different objects (one for `style` and one for `product`) instead of one object for both props?

Also, since you are using `Props` to type the Block below, `product` is expected to be defined in the normal block, which is not always the case. Either `product` needs to be optional, or I would separate the types into `Props` and `LegacyProps` for even further clarity, so future consumers don't think it's ok to send `product` to the badge even when “technically” accepted.

sunyatasattva (changes requested):

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/block.tsx
kmanijak: > Mmm… interesting way to type it, I find it a bit noisy, why did you intersect two different objects (one for style and one for product) instead of one object for both props?

I have nothing to defend myself, honestly, I don't know, I think I rushed 😅 

> I would separate the types into Props and LegacyProps for even further clarity, so future consumers don't think it's ok to send product to the badge even when “technically” accepted.

Done in https://github.com/woocommerce/woocommerce/pull/57889/commits/41e988bc05b51d1c8e6e29bcf002fe34b66aeb74. That's what you meant, right? 🙌

sunyatasattva (approved):
Tested, reviewed and approved. Let's merge and move on to the [other PR](https://github.com/woocommerce/woocommerce/pull/57586).

PR #57891, opened 2025-05-09
Revert "Product Collection - Cross-sell product: fix Product Selector when the selected product is deleted"

gigitux:
Reverts woocommerce/woocommerce#57717. See this comment https://github.com/woocommerce/woocommerce/pull/57717#pullrequestreview-2829440987

With Code Freeze approaching, it’s better to fix this carefully rather than rushing. For this reason, I'm reverting the PR.


kmanijak (approved):
Hmm.. editor selection highlight is back for the Product Picker but not around the whole Product Collection. I think there may be something else going on then. But let's revert and fix with no rush (it's not a big issue anyway).

Approving and merging 👍 Thanks for super quick revert!

PR #57892, opened 2025-05-09
Add alphabetical fallback sorting for Product Collection Block with same menu order

### LINKED ISSUE #53139: [Enhancement]: Product Collection Block - use alphabetical ordering as fallback for manual order

The product collection block allows setting the same menu order for more than one product, if set from within the product editor.

Currently, if the block is set to manual order, products with the same menu order set are being listed based on which product had that menu order set first. It would make more sense for the block to fall back to alphabetical ordering where two or more products have the same menu order set.

_No response_

### Additional context

_No response_

amitraj2203:

- This PR enhances the `Product Collection` Block by adding alphabetical (title-based) sorting as a fallback when multiple products have the same menu order.
- Currently, when products share the same menu order value and `Manual (menu order)` in `Order By` is selected, they appear based on which product had that menu order set first.
- This change makes the sorting more predictable and user-friendly by using product titles as a secondary sort parameter.

Closes #53139 

1. Go to the _Admin Dashboard > Products > Add new product_
2. Create at least 3 products with the same menu order value (e.g., set them all to 7)
3. Or, if products already exist, edit their menu order by editing the product, clicking on `Advanced` tab and changing the `Menu order`
4. Make sure their titles are in different alphabetical order than their creation order
5. Now create a page/post and add a `Product Collection` block
6. Set the block to use `Manual (menu order)` from Inspector control
7. Visit the page/post in frontend
8. Verify that products with the same menu order value are now sorted alphabetically by title

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


KokkieH:
@amitraj2203 Thanks for the PR. I've requested review from the team responsible for maintaining this area of the plugin.

plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/QueryBuilder.php
```diff
@@ -1207,4 +1208,33 @@ function ( $val ) {
...
+		if ( ! $is_product_collection_block ) {
+			return $clauses;
+		}
+
+		$orderby = $query_vars['orderby'] ?? null;
+		if ( 'menu_order' !== $orderby ) {
+			return $clauses;
+		}
+
+		$is_ascending_order = 'asc' === strtolower( $query_vars['order'] ) ?? false;
```
KokkieH: `??` has lower precedence than `===` so this expression is evaluated as  `$is_ascending_order = ('asc' === strtolower( $query_vars['order'] )) ?? false;` This will result in descending ordering if `order` is not set, which is unexpected.

`order` is set to `asc` by default on line 146, but it might be a good idea to be a bit more defensive here.

If we want to default to ascending order, we'd need this instead:

`$is_ascending_order = ! isset( $query_vars['order'] ) || 'asc' === strtolower( $query_vars['order'] );`

KokkieH (approved):
Tests well for me. Also checked that other ordering, and setting different or no manual order values still work as expected.

I left a comment on how we could make the code a bit more defensive, but not a blocker imo.

plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/QueryBuilder.php
amitraj2203: Thanks for the review 🙇. I've updated the code to reflect this approach now.

KokkieH: Thanks for making the update!

I would prefer if someone from the FSE team also review this. Hopefully they'll get to this soon :)

gigitux:
This PR looks good to me! It seems the behavior now matches what's implemented in Storefront ([reference](https://rudrastyh.com/woocommerce/change-defaut-product-sorting.html)).

Should we consider updating the label as well? Storefront uses: `Default sorting (custom ordering + name)`. Maybe something like `Manual (menu order + name)` would be clearer?

Also, there's a failing unit test — could you take a look? Furthermore, it would be great to add an integration test for both the editor and frontend sides too.

I think that it makes sense to include @kmanijak in this discussion.

kmanijak:
Nice improvement @amitraj2203! 🚀

I tested this in two cases:
- Product Catalog template (using default query -> `inherit === true`)
- in post (using custom query -> `inherit === false`)
and it works fine in both editor and frontend.

> Should we consider updating the label as well? Storefront uses: Default sorting (custom ordering + name). Maybe something like Manual (menu order + name) would be clearer?

It's not a bad idea! 🤔 I think we should do it!

Otherwise, let's just address @gigitux's suggestion:

> Also, there's a failing unit test — could you take a look? Furthermore, it would be great to add an integration test for both the editor and frontend sides too.

amitraj2203:
Hi @gigitux @kmanijak,

I'd appreciate your input on the testing approach for the `menu_order` sorting with title fallback feature. I'm thinking of two different testing styles and would like guidance on which is preferred:

### **Approach 1: Integration tests with real products**
- Creates actual products in the database with different `menu_order` values and names
- Executes full `WP_Query` to verify the sorting order
- Tests both frontend and editor contexts with ascending/descending order
- Pros: Tests the complete flow end-to-end
- Cons: Slow execution ~(3000+ ms per test due to database operations)~

### **Approach 2: Unit tests focusing on query building**
- Tests the query parameter construction without creating products
- Verifies that the correct SQL clauses are generated
- Tests the orderby values
- Pros: Fast execution, focused on the actual code changes
- Cons: Doesn't verify the actual sorting behavior with real data

Currently, I have both types of tests implemented so need your guidance which one to use.

---

Also regarding this:
> Should we consider updating the label as well? Storefront uses: Default sorting (custom ordering + name). Maybe something like Manual (menu order + name) would be clearer?

I hope changing the labels here does the job. Let me know if I'm missing anything.

https://github.com/woocommerce/woocommerce/blob/011ab429000cc66bccb71ea0b02e25b9469b2ca0/plugins/woocommerce/includes/customizer/class-wc-shop-customizer.php#L328

https://github.com/woocommerce/woocommerce/blob/011ab429000cc66bccb71ea0b02e25b9469b2ca0/plugins/woocommerce/includes/customizer/class-wc-shop-customizer.php#L499

Thanks!

gigitux:
I’d prefer integration tests, as they help ensure the Product Collection behaves as expected end-to-end. In the future, if we update the code to handle additional queries, it could impact the outcome.

That said, if you’ve already written unit tests specifically for the SQL query, I’m happy to include those as well.

Also, @kmanijak has more experience than me on this part of this codebase, so could have a different opinion. 


> I hope changing the labels here does the job. Let me know if I'm missing anything.

I would keep the `Manual` instead of `Default`, so: `Manual (menu order + name)` would be great!


Thanks for your contribution 🙇

amitraj2203:
@kmanijak @gigitux I've added both the integration tests as well as tests for the SQL query. As this was my first time writing test cases for this codebase, please let me know if they're alright or if something needs to be changed. Thanks for all your support 🙇

plugins/woocommerce/includes/customizer/class-wc-shop-customizer.php
```diff
@@ -496,7 +496,7 @@ public function add_product_catalog_section( $wp_customize ) {
 				'choices'     => apply_filters(
 					'woocommerce_default_catalog_orderby_options',
 					array(
-						'menu_order' => __( 'Default sorting (custom ordering + name)', 'woocommerce' ),
```
gigitux: Sorry for not being clear. My proposal wasn't to update this part of the codebase, but update this part related to the Product Collection block:

https://github.com/woocommerce/woocommerce/blob/aed08a89ef97179b3cdf42717085d18e79859960/plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/edit/inspector-controls/order-by-control/default-query-order-by-control.tsx#L43-L44

https://github.com/woocommerce/woocommerce/blob/aed08a89ef97179b3cdf42717085d18e79859960/plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/edit/inspector-controls/order-by-control/custom-query-order-by-control.tsx#L63-L64

gigitux:
Awesome work! Thanks for adding the unit test! I left just one comment and after that, I think that we can merge this PR.

I have two thoughts, but I don't think we should address in this PR.
1. The `plugins/woocommerce/tests/php/src/Blocks/BlockTypes/ProductCollection.php` file is pretty big. I think that we should separate unit tests regarding the SQL generation query and the integration test regarding the block. 
2. It would be great to start to add some integration tests for the Product Collection for the editor side as well.

gigitux (changes requested):

plugins/woocommerce/includes/customizer/class-wc-shop-customizer.php
amitraj2203: Oh my bad, sorry for the confusion. I've fixed it now :)

gigitux (approved):
LGTM! I'm going to approve this PR, but before to merge it, I will wait for another feedback from @kmanijak.

In any case, awesome work! Thanks for your work! 🙇

amitraj2203:
> I have two thoughts, but I don't think we should address in this PR.
> 
> 1. The `plugins/woocommerce/tests/php/src/Blocks/BlockTypes/ProductCollection.php` file is pretty big. I think that we should separate unit tests regarding the SQL generation query and the integration test regarding the block.
> 2. It would be great to start to add some integration tests for the Product Collection for the editor side as well.

@gigitux Yes, definitely! I'd be happy to work on these as well. It would be great if separate issues could be created for them, so we can track and prioritize the work accordingly. Let me know if you want me to create issues for these.

gigitux:
> > I have two thoughts, but I don't think we should address in this PR.
> > 
> > 1. The `plugins/woocommerce/tests/php/src/Blocks/BlockTypes/ProductCollection.php` file is pretty big. I think that we should separate unit tests regarding the SQL generation query and the integration test regarding the block.
> > 2. It would be great to start to add some integration tests for the Product Collection for the editor side as well.
> 
> @gigitux Yes, definitely! I'd be happy to work on these as well. It would be great if separate issues could be created for them, so we can track and prioritize the work accordingly. Let me know if you want me to create issues for these.

Feel free to create them. Moreover, feel free to ping me in case you need some help/you have questions!

kmanijak (approved):
Thanks for ping @gigitux, I wasn't re-testing the PR but checked the tests and they look fine for me, hence approving as well. And +1 for splitting tests, thank you @amitraj2203 for handling all this!

PR #57893, opened 2025-05-10
[Experimental] Extend Product Details block using Block Hooks API

dinhtungdu:

This PR introduces a mechanism to extend the Product Details block with additional accordions using the Block Hooks API. It allows developers to add custom accordion items to the Product Details block using the `woocommerce_product_details_hooked_items` filter. By providing a title and editor-generated block markup, developers can easily register a new product details item, which is visible not only on the front end but also in the editor.

Some design decisions:
- Using custom metadata to distinguish between Accordion blocks so that we can inject the hooked blocks into the Product Details Accordion only. The metadata is hardcoded in the Product Details inner blocks template. Themes need to ensure their Single Product template has metadata for hooked blocks to work.
   - This may sound fragile, as the user can remove the whole Accordion block in the Site Editor and then add it back, the metadata will be lost. But in that case, even if we don't use metadata, Block Hooks won't work with blocks added after the hooked blocks are registered. In other words, Block Hooks will only add blocks to the anchor blocks defined in the template at the time hooked blocks are registered, and it's a design choice. Because of that, using a custom/hardcoded metadata here is not fragile.
   - Whatever the metadata name we use here, we need to stick with it forever. I'm using `isProductDetailInnerBlock`, but open to change for more consistent, feedback is welcome :).
- Only supporting `last_child` position. It means that developers can't programmatically add an item after `Description` and before `Additional Information`, for example. It doesn't mean we can't support that, but we want to keep the hook simple and encourage changing the order of hooked blocks using the editor. The order of hooked blocks can be manipulated using filter priority, though. I'm open to changing this and supporting more positions, up for debate.
- I looked at modifying the template to inject the custom metadata like WordPress does with Block Hooks API itself, but we can not modify it **before** Block Hooks does its job, so it's not an option.

Closes [WOOPLUG-4235](https://linear.app/a8c/issue/WOOPLUG-4235/add-support-for-extending-product-details-block-with-custom-accordions)

1. Add the `Blockified Product Details` block to the `Single Product` template.
1. Create a simple plugin with the following code:
```php
<?php
/**
 * Plugin Name: WooCommerce Product Details Extension Demo
 * Description: Demonstrates how to extend the WooCommerce Product Details block via Block Hooks API.
 * Version: 1.0.0
 */
add_filter( 'woocommerce_product_details_hooked_blocks', function( $items ) {
    $items[] = [
        'title' => 'Custom Information',
        'content' => '<!-- wp:paragraph --><p>This is custom information added to the product details block using Block Hooks API!</p><!-- /wp:paragraph -->',
    ];
    $items[] = [
        'title' => 'Custom Information',
        'content' => '<!-- wp:paragraph --><p>This is the duplicated item with the same title. This will replace the former one.</p><!-- /wp:paragraph -->',
    ];
    $items[] = [
        'title' => 'Invalid Content',
        'content' => '<p>This item will be filtered out and logged as a WC error because the content is not a valid block markup.</p>',
    ];
    return $items;
});
```
2. Activate the plugin.
3. Visit a product page.
4. Verify that a new "Custom Information" accordion appears alongside the default accordions.
5. Verify that the accordion opens and closes properly when clicked.
6. Edit the `Single Product` template in the Site Editor, see the "Custom Information" accordion added with the paragraph block.
7. Verify the invalid item doesn't show up in the editor or on the front end.

| Editor | Frontend |
| ------ | ----- |
|  image  | image  |

### DIFF
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/product-details/edit.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/product-details/edit.tsx
@@ -21,7 +21,7 @@ const TEMPLATE: InnerBlockTemplate[] = [
 		'woocommerce/accordion-group',
 		{
 			metadata: {
-				isProductDetailsInnerBlock: true,
+				isDescendantOfProductDetails: true,
 			},
 		},
 		[
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/product-details/edit.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/product-details/edit.tsx
@@ -19,7 +19,11 @@ import { ProductDetailsEditProps } from './types';
 const TEMPLATE: InnerBlockTemplate[] = [
 	[
 		'woocommerce/accordion-group',
-		{},
+		{
+			metadata: {
+				isProductDetailsInnerBlock: true,
+			},
+		},
 		[
 			[
 				'woocommerce/accordion-item',
--- a/plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
@@ -4,6 +4,7 @@ namespace Automattic\WooCommerce\Blocks\BlockTypes;
 
 use WP_Block;
 use WP_HTML_Tag_Processor;
+use WP_Block_Supports;
 
 /**
  * BlockifiedProductDetails class.
@@ -23,6 +24,93 @@ class BlockifiedProductDetails extends AbstractBlock {
 	 */
 	protected function get_block_type_style() {
 		return null;
+	}
+
+	/**
+	 * Initialize the block type.
+	 *
+	 * @return void
+	 */
+	protected function initialize() {
+		parent::initialize();
+
+		/**
+		 * Filter the items that are hooked into the Product Details block.
+		 *
+		 * @hook woocommerce_product_details_hooked_items
+		 *
+		 * @param {array} $hooked_items The items that are hooked into the Product Details block.
+		 * @return {array} The items that are hooked into the Product Details block.
+		 */
+		$hooked_items = apply_filters( 'woocommerce_product_details_hooked_items', [] );
+
+		$validated_hooked_items = array_filter( $hooked_items, function( $item ) {
+			return isset( $item['title'] ) && isset( $item['content'] ) && is_string( $item['title'] ) && is_string( $item['content'] );
+		} );
+
+		foreach ( $validated_hooked_items as $item ) {
+			$this->register_product_details_item( $item['title'], $item['content'] );
+		}
+	}
+
+	/**
+	 * Register a product details item using Block Hooks API.
+	 *
+	 * @param string $title The title of the item.
+	 * @param string $content The content of the item.
+	 * @return void
+	 */
+	protected function register_product_details_item($title, $content) {
+		$slug = sanitize_title( $title );
+
+		add_filter(
+			'hooked_block_types',
+			function ( $hooked_block_types, $relative_position, $anchor_block_type, $context ) use ( $slug ) {
+				if ( 'woocommerce/accordion-group' === $anchor_block_type && 'last_child' === $relative_position ) {
+					$hooked_block_types[] = $slug;
+				}
+				return $hooked_block_types;
+			},
+			10,
+			4
+		);
+
+		add_filter(
+			"hooked_block_{$slug}",
+			function (
+				$parsed_hooked_block,
+				$hooked_block_type,
+				$relative_position,
+				$parsed_anchor_block,
+				$context
+			) use ( $title, $content ) {
+
+				if ( is_null( $parsed_hooked_block ) ) {
+					return $parsed_hooked_block;
+				}
+
+				// TODO: Verify that the anchor Accordion Group block is a child of the Product Details block.
+				if ( 'woocommerce/accordion-group' !== $parsed_anchor_block['blockName'] || $relative_position !== 'last_child' ) {
+					return $parsed_hooked_block;
+				}
+
+				$accordion_item_template = '<!-- wp:woocommerce/accordion-item -->
+				<div class="wp-block-woocommerce-accordion-item"><!-- wp:woocommerce/accordion-header -->
+				<h3 class="wp-block-woocommerce-accordion-header accordion-item__heading"><button class="accordion-item__toggle"><span>%1$s</span><span class="accordion-item__toggle-icon has-icon-plus" style="width:1.2em;height:1.2em"><svg width="1.2em" height="1.2em" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg" aria-hidden="true"><path d="M11 12.5V17.5H12.5V12.5H17.5V11H12.5V6H11V11H6V12.5H11Z" fill="currentColor"></path></svg></span></button></h3>
+				<!-- /wp:woocommerce/accordion-header -->
+
+				<!-- wp:woocommerce/accordion-panel -->
+				<div class="wp-block-woocommerce-accordion-panel"><div class="accordion-content__wrapper">%2$s</div></div>
+				<!-- /wp:woocommerce/accordion-panel --></div>
+				<!-- /wp:woocommerce/accordion-item -->';
+
+				$accordion_item_block = parse_blocks( sprintf( $accordion_item_template, $title, $content ) );
+
+				return $accordion_item_block[0];
+			},
+			10,
+			5
+		);
 	}
 
 	/**
--- a/plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
@@ -293,7 +293,7 @@ class BlockifiedProductDetails extends AbstractBlock {
 	private function has_accordion( $parsed_block ) {
 		if (
 			'woocommerce/accordion-group' === $parsed_block['blockName'] &&
-			! empty( $parsed_block['attrs']['metadata']['isProductDetailsInnerBlock'] )
+			! empty( $parsed_block['attrs']['metadata']['isDescendantOfProductDetails'] )
 		) {
 			return true;
 		}
@@ -348,8 +348,7 @@ class BlockifiedProductDetails extends AbstractBlock {
 				if (
 					'woocommerce/accordion-group' !== $parsed_anchor_block['blockName'] ||
 					'last_child' !== $relative_position ||
-					! isset( $parsed_anchor_block['attrs']['metadata']['isProductDetailsInnerBlock'] ) ||
-					! $parsed_anchor_block['attrs']['metadata']['isProductDetailsInnerBlock']
+					empty( $parsed_anchor_block['attrs']['metadata']['isDescendantOfProductDetails'] )
 				) {
 					return array();
 				}
--- a/plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
@@ -291,7 +291,10 @@ class BlockifiedProductDetails extends AbstractBlock {
 	 * @return bool True if the block has an accordion, false otherwise.
 	 */
 	private function has_accordion( $parsed_block ) {
-		if ( 'woocommerce/accordion-group' === $parsed_block['blockName'] ) {
+		if (
+			'woocommerce/accordion-group' === $parsed_block['blockName'] &&
+			! empty( $parsed_block['attrs']['metadata']['isProductDetailsInnerBlock'] )
+		) {
 			return true;
 		}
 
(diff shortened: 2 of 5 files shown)


plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
```diff
@@ -25,6 +25,103 @@ protected function get_block_type_style() {
...
+	 * @return void
+	 */
+	protected function register_product_details_item( $title, $content ) {
+		$slug = sanitize_title( $title );
+
+		add_filter(
+			'hooked_block_types',
+			function ( $hooked_block_types, $relative_position, $anchor_block_type ) use ( $slug ) {
+				if ( 'woocommerce/accordion-group' === $anchor_block_type && 'last_child' === $relative_position ) {
+					$hooked_block_types[] = $slug;
```
ockham: As mentioned in our Slack convo, this deserves highlighting, as it uses an arbitrary string (rather than an actual block type name). Internally, the "hooked block type" passed here is used first and foremost for `ignoredHookedBlocks`.

It turns out that this "hack" actually works pretty well! The reason is that even if the user changes the title in the editor, `ignoredHookedBlocks` will still contain the original title. For example, if the original title was "Custom Information" (per the example in the PR description), the slug generated from it and used in `ignoredHookedBlocks` will be `custom-information`. 

Now even if the user changes the title in the editor to "New Custom Information", `ignoredHookedBlocks` will continue to contain `custom-information` -- thus preventing the original block to be re-inserted ✅

ockham:
The example code in the PR description needs to be updated to use the new filter name `woocommerce_product_details_hooked_blocks` (instead of `_items`), per https://github.com/woocommerce/woocommerce/pull/57893/commits/950bbe2278c2edaf573d6190dfd4068dae69eacf 😊

ockham:
With the example code from the PR description (and after [updating the block filter name](https://github.com/woocommerce/woocommerce/pull/57893#issuecomment-2876212747)), I'm currently getting a block validation error in the editor:

image

The console has these additional details:

image

The reason is probably that we're missing `<!-- wp:paragraph -->` block delimiters for the `<p>` we're adding. So we should probably instruct 3PDs to provide _block_ markup in the `$content` argument.

dinhtungdu:
@ockham Thanks I updated the example. I indeed missed the block tag for paragraph block in the example. The warning shouldn't show up now.

ockham:
Here's a variation of this PR that uses dedicated `create_block` functions for each of the required blocks: https://github.com/woocommerce/woocommerce/pull/57968

This is based on [the exploration in this experimental repo](https://github.com/ockham/woo-block-hooks-experiments/blob/f66cfd132a1361964ac8099e6c1f15b16b27b70b/woo-block-hooks-experiments.php#L50-L59).

dinhtungdu:
@ockham walking through your PR, I was thinking about getting the current accordion item attribute to use for hooked one (as we can pass attribute now with your PR). But I will need to get the attributes from the parsed block, which is already available. I thought we may reuse that parsed block without the need of recreating it. The default template is still parsed from string for simplicity, but as we now can reuse the current accordion item, I think we don't need to pass additional attribute anymore, which remove the need of `create_block` helper.

I updated the accordion template with the ability to create a template from the last accordion item then use it for new creating accordion item. I updated for the compatibility layer but the same method can be share with Block Hooks introduced in this PR: [`770a380` (#57973)](https://github.com/woocommerce/woocommerce/pull/57973/commits/770a380e3f32e32d4caf76ff7ccb27112fcfc429).

Edit: here is the draft PR for Block Hooks: https://github.com/woocommerce/woocommerce/pull/57975

Let me know what you think.

ockham:
TBH I think of using those `create_block` helpers as more of an implementation detail. I'm fine with parsing block markup -- especially at the beginning; I just think it's somewhat opaque and harder to maintain/extend if (when) the Accordion block changes. The `create_block` helpers are an attempt to make these things more transparent and maintainable, but aren't crucial to solving the actual problem at hand.

With that said, I'm skeptical about copying the preceding Accordion Item block's attributes to use for the hooked block. That introduces an assumption (that the user will want the attributes to be the same for both blocks) that's somewhat arbitrary IMO.

FWIW, the original idea for Block Hooks was that they would insert a block in its "default" state -- as if the user had just inserted the block manually into the editor, with no further customizations. Now of course, this doesn't make much sense for most static blocks -- a paragraph block in its default state is just empty, as is an Accordion Item block, which is why we're allowing the user to provide a `$title` and `$content`. However, for any other attribute, I think it makes sense to stick with the default value. (If we decide that that's not enough, we might consider introducing an extra argument to supply additional attributes.)

dinhtungdu:
> With that said, I'm skeptical about copying the preceding Accordion Item block's attributes to use for the hooked block. That introduces an assumption (that the user will want the attributes to be the same for both blocks) that's somewhat arbitrary IMO.

@ockham As mentioned in Slack, I also think that it doesn't make sense to do that anymore.

> FWIW, the original idea for Block Hooks was that they would insert a block in its "default" state -- as if the user had just inserted the block manually into the editor, with no further customizations.

Thanks for the reminder. This will lead some future decision, like the above one.

> The create_block helpers are an attempt to make these things more transparent and maintainable, but aren't crucial to solving the actual problem at hand.

As we're now on the same page about injecting hooked blocks at its default state, the `create_block` helpers make a lot of sense. Can you work on merging your PR to this one?

plugins/woocommerce/client/blocks/assets/js/blocks/product-details/edit.tsx
```diff
@@ -19,7 +19,11 @@ import { ProductDetailsEditProps } from './types';
 const TEMPLATE: InnerBlockTemplate[] = [
 	[
 		'woocommerce/accordion-group',
-		{},
+		{
+			metadata: {
+				isProductDetailsInnerBlock: true,
```
dinhtungdu: We are using `isDescendantOf...` everywhere, so I will update this follow the current convention.

ockham:
> As we're now on the same page about injecting hooked blocks at its default state, the `create_block` helpers make a lot of sense. Can you work on merging your PR to this one?

I think it’s fine to land your PR without the `create_block` helpers 😊 Your PR is kind of feature complete, whereas mine still has some TODOs. So I think we shouldn’t block your work on those but can land the `create_block` helpers as a separate PR later 😄

dinhtungdu:
@woocommerce/woo-fse @ockham this PR is now ready for final review.

plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
```diff
@@ -263,13 +330,55 @@ private function has_accordion( $parsed_block ) {
...
-	 * @param string $key Data to get, or default to everything.
-	 * @return array|string|null
+	 * @param string $title The title of the item.
+	 * @param string $content The content of the item.
+	 * @return void
 	 */
-	protected function get_block_type_script( $key = null ) {
-		return null;
+	private function register_hooked_blocks( $title, $content ) {
+		$slug = sanitize_title( $title );
```
gziolo: Nit: definitely not a blocker as it might not be an issue in practical terms. Would it be necessary to introduce a static variable to keep track of registered slugs. At the moment, there is a risk that the same virtual block type names could get duplicated which might lead to unknown consequences as we deal with filters registration.

dinhtungdu: We can be more defensive by removing the item that have duplicated title with item added before it. I updated my branch and PR description to cover that case: d6b971cecd9d9ff43628cb0a33a25b4ecf033f99

plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
```diff
@@ -263,13 +301,111 @@ private function has_accordion( $parsed_block ) {
...
+	private function validate_hooked_blocks( $hooked_blocks ) {
+		$logger                  = wc_get_logger();
+		$validated_hooked_blocks = [];
+
+		foreach ( $hooked_blocks as $block ) {
+			$invalid = ! is_array( $block ) ||
+			! isset( $block['title'] ) ||
+			! isset( $block['content'] ) ||
+			! is_string( $block['title'] ) ||
+			! is_string( $block['content'] );
```
ockham: Nitpick, I think these need to be indented a bit more.

```suggestion
					! isset( $block['title'] ) ||
					! isset( $block['content'] ) ||
					! is_string( $block['title'] ) ||
					! is_string( $block['content'] );
```

plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
```diff
@@ -263,13 +301,111 @@ private function has_accordion( $parsed_block ) {
...
+
+			if ( array_filter(
+				$validated_hooked_blocks,
+				function ( $hooked_block ) use ( $block ) {
+					return $hooked_block['title'] === $block['title'];
+				}
+			) ) {
+				$logger->error( 'Duplicate hooked block data. Hooked block with title `' . $block['title'] . '` is already registered.', $block );
+				continue;
+			}
```
ockham: This only checks for identical titles, but I think we'd need to check for titles that would result in identical slugs _when  sanitized_; e.g. `sanitize_title( 'Cüstöm Information' )` gives `Custom Information`, but the `$hooked_block['title'] === $block['title']` check won't find that.

ockham: TBH I'm not even sure we need the duplicates check at all, since it might be okay for a block with a title that results in a slug that's already "taken" by another hooked block to override that other block -- it seems more or less of an arbitrary decision to me which block should "win".

An alternative could be to eschew slug collisions by e.g. adding a number (if there already is a `Custom Information`, the next one becomes `Custom Information `1), but I think that would be quite fragile (in case these blocks come from different plugins, and one of those plugins is later deactivated.)

plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
```diff
@@ -263,13 +301,111 @@ private function has_accordion( $parsed_block ) {
...
+	}
+
+	/**
+	 * Register a product details item using Block Hooks API.
+	 *
+	 * @param string $title The title of the item.
+	 * @param string $content The content of the item.
+	 * @return void
+	 */
+	private function register_hooked_blocks( $title, $content ) {
```
ockham: Nitpick: As it stands now, the function only registers one hooked block, so its name should probably reflect that:

```suggestion
	private function register_hooked_block( $title, $content ) {
```

ockham (commented):
This is looking very good! I left a few minor notes, but I think the PR is almost ready for merging 😄

plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
dinhtungdu: I think we need to ensure the final hooked blocks data doesn't have item with duplicated title, otherwise we will have more than one `"hooked_block_{$slug}"` filter for a same `slug`. The last one will be use, but previous callbacks are still executed and will be wasted.

I think the new question now is which one should we choose, first or the last. If we choose the last one, it may be easier for extender to override by setting a bigger priority value of their filter. If we choose the first one, the one hooked using smaller priority value (like `1`) is harder to be overridden.

I'm changing the logic to picking the last duplicated items, but I'm open to change.

ockham: That looks good to me. I particularly like that we're now using the slug as array keys, which allows us to "organically" avoid collisions (by overriding the previous block) for free.

ockham (approved):
LGTM! 🚢 Thank you 😄

dinhtungdu:
@gigitux as you worked on this subject before, can you take a look at this PR?

plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
```diff
@@ -263,13 +298,99 @@ private function has_accordion( $parsed_block ) {
...
+		$validated_hooked_blocks = [];
+
+		foreach ( $hooked_blocks as $block ) {
+			$invalid = ! is_array( $block ) ||
+					! isset( $block['title'] ) ||
+					! isset( $block['content'] ) ||
+					! is_string( $block['title'] ) ||
+					! is_string( $block['content'] );
+
+			$parsed_content = parse_blocks( $block['content'] );
```
gigitux: Should we continue with the flow if the first `$invalid` variable (L-316) is true?

gigitux (approved):
Awesome work! Everything works as expected:


| Editor | Frontend |
|--------|--------|
| image | image | 

I left a comment, but I'm going to approve this PR! 🚀

plugins/woocommerce/src/Blocks/BlockTypes/BlockifiedProductDetails.php
dinhtungdu: we do that below but we should move it earlier, after the variable declaration.

dinhtungdu: Now I get what you mean, yeah, we should

PR #57901, opened 2025-05-12
Add registerTemplateRestrictedBlockType API

### LINKED ISSUE #56198: registerTemplateRestrictedBlockType: Restrict Block Registration to Specific Templates
Some blocks such as the Catalog Sorting block is designed specifically for the WooCommerce Product Catalog template and should not be available for insertion or registration on unrelated templates, such as the homepage block template. This issue extends the discussion in https://github.com/woocommerce/woocommerce/issues/42387, where the block’s unrestricted registration was identified as a problem. To address this, we propose introducing a new JavaScript API for registering blocks with template-specific restrictions.

### Proposed Solution

Create a new JavaScript API, registerTemplateRestrictedBlockType, to allow developers to register blocks that are only available for specific templates. This API will extend the functionality of registerBlockType (from Gutenberg) and registerProductBlockType (from WooCommerce) by adding a templates parameter to specify allowed template IDs.

### API Signature

`registerTemplateRestrictedBlockType( name, settings, templates )`

* `name`: (string) The block’s unique name (e.g., `woocommerce/catalog-sorting`).
* `settings`: (object) The block settings object, consistent with `registerBlockType` and `registerProductBlockType`.
* `templates`: (array) An array of template IDs (e.g., `['product-catalog']`) where the block can be registered and inserted. If the current template is not in this array, the block will not be registered.

amitraj2203:

Closes woocommerce/woocommerce#56198

| Before | After |
| -- | -- |
|  |  |

1. 
2. 
3. 

### Changelog entry

#### Significance

- [ ] Patch
- [ ] Minor
- [ ] Major

#### Type

#### Message

</details>

#### Comment

</details>


tjcafferkey:
Just pinging to let you know I've not forgot about this PR, I will hopefully be able to review this week. If not, next week.

tjcafferkey:
So the reason you're seeing this issue you describe is because when we load the site editor we land on an ineligible template (e.g. `home`) and when you transition out of the edit template view and into the all templates view it does not reset the `getEditedPostId()` value to `undefined` even though you're not editing a template (so it still has the value `home`). Because of this, the catalog block is not registered.

As far as I can tell, there are no data store actions we can subscribe to in order to check this. But we could maybe determine if we are currently editing a template by checking for `canvas=edit` in the URL and if this is not present then we could assume we're on the all templates view. Just a thought...

amitraj2203:
Thanks @tjcafferkey for checking on this.

> As far as I can tell, there are no data store actions we can subscribe to in order to check this. But we could maybe determine if we are currently editing a template by checking for canvas=edit in the URL and if this is not present then we could assume we're on the all templates view.

Yes, this is how I was thinking of handling it as well. I also found this [hook](https://github.com/Automattic/wp-calypso/blob/532c4d6e3baeb3bc59a413cca0fbba11959ec46f/apps/help-center/hooks/use-canvas-mode.js) in Calypso, which uses a similar logic.

However, in our case, I was thinking we might need to create a store that checks and indicates whether we're in template edit mode. Then we can simply subscribe to that store. I tried this approach locally and was able to achieve the desired behavior.

Before moving forward with this, I wanted to get your opinion on whether this is a good approach or if you have any suggestions.

tjcafferkey:
>Yes, this is how I was thinking of handling it as well. I also found this [hook](https://github.com/Automattic/wp-calypso/blob/532c4d6e3baeb3bc59a413cca0fbba11959ec46f/apps/help-center/hooks/use-canvas-mode.js) in Calypso, which uses a similar logic.

Ah interesting, great find. It's reassuring to see that they're using a similar approach to what we thought. Unfortunately we can't use a hook in our case because we're not in a React component here (although you probably know that already).

>However, in our case, I was thinking we might need to create a store that checks and indicates whether we're in template edit mode. Then we can simply subscribe to that store. I tried this approach locally and was able to achieve the desired behavior.

I think this could be a good contribution to the Gutenberg repository(maybe [here](https://github.com/WordPress/gutenberg/tree/trunk/packages/edit-site/src/store)) which we could then use in WooCommerce once it gets released.

In the meantime it might be a good idea to open a separate PR with your WooCommerce data store, with a view that once merged (if it gets approves) we could use it in this PR. One thing we will need to do though is make it clear that this new data store is private and for Woo internal use only. This way once Gutenberg has its own implementation we can remove ours safely.

nerrad:
Closing because this is a stale draft issue

*\*left by Biff (AI agent) on behalf of Darren*

PR #57912, opened 2025-05-12
Product Filters: fix Chips items and Show more buttons font family

dinhtungdu:

This PR fixes the issue with the style of Chips and Checkbox List blocks that uses the incorrect font family with themes that don't set the `font-family` property for the generic `button` element. We reset the button style for Chips items and Show more button using `appearance: none;`, so the `font-family` property can fall back to the browser default. This is not a theme compatibility fix because the Chips items and Show more buttons should look good by itself.

Closes [WOOPLUG-4244](https://linear.app/a8c/issue/WOOPLUG-4244/chips-item-block-uses-incorrect-font-family-with-certain-themes)

| Before | After |
| ------ | ----- |
|   image     |   image    |

1. Before checking out this PR, test `trunk` with [Pixl theme](https://wordpress.org/themes/pixl/).
2. Add Product Filters block to the Product Catalog template.
3. Set Attribute block display style to Chips.
4. Verify the font family of Chips items on the frontend is Arial, not `"DM Mono", sans-serif`.
5. Check out and build this PR.
6. Verify the Chips item font family is `"DM Mono", sans-serif`.

### DIFF
--- a/plugins/woocommerce/changelog/fix-chip-item-font-family
+++ b/plugins/woocommerce/changelog/fix-chip-item-font-family
@@ -1,0 +1,5 @@
+Significance: patch
+Type: fix
+Comment: Product Filters: fix Chips item font family for theme that doesn't set the font-family for generic button element.
+
+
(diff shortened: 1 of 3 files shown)


kmanijak (changes requested):
I confirm Chips have a correct font, however, `Show more` doesn't:

image

Could we apply it to Show More as well? Here are the classes of a button:
- `wc-block-product-filter-checkbox-list__show-more`
- `wc-block-product-filter-chips__show-more`

dinhtungdu:
Great catch! I fixed it in 3419890484c9. Thank you!

kmanijak (approved):
Nice one! 😎 

image

Issue #34358, opened 2022-08-17
[Enhancement]: wc_customer_bought_product() Performance Improvement


nicholasjeberlin:
### Describe the solution you'd like

I have a merchant who has shared that they have found a change that provides a large performance improvement to the `wc_customer_bought_product` function located in woocommerce/includes/wc-user-functions.php.

> The improvement has to do with how the SQL transaction for lookup is ordered. Below is our change SQL transaction. We adjusted how the WHERE clauses were ordered to cut down on the returned data set faster. Could you consider changing this code in a future release of WooCommerce? Thank you.

```
SELECT im.meta_value FROM {$wpdb-&gt;posts} AS p
INNER JOIN {$wpdb-&gt;postmeta} AS pm ON p.ID = pm.post_id
INNER JOIN {$wpdb-&gt;prefix}woocommerce_order_items AS i ON p.ID = i.order_id
INNER JOIN {$wpdb-&gt;prefix}woocommerce_order_itemmeta AS im ON i.order_item_id = im.order_item_id
WHERE pm.meta_key IN ( '_billing_email', '_customer_user' )
AND im.meta_key IN ( '_product_id', '_variation_id' )
AND p.post_status IN ( 'wc-" . implode( "','wc-", $statuses ) . "' )
AND pm.meta_value IN ( '" . implode( "','", $customer_data ) . "' )
AND im.meta_value != 0
```
Leaving this here for input from developers and the community.

nicholasjeberlin:
5485491-zen

kaushikasomaiya:
7070997-zen

Babylon1999:
zd-7156164

https://wordpress.org/support/topic/wc_customer_bought_product-slows-down-page-massively/

PR #57916, opened 2025-05-12
[Performance] wc_customer_bought_product function performance optimization

kalessil:

Closes WOOPLUG-790 #34358

- Adds filtering the post tables by type so that only orders (not pages, attachments, products, refunds, etc.) are accounted for in the joins in the fallback SQL query (no HPOS nor lookup tables enabled).
- Switches caching API from transients to cache.
- Optimizes cache versioning when the function is called for registered users (when no lookup tables are enabled) for longer cache retention.
- Optimizes the function usage across the core codebase.

Explain results: before-after (note the drop in rows fetched from the posts table)

Screenshot from 2025-05-12 16-01-24
Screenshot from 2025-05-12 16-01-42

This translates to equal SQL performance improvements: before-after

```log
[2025-05-12 16:04:42] 2 rows retrieved starting from 1 in 66 ms (execution: 42 ms, fetching: 24 ms)
[2025-05-12 16:04:54] 2 rows retrieved starting from 1 in 55 ms (execution: 32 ms, fetching: 23 ms)
```

Rough calculations: 30% less post table rows extracted = ( ( 42 - 32 ) * 100 / 42 ) = 24% faster SQL (benchmarking method is not great, so it's a rough estimate).

- Enable all three checkboxes in `Enable reviews` section under `WooCommerce -> Settings -> Products`
- Install/enable a drop-in cache for the store
- Create a new customer account (or use an existing one) and log into the store
- Make a test purchase
- Leave a product review for a purchased product
- Verify: the comment meta `verified` is present in `wp_commentmeta` table and the cache entry `wc_customer_bought_product_*` is populated

The cache entry can also be verified by performing the testing steps, adding some breakpoint and refreshing the product page. Break point reference:
```patch
Subject: [PATCH] Performance: verify.
---
Index: plugins/woocommerce/includes/wc-user-functions.php
IDEA additional info:
<+>UTF-8
diff --git a/plugins/woocommerce/includes/wc-user-functions.php b/plugins/woocommerce/includes/wc-user-functions.php
--- a/plugins/woocommerce/includes/wc-user-functions.php	(revision c91510ea3f245236016de921da0a11e2c5ff239c)
+++ b/plugins/woocommerce/includes/wc-user-functions.php	(date 1747143726386)
@@ -444,7 +444,7 @@
 	$cache_value = wp_cache_get( $cache_key, $cache_group );
 
 	if ( isset( $cache_value['value'], $cache_value['version'] ) && $cache_value['version'] === $cache_version ) {
-		$result = $cache_value['value'];
+		$result = $cache_value['value']; // Break point / logging statement
 	} else {
 		$customer_data = array( $user_id );
```


plugins/woocommerce/includes/wc-user-functions.php
```diff
@@ -424,19 +424,27 @@ function wc_customer_bought_product( $customer_email, $user_id, $product_id ) {
...
 	if ( $use_lookup_tables ) {
 		// Lookup tables get refreshed along with the `woocommerce_reports` transient version.
-		$transient_version = WC_Cache_Helper::get_transient_version( 'woocommerce_reports' );
+		// Sync is taking place async, hence we bound to the transient version.
+		// With high orders placement rate, this caching here will be short-lived (suboptimal for BFCM/Christmas and busy stores in general).
+		$cache_version = WC_Cache_Helper::get_transient_version( 'woocommerce_reports' );
+	} elseif ( '' === $customer_email && $user_id ) {
+		// Optimized: for specific customers rely on orders count (customer meta under the hood with in-memory caching in the way to it).
+		// Best-case scenario for caching here, as it only depends on the customer orders placement.
+		$cache_version = wc_get_customer_order_count( $user_id );
```
kalessil, line 434: The bottom line is that the counter is a user meta. At this point the user object is fully populated (the relevant use-cases are for the current user or comments users) and present in memory - hence we can skip transients here altogether.

plugins/woocommerce/includes/wc-user-functions.php
```diff
@@ -424,19 +424,27 @@ function wc_customer_bought_product( $customer_email, $user_id, $product_id ) {
...
 	}
 
-	$transient_value = get_transient( $transient_name );
+	$cache_group = 'orders';
+	$cache_key   = 'wc_customer_bought_product_' . md5( $customer_email . '-' . $user_id . '-' . $use_lookup_tables );
+	$cache_value = wp_cache_get( $cache_key, $cache_group );
 
-	if ( isset( $transient_value['value'], $transient_value['version'] ) && $transient_value['version'] === $transient_version ) {
-		$result = $transient_value['value'];
+	if ( isset( $cache_value['value'], $cache_value['version'] ) && $cache_value['version'] === $cache_version ) {
```
prettyboymp, line 446: We might be able to simplify this if we move the `version` to the cache key, but I'm good with how it is.

prettyboymp (approved):
Confirmed, works as expected.

plugins/woocommerce/includes/wc-user-functions.php
kalessil: I had similar thoughts—it would be a good simplification, but with lookup tables, it spawns too many cache entries (one for each new order placed), so I left storing the version inside.

PR #57917, opened 2025-05-12
Product Filters: fix filter blocks visibility after navigation

dinhtungdu:

In other filter blocks except Price, we inject the hidden attribute through the `get_block_wrapper_attributes()` function, which sets the attribute like this: `hidden="1"`. iAPI keeps this attribute so it's injected to the DOM after navigation.

The Price Filter block doesn't use the `get_block_wrapper_attributes()` function, but declares the `hidden` attribute directly, without the value (`="1"`). iAPI discards that attribute so the "empty" Price filter block is visible after navigation.

This PR fixes the issue by using the `get_block_wrapper_attributes()` function for the Price filter block as well. In addition, we add a `.hidden` class and use it for all filter blocks to better control the visibility of the filter blocks. The `hidden` attribute is still used for accessibility reasons.

Closes [WOOPLUG-4245](https://linear.app/a8c/issue/WOOPLUG-4245/price-filter-block-visibility-issue-after-navigation)

| Before | After |
| ------ | ----- |
|  image      |   image    |

1. Add the Product Filters block to the Product Catalog template.
2. Enable Product Count for Attribute and Status blocks.
3. On the frontend, select a filter that has `1` as the count.
4. On `trunk`, verify the Price title is still visible.
5. On this PR, verify the Price filter block and its title are hidden.

### DIFF
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/style.scss
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/style.scss
@@ -161,6 +161,6 @@
 }
 
 // We don't want zero specificity for this style.
-.wc-block-product-filters .hidden {
+.wc-block-product-filter--hidden {
 	display: none;
 }
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/style.scss
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/style.scss
@@ -159,3 +159,8 @@
 		}
 	}
 }
+
+// We don't want zero specificity for this style.
+.wc-block-product-filters .hidden {
+	display: none;
+}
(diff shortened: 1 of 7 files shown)


plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/style.scss
```diff
@@ -159,3 +159,8 @@
 		}
 	}
 }
+
+// We don't want zero specificity for this style.
+.wc-block-product-filters .hidden {
```
kmanijak: I'd consider a bit more specific class name than `hidden`, e.g. `.wc-block-product-filter--hidden`.

`hidden` I think there's a high chance some 3rd-party code with non scoped selectors (simply targeting `.hidden`) could use this class to actually hide all filters.

kmanijak (approved):
It works correctly! However I strongly suggest to make class more specific to avoid 3rd party styles leaking into filters!

Pre-approving though as it works fine and change would be rather small.

kmanijak (changes requested):
Automerge enabled, removing "pre-approve"!

kmanijak (approved):
Thanks for quickly addressing the comment. Approved! 👍

PR #57920, opened 2025-05-12
Make sure package items name don't break at comma

senadir:

Fixes a bug in which a package with multiple items would break around comma separator, causing a broken UI. Most likely due to the fact that we use Label which uses RawHTML which always create a div, which is by default block, and the comma being on the parent, not on the div.

Another solution would be to move to our own RawHTML component that doesn't create the extra div.
Closes WOOPLUG-4219

| Before | After |
| ------ | ----- |
|    image  |  image  |

1. Install a plugin that can generate multiple packages, I've used WooCommerce Subscriptions, you can use https://wordpress.org/plugins/wc-shipping-packages/#developers
2. Add multiple items to your cart, ensuring at least 2 items will end up in a single package (regular + subscription product will get you there).
3. On Cart, ensure package item names don't break at comma level.
4. On Checkout, ensure the same thing.
5. Change one item name to be a long name, to simulate breaking at end of line, like `Office Chair Ergonomic Desk Chair, 330 LBS Home Mesh Office Desk Chairs with Wheels, Comfortable Gaming Chair, High Back Office Chair for Long Hours (Black)`
6. On Checkout, ensure it doesn't break at the comma and text flows naturally.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Fix issue in which package item names would show up malformed on Checkout.
</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/shipping-rates-control-package/style.scss
```diff
@@ -73,9 +73,13 @@
...
 	margin: 0;
 	padding: 0;
+	> div {
+		// Usage of RawHTML component from WordPress always adds a div wrapper.
+		display: inline;
+	}
 }
 
-.wc-block-components-shipping-rates-control__package-item:not(:last-child)::after {
+.wc-block-components-shipping-rates-control__package-item:not(:last-child) > div::after {
```
tjcafferkey: Nit: I'd also add the comment here about the `RawHTML` generated div wrapped.

tjcafferkey (approved):
Tested with multiple different types of products. Works great. Left a non-blocking comment.

PR #57951, opened 2025-05-13
Render an experimental iAPI powered mini cart behind a feature flag

samueljseay:
## Changes

In our exploration so far of different approaches to migrate the mini cart we explored approaches such as partially rendering some of the React components of the existing mini cart via Preact/interactivity API.

These approaches proved highly complex and problematic. It was impossible to use the existing components "as is" due to dependencies and sub dependencies on `@wordpress/element` and many other issues.

We've decided to take a step back and migrate the mini cart behind a feature flag. But it's our goal not to introduce new experimental blocks that we'll need to migrate the existing mini cart to later. Instead, we're using a feature flag to render the existing block (and eventually inner blocks) with interactivity API directives. In the hope that it might be as simple as switching off the feature flag when we have something ready.

This approach is not perfect, it will introduce some forking code in some places, but since a lot of the existing inner blocks are almost entirely client side rendered at the moment, we expect that forking code and duplication will be minimal. I've tried to limit it as much as possible here in the top level Mini Cart.

The goal of this PR is set up so we can start the work of building the new mini cart, not to provide anything meaningfully functional at this point.

This PR:

1. Add's a feature flag, it is turned off by default in all environments to avoid confusion and problems. If you want to develop the experimental mini cart you'll need to explicitly turn the feature flag on in development. This is different from other feature flags where we might just enable an extra block in the inserter during development. This will break the existing mini cart in development so it's important to opt-in.
2. Add's minimal forking code in the `MiniCart.php` and `MiniCartContents.php` blocks as possible to render a separate iAPI directive powered Mini Cart. I've opted for some duplication to avoid complexity, and I've opted not to render the iAPI mini cart in the existing render to keep it clean and reduce complexity and burden to anyone trying to maintain the existing mini cart. 
3. I've also split out the code that gets the template contents in the MiniCart (this is where I drew the line at too much duplication).
4. Adds a very basic front-end that allows clicking the mini cart badge to open the drawer. This is just POC and we will follow up with PR's that will actually implement this functionality
5. Adds Webpack entry points to build the frontend code for the iAPI powered blocks

## Testing

The experimental mode does not need explicit testing at this point, it has some basic functionality, but that's not important.

The one important thing to test is just that the experimental mode does not interfere or break mini cart functionality:

1. Checkout this PR and build on a WooCommerce site 
2. Load any page with the mini cart, ensure that it works as before, add/remove products, open close the cart, check for dev console errors.

If you like you can also test that switching the flag on also changes the implementation:

1. Install WooCommerce beta tester, enable the `experimental-iapi-mini-cart` feature flag
2. Load any page with mini cart, see that clicking the cart opens a blank drawer and that clicking the drawer close button closes the cart
### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 
Render POC interactivity API powered mini cart behind a feature flag.

</details>


luisherranz (approved):
LGTM! 🎉

I have tested without the feature flag, and in principle, everything seems to be working fine.

A minor detail is that you don't really need to create two different frontend files if the MiniCart block and the MiniCart Contents block are always going to be rendered together. One for the MiniCart would suffice.

samueljseay:
> A minor detail is that you don't really need to create two different frontend files if the MiniCart block and the MiniCart Contents block are always going to be rendered together. One for the MiniCart would suffice.

That's a good point and would simplify things. I'll fix that.

samueljseay:
I've consolidated the f/e code to a single entry point, also adjusted the rendering of `MiniCartContents` which was not rendering the inner blocks, which we'll need for next steps

pnpm-lock.yaml
dinhtungdu, line 1: Is this change necessary?

plugins/woocommerce/changelog/57951-dev-experiment-render-iapi-mini-cart
dinhtungdu, line 1: We probably want to use comment for this type of change, this probably doesn't need to go into changelog.

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartContents.php
```diff
@@ -54,6 +55,42 @@ protected function get_block_type_style() {
...
+	 * @return string Rendered block type output.
+	 */
+	protected function render_experimental_iapi_mini_cart_contents( $attributes, $content, $block ) {
+		// As a temporary workaround we remove the wrapping div from the template.
+		$content = preg_replace(
+			'~<div([^>]*\bclass=["\'][^"\']*\bwp-block-woocommerce-mini-cart-contents\b[^"\']*["\'][^>]*)>~',
+			'',
+			$content,
+			1
+		);
```
dinhtungdu, line 73: In the future, we can remove the wrapper div from the save function and use Deprecation API to update the block in the editor, but for the frontend rendering, we will need to remove the div like this.

plugins/woocommerce/src/Blocks/BlockTypes/MiniCart.php
```diff
@@ -417,17 +422,115 @@ protected function get_include_tax_label_markup( $attributes ) {
...
+		$icon                     = MiniCartUtils::get_svg_icon( $attributes['miniCartIcon'] ?? '', $icon_color );
+		$product_count_visibility = isset( $attributes['productCountVisibility'] ) ? $attributes['productCountVisibility'] : 'greater_than_zero';
+		$wrapper_classes          = sprintf( 'wc-block-mini-cart wp-block-woocommerce-mini-cart %s', $classes_styles['classes'] );
+
+		$template_part_contents = $this->get_template_part_contents();
+
+		ob_start();
+		?>
+		<?php // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
+		<div <?php echo wp_interactivity_data_wp_context( $context ); ?> data-wp-interactive="woocommerce/mini-cart" class="<?php echo esc_attr( $wrapper_classes ); ?>" style="<?php echo esc_attr( $wrapper_styles ); ?>">
```
dinhtungdu, line 469: `$wrapper_styles` is undefined

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
```diff
@@ -0,0 +1,29 @@
...
+import { store, getContext } from '@wordpress/interactivity';
+
+store( 'woocommerce/mini-cart', {
+	state: {
+		get drawerOverlayClass() {
+			const { isOpen } = getContext< { isOpen: boolean } >();
+
+			return ! isOpen
+				? 'wc-block-components-drawer__screen-overlay wc-block-components-drawer__screen-overlay--is-hidden'
+				: 'wc-block-components-drawer__screen-overlay wc-block-components-drawer__screen-overlay';
```
dinhtungdu, line 13: The two classes are identical.

Btw, should we be more declarative

```
<div
    data-wp-bind-class--wc-block-components-drawer__screen-overlay--is-hidden="!context.isOpen"
    class="is-loading wc-block-components-drawer__screen-overlay wc-block-components-drawer__screen-overlay--is-hidden"
    aria-hidden="true">```

**Review by dinhtungdu — COMMENTED** [review]
Thanks for working on this PR. The PR is testing great on my end. About the code change, beside some minor comments, I have a question/idea: what do you think about creating a whole new classes for Mini Cart and Mini Cart content, and load them conditionally based on the flag?

File: plugins/woocommerce/changelog/57951-dev-experiment-render-iapi-mini-cart
Comment (samueljseay) [reply plugins/woocommerce/changelog/57951-dev-experiment-render-iapi-mini-cart:1]: Sure

File: plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts (hunk 3 above)
Comment (samueljseay) [reply plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts:13]: So unfortunately I did update and improve a lot of this in my follow up PR. since it's experimental feature flag code is it OK if we update the approach there? https://github.com/woocommerce/woocommerce/pull/58265/files#diff-e5321b2c65fe5066069deccc33f89005ef1a93a664f49f9791960613fbf35e4fR37

File: plugins/woocommerce/src/Blocks/BlockTypes/MiniCart.php (hunk 2 above)
Comment (samueljseay) [reply plugins/woocommerce/src/Blocks/BlockTypes/MiniCart.php:469]: Ah yeah also fixed in the follow up PR, but I'll fix it here.

File: plugins/woocommerce/src/Blocks/BlockTypes/MiniCartContents.php (hunk 1 above)
Comment (samueljseay) [reply plugins/woocommerce/src/Blocks/BlockTypes/MiniCartContents.php:73]: This quickly became untenable because I needed regexes for all inner blocks and it wasn't easy to reliably strip the outer tags in all inner blocks. In my follow up PR i've allowed overriding the template to one without the wrapper divs:

https://github.com/woocommerce/woocommerce/pull/58265/files#diff-e9aad967c68600c33d8102afce406254c9a0805c3e2bc939c8610612010abd3cR463

https://github.com/woocommerce/woocommerce/pull/58265/files#diff-e9aad967c68600c33d8102afce406254c9a0805c3e2bc939c8610612010abd3cR550

We can move to deprecation API as well later, but is changing the template a viable approach here? 

If it's not we can remove the duplicate classes instead via the `WP_HTML_Tag_Processor`, but we'll still need to adjust CSS to fix broken situations where the nested div breaks things like display flex etc.

Perhaps we can explore how we'll fix this situation nearer to releasing the feature complete iAPI mini cart

File: pnpm-lock.yaml
Comment (samueljseay) [reply pnpm-lock.yaml:1]: Sadly pnpm lock generation is unstable, yeah you're right this isn't needed thanks :)

**Comment by samueljseay** [comment]
> Thanks for working on this PR. The PR is testing great on my end. About the code change, beside some minor comments, I have a question/idea: what do you think about creating a whole new classes for Mini Cart and Mini Cart content, and load them conditionally based on the flag?

@dinhtungdu 

Please could you, if you have an idea like that also include your rationale for why it'd be better? It will just save time in async discussion next time :) 

Trying to infer your rationale for the change:

I certainly considered this and other solutions, but, I think I'm just going to say: no I don't think so this time.

I think what we have here is simple enough, and has allowed me to rebuild [most of the mini cart](https://github.com/woocommerce/woocommerce/pull/58265) in a matter of days. I don't think creating new classes here has additional benefit, and also if we're keeping block names the same we'll need to not register the existing classes for those blocks in development.

That solution doesn't address rendering in any of the inner blocks either (again if you could elaborate your idea more in future how it'd deal with the details it would also help discussion). 

Afaict we'd still need to create new blocks for the whole tree (or it's questionable value, why only create new blocks for mini cart and mini cart contents when the rest is still conditional rendering?). Maybe your concern is that cart and cart contents are more complex, but they aren't really, it's just more rendering code than the other blocks.

The idea is, when this is ready, we can simply remove the old rendering code.

**Review by albarin — APPROVED** [review]
✅ Confirmed the mini-cart works fine and there are no errors in the console when the feature flag is disabled.
✅ Confirmed the blank mini-cart shows the feature flag is enabled.

File: plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts (hunk 3 above)
Comment (dinhtungdu) [reply plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts:13]: Sure, no problem :D

File: plugins/woocommerce/src/Blocks/BlockTypes/MiniCartContents.php (hunk 1 above)
Comment (dinhtungdu) [reply plugins/woocommerce/src/Blocks/BlockTypes/MiniCartContents.php:73]: Yeah, using different template sounds like a better approach. And yes, we can explore it further when it comes to releasing new Mini Cart.

**Comment by dinhtungdu** [comment]
> Please could you, if you have an idea like that also include your rationale for why it'd be better? It will just save time in async discussion next time :)

I'm sorry about this. I will make sure do give more context next times.

The reason I suggest that is to avoid too much branching in the codebase. Instead of having the same condition in many methods, we can have only one in the BlockTypesController. My idea is duplicating the whole Mini Cart and Mini Cart Content classes and editing the duplicated ones like you're doing here. The block name, setting, everything are still the same. When it come to releasing, we can just remove the conditional registration in the BlockTypesController. I think doing so can give us freedom to shape the new class without worrying about accidentally modify the current block.

**Review by dinhtungdu — APPROVED** [review]
I'm approving this one as it looks good and to unblock other PRs. There are some issues but the change is gated behind an experimental flag so it's safe to go. We can give the "duplicating classes" idea a more careful consideration after this PR.

File: plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts (hunk 3 above)
Comment (luisherranz) [reply plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts:13]: A small comment: With the iAPI's SSR, you don't need to add the class name to the markup; the iAPI takes care of that based on the value of `context.isOpen` on the server.

So instead of this:

```html
<div
    data-wp-bind-class--wc-block-components-drawer__screen-overlay--is-hidden="!context.isOpen"
    class="is-loading wc-block-components-drawer__screen-overlay wc-block-components-drawer__screen-overlay--is-hidden"
    aria-hidden="true"
>
```

You can write this:

```html
<div
    data-wp-bind-class--wc-block-components-drawer__screen-overlay--is-hidden="!context.isOpen"
    class="is-loading wc-block-components-drawer__screen-overlay"
    aria-hidden="true"
>
```

samueljseay: That's a good point. I don't know if I concretely thought of that here, but I did avoid class binding in general so far because the class names are just so long, which makes the markup look quite unwieldy.

maybe when we're closer to release we can also consider refactoring the css and reduce the length of some of the class names.

PR #57952, opened 2025-05-13
[Payments NOX] Harden code against failures to determine gateway details

### LINKED ISSUE #56267: Uncaught TypeError in PaymentGateway::get_plugin_file()
### Prerequisites

### Describe the bug

A `TypeError` can occur in `Automattic\WooCommerce\Internal\Admin\Settings\PaymentProviders\PaymentGateway::get_plugin_file()` if `PluginsHelper::get_plugin_path_from_slug()` returns `false` [here](https://github.com/woocommerce/woocommerce/blob/bf4226aa1ac4619eff5da8faca7311b1b90d57c1/plugins/woocommerce/src/Internal/Admin/Settings/PaymentProviders/PaymentGateway.php#L444).

Example error:

```
Fatal error: Uncaught TypeError: Automattic\WooCommerce\Internal\Admin\Settings\PaymentProviders\PaymentGateway::get_plugin_file(): Return value must be of type string, bool returned
```

`PluginsHelper::get_plugin_path_from_slug()` can return `false` if `PaymentGateway::get_plugin_file()` was called with a gateway instance whose class file is in a non-standard location like `/domXYZ/wp-content/mu-plugins/vendor/org/project/src/Path/To/Gateway.php`.

In that case, on WooCommerce 9.7.1, `PaymentGateway::get_plugin_file()` would be called with the gateway instance and a `'vendor'` plugin slug, which is not enough for `PluginsHelper::get_plugin_path_from_slug()` to find the plugin path.

### Expected behavior

`PaymentGateway::get_plugin_file()` should return an empty string.

### Actual behavior

```
Fatal error: Uncaught TypeError: Automattic\WooCommerce\Internal\Admin\Settings\PaymentProviders\PaymentGateway::get_plugin_file(): Return value must be of type string, bool returned
```

### Steps to reproduce

The path that leads to the execution of `Automattic\WooCommerce\Internal\Admin\Settings\PaymentProviders\PaymentGateway::get_plugin_file()` seems to be behind a feature flag that I don't know how to properly toggle on.

However, if you execute the following snippet on a site running WooCommerce 9.7.1, you should get a `Fatal error`:

```php
$pg = new Automattic\WooCommerce\Internal\Admin\Settings\PaymentProviders\PaymentGateway();

$pg->get_plugin_file(WC()->payment_gateways->payment_gateways()['cod'], 'vendor');
```

### WordPress Environment

[SystemStatusReport_2025-03-10T15-48-17.txt](https://github.com/user-attachments/files/19166200/SystemStatusReport_2025-03-10T15-48-17.txt)

### Isolating the problem

vladolaru:

As we received reports from our users about issues they encountered with the new Payments Settings page, we discovered places where we needed to better handle failures around extracting payment gateway details.
This PR introduces missing defenses around missing or wrong type gateway details and ensures we cover these cases with PHPUnit tests.

Closes WOOPLUG-4264
Closes WOOPLUG-3393

1. The code changes should make sense.
2. The unit tests should pass.
3. Smoke test the new Payments Settings page.

#### Testing with a theme-provided gateway

1. Create a new JN site with WooCommerce on `trunk` (here is a [direct create link](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce=trunk))
1.  Go to the new site's WP admin, note the SSH credentials, and connect through SSH to the JN site.
1. Go to Appearance > Themes > Add theme and install&activate this customized Storefront version 
[custom-storefront-with-gateway.zip](https://github.com/user-attachments/files/20205667/custom-storefront-with-gateway.zip), which registers a custom BACS gateway - a BAD way to do it.
1. When you activate the theme, the WP admin will crash with `There has been a critical error on this website....` error.
1. Via SSH, activate the default theme by running the following command: `wp theme activate twentytwentyfour`
1. Go to your WP admin, and it should not be crashing anymore.
1. Go to Jetpack > Beta Tester > WooCommerce and click Manage. Once on the Jetpack Beta Tester > WooCommerce page, search for `gateway plugin` in the Feature branch field and activate this PR's build:
Screenshot 2025-05-14 at 13 33 56
1. Via SSH, activate the custom Storefront theme by running the following command: `wp theme activate custom-storefront`
1. Go to your WP admin, and it should not be crashing.
1. Go to WooCommerce > Settings > Payments and you should see a "[Custom - From Storefront] Direct bank transfer" provider:
Screenshot 2025-05-14 at 13 40 46

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Internal/Admin/Settings/PaymentProviders.php
```diff
@@ -1074,7 +1074,7 @@ private function enhance_extension_suggestion( array $extension ): array {
 				if ( PluginsHelper::is_plugin_installed( $plugin_slug ) ) {
 					// Make sure we put in the actual slug and file path that we found.
 					$extension['plugin']['slug'] = $plugin_slug;
-					$extension['plugin']['file'] = PluginsHelper::get_plugin_path_from_slug( $plugin_slug );
+					$extension['plugin']['file'] = (string) PluginsHelper::get_plugin_path_from_slug( $plugin_slug );
```
dmallory42: It looks like this function will return `false` when it is not able to determine the path. Rather than casting to a string, I wonder if it would be better to either:
- Implement a new function which falls back to a sensible default instead of returning false
- Use a coalescence operator to fall back to a sensible default like `PluginsHelper::get_plugin_path_from_slug( $plugin_slug ) ?? ''`;

vladolaru: The coalescing operator is only for `null`, not `false`, so it doesn't help us here. I decided to go with casting since `false` will be cast to an empty string.

> Implement a new function which falls back to a sensible default instead of returning false

I think the behavior of `PluginsHelper::get_plugin_path_from_slug` is useful since it doesn't hide failures behind empty strings. And it would not make sense to have a `default` like parameter, either, since a plugin file path doesn't have a default - we either have it or we don't.

WDYT?

dmallory42: > I think the behavior of PluginsHelper::get_plugin_path_from_slug is useful since it doesn't hide failures behind empty strings. And it would not make sense to have a default like parameter, either, since a plugin file path doesn't have a default - we either have it or we don't.

This makes sense, I think upon reflection the current function behaviour is sensible.

> The coalescing operator is only for null, not false, so it doesn't help us here. I decided to go with casting since false will be cast to an empty string.

Good point! This is nitpicky, but I prefer it to be more verbose/explicit that the value will be `''` in case `false` is returned - but this is mainly just personal preference/because I always forget how PHP casting rules work. 😅

vladolaru: @dmallory42 I've made changes to be explicit about the string type.

dmallory42: Thanks, I think that the intent is much clearer now!

oaratovskyi (approved):
💪 Thanks for improving the gateway details, it's much more robust now.

The changes look good to me. Thanks for going extra mile with tests 🚀 

I've tested according to the instructions and additionally smoke tested enabling/disabling, activating/deactivating the gateway. Created WooPayments and Stripe account - all good :shipit:

PR #57953, opened 2025-05-13
Add email toggle into the email editor sidebar

pavel-mailpoet:

Added email status toggle in the email editor sidebar. Added the ability to enable/disable transactional emails directly from the email editor

| Before | After |
| ------ | ----- |
|    Screenshot 2025-05-07 at 15 13 08    |     Screenshot 2025-05-13 at 13 20 19  |

1. Make sure the Block Email Editor feature is enabled in WooCommerce > Settings > Advanced > Features
2. Navigate to WooCommerce > Settings > Emails
3. Open emails with different statuses in the email editor and verify that the status is displayed correctly in the sidebar
4. Change the status in the sidebar of the email editor and verify that the change is propagated in the list of the emails

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/email_status.tsx
```diff
@@ -0,0 +1,156 @@
+/**
```
yuliyan: Is using `snake_case` for the file name here intentional? I noticed that `sidebar_settings.tsx` also uses `snake_case`. While I couldn't find any specific rules about file naming in [the naming convention docs](https://github.com/woocommerce/woocommerce/blob/fc221eddf43ca4955f74f73f11842ceb8f9abbd8/docs/quality-and-best-practices/naming-conventions.md), it looks like most of the repo uses `kebab-case`. It might be better to stick with that for consistency.

plugins/woocommerce/src/Internal/EmailEditor/EmailApiController.php
```diff
@@ -52,6 +52,8 @@ public function get_email_data( $post_data ): array {
 		$email       = $this->get_email_by_type( $email_type );
 
 		return array(
+			'enabled'         => isset( $post_option['enabled'] ) && 'yes' === $post_option['enabled'] ? true : false,
```
yuliyan: It looks like the current logic defaults to `false` when `'enabled'` isn't set in the post options. However, `WC_Email` [defaults this property to `'yes'`](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/emails/class-wc-email.php#L990-L995), which results in reporting different status in the editor compared to the listing (and likely when used to determine whether the email should be sent).

Would it make sense to use `$email->is_enabled()` instead?

plugins/woocommerce/client/admin/client/settings-email/settings-email-listing-status.tsx
```diff
@@ -15,16 +15,25 @@ export const EMAIL_STATUSES = [
...
 		icon: published,
+		description: __(
+			'Email would be sent if trigger is met',
+			'woocommerce'
+		),
 	},
 	{
 		value: 'disabled',
 		label: __( 'Disabled', 'woocommerce' ),
 		icon: cancelCircleFilled,
```
yuliyan, line 26: Not related to the changes in this PR, but while checking the post status in Core, I noticed that one of the statuses uses `notAllowed`, which may align a little bit better with the style (non-filled icon) of the icons for `enabled` and `manual`.

plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/email_status.tsx
```diff
@@ -0,0 +1,156 @@
...
+		);
+	};
+
+	const renderDropdownContent = ( {
+		onClose,
+	}: {
+		onClose: () => void;
+	} ): JSX.Element => (
+		<div style={ { minWidth: 230 } }>
+			<Flex
```
yuliyan: Can we use `__experimentalInspectorPopoverHeader` here instead, so that we match the styles in the Core editor for these type of popovers?

For reference, here's how it's implemented for the post status - https://github.com/WordPress/gutenberg/blob/7e47016787a2bef92da639c6e1f33208ccc5b2a1/packages/editor/src/components/post-status/index.js#L193.

plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/email_status.tsx
```diff
@@ -0,0 +1,156 @@
...
+ */
+export function EmailStatus( { className }: EmailStatusProps ): JSX.Element {
+	const [ woocommerce_email_data ] = useEntityProp(
+		'postType',
+		'woo_email',
+		'woocommerce_data'
+	);
+
+	const isManual = woocommerce_email_data?.is_manual;
+	let statusValue = 'disabled';
```
yuliyan: To maintain consistency with the `WC_Email` object, we can set the default status to `'enabled'`, even though the `enabled` property should always be present in `woocommerce_email_data`, and there should be no scenarios where its behavior differs.

yuliyan (changes requested):
Updating the status from the sidebar works well, but the issue with the default value for the enabled status should be resolved before merging the changes.

plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/email_status.tsx
pavel-mailpoet: I was not sure which to use, as I saw both styles and didn't look into the documentation. Renaming.

plugins/woocommerce/client/admin/client/wp-admin-scripts/email-editor-integration/email_status.tsx
pavel-mailpoet: I am not sure. I saw the implementation and decided not to use the experimental feature. It could be a problem with maintenance when the experimental features change in the future.

yuliyan (approved):
Thanks for addressing the comments, LGTM!

https://github.com/user-attachments/assets/c6d37d54-fef8-43bd-b4d5-a567741c7bf1

PR #57958, opened 2025-05-13
Add support for Personalization Tags in the Button block

yuliyan:

This PR includes the following updates:
* Resolves the issue with inserting personalization tag as a text for the Button block.
* Adds new action "Set as URL" in the Personalization Tags modal for the Button block.
* Implements a workaround to allow adding or editing personalization tags manually from the link control component without prefixing them with `http://`.
* Adds support for personalization tags in `href` attributes for `<a>` tags.

Closes [WOOPLUG-3654: Personalization Tags does not support buttons](https://linear.app/a8c/issue/WOOPLUG-3654/personalization-tags-does-not-support-buttons) .

| Screenshots | |
| ----- | ----- |
| AZOAZ002kwiDY4jB | Sk0k0dIMHcFHLztn |

1. Enable the block-based email editor from WooCommerce → Settings → Advanced → Features → **Block Email Editor (alpha)**.
2. Open an email in the editor WooCommerce → Settings → **Emails**.
3. Add a Button block and open the Personalization Tags modal from the toolbar.
4. Confirm the "Insert" inserts the tag correctly as a button text.
5. Confirm the "Set as URL" action, available on URL related personalization tags, sets the URL for the button correctly.
6. Confirm editing the personalization tag manually from the link control doesn't prefix the tag with `http://`.
7. Preview the email and confirm the personalization tags used as button text or URL are replaced with the expected values.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

Allow using Personalization Tags as button text or URL

</details>

### DIFF
--- a/packages/js/email-editor/src/components/personalization-tags/category-section.tsx
+++ b/packages/js/email-editor/src/components/personalization-tags/category-section.tsx
@@ -3,6 +3,8 @@
  */
 import { Button } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
+import { useDispatch, useSelect } from '@wordpress/data';
+import { store as blockEditorStore } from '@wordpress/block-editor';
 
 /**
  * Internal dependencies
@@ -24,6 +26,16 @@ const CategorySection = ( {
 	closeCallback: () => void;
 	openLinkModal: ( tag: PersonalizationTag ) => void;
 } ) => {
+	const { updateBlockAttributes } = useDispatch( blockEditorStore );
+	const selectedBlockId = useSelect( ( select ) =>
+		select( blockEditorStore ).getSelectedBlockClientId()
+	);
+	const selectedBlock = useSelect( ( select ) =>
+		// @ts-expect-error getBlock expects one argument, but TS thinks it expects none
+		select( blockEditorStore ).getBlock( selectedBlockId )
+	);
+	const canSetURL = [ 'core/button' ].includes( selectedBlock?.name );
+
 	const categoriesToRender: [ string, PersonalizationTag[] ][] =
 		activeCategory === null
 			? Object.entries( groupedTags ) // Render all categories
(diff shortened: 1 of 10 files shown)


packages/js/email-editor/src/components/personalization-tags/category-section.tsx
```diff
@@ -24,6 +26,16 @@ const CategorySection = ( {
 	closeCallback: () => void;
 	openLinkModal: ( tag: PersonalizationTag ) => void;
 } ) => {
+	const { updateBlockAttributes } = useDispatch( blockEditorStore );
+	const selectedBlockId = useSelect( ( select ) =>
+		select( blockEditorStore ).getSelectedBlockClientId()
+	);
+	const selectedBlock = useSelect( ( select ) =>
+		// @ts-expect-error getBlock expects one argument, but TS thinks it expects none
```
pavel-mailpoet: Do you know why TS thinks that? Would it be possible to fix this globally, rather than adding a TS ignore?

packages/php/email-editor/src/Engine/class-personalizer.php
```diff
@@ -122,6 +122,25 @@ public function personalize_content( string $content ): string {
 					$content_processor->remove_attribute( 'data-link-href' );
 					$content_processor->remove_attribute( 'contenteditable' );
 				}
+			} elseif ( $content_processor->get_token_type() === '#tag' && $content_processor->get_tag() === 'A' ) {
+				$href = $content_processor->get_attribute( 'href' );
+
+				if ( ! $href || ! preg_match( '/\[[a-z-\/]+\]/', urldecode( $href ), $matches ) ) {
```
pavel-mailpoet, line 128: I am not sure about the syntax. We used the HTML comments, but here we are using only the syntax `[]`. Should we be more consistent and use the html comments here as well?

packages/js/email-editor/src/components/personalization-tags/category-section.tsx
```diff
@@ -52,7 +52,7 @@ const CategorySection = ( {
 						<div className="woocommerce-personalization-tags-modal-category-group">
 							{ items.map( ( item ) => {
 								// TODO: Improve logic for detecting URL tags.
```
pavel-mailpoet: Do we still need this here? If it could be deleted, it would be better, but if not, I think we need more details.

packages/js/email-editor/src/components/personalization-tags/category-section.tsx
```diff
@@ -52,7 +52,7 @@ const CategorySection = ( {
...
-								const isURLTag = /\Wurl\W/.test( item.token );
```
pavel-mailpoet: The commit message doesn't explain this change at all. What was the reason we needed this change? In what scenario is this necessary?

pavel-mailpoet (commented):
I have a couple of comments

pavel-mailpoet:
The functionality is not working properly. If I try to use it, nothing happens and I get an error in the console:

Screenshot 2025-05-14 at 15 46 38

packages/php/email-editor/src/Engine/class-personalizer.php
yuliyan: On discussing the ticket, we decided to use the syntax without the comment part in this specific case, because it looks less confusing when displayed in the link control, and it's similar to the shortcode syntax that users might be more familiar with.

packages/js/email-editor/src/components/personalization-tags/category-section.tsx
yuliyan: I added the comment because I'm not entirely satisfied with this implementation. It works all right, but I assumed that at some point we may need to introduce better categorization for the personalization tags so that we can better filter filter the actions (e.g. "Set as URL") based on the context.

packages/js/email-editor/src/components/personalization-tags/category-section.tsx
yuliyan: The pattern was introduced in earlier commit of this PR, and was changed based on Copilot suggestion. I thought that latter regex made a little more sense since `\b` is non-capturing token. However, I don't think it makes any difference in any scenario since it's currently used just to `.test()`, rather using the matched string.

packages/js/email-editor/src/components/personalization-tags/category-section.tsx
yuliyan: I came across this "fix" used elsewhere in the package and followed it, assuming the problem was in the original selector definition. On a second look after your comment, I found that we have type definitions for the `@wordpress ` modules in the email editor package, that are probably causing the conflict. I'll investigate further and try to fix it.

yuliyan:
> The functionality is not working properly. If I try to use it, nothing happens and I get an error in the console:

@pavel-mailpoet, I was unable to reproduce the error testing the branch on a refreshed local environment. Do you always get the error when attempting to insert the tag, or only in a specific scenario? Also, do you get an error if you try to insert in a paragraph?

packages/js/email-editor/src/components/personalization-tags/category-section.tsx
pavel-mailpoet: I understand your reasoning, but I don't find the comment very helpful. It doesn't explain it. Based on my previous experience, TODO comments often remain in the codebase for years without anyone taking action on them. Anyway, I am OK with keeping it here, but I would try to explain better what exactly "Improve logic" means. Because without any context, it is hard to understand.

packages/js/email-editor/src/components/personalization-tags/category-section.tsx
pavel-mailpoet: I understand; I'm not a big fan of those comments. I'm okay with you keeping it here if you just add another instance of it, but we should probably make an effort to clean them up at some point.

pavel-mailpoet:
> I was unable to reproduce the error testing the branch on a refreshed local environment. Do you always get the error when attempting to insert the tag, or only in a specific scenario? Also, do you get an error if you try to insert in a paragraph?

Never mind, it is working now, maybe it was an issue with my environment:

Screenshot 2025-05-15 at 9 20 33

Screenshot 2025-05-15 at 9 20 37

pavel-mailpoet (approved):

packages/js/email-editor/src/components/personalization-tags/category-section.tsx
yuliyan: I did some digging and found that we override the module declaration for `@wordpress/data` and use `CurriedSelectorsOf` to transform the selector definitions. `CurriedSelectorsOf` is used to remove the first argument from the selector, which is expected to be `state`. However, in our case, we use the selectors types from `@types/wordpress__block-editor` which already have the state argument removed, and thus `CurriedSelectorsOf` removes another argument, which breaks the types.

Fixed it in https://github.com/woocommerce/woocommerce/pull/57958/commits/ed5d16536bd3da3cbb9b1667586d6307febeb8dc.

packages/js/email-editor/src/components/personalization-tags/category-section.tsx
yuliyan: Updated the comment to be more descriptive and removed that "TODO" part as it is indeed not very useful.

PR #53526, opened 2024-12-06
Make nullable parameters explicitly nullable in various admin packages

jorgeatorres:
### Submission Review Guidelines:

-   I have followed the [WooCommerce Contributing Guidelines](https://github.com/woocommerce/woocommerce/blob/trunk/.github/CONTRIBUTING.md) and the [WordPress Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/).
-   I have checked to ensure there aren't other open [Pull Requests](https://github.com/woocommerce/woocommerce/pulls) for the same update/change.
-   I have reviewed my code for [security best practices](https://developer.wordpress.org/apis/security/).
-   Following the above guidelines will result in quick merges and clear and detailed feedback when appropriate.

### Changes proposed in this Pull Request:

Beginning with PHP 8.4, implicitly nullable parameters in function signatures (such as `function f( int $x = null )`) [will trigger a deprecation notice](https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated). Those parameters now have to be explicitly made nullable (i.e. ` function f( ?int $x = null )`).

To reduce the amount of noise when running under this PHP version, this PR makes the relevant changes to code related to various admin functionality (particularly, marketing related stuff).

In particular, it fixes deprecation notices such as the following:

```
PHP Deprecated:  Automattic\WooCommerce\Admin\PluginsHelper::install_plugins(): Implicitly marking parameter $logger as nullable is deprecated, the explicit nullable type must be used instead in [...]/src/Admin/PluginsHelper.php on line 220
PHP Deprecated:  Automattic\WooCommerce\Admin\PluginsHelper::activate_plugins(): Implicitly marking parameter $logger as nullable is deprecated, the explicit nullable type must be used instead in [...]/src/Admin/PluginsHelper.php on line 42
PHP Deprecated:  Automattic\WooCommerce\Admin\Features\PaymentGatewaySuggestions\Init::get_suggestions(): Implicitly marking parameter $specs as nullable is deprecated, the explicit nullable type must be used instead in [...]/src/Admin/Features/PaymentGatewaySuggestions/Init.php on line 36
PHP Deprecated:  Automattic\WooCommerce\Admin\Features\MarketingRecommendations\Init::evaluate_specs(): Implicitly marking parameter $specs as nullable is deprecated, the explicit nullable type must be used instead in [...]/src/Admin/Features/MarketingRecommendations/Init.php on line 85
```

[...]

PR #52296, opened 2024-10-24
Introduce a simplified dependency injection container

Konamiman:
### Changes proposed in this Pull Request:

This pull request introduces a simplified dependency injection container as a replacement of the PHP League's Container package. The motivation for this change is the performance gain, and the fact that we are effectively only using the container for one task: resolving single-instance classes by class name (we don't need a full-blown dependency injection engine for that).

This pull request introduces two new classes:

* `RuntimeContainer`: this class is a replacement for `ExtendedContainer` as the underlying dependency injection container used by WooCommerce at runtime. It doesn't implement explicit class registration, instead it considers any class belonging to the `Automattic\WooCommerce` namespace as implicitly registered as a single-instance class. It uses reflection to resolve (recursively) the dependencies passed as parameters to the `init` method of the resolved classes.
* `TestingContainer`: inheriting from the previous one, includes additional methods to replace the regular classes with mocks and it's intended for usage when running unit tests (being, too, a replacement for `ExtendedContainer`).

`RuntimeContainer` is aware of the containers that the Store API and Blocks implement on their own, and will delegate class resolutions to these containers for classes within the respective namespaces.

Temporarily it's possible to force WooCommerce to still use `ExtendedContainer` at runtime by either creating a `WOOCOMMERCE_USE_OLD_DI_CONTAINER` constant with a value of `true`, or hooking on the `woocommerce_use_old_di_container` filter and having it returning true. This should be done only if issues are found in the new container and the site is broken as a result. `ExtendedContainer` (along with the PHP Lague's container package and all the supporting code) will be removed in WooCommerce 10.0.

Smilarly, it's possible to force WooCommerce to still use `ExtendedContainer` for unit tests by defining the `USE_OLD_DI_CONTAINER` environment variable before triggering the tests run.

### How to test the changes in this Pull Request:

1. Verify that the WooCommerce critical flows (like placing an order from the store and managing products and orders from admin) work with and without adding `define('WOOCOMMERCE_USE_OLD_DI_CONTAINER', true);` in `wp-config.php` (or alternatively, with and without adding this snippet: `add_filter('woocommerce_use_old_di_container', '__return_true');`
2. Verify that all the unit tests pass.
3. Temporarily add the following method to the `Container` class (`src/Container.php`), then run `wp eval 'wc_get_container()->instantiate_everything();'`, again with both the old and the new containers (see step 1):

```php
// Use the current container to retrieve an instance of each of the classes that are registered with the old container.
public function instantiate_everything() {
[...]

PR #57966, opened 2025-05-13
Remove the old dependency injection container and all the related infrastructure

Konamiman:
In https://github.com/woocommerce/woocommerce/pull/52296 we introduced a simplified dependency injection container and we left the old one in place ready to be reactivated with a filter just in case, but [we announced](https://developer.woocommerce.com/2024/11/15/developer-advisory-changes-to-the-dependency-injection-container-in-woocommerce/) that we would definitely remove the old container and the related infrastructure in WooCommerce 10.0. This pull request does just that:

- Removes the PHP League's Container package from `lib`.
- Deletes the `ExtendedContainer` class.
- Deletes all the service providers and the `AbstractServiceProvider` class.
- Modifies the `Container` class so that it always instantiates a `RuntimeContainer`.
- Removes the `should_use` method from `RuntimeContainer`.
- Deletes the unit testing support classes related to `ExtendedContainer`.
- Removes the code from `bootstrap.php` that allowed using `ExtendedContainer` in unit tests.
- Updates all the relevant documentation (class/function docblocks and README files).

The `psr/container` package is kept even if no longer used to prevent backward compatibility issues (it's a very small package anyway).

Additionally, `RuntimeContainer` is improved to throw a proper exception (instead of a fatal error) when trying to resolve a class that doesn't have a public constructor or has a public constructor with non-optional arguments.

Important notes:

1. Now _"psr/container"_ MUST be the first entry in the _"extra - mozart - packages"_ in `lib/composer.json`. Otherwise
that package won't get the `Automattic\WooCommerce\Vendor` namespace when `composer run-script build-lib` runs.

2. https://github.com/woocommerce/woocommerce/pull/53526#issuecomment-2875522696 - the `lib/README.md` file has been updated to warn about this.

1. Basically smoketesting. Verify that the WooCommerce critical flows (like placing an order from the store and managing products and orders from admin) still work.
2. Review the updated documentation for accuracy and clarity.
3. Add the following code to the `Container` class (`src/Container.php`), then run `wp eval 'wc_get_container()->instantiate_everything();'`. You will see that each class is either instantiated, or an exception is thrown, but there are no fatal errors:

```php
public function instantiate_everything() {
	$classes = $this->getAllClassesFromDirectory( __DIR__ );

	foreach ( $classes as $class ) {
		if ( ! class_exists( $class ) ) {
			continue;
		}
		try {
			echo 'Ok: ' . get_class( $this->get( $class ) ) . "\n";
		} catch ( \Exception $e ) {
			echo '*** ' . $class . ': ' . $e->getMessage() . "\n";
		}
	}
}

private function getAllClassesFromDirectory( $directory ) {
	$classes = array();

	$iterator = new \RecursiveIteratorIterator(
		new \RecursiveDirectoryIterator( $directory )
	);

	foreach ( $iterator as $file ) {
		if ( $file->getExtension() !== 'php' ) {
			continue;
		}

		$className = $this->getFullClassNameFromFile( $file->getPathname() );
		if ( $className !== null ) {
			$classes[] = $className;
		}
	}

	return $classes;
}

private function getFullClassNameFromFile( $filePath ) {
	$content = file_get_contents( $filePath );

	$namespace = null;
	if ( preg_match( '/^\s*namespace\s+([^;]+);/m', $content, $matches ) ) {
		$namespace = $matches[1];
	}

	if ( preg_match( '/^\s*class\s+([a-zA-Z0-9_]+)/m', $content, $matches ) ) {
		$className = $matches[1];
		return $namespace ? $namespace . '\\' . $className : $className;
	}

	return null;
}
```

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


adimoldovan:
@Konamiman can you check the PHP test failures?

Konamiman:
@adimoldovan There was actually a bug in the extra checks added to `RuntimeContainer`: it was not possible to resolve a class with optional constructor arguments. It's fixed now.

adimoldovan (commented):
> Add the following code to the Container class (src/Container.php), then run wp eval 'wc_get_container()->instantiate_everything();'. You will see that each class is either instantiated, or an exception is thrown, but there are no fatal errors:

@Konamiman I did get a fatal here:

```
PHP Fatal error:  Uncaught Error: Class 'Theme_Upgrader' not found in /var/www/html/wp-content/plugins/woocommerce/src/Admin/Overrides/ThemeUpgrader.php:13
Stack trace:
#0 /var/www/html/wp-content/plugins/woocommerce/vendor/jetpack-autoloader/class-php-autoloader.php(102): require()
#1 [internal function]: Automattic\Jetpack\Autoloader\jp67abb2f5c15b32a13611c55273717232\al5_0_0\PHP_Autoloader::load_class('Automattic\\WooC...')
#2 [internal function]: spl_autoload_call('Automattic\\WooC...')
#3 /var/www/html/wp-content/plugins/woocommerce/src/Container.php(85): class_exists('Automattic\\WooC...')
#4 phar:///usr/local/bin/wp/vendor/wp-cli/eval-command/src/Eval_Command.php(39) : eval()'d code(1): Automattic\WooCommerce\Container->instantiate_everything()
#5 phar:///usr/local/bin/wp/vendor/wp-cli/eval-command/src/Eval_Command.php(39): eval()
#6 [internal function]: Eval_Command->__invoke(Array, Array)
#7 phar:///usr/local/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Dispatcher/CommandFactory.php(100): call_user_func(Array, Array, Array) in /var/www/html/wp-content/plugins/woocommerce/src/Admin/Overrides/ThemeUpgrader.php on line 13
```

Konamiman:
@adimoldovan Before instantiating this particular class the following is needed first:

```php
include_once ABSPATH . '/wp-admin/includes/class-wp-upgrader.php';
include_once ABSPATH . '/wp-admin/includes/class-theme-upgrader.php';
```

That's because the class is declared as `class ThemeUpgrader extends \Theme_Upgrader` but that `\Theme_Upgrader` (provided by WordPress core) isn't included by default. It's not the fault of the container, though; a simple `new ThemeUpgrader()` fails in the same way without those includes.

adimoldovan (approved):
✅ Smoke tested, no issues found. e2e tests passed
✅ Doc updates make sense to me
✅ `wc_get_container()->instantiate_everything();` output as expected

PR #57979, opened 2025-05-14
[Experimental] Blockified Add to Cart with Option: Improved layout of template part

MaggieCabrera:

While working on #57927 I noticed how strange the layout of the template part looks when we have the checkbox available, so I decided to align the first column to the left and the last one to the right. We don't have much control to style the checkbox case specifically but I think this is a better default than what we started with. I would love to hear design input on this

**Site Editor**
| Before | After |
| ------ | ----- |
|    Screenshot 2025-05-14 at 09 51 27    |    Screenshot 2025-05-14 at 09 46 00   |

**Frontend**
| Before | After |
| ------ | ----- |
|    Screenshot 2025-05-14 at 09 51 36    |    Screenshot 2025-05-14 at 09 45 51    |    

1. Create a grouped product. I use the default woocommerce export dummy products and the Logo Collection product.
2. I changed the Beanie to be set as sold individually under inventory
3. I added the V-Neck T-Shirt to the Logo Collection group of products (this is a variable product). This gets me all 3 cases to test
4. On the single product template, update the Add to Cart block to use the blockified version
5. Check the new styles both on the site editor and the frontend

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
[Experimental] Blockified Add to Cart with Option: Improved layout of template part
</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/57979-try-style-grouped-block-template
+++ b/plugins/woocommerce/changelog/57979-try-style-grouped-block-template
@@ -1,0 +1,4 @@
+Significance: minor
+Type: enhancement
+
+[Experimental] Blockified Add to Cart with Option: Improved layout of template part
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-cta/style.scss
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-cta/style.scss
@@ -1,8 +1,6 @@
 @import "../../../../base/components/quantity-selector/style";
 
 .wc-block-add-to-cart-with-options-grouped-product-selector-item-cta {
-	text-align: center;
-
 	.button {
 		display: block;
 	}
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-cta/style.scss
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-cta/style.scss
@@ -2,6 +2,6 @@
 
 .wc-block-add-to-cart-with-options-grouped-product-selector-item-cta {
 	.button {
-		display: block;
+		display: inline-flex;
 	}
 }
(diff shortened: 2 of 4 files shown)


MaggieCabrera:
I went with a wrapping group block for this because the CTA block doesn't have alignment options, but we could add support to that and remove the group block instead

gigitux:
Code-wise the PR LGTM! 

> I went with a wrapping group block for this because the CTA block doesn't have alignment options, but we could add support to that and remove the group block instead

If it is possible, I would follow this approach to make the block structure simpler.

Aljullu:
I agree with Luigi that it would be great to add alignment options to the _Grouped Product: Item Selector_ block. As an interim solution, however, we could simply remove this CSS line:

https://github.com/woocommerce/woocommerce/blob/78632db69e88c0d7aef4c4ac6b88e35ec334fa86/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/product-item-cta/style.scss#L2

@MaggieCabrera would you like us to take ownership of this PR?

MaggieCabrera:
> @MaggieCabrera would you like us to take ownership of this PR?

No need to ask for permission on any of my PRs! specially if I'm AFK

Aljullu:
> I went with a wrapping group block for this because the CTA block doesn't have alignment options, but we could add support to that and remove the group block instead

I think that's not as straightforward as we initially thought because the block is inside a _Grid_ block, so even if we add alignment support, the controls won't show up.

I went for the alternative solution to remove the `text-alignment` CSS rule in 2ec0620222c264dfc4fe177a929299eae41f1e94.

@gigitux can you take another look? :pray:

gigitux:
I'm happy to approve this PR, but do you think that it makes sense that the button size is the entire column? 


It looks pretty misaligned:

image


Unfortunately, I played with `Grid` block, and it looks like that we can't fix it via Global Styles.


NIT: Should we keep the order of the elements in the same order?

| Editor | Frontend |
|--------|--------|
| image |image |

Aljullu:
Good catch, @gigitux! That's because we had a `display: block` in the CSS that made them take all the available space. I replaced it with `inline-flex` (to keep it consistent with the Product Button block), and it looks better now:

image

(I feel font sizes are still not properly balanced, but I opened a separate issue for it: #58465 / WOOPLUG-4480).

> NIT: Should we keep the order of the elements in the same order?

Hmm, that's weird, they already are in my case:

Editor | Frontend
--- | ---
image | image

I will keep an eye in case I can reproduce. Or if you have steps to reproduce, let me know! :pray:

gigitux (approved):
LGTM! 🚀 

| Editor | Frontend |
|--------|--------|
| image | image | 


> I will keep an eye in case I can reproduce. Or if you have steps to reproduce, let me know! 🙏

Thanks! I don't have particular steps to reproduce it.

PR #57989, opened 2025-05-14
Add shipment information box on fulfillment editor

### LINKED ISSUE #57405: [Order fulfillments] Create shipping information entry form
**Issue:** We need a form to allow our users to add tracking information manually for a fulfillment.

**Proposed solution:** Create the form using the design in [https://www.figma.com/design/2xydCrrB4yb2TAqpXt607y/New-Shipping-Fulfillments-Entity?node-id=3274-51696&p=f&t=Hk9hU6M2KNCPTs23-0](https://www.figma.com/design/2xydCrrB4yb2TAqpXt607y/New-Shipping-Fulfillments-Entity?node-id=3274-51696&p=f&t=Hk9hU6M2KNCPTs23-0) and fill the providers list using the taxonomy terms and data returned from the filter added in [WOOPLUG-3999](https://linear.app/a8c/issue/WOOPLUG-3999/[order-fulfillments]-add-shipping-providers-support). When the fulfillment is saved as draft, or fulfilled, with the form being filled, we should save the inputs to the relative metadata.

tpaksu:

This PR adds the shipping information entry and display boxes, and their tests to the fulfillment drawer. 

- The user can add this information while creating the fulfillment, or skip it (choose no information provided) on the creation, and add it with updating the fulfillment. 
- There are some shipping providers implemented as JS object for demo purposes, normally it will be received via a global variable like `wcSettings` from the PHP side. This will be handled on a separate issue alongside the currency symbols, the fulfillment status strings, the meta key strings, etc.
- I also added a demo lookup response, which shows up when you enter "12345678" to the tracking number look up form, this will be deleted before releasing the feature.

Closes WOOPLUG-4000

image
image
image
image
image
image

- Checkout this branch
- Go to `plugins/woocommerce` folder, and run `pnpm run watch:build` to build the assets.
- Create some orders if you don't have any in your test environment.
- Open WooCommerce > Orders page.

#### Tracking number lookup with a new fulfillment

- Test the tracking number lookup with no tracking number entered. You should see an error.
- Test the tracking number lookup with a random tracking number. You should see the same error.
- Test the tracking number lookup with "12345678", you should see a shipment provider, and a tracking link returned. Also the tracking number input will convert to a label with an edit icon.
- Note the shipment provider and tracking URL and save the fulfillment. Open the saved fulfillment, and you should see the same information displayed.
- Edit the fulfillment, and you should see the same information on the shipping information edit form.

#### Manual entry with a new fulfillment

> [!Note] 
> Those fields are not validated, and we allow any field to be empty on save. The user will have the control over that, and it's their responsibility to have them filled. We don't know the possible scenarios the user might have.

- Test the manual entry with a tracking number, a shipment provider from the list, and a tracking link. 
- Store the fulfillment and see if they are saved correctly.
- Edit the fulfillment, and see if the form opens with manual entry option selected, and the inputs have correct values.
- Test it again with selecting "Other" at the bottom of the shipment provider list. There should be a new input visible allowing you to enter a name for the shipping provider.
- Fill it with a provider that isn't available on the list, fill other fields and save.
- Check if the fulfillment displays the correct values
- Check if editing the fulfillment shows the correct values

#### No information entered

- Test saving and editing a new fulfillment with "No information" radio selected. In all steps, it should show the correct values.

#### Testing edits

For all scenarios above, create a new fulfillment with "No information provided", selected, and test the scenario applied with editing the fulfillment. It should show the correct values after editing and saving the fulfillment. 

#### Check design

Open this 2xydCrrB4yb2TAqpXt607y-fi-3939_19728 and check if the designs match.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

Adds a shipment information editor for the fulfillments.

</details>


plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/components/shipment-form/shipment-manual-entry-form.tsx
```diff
@@ -0,0 +1,139 @@
...
+						__nextHasNoMarginBottom
+					/>
+					<div className="woocommerce-fulfillment-shipment-provider-search-icon">
+						<SearchIcon />
+					</div>
+				</div>
+			</div>
+			{ shipmentProvider === 'other' && (
+				<div className="woocommerce-fulfillment-input-container">
+					<h4>{ __( 'Name', 'woocommerce' ) }</h4>
```
ismaeldcom: Why using `h4` instead of the `label` property from `TextControl`?

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/utils/icons.tsx
```diff
@@ -0,0 +1,108 @@
...
+		<path
+			fillRule="evenodd"
+			clipRule="evenodd"
+			d="M0 2C0 0.89543 0.895431 0 2 0H16C17.1046 0 18 0.895431 18 2V12C18 13.1046 17.1046 14 16 14H2C0.89543 14 0 13.1046 0 12V2ZM2 1.5H16C16.2761 1.5 16.5 1.72386 16.5 2V2.93754L9.00005 8.5625L1.5 2.93746V2C1.5 1.72386 1.72386 1.5 2 1.5ZM1.5 4.81246V12C1.5 12.2761 1.72386 12.5 2 12.5H16C16.2761 12.5 16.5 12.2761 16.5 12V4.81254L9.00005 10.4375L1.5 4.81246Z"
+			fill="#1E1E1E"
+		/>
+	</svg>
+);
+
+export const CopyIcon = ( { copyText }: { copyText: string } ) => {
```
ismaeldcom, line 70: We should consider using a button here to ensure accessibility.

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/components/user-interface/fulfillments-card/card.tsx
ismaeldcom, line 1: Did you evaluate using the existing [Panel component](https://wordpress.github.io/gutenberg/?path=/docs/components-panel--docs)?

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/context/shipment-form-context.tsx
```diff
@@ -0,0 +1,123 @@
...
+		defaultContextProps.shipmentProvider
+	);
+	const [ trackingUrl, setTrackingUrl ] = React.useState(
+		defaultContextProps.trackingUrl
+	);
+	const [ providerName, setProviderName ] = React.useState(
+		defaultContextProps.providerName
+	);
+
+	// fulfillment değiştiğinde state'leri güncelle
```
ismaeldcom: This needs translation for those who don't understand Turkish. 🙂

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/context/shipment-form-context.tsx
```diff
@@ -0,0 +1,123 @@
...
+			getFulfillmentMeta( fulfillment, TRACKING_URL_META_KEY, '' )
+		);
+		setProviderName(
+			getFulfillmentMeta( fulfillment, PROVIDER_NAME_META_KEY, '' )
+		);
+	}, [ fulfillment ] );
+
+	return (
+		<ShipmentFormContextValue.Provider
+			value={ {
```
ismaeldcom: Both this and the fulfillment context will benefit from `useMemo` here. Maybe is not an issue right now, but any component consuming this provider will rerender every time any these properties change even if they are not using them.

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/utils/icons.tsx
```diff
@@ -0,0 +1,108 @@
...
+	<svg
+		width="16"
+		height="16"
+		viewBox="0 0 12 14"
+		fill="none"
+		xmlns="http://www.w3.org/2000/svg"
+	>
+		<path
+			d="M11.8333 2.83301L9.33329 0.333008L2.24996 7.41634L1.41663 10.7497L4.74996 9.91634L11.8333 2.83301ZM5.99996 12.4163H0.166626V13.6663H5.99996V12.4163Z"
+			fill="#3858E9"
```
ismaeldcom: Remember to remove the `fill` here if you need to change it's color from CSS.

ismaeldcom (commented):
Thanks for splitting the PR to make it easier to review. The testing instructions work as intented but I've noticed some issues, and although I'm not sure if all of them fall into this PR I'll share them here.

#### UX issues
- In an order with a product x2, create a fulfillment for just one of them. Editing the fullfilment later will select the two instead of just one.
- After opening the fulfillment modal for a second order, the products' checkbox doesn't work, I need to first click on `Select all` to be able to use them again.
- If you switch to a different order fulfillment while editing one, you end up with the sidebar in edit state (header disabled)

#### Design review
Not sure if you prefer to ask for a review from Design, but here is what I've noticed by comparing the current implementation with the Figma one.

- There are no items selected by default.
- The `Select all(n)` button should display `N Selected` instead, in any case.
- After clicking on **Find info** in **Tracking Number**, provider should display the icon and name, and tracking URL should by an anchor. Right now these are disabled inputs.
- The edit (pen) button should be grey instead of blue.
- Fulfillment status should be capitalized.
- While editing a fulfillment, other fulfillments should be disabled but not the header.

It's not clear to me from the designs, but I think we should unwrap by default a fulfillment if there are no products left. Otherwise you only see a collapsed **Fulfillment 1** on the sidebar.

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/context/shipment-form-context.tsx
tpaksu: 🤦‍♂️ Good catch 😬

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/utils/icons.tsx
tpaksu: I'll leave this to later design review and change if needed.

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/context/shipment-form-context.tsx
tpaksu: Addressed in [71d7b63](https://github.com/woocommerce/woocommerce/pull/57989/commits/71d7b6384a44cf3d74227aa498a388d288df0bcf).

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/context/shipment-form-context.tsx
tpaksu: Fixed in [71d7b63](https://github.com/woocommerce/woocommerce/pull/57989/commits/71d7b6384a44cf3d74227aa498a388d288df0bcf).

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/utils/icons.tsx
tpaksu: Addressed in [71d7b63](https://github.com/woocommerce/woocommerce/pull/57989/commits/71d7b6384a44cf3d74227aa498a388d288df0bcf).

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/components/user-interface/fulfillments-card/card.tsx
tpaksu: Yep, I've tried that before, just tried that yesterday again to see what the issue was, and confirmed that the panel component doesn't support disabled state (need to disable other tabs when editing a fulfillment), and custom components on the title (we need to display the fulfillment status badge on the title). Once that features are added, we can replace those with the existing panel components.

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/components/shipment-form/shipment-manual-entry-form.tsx
tpaksu: Good catch, addressed in [71d7b63](https://github.com/woocommerce/woocommerce/pull/57989/commits/71d7b6384a44cf3d74227aa498a388d288df0bcf).

tpaksu:
> In an order with a product x2, create a fulfillment for just one of them. Editing the fullfilment later will select the two instead of just one.

Thanks for noticing, fixed it on the parent PR. 

> After opening the fulfillment modal for a second order, the products' checkbox doesn't work, I need to first click on Select all to be able to use them again.

Yep, that was also an issue which surfaced on the parent PR, fixed it there.

> If you switch to a different order fulfillment while editing one, you end up with the sidebar in edit state (header disabled)

I'm not sure how to fix this one, maybe I should check if the _editMode_ is active and prevent the order ID change, but the editMode is in the inner context. We should do a brainstorm on this.

> Not sure if you prefer to ask for a review from Design, but here is what I've noticed by comparing the current implementation with the Figma one.

I'll ask for a design review once all the components and flow are completed, and create another PR for fixing all the issues. 

> There are no items selected by default.

Thanks for noticing. I've rewritten the logic for the items tree, which should fix it in the parent PR. 

> The Select all(n) button should display N Selected instead, in any case.

Can you check the latest update if it works as you mentioned?

> After clicking on Find info in Tracking Number, provider should display the icon and name, and tracking URL should by an anchor. Right now these are disabled inputs.
> The edit (pen) button should be grey instead of blue.

Yep, fixed them in this PR. 

> Fulfillment status should be capitalized.

This is waiting for getting the translated string representations from the backend, these shown are only keys for now.

> While editing a fulfillment, other fulfillments should be disabled but not the header.

Fixed it. 

> It's not clear to me from the designs, but I think we should unwrap by default a fulfillment if there are no products left. Otherwise you only see a collapsed Fulfillment 1 on the sidebar.

What do you mean by "unwrap"? Showing all the fulfillments as expanded? **Edit:** Yep, it is visible to me too, but as you mentioned, I don't know how to proceed. I'll also ask this one on the design review.

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/components/user-interface/fulfillments-card/card.tsx
ismaeldcom: Thanks for checking it!

ismaeldcom (approved):
Thanks for taking care of every comment! Everything is fixed and works as expected. I noticed a small issue, but it's not related to this PR; more on that below.

>I'm not sure how to fix this one, maybe I should check if the editMode is active and prevent the order ID change, but the editMode is in the inner context. We should do a brainstorm on this.

The easiest way to fix it is adding a `key` to [the provider here](https://github.com/woocommerce/woocommerce/blob/0ab1940d042bfe47cddaac1935e2df6061c9d030/plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/components/user-interface/fulfillment-drawer/fulfillment-drawer.tsx#L36). Using `orderId` as the key will create a different provider per order, instead of reusing the same one and preserving its state.

While looking at `FulfillmentDrawerProvider`, I've noticed that you're using `setState` within `useSelect` while having these states in the dependencies, which causes an infinite loop that you prevented using `isEqual`.

We should return the store date, like this:
```tsx
const { order, fulfillments} = useSelect(
  ( select ) => {
    const store = select( FulfillmentsStore );
    return {
      order: store.getOrder( orderId ),
      fulfillments: store.readFulfillments( orderId ) ?? [],
    };
  },
  [ orderId ]
);
```
If you move the `orderId` check outside of the provider, you can greatly simplify it. If I'm not wrong, order and fulfillment states were only used within the provider, so they don't need to be exposed and can be removed.

PR #58001, opened 2025-05-14
Prevent the possibility of rendering multiple instances of checkout on the page

### LINKED ISSUE #57791: Safe guard Block Checkout from creating multiple concurrent requests when odd/corrupted block code is present
### Prerequisites

### Describe the bug

In 9566861-zen we noted mysteriously a single order is charged multiple times via WooPayments, there are also intermittent 400,500 and 429 request responses seen on `wp-json/wc/store/v1/checkout?_locale=site` end point.

Upon investigating I found the checkout page is actually generating multiple `wp-json/wc/store/v1/checkout?_locale=site` requests on pressing Place Order button

<img width="1246" alt="Image" src="https://github.com/user-attachments/assets/61127c48-b7a7-4ff4-a5af-dc5bb1febd45" />

This was because the usual block checkout code was affected/corrupted due to external factors

```
<!-- wp:columns {"wpmlTranslatedContent":"1"} -->
<div class="wp-block-columns"><!-- wp:column {"width":"16.5%","wpmlTranslatedContent":"1"} -->
<div class="wp-block-column" style="flex-basis:16.5%"></div>
<!-- /wp:column -->

<!-- wp:column {"width":"67%","wpmlTranslatedContent":"1"} -->
<div class="wp-block-column" style="flex-basis:67%"><!-- wp:heading {"level":1,"dynamicAttributes":{"toolsetDSVersion":"257000"},"translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<h1 class="wp-block-heading" id="h-checkout">Kasse</h1>
<!-- /wp:heading --></div>
<!-- /wp:column -->

<!-- wp:column {"width":"16.5%","wpmlTranslatedContent":"1"} -->
<div class="wp-block-column" style="flex-basis:16.5%"></div>
<!-- /wp:column --></div>
<!-- /wp:columns --><!-- wp:columns {"wpmlTranslatedContent":"1"} -->
<div class="wp-block-columns"><!-- wp:column {"width":"16.5%","wpmlTranslatedContent":"1"} -->
<div class="wp-block-column" style="flex-basis:16.5%"></div>
<!-- /wp:column -->

<!-- wp:column {"width":"67%","wpmlTranslatedContent":"1"} -->
<div class="wp-block-column" style="flex-basis:67%"><!-- wp:woocommerce/checkout {"className":"wc-block-checkout","translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout alignwide wc-block-checkout is-loading"><!-- wp:woocommerce/checkout-fields-block {"translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout-fields-block"><!-- wp:woocommerce/checkout-express-payment-block {"showButtonStyles":true,"buttonBorderRadius":"5","translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout alignwide wc-block-checkout is-loading"><br><br></div><br>
<!-- /wp:woocommerce/checkout-express-payment-block -->

<!-- wp:woocommerce/checkout-contact-information-block {"wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout-contact-information-block"><!-- wp:woocommerce/mailchimp-newsletter-subscription {"gdprStatus":"uncheck","translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout alignwide wc-block-checkout is-loading"><br><br></div><br>
<!-- /wp:woocommerce/mailchimp-newsletter-subscription --></div>
<!-- /wp:woocommerce/checkout-contact-information-block -->

<!-- wp:woocommerce/checkout-shipping-method-block {"translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout alignwide wc-block-checkout is-loading"><br><br></div><br>
<!-- /wp:woocommerce/checkout-shipping-method-block -->

<!-- wp:woocommerce/checkout-pickup-options-block {"translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout alignwide wc-block-checkout is-loading"><br><br></div><br>
<!-- /wp:woocommerce/checkout-pickup-options-block -->

<!-- wp:woocommerce/checkout-shipping-address-block {"wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout-shipping-address-block"></div>
<!-- /wp:woocommerce/checkout-shipping-address-block -->

<!-- wp:woocommerce/checkout-billing-address-block {"translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout alignwide wc-block-checkout is-loading"><br><br></div><br>
<!-- /wp:woocommerce/checkout-billing-address-block -->

<!-- wp:woocommerce/checkout-shipping-methods-block {"wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout-shipping-methods-block"></div>
<!-- /wp:woocommerce/checkout-shipping-methods-block -->

<!-- wp:woocommerce/checkout-payment-block {"wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout-payment-block"></div>
<!-- /wp:woocommerce/checkout-payment-block -->

<!-- wp:woocommerce/checkout-additional-information-block {"translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout alignwide wc-block-checkout is-loading"><br><br></div><br>
<!-- /wp:woocommerce/checkout-additional-information-block -->

<!-- wp:woocommerce/checkout-terms-block {"wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout-terms-block"></div>
<!-- /wp:woocommerce/checkout-terms-block -->

<!-- wp:woocommerce/checkout-actions-block {"translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout alignwide wc-block-checkout is-loading"><br><br></div><br>
<!-- /wp:woocommerce/checkout-actions-block --></div>
<!-- /wp:woocommerce/checkout-fields-block -->

<!-- wp:woocommerce/checkout-totals-block {"translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout-totals-block"><!-- wp:woocommerce/checkout-order-summary-block {"translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout-order-summary-block"><!-- wp:woocommerce/checkout-order-summary-cart-items-block {"translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout alignwide wc-block-checkout is-loading"><br><br></div><br>
<!-- /wp:woocommerce/checkout-order-summary-cart-items-block -->

<!-- wp:woocommerce/checkout-order-summary-coupon-form-block {"translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout alignwide wc-block-checkout is-loading"><br><br></div><br>
<!-- /wp:woocommerce/checkout-order-summary-coupon-form-block -->

<!-- wp:woocommerce/checkout-order-summary-totals-block {"translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout-order-summary-totals-block"><!-- wp:woocommerce/checkout-order-summary-subtotal-block {"translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout alignwide wc-block-checkout is-loading"><br><br></div><br>
<!-- /wp:woocommerce/checkout-order-summary-subtotal-block -->

<!-- wp:woocommerce/checkout-order-summary-fee-block {"translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout alignwide wc-block-checkout is-loading"><br><br></div><br>
<!-- /wp:woocommerce/checkout-order-summary-fee-block -->

<!-- wp:woocommerce/checkout-order-summary-discount-block {"translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout alignwide wc-block-checkout is-loading"><br><br></div><br>
<!-- /wp:woocommerce/checkout-order-summary-discount-block -->

<!-- wp:woocommerce/checkout-order-summary-shipping-block {"translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout alignwide wc-block-checkout is-loading"><br><br></div><br>
<!-- /wp:woocommerce/checkout-order-summary-shipping-block -->

<!-- wp:woocommerce/checkout-order-summary-taxes-block {"translatedWithWPMLTM":"1","wpmlTranslatedContent":"1"} -->
<div class="wp-block-woocommerce-checkout alignwide wc-block-checkout is-loading"><br><br></div><br>
<!-- /wp:woocommerce/checkout-order-summary-taxes-block --></div>
<!-- /wp:woocommerce/checkout-order-summary-totals-block --></div>
<!-- /wp:woocommerce/checkout-order-summary-block --></div>
<!-- /wp:woocommerce/checkout-totals-block --></div>
<!-- /wp:woocommerce/checkout --></div>
<!-- /wp:column -->

<!-- wp:column {"width":"16.5%","wpmlTranslatedContent":"1"} -->
<div class="wp-block-column" style="flex-basis:16.5%"></div>
<!-- /wp:column --></div>
<!-- /wp:columns --><!-- wp:paragraph {"dynamicAttributes":{"toolsetDSVersion":"257000"},"wpmlTranslatedContent":"1"} -->
<p></p>
<!-- /wp:paragraph -->
```

### Expected behavior

NA

### Actual behavior

NA

### Steps to reproduce

1. Use block checkout
2. Edit the code and replace with code above in the description 
3. Go to the checkout page - simply press Place Order without filling anything and see multiple concurrent requests made

### WordPress Environment

NA

### Isolating the problem

straku:

Closes #57791 

We use [`renderParentBlock`](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/atomic/utils/render-parent-block.tsx#L283) to render a checkout (but also cart, mini-cart, etc.) block on the frontend accepts a selector that is used to target a block received from the server. If this selector matches multiple DOM elements, it tries to render them multiple times (see `querySelectorAll` usage in [`renderFrontend`](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/base/utils/render-frontend.tsx#L229) function that is used to do all heavy lifting).

Since the checkout block is configured to allow only one instance of it on a page in the editor ([see here](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/blocks/checkout/block.json#L15)) I think we should do the same when rendering and only render it once.

This PR adds this constraint. See [this comment](https://github.com/woocommerce/woocommerce/pull/58001#issuecomment-2880972300) on how it was figured out which selector to use.

I am adding first easy reproduction steps for the issue if you want to go through them on `trunk` to see the bug in action.

1. Go to an editor of the page containing checkout block
2. Open code editor
3. At the bottom of the page add the following markup multiple times `<div class="wp-block-woocommerce-checkout"></div>`, to encounter the bug it is enough to add it twice, but you can go crazy if you want
4. Prepare for testing checkout flow on the frontend by having your store open with devtools on network tab, enable "Preserve log" option in the network tab
5. Go through the checkout flow, start paying attention to the network tab before clicking "Place Order"
6. After clicking "Place Order" notice that the request to `/wc/store/v1/checkout` happened multiple times (the number of the calls should be aligned with the amount of `.wp-block-woocommerce-checkout` you've added), with some of them failing (you might see an error state for a brief moment, but the flow should proceed to the order confirmation page)

Now to confirm the fix from this branch:

1. Repeat steps 1-5 from the above list
2. After clicking "Place Order" notice that there's only one request to `/wc/store/v1/checkout` and no error state visible briefly before continuing to the order confirmation page.

#### Cart

1. Go to editor of your test store cart page (or any other emtpy page)
2. Open code editor and replace existing structure with the following code

```
<!-- wp:woocommerce/cart -->
<div class="wp-block-woocommerce-cart"></div>
<!-- /wp:woocommerce/cart -->

<!-- wp:woocommerce/cart -->
<div class="wp-block-woocommerce-cart"></div>
<!-- /wp:woocommerce/cart -->

<!-- wp:woocommerce/cart -->
<div class="wp-block-woocommerce-cart"></div>
<!-- /wp:woocommerce/cart -->
```

3. Save the page (if you exit code editor now, it will complain that the blocks are malformed – the editor is right, but dismiss it for now)
4. Visit the frontend of the page you just saved
5. Verify that you can see only one cart on the page and that it is functional

#### Mini-Cart

We're going to do something really weird here. In my testing, it completely breaks the editor, but at least the frontend still looks great.

1. Go to the editor of your test store.
2. Find or add a Mini-Cart block to the page that you're editing.
3. Open Mini-Cart block settings and scroll to the Cart Drawer section. Click "Edit Mini-Cart Drawer template↗" link.
4. Now you're editing the Mini-Cart Drawer template. Go to the code editor.
5. In the code editor, duplicate the whole template (copy and paste below existing code). Save.
6. Open the frontend, click on Mini-Cart and verify that it still opens correctly. Is there any duplication that you can see?

I followed the steps listed above. I didn't try setting up the plugin (https://wpml.org/) that caused the issues for the reporter.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where multiple instances of the checkout process could be rendered at the same time, ensuring only one checkout block is displayed.

- **New Features**
  - Added support for controlling whether blocks can be rendered multiple times based on block settings.
  - Introduced new metadata configurations for the Cart and Mini-Cart Contents blocks, improving block registration and customization.

- **Refactor**
  - Updated block registration to use imported metadata objects for configuration, streamlining block setup.
  - Enhanced rendering utilities to support array-based iteration and new options for rendering single or multiple block instances.
  - Improved type definitions for greater consistency and flexibility in block rendering components.

- **Style**
  - Improved component prop types for better compatibility with React node structures.


opr:
> Maybe we should find all matching elements, but perform additional validation to find the one that matches our expected structure? What do you think?

I like this idea better than rendering an empty checkout, though I am wondering does block forcing not come into play in the first case? Would it not force the missing inner blocks to be added?

If we do go down the route of checking individual children, then I think we should have fairly loose requirements about what we believe is "valid", e.g. flattening the structure, checking that the children exist in any order, disregarding anything that isn't a WC Checkout specific element. If the store is using plugins that modify how blocks are saved we could potentially break a lot of stuff, so aggressively normalizing the markup should be a prerequisite.

straku:
> I am wondering does block forcing not come into play in the first case? Would it not force the missing inner blocks to be added?

~~I didn't notice any functionality in our rendering logic that would do something like this 🤔~~

Taking it back, taking a look at `renderForcedBlocks` now!

straku:
@opr I got my example partially wrong. This is what is happening on `trunk`.

This will render checkout without any of its children, forced blocks won't be applied:

```html
<div class="wp-block-woocommerce-checkout"></div>
```

---

This will render checkout with forced children, giving you a functional checkout on the page:

```html
<!-- wp:woocommerce/checkout -->
<div class="wp-block-woocommerce-checkout"></div>
<!-- /wp:woocommerce/checkout -->
```

---

This will render:
- a checkout without children
- and then checkout with forced blocks

It results in 2 instances and is the basis of the bug we're trying to address in this PR.

```html
<div class="wp-block-woocommerce-checkout"></div>

<!-- wp:woocommerce/checkout -->
<div class="wp-block-woocommerce-checkout"></div>
<!-- /wp:woocommerce/checkout -->
```

---

It might come in different variants. As an example, this is properly rendered on the page, but in reality there are 5 instances of checkout running, leading to various issues.

```html
<div class="wp-block-woocommerce-checkout"></div>

<!-- wp:woocommerce/checkout -->
<div class="wp-block-woocommerce-checkout"></div>
<!-- /wp:woocommerce/checkout -->

<div class="wp-block-woocommerce-checkout"></div>
<div class="wp-block-woocommerce-checkout"></div>
<div class="wp-block-woocommerce-checkout"></div>
```

---

For a full picture, this will render 2 visible checkouts on the page, but is warned as invalid in Gutenberg due to the expectancy of only one checkout block on the page.

```html
<!-- wp:woocommerce/checkout -->
<div class="wp-block-woocommerce-checkout"></div>
<!-- /wp:woocommerce/checkout -->

<!-- wp:woocommerce/checkout -->
<div class="wp-block-woocommerce-checkout"></div>
<!-- /wp:woocommerce/checkout -->
```

---

Based on the above, my proposition would be to not target the first DOM element that matches the selector, but to match the first occurrence of a proper block with enclosing comment. Let's look at a very contrived example:

```html
<!-- based on logic from this PR we would select this -->
<div class="wp-block-woocommerce-checkout"></div>

<!-- let's select this instead and ignore everything else -->
<!-- wp:woocommerce/checkout -->
<div class="wp-block-woocommerce-checkout"></div>
<!-- /wp:woocommerce/checkout -->

<div class="wp-block-woocommerce-checkout"></div>

<!-- wp:woocommerce/checkout -->
<div class="wp-block-woocommerce-checkout"></div>
<!-- /wp:woocommerce/checkout -->

<div class="wp-block-woocommerce-checkout"></div>
```

**EDIT:** This can be achieved by checking for the `data-block-name` attribute on the element. For checkout specifically, it would be set to `"woocommerce/checkout"`. Selector was changed to this pattern in https://github.com/woocommerce/woocommerce/pull/58001/commits/cc01df282f668e026fbb669a24f2d76554997893

opr:
> Based on the above, my proposition would be to not target the first DOM element that matches the selector, but to match the first occurrence of a proper block with enclosing comment. Let's look at a very contrived example:

This sounds good 👌🏼

opr (commented):
@straku what is the expected behaviour if you copy the checkout block in the editor using code view and paste it again?

so the markup in the code editor would be:

```html
<!-- wp:woocommerce/checkout -->
<div class="wp-block-woocommerce-checkout alignwide wc-block-checkout is-loading"><!-- wp:woocommerce/checkout-totals-block -->
<div class="wp-block-woocommerce-checkout-totals-block"><!-- wp:woocommerce/checkout-order-summary-block -->
<div class="wp-block-woocommerce-checkout-order-summary-block"><!-- wp:woocommerce/checkout-order-summary-cart-items-block -->
<div class="wp-block-woocommerce-checkout-order-summary-cart-items-block"></div>
<!-- /wp:woocommerce/checkout-order-summary-cart-items-block -->

<!-- wp:woocommerce/checkout-order-summary-coupon-form-block -->
<div class="wp-block-woocommerce-checkout-order-summary-coupon-form-block"></div>
<!-- /wp:woocommerce/checkout-order-summary-coupon-form-block -->

<!-- wp:woocommerce/checkout-order-summary-totals-block -->
<div class="wp-block-woocommerce-checkout-order-summary-totals-block"><!-- wp:woocommerce/checkout-order-summary-subtotal-block -->
<div class="wp-block-woocommerce-checkout-order-summary-subtotal-block"></div>
<!-- /wp:woocommerce/checkout-order-summary-subtotal-block -->

<!-- wp:woocommerce/checkout-order-summary-fee-block -->
<div class="wp-block-woocommerce-checkout-order-summary-fee-block"></div>
<!-- /wp:woocommerce/checkout-order-summary-fee-block -->

<!-- wp:woocommerce/checkout-order-summary-discount-block -->
<div class="wp-block-woocommerce-checkout-order-summary-discount-block"></div>
<!-- /wp:woocommerce/checkout-order-summary-discount-block -->

<!-- wp:woocommerce/checkout-order-summary-shipping-block -->
<div class="wp-block-woocommerce-checkout-order-summary-shipping-block"></div>
<!-- /wp:woocommerce/checkout-order-summary-shipping-block -->

<!-- wp:woocommerce/checkout-order-summary-taxes-block -->
<div class="wp-block-woocommerce-checkout-order-summary-taxes-block"></div>
<!-- /wp:woocommerce/checkout-order-summary-taxes-block --></div>
<!-- /wp:woocommerce/checkout-order-summary-totals-block --></div>
<!-- /wp:woocommerce/checkout-order-summary-block --></div>
<!-- /wp:woocommerce/checkout-totals-block -->

<!-- wp:woocommerce/checkout-fields-block {"metadata":{"ignoredHookedBlocks":["shipping-workshop/shipping-workshop-block"]}} -->
<div class="wp-block-woocommerce-checkout-fields-block"><!-- wp:woocommerce/checkout-express-payment-block -->
<div class="wp-block-woocommerce-checkout-express-payment-block"></div>
<!-- /wp:woocommerce/checkout-express-payment-block -->

<!-- wp:woocommerce/checkout-contact-information-block -->
<div class="wp-block-woocommerce-checkout-contact-information-block"></div>
<!-- /wp:woocommerce/checkout-contact-information-block -->

<!-- wp:woocommerce/checkout-shipping-method-block {"localPickupText":"Local pickup by thomas"} -->
<div class="wp-block-woocommerce-checkout-shipping-method-block"></div>
<!-- /wp:woocommerce/checkout-shipping-method-block -->

<!-- wp:woocommerce/checkout-pickup-options-block -->
<div class="wp-block-woocommerce-checkout-pickup-options-block"></div>
<!-- /wp:woocommerce/checkout-pickup-options-block -->

<!-- wp:woocommerce/eu-vat-number -->
<div class="wp-block-woocommerce-eu-vat-number"></div>
<!-- /wp:woocommerce/eu-vat-number -->

<!-- wp:woocommerce/checkout-shipping-address-block -->
<div class="wp-block-woocommerce-checkout-shipping-address-block"></div>
<!-- /wp:woocommerce/checkout-shipping-address-block -->

<!-- wp:woocommerce/checkout-billing-address-block -->
<div class="wp-block-woocommerce-checkout-billing-address-block"></div>
<!-- /wp:woocommerce/checkout-billing-address-block -->

<!-- wp:shipping-workshop/shipping-workshop-block /-->

<!-- wp:woocommerce/checkout-shipping-methods-block {"metadata":{"ignoredHookedBlocks":["shipping-workshop/shipping-workshop-block"]}} -->
<div class="wp-block-woocommerce-checkout-shipping-methods-block"></div>
<!-- /wp:woocommerce/checkout-shipping-methods-block -->

<!-- wp:woocommerce/checkout-payment-block -->
<div class="wp-block-woocommerce-checkout-payment-block"></div>
<!-- /wp:woocommerce/checkout-payment-block -->

<!-- wp:woocommerce/checkout-additional-information-block -->
<div class="wp-block-woocommerce-checkout-additional-information-block"></div>
<!-- /wp:woocommerce/checkout-additional-information-block -->

<!-- wp:woocommerce/checkout-order-note-block -->
<div class="wp-block-woocommerce-checkout-order-note-block"></div>
<!-- /wp:woocommerce/checkout-order-note-block -->

<!-- wp:woocommerce/checkout-terms-block -->
<div class="wp-block-woocommerce-checkout-terms-block"></div>
<!-- /wp:woocommerce/checkout-terms-block -->

<!-- wp:woocommerce/checkout-actions-block -->
<div class="wp-block-woocommerce-checkout-actions-block"></div>
<!-- /wp:woocommerce/checkout-actions-block --></div>
<!-- /wp:woocommerce/checkout-fields-block --></div>
<!-- /wp:woocommerce/checkout -->

<!-- wp:woocommerce/checkout -->
<div class="wp-block-woocommerce-checkout alignwide wc-block-checkout is-loading"><!-- wp:woocommerce/checkout-totals-block -->
<div class="wp-block-woocommerce-checkout-totals-block"><!-- wp:woocommerce/checkout-order-summary-block -->
<div class="wp-block-woocommerce-checkout-order-summary-block"><!-- wp:woocommerce/checkout-order-summary-cart-items-block -->
<div class="wp-block-woocommerce-checkout-order-summary-cart-items-block"></div>
<!-- /wp:woocommerce/checkout-order-summary-cart-items-block -->

<!-- wp:woocommerce/checkout-order-summary-coupon-form-block -->
<div class="wp-block-woocommerce-checkout-order-summary-coupon-form-block"></div>
<!-- /wp:woocommerce/checkout-order-summary-coupon-form-block -->

<!-- wp:woocommerce/checkout-order-summary-totals-block -->
<div class="wp-block-woocommerce-checkout-order-summary-totals-block"><!-- wp:woocommerce/checkout-order-summary-subtotal-block -->
<div class="wp-block-woocommerce-checkout-order-summary-subtotal-block"></div>
<!-- /wp:woocommerce/checkout-order-summary-subtotal-block -->

<!-- wp:woocommerce/checkout-order-summary-fee-block -->
<div class="wp-block-woocommerce-checkout-order-summary-fee-block"></div>
<!-- /wp:woocommerce/checkout-order-summary-fee-block -->

<!-- wp:woocommerce/checkout-order-summary-discount-block -->
<div class="wp-block-woocommerce-checkout-order-summary-discount-block"></div>
<!-- /wp:woocommerce/checkout-order-summary-discount-block -->

<!-- wp:woocommerce/checkout-order-summary-shipping-block -->
<div class="wp-block-woocommerce-checkout-order-summary-shipping-block"></div>
<!-- /wp:woocommerce/checkout-order-summary-shipping-block -->

<!-- wp:woocommerce/checkout-order-summary-taxes-block -->
<div class="wp-block-woocommerce-checkout-order-summary-taxes-block"></div>
<!-- /wp:woocommerce/checkout-order-summary-taxes-block --></div>
<!-- /wp:woocommerce/checkout-order-summary-totals-block --></div>
<!-- /wp:woocommerce/checkout-order-summary-block --></div>
<!-- /wp:woocommerce/checkout-totals-block -->

<!-- wp:woocommerce/checkout-fields-block {"metadata":{"ignoredHookedBlocks":["shipping-workshop/shipping-workshop-block"]}} -->
<div class="wp-block-woocommerce-checkout-fields-block"><!-- wp:woocommerce/checkout-express-payment-block -->
<div class="wp-block-woocommerce-checkout-express-payment-block"></div>
<!-- /wp:woocommerce/checkout-express-payment-block -->

<!-- wp:woocommerce/checkout-contact-information-block -->
<div class="wp-block-woocommerce-checkout-contact-information-block"></div>
<!-- /wp:woocommerce/checkout-contact-information-block -->

<!-- wp:woocommerce/checkout-shipping-method-block {"localPickupText":"Local pickup by thomas"} -->
<div class="wp-block-woocommerce-checkout-shipping-method-block"></div>
<!-- /wp:woocommerce/checkout-shipping-method-block -->

<!-- wp:woocommerce/checkout-pickup-options-block -->
<div class="wp-block-woocommerce-checkout-pickup-options-block"></div>
<!-- /wp:woocommerce/checkout-pickup-options-block -->

<!-- wp:woocommerce/eu-vat-number -->
<div class="wp-block-woocommerce-eu-vat-number"></div>
<!-- /wp:woocommerce/eu-vat-number -->

<!-- wp:woocommerce/checkout-shipping-address-block -->
<div class="wp-block-woocommerce-checkout-shipping-address-block"></div>
<!-- /wp:woocommerce/checkout-shipping-address-block -->

<!-- wp:woocommerce/checkout-billing-address-block -->
<div class="wp-block-woocommerce-checkout-billing-address-block"></div>
<!-- /wp:woocommerce/checkout-billing-address-block -->

<!-- wp:shipping-workshop/shipping-workshop-block /-->

<!-- wp:woocommerce/checkout-shipping-methods-block {"metadata":{"ignoredHookedBlocks":["shipping-workshop/shipping-workshop-block"]}} -->
<div class="wp-block-woocommerce-checkout-shipping-methods-block"></div>
<!-- /wp:woocommerce/checkout-shipping-methods-block -->

<!-- wp:woocommerce/checkout-payment-block -->
<div class="wp-block-woocommerce-checkout-payment-block"></div>
<!-- /wp:woocommerce/checkout-payment-block -->

<!-- wp:woocommerce/checkout-additional-information-block -->
<div class="wp-block-woocommerce-checkout-additional-information-block"></div>
<!-- /wp:woocommerce/checkout-additional-information-block -->

<!-- wp:woocommerce/checkout-order-note-block -->
<div class="wp-block-woocommerce-checkout-order-note-block"></div>
<!-- /wp:woocommerce/checkout-order-note-block -->

<!-- wp:woocommerce/checkout-terms-block -->
<div class="wp-block-woocommerce-checkout-terms-block"></div>
<!-- /wp:woocommerce/checkout-terms-block -->

<!-- wp:woocommerce/checkout-actions-block -->
<div class="wp-block-woocommerce-checkout-actions-block"></div>
<!-- /wp:woocommerce/checkout-actions-block --></div>
<!-- /wp:woocommerce/checkout-fields-block --></div>
<!-- /wp:woocommerce/checkout -->
```

In this case, the checkout block _is_ rendered twice on the front end.

straku:
> What is the expected behaviour if you copy the checkout block in the editor using code view and paste it again?

@opr I don't know if I would call it an expected behavior, but the change introduced in this PR doesn't change the current `trunk` behavior. As you've experienced, cart or checkout will be rendered twice, but the editor will properly inform you that the markup is malformed.

We can account for this case and add a `:first-of-type` to the selector, but should we?

opr:
> > What is the expected behaviour if you copy the checkout block in the editor using code view and paste it again?
> 
> @opr I don't know if I would call it an expected behavior, but the change introduced in this PR doesn't change the current `trunk` behavior. As you've experienced, cart or checkout will be rendered twice, but the editor will properly inform you that the markup is malformed.
> 
> We can account for this case and add a `:first-of-type` to the selector, but should we?

Yes it warns that it's incorrect, but multiple blocks are rendered on the front end too. Would `:first-of-type` help here if there was additional markup wrapping the other checkout block (just trying to think of edge cases).

```html
<!-- wp:woocommerce/checkout -->
  <div class="wp-block-woocommerce-checkout"></div>
<!-- /wp:woocommerce/checkout -->

<div>
  <!-- wp:woocommerce/checkout -->
    <div class="wp-block-woocommerce-checkout"></div>
  <!-- /wp:woocommerce/checkout -->
</div>
```

Could we perhaps add something to `renderParentBlock` (or the functions it calls) to track the number of times something has been rendered, and compare this to the block's `supports.multiple` value, if `supports.multiple` is false, and the block has already been rendered, skip rendering

straku:
> Would :first-of-type help here if there was additional markup wrapping the other checkout block (just trying to think of edge cases).

No, it would not. `:first-of-type` "represents the first element of its type (tag name) among a group of sibling elements" (from [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/:first-of-type)).

> Could we perhaps add something to `renderParentBlock` (or the functions it calls) to track the number of times something has been rendered, and compare this to the block's `supports.multiple` value, if `supports.multiple` is false, and the block has already been rendered, skip rendering 

Yes, it is possible. A simpler alternative would be for it to accept an additional parameter that would prevent rendering multiple occurrences of the block, and either: set it explicitly at the call site for affected blocks, or add the logic you described and read the block configuration. Doing it at call site feels better to me.

@opr I will explore this angle today.

plugins/woocommerce/client/blocks/assets/js/base/utils/render-frontend.tsx
```diff
@@ -184,18 +182,27 @@ const renderBlockOutsideWrappers = <
...
-		Array.prototype.filter.call( containers, ( el ) => {
+		containers = containers.filter( ( el ) => {
 			return ! isElementInsideWrappers( el, wrappers );
 		} );
 	}
 
+	// Limit to first element if multiple option is false
+	if ( options?.multiple === false ) {
+		containers = containers.slice( 0, 1 );
+	}
```
straku, line 204: The clue of the change is here. The rest is a cleanup that is the result of a decision to convert `NodeList` to a standard array right after querying for the elements.

plugins/woocommerce/client/blocks/assets/js/base/utils/render-frontend.tsx
```diff
@@ -225,8 +232,17 @@ const renderBlockInsideWrapper = <
...
 }: RenderBlockInsideWrapperParams< TProps, TAttributes > ): void => {
-	const containers = wrapper.querySelectorAll( selector );
+	let containers: HTMLElement[] = Array.from(
+		wrapper.querySelectorAll( selector )
+	);
+
+	// Limit to first element if multiple option is false
+	if ( options?.multiple === false ) {
+		containers = containers.slice( 0, 1 );
+	}
```
straku, line 244: And here

plugins/woocommerce/client/blocks/assets/js/base/utils/render-frontend.tsx
```diff
@@ -266,23 +289,30 @@ export const renderFrontend = <
...
+		document.body.querySelectorAll( selectorsToSkipOnLoad.join( ',' ) )
 	);
 
-	const { Block, getProps, getErrorBoundaryProps, selector } = props;
+	const {
+		Block,
+		getProps,
+		getErrorBoundaryProps,
+		selector,
+		options = { multiple: true },
```
straku, line 301: The default is set to be consistent with the current behavior

straku:
@opr One "downside" to this approach is that it prevents rendering on the frontend but doesn't do anything on the server to strip malformed markup (I am not sure if this is something we would ever approach in the WordPress ecosystem?). So, any markup rendered on the server still ends up on the frontend, but we don't hydrate it in any way. This prevents business logic issues that were mentioned in the original ticket, but doesn't prevent some markup being rendered on the frontend.

In the case of the markup you provided, we will see an additional checkout in loading state below the properly rendered (and hydrated) one. It won't be used by frontend rendering logic, but it's still there, and usually, it matches some of the loaded CSS styles, so some parts of it get a visual treatment.

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/frontend.tsx
```diff
@@ -56,11 +56,15 @@ const Wrapper = ( {
...
+	selector:
+		'.wp-block-woocommerce-checkout[data-block-name="woocommerce/checkout"]',
 	getProps,
 	blockMap: getRegisteredBlockComponents( blockName ) as Record<
 		string,
 		React.ReactNode
 	>,
 	blockWrapper: Wrapper,
+	options: {
+		multiple: false,
```
opr: is it possible to get this from the metadata in block.json instead of defining it here? Ideally, this logic would apply to all of our blocks where `multiple: false` is set not just the checkout.

opr (changes requested):
Thanks Łukasz I like the approach. I agree trying to limit/change what we output from the server seems like we're overstepping our boundaries a bit. I think if the store has markup for two checkout blocks, then preventing rendering of more than one into a React component should be the limit of our responsibility.

I added a comment about how `multiple` is derived. We should be getting this from block metadata if possible, so that it applies to all blocks. Testing steps should be updated to include the Cart block also if we do this.

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/frontend.tsx
straku: The usage of `renderParentFrontend` throughout the project is minimal (4 places). I would prefer not to complicate it more by passing a pretty extensive configuration object that will be mostly unused. As an alternative/compromise, I propose:
- to make `options` required parameter (for it to be enforced we would have to fix some TS issues that we have with the current usage of this function, as other errors are preventing from checking the presence of all required arguments)
- not provide a literal `multiple: false` for options but import the value from existing metadata

The above proposition is pushed in https://github.com/woocommerce/woocommerce/pull/58001/commits/97f9dcafd7744ec1c4a102effca20c10c1fc11db so you can inspect it in practice.

Edit: Updated testing steps to cover Cart and Mini-Cart Drawer too.

plugins/woocommerce/client/blocks/assets/js/blocks/filter-wrapper/frontend.ts
```diff
@@ -14,4 +14,8 @@ renderParentBlock( {
 	selector: '.wp-block-woocommerce-filter-wrapper',
 	Block,
 	blockMap: getRegisteredBlockComponents( 'woocommerce/filter-wrapper' ),
+	options: {
+		multiple: true,
+	},
+	getProps: () => ( {} ),
```
opr, line 20: Why these changes in this PR?

opr (changes requested):
@straku thanks, I tested this and it seems pretty good, there are some E2E fails that seem to be happening consistently (I think?) I re-ran them a few times and still seeing failures, could you investigate.

Code changes look fairly safe to me and I doubt they would influence the tests, but better to check.

Also left one comment on some code I'm not sure why it was included.

plugins/woocommerce/client/blocks/assets/js/blocks/filter-wrapper/frontend.ts
straku: `getProps` is required (TypeScript), safer to update in this one call site rather than making the property optional globally (I think)

opr (approved):
Thanks @straku looks good now, I tested with multiple blocks, mini cart looks good, Cart/Checkout load fine, so do other FSE blocks and filters.

🚢 it!

PR #57627, opened 2025-04-30
Fix Add to Cart with Options Stepper style having missing styles

Aljullu:
### Changes proposed in this Pull Request:

Fixes an issue that caused _Stepper_ styles not to be loaded in the _Add to Cart with Options_ block.

Kudos to @AhmarZaidi for finding this issue (https://github.com/woocommerce/woocommerce/pull/57570#issuecomment-2841346423).

### How to test the changes in this Pull Request:

1. In a classic theme (ie: Storefront).
2. Create a post or page and add the _Single Product_ block.
3. Select any product and change the _Add to Cart with Options_ block display style to _Stepper_.
4. In the frontend, verify the buttons look correct:

Before | After
--- | ---
imatge | imatge

5. Switch to a block theme (ie: Twenty-Twenty Five). Remove the _Mini-Cart_ block from the header if it's there.
6. Reload the same post and verify the buttons look correct as well:

Before | After
--- | ---
imatge | imatge

7. Install and activate the [WooCommerce Beta Tester](https://woocommerce.com/products/woocommerce-beta-tester/) extension.
8. Go to _WooCommerce Admin Test Helper_ > _Features_ and enable `experimental-blocks` and `blockified-add-to-cart`.
9. Go to _Appearance_ > _Editor_ > _Templates_ > _Single Product_.
10. Click on the _Add to Cart with Options_ block and update to the blockified version.
11. Visit a product template in the frontend and verify the styles are also loaded correctly:

Before | After
--- | ---
imatge | imatge

PR #58002, opened 2025-05-14
[Experimental] Blockified Add to Cart with Options block: make it possible to add grouped products to cart

### LINKED ISSUE #55697: Blockified Add to Cart with Options block: make it possible to add grouped products to cart
Similar to https://github.com/woocommerce/woocommerce/issues/54416 but for grouped products.

Once we have all the inner blocks of the blockified Add to Cart with Options block ready, we will need to make them actually work. This issue is focused exclusively on grouped products.

As an outcome, it should be possible to add the blockified _Add to Cart with Options_ block in the Single Product template (or the Single Product block when a grouped product is selected) and, from the frontend, add to cart without causing a page reload.

mikachan:

Closes https://github.com/woocommerce/woocommerce/issues/55697.

This PR allows grouped products to be added to the cart without the need for a page reload. It covers the following:

- Enables `ajax_add_to_cart` for grouped products.
- Adds grouped product context to the `addToCartWithOptionsStore` store and the Add to Cart button, which means that the child product IDs and quantities can be applied to the cart. 

https://github.com/user-attachments/assets/258eaf31-2cb9-42af-8b57-1b15c3f40098

6. Then click `Save` to publish the changes.
7. Go to the store and click on any grouped product or create a new one.
8. Change some of the quantities of the child products.
9. Click add to cart.
10. Check that the correct products and quantities are added to the cart, and that the page does not reload.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Blockified Add to Cart with Options block: make it possible to add grouped products to cart

</details>

<details>

#### Comment 

</details>


mikachan:
[These e2e tests](https://github.com/woocommerce/woocommerce/actions/runs/15066507406/job/42355584594?pr=58002) seem to be consistently timing out on this PR - trying to work out how to fix them 🤔

dinhtungdu:
Just so you know that we're updating the Product Button store at https://github.com/woocommerce/woocommerce/pull/58097, which likely to affect this PR.

mikachan:
Thanks @dinhtungdu, I'll wait for that PR to land before continuing with this one.

dinhtungdu:
@mikachan https://github.com/woocommerce/woocommerce/pull/58097 is merged :).

mikachan:
Thanks @dinhtungdu! I've merged in trunk, and I think I've fixed the related e2e tests now; the remaining failure seems to be caused by a flaky test that's happening on other PRs as well.

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
```diff
@@ -120,10 +120,37 @@ protected function render( $attributes, $content, $block ) {
...
 		);
 
-		$number_of_items_in_cart  = $this->get_cart_item_quantities_by_product_id( $product->get_id() );
+		$is_grouped              = $product->is_type( 'grouped' );
+		$grouped_product_ids     = $product->get_children();
+		$is_product_purchasable  = $product->is_purchasable();
+		$number_of_items_in_cart = $this->get_cart_item_quantities_by_product_id( $product->get_id() );
+
+		if ( $is_grouped ) {
+			$children = $product->get_children();
```
Aljullu: Nit: `$grouped_product_ids` and `$children` hold the same value, no? I guess we can remove one of the two variables.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
```diff
@@ -205,24 +212,83 @@ const addToCartWithOptionsStore = store(
...
-				const currentQuantity = product?.quantity || 0;
+				if (
+					productType === 'grouped' &&
+					groupedProductIds.length > 0
+				) {
+					// Handle grouped products
+					const form = event.target as HTMLFormElement;
+					const quantityInputs = form.querySelectorAll(
+						'.input-text.qty.text'
+					);
```
Aljullu: I wonder if instead of querying the input HTML elements here, it might be better to have each product's quantity saved in the store. So `quantity`, instead of being an int, it could be an object matching `productId => quantity`.

I'm fine leaving this for a follow-up PR if you prefer, as the current approach works, but I feel that saving the quantities in the store is a more iAPI-_ish_ approach. What do you think?

Aljullu (commented):
Thanks for working on this, @mikachan! Really great to see that it's possible to add grouped products to cart. :raised_hands: 

I left some comments below, but happy to discuss them:

---

Not a strong opinion, and we might want to ask design, but I expected the Product Button not to show the "X in cart" text in grouped products. I feel it might be confusing:

https://github.com/user-attachments/assets/68f83346-5e60-4391-9203-376547158098

Looking at the screenshot above, when the button says "4 in cart" I would expect to have 4 _Hoodie with Logo_, 4 _T-Shirt_ and 4 _Beanie_ in the cart.

Instead, I think it could simply change from `Add to cart` to `Added to cart` or something like this, to signal that the action was completed.

But as mentioned, not a strong opinion and happy to discuss. :slightly_smiling_face:

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
```diff
@@ -120,10 +120,37 @@ protected function render( $attributes, $content, $block ) {
 			)
 		);
 
-		$number_of_items_in_cart  = $this->get_cart_item_quantities_by_product_id( $product->get_id() );
```
dinhtungdu, line 123: I think we should move the logic calculating the number of product in cart of grouped product also in `get_cart_item_quantities_by_product_id`, leaving this render method unchanged. It makes sense to me to provide a product id and get the number of product in cart back, regardless of product type.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/GroupedProductSelector.php
```diff
@@ -32,9 +32,31 @@ protected function render( $attributes, $content, $block ): string {
...
+	 * @param  bool        $supports If features are already supported or not.
+	 * @param  string      $feature  The feature to check if is supported.
+	 * @param  \WC_Product $product  The product to check.
+	 * @return bool True if the product supports the feature, false otherwise.
+	 * @since  9.9.0
+	 */
+	public function check_product_supports( $supports, $feature, $product ) {
+		if ( 'ajax_add_to_cart' === $feature ) {
+			return true;
+		}
```
dinhtungdu: Aren't we using iAPI to add the product to cart, why do we need to check for `ajax_add_to_cart` support? I thought they're two things, is that correct?

dinhtungdu: Also can we have more explicit name, this function only enable `ajax_add_to_cart` support, so we should mentioned that in the function name I think.

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
```diff
@@ -166,11 +193,13 @@ protected function render( $attributes, $content, $block ) {
...
-			'productId'       => $product->get_id(),
-			'addToCartText'   => $add_to_cart_text,
-			'tempQuantity'    => $number_of_items_in_cart,
-			'animationStatus' => 'IDLE',
+			'quantityToAdd'     => $default_quantity,
+			'productId'         => $product->get_id(),
+			'addToCartText'     => $add_to_cart_text,
+			'tempQuantity'      => $number_of_items_in_cart,
+			'isGrouped'         => $is_grouped,
+			'groupedProductIds' => $grouped_product_ids,
```
dinhtungdu: I think we can get this data from the parent context, as we're already declared the same in the Add to Cart + Options block.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
```diff
@@ -205,24 +212,83 @@ const addToCartWithOptionsStore = store(
...
+						const product = wooState.cart?.items.find(
+							( cartItem ) => cartItem.id === item.id
+						);
+						const currentQuantity = product?.quantity || 0;
+
+						yield actions.addCartItem( {
+							id: item.id,
+							quantity: currentQuantity + item.quantity,
+							variation: item.variation,
+						} );
```
dinhtungdu: FYI, StoreAPI supports batching, I think we should leverage that ability instead of making multiple request like this: https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/StoreApi/docs/cart.md#add-item

dinhtungdu (changes requested):
Thanks for working on this. I tested the PR and can verify that I can add grouped product to the cart, but the products are added one by one, which I think we should fixed by batching the Store API request (as mentioned below).

https://github.com/user-attachments/assets/604c2209-29b4-4223-b724-a0c230d4746e

I'm not sure this is raised elsewhere but mentioned it here to not lose the issue:

When the Mini Cart isn't used on the page, the quantity step style is missing.
<details><summary>Screenshot</summary>
<p>

image

</p>
</details>

Aljullu:
> When the Mini Cart isn't used on the page, the quantity step style is missing.

Ups, we fixed that for some blocks in https://github.com/woocommerce/woocommerce/pull/57627, but I forgot about this one. :grimacing: Given that this is also reproducible in `trunk`, I fixed it in a separate PR here: https://github.com/woocommerce/woocommerce/pull/58396.

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
mikachan: Thanks! I've ended up removing `$is_grouped` for now so these variables are no longer used.

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
mikachan: That makes sense - I've moved the grouped product quantity logic to `get_cart_item_quantities_by_product_id ` in https://github.com/woocommerce/woocommerce/pull/58002/commits/016f5786ff1a1f33788f3a8869a36ef2260150b1.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/GroupedProductSelector.php
mikachan: I'm not sure on the details, and this was based on the variable product implementation in https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/VariationSelector.php#L56.

If I remove the check for this feature, then the page reloads when adding to cart. Do you know if there's another way to prevent the page reload?

I've made the function name more specific in https://github.com/woocommerce/woocommerce/pull/58002/commits/7e5f56704a0466eef1a6dae833fb0e3489043f98.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
mikachan: Oh nice, I didn't know about this. I've added a new `batchAddCartItems` action in https://github.com/woocommerce/woocommerce/pull/58002/commits/51bd56cccce2633ce140f4fb016b43567628facd. Is this what you had in mind?

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
mikachan: I've removed the `isGrouped` variable, and I've tweaked how the `groupedProductIds` is applied so it's not added for every product. I think this is looking better but I'll see if there's anything else we can improve here.

I'm using the `groupedProductIds` for checking the quantity of items in the cart. Otherwise, I think the only product ID available in context is the parent ID, so it's not possible to check these quantities, and I'm not sure how else to query for child IDs on the frontend of the button. The quantity check is for displaying the different Add to Cart text ("Add to Cart" or "Added to Cart").

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/GroupedProductSelector.php
dinhtungdu: @Aljullu can you share more insights here?

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
```diff
@@ -98,32 +98,64 @@ protected function render( $attributes, $content, $block ) {
...
+			if ( $product->is_type( 'grouped' ) ) {
+				return __( 'Added to cart', 'woocommerce' );
+			}
+
+			return sprintf(
+				/* translators: %s: product number. */
+				__( '%s in cart', 'woocommerce' ),
+				'###'
+			);
+		}
```
dinhtungdu: We can't declare a function like this, there may be multiple Product Button on the same page, the render method can be executed multiple time (like Product Collection), resulting a fatal error.

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
```diff
@@ -98,32 +98,64 @@ protected function render( $attributes, $content, $block ) {
...
+							array_map(
+								function ( $child_product_id ) {
+									return $this->get_cart_item_quantities_by_product_id( (int) $child_product_id );
+								},
+								$grouped_product_ids
+							)
+						);
+
+						return $grouped_product_ids_in_cart > 0 ? __( 'Added to cart', 'woocommerce' ) : $add_to_cart_text;
+					}
```
dinhtungdu: I think we don't need to check for the sum, we only need to check if there is a child product in cart is enough. So instead of the sum, we can loop through child product and return early if the product is in cart.

dinhtungdu: Performance-wise, it may be negligible, though.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
```diff
@@ -77,9 +84,32 @@ const productButtonStore = {
...
+						const product = wooState.cart?.items.find(
+							( item ) => item.id === productId
+						);
+						return product?.quantity || 0;
+					}
+				);
+				if ( groupedProductIdsInCart?.some( ( qty ) => qty > 0 ) ) {
+					return state.inTheCartText;
+				}
+				return addToCartText;
```
dinhtungdu, line 104: Looks like this return statement can be removed.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
```diff
@@ -77,9 +84,32 @@ const productButtonStore = {
...
-			if ( quantity === 0 ) return addToCartText;
+			if ( productType !== 'grouped' ) {
+				if ( quantity > 0 ) {
+					return state.inTheCartText.replace(
+						'###',
+						quantity.toString()
+					);
+				}
+				return addToCartText;
+			}
```
dinhtungdu: I think we should handle specific case first (grouped product), then the remaining cases later. I mean that moving this block of code below the grouped product condition and remove the if statement, saving us one level indentation.

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
```diff
@@ -301,10 +338,49 @@ private function get_cart_item_quantities_by_product_id( $product_id ) {
...
+			return array_sum(
+				array_map(
+					function ( $child_product_id ) {
+						return $this->get_cart_item_quantities_by_product_id( (int) $child_product_id );
+					},
+					$grouped_product_ids
+				)
+			);
+		}
+
```
dinhtungdu: Is this change needed anymore?

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
```diff
@@ -142,6 +143,102 @@ const { state, actions } = store< Store >(
...
+							// Optimistically updates the number of items in the cart.
+							existingItem.quantity = item.quantity;
+							if ( existingItem.key ) {
+								quantityChanges.cartItemsPendingQuantity = [
+									existingItem.key,
+								];
+							}
+							return {
+								method: 'POST',
+								path: `/wc/store/v1/cart/add-item`,
```
dinhtungdu: Should this be `update-item`?

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
```diff
@@ -142,6 +143,102 @@ const { state, actions } = store< Store >(
...
+						}
+
+						// Adds new cart item.
+						item = {
+							id: item.id,
+							quantity: item.quantity,
+							variation: item.variation,
+						} as OptimisticCartItem;
+						state.cart.items.push( item );
+						quantityChanges.productsPendingAdd = [ item.id ];
```
dinhtungdu: By doing so, we're not properly tracking the pending items, as we always replace the array, not update, resulting only one pending item at the time, while there may be more. I don't know the consequent of this can be, tbh, as I don't see we use it anywhere. Maybe @Aljullu has more context.

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
```diff
@@ -142,6 +143,102 @@ const { state, actions } = store< Store >(
...
+
+					// Checks if the response was successful, but still contains some errors.
+					json.errors?.forEach( ( error ) => {
+						actions.showNoticeError( error );
+					} );
+
+					// Ensure we only get a single cart response
+					const cartResponse = Array.isArray( json.responses )
+						? json.responses[ 0 ].body
+						: json;
```
dinhtungdu: Can you explain why single response? I checked the response from the batch endpoint adding 3 products inside a grouped product, I got 3 items returned.

image

dinhtungdu (changes requested):
Thanks for the updates! I can verify that that new action adding products in batch is working as expected and better than adding one by one. I also agree that using `Added to cart` text instead of explicit quantity for grouped product is better and less confusing.

There are some issue with the current PR that I detailed in the comments, please take a look at this. For complex PR like this and the review is getting longer, if you think a pair may help, please ping me, my calendar is pretty much empty :D.

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
mikachan: Ah sorry, this was meant to be temporary! Moved the function in https://github.com/woocommerce/woocommerce/pull/58002/commits/4be89b5f4e01053fa573b73180d8802f40c922e2.

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
mikachan: That makes sense. I've updated this so it only checks if a product is in the cart, rather than totalling the quantities: https://github.com/woocommerce/woocommerce/pull/58002/commits/58d8879a14db6fe0b97f1ed4b1721a7e45f21a5a.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
mikachan: Handled in https://github.com/woocommerce/woocommerce/pull/58002/commits/de42784352b8097001846eacd09a79254f8de15a.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
mikachan: Yes, agreed, thank you! I've re-ordered these cases and also removed the unnecessary return mentioned above. I think this looks better: https://github.com/woocommerce/woocommerce/pull/58002/commits/de42784352b8097001846eacd09a79254f8de15a.

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
mikachan: Nope! Removed in https://github.com/woocommerce/woocommerce/pull/58002/commits/643380a9581be09d5415ec6eddd9eeb49fc5943d.

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
mikachan: Yes I think it should be `update-item`, but when I try this I'm seeing an error about the cart item not existing:

image

I'm wondering if the batch endpoint only works with `add-item`, as I only see `add-item` used as an example in the [docs](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/StoreApi/docs/cart.md#add-item). Or maybe I've missed something from the implementation.

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
mikachan: I did this because I think `state.cart` expects to be a single cart object, rather than an array of cart objects. As the batch endpoint returns an array, I didn't think we'd want this applied to the local state. But maybe I'm missing something with how to handle this response.

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
mikachan: Nevermind, I think I've figured this out, I was doing it wrong! Fixed in https://github.com/woocommerce/woocommerce/pull/58002/commits/ba8b7e93bf9f2eeabdf7a85c26468f84ce443be9.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
mikachan: Great idea, thank you. I've tried this in https://github.com/woocommerce/woocommerce/pull/58002/commits/9a60a1eba5a522f7e4429633339f10845e468350. I'm hoping I've understood correctly - let me know otherwise!

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
```diff
@@ -208,24 +216,73 @@ const addToCartWithOptionsStore = store(
...
+							( item ) => item.id === childProductId
+						);
+						const currentQuantity = existingProduct?.quantity || 0;
+
+						// Skip adding variable products to the cart when grouped.
+						if (
+							groupedVariableProductIds.includes( childProductId )
+						) {
+							continue;
+						}
```
mikachan: During testing, I noticed that if a variable product is included in a grouped product, it causes an add-to-cart error (No matching variation found.), and the variable product isn't added to cart. For the time being, I've prevented variable products from being added to the cart along with the other child products, but maybe there's a better way to handle this.

There's a button that's shown for variable products that links to a separate product page, so I was working on the assumption that a variable product should be handled from its own product page.

image

Maybe this would be better handled in a follow-up PR.

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
dinhtungdu: I see. We probably want the last cart object instead of the first one, after all items are added to the cart.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
Aljullu: > For the time being, I've prevented variable products from being added to the cart along with the other child products, but maybe there's a better way to handle this.

Yes, I think that's how it should be handled. Only the products which display an input should be added to cart when _Add to cart_ is clicked. That means variable products, other grouped products, out of stock products, products with no price, etc. shouldn't be added to cart.

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
mikachan: Yes I agree, we're using the last cart object [here](https://github.com/woocommerce/woocommerce/pull/58002/files#diff-94da99060e18b7a1c5747fc964ede9549441e0ca5e73a0e22e69bbdbcf6fdcb0R221-R224).

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
mikachan: That makes sense. I've updated the logic so that any grouped child products that are variable, grouped, out of stock, or not purchasable will not be added to the cart in https://github.com/woocommerce/woocommerce/pull/58002/commits/f903bfcc26174e8aa00e2ee8098531b100207b3b.

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
mikachan: I think I've addressed this by spreading the existing `productsPendingAdd` and appending the new item, instead of replacing the array: https://github.com/woocommerce/woocommerce/pull/58002/commits/fd7958da945c4cc4ace8d847c023ae9fefa242e0.

mikachan:
Thanks @dinhtungdu and @Aljullu for all the feedback and reviews so far 🙇 

I've made some more changes, including improving how the `quantity` context works for grouped products, and I think this PR is working well now.

There's still one question outstanding: Is there an alternative for the  `ajax_add_to_cart` feature check to prevent the page reload when adding to cart? ([Related comment](https://github.com/woocommerce/woocommerce/pull/58002#discussion_r2113446219).) But perhaps it's something we can address in a follow-up, as it looks like variable products are using the same approach.

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
```diff
@@ -142,6 +143,111 @@ const { state, actions } = store< Store >(
...
+
+					// Checks if the response was successful, but still contains some errors.
+					json.errors?.forEach( ( error ) => {
+						actions.showNoticeError( error );
+					} );
+
+					// Use the last cart response to update the local cart.
+					const cartResponse = Array.isArray( json.responses )
+						? json.responses[ json.responses.length - 1 ].body
+						: json;
```
dinhtungdu: The response can contains other item than cart, so we need to pick the last cart response with biggest item count instead.

image

You can reproduce the issue by setting the last product stock to sold individually.

dinhtungdu (dismissed):
Thank you for all the updates. Other than the review comment, this is LGTM.

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
mikachan: Thank you! I managed to reproduce this. I've updated this so that it uses the last successful response instead: https://github.com/woocommerce/woocommerce/pull/58002/commits/62b165fc0f6c5a8b5b7430de1676f17a34a78bc8.

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/GroupedProductSelector.php
Aljullu: We briefly discussed it here: https://github.com/woocommerce/woocommerce/pull/56378/files#r1999016470.

I think we could update the Product Button block to ignore the `ajax_add_to_cart` value when it's inside the _Add to Cart + Options_ block. But probably something for a separate PR, given that we are already overriding it in variable products, as @mikachan said.

I opened a separate issue for this: https://github.com/woocommerce/woocommerce/issues/58559 (WOOPLUG-4516).

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
```diff
@@ -142,6 +143,119 @@ const { state, actions } = store< Store >(
...
+							method: 'POST',
+							headers: {
+								Nonce: state.nonce,
+								'Content-Type': 'application/json',
+							},
+							body: JSON.stringify( { requests } ),
+						}
+					);
+
+					const json: Cart = yield res.json();
```
dinhtungdu: The type of this data is not `Cart`, but the following (see https://github.com/WordPress/wordpress-develop/blob/1651947e3dc6ba14e227cf5f2045df6c16908045/src/wp-includes/rest-api/class-wp-rest-server.php#L1816-L1822 and https://github.com/WordPress/wordpress-develop/blob/1651947e3dc6ba14e227cf5f2045df6c16908045/src/wp-includes/rest-api/class-wp-rest-server.php#L1862)
```ts
{
    responses: unknown[];
    failed: 'validation'
}
```

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
```diff
@@ -142,6 +143,119 @@ const { state, actions } = store< Store >(
...
+						? json.responses.filter(
+								( response ) => response.status === 200
+						  )
+						: [];
+					const cartResponse =
+						successfulResponses.length > 0
+							? successfulResponses[
+									successfulResponses.length - 1
+							  ].body
+							: json;
```
dinhtungdu: We can not use the `json` here as the fallback. We need to ensure that we only set the valid cart response. As I said above, `json` here isn't a valid `Cart` response, so if we don't find any valid response, we shouldn't update the cart.

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
```diff
@@ -105,25 +105,37 @@ protected function render( $attributes, $content, $block ) {
 					$context = wp_interactivity_get_context();
 					$quantity = $context['tempQuantity'];
 					$add_to_cart_text = $context['addToCartText'];
+					$product_type = $context['productType'];
+
+					if ( 'grouped' === $product_type ) {
+						$product = wc_get_product( $context['productId'] );
```
Aljullu: Nit: not a strong opinion, but I think it would be clearer (and more performant?) to use the `$product` that has already been fetched in the upper function. So instead of calling `wc_get_product()` here again, we could update L104 to:

```diff
- 'addToCartText' => function () {
+ 'addToCartText' => function () use ( $product ) {
```

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
```diff
@@ -301,6 +318,58 @@ private function get_cart_item_quantities_by_product_id( $product_id ) {
...
+		);
+	}
+
+	/**
+	 * Prevent caching on certain pages
+	 */
+	private function prevent_cache() {
+		\WC_Cache_Helper::set_nocache_constants();
+		nocache_headers();
+	}
```
Aljullu: Do you have some more context on why this method is needed? :thinking: My understanding is that it's already provided by `BlocksSharedState`, no?

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
```diff
@@ -105,25 +105,37 @@ protected function render( $attributes, $content, $block ) {
...
+						$grouped_product_ids = $product->get_children();
+
+						foreach ( $grouped_product_ids as $child_product_id ) {
+							if ( $this->get_cart_item_quantities_by_product_id( (int) $child_product_id ) > 0 ) {
+								return __( 'Added to cart', 'woocommerce' );
+							}
+						}
+
+						return $add_to_cart_text;
+					}
```
Aljullu: We don't have clear guidelines on this, so just thinking out loud, but I would only display 'Added to cart' after the button is pressed. So if the products are already in the cart, but the grouped product single page has just been loaded, I wouldn't display 'Added to cart'.

In the way it's implemented right now, I feel it's a bit confusing. For example, here I only have 'T-Shirt' in my cart, but the whole grouped product shows 'Added to cart':

imatge

But definitely not a strong opinion neither a blocker. :slightly_smiling_face:

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
```diff
@@ -142,6 +143,119 @@ const { state, actions } = store< Store >(
 					actions.showNoticeError( error as Error );
 				}
 			},
+			*batchAddCartItems( items: OptimisticCartItem[] ) {
+				const previousCart = JSON.stringify( state.cart );
```
Aljullu: TIL about it, so not sure if it will work. But wondering if we could use [`structuredClone()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone) here.

So instead of:

```JS
const previousCart = JSON.stringify( state.cart );
...
state.cart = JSON.parse( previousCart );
```

We do:

```JS
const previousCart = structuredClone( state.cart );
...
state.cart = previousCart;
```

Aljullu (commented):
Thanks for all the updates, @mikachan! This is working great. I just found a small glitch that the _Added to cart_ text disappears after adding products:

https://github.com/user-attachments/assets/0d93b786-e672-4ed1-aace-4221b064fe3a

Can you reproduce as well?

Besides that, I left some inline comments, but they are mostly questions and suggestions.

Btw, it can be done in a follow-up, but we should add an e2e test to verify grouped products can be added to cart. We have similar ones for simple and variable products:

https://github.com/woocommerce/woocommerce/blob/4e08bee588c97aec594f0043b9f2c8e94e85a459/plugins/woocommerce/client/blocks/tests/e2e/tests/add-to-cart-with-options/add-to-cart-with-options.block_theme.spec.ts#L84

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
mikachan: It looks like the other nearby logic here for the error response also needs updating to handle the batch responses. I've done that in https://github.com/woocommerce/woocommerce/pull/58002/commits/fe24c5532e91c49480d73f94bb5742d11ea6f8a6, and I've created a new `BatchResponse` type for `json`.

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
mikachan: Nice! Done in https://github.com/woocommerce/woocommerce/pull/58002/commits/0d46065179f320f5dc10b51d326317750e6ecc40.

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
mikachan: Ah, looks like this was removed in https://github.com/woocommerce/woocommerce/pull/58265, so this must be from a merge conflict. I've removed it in https://github.com/woocommerce/woocommerce/pull/58002/commits/23e732a4b5ef574aec3d71229d5173c09a99e2fd.

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
mikachan: Nice, looks like that works and also TIL. Done in https://github.com/woocommerce/woocommerce/pull/58002/commits/d11b2db0feb0d4b7d6eccbedc2460105923184cc.

mikachan:
Thanks @Aljullu!

> I just found a small glitch that the Added to cart text disappears after adding products:

This happens if there is an unhandled error with the batch add to cart, and I think it's the same issue @dinhtungdu saw in [this comment](https://github.com/woocommerce/woocommerce/pull/58002#discussion_r2128122958). It's hopefully fixed now with the latest commits 🤞 

> Btw, it can be done in a follow-up, but we should add an e2e test to verify grouped products can be added to cart. 

I'll add some e2e tests as an immediate follow-up.

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
mikachan: I agree that this might be confusing. What do you think the text should be if all items are already in the cart? I'm thinking this might be the only case where it may already show 'Added to cart' on page load, but for all other cases it only changes the text after the button is pressed.

Aljullu: To me, the text being the regular _Add to cart_ even if all products are in cart would be good enough. I think grouped products are a special case and it's expected the button can't always represent the current cart state.

> I'm thinking this might be the only case where it may already show 'Added to cart' on page load, but for all other cases it only changes the text after the button is pressed.

That would work for me too. :slightly_smiling_face:

Aljullu (commented):
Confirming the _Added to cart_ text after adding to cart is fixed. I noticed a couple of glitches more:

1. When a product is sold individually, it gets added to cart even if the checkbox is unchecked.
2. For some reason, the `quantity` of the first product is ignored if there is a following product with a quantity higher than 1.

In this video, I'm trying to add 3 _Hoodie with Logo_, 0 _T-Shirts_ and 2 _Caps_. Instead, 1 _Hoodie with Logo_, 1 _T-Shirt_ and 2 _Caps_ are added.

https://github.com/user-attachments/assets/79b4ad6b-8e3b-4ddc-bcc3-cc4008764f21

Can you reproduce as well?

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/GroupedProductSelector.php
```diff
@@ -32,9 +32,31 @@ protected function render( $attributes, $content, $block ): string {
 		global $product;
 
 		if ( $product instanceof \WC_Product && $product->is_type( 'grouped' ) ) {
+			add_filter( 'woocommerce_product_supports', array( $this, 'check_product_supports_ajax_add_to_cart' ), 10, 3 );
```
Aljullu: @mikachan heads-up that https://github.com/woocommerce/woocommerce/pull/58588 has been merged. So now we will need to remove this hook.

mikachan: Thanks for the heads-up! I've removed that hook in https://github.com/woocommerce/woocommerce/pull/58002/commits/f92d03d7ff4bb294d9ed74fb82a3dde26bdf3b21.

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
mikachan: I've updated this so that 'Added to cart' is only shown when the add to cart button has been clicked; otherwise, 'Add to cart' will be used regardless of the products already in the cart: https://github.com/woocommerce/woocommerce/pull/58002/commits/37531bca3f843f7632ebee8a1a22b4b6ecc94189.

mikachan:
Thanks @Aljullu! 🙇 

> When a product is sold individually, it gets added to cart even if the checkbox is unchecked.
For some reason, the quantity of the first product is ignored if there is a following product with a quantity higher than 1.

I could reproduce both of these. For the first one, it looks like the Store API always adds a quantity of 1, even if 0 is sent. For now I've prevented quantities of 0 being sent to the API in https://github.com/woocommerce/woocommerce/pull/58002/commits/b6b785f7a23f89be1bb3a29dc63697905b5c9371, but it would be good for this to be fixed upstream too.

For the second one, I think this should be fixed with some improvements I made to `setQuantity` in https://github.com/woocommerce/woocommerce/pull/58002/commits/952524fd4b2037115cda51629519fff4a189f9f4.

plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php
```diff
@@ -301,6 +307,50 @@ private function get_cart_item_quantities_by_product_id( $product_id ) {
...
+					continue;
+				}
+				if ( $child_product->is_purchasable() && $child_product->is_in_stock() ) {
+					$is_product_purchasable = true;
+					break;
+				}
+			}
+		}
+
+		return $is_product_purchasable;
```
Aljullu: Not a big deal, but this could be simplified with early returns. I think it makes it easier to read and it should be a bit more performant:

```suggestion
		if ( $product->is_type( 'grouped' ) ) {	
			$grouped_product_ids = $product->get_children();
			foreach ( $grouped_product_ids as $child ) {
				$child_product = wc_get_product( $child );
				if ( ! $child_product instanceof \WC_Product ) {
					continue;
				}
				if ( $child_product->is_purchasable() && $child_product->is_in_stock() ) {
					return true;
				}
			}

			return false;
		}
		
		return $product->is_purchasable() && $product->is_in_stock();
```

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
```diff
@@ -77,9 +87,30 @@ const productButtonStore = {
...
+							( item ) => item.id === productId
+						);
+						return product?.quantity || 0;
+					}
+				);
+				if ( groupedProductIdsInCart?.some( ( qty ) => qty > 0 ) ) {
+					return hasPressedButton
+						? state.inTheCartTextGrouped
+						: state.inTheCartText;
+				}
```
Aljullu: In grouped products `inTheCartText` is going to have the same text as `addToCartText`, no? So I wonder if we could use that one instead.

If we do that, I think we could also get rid of `inTheCartTextGrouped` altogether and instead use `inTheCartText` to hold the `Added to cart` string (that would require changes in `ProductButton.php`).


```suggestion
				if (
					groupedProductIdsInCart?.some( ( qty ) => qty > 0 ) &&
					hasPressedButton
				) {
					return state.inTheCartText;
				} else {
					return addToCartText;
				}
```

Not a big deal, but I think it will help simplify the code because we will have one less variable to care about. Let me know if it doesn't make sense. :sweat_smile:

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
```diff
@@ -139,6 +170,10 @@ const productButtonStore = {
 				context.animationStatus = AnimationStatus.IDLE;
 			}
 		},
+		handleClick() {
```
Aljullu: Nit: I think that if we have a `handleClick()` action we should add the `addToCartItem()` call inside it. Otherwise, we could rename this `handleClick()` to `handlePressedState()` or something along those lines, to make it clear it doesn't handle all the callbacks after a click.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
```diff
@@ -150,6 +185,7 @@ const productButtonStore = {
 			// eslint-disable-next-line react-hooks/rules-of-hooks
 			useLayoutEffect( () => {
 				context.tempQuantity = state.quantity;
+				context.hasPressedButton = false;
```
Aljullu: Do you know why this is needed?

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
```diff
@@ -32,18 +39,26 @@ const { state: wooState } = store< WooCommerce >(
...
+		inputElement = event.target.parentElement?.querySelector(
+			'.input-text.qty.text'
+		);
+	}
 
-	const inputElement = target.parentElement?.querySelector(
-		'.input-text.qty.text'
-	) as HTMLInputElement | null;
+	if ( event.target instanceof HTMLInputElement ) {
+		inputElement = event.target.closest( 'input[type="checkbox"]' );
```
Aljullu: Do you know why we need to use `event.target.closest()` if `event.target` is already the input element?

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/AddToCartWithOptions.php
```diff
@@ -194,6 +214,31 @@ function ( $variation ) {
...
+					$child_product = wc_get_product( $child_product_id );
+					if ( $child_product && $this->is_child_product_purchasable( $child_product ) ) {
+						$context['groupedProductIds'][] = $child_product_id;
+					}
+				}
+
+				// Add quantity context for purchasable child products.
+				$context['quantity'] = array_fill_keys(
+					$context['groupedProductIds'],
+					$default_quantity
```
Aljullu, line 230: This can be done in a follow-up because it's like that in `trunk`, but I'm not sure why `$default_quantity` defaults to `1` and uses the `woocommerce_add_to_cart_quantity` filter (line 186). I think it should be instead:

```PHP
apply_filters( 'woocommerce_quantity_input_min', $product->get_min_purchase_quantity(), $product )
```

Because we want the default value to be the minimum that can be introduced in the input.

Aljullu (commented):
Thanks for the update, @mikachan! This is working great and I couldn't find any issues. I left some inline comments about coding changes, but they are not strong opinions, so I'm fine if there is any reason to leave things as they are.

> I could reproduce both of these. For the first one, it looks like the Store API always adds a quantity of 1, even if 0 is sent. For now I've prevented quantities of 0 being sent to the API in https://github.com/woocommerce/woocommerce/pull/58002/commits/b6b785f7a23f89be1bb3a29dc63697905b5c9371, but it would be good for this to be fixed upstream too.

Oooh, I wasn't aware of that, good catch! Could you create an issue for this so we don't forget?

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
mikachan: Yes, that works well, thanks! I was seeing 'Add to cart' persistently on the client-side, so I think that's why I created a new variable just for grouped products. But this is working well, thank you. Done in https://github.com/woocommerce/woocommerce/pull/58002/commits/d1122b0ac923ec0ed264f6359fd8b44f691309fa.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
mikachan: I like `handlePressedState()` - changed in https://github.com/woocommerce/woocommerce/pull/58002/commits/7c19e88e701e9fde3e570bfd42afce8fb06e5646.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
mikachan: I think this was because I was seeing the button context persist on initial load, so the add to cart text was wrong sometimes. But I'm not seeing this issue now (I'm wondering if it's related to the improvements from [here](https://github.com/woocommerce/woocommerce/pull/58002#discussion_r2135228142)), so I've removed it in https://github.com/woocommerce/woocommerce/pull/58002/commits/54b996a3217e73d50bcca9ecb38d670d5247bbc8.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
mikachan: I don't think we need `closest()` at all, I think this is left behind from me testing the different `event.target` results. I've removed this selector in https://github.com/woocommerce/woocommerce/pull/58002/commits/6a5be3fa0a5e9136bfdd003a4d16ceba78b5d03b, thanks!

mikachan:
Thanks @Aljullu! I think I've addressed all your recent feedback.

> Could you create an issue for this so we don't forget?

Yes! I've created an issue here: https://github.com/woocommerce/woocommerce/issues/58635.

Aljullu (approved):
Everything looks good on my end, thanks for all your work, @mikachan! :raised_hands: 

<details>

* Theme: Twenty Twenty-Five
* Store size: small
</details>

mikachan:
Thanks so much, @Aljullu!

Here is a list of follow-up tasks for when this is merged:

- [x] Add e2e tests for grouped products: https://github.com/woocommerce/woocommerce/pull/58727
- [x] Change `$default_quantity` to use minimum purchase quantity ([related comment](https://github.com/woocommerce/woocommerce/pull/58002#discussion_r2135302391)): https://github.com/woocommerce/woocommerce/pull/58729
- [x] Address issue with how `0` quantities are handled in batch Store API requests ([related issue](https://github.com/woocommerce/woocommerce/issues/58635)): https://github.com/woocommerce/woocommerce/pull/58658

mikachan:
Sorry to make more changes after an approval - I wasn't happy with how the `lastSuccessfulCartResponse` was handled in `cart.ts`, as it could still potentially be `undefined`. I've updated this so that an error is thrown if it's `undefined` (https://github.com/woocommerce/woocommerce/pull/58002/commits/27b90eea4045dcd53b1b269d003f5676eb335223), otherwise it will continue with the successful cart response as before.

I also made a small change to the way the quantities are handled for the `batchAddCartItems` request. There should be no change to the logic - I just noticed that some of the code was repeated, so I've moved it to a new function called `getNewQuantity ` (https://github.com/woocommerce/woocommerce/pull/58002/commits/f6fb8751244a9f63fef64aae87a667197fc89595).

mikachan:
[This test failure](https://github.com/woocommerce/woocommerce/actions/runs/15539858498/job/43747861698?pr=58002) seems unrelated to this PR - it looks like it's perhaps related to [a flaky test](https://github.com/woocommerce/woocommerce/pull/58593), and it seems to be failing on other PRs as well. It's passing for me locally!

mikachan:
I have a potential fix for the test here: https://github.com/woocommerce/woocommerce/pull/58661

PR #57880, opened 2025-05-09
Core API tests - update flaky refund tests

mikejolley:
### Submission Review Guidelines:

-   I have followed the [WooCommerce Contributing Guidelines](https://github.com/woocommerce/woocommerce/blob/trunk/.github/CONTRIBUTING.md) and the [WordPress Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/).
-   I have checked to ensure there aren't other open [Pull Requests](https://github.com/woocommerce/woocommerce/pulls) for the same update/change.
-   I have reviewed my code for [security best practices](https://developer.wordpress.org/apis/security/).
-   Following the above guidelines will result in quick merges and clear and detailed feedback when appropriate.

### Changes proposed in this Pull Request:

Cause of the flakiness is not 100% clear but I noticed that the tests are written in a way that require them to be ran in sequence, and based on the logs (https://github.com/woocommerce/woocommerce/actions/runs/14775338381/job/41482508330) it looks like they can run in parallel. So to address the issue I've tweaked the tests to run independently of one another so each test creates an order, does the refund, then cleans up after itself.

Closes #57667

### How to test the changes in this Pull Request:

N/A. CI should pass (Core API Tests).

### Testing that has already taken place:

Running CI locally, but I couldn't get them to fail.

### Changelog entry

-   [ ] Automatically create a changelog entry from the details below.

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Details</summary>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

-   [ ] Fix - Fixes an existing bug
-   [ ] Add - Adds functionality
-   [ ] Update - Update existing functionality
-   [ ] Dev - Development related task
-   [ ] Tweak - A minor adjustment to the codebase
-   [ ] Performance - Address performance issues
-   [ ] Enhancement - Improvement to existing functionality

#### Message 

</details>

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

PR #58021, opened 2025-05-15
Core API tests - fix flaky refund tests (cherry pick #57880)

adimoldovan:
Cherry picks https://github.com/woocommerce/woocommerce/issues/57880 into 9.9

## Testing instruction

CI should be green.
The failures where happening with HPOS off, which is not checked by CI. Tested this locally and tests passed.

```
pnpm test:api:hpos
```


jorgeatorres (approved):
LGTM.

PR #58023, opened 2025-05-15
Add initial CheckoutProcessor test setup and tests

tjcafferkey:

This PR adds initial test setup and some basic test coverage for the `<CheckoutProcessor />` provider. This provides the basis for further test coverage and iterations.

1. Ensure test cases are valid and correctly setup/executed.
2. Ensure CI passes

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


straku:
I very much appreciate the effort, but I am worried that the amount of mocking won't set us up for success in continuing to increase test coverage for this area. This might be fuelled by my personal vendetta against mocks (that I would like to believe is informed by my previous experiences), but I am sure there's some objective merit to it as well.

I am approaching it from the "setting up for success" angle, as my problem with mocking is maintainability:
- Mocking of this level is tightly coupled with implementation details. This often leads to test failures during refactoring that aren’t caused by actual bugs, but by mocks no longer aligning with reality. I don't think adding tests should make us more reluctant to refactor, ideally, it should have the opposite effect.
- This might be subjective, but I find tests with heavy mocking harder to reason about, which makes me more hesitant to add or modify tests.

What I would prefer instead is for the component to work with real stores that are powered by mock data (ideally reused from a real test store, rather than handcrafted for each test). We can't avoid mocks entirely, but we can target lower-level APIs – think `fetch`, `@wordpress/api-fetch`, and anything else that is not realistic to rely on in our current test environment.

tjcafferkey:
I really appreciate your feedback and concerns @straku, I think you've touched on a few different aspects here so Im going to try and address each one separately.

**Heavy mocking is tightly coupled to implementation details, leading to test failures during refactoring that aren’t caused by bugs:**

I agree that excessive mocking can tie tests to implementation details, potentially making refactoring harder. However, the `CheckoutProcessor` is a complex provider that handles multiple stores (checkout, payment, validation) and external dependencies (`@wordpress/api-fetch`, payment methods, etc.). The current level of mocking is necessary to isolate the component’s logic for unit testing, ensuring we can verify its behavior without external dependencies. For example, mocking `hasValidationErrors` and `isProcessing` allows us to test how the component handles validation errors or processing states, which are critical to its functionality. Without this, we’d struggle to simulate edge cases like validation failures or API errors in a controlled way.

**Heavy mocking makes tests harder to reason about, reducing willingness to modify or add tests.**

I understand that mocks can make tests less readable, and I share the goal of making tests approachable. The mocks in this PR are structured to mirror the component’s dependencies (e.g., WordPress data stores, API fetch), with clear setups in `beforeEach` to reset state and provide defaults. This aims to make the test’s intent - verifying the component’s rendering and order processing logic - clearer. That said, I’m open to exploring ways to simplify the mock structure.

**Use real stores powered by mock data from a test store.**

I see the appeal of using real stores with mock data to reduce reliance on high-level mocks. However, I’m concerned about the reliability of external test data. If we depend on a shared test store, changes to that data (e.g., values, schema updates, new fields) could inadvertently break our tests, even if the component’s logic remains correct. This could introduce flakiness, undermining test stability. Additionally, using real stores might shift us away from unit testing toward integration testing, requiring a full environment setup, which could slow down test execution.

**Tests should encourage refactoring, not hinder it.**

I completely agree that tests should support refactoring. The current tests focus on the component’s public behavior - rendering without crashing and skipping order processing when validation fails - rather than internal implementation details. By mocking dependencies, we can test the component in isolation, ensuring that refactors (e.g., changing how stores are accessed) don’t break core functionality. If we find specific mocks are too brittle during refactoring, we can revisit them.

I’m very open to exploring alternative methods, such as reusable mock utilities or mock data factories, to address maintainability concerns while preserving test stability. Could you take another look at the PR in light of these points along with their wider usage in the codebase? I’d also love to discuss any specific ideas you have for reducing mock complexity or improving our testing strategy to set us up for success.

tjcafferkey:
We should also consider that since `CheckoutProcessor` is being used in production now does this approach improve or reduce confidence?

Issue #57995, opened 2025-05-14
Remove the Shipping Calculation Section in the Cart Block


mikejolley:
We will remove the "shipping calculation" section from the cart block and instead show text stating that shipping will be added during checkout.

See discussion: [https://github.com/woocommerce/woocommerce/discussions/57456](https://github.com/woocommerce/woocommerce/discussions/57456#discussioncomment-12920937)

### **Problem**

[The discussion ](https://github.com/woocommerce/woocommerce/discussions/57456#discussion-8229997)summarises the problems with the current approach. To summarise:

* There are inconsistencies between cart and mini-cart totals areas
* Having shipping calculator/address form on cart is complex and difficult to maintain alongside checkout
* Partial addresses do not return accurate rates and can confuse customers
* Presentation of local pickup alongside shipping rates is also confusing
* The additional steps and confusion might delay customers from proceeding to checkout, negatively affecting conversion rates. 

### Proposal

* The shipping calculation section will be removed
* Some text will be added to state that shipping will be added during checkout. This will be customisable in the editor.
* Settings UI will be updated to state that certain settings only apply to the shortcode

Unknown items to be worked out during implementation, or things that need design input @yoann.jailin 

* The exact text used stating that shipping will be added during checkout
* What totals are shown on cart—will need to look into the technical side to see if shipping total can be excluded from the order total, or if that needs to also be left out (just show subtotal)
* Migration to avoid breakage

Styling will follow existing patterns.

Ertuit:
Hello,
I dont agree at all with the removal of this. This is quite a essential feature and it reduce pain for international shipping or country/region where shipping cost is the same across large areas. Actually, I think it need more options. I would really like that the customer just has to choose his country and is offered shipping cost (instead of entering their city and postal code when its irrelevant).

To solve the "misleading result" **we can put a range. (ex : " ~$10 to $25"),**
The "not accurate" and "Rate Presentation" argument is wrong as local pickup is always free (I have yet to see a merchant who charge for local pickup) so a price calculation, any price calculation, is irrelevant for this as no matter what, the customer know that it will be free anyway.

I think a better way to do this is to **allow displaying range of shipping cost.** And then refuse to display a subtotal/total of the order.
Also, when two delivery methods are present, (ex : local pickup and parcel shipping) put a selector with a radio to choose between them and adjust the shipping cost.

The biggest no for me is the Conversion Impact. Actually **mysterious shipping cost is the first reason of cart abandonment.**
[...]

PR #58030, opened 2025-05-15
Update address JS utils & remove filter

ralucaStan:

- - This PR converts the `addressFieldsForShippingRates` setting introduced in #57374 to a JS constant kept in the address utils file.

-- - This PR also removes the `woocommerce_address_fields_for_shipping_rates` filter introduced in #57374. The filter is not released, so no deprecation is needed. This was an optimistically added filter. We're not accounting for it in the shipping calculator (which is pending removal https://github.com/woocommerce/woocommerce/issues/57995) and the WC shipping logic doesn't account for additional shipping fields. It was only intended for front-end use in the checkout block to trigger loading states. If a need arises we'd need to create a holistic approach at the core level

- This PR updates the address utils to account for `addressFieldsForShippingRates` setting introduced in #57374 instead of the hardcoded values for `[ 'state', 'country', 'postcode', 'city' ]`.

- The `isAddressComplete` function was renamed and refactored to use the shipping fields setting; we're not using it for checking random fields, it was used only for the fields from addressFieldsForShippingRates.

This PR does not account for the hardcoded variables on the PHP side, this would be handled in another PR.

#### a. No shipping rates defined for a complete address
1. Define shipping rates only for one country
2. Checkout with a product, and select another country
3. Notice that checkout shows a message that there are no shipping options
4. Notice the order summary show a message about missing delivery
5. Fill in the shipping address, select shipping option and checkout

#### b. Guest user with no address
1. Set in WooCommerce > General >  Default customer location > No location by default
2. Go to a guest checkout and notice the shipping area and the Order summary message prompting to enter an address
Screenshot 2025-05-15 at 16 03 43
3. Fill in the shipping address, select shipping option and checkout

### Shipping calculator
Testing for this is covered by this file https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/tests/e2e/tests/cart/cart-checkout-block-shipping.block_theme.spec.ts

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/types/type-defs/cart-response.ts
```diff
@@ -163,6 +163,11 @@ export interface CartResponseExtensionItem {
 	[ key: string ]: unknown;
 }
 
+export interface ShippingAddressFieldsForShippingRates
+	extends Omit< ResponseBaseAddress, 'address_1' | 'address_2' > {
+	[ key: string ]: string;
+}
+
```
opr: I don't think this type is correct because it's saying that `ShippingAddressFieldsForShippingRates` is an object of strings, keyed by strings, whereas it is an array of strings. We should probably use the keys of `ResponseBaseAddress`without `address_1` and `address_2`.

e.g. 
```suggestion
export type ShippingAddressFieldsForShippingRates = keyof Omit<
	ResponseBaseAddress,
	'address_1' | 'address_2'
>;
```

plugins/woocommerce/src/Blocks/BlockTypes/Cart.php
```diff
@@ -252,6 +252,8 @@ protected function enqueue_data( array $attributes = [] ) {
 		$this->asset_data_registry->add( 'collectableMethodIds', $local_pickup_method_ids );
 		$this->asset_data_registry->add( 'shippingMethodsExist', CartCheckoutUtils::shipping_methods_exist() > 0 );
 
+		$this->asset_data_registry->add( 'addressFieldsForShippingRates', [ 'state', 'country', 'postcode', 'city' ] );
```
opr: What is the purpose of adding this to the asset_data_registry? If it's a static value, and only consumed on the front end, should it just be kept as a static value in `address.ts` ? We can remove the PHP tests checking for this value too if we do that.

opr (changes requested):
Thanks for working on this, I tested and it's working fine. I agree with removing the filters for the address fields, but I think it's safe to stop defining this in PHP if we're not filtering it. Do you agree?

If so, I can re-test it once the `addressFieldsForShippingRates` is defined in the front-end.

plugins/woocommerce/src/Blocks/BlockTypes/Cart.php
ralucaStan: You're right. I've moved it to the address utilities. Thank you for catching that.

plugins/woocommerce/client/blocks/assets/js/types/type-defs/cart-response.ts
ralucaStan: Nice catch! I've based it on BaseAddress, as I've learned that the response types are for data coming from Store API. See https://github.com/woocommerce/woocommerce/pull/58030/commits/f8200fec0b381e135c50465ea3a4b78a38125975

ralucaStan:
@opr thank you for your review! I've addressed your code changes. There is an email E2E test failing, but I've reported it and it should be fixed soon.

opr (approved):
Thanks Raluca, looks good now!

PR #55879, opened 2025-02-26
[Experimental] Moves schema based `rules` for additional fields to the top level

mikejolley:
### Submission Review Guidelines:

-   I have followed the [WooCommerce Contributing Guidelines](https://github.com/woocommerce/woocommerce/blob/trunk/.github/CONTRIBUTING.md) and the [WordPress Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/).
-   I have checked to ensure there aren't other open [Pull Requests](https://github.com/woocommerce/woocommerce/pulls) for the same update/change.
-   I have reviewed my code for [security best practices](https://developer.wordpress.org/apis/security/).
-   Following the above guidelines will result in quick merges and clear and detailed feedback when appropriate.

### Changes proposed in this Pull Request:

Moves schema based `rules` for additional fields to the top level (required, hidden, validation). Updates client, server, and tests.

Closes #54875

### How to test the changes in this Pull Request:

Covered by unit tests and behind a feature flag.

Developers can test with an experimental build and a field snippet:

```php
add_action(
	'woocommerce_init',
	function () {
		$fields = array(
			array(
				'id'                => 'plugin-namespace/gov-id',
				'label'             => 'Government ID',
				'location'          => 'address',
				'type'              => 'text',
				'required'          => true,
				'attributes'        => array(
					'title'          => 'This is a gov id',
					'autocomplete'   => 'gov-id',
					'autocapitalize' => 'none',
					'maxLength'      => '30',
				),
				'sanitize_callback' => function ( $value ) {
					return trim( $value );
				},
				'validate_callback' => function ( $value ) {
					return strlen( $value ) > 3;
				},
			),
			array(
				'id'       => 'plugin-namespace/job-function',
				'label'    => 'What is your main role at your company?',
				'location' => 'contact',
				'required' => true,
				'type'     => 'select',
				'options'  => array(
					array(
						'label' => 'Director',
						'value' => 'director',
					),
					array(
						'label' => 'Engineering',
						'value' => 'engineering',
					),
					array(
						'label' => 'Customer Support',
						'value' => 'customer-support',
					),
					array(
						'label' => 'Other',
						'value' => 'other',
					),
				),
			),
			array(
				'id'       => 'plugin-namespace/leave-on-porch',
				'label'    => __( 'Please leave my package on the porch if I\'m not home', 'woocommerce' ),
				'location' => 'order',
				'type'     => 'checkbox',
			),
			array(
				'id'         => 'namespace/vat-number',
[...]

PR #58035, opened 2025-05-15
Prevent hidden fields being validated during checkout

opr:

- If a field is marked as hidden in locale, validation will be skipped. This prevents fields that are required, but hidden from blocking the checkout.
- This PR also sets the newly added `cart_context` when hydrating cart/checkout routes on non-store API requests.
- Update `OrderController` tests to use a wrapper class to allow testing a protected method.

Closes WOOPLUG-4295

(For Bug Fixes) Bug introduced in PR: possibly https://github.com/woocommerce/woocommerce/pull/55879/

| Before | After |
| ------ | ----- |
| https://www.loom.com/share/7b0597cfc83d456d836b5ebc72e762a1 | https://www.loom.com/share/c1537499447741ecaef7fe38d68d64da) |

1. Add this snippet
```php
add_filter( 'woocommerce_get_country_locale', function( $locale ){
	$locale['GB']['postcode']['hidden'] = true;
	$locale['GB']['postcode']['required'] = true;
	return $locale;
} );
```

2. Check out using a UK address, postcode should be hidden and you should be able to check out without errors.
3. Check out using a different address, e.g. one in US. Do not fill ZIP Code, ensure it shows, and is required. Ensure an error is shown when trying to check out without entering ZIP code.
4. Change the snippet to:
```php
add_filter( 'woocommerce_get_country_locale', function( $locale ){
	$locale['GB']['postcode']['required'] = false;
	return $locale;
} );
```
5. Check out using a UK address, ensure you can check out without entering a postcode.
6. Repeat these tests on the Shortcode checkout.

#### Testing with additional fields (blocks only)

1. Install and activate https://github.com/woocommerce/additional-checkout-fields-tester
2. Check out ensuring the required fields from that plugin are required, and the optional ones do not block checkout when not given a value.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/wooplug-4295-hidden-but-required-fields-prevent-checkout
+++ b/plugins/woocommerce/changelog/wooplug-4295-hidden-but-required-fields-prevent-checkout
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Prevent hidden fields being considered during validation
--- a/plugins/woocommerce/includes/class-wc-customer.php
+++ b/plugins/woocommerce/includes/class-wc-customer.php
@@ -311,6 +311,16 @@ class WC_Customer extends WC_Legacy_Customer {
 				continue;
 			}
 
+			// If the field is hidden in the country-specific locale, we can skip it.
+			if ( isset( $country_locale[ $key ]['hidden'] ) && true === wc_string_to_bool( $country_locale[ $key ]['hidden'] ) ) {
+				continue;
+			}
+
+			// Check if the field is hidden in the default locale, if so, we can skip too (because it wasn't hidden in country-specific locale).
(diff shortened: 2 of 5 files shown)


opr:
Closing because I think this needs a deeper look when it comes to additional fields. I think we should prevent a field going hidden if it's required. I will update the issue to mention this.

senadir:
@opr in additional fields we made the decision that hidden takes precedence over required so I think you can proceed with that

opr:
Thanks @senadir good to know, glad I reopened!

senadir (approved):
This is testing fine but I wonder why we didn't include this check in our refactor, I can't recall anything, @mikejolley any idea why we don't check for hidden fields, only required (for core ones)?

mikejolley:
@senadir Was it it because they are not posted or rendered? I recall doing some filtering when getting groups of fields.

Issue #57918, opened 2025-05-12
Woocommerce Session Cookie blocked by AWS WAF


yberges-hiberus:
### Prerequisites

- [x] I have carried out troubleshooting steps and I believe I have found a bug.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

When implementing the AWS WAF with AWS-AWSManagedRulesSQLiRuleSet, the website starts being blocked by AWS WAF. On the "Sampled Requests" feature on AWS, I can see the requests are being blocked because of "AWS#AWSManagedRulesSQLiRuleSet#SQLi_COOKIE". 

Cleaning the cookies and refreshing the site gets me a web with errors (no css, js or other static file is loaded because 403 errors). When inspecting the cookies I can only see the "cokieyes" cookie and "wp_woocommerce_session" cookie. If I try to refresh the site, I get a 403 again.

Editting the value of the "wp_woocommerce_session" cookie and removing the pipes (in URLencode '%7C') fixes the problem (when removing the character "%" from the URLencodec way also makes the trick).

I've found an issue related with cookies and WAF problems (https://github.com/woocommerce/woocommerce/issues/43413), but that fix does not seem to apply to my case since the described cookies on the fix are not the ones related with my problem.

### Expected behavior

The expected behavior is to not get blocked by the WAF without disabling or allowing AWS SQLi from cookies.

### Actual behavior

The WAF is blocking all requests since the "wp_woocommerce_session" cookie is in every request.

### Steps to reproduce

1. Install a woocommerce site and enable wp_woocommerce_session cookies.
2. Configure AWS WAF in front of the woocommerce site.

### WordPress Environment
WooCommerce 9.8.4
WordPress 6.7.1

### Isolating the problem

- [x] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [ ] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/products/storefront/).
- [x] I can reproduce this bug consistently using the steps above.

nerrad:
Have you tried creating a [custom AWS WAF rule](https://docs.aws.amazon.com/waf/latest/developerguide/waf-rules.html) to exclude the cookie from the SQLi ruleset?

yberges-hiberus:
That's what I currently have configured on the WAF.

![Image](https://github.com/user-attachments/assets/067fd7dc-952c-4340-8f43-212c21ed09a9)

[...]

PR #58039, opened 2025-05-15
Fix: Update session cookie delimiter handling in MockSessionHandler

juanjoache:
# Implementation of Unit Tests for `MockSessionHandler`

## Ticket Reference  
- [#57918 - Woocommerce Session Cookie blocked by AWS WAF](https://github.com/woocommerce/woocommerce/issues/57918)

## Description  
This update addresses the issue of WooCommerce session cookies being flagged by AWS WAF (`AWSManagedRulesSQLiRuleSet`) due to the usage of pipe characters (`||`).  
To resolve this, the cookie delimiter has been modified to a single pipe (`|`).

## Changes  
- Modified the cookie delimiter in the `MockSessionHandler` class from `||` to `|`.  
- Added a filter `woocommerce_session_cookie_delimiter` to allow customization of the delimiter.  
- Implemented unit tests to verify both the default and custom delimiter configurations.  
- Executed PHPUnit tests with successful outcomes.

## Testing  
- PHPUnit executed with the following command:  

  ```bash
  phpunit --filter MockSessionHandlerTest

- All tests passed successfully with no errors or warnings.

## Impact
The delimiter change resolves the conflict with AWS WAF but introduces a potential backward compatibility consideration for existing cookies using the old || delimiter.

Additional logic can be implemented to handle both delimiters during the transition period if needed.

Example of New Filter Implementation:


```php
add_filter('woocommerce_session_cookie_delimiter', function() {
    return '|';
});
```

## Relevant Files Modified:
- tests/php/src/Blocks/StoreApi/MockSessionHandler.php
- composer.json
- composer.lock
- bin/post-update.sh
- bin/composer/phpcs/composer.lock

### DIFF
--- a/plugins/woocommerce/tests/php/src/Blocks/StoreApi/MockSessionHandler.php
+++ b/plugins/woocommerce/tests/php/src/Blocks/StoreApi/MockSessionHandler.php
@@ -155,9 +155,16 @@ class MockSessionHandler extends WC_Session_Handler {
 	 */
 	public function set_customer_session_cookie( $set ) {
 		if ( $set ) {
-			$to_hash           = $this->_customer_id . '|' . $this->_session_expiration;
+			$delimiter = apply_filters('woocommerce_session_cookie_delimiter', '||');
+			$to_hash           = $this->_customer_id . $delimiter . $this->_session_expiration;
 			$cookie_hash       = hash_hmac( 'md5', $to_hash, wp_hash( $to_hash ) );
-			$cookie_value      = $this->_customer_id . '||' . $this->_session_expiration . '||' . $this->_session_expiring . '||' . $cookie_hash;
+			$cookie_value = implode($delimiter, [
+				$this->_customer_id,
+				$this->_session_expiration,
+				$this->_session_expiring,
+				$cookie_hash
+			]);
+
 			$this->_has_cookie = true;
 
 			if ( ! isset( $_COOKIE[ $this->_cookie ] ) || $_COOKIE[ $this->_cookie ] !== $cookie_value ) {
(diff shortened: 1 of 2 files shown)


juanjoache:
@mikejolley @nerrad 
I've implemented the fix for the session cookie delimiter issue as discussed in #57918. 
The changes are isolated to `MockSessionHandler.php`, and test files. 

Could you confirm who should be tagged as reviewers for this change? 
Or, is there a specific team or group responsible for validating this type of update?

nerrad (commented):
This pull request is just updating mocks used in tests?

juanjoache:
> This pull request is just updating mocks used in tests?

Yes, the primary change in this PR is updating the MockSessionHandler class to handle the cookie delimiter through a new filter woocommerce_session_cookie_delimiter

This update is aimed at resolving the AWS WAF issue (reported in https://github.com/woocommerce/woocommerce/issues/57918) where the || delimiter was flagged by the AWSManagedRulesSQLiRuleSet. The new implementation changes the default delimiter to a single | and includes unit tests to verify both default and custom delimiters.

Let me know if further clarification is needed or if additional adjustments are required.

mikejolley:
Hi @juanjoache I don't think this addresses https://github.com/woocommerce/woocommerce/issues/57918. The mock handler is only used by unit tests. The actual cookie class needs to be updated to use a single pipe. I'll be looking at this later in the week if you do not have a chance to get to it, but thanks for raising a PR.

nerrad:
Closing because as Mike points out this doesn't address the issue. Besides, we've shipped a separate fix.

PR #58044, opened 2025-05-15
MIC: Introduce Email Templates for Back In Stock Notifications

xristos3490:

Follow up on https://github.com/woocommerce/woocommerce/pull/57062

This PR is part of a broader effort to introduce Back In Stock (BIS) notifications. It adds the necessary email templates used to inform customers when a previously out-of-stock product is restocked.

#### Email Templates
Three new template files have been introduced:
- `customer-stock-notification.php`: This is the default template for sending back-in-stock notifications to customers.
- `customer-stock-notification-verify.php`: This template is used to confirm back-in-stock notifications from customers through an OTP.
- `customer-stock-notification-confirm.php`: This template is employed to confirm the subscription after successful verification.

Each of these also includes plain templates that correspond to them.

Screenshot 2025-05-19 at 7 28 09 PM

#### Specification

This PR introduces two controllers: EmailsController and TemplatesController for better management of email-related hooks and actions. The EmailsController will manage email hooks, while the TemplatesController will oversee all front-facing template hooks related to this feature. Currently, I have included the sections that are printed within the email templates.

Additionally, this PR simplifies and restructures the existing templates in the extension to align with the lean format currently used in emails. Specifically:
- Nine templates in the extension (six HTML and three plain) have been consolidated into six (three HTML and three plain).
- Sanity checks previously included in each template are removed; instead, this logic will be incorporated into each flow utilizing the emails.
- Most importantly, BIS emails are no longer categorized as "transactional emails". Triggering will occur within the EmailsController using dedicated methods, rather than the previous "[hook]_notification" _(via the `woocommerce_email_actions` hook) approach for transactional emails. This change is essential because these emails do not fit the transactional category*.
_*: This also fixes an issue where a store is set to "[defer transactional emails](https://github.com/woocommerce/woocommerce/blob/45b8a66cdd899c3ef0d96d8d06567b82f93a92de/plugins/woocommerce/includes/class-wc-emails.php#L127)," and having a dedicated "background mailer" for BIS notifications could result in a backlog of background jobs to handle._

There are some conceptual uncertainties I had, which I described below under the "open questions" section. 

#### Open questions

- Currently, all `WC_Email` classes reside within internal namespaces, and their instances are directly created in the WC()->mailer emails property. I’m wondering whether we should incorporate these email singletons into the container.
- Having `WC_Email`s in the Internal namespace breaks the `WC()->mailer`'s _[classname => instance]_ format. Adding the complete namespace as key isn't feasible because of the way the ["sections" are connected](https://github.com/woocommerce/woocommerce/blob/8da790103a4f66abe6c45db5492a0ac6fe0cdd84/plugins/woocommerce/includes/admin/settings/class-wc-settings-emails.php#L456) on the settings page. The only viable alternative would be to relocate the `WC_Emails` to the legacy "includes" folder.

- Navigate to **WooCommerce > Settings > Emails**
- Ensure both HTML and plain versions render correctly.
- Manually trigger a BIS notification using the preview.
- Test overrides by placing modified versions in the active theme folder.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Internal/StockNotifications/EmailsController.php
```diff
@@ -0,0 +1,282 @@
...
+	/**
+	 * Registers custom emails classes.
+	 *
+	 * @param array $emails Array of email classes.
+	 * @return array
+	 */
+	public function email_classes( $emails ) {
+		$emails['WC_Email_Stock_Notification_Receive'] = new StockNotificationEmail();
+		$emails['WC_Email_Stock_Notification_Confirm'] = new StockNotificationConfirmEmail();
+		$emails['WC_Email_Stock_Notification_Verify']  = new StockNotificationVerifyEmail();
```
xristos3490: **Note to the reviewer:** The "classname => instance" structure is broken here. The reason is that I cannot add a namespace in the emails key due to the admin settings, which [escape this key as a URL parameter](https://github.com/woocommerce/woocommerce/blob/8da790103a4f66abe6c45db5492a0ac6fe0cdd84/plugins/woocommerce/includes/admin/settings/class-wc-settings-emails.php#L456), leading to a 404 error.

This makes me wonder if we need to have the WC_Emails under the legacy namespace in the `includes/emails` folder.

plugins/woocommerce/src/Internal/StockNotifications/Notification.php
```diff
@@ -29,7 +29,7 @@ class Notification extends \WC_Data {
 	 *
 	 * @var \WC_Product
 	 */
-	private $product;
+	public $product;
```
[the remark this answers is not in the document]
xristos3490: Ignore

plugins/woocommerce/src/Internal/StockNotifications/TemplatesController.php
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	/**
+	 * Add template hooks.
+	 */
+	public function add_email_template_hooks() {
+		add_action( 'woocommerce_email_stock_notification_product', array( $this, 'email_product_image' ), 10, 3 );
+		add_action( 'woocommerce_email_stock_notification_product', array( $this, 'email_product_title' ), 20, 3 );
+		add_action( 'woocommerce_email_stock_notification_product', array( $this, 'email_product_attributes' ), 30, 3 );
+		add_action( 'woocommerce_email_stock_notification_product', array( $this, 'email_product_price' ), 40, 3 );
```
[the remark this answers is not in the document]
xristos3490: Ignore. The hook does indeed support the WC_Email as a fourth argument, but the templates here don't need to know that.

xristos3490:
Neat! The copilot uncovered [a typo](https://github.com/woocommerce/woocommerce/pull/58044/commits/47ae6b8ecafeb1481fbeb196fb8ab20332e896b1) that's been hanging around in the extension for years—it's hiding in the "low confidence" comments! 🧌 cc @PanosSynetos

plugins/woocommerce/src/Internal/StockNotifications/EmailsController.php
```diff
@@ -0,0 +1,282 @@
...
+		return $emails;
+	}
+
+	/**
+	 * Restore customer data from notification's metadata, if applicable.
+	 *
+	 * @param  Notification $notification The notification object.
+	 * @return void
+	 */
+	public function maybe_restore_customer_data( $notification ) {
```
jimjasson: I needed to do some head scratching to understand what this function does, and I would like some confirmation if my assumption is correct:

When a product goes back in stock, its price is listed in the notification e-mail. As products may have different taxes depending on the customer location, you would like to show the price based on the customer's location to avoid any price differences when customers click on the action link. 

Is that right? 

If so, then can we rephrase the comments and the function name to make this more clear? Perhaps `maybe_restore_customer_tax_location_data`?

plugins/woocommerce/src/Internal/StockNotifications/Notification.php
```diff
@@ -368,4 +368,92 @@ public function save() {
...
+	 * @return string
+	 */
+	public function get_product_formatted_variation_list( $flat = false, $context = 'view' ) {
+
+		$product = $this->get_product();
+		if ( ! $product ) {
+			return false;
+		}
+
+		$formatted_variation_list = wc_get_formatted_variation( $product, $flat, true, true );
```
jimjasson: Can you provide some more details as to why this needs to run for non-variation products?

`wc_get_formatted_variation` is always called with a `WC_Product` object. If this object is not a variation, then `wc_get_formatted_variation` will attempt to run this code: https://github.com/woocommerce/woocommerce/blob/d5c53efc7ba616ccdd0c4a956ceb98ede07736d5/plugins/woocommerce/includes/wc-product-functions.php#L499-L509. 

However, since the function argument is not an array: https://github.com/woocommerce/woocommerce/blob/d5c53efc7ba616ccdd0c4a956ceb98ede07736d5/plugins/woocommerce/includes/wc-product-functions.php#L504 but a `WC_Product` object, this function won't do anything. Right?

plugins/woocommerce/src/Internal/StockNotifications/Emails/StockNotificationEmail.php
```diff
@@ -0,0 +1,301 @@
...
+/**
+ * Back in stock notification email class.
+ */
+class StockNotificationEmail extends WC_Email {
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		$this->id             = 'stock_notification_receive';
```
jimjasson: I am more in favor of using `stock_notification_received`, as we had in the extension. `receive` indicates that some further action is required, as in `confirm`.

plugins/woocommerce/templates/emails/customer-stock-notification-confirm.php
```diff
@@ -0,0 +1,95 @@
...
+		 * Hook: woocommerce_email_stock_notification_product.
+		 *
+		 * @since 0.0.0
+		 *
+		 * @hooked \Automattic\WooCommerce\Internal\StockNotifications\Templates::email_product_image - 10
+		 * @hooked \Automattic\WooCommerce\Internal\StockNotifications\Templates::email_product_title - 20
+		 * @hooked \Automattic\WooCommerce\Internal\StockNotifications\Templates::email_product_attributes - 30
+		 * @hooked \Automattic\WooCommerce\Internal\StockNotifications\Templates::email_product_price - 40
+		 */
+		do_action( 'woocommerce_email_stock_notification_product', $product, $notification, $plain_text, $email );
```
jimjasson: Good call to consolidate these!

jimjasson (commented):
Hey @xristos3490, 

This PR is very well thought-through in terms of architecture!

I left some comments is some parts of the code -- see below. 

I also compared the back-in-stock e-mails functionality we are introducing here with the one that the Back In Stock Notifications plugin had and I noticed some differences: 
- **Triggering e-mail notifications**: It is not clear to me how these e-mail notifications are triggered with the code in this PR, but I guess that this will be part of a different PR, right? For example, where is the `send_stock_notification_email` called from.
- **Force send notifications**: The plugin had a way to force send an e-mail notification even if not all conditions had been met. Will this happen in core at a different point perhaps?
- **Validating notifications before they are sent**: The plugin validated the state of the notification before sending out the e-mail, for example if the notification is still active or already verified. I didn't find any similar validations in this PR -- will this perhaps happen at the point where the notifications are triggered? 
- In general, I noticed that a lot of filters that used to exist in the plugin have been consolidated or removed. I think that the usage of filters in the plugin was a bit excessive -- for example, I noticed that there used to be a filter even for changing which dummy product would be used in e-mail previews. Should we perhaps document somewhere all the filters that have been deprecated or removed, as an advisory for the developers who migrate from the plugin to the core feature? 
----------------------------------

Then, I tested all the e-mails both in the preview and the actual e-mail. My tests included both plain e-mails and HTML e-mails. I also tested all the available placeholders and confirmed that the results are the same as with the e-mails generated by the plugin. 

| E-mail type | Screenshot |
|--------|--------|
| Received email - HTML - Preview | Core - Received - HTML - Preview |
| Received email - HTML - Mail | Core - Received - HTML - Mail |
| Received email - Plain - Preview | Core - Received - Plain - Preview |
| Received email - Plain - Mail | Core - Received - Plain - Mail |
| Confirm email - HTML - Preview | Core - Confirm - HTML - Preview | 
| Confirm email - HTML - Mail | Core - Confirm - HTML - Mail |
| Confirm email - Plain - Preview | Core - Confirm - Plain - Preview |
| Confirm email - Plain - Mail | Core - Confirm - Plain - Mail |
| Verify email - HTML - Preview | Core - Verify - HTML - Preview |
| Verify email - HTML - Mail | Core - Verify - HTML - Mail |
| Verify e-mail - Plain - Preview | Core - Verify - Plain - Preview |
| Verify e-mail - Plain - Mail | Core - Verify - Plain - Mail |

plugins/woocommerce/src/Internal/StockNotifications/Notification.php
xristos3490: Thanks for this! This was taken directly from the extension, but it needs some love. 

1. We only need variations. You are right. 
2. The idea is that it accepts a variation _or_ an array of "cart"-item data to display in a formatted way. The current `wc_get_formatted_variation()` function is kinda convoluted, and honestly, I feel the same way about this function in the Notifications. The more I look at this, the more it makes sense as a "template function helper" rather than a model method.

For now, I have refactored this for sanity and performance and written a few unit tests [here](https://github.com/woocommerce/woocommerce/pull/58044/commits/cc8c3b87742c4997539b12b67a08e838478d2d7f). Let me know what you think! Definitely room for improvement on a follow-up!

plugins/woocommerce/src/Internal/StockNotifications/EmailsController.php
xristos3490: Nice! Done in https://github.com/woocommerce/woocommerce/pull/58044/commits/18665724fb89e66119a71ea7fcdf51f2e4ea0be4

plugins/woocommerce/src/Internal/StockNotifications/Emails/StockNotificationEmail.php
xristos3490: As discussed, an idea would be:
1. `customer_stock_notification`
2. `customer_stock_notification_verify`
3. `customer_stock_notification_confirm`


Thinking outside the box (=extension), this "confirm" as the "confirmation with no CTA" feels somewhat off. An alternative might be:
1. `customer_stock_notification`
2. `customer_stock_notification_confirm` _(The current verify)_
3. `customer_stock_notification_confirmed` _(The current confirm)_

xristos3490:
Thanks for this thorough review, @jimjasson! 

> It is not clear to me how these e-mail notifications are triggered with the code in this PR, but I guess that this will be part of a different PR, right? For example, where is the send_stock_notification_email called from.

Yup! But I did make a few more changes [here](https://github.com/woocommerce/woocommerce/pull/58044/commits/97486ee98f50a7a25b703508ed73499f894ba981). 

The idea is that each email would be triggered by calling a method (within the AS later on):
```php
$EmailManager->send_stock_notification_email( Notification $notification );
```

The verification and confirmation emails will be sent during stream processing (e.g., through the signup flow). To allow for [deferring these emails](https://github.com/woocommerce/woocommerce/blob/0a25b2f633d57f71d4dfd1f86922b2430e32dad5/plugins/woocommerce/includes/class-wc-emails.php#L127), they need to fall under Woo's "transactional" emails, meaning they would trigger as follows:
```php
do_action( 'woocommerce_customer_stock_notification_confirm`, $notification );
```

> Missing:
> - Force send notifications: 
> - Validating notifications before they are sent:

Absolutely! To keep things consistent and maintain a clear separation of concerns, I've simplified the email logic from the extension quite a bit. That being said, as we implement the logic that triggers the emails, we must remember to transfer anything we still need from the extension carefully or have them side by side while reviewing. 😊

> In general, I noticed that a lot of filters that used to exist in the plugin have been consolidated or removed.

Yes. To be honest, the extension has quite a few hooks. I'm focusing on the really important ones for now, but we can definitely revisit DX later on.

When it comes to documenting what might be missing, I'm a bit uncertain. There are quite a few unnecessary hooks, and I worry that a list like this could confuse folks. Perhaps starting fresh with new documentation on stock notifications would be clearer! Thoughts?

plugins/woocommerce/src/Internal/StockNotifications/Notification.php
jimjasson: Nice! Thanks for adding the unit test too, it makes it easier to understand the different flows.

plugins/woocommerce/src/Internal/StockNotifications/EmailsController.php
jimjasson: Thanks @xristos3490! Looking good! 

I think that we can still improve the docblock for this function to clearly state why this restoration is needed. 

E.g.: Restore customer tax location data from notification's metadata to display product prices in emails using the customer's tax location, if applicable.

plugins/woocommerce/src/Internal/StockNotifications/Emails/StockNotificationEmail.php
jimjasson: Nice -- I think verify/verified would also work here, up to you!

jimjasson:
Thanks for the explanations and the changes @xristos3490! 

> There are quite a few unnecessary hooks, and I worry that a list like this could confuse folks. Perhaps starting fresh with new documentation on stock notifications would be clearer! Thoughts?

I agree that the new documentation shouldn't include any legacy information from the plugin. 

However, I imagine that we will write a developer blog post about migrating from the Back In Stock Notifications plugin to the core feature. It would be nice to have a section there explaining which filters and settings have been removed and if there are any alternatives that can be used.

plugins/woocommerce/src/Internal/StockNotifications/Emails/StockNotificationEmail.php
xristos3490: >  I think verify/verified would also work here, up to you!

Yup, I like this better 🚀

plugins/woocommerce/src/Internal/StockNotifications/EmailsController.php
xristos3490: Thanks! Done in https://github.com/woocommerce/woocommerce/pull/58044/commits/4a570337d6b443bb3a56f0d0e9d2b971ad81de9e

plugins/woocommerce/src/Internal/StockNotifications/Emails/StockNotificationEmail.php
xristos3490: Done in https://github.com/woocommerce/woocommerce/pull/58044/commits/4a570337d6b443bb3a56f0d0e9d2b971ad81de9e and https://github.com/woocommerce/woocommerce/pull/58044/commits/7929c05b7ddda365a0b4e0331e760e6026531f1b

xristos3490:
> However, I imagine that we will write a developer blog post about migrating from the Back In Stock Notifications plugin to the core feature. It would be nice to have a section there explaining which filters and settings have been removed and if there are any alternatives that can be used.

Oh, you're right, @jimjasson! We need to post one about the migrator plugin; we can definitely experiment with a list of replacements here. Thanks for bringing this up!

Just FYI, for the review, I made [another change](https://github.com/woocommerce/woocommerce/pull/58044/commits/d92e193439d9a1ba116dccacadfebb7e09a3eb6f) to the `get_formatted_variation_list` and repurposed it as a simple wrapper for the existing `wc_get_formatted_variation()` that will also provide the notification's posted attributes during the sign-up process. Essentially, I moved the "email-rendering" context outside this function and integrated it into the EmailTemplatesController, where it belongs. This should hopefully simplify the data model.

Let me know what you think! 🙌

jimjasson (approved):
Thank you for addressing all the comments above @xristos3490 and doing some additional improvements, such as separating the logic of creating and displaying variation product titles in different classes to be more context specific. 

I re-tested the 3 e-mails types after the latest changes and confirmed that there aren't any issues with previewing the e-mails and sending them. Moreover, the code looks good -- even more readable now.

PR #58045, opened 2025-05-15
Fix: Incorrect shipping notice when using Local Pickup with multiple packages

### LINKED ISSUE #57858: Shipping notice shown incorrectly when using local pickup and multiple packages
## **Describe the bug**

hen using  [Multiple Packages for WooCommerce](https://wordpress.org/plugins/multiple-packages-for-woocommerce/) and Local Pickup, switching from Local Pickup back to "Ship" results in a notice being shown incorrectly.

Go to WooCommerce -> Settings -> Multiple Packages and enable it. Set "Group by" to "Product (Individual).

## **To Reproduce**

1. Install [Multiple Packages for WooCommerce](https://wordpress.org/plugins/multiple-packages-for-woocommerce/)
2. Go to WooCommerce -> Settings -> Multiple Packages and enable it. Set "Group by" to "Product (Individual).
3. Go to WooCommerce -> Settings -> Shipping -> Local Pickup. Enable local pickup and add some locations.
4. Go to WooCommerce -> Settings -> Shipping and add some shipping rates. Add two.
5. Go to the front-end, add an item to your cart and go to the Checkout block.
6. Switch to Local Pickup.
7. Switch back to shipping. Observe the notice above the order summary.

## **Expected behavior**

When switching from Local Pickup to shipping, the notice should not appear, since the packages are both using a correct shipping rate. This notice was intended for use when switching from Local Pickup to Delivery *when the address is not filled in, but shipping rates require it.*

Technically, the local pickup method is still selected because no shipping rate is available to be selected, and there is a mismatch between what the user has chosen (Shipping) vs. what is actually chosen (Local pickup).

## **Screenshots**

[Shipping notice shown incorrectly when using local pickup and multiple packages](https://www.loom.com/share/2cbc37da07e641788a57d38c839479f8)

## Environment

Please fill out any of the below that applies

**Operating System/Device**

- [ ] Windows
- [ ] MacOS
- [ ] Linux
- [ ] iPhone
- [ ] Samsung
- [ ] Other (add in additional context field)

**Version:** 

**Browser**

- [ ] Desktop Edge
- [ ] Desktop Chrome/Chromium
- [ ] Desktop Safari
- [ ] Mobile Edge
- [ ] Mobile Chrome/Chromium
- [ ] Mobile Safari
- [ ] Other (add in additional context field)

**Version:**

BurakParsAydin:

Closes #57858 .

1. Please refer to the issue above and follow the steps there while on the trunk branch.
2. Then, switch to this branch, and you should see that the notice no longer appears.

- Keep in mind that you'll need to run the following command to rebuild the JavaScript files after switching branches: 
`pnpm --filter='@woocommerce/plugin-woocommerce' build:blocks`

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

Hide shipping warning when valid rates are available after switching from Local Pickup.

</details>

### DIFF
--- a/plugins/woocommerce/changelog/fix-57858
+++ b/plugins/woocommerce/changelog/fix-57858
@@ -1,0 +1,4 @@
+Significance: minor
+Type: fix
+
+Hide shipping warning when valid rates are available after switching from Local Pickup.
--- a/plugins/woocommerce/client/blocks/assets/js/base/context/hooks/use-show-shipping-total-warning.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/base/context/hooks/use-show-shipping-total-warning.ts
@@ -38,10 +38,11 @@ export const useShowShippingTotalWarning = () => {
 				( notice: WPNotice ) => notice.id === errorNoticeId
 			);
 		const hasMismatch =
-			!prefersCollection &&
+			! prefersCollection &&
 			hasSelectedLocalPickup &&
 			shippingRates?.every(
-				(rate) => rate.method_id && rate.method_id.includes('local_pickup')
+				( rate ) =>
+					rate.method_id && rate.method_id.includes( 'local_pickup' )
 			);
 
 		if ( ! hasRates || isRateBeingSelected ) {
--- a/plugins/woocommerce/client/blocks/assets/js/base/context/hooks/use-show-shipping-total-warning.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/base/context/hooks/use-show-shipping-total-warning.ts
@@ -37,7 +37,12 @@ export const useShowShippingTotalWarning = () => {
 			shippingNotices.some(
 				( notice: WPNotice ) => notice.id === errorNoticeId
 			);
-		const hasMismatch = ! prefersCollection && hasSelectedLocalPickup;
+		const hasMismatch =
+			!prefersCollection &&
+			hasSelectedLocalPickup &&
+			shippingRates?.every(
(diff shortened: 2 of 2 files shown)


opr (changes requested):
Hi @BurakParsAydin thanks for opening this, I don't think it quite fixes the issue, for example if you have the following setup:

- No fallback shipping rates
- Rates only in Netherlands
- Local pickup enabled
- No customer location by default

image

- Add an item to cart and go to Checkout, select Local pickup and then switch back to shipping
- Shopper is on "Shipping", but notice in the order summary, local pickup is showing
- This is when the notice should show, since the "real" selected method does not match the UI.

I think the actual bug is somewhere in `useShippingData` related to the way `hasSelectedLocalPickup` is calculated. Might be a good place to start looking.

BurakParsAydin:
@opr Thanks for the review. I couldn’t reproduce the issue with the setup you described. However, what do you think about adjusting the code like this:

```
const hasMismatch =
	! prefersCollection &&
	hasSelectedLocalPickup &&
	shippingRates?.length > 0 &&
	shippingRates.every(
		( rate ) =>
		rate?.selected && rate.method_id.includes( 'local_pickup' )
	);
```

This strengthens the check by ensuring that shipping rates exist and that each selected rate is specifically for local pickup.
Do you think this would fully address the issue?

ralucaStan:
Closing as the bug isn't replicable anymore. Thank you @BurakParsAydin for trying to fix this one.

PR #58053, opened 2025-05-16
Fix: WooCommerce Blocks Storybook bugs

### LINKED ISSUE #51785: [Epic] WooCommerce Blocks Storybook bugs
In the WooCommerce blocks Storybook instance there are some issues that I found while testing:

- [x] `CheckboxControl` - can’t click to toggle the checkbox
- [ ] `RemovableChip` - can't remove the chip when pressing the `x`
- [ ] `RadioControlAccordion` and `RadioControl` - styling seems to be missing or just needs extra padding (See image 1)
- [ ] `TextInput` - comes with an error by default (not sure if intended, but would be good to investigate)
- [ ] `ValidatedTextInput` - can’t get error state to show, the “With error” variation also doesn’t work
- [ ] `Title` - changing heading level doesn’t affect heading size
- [ ] `CountryInput` - docs page goes off the screen (See image 2). Also label seems incorrect, should be "Select a country".
- [ ] `QuantitySelector` - the `editable` prop should be on by default. Also add a new variation so we have two versions:  editable/not editable

#### Image 2
image

amitraj2203:

This PR addresses several UI component issues in the WooCommerce Blocks Storybook implementation. The changes fix various functionality and styling problems to improve the developer experience when using the Storybook components.

Specifically, the PR:

- **RadioControlAccordion**: Created a style.scss file and added styles for top and bottom padding similar to RadioControl
- **RemovableChip**: Fixed functionality to make the chip properly removable and added Disabled and RemoveOnAnyClick variations
- **TextInput**: Fixed default error state display and added a WithHelp variation
- **Title**: Fixed heading size to properly change when modifying the heading level tags prop in Storybook
- **CountryInput**: Changed the label from "Select a countries" to "Select a country"
- **QuantitySelector**: Set the editable prop to true by default and added new variations: NonEditable, WithMinimum, and WithMaximum
- **ValidatedTextInput**: Fixed the showError prop functionality and repaired the "With error" variation

Closes #51785 

| Component | Image |
| ------- | ------ |
| RadioControlAccordion | RadioControlAccordion after fix |
| Title | Title after fix |
| Countrylnput | Countrylnput after fix |
| QuantitySelector | QuantitySelector after fix |
| ValidatedTextInput | ValidatedTextInput after fix |

1. Run the Storybook instance with `npm run storybook` and go to `http://localhost:6006/`
2. Test the fixed components:
   - Verify `RadioControlAccordion` displays with proper padding
   - Check that `RemovableChip` can be removed by clicking the 'x'
   - Confirm `TextInput` no longer shows errors by default
   - Verify `Title` component changes font size when changing heading levels
   - Check `CountryInput` displays with the correct `Select a country: ` label
   - Verify `QuantitySelector` is editable by default and new variations work correctly
   - Test `ValidatedTextInput` error state displays correctly and the `With error` variation works now

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/packages/components/radio-control-accordion/style.scss
```diff
@@ -0,0 +1,12 @@
+.wc-block-components-radio-control-accordion-option {
```
opr: This affects the actual block on the front end, and makes spacing too big.

I'd recommend investigating ig moving this to a file called `style.stories.css` and importing this file in the radio-control-accordion _stories_ file rather than the main component solves this.

image

plugins/woocommerce/client/blocks/assets/js/base/components/quantity-selector/stories/index.stories.tsx
```diff
@@ -15,6 +15,7 @@ export default {
 	args: {
 		itemName: 'widgets',
 		quantity: 1,
+		maximum: 10,
```
opr: Why did we need to add the maximum here? The default component should have no maximum, and the `WithMaximum` variant adding it should be fine right?

plugins/woocommerce/client/blocks/packages/components/chip/stories/removable-chip.stories.tsx
```diff
@@ -21,13 +22,38 @@ export default {
...
-);
+const Template: StoryFn< RemovableChipProps > = ( args ) => {
+	const [ isVisible, setIsVisible ] = useState( true );
+	const handleRemove = () => {
+		setIsVisible( false );
+	};
+
+	if ( ! isVisible ) {
+		return <></>;
+	}
```
opr: What do you think of allowing the chip to be restored like so?

```suggestion
	if ( ! isVisible ) {
		return (
			<button onClick={ () => setIsVisible( true ) }>
				<em>Chip was removed, click to reset</em>
			</button>
		);
	}
```

opr (changes requested):
Thanks a lot @amitraj2203 looking great so far except a few snags, please let me know if any of my comments are unclear.

By the way I also noticed the `With error` variation of the `ValidatedTextInput` does not show the value after typing, would you be willing to tackle that bug in this PR too? 🙏🏼

plugins/woocommerce/client/blocks/packages/components/chip/stories/removable-chip.stories.tsx
amitraj2203: Sure

plugins/woocommerce/client/blocks/assets/js/base/components/quantity-selector/stories/index.stories.tsx
amitraj2203: My bad, it shouldn't be there, removed it now.

plugins/woocommerce/client/blocks/packages/components/radio-control-accordion/style.scss
amitraj2203: Thanks for the suggestion! Just to clarify, when you mentioned moving this to a file called `style.stories.css`, did you mean renaming `style.scss` to `style.stories.css`?

Also, I tried importing this stylesheet in `radio-control-accordion/stories/index.stories.tsx`, and that does seem to resolve the issue. Let me know if that’s what you had in mind!

amitraj2203:
> By the way I also noticed the With error variation of the ValidatedTextInput does not show the value after typing

I see, the issue was with all the variations present in the Docs page.
Screenshot 2025-05-21 at 1 23 22 AM
I tried to fix it in [40471b1](https://github.com/woocommerce/woocommerce/pull/58053/commits/40471b1758a5a681c321c4d269c017ec43e01d0b) could you please review it again? Thanks!

plugins/woocommerce/client/blocks/packages/components/radio-control-accordion/style.scss
opr: Yes, I did mean renaming it to `style.stories.css` just to make sure it's clear that these styles are specific to the storybook rendering. If we call it just `style.css` then it may mislead folks thinking this applies on the front-end.

Thanks!

plugins/woocommerce/client/blocks/packages/components/title/style.scss
```diff
@@ -1,7 +1,6 @@
 // Extra class for specificity to overwrite editor styles.
 .wc-block-components-title.wc-block-components-title {
 	@include reset-box();
-	@include font-size(large);
```
opr: Why did we remove this? We should be careful about changing CSS files that are consumed on the front end. This change causes the titles on checkout to be too large. Could you add it back please?

opr (changes requested):
Thanks for looking at those things I mentioned! A couple more comments, but looks almost ready to merge 👍🏼

plugins/woocommerce/client/blocks/packages/components/title/style.scss
amitraj2203: Apologies for removing this, I removed it to solve the issue of:
`Title - changing heading level doesn’t affect heading size`.

But since we shouldn't make changes to the `style.scss` I think we can make a `style.stories.css` similar to the `RadioControlAccordion` and add the following styles in it. Let me know if that sounds good.
```scss
.wc-block-components-title {
	&[class*="h1"] {
		font-size: 2em;
	}
	&[class*="h2"] {
		font-size: 1.5em;
	}
	&[class*="h3"] {
		font-size: 1.17em;
	}
	&[class*="h4"] {
		font-size: 1em;
	}
	&[class*="h5"] {
		font-size: 0.83em;
	}
	&[class*="h6"] {
		font-size: 0.67em;
	}
}
```

Also just wanted to let you know that naming the file `style.stories.css` does not working but `style.stories.scss` works. So would it be good to keep it a SCSS file?

opr: Sounds good! If it helps the storybook look cleaner I say go for it. The naming is fine too, I just made a mistake in the file extension 😄 

Please double check the different stories and front-end after making this change to ensure the changes improve the storybook and also don't affect the front-end.

Thanks!

amitraj2203: Thanks for the review. I've created `style.stories.scss` for both `Title` and `RadioControlAccordion`, and imported that file only in the _stories_ file. I've also tested it and can confirm that it does not affect the front end.

opr (approved):
@amitraj2203 this looks good now, really impressive work. I've just tested the titles and they are working nicely.

Thanks again for your contribution, it's greatly appreciated!

PR #58066, opened 2025-05-16
Remove the shipping calculation section in the cart block

### LINKED ISSUE #57995: Remove the Shipping Calculation Section in the Cart Block

mikejolley:

This is a major refactor of the cart block which removes the shipping calculator section and rate selector in favour of showing only a summary of estimated totals. 

Reasoning behind this change can be found here:

- https://github.com/woocommerce/woocommerce/issues/57995
- https://github.com/woocommerce/woocommerce/discussions/57456#discussioncomment-12920937

The main goals of this PR were to remove the complexity of showing a mix of local pickup and shipping rates in the cart, and remove confusing UIs in favour of pushing customers to the checkout where accurate shipping costs can be displayed.

**Legacy/shortcode flow is untouched.** What was touched before by filters is now excluded due to cart_context. Local pickup gets no special handling in the legacy cart/checkout.

**Hide shipping costs until an address is entered** is no longer restricted by local pickup. Local pickup rates are no longer auto selected as default, and since the cart does not display rates we don't need to impose restrictions. Settings UI updated. 

Closes #57995

No address, hide shipping costs before an address is provided:

Screenshot 2025-05-16 at 11 30 24

Showing shipping costs, or going back to cart after selecting shipping at checkout:

Screenshot 2025-05-16 at 11 31 14

As above but with a coupon applied:

Screenshot 2025-05-16 at 11 31 23

There are quite a few different flows involved here so extensive testing is required using different configurations. The main settings that affect this feature are:

1. Your shipping zone/shipping methods
2. Local Pickup on or off
3. "Hide shipping costs before an address is entered" on or off
4. Guest vs customer session

#### Guest flow (incognito session)

Before testing:
- Enable "Hide shipping costs before an address is entered"
- Ensure you have several zones of rates and local pickup enabled
- Block checkout as primary checkout

1. Add item to cart, go to cart page
2. Confirm the totals area shows only the estimated total—no shipping, no subtotal
3. Apply a coupon. Confirm subtotal and discount amount is displayed. Remove coupon when done.
4. Go to page containing cart shortcode. Check no rates are displayed there either.
5. Go to checkout. 
     - Confirm order summary area shows correct values, shipping should prompt for address
     - **Confirm on entry to checkout, Local Pickup was not selected. You should be on the shipping form with no shipping selection. This was a related fix.**
6. Select local pickup
7. Back to the cart, now a shipping row appears with local pickup as the row name.
8. Back to checkout, enter address, select other shipping method as normal.
9. Back to cart, the shipping method you chosen is shown in the totals area

Now turn off "Hide shipping costs before an address is entered" and start a new incognito session.

1. Add item to cart.
2. Go to cart. A shipping row and subtotal row exists. One of the shipping methods is preselected.
3. As admin, edit the matching shipping zone and disable shipping rates for that zone.
4. Back as guest, refresh. Notice no shipping is shown. Local pickup is not auto selected.
5. Go to checkout. You are on the address tab, not the local pickup tab, and can continue as normal with checkout

#### Customer flow (logged in)

Similar process as above (smoke test), but note that if your account has an address, shipping rates are selected and shown on cart when "Hide shipping costs before an address is entered" is on. This is expected—you have an address.

#### Other configurations/edge cases

Smoke test with multiple shipping packages extension. This will impact the name of the shipping row in cart and checkout. If there are multiple packages the row will be called "Shipping" with selected rates listed below.

Screenshot 2025-05-16 at 13 02 15

Check the settings area. I've updated the settings to not mention local pickup, and to disable the shipping calculator setting when using block cart:

Screenshot 2025-05-16 at 12 24 45

Check for regressions in legacy cart and ensure our block specific logic is not running there:

Screenshot 2025-05-16 at 11 31 39

Test other related flows you feel may be impacted that I may have missed.

Above steps, checking for regressions in cart, but there are other edge cases that others may test.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/base/utils/shipping-rates.ts
```diff
@@ -122,11 +136,15 @@ export const getTotalShippingValue = ( values: {
 export const getSelectedShippingRateNames = (
 	shippingRates: CartShippingRate[]
 ): string[] => {
-	return shippingRates.flatMap( ( shippingPackage ) => {
-		return shippingPackage.shipping_rates
-			.filter( ( rate ) => rate.selected )
-			.flatMap( ( rate ) => rate.name );
-	} );
+	return Array.from(
```
mikejolley, line 137: (add docblock) Return unique values in the array

plugins/woocommerce/client/blocks/assets/js/blocks/cart/inner-blocks/cart-order-summary-fee/block.tsx
```diff
@@ -5,8 +5,14 @@ import { TotalsFees, TotalsWrapper } from '@woocommerce/blocks-components';
 import { getCurrencyFromPriceResponse } from '@woocommerce/price-format';
 import { useStoreCart } from '@woocommerce/base-context/hooks';
 
-const Block = ( { className }: { className: string } ): JSX.Element => {
+const Block = ( { className }: { className: string } ): JSX.Element | null => {
```
mikejolley: Can this be inferred?

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
```diff
@@ -70,7 +72,9 @@ const TotalsFooterItem = ( {
 
 	const label = applyCheckoutFilter( {
 		filterName: 'totalLabel',
-		defaultValue: __( 'Total', 'woocommerce' ),
+		defaultValue: isCart
+			? __( 'Estimated total', 'woocommerce' )
+			: __( 'Total', 'woocommerce' ),
 		extensions: cart.extensions,
 		arg: { cart },
```
senadir, line 83: Argument should include context for this label (cart/checkout).

senadir:
A note to whoever will test this is to test it with Subscpritons and possibly booking or deposit, since they filter out the total label value and now it's different.

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
ralucaStan: This is what we show in the Mini-Cart. I think we should stick with `Subtotal` instead of `Estimated total`
Screenshot 2025-05-26 at 16 37 53

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
```diff
@@ -123,20 +127,29 @@ const TotalsFooterItem = ( {
...
+										value={ parsedTaxValue }
+									/>
+								),
+							} ) }
+						</p>
+					) }
+					{ isCart && (
+						<p className="wc-block-components-totals-footer-item-shipping">
+							{ __(
+								'Shipping will be calculated at checkout',
```
ralucaStan, line 154: Small change: Mini-Cart says: Shipping, taxes, and discounts calculated at checkout.

We should be consistent and use `Shipping calculated at checkout.` Actually, we should use delivery instead of shipping:  `Delivery calculated at checkout`.

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/test/__snapshots__/index.tsx.snap
```diff
@@ -77,7 +77,7 @@ exports[`TotalsFooterItem Shows the "including %s TAX LABEL" line with single ta
       <p
         class="wc-block-components-totals-footer-item-tax"
       >
-        Including £1.00 10% VAT
```
ralucaStan: We'll have to replace these snapshot tests with some component unit tests

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
```diff
@@ -111,7 +115,7 @@ const TotalsFooterItem = ( {
 						} )
```
ralucaStan: Although not brought in by this PR I think we should stick with the price being displayed last, to match the structure of the whole order summary table. I added a nice-to-have version of the area with modified title, and texts to match the Mini Cart
Screenshot 2025-05-26 at 17 07 49

plugins/woocommerce/client/blocks/tests/e2e/tests/checkout/checkout-block.shopper.block_theme.spec.ts
```diff
@@ -225,11 +225,9 @@ test.describe( 'Shopper → Local pickup', () => {
 			'page=wc-settings&tab=shipping&section=options'
 		);
 
-		await expect(
```
ralucaStan, line 230: there are some utils that need to be removed from the e2e tests https://github.com/woocommerce/woocommerce/blob/e1fa42b054d0586b3e799e02ff02fa6d2cebed68/plugins/woocommerce/client/blocks/tests/e2e/utils/shipping/shipping-utils.page.ts#L27-L53

plugins/woocommerce/client/blocks/assets/js/blocks/cart/style.scss
```diff
@@ -25,6 +25,18 @@
...
+		border-bottom: 1px solid $universal-border-light;
+		margin-bottom: $gap;
+
+		.wc-block-components-totals-item,
+		.wc-block-components-panel,
+		.wc-block-components-totals-coupon {
+			padding-left: 0;
+			padding-right: 0;
+		}
+	}
```
ralucaStan, line 35: there is some 	css for `.wc-block-components-shipping-calculator` in this file

ralucaStan: Some tests also mention this class wp-block-woocommerce-shipping-calculator, so you should remove it from there as well

ralucaStan:
Some documentation & comments will need to be updated as well: 

- https://github.com/woocommerce/woocommerce/blob/e1fa42b054d0586b3e799e02ff02fa6d2cebed68/docs/cart-and-checkout-blocks/overview-of-data-flow.md#L132-L134
- multiple mentions of the calculator like here https://github.com/woocommerce/woocommerce/blob/e1fa42b054d0586b3e799e02ff02fa6d2cebed68/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/form/use-form-fields.ts#L28  
- also some cleanup might be needed here https://github.com/woocommerce/woocommerce/blob/e1fa42b054d0586b3e799e02ff02fa6d2cebed68/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/form/form.tsx#L95-L96

plugins/woocommerce/client/blocks/assets/js/blocks/cart/inner-blocks/cart-order-summary-totals/style.scss
```diff
@@ -1,6 +1,12 @@
 .wp-block-woocommerce-cart-order-summary-totals-block {
 	border-top: 1px solid $universal-border-light;
 	padding-bottom: $gap;
+
+	&:empty,
```
ralucaStan, line 9: a comment might help with understanding the case this selectors cover

plugins/woocommerce/tests/e2e-pw/tests/api-tests/settings/settings-crud.test.js
```diff
@@ -1244,65 +1244,6 @@ test.describe( 'Settings API tests: CRUD', () => {
 					} ),
 				] )
 			);
-			expect( responseJSON ).toEqual(
```
ralucaStan, line 1247: I don't think these should be removed.

plugins/woocommerce/src/StoreApi/Utilities/CartController.php
```diff
@@ -28,13 +28,11 @@ class CartController {
 	 * Makes the cart and sessions available to a route by loading them from core.
 	 */
 	public function load_cart() {
-		if ( did_action( 'woocommerce_load_cart_from_session' ) ) {
```
ralucaStan, line 31: what is the purpose for these changes?

plugins/woocommerce/src/StoreApi/Schemas/V1/CartSchema.php
```diff
@@ -338,11 +337,8 @@ public function get_item_response( $cart ) {
 		// Get cart errors first so if recalculations are performed, it's reflected in the response.
 		$cart_errors = $this->get_cart_errors( $cart );
 
-		// Get shipping packages to return in the response from the cart. Always get the shipping packages if local
-		// pickup is enabled and has methods. If the address is required then regular shipping rates will be filtered
-		// out later.
-		$has_local_pickup_methods = LocalPickupUtils::is_local_pickup_enabled() && count( LocalPickupUtils::get_local_pickup_method_ids() ) > 0;
-		$shipping_packages        = $cart->has_calculated_shipping() || $has_local_pickup_methods ? $controller->get_shipping_packages() : [];
+		// Get shipping packages to return in the response from the cart.
+		$shipping_packages = $cart->has_calculated_shipping() ? $controller->get_shipping_packages() : [];
```
ralucaStan, line 341: are we keeping these for backwards compatibility?

ralucaStan (commented):
I took a high level look at the PR and tested a bit the shopper experience. I still need to do a deep dive into the shipping rates logic and order summary changes and test with some extensions as well. 

I am ending my day soon, so I'll leave some of the findings from today and continue tomorrow with a more in depth review.

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/shipping/index.tsx
```diff
@@ -27,20 +29,24 @@ export const TotalsShipping = ( {
 	collaterals = null,
 }: TotalShippingProps ): JSX.Element | null => {
 	const { cartTotals, shippingRates } = useStoreCart();
-	const hasRates = hasShippingRate( shippingRates );
+	const hasSelectedRates = hasSelectedShippingRate( shippingRates );
```
ralucaStan, line 32: shipping rates throws a TS warning here

plugins/woocommerce/client/blocks/assets/js/types/type-defs/cart-response.ts
```diff
@@ -66,14 +66,7 @@ export interface CartResponseShippingPackageShippingRate
...
 
-export interface CartResponseShippingRate {
-	/* PackageId can be a string, WooCommerce Subscriptions uses strings for example, but WooCommerce core uses numbers */
-	package_id: number | string;
-	name: string;
-	destination: ResponseBaseAddress;
-	items: Array< ShippingRateItem >;
-	shipping_rates: Array< CartResponseShippingPackageShippingRate >;
-}
+export type CartResponseShippingRate = CartShippingRate;
```
ralucaStan, line 69: Curious why you exported CartResponseShippingRate? I was a little confused by the fact that we have 2 cart shipping rates types before reaching this line

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
ralucaStan: When there is a selection for the shipping rate in checkout and I navigate back we show the same text. `Shipping will be calculated at checkout`.

Screenshot 2025-05-27 at 11 17 52

Screenshot 2025-05-27 at 11 24 51


The Mini-Cart ignores completely the shipping selection in the summary, so there the text still makes sense.

We have 2 options:
- we change the text: Shipping can be changed at checkout
- or we don't display the selected shipping option

Something to also discuss with @yjailin .

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
```diff
@@ -39,6 +39,7 @@ export interface TotalsFooterItemProps {
 	 */
 	values: LooselyMustHave< CartResponseTotals, 'total_price' | 'total_tax' >;
 	className?: string;
+	isCart?: boolean;
```
ralucaStan: I'm thinking we could rename this `isCart` for better clarity  to `isEstimatedTotal` or `isSubtotal` if we change it. Or we could say `isCompleteTotal` or `isFinalizedTotal` and change the logic a bit

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
mikejolley: @ralucaStan its different in the context of mini cart because in mini cart that is the subtotal (product prices combined). In the cart we're factoring in data we have such as discounts and shipping (if already selected). This was suggested by design, its in figma—Ill DM you the link because its not shared publically.

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
mikejolley: We have to display the selected option unless we want to manually do calculations in the client, which I think we should avoid. If shipping is selected, the "total" price will include the shipping price _on the server_.

I personally think its fine to keep the line saying shipping will be calculated at checkout because thats your only opportunity to recalculate or reselect a different shipping method. I left this in intentionally.

plugins/woocommerce/tests/e2e-pw/tests/api-tests/settings/settings-crud.test.js
mikejolley: You can't see in the diff but these tests are duplicated a few lines prior.

plugins/woocommerce/src/StoreApi/Utilities/CartController.php
mikejolley: 1. Exit early
2. It prevented the context `$cart->cart_context = 'store-api';` being set if the cart was already initialised. That was a bug.

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
mikejolley: I think I prefer the natural sentence structure in place now—but you could raise this as a separate issue for feedback from design. We can revert this back to `Including` if it reduces scope.

plugins/woocommerce/src/Blocks/Shipping/ShippingController.php
```diff
@@ -222,13 +203,13 @@ public function show_local_pickup_details( $return_value, $order ) {
...
+		if ( CartCheckoutUtils::is_cart_block_default() ) {
 			foreach ( $settings as $index => $setting ) {
-				if ( 'woocommerce_shipping_cost_requires_address' === $setting['id'] ) {
-					$settings[ $index ]['desc'] = sprintf(
+				if ( 'woocommerce_enable_shipping_calc' === $setting['id'] ) {
+					$settings[ $index ]['desc_tip'] = sprintf(
 					/* translators: %s: URL to the documentation. */
-						__( 'Hide shipping costs until an address is entered (Not available when using the <a href="%s">Local pickup options powered by the Checkout block</a>)', 'woocommerce' ),
-						'https://woocommerce.com/document/woocommerce-blocks-local-pickup/'
+						__( 'This feature is not available when using the <a href="%s">Cart and checkout blocks</a>. Shipping will be calculated at checkout.', 'woocommerce' ),
```
ralucaStan, line 211: Let's replace Shipping with Delivery here as well. It work no matter the available shipping options or local pickup settings

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
senadir: and mini cart one is not filtered to my knowledge

plugins/woocommerce/src/StoreApi/Schemas/V1/CartSchema.php
mikejolley: We're keeping this because shipping packages are included in the API response. We've removed the local pickup handling here; we don't need to manipulate the response any longer.

plugins/woocommerce/src/Blocks/Shipping/ShippingController.php
mikejolley: Delivery sounds so wrong to me, I know we mix things up on block checkout but its still `shipping` on legacy checkout at least so I think we can keep this as `shipping`.

I would actually prefer we call it shipping everywhere with the exception of the Delivery toggle section on checkout where you choose between "Shipping" and "Collection". In general, "collection" isn't _calculated_ anyway.

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
mikejolley: `isEstimate`? I think that reflects what we're showing rather than "where" we're showing it.

mikejolley: Changed in https://github.com/woocommerce/woocommerce/pull/58066/commits/475faf72c2a5dc414c49773c12c499d367e1b1bb

plugins/woocommerce/client/blocks/assets/js/types/type-defs/cart-response.ts
mikejolley: I noticed we had 2 types for the same thing. I wanted to merge them without breaking existing usage of the `CartResponseShippingRate` type. ~~Looking again, its only used in a few files so I'll just remove usage of the `CartResponseShippingRate` type.~~

Edit: I saw some other examples of this such as `export type CartResponseItem = CartItem;`. I will leave it. While the types are the same, one is specifically for api responses. While the two types have parity right now it may not always be the case.

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/shipping/index.tsx
mikejolley: I'm not seeing a type warning in this file. Can you show me the output?

plugins/woocommerce/client/blocks/assets/js/blocks/cart/inner-blocks/cart-order-summary-totals/style.scss
mikejolley: Added in https://github.com/woocommerce/woocommerce/pull/58066/commits/17a9e518e874a9702a59a708390e1a6689bc0e3f

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
ralucaStan: > If shipping is selected, the "total" price will include the shipping price on the server.

Gotcha! I understand the reasoning behind it, but I was speaking mostly about the user experience, as I expect some shopper confusion, which is why I was suggesting the quick fix of changing the text to something like `Shipping can be changed at checkout`

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
ralucaStan: Makes sense, thank you for the context

plugins/woocommerce/client/blocks/assets/js/blocks/cart/style.scss
mikejolley: Done in https://github.com/woocommerce/woocommerce/pull/58066/commits/e5e19128be5df05ea3dd0e4ecfe605523a581e17 and https://github.com/woocommerce/woocommerce/pull/58066/commits/5dae165a859209103fe3cc227950f3e84594f812

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/shipping/index.tsx
ralucaStan: I'm no longer seeing the error, but this is what it was showing me. 

```
Argument of type 'CartResponseShippingRate[]' is not assignable to parameter of type 'CartShippingRate[]'.
  Type 'CartResponseShippingRate' is not assignable to type 'CartShippingRate'.
    Types of property 'shipping_rates' are incompatible.
      Type 'CartResponseShippingPackageShippingRate[]' is not assignable to type 'CartShippingPackageShippingRate[]'.
        Type 'CartResponseShippingPackageShippingRate' is not assignable to type 'CartShippingPackageShippingRate'.
          Types of property 'currency_code' are incompatible.
            Type '"" | CurrencyCode' is not assignable to type 'CurrencyCode'.
              Type '""' is not assignable to type 'CurrencyCode'.
```

mikejolley: Ah that shouldn't happen as we made those 2 types the same—I think I saw it prior to that change. I'm not seeing it now so I assume its fixed by making CartResponseShippingRate equal to CartShippingRate.

plugins/woocommerce/client/blocks/tests/e2e/tests/checkout/checkout-block.shopper.block_theme.spec.ts
mikejolley: Removed in https://github.com/woocommerce/woocommerce/pull/58066/commits/8e1de60cbaf189215594addade8b6582d4e2d318

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
mikejolley: https://github.com/woocommerce/woocommerce/pull/58066/commits/b0089b33dfc82a3ec52d96ef339433f39b45c274

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
mikejolley: In the mocks this line is not present when its been calculated already. Ill remove it.

mikejolley: Done in https://github.com/woocommerce/woocommerce/pull/58066/commits/fdabb2d4378295ee2389145f21219cb5e233cf79

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/shipping/index.tsx
ralucaStan: It complanins about [this type ](https://github.com/woocommerce/woocommerce/blob/e1fa42b054d0586b3e799e02ff02fa6d2cebed68/plugins/woocommerce/client/blocks/assets/js/types/type-defs/cart-response.ts#L55), which we aren't using anywhere. We use [this type ](https://github.com/woocommerce/woocommerce/blob/e1fa42b054d0586b3e799e02ff02fa6d2cebed68/plugins/woocommerce/client/blocks/assets/js/types/type-defs/cart.ts#L56)instead

mikejolley:
@ralucaStan @senadir I think thats most of the feedback addressed. Let me know what else you find.

plugins/woocommerce/client/blocks/assets/js/blocks/cart/inner-blocks/cart-order-summary-totals/style.scss
```diff
@@ -1,6 +1,16 @@
...
+	/**
+	* Hide the block if it has no children. `:empty` does not consider white-space, but this will change with CSS selectors level 4.
+	* `&:not(:has(*))` is a more robust selector that will work in all browsers.
+	*/
+	&:empty,
+	&:not(:has(*)) {
+		display: none;
+	}
+
 	.wc-block-components-totals-wrapper {
```
ralucaStan, line 14: Screenshot 2025-05-27 at 17 27 10
A double border appeared, coming from a div with this class. Also happens on the checkout block.

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
```diff
@@ -123,20 +133,29 @@ const TotalsFooterItem = ( {
...
+									<FormattedMonetaryAmount
+										className="wc-block-components-totals-footer-item-tax-value"
+										currency={ currency }
+										value={ parsedTaxValue }
+									/>
+								),
+							} ) }
+						</p>
+					) }
+					{ isEstimate && ! hasSelectedRates && (
```
ralucaStan: I'd also check for `cartNeedsShipping` here to make sure we are showing the message only for carts that require it; for example a digital product won't need to display this message

plugins/woocommerce/client/blocks/assets/js/blocks/cart/inner-blocks/cart-order-summary-totals/style.scss
ralucaStan: We shouldn't be displaying a border above the Button, see how it looks on trunk
Screenshot 2025-05-27 at 17 41 24

plugins/woocommerce/client/blocks/assets/js/blocks/cart/inner-blocks/cart-order-summary-shipping/block.tsx
```diff
@@ -2,87 +2,29 @@
...
-	const [ isShippingCalculatorOpen, setIsShippingCalculatorOpen ] =
-		useState( false );
-
-	if ( ! cartNeedsShipping ) {
-		return null;
-	}
+	const hasSelectedRates = hasSelectedShippingRate( shippingRates );
 
-	if ( isEditor && getShippingRatesPackageCount( shippingRates ) === 0 ) {
+	if ( ! cartNeedsShipping || ! hasSelectedRates ) {
```
ralucaStan: small improvement to return earlier if !cartNeedsShipping and skip calculating `hasSelectedRates` if so, the way it initially was

plugins/woocommerce/client/blocks/assets/js/base/utils/shipping-rates.ts
```diff
@@ -76,6 +76,20 @@ export const hasShippingRate = (
...
 
+export const hasSelectedShippingRate = (
+	shippingRates: CartShippingRate[]
+): boolean => {
+	if ( ! hasShippingRate( shippingRates ) ) {
+		return false;
+	}
+	return shippingRates.some(
+		( shippingRatesPackage ) =>
+			!! shippingRatesPackage.shipping_rates.find(
```
ralucaStan: You can use some instead of find here: `shippingRatesPackage.shipping_rates.some(( rate ) => rate.selected)`

ralucaStan (changes requested):
I've taken my time with this PR as it's impact is pretty big. I'm submitting the last code changes request. I'll add separate comments for the manual testing I've done

ralucaStan:
I've tested a store with shortcodes on both cart and checkout pages.
I've enabled the shipping calculator.
I've defined Shippind rates for 2 countries, one country has to be the default for the checkout.

Incognito, I am doing the following:
- I am in the cart shortcode and I click Calculate shipping -> notice the default shop country in the calculator
- I select another country, the other one that has the shipping rate defined
- I select the shipping rate shown for the new country
- I go to checkout
- I don't see my new country, nor the selected shipping option
- The form shows the default country

On trunk the checkout shows my calculator selection.

senadir:
I tested with a subscription, a booking product, and a deposit product:

Deposit:
image

image

Booking:
image

Subscription:
image

Do we need to change anything about them?

ralucaStan:
I also tested with subscriptions, but it gets weird and confusing when testing with local pickup enabled, or when there are multiple shipping options available.

# Multiple shipping options available
For example, on trunk I can select different shipping options for the subscription shippements, see photo. But on this branch 
the selection is synced between the two areas: when I change in one place, it also changes in the other
Screenshot 2025-05-27 at 19 56 47

# With Local Pickup
I think subscriptions does not support Local pickup, testing on trunk this is what I see. When I select local pickup it is not reflected in the order summary, so I can't test this scenario on your branch
Screenshot 2025-05-27 at 20 07 11
Screenshot 2025-05-27 at 20 08 52

ralucaStan:
I followed the testing instructions. 
Additionally tested with digital products, only local pickup and only shipping and both.
Tested shortcode only and shortcode in parallel.
Tested only with the Subscription extension. 
I reported my findings on the flows that didn't work as expected.
Could you provide some testing steps for the Multiple packages extension. I am not that familiar with it. Thank you

senadir:
Another thing is that it's strange that you're seeing broken package names, I fixed that 2 weeks ago
https://github.com/woocommerce/woocommerce/pull/57920

plugins/woocommerce/client/blocks/assets/js/blocks/cart/inner-blocks/cart-order-summary-totals/style.scss
mikejolley: There should be a single border like in figma.

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
mikejolley: https://github.com/woocommerce/woocommerce/pull/58066/commits/0e88eacafcaa4d96374ea7e7f2263379264e8c8b

plugins/woocommerce/client/blocks/assets/js/blocks/cart/inner-blocks/cart-order-summary-shipping/block.tsx
mikejolley: https://github.com/woocommerce/woocommerce/pull/58066/commits/c288ff56e20a8de6737f43936a7567bd01c7a6bb

plugins/woocommerce/client/blocks/assets/js/base/utils/shipping-rates.ts
mikejolley: https://github.com/woocommerce/woocommerce/pull/58066/commits/6f4bb8d40d2291659d887678eba59578c5613696

plugins/woocommerce/client/blocks/assets/js/blocks/cart/inner-blocks/cart-order-summary-totals/style.scss
mikejolley: Double border was an empty div with top border. Fixed in https://github.com/woocommerce/woocommerce/pull/58066/commits/35a434fac87029d63fc729204261c3fb30e7cb5f

mikejolley:
@ralucaStan @senadir Package names work on my branch but shipping selection is funky when toggling between pickup and shipping when subs is active. That seems to be due to request cancellation:

Screenshot 2025-05-28 at 11 26 39

Only shipping selection for one package makes it through to server.

mikejolley:
> Could you provide some testing steps for the Multiple packages extension. I am not that familiar with it. Thank you

- Enable "Multiple Packages for WooCommerce"
- WC > Settings > Multiple Packages
- Set "Group By" to product individual
- Add 2 items to cart
@ralucaStan

mikejolley:
@ralucaStan Subscriptions is broken on trunk as well. I'm not sure whats up with your build but given you're missing package names that was fixed in trunk, I'd suggest rebuilding and ensuring you have subs v7.5. I don't have the issue you describe, but on this branch and trunk I see these issues relating to subscriptions:

1. Choosing pickup in the main toggle does not sync to recurring packages. When you change pickup location it has no effect on the recurring packages.
2. Toggling pickup to shipping does not persist. The requests are aborted (see console XHR log).
3. Choosing pickup on a recurring package changes the sidebar total for the main package (it syncs), but the radio buttons do no update so the UI is stale.

I think this is way out of scope. It does need logging and it does need resolving, but not here. I think we need to find a way to hide the Local Pickup/Shipping toggle if multiple packages are used, or if the slot fill for packages is used.

- `selectShippingRate` has syncing logic if pickup is chosen. Review this.
- `ExperimentalOrderShippingPackages` slotfill adds the recurring packages. We need a way to reliably detect if this is used.
- Once we detect that ^ disable the pickup/shipping toggle and just show rates per package.

I think this should be a priority given the buggy experience right now. Some changes may be needed in subs as well.

cc @tjcafferkey 

@ralucaStan could you check trunk and if the same occurs, can we move ahead with the other changes in this PR?

mikejolley:
Looks like removing the toggle is inline with this p2 pfHfG4-1ic-p2

ralucaStan:
## Multiple packages 
@mikejolley thank you for the testing steps, I've tested with Multiple packages and it works as expected.
Screenshot 2025-05-28 at 18 45 31

## Subscriptions
I've updated the plugin, turns out I was behind. 
> For example, on trunk I can select different shipping options for the subscription shippements, see photo. But on this branch
the selection is synced between the two areas: when I change in one place, it also changes in the other
I can no longer replicate this, so we're 👌 

I did however find a bug for the recurring subscription.
1. Add a free and a flat rate shipping for store
2. Don't default to any country and add to cart a subs as a guest
3. See the cart
4. Notice that initial payment says shipping will be calculated, but recurring says free
Screenshot 2025-05-28 at 19 01 49
5. Go to checkout, notice same order summary but no selection for shipping in place

Screenshot 2025-05-28 at 19 07 00

## Subscriptions with Local pickup
I confirm the issues you listed. 
For me Local Pickup only gets listed under the recurring package, it's not an option for the main delivery.
All and all I agree that Subs needs some fixes, so we can't block this PR because of it.

ralucaStan:
> * I am in the cart shortcode and I click Calculate shipping -> notice the default shop country in the calculator
> * I select another country, the other one that has the shipping rate defined
> * I select the shipping rate shown for the new country
> * I go to checkout
> * I don't see my new country, nor the selected shipping option
> * The form shows the default country

@mikejolley  👍 Tested this again, and I can't replicate

ralucaStan (approved):
I'm approving this PR. I appreciate you answering all my questions and the quick feedback loop. This is an important change, so thank you for taking the initiative and making it happen via this PR.

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/shipping/index.tsx
```diff
@@ -27,20 +29,24 @@ export const TotalsShipping = ( {
...
+	const rateNames = getSelectedShippingRateNames( shippingRates );
+	const hasMultipleRates = rateNames.length > 1;
+	const rowLabel =
+		! hasSelectedRates || hasMultipleRates ? label : rateNames[ 0 ];
```
straku, line 36:
```suggestion
		( ! hasSelectedRates || hasMultipleRates ) ? label : rateNames[ 0 ];
```
A tiny suggestion, operator precedence is not yet engrained in my brain, even after years doing JS 😅

plugins/woocommerce/client/blocks/packages/components/totals/item/index.tsx
```diff
@@ -14,7 +14,7 @@ import FormattedMonetaryAmount from '../../formatted-monetary-amount';
 
 export interface TotalsItemProps {
 	className?: string | undefined;
-	currency: Currency;
+	currency?: Currency | undefined;
```
straku, line 17:
```suggestion
	currency?: Currency;
```
I think both additions (`?` and `| undefined`) are doing the same thing?

opr: having `currency?: Currency | undefined;` allows the currency prop to be explicitly set as undefined, e.g.

```jsx
<TotalsItemProps currency={ undefined } ... />
```

whereas just `currency?: Currency;` does not allow this, it either has to be missing, or `Currency`.

```jsx
<TotalsItemProps currency={ someCurrencyObject } ... />
```

or

```jsx
<TotalsItemProps ... />
```

straku: My understanding is that optional (`?`) still allows you to set it explicitly as `undefined`. See [this example](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAKhDOwoF4oG8CwAoKUD2YwAlngHYCGANgPwBcUiATkaQOYDc2AvttgPR8oASSgBjcqQYQkRJBAAeYSkVGzKIKOXhQArqQAmEAGYsI+7KLKJN9ONdSYc+QiQqV6ewydJnuvLAL4jAwA1kRgULIWVkgARrYISA48WNigkFAAgqR4wAAWEIx2SejYuATEZFT0TCysUAA+ugbGpvqcWCn8giLikvDSkXKKyqrA6pranq0+5liWpNbkAEz02bkFRYkopU4VrtXNXm1+qQGCfQDkSPBhEfl4Oqx50Ytxq1k5+YXFO2hcQA).

straku:
I went through the code changes (left tiny comments) and tested logged-out and logged-in flows. All looks good on my end, with a note that I didn't go as deep as other reviewers.

One question I have from my testing is why are we referring to the total as "Estimated total" in the cart, even after a proper shipping method or local pickup is selected?

plugins/woocommerce/client/blocks/packages/components/totals/item/index.tsx
mikejolley: I get an error with this change..

```
Type 'Currency | undefined' is not assignable to type 'Currency'.
```

mikejolley:
> One question I have from my testing is why are we referring to the total as "Estimated total" in the cart, even after a proper shipping method or local pickup is selected?

I think it reflects that the totals there are no final and will be subject to change based on the actual values you provide at checkout. You also cannot change shipping rate in cart, so I think its good that its not inferred to be final.

plugins/woocommerce/client/blocks/assets/js/blocks/cart/inner-blocks/cart-order-summary-discount/block.tsx
```diff
@@ -23,9 +23,15 @@ const DiscountSlotFill = (): JSX.Element => {
...
 
-const Block = ( { className }: { className: string } ): JSX.Element => {
+const Block = ( { className }: { className: string } ) => {
 	const { cartTotals, cartCoupons } = useStoreCart();
 	const { removeCoupon, isRemovingCoupon } = useStoreCartCoupons( 'wc/cart' );
+
+	// Hide if there are no coupons to show.
+	if ( ! cartCoupons.length ) {
+		return null;
+	}
```
opr, line 33: This is hiding the slot/fill - will raise a PR to fix

mikejolley: Is there a risk that will show a form (subscriptions) after we've removed it?

opr: I checked, subscriptions uses `ExperimentalOrderShippingPackages`  and `ExperimentalOrderMeta` the one I am proposing to return is `ExperimentalDiscountsMeta` - a plugin should not display shipping-related UI in this area.

Also, even if plugins _did_ show a form here, it would render anyway if the shopper added a coupon.

Issue #56080, opened 2025-03-04
Remove core profiler free features fallback


adrianduffell:
## Issue Description

To ensure partnerships changes can be handled efficiently, remove the fallback system that will show core-provided recommendations in edge-case scenarios.

## Acceptance Criteria

* Remove core profiler's usage of a core fallback for recommendations. Don't show the free features screen when suggestions are disabled or the WooCommerce.com REST API is unavailable.
* Ensure that strings for recommendation copy still exist in Core. They need to be found by WooCommerce.com to make translations in its REST API responses. A "strings file" may be needed for this.
* Wrap the removal around a feature flag, in case it needs to be easily reverted.
* Enable the feature flag for 50% of users initially.

PR #58071, opened 2025-05-16
Update MailPoet and Klaviyo variant assignment to 50/50 split

nerrad:

-   [x] I have followed the [WooCommerce Contributing Guidelines](https://github.com/woocommerce/woocommerce/blob/trunk/.github/CONTRIBUTING.md) and the [WordPress Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/).
-   [x] I have checked to ensure there aren't other open [Pull Requests](https://github.com/woocommerce/woocommerce/pulls) for the same update/change.
-   [x] I have reviewed my code for [security best practices](https://developer.wordpress.org/apis/security/).

This PR updates the variant assignment configuration for MailPoet and Klaviyo plugins to create an equal 50/50 split in visibility. Previously, the configuration was set to a 70/30 split (MailPoet: 70%, Klaviyo: 30%).

Closes #522.

| Before | After |
| ------ | ----- |
| MailPoet visible for 70% of stores (variant 1-84) | MailPoet visible for 50% of stores (variant 1-60) |
| Klaviyo visible for 30% of stores (variant 85-120) | Klaviyo visible for 50% of stores (variant 61-120) |

1. Set up a fresh WooCommerce installation
2. Verify the `woocommerce_remote_variant_assignment` option is set (it should be a random number between 1-120)
3. Test with different variant assignments:
   - For values 1-60: Verify MailPoet is visible
   - For values 61-120: Verify Klaviyo is visible
4. Verify that the visibility is mutually exclusive (stores should only see one of the plugins)

Environment details:
- Local development environment
- WordPress 6.4
- WooCommerce trunk
- Default theme
- No additional plugins installed

Testing performed:
- Verified variant assignments work as expected for both plugins
- Confirmed no existing tests were affected as there were no specific tests for the previous 70/30 split
- Tested visibility rules for both plugins across the full range of variant assignments

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Update MailPoet and Klaviyo variant assignment to 50/50 split. This change updates the visibility configuration for MailPoet and Klaviyo plugins to create an equal 50/50 split in visibility, replacing the previous 70/30 split configuration.

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/58071-update-wooprd-522-update-klaviyo-visibility-in-core-profiler
+++ b/plugins/woocommerce/changelog/58071-update-wooprd-522-update-klaviyo-visibility-in-core-profiler
@@ -1,0 +1,4 @@
+Significance: minor
+Type: update
+
+Update MailPoet and Klaviyo variant assignment to 50/50 split. This change updates the visibility configuration for MailPoet and Klaviyo plugins to create an equal 50/50 split in visibility, replacing the previous 70/30 split configuration.
--- a/plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/DefaultFreeExtensions.php
+++ b/plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/DefaultFreeExtensions.php
@@ -215,7 +215,7 @@ class DefaultFreeExtensions {
 					array(
 						'type'        => 'option',
 						'option_name' => 'woocommerce_remote_variant_assignment',
-						'value'       => array( 1, 84 ), // 70% segment with klaviyo
+						'value'       => array( 1, 60 ), // 50% segment with klaviyo
 						'default'     => false,
 						'operation'   => 'range',
 					),
@@ -254,7 +254,7 @@ class DefaultFreeExtensions {
 					array(
 						'type'        => 'option',
 						'option_name' => 'woocommerce_remote_variant_assignment',
-						'value'       => array( 85, 120 ), // 30% segment with mailpoet
+						'value'       => array( 61, 120 ), // 50% segment with mailpoet
 						'default'     => false,
 						'operation'   => 'range',
 					),


nerrad:
Hi @chihsuan - besides asking for your review, I have a couple additional questions/things I'd like you to confirm:

- I [see here](https://github.com/woocommerce/woocommerce/blob/febfbc4b24b0d3c9ee73c2fe8cca468c66b158e1/plugins/woocommerce/src/Internal/Admin/RemoteFreeExtensions/DefaultFreeExtensions.php#L99-L112) that if the `disable-core-profiler-fallback` is true (which [it is](https://github.com/woocommerce/woocommerce/blob/d77b9556e823f9ecc2322586692a27c1fe433b63/plugins/woocommerce/client/admin/config/core.json#L46)), the core profiler fallback will never be shown for the set of users falling under 60 as the value for the `woocommerce_remote_variant_assignment` option. Effectively that means that for the change here, neither Klaviyo nor MailPoet would show up in the free plugin options if the fallback is invoked. Is that correct? If so, I'm not sure I understand the purpose of this logic. Whey would we never want to show plugins matching a variant under 60?
- I understand that along with the changes here, we need to update WooCommerce.com to serve the correct variants when requested remotely. I'll do a follow up PR there.

chihsuan (approved):
I’ve reviewed the changes and tested the PR locally—everything works as expected. 🚢

chihsuan:
> Effectively that means that for the change here, neither Klaviyo nor MailPoet would show up in the free plugin options if the fallback is invoked. Is that correct?
 
@nerrad Yes, that's correct.

> If so, I'm not sure I understand the purpose of this logic. Whey would we never want to show plugins matching a variant under 60?

The purpose of this logic is to test the impact of removing the core profiler fallback recommendations. 

We aim to eliminate the fallback recommendations due to the inefficiency of maintaining updates (Issue: https://github.com/woocommerce/woocommerce/issues/56080), as it requires updating both the fallback and remote endpoints.

The fallback is only used when:

1. "Show Suggestions" option is disabled in Advanced Settings. The default value is true and users are unlikely to disable it before completing the onboarding.
2. The remote endpoint is not available. We assume this scenario is rare and if any issue arises, we can fix it immediately without WC release.

> I understand that along with the changes here, we need to update WooCommerce.com to serve the correct variants when requested remotely. I'll do a follow up PR there.

Yes, thank you! 🙌

nerrad:
Okay so this is being tested for 50% of users initially. But the goal is to eventually remove any fallbacks.

chihsuan:
Yes, that's correct.

Issue #58064, opened 2025-05-16
PHP Warning: array_sum(): Addition is not supported on type string on `WC_Cart::calculate_shipping`


jimjasson:
### Prerequisites

- [x] I have carried out troubleshooting steps and I believe I have found a bug.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

When a store uses PHP 8.4 and the latest `nightly` WooCommerce version (after `9.8.5`) and a Local Pickup option is selected in the cart, then the following warning is thrown: 

```
[16-May-2025 11:44:23 UTC] PHP Warning:  array_sum(): Addition is not supported on type string in /Users/jasonkytros/Local Sites/wp/app/public/wp-content/plugins/woocommerce-mono/plugins/woocommerce/includes/class-wc-cart.php on line 1514
[16-May-2025 11:44:23 UTC] PHP Stack trace:
[16-May-2025 11:44:23 UTC] PHP   1. {main}() /Users/jasonkytros/Local Sites/wp/app/public/index.php:0
[16-May-2025 11:44:23 UTC] PHP   2. require() /Users/jasonkytros/Local Sites/wp/app/public/index.php:17
[16-May-2025 11:44:23 UTC] PHP   3. wp($query_vars = *uninitialized*) /Users/jasonkytros/Local Sites/wp/app/public/wp-blog-header.php:16
[16-May-2025 11:44:23 UTC] PHP   4. WP->main($query_args = '') /Users/jasonkytros/Local Sites/wp/app/public/wp-includes/functions.php:1342
[16-May-2025 11:44:23 UTC] PHP   5. WP->parse_request($extra_query_vars = '') /Users/jasonkytros/Local Sites/wp/app/public/wp-includes/class-wp.php:818
[16-May-2025 11:44:23 UTC] PHP   6. do_action_ref_array($hook_name = 'parse_request', $args = [0 => class WP { public $public_query_vars = [...]; public $private_query_vars = [...]; public $extra_query_vars = [...]; public $query_vars = [...]; public $query_string = ''; public $request = 'wp-json/wc/store/v1/cart/select-shipping-rate'; public $matched_rule = '^wp-json/(.*)?'; public $matched_query = 'rest_route=/wc%2Fstore%2Fv1%2Fcart%2Fselect-shipping-rate'; public $did_permalink = TRUE }]) /Users/jasonkytros/Local Sites/wp/app/public/wp-includes/class-wp.php:418
[...]

Issue #26654, opened 2020-06-01
Shipping tax rounding is incorrect when product prices are entered inclusive of tax


madeincosmos:
**Describe the bug**
Even though shipping prices are normally entered exclusive of tax, making product prices tax-inclusive causes a problem with shipping tax rounding. This problem doesn't happen when product prices are entered exclusive of tax.

**To Reproduce**
1. Set up a flat shipping rate of $75.10
2. Create a standard tax rate of 25%.
3. In **WooCommerce > Settings > Tax**, make product prices display **exclusive of tax**:

![Screenshot](https://d.pr/i/yUHzUU+)
Screenshot: https://d.pr/i/yUHzUU

4. Add a free product to cart and observe tax on shipping is rounded to $18.78 (which is the correct rounding of $75.10 * 25% = $18.775):

![Screenshot](https://d.pr/i/O6ViUE+)
Screenshot: https://d.pr/i/O6ViUE

5. Change tax display to prices **inclusive of tax**:

![Screenshot](https://d.pr/i/K7vxZR+)
Screenshot: https://d.pr/i/K7vxZR

6. Observe the tax amount is rounded now to $18.77, which is incorrect:

![Screenshot](https://d.pr/i/QZRJNl+)
Screenshot: https://d.pr/i/QZRJNl

**Screenshots**
See above.

**Expected behavior**
Shipping tax is always rounded the same way, no matter how product prices are entered.

**Isolating the problem (mark completed items with an [x]):**
- [x] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [x] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/storefront/).
- [x] I can reproduce this bug consistently using the steps above.

**WordPress Environment**
<details>
```
`
### WordPress Environment ###

WordPress address (URL): http://new.test
Site address (URL): http://new.test
WC Version: 4.1.1
REST API Version: ✔ 1.0.7
WC Blocks Version: ✔ 2.5.16
Action Scheduler Version: ✔ 3.1.6
WC Admin Version: ✔ 1.1.3
Log Directory Writable: ✔
WP Version: 5.4.1
WP Multisite: –
WP Memory Limit: 512 MB
WP Debug Mode: ✔
WP Cron: ✔
Language: en_US
External object cache: –

### Server Environment ###

Server Info: nginx/1.17.3
PHP Version: 7.3.10
PHP Post Max Size: 128 MB
[...]

PR #58073, opened 2025-05-16
[dev] PHP 8.4 environment for CI testing PHPUnit jobs

### LINKED ISSUE #52081: PHP 8.4 compatibility
Cross-team parent issue for tracking efforts relating to PHP 8.4 compatibility. Those with access, please feel free to add checklist items as needed or desired.

- [ ] Research and look for possible or actual problems when running under PHP 8.4.
- [ ] Implicitly nullable types are deprecated.
    - [x] Resolve for Action Scheduler.
    - [x] Resolve for PaymentGatewaySuggestions, PluginsHelper, RemoteSpecs (Ghidorah).

PRs can reference this issue but, in general, individual PRs should not *close* this issue.

kalessil:

Closes #52081

Adds PHP 8.4 testing environments for PHPUnit and core E2E tests.

- CI-checks shall pass


plugins/woocommerce/includes/class-wc-cart.php
```diff
@@ -1511,8 +1511,8 @@ public function calculate_shipping() {
 			}
 		}
 
-		$this->set_shipping_total( array_sum( $shipping_costs ) );
-		$this->set_shipping_tax( array_sum( $merged_taxes ) );
+		$this->set_shipping_total( array_sum( array_filter( $shipping_costs ) ) );
+		$this->set_shipping_tax( array_sum( array_filter( $merged_taxes ) ) );
```
kalessil: https://github.com/woocommerce/woocommerce/issues/58064

plugins/woocommerce/includes/wc-core-functions.php
```diff
@@ -2033,9 +2033,10 @@ function wc_add_number_precision( ?float $value, bool $round = true ) {
 		return 0.0;
 	}
 
-	$cent_precision = pow( 10, wc_get_price_decimals() );
-	$value          = $value * $cent_precision;
-	return $round ? NumberUtil::round( $value, wc_get_rounding_precision() - wc_get_price_decimals() ) : $value;
+	$result          = $value * pow( 10, wc_get_price_decimals() );
+	$round_precision = $round ? wc_get_rounding_precision() - wc_get_price_decimals() : wc_get_rounding_precision();
```
kalessil, line 2038: https://github.com/woocommerce/woocommerce/issues/26654 re-appeared on PHP 8.4

plugins/woocommerce/includes/class-wc-cart.php
PanosSynetos: Hey @kalessil - Is this planned for WC 9.9 ?

kalessil: @PanosSynetos : don't know yet - primary reason for the change is to get the tests passing and connect dots to any existing issues. Now I need to coordinate with the teams on how to proceed with my fixes, so if you have any inputs or want to pull anyone in on the conversation, that would be helpful!

plugins/woocommerce/includes/wc-core-functions.php
kalessil: @jorgeatorres @Konamiman : I'd love to hear your thoughts on this change (fixes a regression on PHP 8.4) - both on technical side and which core version it should target.

plugins/woocommerce/includes/class-wc-cart.php
kalessil: @tjcafferkey, since you are assigned to the referenced ticket, any insight about when ya folks plan to take it into work? Asking, as my patch may be inaccurate, and my preference would be to rely on teams here while not delaying progress here more than for one week.

plugins/woocommerce/includes/wc-core-functions.php
kalessil: Domain-wise: if no rounding is requested, the extended rounding still takes place (`price decimals + 2` by default)

plugins/woocommerce/includes/class-wc-cart.php
PanosSynetos: This issue is causing Conditional Shipping and Payments unit tests to fail.

Let me tag @mikejolley too, as he discussed with @jimjasson over slack on Friday

What is important to note here, is that this doesn't cause a warning with PHP8.5 and WC 9.8.5, and all @mikejolley did, was to move a line further up in this PR.
- https://github.com/woocommerce/woocommerce/pull/55804

So, something else must have changed elsewhere, and I wonder if by doing an `array_filter` we mask/hide another issue.

mikejolley: You are correct the line just moved. So could be a different change. Not traced it yet.

It seems prices coming from shipping rates are numeric strings e.g. `"10"` and I'm wondering if something is returning an empty string here, which a `array_filter` would fix. I don't think there is any problem using `array_filter` in this context so for the purposes of fixing the notice, this is fine.

jimjasson: > wondering if something is returning an empty string here

From our tests, we noticed that Local Pickup methods that don't have a price have an empty string as the `cost`.

kalessil: :arrow_up: is pretty much what triggers the warning, and one test to fail. I narrowing down the changes in 8549a6194a62dc5ca318611e2f454a814139cda1

Konamiman (approved):

Issue #58084, opened 2025-05-16
Checkout can get into an invalid state if "Ship to specific countries" changes and the shopper has a saved address


opr:
## **Describe the bug**

If a store was set to ship to all countries, and a shopper from country A places an order and makes an account, their address would be saved to their account.

If, between orders (or even during the same shopping session) the store's "Ship to specific countries" option changes, and the new value does not include the country in the shopper's saved address, they are unable to proceed and must log out to be able to place an order.

## **Steps To Reproduce**

* Go to WC -> Settings -> General
* Set the following options
* Add shipping rates for USA and UK

<img src="https://uploads.linear.app/f629fcde-6f9b-48e8-8af3-60eed8823bff/7aa87109-fc0c-47b2-b253-5f78734cb791/251ad637-c9b1-4fd1-8a91-2b0d60120fc7?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2Y2MjlmY2RlLTZmOWItNDhlOC04YWYzLTYwZWVkODgyM2JmZi83YWE4NzEwOS1mYzBjLTQ3YjItYjI1My01Zjc4NzM0Y2I3OTEvMjUxYWQ2MzctYzliMS00ZmQxLThhOTEtMmIwZDYwMTIwZmM3IiwiaWF0IjoxNzU1NjAzODM3LCJleHAiOjMzMzI2MTYzODM3fQ.Yaz4bvyujX4Sp5WhjglOgEjDOJA5wu6npD-LR59LJ7A " alt="image.png" width="652" data-linear-height="162" />

* In incognito, create a new account, log into it, add an item to your cart and check out using a UK address.
* Go to WC -> Settings -> General and change the settings to this:

[...]

PR #58085, opened 2025-05-16
Display country as disabled select if there's a single option

opr:

- Update the country select field on the Checkout block to have: no arrow, no interaction, and no disabled state if only one country is available.
- If there is only one country available to sell to and ship to, default the shopper's address to that country.

Closes WOOPLUG-1961

| Before | After |
| ------ | ----- |
| https://www.loom.com/share/0003a2254730447eb6d2c1e99f753891 |  https://www.loom.com/share/a1d0f888bd8b4259bfdd329a1ea65b69 |

1. Go to WooCommerce -> Settings -> General -> Selling location(s). Set it to "Sell to specific countries only" - select one country. Ensure "Shipping location(s)" is set to "Ship to all countries you sell to"
2. Add an item to the cart and go to the Checkout block.
3. Ensure the country in shipping and billing is set and cannot be changed.
4. Use a screen reader and ensure the label for the field is read out and includes 'Country/Region, Cannot be changed'
5. Add this snippet
```php
add_filter('woocommerce_get_country_locale', function( $locale ){
        foreach ( $locale as $country => $value ) {
			$locale[$country]['country'] = [
				'label' => 'Custom country',
			];
	}
	return $locale;
}, 10, 1);
```

6. Repeat step 4 and ensure the correct label (Custom country) is read out.
7. Go to WooCommerce -> Settings -> General -> Selling location(s) and set it to "Sell to all countries".
8. Go to the Checkout block and ensure the country select can be changed. Test billing and shipping.
9. Go to WooCommerce -> Settings -> General -> Selling location(s) and change it to specific countries then choose a country
10. Go to WooCommerce -> Settings -> General -> Shipping location(s) and change it to specific countries, choose a _different_ country.
11. Go to the Checkout block and ensure the country field can be edited. It will include both countries. (Note a follow up issue will be created to take a look into this behaviour)
12. Go to WooCommerce -> Settings -> General -> Selling location(s)/Shipping location(s) and set the same country set in both options
13. Go to the Checkout block and ensure the country select cannot be changed.
14. Try various combinations for these shipping and selling options and ensure the expected behaviour occurs.
15. Test Cart/Checkout Shortcodes and ensure orders can be placed from both Blocks and Shortcodes.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/includes/wc-core-functions.php
+++ b/plugins/woocommerce/includes/wc-core-functions.php
@@ -1381,6 +1381,11 @@ function wc_get_customer_default_location() {
 		$default_location = wc_get_customer_geolocation( $default_location );
 	}
 
+	// If only one country is valid, preselect it.
+	if ( 1 === count( $allowed_countries ) ) {
+		$default_location['country'] = key( $allowed_countries );
+	}
+
 	/**
 	 * Filter the customer default location after geolocation.
 	 *
--- a/plugins/woocommerce/includes/wc-core-functions.php
+++ b/plugins/woocommerce/includes/wc-core-functions.php
@@ -1367,7 +1367,8 @@ function wc_get_customer_default_location() {
 	);
 
 	// Ensure defaults are valid.
-	$allowed_countries = WC()->countries->get_allowed_countries();
+	$allowed_countries          = WC()->countries->get_allowed_countries();
+	$allowed_shipping_countries = WC()->countries->get_shipping_countries();
 
 	if ( ! in_array( $default_location['country'], array_keys( $allowed_countries ), true ) ) {
 		$default_location = array(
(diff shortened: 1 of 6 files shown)


plugins/woocommerce/includes/wc-core-functions.php
```diff
@@ -1381,6 +1381,11 @@ function wc_get_customer_default_location() {
 		$default_location = wc_get_customer_geolocation( $default_location );
 	}
 
+	// If only one country is valid, preselect it.
+	if ( 1 === count( $allowed_countries ) ) {
+		$default_location['country'] = key( $allowed_countries );
+	}
+
```
senadir: Nice!

mikejolley (changes requested):
Functionality and testing; mostly ok. Happy paths worked.

You do however get locked in if you change settings e.g. shop country UK, allowed country US:

Screenshot 2025-07-30 at 11 55 50

Even though the input says US, in the backend and data store you're in the UK so you get this unrecoverable state.

Also I'm unsure how I feel about this usability wise. It seems like you should be able to click the input but that doesn't work so it gives the impression its broken. I think it needs to be er visibly disabled as its not an input?

ralucaStan:
> Also I'm unsure how I feel about this usability wise. It seems like you should be able to click the input but that doesn't work so it gives the impression its broken. I think it needs to be er visibly disabled as its not an input?

I agree with Mike here. We shouldn't disable the field. It's a common pattern to see in the select the countries a shop sells to. Being able to open that select, and see the list of countries, is a normal user behaviour.
Screenshot 2025-08-04 at 16 09 19

> Even though the input says US, in the backend and data store you're in the UK so you get this unrecoverable state.

There's a ticket for this https://github.com/woocommerce/woocommerce/issues/58084

PR #58093, opened 2025-05-17
[Blueprint] Prevent unsaved changes popup after successful export

### LINKED ISSUE #58047: [Blueprint] do not display unsaved changes popup after toggle export settings
in woocommerce/woocommerce#57657 We prevent unsaved changes popup after successful blueprint **import.**

We should do the same for the export section.

1. Go to WooCommerce > Settings > Advanced > Blueprints (beta)
2. Toggle export options
3. Navigate to any other section in the WP Admin dashboard
4. Verify that unsaved changes popup appears

BurakParsAydin:

Closes #58047  

1 .Go to **WooCommerce → Settings → Advanced → Blueprints (Beta)**
2. In the Export section, toggle one or more export options
3. Click the "Export" button
4. After the export completes, navigate to another page in the WP Admin (e.g., Dashboard)
5. **Expected**: No unsaved changes popup should appear

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Prevented the browser's 'unsaved changes' warning from appearing after a successful export in the Blueprint settings page

</details>

<details>

#### Comment 

</details>

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved an issue where the browser's "unsaved changes" warning appeared after successfully exporting settings on the Blueprint settings page. The warning will no longer display incorrectly after export.


chihsuan (commented):
Thanks for jumping in and helping out! @BurakParsAydin 

I can see that the current approach fixes the issue when user clicks the export button. However, we should prevent the unsaved changes warning when the user toggles the export options as well.

I believe the best way to do a similar change in #57657 is to add a wc-settings-prevent-change-event class around the export section.

Let me know if you have any questions.

BurakParsAydin:
@chihsuan Thanks for the feedback!

I understand your point and appreciate the suggestion to use. However, based on my local testing again, I’ve found that the current approach in this PR still prevents the unsaved changes warning when toggling options - not just on clicking the export button.

When I initially tested, I didn’t see the fix working until I did the following:

1. Ran `pnpm --filter='@woocommerce/plugin-woocommerce' build:admin`
2. Restarted Docker containers
3. Cleared the browser cache

After these steps, everything compiled cleanly, and I saw the warning on toggles if I don't click export button. 

That said, if there’s a specific flow where the warning still appears on your side, I’d be happy to revisit and explore alternative fixes like adding the class wrapper.

Let me know your thoughts!

chihsuan:
That's interesting. 🤔  I'm still seeing the warning on my end, and I verified that my local setup is running the most recent code.

As I understand it, the code you added is triggered only when the export button is clicked, which suggests it does not resolve the problem of the warning appearing when toggling options without pressing the export button. Anything I misunderstand?

BurakParsAydin:
@chihsuan Yes, the current implementation only addresses the case when the export button is clicked. I'll look into expanding the fix.

BurakParsAydin:
@chihsuan FYI, I’ve tried a few different approaches to avoid directly using `window.onbeforeunload = '';,` but unfortunately, none of them solved the issue as intended.

Here’s what I’ve attempted so far:
1. Wrapped the export section with `.wc-settings-prevent-change-event` – this successfully suppressed the unsaved changes warning, but it also disabled the warning when toggling options, which we still want to trigger.

2. Tried delegating a click handler on the export button using:
```
$('body').on('click', '.blueprint-settings-export-button', function () {
  window.onbeforeunload = '';
});
```
But this approach didn’t work either -  the warning still shows up when navigating away after clicking the export button.

3. I also looked into triggering WooCommerce’s form-based logic via` .submit :input`, but the export UI is rendered via a React slot-fill and not inside a native form or tracked input group.

At this point, I haven't found a viable alternative that clears the warning only when clicking the export button — without relying on `window.onbeforeunload = '';`.

Do you happen to have another idea in mind, or a pattern WooCommerce prefers for this case?

Thanks!

chihsuan:
@BurakParsAydin Thanks for exploring the options.

What do you mean by the "warning when toggling options"? When you toggle the options, you shouldn't see the warning. Or do you mean the warning when navigating away from the page? We don't need to show the warning when navigating away from the blueprint page.

BurakParsAydin:
@chihsuan Thanks for the follow-up!

I'm aware that we don’t need to show that warning when navigating away from the blueprint page. 
To clarify, by "warning when toggling options," I meant the **unsaved changes warning that appears when navigating away from the page** after interacting with toggle elements.

chihsuan:
> @chihsuan Thanks for the follow-up!
> 
> I'm aware that we don’t need to show that warning when navigating away from the blueprint page. To clarify, by "warning when toggling options," I meant the **unsaved changes warning that appears when navigating away from the page** after interacting with toggle elements.

@BurakParsAydin We don't need to show the warning when navigating away from the page after interacting with toggle elements. Hope this clarifies things.

BurakParsAydin:
@chihsuan Really appreciate your thoughtful review and guidance!
FYI, I've improved the logic. Looking forward to your feedback.

chihsuan:
> @chihsuan Really appreciate your thoughtful review and guidance! FYI, I've improved the logic. Looking forward to your feedback.


Thanks for the update. @BurakParsAydin I wonder why we can't wrap the export section with `.wc-settings-prevent-change-event? 🤔  As mentioned above, we want to prevent the unsaved changes popup for all cases.

BurakParsAydin:
@chihsuan Thanks for the clarification - that last note really helped clarify the requirement. I had initially interpreted it as needing to preserve the warning on some cases.

Just to confirm: we need to completely remove the unsaved changes warning for the entire export section - **no popups at all**, regardless of whether the user toggles options or clicks the export button. Am I understanding that correctly?

chihsuan:
> Just to confirm: we need to completely remove the unsaved changes warning for the entire export section - no popups at all, regardless of whether the user toggles options or clicks the export button. Am I understanding that correctly?

Yes, that’s correct!

BurakParsAydin:
@chihsuan FYI, I've applied the same class you used for the import section — it prevents the popup as expected. Looking forward to your feedback.

chihsuan (approved):
Thanks, tested well! 👍

PR #57859, opened 2025-05-08
[Experimental] Blockified Add to Cart with Options block: Make 'X in cart' button text work for variable products

Aljullu:
### Changes proposed in this Pull Request:

Closes #57726.

This PR fixes the "X in cart" text in the Add to Cart button for variable products.

### Screenshots or screen recordings:

https://github.com/user-attachments/assets/c52d5ea9-cd0a-421d-8ad4-cd5e292cdf7f

#### Preparation

0. Install and activate the [WooCommerce Beta Tester](https://woocommerce.com/products/woocommerce-beta-tester/) extension.
0. Go to _WooCommerce Admin Test Helper_ > _Features_ and enable `experimental-blocks` and `blockified-add-to-cart`.
1. Go to _Appearance_ > _Editor_ > _Templates_ > _Single Product_.
2. Click on the _Add to Cart with Options_ block and update to the blockified version.

#### Testing the PR

1. In the frontend, visit the template of a variable product.
2. Select some attributes and add them to cart.
3. Switch between attributes and verify the Add to Cart button text updates properly and always displays the correct number of products in cart.

PR #58097, opened 2025-05-19
Product Button: prepare to detect selected variation when used inside the Add to Cart Form + Options block

dinhtungdu:

The Product Button block is a production block, while the Add to Cart + Options block is an experimental block, so I split the changes to Product Button in https://github.com/woocommerce/woocommerce/pull/57859 into a separate PR for proper testing and history. This also makes it easier to revert if unexpected things happen.

I took this chance to change the implementation of the Product Button to handle selected variation ID. Previously, in https://github.com/woocommerce/woocommerce/pull/57859, we updated the `state.quantity` getter to retrieve the variation ID directly from there. We also needed a new context called `isDescendantOfAddToCartWithOptions` to determine if we were inside the Add to Cart Form + Options block.

```ts
			get quantity() {
				const { productId, isDescendantOfAddToCartWithOptions } =
					getContext();

				const itemId =
					isDescendantOfAddToCartWithOptions &&
					addToCartWithOptionsState?.productOrVariationId
						? addToCartWithOptionsState.productOrVariationId
						: productId;

				const product = wooState.cart?.items.find(
					( item ) => item.id === itemId
				);

				return product?.quantity || 0;
			},
```

In this PR, I took another direction, which does not use that new context but instead uses the result of `getContext('woocommerce/add-to-cart-with-options')` to determine if we are inside the Add to Cart Form with Options block. After that, I modified the `context.productId` of the button according to the selected variation.

Apart from the above change, this PR brings the naming changes as they are from https://github.com/woocommerce/woocommerce/pull/57859.

1. Smoke testing the Product Button block to verify no regression was introduced.
2. Test https://github.com/woocommerce/woocommerce/pull/57859 which includes change from this PR to verify the logic updating selected variation.

### DIFF
--- a/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
@@ -177,6 +177,8 @@ const { state } = store< Store >(
 				}
 			},
 			syncProductId() {
+				// This is intentionally not typed as we don't know if the block
+				// is inside the Add to Cart Form + Options block.
 				const addToCartContext = getContextFn(
 					'woocommerce/add-to-cart-with-options'
 				);
--- a/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
@@ -1,11 +1,7 @@
 /**
  * External dependencies
  */
-import {
-	store,
-	getContext as getContextFn,
-	useLayoutEffect,
-} from '@wordpress/interactivity';
+import { store, getContext, useLayoutEffect } from '@wordpress/interactivity';
 import type { Store as WooCommerce } from '@woocommerce/stores/woocommerce/cart';
 
 /**
@@ -55,8 +51,6 @@ interface Store {
 	};
 }
 
-const getContext = () => getContextFn< Context >();
-
 const { state: wooState } = store< WooCommerce >(
 	'woocommerce',
 	{},
@@ -68,23 +62,23 @@ const { state } = store< Store >(
 	{
 		state: {
 			get quantity() {
-				const { productId } = getContext();
+				const { productId } = getContext< Context >();
 				const product = wooState.cart?.items.find(
 					( item ) => item.id === productId
 				);
 				return product?.quantity || 0;
 			},
 			get slideInAnimation() {
-				const { animationStatus } = getContext();
+				const { animationStatus } = getContext< Context >();
 				return animationStatus === AnimationStatus.SLIDE_IN;
 			},
 			get slideOutAnimation() {
-				const { animationStatus } = getContext();
+				const { animationStatus } = getContext< Context >();
 				return animationStatus === AnimationStatus.SLIDE_OUT;
 			},
 			get addToCartText(): string {
 				const { animationStatus, tempQuantity, addToCartText } =
-					getContext();
+					getContext< Context >();
 
 				// We use the temporary quantity when there's no animation, or
 				// when the second part of the animation hasn't started yet.
@@ -103,14 +97,14 @@ const { state } = store< Store >(
 				);
 			},
 			get displayViewCart(): boolean {
-				const { displayViewCart } = getContext();
+				const { displayViewCart } = getContext< Context >();
 				if ( ! displayViewCart ) return false;
 				return state.quantity > 0;
 			},
 		},
 		actions: {
 			*addCartItem() {
-				const context = getContext();
+				const context = getContext< Context >();
 				const { productId, quantityToAdd } = context;
 
 				// Todo: Use the module exports instead of `store()` once the
@@ -142,7 +136,7 @@ const { state } = store< Store >(
 				actions.refreshCartItems();
 			},
 			handleAnimationEnd( event: AnimationEvent ) {
-				const context = getContext();
+				const context = getContext< Context >();
 				if ( event.animationName === 'slideOut' ) {
 					// When the first part of the animation (slide-out) ends, we move
 					// to the second part (slide-in).
@@ -158,7 +152,7 @@ const { state } = store< Store >(
 		},
 		callbacks: {
 			syncTempQuantityOnLoad() {
-				const context = getContext();
+				const context = getContext< Context >();
 				// When we instantiate this element, we sync the temporary
 				// quantity with the quantity in the cart to avoid triggering
 				// the animation. We do this only once, and we use
@@ -170,7 +164,7 @@ const { state } = store< Store >(
 				}, [] );
 			},
 			startAnimation() {
-				const context = getContext();
+				const context = getContext< Context >();
 				// We start the animation if the temporary quantity is out of
 				// sync with the quantity in the cart and the animation hasn't
 				// started yet.
@@ -183,7 +177,7 @@ const { state } = store< Store >(
 			},
 			syncProductId() {
 				const addToCartContext =
-					getContextFn< AddToCartWithOptionsContext >(
+					getContext< AddToCartWithOptionsContext >(
 						'woocommerce/add-to-cart-with-options'
 					) as AddToCartWithOptionsContext | undefined; // Product Button may not be nested inside Add To Cart + Options block.
 
@@ -191,7 +185,7 @@ const { state } = store< Store >(
 					return;
 				}
 
-				const context = getContext();
+				const context = getContext< Context >();
 				const { productId, variationId } = addToCartContext;
 
 				// The `variationId` is only available when the product is a
--- a/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
@@ -46,6 +46,7 @@ interface Store {
 	callbacks: {
 		startAnimation: () => void;
 		syncTempQuantityOnLoad: () => void;
+		syncProductId: () => void;
 	};
 }
 
@@ -173,6 +174,33 @@ const { state } = store< Store >(
 					context.animationStatus === AnimationStatus.IDLE
 				) {
 					context.animationStatus = AnimationStatus.SLIDE_OUT;
+				}
+			},
+			syncProductId() {
+				const addToCartContext = getContextFn(
+					'woocommerce/add-to-cart-with-options'
+				);
+
+				if (
+					! addToCartContext ||
+					! ( 'productId' in addToCartContext ) ||
+					! ( 'variationId' in addToCartContext ) ||
+					typeof addToCartContext.productId !== 'number' ||
+					typeof addToCartContext.variationId !== 'number' ||
+					! addToCartContext.productId
+				) {
+					return;
+				}
+
+				const context = getContext();
+				const { productId, variationId } = addToCartContext;
+
+				// The `variationId` is only available when the product is a
+				// variable product and the user has selected a variation.
+				if ( variationId > 0 ) {
+					context.productId = variationId;
+				} else {
+					context.productId = productId;
 				}
 			},
 		},
--- a/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
@@ -8,6 +8,11 @@ import {
 } from '@wordpress/interactivity';
 import type { Store as WooCommerce } from '@woocommerce/stores/woocommerce/cart';
 
+/**
+ * Internal dependencies
+ */
+import type { Context as AddToCartWithOptionsContext } from '../../../../blocks/add-to-cart-with-options/frontend';
+
 // Stores are locked to prevent 3PD usage until the API is stable.
 const universalLock =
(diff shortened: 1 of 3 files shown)


dinhtungdu:
@samueljseay @luisherranz tagging you as the reviewers, it's great to hear what you think about the pattern I used in this PR.

I wonder if iAPI provides a parent-agnostic approach to passing the data down, from parent to children, something like React prop. It'd be great incase we want to use the Product Button block inside other block that also provide product and variation ID.

Right now, what I can do with iAPI is getting the data from the children with a specific parent region (`woocommerce/add-to-cart-with-options`). And this works well. But we will have to do the same when we want to use the Product Button block inside another block. It's not the case now for sure, we don't need to do any premature optimization now, but I'm still wondering if iAPI provides other alternative.

_Mon 19 May, 11:54 UTC+7: Edited to clarify my point._

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
```diff
@@ -175,6 +176,35 @@ const { state } = store< Store >(
 					context.animationStatus = AnimationStatus.SLIDE_OUT;
 				}
 			},
+			syncProductId() {
+				// This is intentionally not typed as we don't know if the block
+				// is inside the Add to Cart Form + Options block.
+				const addToCartContext = getContextFn(
```
samueljseay: I want to say a couple things here.

1. wrapping getContext is a typing convenience and it's barely useful unless you explicitly only get exactly one context. You'll find over time that you want to access multiple scopes. If we're going to wrap it lets at least name it better.

2. I think there is a flaw in the typing of `getContext`. In JS APIs that TS has type signatures for, it's always stricter about possible null or undefined values (think document.querySelector etc), but the `getContext` type signature always returns `T`. This is incorrect, and it should be `T | undefined` and that will make all this code a lot clearer.

For example, you should be able to type getContextFn by passing the generic, then you can just ask: `if (addToCartContext)` and TS will narrow the type accordingly. Right now the type is misleading that you'll always have a context.

Also: it's not correct not to type this here. As mentioned `getContext` should tell you it can return undefined, but you can still do a check for undefined and that should be sufficient.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
```diff
@@ -175,6 +176,35 @@ const { state } = store< Store >(
...
+					'woocommerce/add-to-cart-with-options'
+				);
+
+				if (
+					! addToCartContext ||
```
samueljseay: This feels way too defensive. if we have a cart context it means the addToCartContext was resolved from `'woocommerce/add-to-cart-with-options'`, is every one of these checks needed?

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
luisherranz: I see your point, but in most blocks, the context cannot _not_ exist. If that happens, it's not a problem with the types, it's a problem with the HTML, and that should be fixed elsewhere. Adding `if (addToCartContext)` could even be more confusing since the developer will not see their JavaScript failing when that happens, unless they throw errors on the `else`. But those errors, apart from adding boilerplate, would actually exist only for the development phase, since those errors would only be reached if the HTML is broken, and when that happens, the block is broken and shouldn't be published.

If there really is a block where the context may or may not exist in production (when the block is properly working), then the developer can use the types to indicate it.

```ts
const context = getContext<MyContext | undefined>();

if (context) {
  context.someValue = newValue;
} else {
  state.someValue = newValue;
}
```

Also, I know that using an `if` to narrow down the types when they exist but TypeScript doesn't know it is a common pattern (even in production code), but for me that has always been a limitation in the typing system and the reason why TypeScript decided to include the [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator), which has no counterpart in JavaScript.

That's simply my opinion, anyway. We can disagree 🙂

samueljseay: > I see your point, but in most blocks, the context cannot not exist. If that happens, it's not a problem with the types, it's a problem with the HTML, and that should be fixed elsewhere.

I disagree with this and I think it's against the design of TS in general. For one thing the HTML is entirely decoupled from the JavaScript. There is no contract or guarantee given in the JavaScript, much like there isn't for the presence of an HTML element when you call any HTML query selector or any other HTML API. If my HTML element is not present that's certainly something I can "fix" but doesn't change the nature of the frontend API and the true return type (which we should (imo) honor for clarity).

Saying something "should be fixed in HTML" is also implying something is wrong. Here we're looking at a simple scenario where you may or may not be nested within another context but you want to check for it. Its valid and it will return undefined, but our TS type signature will never correctly tell you this. Now we have a situation where the TS signature is misleading, and doesn't explain to you how the API works either.

I think looking at TS types from the angle of "where you should fix the problem in your iAPI code" is not correct. TS is about program correctness and also the type signature self declares the behaviour of the API so you can understand it.

I can agree with the workaround to an extent, although I still don't think its correct. The real type signature in TS is (imo) always the one we should use to describe the API.

samueljseay: For this PR though, what i think I can agree on is that we should put `| undefined` in the type signature of this PR's call to getContext where context is conditionally not available depending on how blocks are nested.

dinhtungdu: > For this PR though, what i think I can agree on is that we should put | undefined in the type signature of this PR's call to getContext where context is conditionally not available depending on how blocks are nested.

I agree with this, too.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
dinhtungdu: Right! Let's remove everything except the first check.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
dinhtungdu: We can't pass `AddToCartWithOptionsContext | undefined` as the generic. I ended up casting the value instead.

samueljseay: @luisherranz what do you think about relaxing the type constraint of `T` so we can pass undefined? The comment should also be updated imo at least to inform that context can be undefined. 

This feels so weird to me though. I don't know why we're against the API documenting what it does. Sorry but I don't get the angle on this one. I wouldn't make a type signature and code comment for a function that don't match.

luisherranz: If the goal is to inform the developer when a context isn't available in the current scope, perhaps the consideration should be to make `getContext` throw an error. In that case, the final code would be:

```ts
try () {
  const context = getContext<MyContext>();
  context.someValue = newValue;
} else {
  state.someValue = newValue;
}
```

This would be a non-backwards-compatible change, however.

Again, I reiterate that this is just my opinion. I'm not against adding `| undefined` to the types.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
```diff
@@ -175,6 +175,27 @@ const { state } = store< Store >(
...
+				const { productId, variationId } = addToCartContext;
+
+				// The `variationId` is only available when the product is a
+				// variable product and the user has selected a variation.
+				if ( variationId ) {
+					context.productId = variationId;
+				} else {
+					context.productId = productId;
+				}
+			},
```
luisherranz: We should avoid all manual state synchronization. If the state already exists somewhere in the application, the new reference/value should always be derived state.

```js
  get productId() {
    const addToCartContext = getContext( 'woocommerce/add-to-cart-with-options');
    return ! addToCartContext
      ? getContext().productId
      : addToCartContext.variationId || addToCartContext.productId;
  }
```

Then switch from using `context.productId` to `state.productId` in all the Product Button references.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
samueljseay: > This would be a non-backwards-compatible change, however.

Yes I had considered that runtime API changes could improve this, but felt that a type change would be less impact. For example React's createContext lets you provide a fallback value when a context can't be found, but that would also be a backward incompatible change.

For my curiosity though, Is it a valid use-case though to try and get context when there might not be one? Throwing  an error seems extreme if its not considered a critical error state?

luisherranz (dismissed):
LGTM, except that we should use derived state instead of manually synchronizing the states.

Furthermore, we don't need to do this now, but I think this would be a good example of how extensibility should work when reusing a child block for multiple purposes. Instead of delegating the functionality directly to the child block and hardcoding it, it would be better to identify patterns that allow us to control it from the parent block.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
luisherranz: Normally, if a block cannot function without a context but the context is missing, it means that something is _very wrong_, and an error could be (or have been) the correct way to inform the developer about it.

Anyway, I think you're right. We can add the type modification in Woo's iAPI and see how it feels.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
dinhtungdu: Right! Thanks for the reminder. I did the same with Filters but didn't do it this time as it seems to be out of scope for this PR. But admittedly, having a callback to sync the context is not a good approach. I will update this to use derived state.

dinhtungdu:
> Furthermore, we don't need to do this now, but I think this would be a good example of how extensibility should work when reusing a child block for multiple purposes. Instead of delegating the functionality directly to the child block and hardcoding it, it would be better to identify patterns that allow us to control it from the parent block.

Yes. It's what I'm asking [above](https://github.com/woocommerce/woocommerce/pull/58097#issuecomment-2889531834), too. It's may not a problem for filter that does the navigation everytime, but it will be for case like this when we want to use the Product Button block in other place.

luisherranz:
> Yes. It's what I'm asking [above](https://github.com/woocommerce/woocommerce/pull/58097#issuecomment-2889531834), too. It's may not a problem for filter that does the navigation everytime, but it will be for case like this when we want to use the Product Button block in other place.

Oh, I missed that part, sorry. Yeah, exactly what you said but extended to callbacks and other properties. I've opened a post so we can discuss it.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
dinhtungdu: Updated in [`fc8f7a5` (#58097)](https://github.com/woocommerce/woocommerce/pull/58097/commits/fc8f7a5596dbe06bbefb97783a92f4185adddce8). For state that referring another state (indirectly) in their returns, I have to manually type the return type to help TS to resolve the type.

luisherranz: > I have to manually type the return type to help TS to resolve the type.

Yes, that's a caveat documented [here](https://developer.wordpress.org/block-editor/reference-guides/interactivity-api/core-concepts/using-typescript/#typing-the-derived-state).

dinhtungdu: @luisherranz I updated the type of the Product Button store in [614b38b](https://github.com/woocommerce/woocommerce/pull/58097/commits/614b38b1d33b969d17fd287752e55972beb20af1), please take a look at it.

dinhtungdu:
@samueljseay @luisherranz this PR is ready for another review. Please take a look! Thanks!

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/button/frontend.ts
luisherranz: That's great, thanks.

Just so you know, I'd like to test with a helper `AsyncAction<ReturnType>` to replace the `Generator` type:

- https://github.com/woocommerce/gutenberg/pull/15

We can update those types once the new iAPI version is used without the feature flag.

dinhtungdu:
Failing tests doesn't look flaky, investigating 👀 
Update: failed tests are still passing locally.
Update: fixed the issue. Not sure how can I missed checking for parent context 🤦🏽

Aljullu (approved):
Code changes look good, thanks for cleaning that up! I did some smoke testing and couldn't find any regressions either, so LGTM.

<details>

* Theme: Twenty Twenty-Five
* Store size: small
</details>

dinhtungdu:
Merging this PR to unblock https://github.com/woocommerce/woocommerce/pull/57859 and https://github.com/woocommerce/woocommerce/pull/58002

PR #58135, opened 2025-05-20
Add static analysis for email-editor PHP package 

lysyjan:

Closes WOOPLUG-3234.

- This PR reintroduces the PHPStan static analysis command.
- The command should be executed as part of the CI job when PHP files in the email editor package are changed.

1. Checkout the branch
2. ~Install composer dependencies in the PHP email-editor package~
3. Execute ~`composer run phpstan` in `packages/php/email-editor` or~ `pnpm --filter='@woocommerce/email-editor-php-config' phpstan`
4. For testing on PHP 7.4 you can use `pnpm --filter='@woocommerce/email-editor-php-config' phpstan:php7` and for PHP 8.4 `pnpm --filter='@woocommerce/email-editor-php-config' phpstan` or `pnpm --filter='@woocommerce/email-editor-php-config' phpstan:php8`
5. The command `pnpm run phpstan` can be used directly in `packages/php/email-editor`
6. The command installs its own dependencies in `packages/js/email-editor/tasks/phpstan/vendor` with `stubs` packages. There is an optional parameter `--cleanup` removing those dependencies to avoid unwanted autocomplete in IDE.

I tested the command locally with PHP 8.1.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


packages/php/email-editor/composer.json
```diff
@@ -47,7 +47,7 @@
 		"test:integration": "wp-env run tests-cli --env-cwd=wp-content/plugins/email-editor ./vendor/bin/phpunit --configuration phpunit-integration.xml.dist",
 		"phpcs": "phpcs -s -p",
 		"phpcbf": "phpcbf -p",
-		"phpstan": "php ./tasks/run-phpstan.php"
+		"phpstan": "vendor/bin/phpstan analyse -c phpstan.neon"
```
costasovo: I would suggest we add a parameter for memory to the command

```
"phpstan": "vendor/bin/phpstan analyse -c phpstan.neon --memory-limit=2G"
```

This will avoid memory limit errors when someone has PHP configured with a strict memory limit locally.

costasovo (commented):
@lysyjan Thanks, you for adding the analysis and for fixing the issues. Great job!

There is an issue with the core and woo stubs I think we need to address. Please see my other comment.

I was clicking through the CI jobs below this PR, but was not able to find where the PHPStan ran. Could you please point me to the job? I assumed it there should be a run because PHP files changed.

What PHP version will be used in the CI job?

I tested it locally with phpVersions from 7.4 to 8.4 set in the phpstan.neon. All were green. Only the 8.4 reported two errors, but we don't need to address those now:
```
21     Deprecated in PHP 8.4: Parameter #3 $previous (Throwable) is implicitly nullable via default value null.  
         🪪 parameter.implicitlyNullable                                                                           
  26     Deprecated in PHP 8.4: Parameter #1 $previous (Throwable) is implicitly nullable via default value null.  
         🪪 parameter.implicitlyNullable  
```

lysyjan:
@costasovo I fixed the mentioned issues.

I updated the command names and tried to use the most straightforward approach for running PHPStan with different PHP versions. Let me know if you prefer a different solution.

I also needed to make minor adjustments in the README file because the markdown validation job failed.

costasovo:
@lysyjan Thanks for the update! I can now see the jobs below the PR!

There is still the issue with stubs. I can't see my comment I was writing yesterday so it looks like I forgot to submit that particular one 🤦 Sorry, my bad!

The problem is that the Woo and WordPress stubs we install for PHPStan get picked by IDEs and it causes for example, that they are offered by the IDE to developers when they go to the function definition. See screenshot where I try to click get to  `getSubject` source. Instead of going directly to the source, I get these three options to choose from.

Screenshot 2025-05-22 at 7 37 02

I'm afraid we can't merge it like this because this will affect the dev experience for all users of the monorepo. I'm not sure what is the best solution. I think we could install/unzip? these stubs only for the job and then remove them. Or perhaps there might be another way to prevent IDEs from picking them up.

lysyjan:
@costasovo Thank you. I added two new commits and rebased the branch.
Please test if the commands work as expected.

I decided to create a simple bash script for better readability.

Konamiman:
I'm having some trouble with the testing:

> Install composer dependencies in the PHP email-editor package

I guess this means `cd packages/php/email-editor` from the root of the repository and then `composer install`? That just tells me "Nothing to install, update or remove"

> `composer run phpstan`

This fails with "Script "phpstan" is not defined in this package"

> `pnpm --filter='@woocommerce/email-editor-php-config' phpstan`

This seems to work (progress bar reaches 100%) but at the end it fails with "Internal error: This enum case does not have a value while analysing file"

Edit: the last one only fails on PHP 7.4, it works fine on PHP 8.0+. Is this expected? If so, shouldn't the PHP version be checked at the beginning of the process?

costasovo (commented):
@lysyjan Thanks for the updates! It works nicely!

I see I need to use `pnpm phpstan:php7 --cleanup` to run it. So you may need to update README.md which says composer phpstan...

I suggest making the `--cleanup` argument default behavior and perhaps changing the logic to `--skip-cleanup`. 
The files are locally installed from composer cache, so I don't observe any difference in the speed, and I think it might be better for someone who would work with the package only occasionally and may forget or just don't know that they need to clean up.
But it is only a suggestion. I'm also fine with keeping it as is as I can simply use it with the argument.

lysyjan:
@Konamiman Thank you for testing. I need to apologize because I forgot to update the testing instructions in the description. After some @costasovo suggestions, we decided to create a separate `composer.json`. The main reason was loading the stubs packages by IDEs. More details are [here](https://github.com/woocommerce/woocommerce/pull/58135#issuecomment-2899991597).

> This seems to work (progress bar reaches 100%) but at the end it fails with "Internal error: This enum case does not have a value while analysing file"

If you run PHPStan on PHP 7.4, you should use the command `phpstan:php7`. I updated the testing instructions and the package readme file.

@costasovo I updated the readme file and changed the option `cleanup` to `skip-cleanup`

Konamiman:
@lysyjan Thanks, tested with the updated instructions with PHP 7.4, 8.0 and 8.4 and it works as expected.

Konamiman (approved):

PR #58139, opened 2025-05-20
Prevent extra space rendering under local pickup options in Checkout Block

opr:

- Add a new `selectedOption` arg for the `renderPickupLocation` method. We use to generate the react components that show the pickup instructions, however these instructions only show when the method is selected. If the method is not selected, the `ReadMore` component is still rendered, but it is empty, however there is spacing applied to it with CSS, even when empty.
- The method will skip rendering the component if it's not selected.

| Before | After |
| ------ | ----- |
| image |  image |

1. Go to WC -> Settings -> Shipping -> Local Pickup and add multiple methods, some of these methods should have an empty "Pickup details" option.
image

2. Add an item to your cart and go to the Checkout block.
3. Select Local Pickup and select different methods, ensure they show correctly.
4. Ensure other components on the page using Radio buttons work OK (Shipping, Payment).

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/local-pickup-select/index.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/local-pickup-select/index.tsx
@@ -47,7 +47,11 @@ export const LocalPickupSelect = ( {
 				highlightChecked={ true }
 				selected={ selectedOption }
 				options={ pickupLocations.map( ( location ) =>
-					renderPickupLocation( location, packageCount )
+					renderPickupLocation(
+						location,
+						packageCount,
+						selectedOption
+					)
 				) }
 			/>
 		</div>
--- a/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/local-pickup-select/index.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/local-pickup-select/index.tsx
@@ -47,11 +47,7 @@ export const LocalPickupSelect = ( {
 				highlightChecked={ true }
 				selected={ selectedOption }
 				options={ pickupLocations.map( ( location ) =>
-					renderPickupLocation(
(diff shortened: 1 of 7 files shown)


plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/local-pickup-select/index.tsx
```diff
@@ -47,7 +47,11 @@ export const LocalPickupSelect = ( {
 				highlightChecked={ true }
 				selected={ selectedOption }
 				options={ pickupLocations.map( ( location ) =>
-					renderPickupLocation( location, packageCount )
+					renderPickupLocation(
```
straku: Suggestion unrelated to the change – this should be a React component.

straku (approved):
Works for me, the fix looks straightforward. You might consider this an opportunity to increase test coverage and add a test, but I leave it up to you.

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/local-pickup-select/index.tsx
```diff
@@ -47,7 +47,11 @@ export const LocalPickupSelect = ( {
...
+						location,
+						packageCount,
+						selectedOption
```
straku: Day after feedback: instead of passing `selectedOption` here, it would be more fitting to pass `isSelected` instead. It's the parent component that decides which option is selected (hence having access to `selectedOption`), having all related logic to this aspect in the parent makes sense as well.

opr: I think we can derive that from `location` so I removed the prop. Thanks!

PR #56797, opened 2025-03-28
Remove declare(strict_types=1) from wc-product-functions.php

tjcafferkey:
### Changes proposed in this Pull Request:

Removes `declare(strict_types=1);` from `wc-product-functions.php` to prevent fatal TypeError exceptions when non-numeric values are passed to `wc_get_related_products()`. This declaration was causing PHP to enforce strict type checking, which would fail when performing arithmetic operations on the `$limit` parameter if it was passed as a string.

Closes https://github.com/woocommerce/woocommerce/issues/56485

(For Bug Fixes) Bug introduced in PR https://github.com/woocommerce/woocommerce/pull/55843

### Screenshots or screen recordings:

| Before | After |
| ------ | ----- |
|        |       |

### How to test the changes in this Pull Request:

Using the [WooCommerce Testing Instructions Guide](https://github.com/woocommerce/woocommerce/wiki/Writing-high-quality-testing-instructions), include your detailed testing instructions:

1. Edit `wc_get_related_products()` in `wc-product-functions.php` 
2. As the first line _within_ the function add `$limit = '5';` as shown below.
3. Ensure no fatal error occurs on the product page.

```php
function wc_get_related_products( $product_id, $limit = 5, $exclude_ids = array(), $related_by = array() ) {
	$limit          = '5';
```

### Changelog entry

-   [ ] Automatically create a changelog entry from the details below.

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Details</summary>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

-   [ ] Fix - Fixes an existing bug
-   [ ] Add - Adds functionality
-   [ ] Update - Update existing functionality
-   [ ] Dev - Development related task
-   [ ] Tweak - A minor adjustment to the codebase
-   [ ] Performance - Address performance issues
-   [ ] Enhancement - Improvement to existing functionality

#### Message 

</details>

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

PR #58140, opened 2025-05-20
[Checkout] Support for shareable checkout URLs which pre-populate the cart/coupons

mikejolley:

## Changes proposed in this Pull Request

This PR introduces the ability to generate a cart from a link which redirects to checkout with a unique session ID, or load a cart via a token in the query string. 

An example link: 

```
https://store.local/checkout-link/?products=18%3A2%2C19%3A1&coupon=TEST
```

or without encoding:

```
https://store.local/checkout-link/?products=18:2,19:1&coupon=TEST
```

The main changes include:

- New `/checkout-link` endpoint that allows users to pre-populate their cart by visiting a URL with product IDs and product Qty, and optional coupon code.
- Support for loading a cart session via a token in the query string, enabling persistent and shareable carts.
- Improved error handling and user notices for invalid products and coupons, with more user-friendly messages (e.g., “Product with ID '999999' was not found and cannot be added to the cart.” and “Coupon 'INVALID_COUPON' cannot be applied because it does not exist.”).
- Refactored session and cart token utilities to support the new flows.
- E2E Playwright tests for:
  - Valid cart and coupon via checkout link
  - Invalid coupon via checkout link
  - Invalid product(s) via checkout link (both mixed and only invalid)
- Updated error messages in `class-wc-coupon.php` and `class-wc-discounts.php` for consistency and clarity.

## Benefits

- Enables external sites, emails, or marketing campaigns to link directly to a pre-filled WooCommerce cart or checkout.
- Supports persistent, shareable carts via tokens.
- Improves UX with clear, actionable error messages for invalid cart/coupon scenarios.

## Testing instructions

1. **Valid Cart Link**
   - Visit `/checkout-link?products=<valid_id_1>,<valid_id_2>&coupon=<valid_coupon>`
   - You should be redirected to checkout with both products in the cart and the coupon applied.

2. **Valid Cart Link w/ product quantities**
   - Visit `/checkout-link?products=<valid_id_1>:2,<valid_id_2>&coupon=<valid_coupon>`
   - You should be redirected to checkout with both products in the cart and the coupon applied.
   - <valid_id_1> should have a quantity of 2
   
3. **Invalid Coupon**
   - Visit `/checkout-link?products=<valid_id_1>,<valid_id_2>&coupon=INVALID_COUPON`
   - You should see a notice: `Coupon "INVALID_COUPON" cannot be applied because it does not exist.`

4. **Invalid Product (mixed)**
   - Visit `/checkout-link?products=<valid_id_1>,999999`
   - You should see a notice: `Product with ID "999999" was not found and cannot be added to the cart.`
   - The valid product should still be in the cart.

5. **Only Invalid Product**
   - Visit `/checkout-link?products=999999`
   - You should be redirected to the cart page.
   - You should see a notice: `The provided checkout link was out of date or invalid. No products were added to the cart.`

6. **Cart Token** (logged out)
   - Repeat step 1. You'll be on the checkout.
   - Copy the page address. Notice `session=` is part of the URL.
   - Open a new incognito window or different browser.
   - Visit the link. You should see the same cart contains and can proceed with ordering.
   - Check the application panel in console and look at the session cookie. Compare to the original window. The cookies will be difference because the session was cloned when launching incognito session.

7. **E2E Tests**
   - Run Playwright E2E tests in `plugins/woocommerce/tests/e2e-pw/tests/checkout/checkout-link.spec.js` to verify all scenarios.

## Checklist

- [x] My code is tested and passes existing unit/E2E tests
- [x] I have added new tests where applicable
- [x] I have updated relevant documentation
- [x] I have reviewed the [contributing guidelines](https://github.com/woocommerce/woocommerce/blob/trunk/CONTRIBUTING.md)
- [x] This PR adheres to the project spec ([external_-checkout-experience-implementation-feb-2025.pdf](external_-checkout-experience-implementation-feb-2025.pdf))

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Blocks/Domain/Services/CheckoutLink.php
```diff
@@ -0,0 +1,115 @@
...
+		$controller->empty_cart();
+
+		// Populate cart with products.
+		$products = wp_parse_id_list( get_query_var( 'products' ) ?? '' );
+		foreach ( $products as $product_id ) {
+			try {
+				$controller->add_to_cart(
+					[
+						'id'       => $product_id,
+						'quantity' => 1,
```
senadir: quantity is passed in, so it should be handled.

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutLink.php
```diff
@@ -0,0 +1,115 @@
...
+	 * Generate a cart token.
+	 *
+	 * @return string
+	 */
+	private function generate_cart_token() {
+		return JsonWebToken::create(
+			[
+				'user_id' => SessionUtils::generate_customer_id(),
+				'exp'     => time() + DAY_IN_SECONDS,
+				'iss'     => 'checkout-link',
```
senadir: Would this matter here? would going from Store API to link break? I'd love as a follow up to include this link in Store API cart/checkout reponses.

plugins/woocommerce/changelog/integra-28-add-ability-to-generate-a-cart-from-link-or-load-up-via
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: enhancement
+
+Added new `checkout-link` endpoint which accepts a command separated list of product ids and a coupon code to prepopulate the cart and redirect to checkout with a new session. Allows for shareable carts. Supports simple products without additional options.
```
senadir:
```suggestion
Added new `checkout-link` endpoint which accepts a comma separated list of product ids and a coupon code to prepopulate the cart and redirect to checkout with a new session. Allows for shareable carts. Supports simple products without additional options.
```

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutLink.php
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	/**
+	 * Process the query params and return the checkout link to redirect to complete with session token.
+	 *
+	 * @return string The checkout link.
+	 */
+	protected function get_checkout_link() {
+		$controller = new CartController();
+		$controller->empty_cart();
```
senadir: Do we need to do this that early? Can't we wait until we validate that params are on the "correct" shape and possibly, having no errors?

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutLink.php
```diff
@@ -0,0 +1,128 @@
...
+				list( $product_id, $qty ) = explode( ':', $product_id_qty );
+			} else {
+				$product_id = $product_id_qty;
+				$qty        = 1;
+			}
+			if ( ! absint( $product_id ) ) {
+				continue;
+			}
+			try {
+				$controller->add_to_cart(
```
senadir, line 141: Correct me if I'm wrong but this should just work with variation product ids right? I will test regardless.

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutLink.php
```diff
@@ -0,0 +1,128 @@
...
+				return add_query_arg( 'wc_error', $empty_cart_notice, wc_get_cart_url() );
+			}
+
+			return wc_get_cart_url();
+		}
+
+		$redirect_url = wc_get_checkout_url();
+
+		// If the user is logged in, the session is tied to the user ID. Do not use a cart token.
+		if ( ! is_user_logged_in() ) {
```
senadir, line 195: 🤔 I agree that session is tied to the logged in user, I think we need a follow up here in which we still attach the cart token regardless, so that taking that url to a different browser will prompt you to login (redirect to login screen). 

We have some possible outcomes here:
- Cart token becomes powerful enough/trusted-enough that we just login the customer using the token. Continues from where their left in their browser.
- Cart token stays the same, we prompt the user to login at that screen, continue where you left.
- We duplicate the cart (A la Shopify) to that user continues from where they left, this would allow safer link sharing, and is how Shopify works. I think the logic of copying will need to be iterated slightly, includes keeping things like stock and possibly private products into account, cart item customization should ideally be copied, coupons should not. If cart cannot be fully copied, we prompt the user to login.

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutLink.php
```diff
@@ -0,0 +1,128 @@
...
+			$session_token = CartTokenUtils::get_cart_token( wc()->session->get_customer_id() );
+			$redirect_url  = add_query_arg( 'session', $session_token, $redirect_url );
```
senadir, line 197: One thing we discussed in slack is that instead of creating a new link, we should preserve query params, can we copy all query params to the new link? I think that's doable, if not, replace the `checkout-link` with `checkout` (or whatever is called), remove `products` and `coupon` and add `session`.

plugins/woocommerce/src/StoreApi/SessionHandler.php
```diff
@@ -78,13 +79,14 @@ public function get_session( $customer_id, $default = false ) {
 
 		$value = $wpdb->get_var(
 			$wpdb->prepare(
-				"SELECT session_value FROM $this->table WHERE session_key = %s", // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+				'SELECT session_value FROM %i WHERE session_key = %s',
```
senadir, line 82: [TIL](https://developer.wordpress.org/reference/classes/wpdb/prepare/#changelog)

plugins/woocommerce/includes/class-wc-session-handler.php
```diff
@@ -88,81 +161,97 @@ public function init() {
...
+		 * Note: If the filtered data is empty, the session will be destroyed and the
+		 * guest's session cookie will be removed. This can be useful for high-traffic
+		 * sites that prioritize page caching over maintaining all session data.
+		 *
+		 * @since 9.9.0
+		 *
+		 * @param array $session_data The session data loaded from storage.
+		 * @return array Modified session data to be used for initialization.
+		 */
+		$this->_data = apply_filters( 'woocommerce_restored_session_data', $this->get_session_data() );
```
senadir: I'm slightly torn here, we might have a privacy issue here. By loading the session as is, we risk exposing the customer address, this can be tricky in a sniffing attack because the token is in the URL and the URL is not private.

What Shopify does is copy the cart but leave out the address part, so copying the link in a new window only adds items. Modifying cart (adding more items) seems to generate a new cart/cart token, so those are not carried over.

Customer session is identified further via cookies it seems (not a part we would care about just yet).

My suggestion would be to do the following:
- Session restored via cookies get all the session.
- Session restored via request gets only items, possible fork:
    - Reading the token payload, token generated via Store API that get added as a header restores all the session (acts like cookies).
    - Token generated via redirect only copies cart.

This also possibly addresses my comment below(/above?) about copying cart for logged in customers.

plugins/woocommerce/includes/class-wc-session-handler.php
```diff
@@ -210,14 +330,14 @@ public function maybe_set_customer_session_cookie() {
 	 */
 	public function set_customer_session_cookie( $set ) {
 		if ( $set ) {
-			$to_hash           = $this->_customer_id . '|' . $this->_session_expiration;
-			$cookie_hash       = hash_hmac( 'md5', $to_hash, wp_hash( $to_hash ) );
-			$cookie_value      = $this->_customer_id . '||' . $this->_session_expiration . '||' . $this->_session_expiring . '||' . $cookie_hash;
-			$this->_has_cookie = true;
+			$cookie_hash  = $this->hash( $this->_customer_id . '|' . $this->_session_expiration );
+			$cookie_value = $this->_customer_id . '|' . $this->_session_expiration . '|' . $this->_session_expiring . '|' . $cookie_hash;
```
senadir, line 362: Any idea why `||` was used? Would it matter that we invalidate existing cookies? I think that's fine, they expire in 2 days anyway.

plugins/woocommerce/includes/class-wc-session-handler.php
```diff
@@ -449,19 +553,6 @@ public function forget_session() {
...
 	}
 
-	/**
-	 * When a user is logged out, ensure they have a unique nonce by using the customer/session ID.
-	 *
-	 * @deprecated 5.3.0
-	 * @param int $uid User ID.
-	 * @return int|string
-	 */
-	public function nonce_user_logged_out( $uid ) {
```
senadir, line 459: Couldn't find public usage of this function, but there's a hook with a similar name,

senadir (changes requested):
Thank you @mikejolley for a great PR. I left an initial review that involves mainly the code and the logic flow, I will follow up with a functional testing one.

I requested some changes that slightly increased the scope of this PR but that I deem necessary for now

plugins/woocommerce/includes/class-wc-session-handler.php
mikejolley: No valid reason. There is a separate issue for this, I just applied the fix here while I was touching the file. Can be addressed separately and rebased later.

plugins/woocommerce/includes/class-wc-session-handler.php
mikejolley: It was deprecated 4 years ago and replaced. Used to be hooked in.

plugins/woocommerce/includes/class-wc-session-handler.php
senadir: Fine to leave it

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutLink.php
```diff
@@ -0,0 +1,128 @@
...
+				wc_add_notice( $e->getMessage(), 'error' );
+			}
+		}
+
+		// Apply coupon if provided.
+		$coupon = wc_format_coupon_code( wp_unslash( $_GET['coupon'] ?? '' ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+
+		if ( wc_coupons_enabled() && ! empty( $coupon ) ) {
+			try {
+				$controller->apply_coupon( $coupon );
```
senadir, line 171: Can you add coupons to an empty cart? in my testing that wasn't possible, not providing products or providing invalid products ended up not adding the coupon, but it did't effectively empty the cart, adding a product manually would surface the coupon.

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutLink.php
```diff
@@ -0,0 +1,128 @@
...
+		if ( wc()->cart->is_empty() ) {
+			$empty_cart_notice = __( 'The provided checkout link was out of date or invalid. No products were added to the cart.', 'woocommerce' );
+			wc_add_notice( $empty_cart_notice, 'error' );
+
+			if ( ! wc()->session->has_session() ) {
+				return add_query_arg( 'wc_error', $empty_cart_notice, wc_get_cart_url() );
+			}
+
+			return wc_get_cart_url();
+		}
```
senadir: At some point I noticed several notices, likely left over from not clearing up the session? 
image

senadir (commented):
Took this through a round of manual testing and it works very well, thank you for adding E2E and PHP tests, the code you wrote is thoughtful, and expansive and safe, I left some comments in my previous technical review that I hope we figure a path for, but the main concept and process stays the same.

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutLink.php
mikejolley: Maybe after a small check for required params, but ultimately given this is a very deliberate endpoint you cannot reach naturally I don't think it matters so much.

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutLink.php
mikejolley: It should work with anything that does not have special options. You'll get notices if you try to add a variable product. Not sure if we want to add a more deliberate check here.

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutLink.php
mikejolley: https://github.com/woocommerce/woocommerce/pull/58140/commits/abd0e06da54cbd16e9cf4390cef8ccac911da4f2

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutLink.php
mikejolley: We can swap it so it adds coupon after checking cart is empty. Done in https://github.com/woocommerce/woocommerce/pull/58140/commits/18d7493cd40885ea849cfeb261c86a100bd8614b

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutLink.php
mikejolley: https://github.com/woocommerce/woocommerce/pull/58140/commits/0b6e17a969df7c67767e2e06f3a1722050b53469

plugins/woocommerce/src/Blocks/Domain/Services/CheckoutLink.php
mikejolley: I think its due to no session. https://github.com/woocommerce/woocommerce/pull/58140/commits/fe7bdda6e6901bc589be0ed69a1abd01e0442867 should clear it up.

plugins/woocommerce/includes/class-wc-session-handler.php
mikejolley: We should discuss this Monday. Its possible to have an allowlist of date to restore from session, but that means data from the cookie session would be ignored so some stuff like address would be empty every time you enter checkout.

mikejolley: Excludes customer data when restoring from token https://github.com/woocommerce/woocommerce/pull/58140/commits/9f8f33892040cd6d4b515973f64e86b611483404

plugins/woocommerce/src/StoreApi/SessionHandler.php
opr: @straku just pointing out this is cool, it could have been used on Points and rewards 👀

plugins/woocommerce/tests/e2e-pw/tests/checkout/checkout-link.spec.js
```diff
@@ -0,0 +1,172 @@
...
+					'The provided checkout link was out of date or invalid. No products were added to the cart.'
+				)
+			).toBeVisible();
+
+			// Cart should be empty
+			await expect(
+				page.getByText( 'Your cart is currently empty!' )
+			).toBeVisible();
+		}
+	);
```
mikejolley: Add test for logged in user

plugins/woocommerce/includes/class-wc-session-handler.php
```diff
@@ -6,11 +6,14 @@
  * From 2.5 this uses a custom table for session storage. Based on https://github.com/kloon/woocommerce-large-sessions.
  *
  * @class    WC_Session_Handler
- * @version  2.5.0
  * @package  WooCommerce\Classes
  */
 
+declare(strict_types=1);
```
mikejolley, line 12: Check this

plugins/woocommerce/includes/class-wc-session-handler.php
```diff
@@ -210,14 +358,14 @@ public function maybe_set_customer_session_cookie() {
 	 */
 	public function set_customer_session_cookie( $set ) {
 		if ( $set ) {
-			$to_hash           = $this->_customer_id . '|' . $this->_session_expiration;
-			$cookie_hash       = hash_hmac( 'md5', $to_hash, wp_hash( $to_hash ) );
-			$cookie_value      = $this->_customer_id . '||' . $this->_session_expiration . '||' . $this->_session_expiring . '||' . $cookie_hash;
-			$this->_has_cookie = true;
+			$cookie_hash  = $this->hash( $this->_customer_id . '|' . $this->_session_expiration );
+			$cookie_value = $this->_customer_id . '|' . $this->_session_expiration . '|' . $this->_session_expiring . '|' . $cookie_hash;
```
mikejolley, line 362: Double pipe handling

plugins/woocommerce/includes/class-wc-session-handler.php
```diff
@@ -340,15 +474,15 @@ public function get_customer_unique_id() {
...
 	public function get_session_cookie() {
-		$cookie_value = isset( $_COOKIE[ $this->_cookie ] ) ? wp_unslash( $_COOKIE[ $this->_cookie ] ) : false; // @codingStandardsIgnoreLine.
+		$cookie_value = isset( $_COOKIE[ $this->_cookie ] ) ? wc_clean( wp_unslash( (string) $_COOKIE[ $this->_cookie ] ) ) : '';
 
-		if ( empty( $cookie_value ) || ! is_string( $cookie_value ) ) {
+		if ( empty( $cookie_value ) ) {
 			return false;
 		}
 
-		$parsed_cookie = explode( '||', $cookie_value );
```
mikejolley, line 349: Double pipes

opr (approved):
We pair reviewed this today, highlighting that I mentioned the following concerns:

- Cookies that existed before 10.0 would have double pipes, so would break if WC was upgraded during a session.
- Check the `declare(strict_types=1);` added to `WC_Session_Handler` since it could cause bugs (e.g. https://github.com/woocommerce/woocommerce/pull/56797)
- Mentioned that sharing a cart URL and keeping the same session could be a valid use-case (e.g. collaborative shopping) but noted it is out of scope for this work.
- Double checked the out of stock, sold individually, product doesn't exist etc. warnings show if invalid products are added via checkout-link.
- E2E tests are necessary and valid here, I'd say this is a critical flow.
- Try to run the same E2E tests as logged in and logged out user.

The changes made in the session handler look good.

Some things we may want to add tests for:

- There doesn't seem to be explicit tests for `init_session_from_request` 
- There doesn't seem to be explicit tests for session cloning when existing cookie sessions conflict
- What happens when `CartTokenUtils::validate_cart_token()` or `get_cart_token_payload()` fail? (I think the main fail point would be in `get_parts`) should we harden that?
- No test for the || to | cookie format migration. Probably not important, since we're removing it next version anyway.
`wp_fast_hash` fallback logic not tested, is it important?

Approving, because it's working for me, and the code looks good, maybe the tests I suggested can be added as a follow up, or if you add them in here dismiss my review and request a new one 👌🏼

senadir:
> Mentioned that sharing a cart URL and keeping the same session could be a valid use-case (e.g. collaborative shopping) but noted it is out of scope for this work.

A valid case still doesn't justify the risk behind it, URLs are not private and can be spoofed over the network regardless of SSH, if we're to allow this as is, it needs to be a single use link.

I will be doing another review today and hopfully we can merge this soon.

opr:
> A valid case still doesn't justify the risk behind it, URLs are not private and can be spoofed over the network regardless of SSH, if we're to allow this as is, it needs to be a single use link.

Agree, not saying we should include it now, but it is definitely a feature that could be useful to shoppers.

senadir (approved):
Thank you Mike! I'm going ahead and approving this as per our discussion, a follow up PR will make the token handling more resilient and solve the replay usage issue we talked about.

senadir:
Follow up feedback we got from X:

- Allow `checkout-link` to be changed, like we do with other endpoints in Woo. My opinion on this is that we should look into moving this logic into `checkout` endpoint, and if we can't or it doesn't make sense, and we settle on a dedicated endpoint, at that point we make it editable in admin.

mikejolley:
@senadir We could piggy back on checkout if it runs early enough. This URL is redirected away from once the cart is populated anyway.

senadir:
I prefer we do that, will add it to the follow up PR.

alewolf:
@mikejolley May I ask what page type this will be? I couldn't find anything in the PR. Will `is_checkout()` return `true`? Or will this be another page type entirely?

I'm asking as I want to implement this with my [Pixel Manager for WooCommerce](https://wordpress.org/plugins/woocommerce-google-adwords-conversion-tracking-tag/) and trigger an `add-to-cart` event when that specific URL with those parameters is loaded.

mikejolley:
@alewolf There is no page type; its a redirect early during `template_redirect`. `is_checkout()` won't work here. Any add-to-cart actions would still fire, since this uses standard add to cart methods to populate the cart.

alewolf:
@mikejolley 
> Any add-to-cart actions would still fire, since this uses standard add to cart methods to populate the cart.

Does that include front-end JavaScript/jQuery `add-to-cart` events?

mikejolley:
@alewolf Server side `woocommerce_add_to_cart`. The redirect happens too early for JavaScript.

dabesa:
How does this differ from the existing add-to-cart URL parameter?
es:
https://store.local/checkout/?add-to-cart=233

helgatheviking:
>Any add-to-cart actions would still fire, since this uses standard add to cart methods to populate the cart

How would you support customizable products, like product addons, bundles, name your price, etc
https://store.local/checkout-link/?products=18%3A2%2C19%3A1&coupon=TEST

For example if product 18 is name your price how would I support passing an additional parameter for price?

sverleis:
> How does this differ from the existing add-to-cart URL parameter? es: https://store.local/checkout/?add-to-cart=233

From my understanding, one element that this has is the ability to add coupons, which the add-to-cart feature lacks. This can also add multiple different products, rather than just 1.

pacotole:
Please, allow multiple coupons to be added.

LuigiPulcini:
> How would you support customizable products, like product addons, bundles, name your price, etc

@helgatheviking About your question – which I am interested in as well –, a viable approach would be to add to the cart page a "Share this cart" custom button that generates the cart token for the current cart. This method also simplifies the creation of the link for complex cart contents, so you don't have to compile a URL manually.

The process would be this:
* you visit your website as a guest (this is necessary because of [this check](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/class-wc-session-handler.php#L127-L129));
* you add any products to the cart, configuring each product with the appropriate options/metadata;
* once you are satisfied with the cart content, you click the "Share this cart" button and use that URL.

I created a proof of concept with the code snippet below and it works on the Classic Cart.

```php
add_action( 'woocommerce_proceed_to_checkout', 'dev_test_woocommerce_proceed_to_checkout', 21 );
function dev_test_woocommerce_proceed_to_checkout() {
	if ( ! class_exists( 'Automattic\WooCommerce\StoreApi\Utilities\CartTokenUtils' ) ) {
		return;
	}
	
	$session_token     = Automattic\WooCommerce\StoreApi\Utilities\CartTokenUtils::get_cart_token( (string) wc()->session->get_customer_id() );
	$checkout_link_url = add_query_arg( 'session', $session_token, wc_get_checkout_url() );
	?>
		<a href="<?php echo esc_url( $checkout_link_url ); ?>" class="button alt<?php echo esc_attr( wc_wp_theme_get_element_class_name( 'button' ) ? ' ' . wc_wp_theme_get_element_class_name( 'button' ) : '' ); ?>">
			<?php esc_html_e( 'Share this cart', 'woocommerce' ); ?>
		</a>
	<?php
}
```

helgatheviking:
@LuigiPulcini that's clever, I like it! Is the session token limited to the user who made the link? Or you can send it to anybody?

I also took a pass at adding support for [additional params to the checkout link in this PR](https://github.com/woocommerce/woocommerce/pull/59341). I was able to get it working with my name your price plugin (so you can add a product with a custom price to the cart via url), but another of my plugins is a bit more complex and would require more integration.

LuigiPulcini:
> Is the session token limited to the user who made the link? Or you can send it to anybody?

@helgatheviking As long as the user creating the cart is a guest (i.e. you are not logged in), the cart token can be shared to anyone. You can easily verify that by creating a link and use it in a different browser or computer. The check I mentioned in my previous comment bails out if the user who created the cart is a registered user of the website.

@pacotole The technique I described [here](https://github.com/woocommerce/woocommerce/pull/58140#issuecomment-3078547820) also allows the use of multiple coupons because the cart is captured as a whole via its cart token and that includes any coupons added to it, no matter how many.

LuigiPulcini:
@helgatheviking About your other PR, I think that manually adding fully-configured products to a cart and generating a cart token for it is much less laborious than any attempt at compiling the URL, even when that is done programmatically (which would require additional code anyway).

Leveraging WooCommerce's built-in capability of storing guest carts in the database makes even more sense now that we have the `CartTokenUtils` class and the updated session handler.

helgatheviking:
I'm inclined to agree with you, esp if there's a way to support generating a guest token even while logged in- since I'm guessing store owners would want to generate links for customers.

but I think we could do both. If the checkout url feature exists then it ought to be extensible and support plugins. I'm pretty much always on that soapbox.

KPYahoo:
@LuigiPulcini I am not a developer. May I request you to explore some option that this could be seen / created ONLY by Admin and not been seen by all customers in the front-end? Do you think this can be done?

Edit: I am a store owner and we have been craving this since years.

LuigiPulcini:
@helgatheviking I agree that the perfect method should simplify how site owners can create links for customers. The reason behind the use of guest users is that the `wp_woocommerce_sessions` table in the database can only have a single row per registered user because the `session_key` field contains the user ID. With guest users, you have a virtually unlimited number of cart sessions that you can create and share. Other approaches could be evaluated, but the key point would be trying to leverage as much as possible this new feature.

@KPYahoo A simple way to hide the button from regular visitors is adding a query argument (e.g. `?share`) and checking if that argument is set with `isset( $_GET['share'] )` before printing the button on the page. This way, the added button will be visible only if you visit the cart page with the URL `www.example.com/cart/?share`.

helgatheviking:
@LuigiPulcini what happens in your snippet if you do not pass a customer ID when generating the token? 

And instead you generate the random hash yourself? 
`wc_rand_hash( 't_', 30 );`

[Which is how the customer ID is generated for non -logged in users.
](https://github.com/woocommerce/woocommerce/blob/1d48db92ad0c5393712d5e78e584c1444c6ccc47/plugins/woocommerce/includes/class-wc-session-handler.php#L545) if I'm reading it right

Which would also allow for restricting display to admins. I have an idea but I'm away from my keyboard. 😅

LuigiPulcini:
@helgatheviking As long as you pass a customer ID that:
1. is present in the `session_key` column of the `wp_woocommerce_sessions` table, **AND**
2. corresponds to a guest user

then, the new `WC_Session_Handler::init_session_from_request()` method will be able to load its session data and clone it into the current session.

The problem, though, is how to make it usable from the currently logged-in admin user which doesn't store its session data as a guest user in the `wp_woocommerce_sessions` table. So, my idea for a more advanced snippet – something I am currently experimenting with – is this:
1. Load the session data of the current user (even logged-in ones)
2. Clone the session data of the current user to a new session after generating a random guest customer ID with `wc_rand_hash( 't_', 30 )`
3. Create a cart token for the cloned session (which is now assigned programmatically to a guest customer)
4. Print the button with the redirection URL for that cart token

This should allow us to effortlessly create checkout links while being logged in as admins and limit the presence of the button with a check to `current_user_can()`.

LuigiPulcini:
@helgatheviking It works! The following code snippet adds a button that is visible to admins or shop managers only and allows them to generate checkout links from the cart page. The link can be shared with any customers and is valid for 48 hours.

```php
add_action( 'woocommerce_proceed_to_checkout', 'dev_test_woocommerce_proceed_to_checkout', 21 );
function dev_test_woocommerce_proceed_to_checkout() {
	if ( ! class_exists( 'Automattic\WooCommerce\StoreApi\Utilities\CartTokenUtils' ) || ! current_user_can( 'manage_woocommerce' ) ) {
		return;
	}

	global $wpdb;
	$customer_id   = wc()->session->get_customer_id();
	$session_data  = wc()->session->get_session( $customer_id, array() );
	$session_data  = array_diff_key( $session_data, array( 'customer' => true ) );
	$guest_user_id = wc_rand_hash( 't_', 30 );
	$expiration    = time() + intval( apply_filters( 'wc_session_expiration', 2 * DAY_IN_SECONDS ) );

	$wpdb->query(
		$wpdb->prepare(
			'INSERT INTO %i (`session_key`, `session_value`, `session_expiry`) VALUES (%s, %s, %d)
						ON DUPLICATE KEY UPDATE `session_value` = VALUES(`session_value`), `session_expiry` = VALUES(`session_expiry`)',
			$wpdb->prefix . 'woocommerce_sessions',
			$guest_user_id,
			maybe_serialize( $session_data ),
			$expiration
		)
	);
	wp_cache_set( WC_Cache_Helper::get_cache_prefix( WC_SESSION_CACHE_GROUP ) . $guest_user_id, $session_data, WC_SESSION_CACHE_GROUP, $expiration - time() );

	$session_token     = Automattic\WooCommerce\StoreApi\Utilities\CartTokenUtils::get_cart_token( (string) $guest_user_id );
	$checkout_link_url = add_query_arg( 'session', $session_token, wc_get_checkout_url() );
	?>
		<a href="<?php echo esc_url( $checkout_link_url ); ?>" class="button alt<?php echo esc_attr( wc_wp_theme_get_element_class_name( 'button' ) ? ' ' . wc_wp_theme_get_element_class_name( 'button' ) : '' ); ?>">
			<?php esc_html_e( 'Share this cart', 'woocommerce' ); ?>
		</a>
	<?php
}
```

Please note that a new guest session data row is created every time the cart page gets reloaded. Nevertheless, all those sessions will eventually disappear from the table when their expiration time comes up.

helgatheviking:
Awesome! I totally forgot you would need to save the session token to the DB

sverleis:
> @KPYahoo : May I request you to explore some option that this could be seen / created ONLY by Admin and not been seen by all customers in the front-end? Do you think this can be done?
Edit: I am a store owner and we have been craving this since years.

Created a plugin that for now can do simple and variable product. See if it works for you, and leave a review :)

https://wordpress.org/plugins/link-wizard-for-woocommerce/

Issue #56289, opened 2025-03-11
Add to Cart with Options block: not visible in the frontend when selecting a specific variation in the Single Product block


Aljullu:
The Add to Cart with Options block is not visible in the frontend when selecting a specific variation in the Single Product block. That affects the non-blockified and the blockified Add to Cart with Options blocks.

## Steps to reproduce

1. Create a post or page.
2. Add the _Single Product_ block.
3. Select a specific product variation.

![Image](https://github.com/user-attachments/assets/cef178d5-c5fe-48ab-8338-8a2060f6f80b)

4. Notice the _Add to Cart with Options_ block renders correctly in the editor:

![Image](https://github.com/user-attachments/assets/3cb8c4d2-0052-46e8-ac8b-f31c3af4d25a)

5. Visit the post or page in the frontend and notice the _Add to Cart with Options_ block isn't rendered at all.

![Image](https://github.com/user-attachments/assets/16a0946d-e6e1-4ddc-8069-d53c1dcdf519)

### Expected behavior

I would expect one of these two options:

1. The Add to Cart with Options block should work in the frontend when a specific variation is selected or
2. The Add to Cart with Options block should display a notice in the editor mentioning that it won't render anything in the frontend for specific variations.

Kallyan01:
I think this issue is a similer to https://github.com/woocommerce/woocommerce/issues/56846

Aljullu:
Thanks, @Kallyan01! I think they are a bit different, this one is about the form not displaying at all when a specific variation is selected, while the other one is about a visual glitch in some themes when the form is displayed.

PraveenKum11:
Hey, I was able to replicate the issue and traced the root cause to the AddToCartForm.php file. The issue arises due to the dynamic hook name generation using:

`do_action('woocommerce_' . $product->get_type() . '_add_to_cart');`

This works as expected for most product types. For example, a simple product results in the hook `woocommerce_simple_add_to_cart`, which is correctly handled. However, for a product of type `variation`, this generates `woocommerce_variation_add_to_cart`, which doesn't exists - leading to no output or functionality.

Proposed Fix:
I've added a new hook, `woocommerce_variation_add_to_cart`, and mapped it to call the existing `woocommerce_simple_add_to_cart()` function. Since a single product variation is essentially a simple product with additional metadata. After this change, the add to cart UI and functionality behave correctly for variations as well. Here is a before and after comparison.

Before:

[...]

Issue #53014, opened 2024-11-20
[Epic]: Blockify Add to Cart with Options block


octaedro:
### Description

This epic encompasses all tasks related to blockifying the `Add to Cart Form`.

* [GitHub Board 1](https://github.com/orgs/woocommerce/projects/284)
* [GitHub Board 2](https://github.com/orgs/woocommerce/projects/294)
* [GitHub Board 3](https://github.com/orgs/woocommerce/projects/305)

helgatheviking:
Following!

helgatheviking:
Would you want to add asynchronous add to cart? I think you'd need to add support to the [`addToCart` function](https://github.com/woocommerce/woocommerce/blob/25fb13a2bf44fe92fef2e764adee53899422cb63/plugins/woocommerce-blocks/assets/js/data/cart/thunks.ts#L204-L214) for more parameters. Other product types will have more fields than ID and Quantity. I opened https://github.com/woocommerce/woocommerce/issues/53357 about this.

I would also like to suggest that the form have interactivity that includes some kind of validation.... so plugins can tell the form it's not valid and therefore should not be add to the cart. This will be needed by variable products anyway so it would be handy to have it accessible from other plugins from start. For example, in Name Your Price the add to cart button prevents submitting the form if the entered price is below the minimum. Same would be needed for Product Addons that are required... etc.

Aljullu:
As usual, thanks for the feedback, @helgatheviking!

> Would you want to add asynchronous add to cart? I think you'd need to add support to the [addToCart function](https://github.com/woocommerce/woocommerce/blob/25fb13a2bf44fe92fef2e764adee53899422cb63/plugins/woocommerce-blocks/assets/js/data/cart/thunks.ts#L204-L214) for more parameters. Other product types will have more fields than ID and Quantity. I opened https://github.com/woocommerce/woocommerce/issues/53357 about this.

I think internally we will likely use the same Store API endpoint to handle Add to Cart events, but via the Interactivity API.

For 3rd-party extensions that only register some fields in the form, I hope we can make this transparent so they don't need to hook into that function or anything similar. IMO we should ideal detect the `name` property of every input and send it with the request.

For 3rd-party extensions that create new product types, I think it's still too early to say how it will look like. I still expect the wrapper block will be shared and hopefully that block will handle most of the logic of Adding to Cart. But I expect we will need to work on a few extensions until we find a flexible API that doesn't require extensions to rewrite the same logic over and over again.

[...]

PR #58148, opened 2025-05-21
Fix Add to Cart with Options block visibility on frontend when selecting a specific variation in the Single Product block

PraveenKum11:

Adds a new hook, `woocommerce_variation_add_to_cart`, which maps to the existing `woocommerce_simple_add_to_cart()` function. This hook is available only within the "Add to Cart with Options" block to avoid potential conflicts with existing code or third-party extensions.

This implementation works because a single product variation is essentially a simple product with additional metadata. As a result, the add-to-cart UI and functionality now behave correctly for variable products as well.

Part of #56289

Before

https://github.com/user-attachments/assets/abb0ca86-3fa9-482e-b415-d5b8c74c730d

After

https://github.com/user-attachments/assets/7bd8708d-48fb-41ea-9106-adf586024071

1. Create a post or page.
2. Add the Single Product block.
3. Select a specific product variation.
image

5. Save the post or page.
6. Visit the post or page in the frontend and ensure that the Add to Cart with Options blocks renders and functions properly on the frontend.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


Aljullu (commented):
Thanks for working on this @PraveenKum11! Overall it looks great. I only noticed one issue if the variation accepts 'Any' as one of the attributes:

imatge

In that case, adding to cart doesn't work because one attribute is missing. I think if we detect the selected variation has one attribute not defined (marked as 'any'), we shouldn't render the form as we were doing until now. In other words, we only render the form in variations where all attributes have a defined value. How does it sound?

> Closes https://github.com/woocommerce/woocommerce/issues/56289

I'm fine not doing it in this PR, but in order to mark #56289 as completed, we would also need to fix this in the blockified _Add to Cart + Options_ block (you can enable it following [these steps](https://github.com/woocommerce/woocommerce/issues/53014#issuecomment-2821133682)). So I would mark this PR as:

> Part of https://github.com/woocommerce/woocommerce/issues/56289.

PraveenKum11:
@Aljullu
> In that case, adding to cart doesn't work because one attribute is missing. I think if we detect the selected variation has one attribute not defined (marked as 'any'), we shouldn't render the form as we were doing until now. In other words, we only render the form in variations where all attributes have a defined value. How does it sound?

Great catch! I will try to fix it.

> we would also need to fix this in the blockified Add to Cart + Options block

Makes sense, I will try fixing it.

PraveenKum11:
Hey @Aljullu,

Apologies for the delay, got caught up with some other work.

> In that case, adding to cart doesn't work because one attribute is missing. I think if we detect the selected variation has one attribute not defined (marked as 'any'), we shouldn't render the form as we were doing until now

I've managed to fix this: if a block has any attribute set to "any", the Add to Cart form will no longer render for that block.

However, I wasn't able to address the issue with the blockified Add to Cart + Options block, so I've marked this PR as "Part of", as you suggested.

Let me know if you'd like any further changes!

Aljullu (approved):
Thanks for the updates, @PraveenKum11!  I just tested and it works as described, if the variation has an attribute which allows 'any' value, the Add to Cart Form doesn't show up, otherwise, it's rendered as expected:

Variation with 'any' | Variation without 'any'
--- | ---
imatge | imatge

<details>

* Theme: Twenty Twenty-Five
* Store size: small
</details>

Issue #57959, opened 2025-05-13
Add back button to the email editor


costasovo:
In this task, we want to add a back button to the email editor. The black cross in the editor 

<img src="https://uploads.linear.app/f629fcde-6f9b-48e8-8af3-60eed8823bff/7cff5424-8150-465a-8f38-3fb4784b0b01/dc088ebc-ba16-457c-a2ee-ae8f0cad9e04?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2Y2MjlmY2RlLTZmOWItNDhlOC04YWYzLTYwZWVkODgyM2JmZi83Y2ZmNTQyNC04MTUwLTQ2NWEtOGYzOC0zZmI0Nzg0YjBiMDEvZGMwODhlYmMtYmExNi00NTdjLWEyZWUtYWU4ZjBjYWQ5ZTA0IiwiaWF0IjoxNzQ3MTQ1Mzk4LCJleHAiOjMzMzE3NzA1Mzk4fQ.a1f_GuV1N5wuhXmWFyGF8hiyibypqDnc3xQHkNB_JUI " alt="main-dashboard-button-close-icon-example.webp" width="859" data-linear-height="375" />

Use the same background as in the site editor. By default, the icon is `wordpress` and it closes Fullscreen mode. It is displayed only in full-screen mode. This aligns with the core and should make the next migration step to the post-editor easier.<br><br>The third party can replace the icon and the callback function. <br><br>For WooCommerce integration:

* We want to enforce the full-screen mode (via config or filter) so that the back button is always displayed
* The icon will be `arrowLeft`
* In the action we will redirect to `wp-admin/admin.php?page=wc-settings&tab=email`
* We will also need to change where the validation happens. We want to trigger it on save.
* Remove the "Save email" button. We want to keep only the Save button. When all changes are save the Save button has disabled state.

PR #58154, opened 2025-05-21
Add back button to the email editor

pavel-mailpoet:

This adds a back button to the WooCommerce Email Editor, allowing navigation back to the email listings page from the editor. The back button appears when the editor is in full-screen mode, which is now enforced by default.

- Added a new `BackButtonContent` component that renders a back button
- Updated the editor component to conditionally render the back button when in full-screen mode
- Modified `PageRenderer.php` to force full-screen mode by adding the isFullScreenForced setting

Closes WOOPLUG-4268 #57959.

| Before | After |
| ------ | ----- |
|    Screenshot 2025-05-21 at 10 14 41   |   Screenshot 2025-05-21 at 10 15 01    |

1. Navigate to the WooCommerce Email Editor
2. Verify that the editor opens in full-screen mode
3. Check that the back button appears in the top-left corner
4. Hover over the back button to see the animation effect
5. Click the back button to confirm it navigates back to the email listings page

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/packages/js/email-editor/src/components/block-editor/editor.tsx
+++ b/packages/js/email-editor/src/components/block-editor/editor.tsx
@@ -69,7 +69,7 @@ export function InnerEditor( {
 						? getEditedPostTemplate()
 						: null,
 				post: postObject,
-				isFullscreenEnabled: true,
+				isFullscreenEnabled: settings.fullScreen,
 			};
 		},
 		[ currentPost.postType, currentPost.postId ]
--- a/packages/js/email-editor/src/components/block-editor/editor.tsx
+++ b/packages/js/email-editor/src/components/block-editor/editor.tsx
@@ -54,30 +54,35 @@ export function InnerEditor( {
 		'post-only'
 	);
 
-	const { post, template, isFullScreenForced } = useSelect(
-		( select ) => {
-			const { getEntityRecord } = select( coreStore );
-			const { getEditedPostTemplate } = select( storeName );
-			const postObject = getEntityRecord(
-				'postType',
+	// isFullScreenForced – comes from settings and cannot be changed by the user
+	// isFullscreenEnabled – indicates if a user has enabled fullscreen mode
+	const { post, template, isFullScreenForced, isFullscreenEnabled } =
+		useSelect(
+			( select ) => {
+				const { getEntityRecord } = select( coreStore );
+				const { getEditedPostTemplate } = select( storeName );
+				const postObject = getEntityRecord(
+					'postType',
+					currentPost.postType,
+					currentPost.postId
+				);
+				return {
+					template:
+						currentPost.postType !== 'wp_template'
+							? getEditedPostTemplate()
+							: null,
+					post: postObject,
+					isFullscreenEnabled:
+						select( storeName ).isFeatureActive( 'fullscreenMode' ),
+					isFullScreenForced: settings.isFullScreenForced,
+				};
+			},
+			[
 				currentPost.postType,
-				currentPost.postId
-			);
-			return {
-				template:
-					currentPost.postType !== 'wp_template'
-						? getEditedPostTemplate()
-						: null,
-				post: postObject,
-				isFullScreenForced: settings.isFullScreenForced,
-			};
-		},
-		[
-			currentPost.postType,
-			currentPost.postId,
-			settings.isFullScreenForced,
-		]
-	);
+				currentPost.postId,
+				settings.isFullScreenForced,
+			]
+		);
 
 	// @ts-expect-error Type is missing in @types/wordpress__editor
 	const { removeEditorPanel } = useDispatch( editorStore );
@@ -135,7 +140,9 @@ export function InnerEditor( {
 					<TemplateSelection />
 					<StylesSidebar />
 					<SendPreview />
-					<FullscreenMode isActive={ isFullScreenForced } />
+					<FullscreenMode
+						isActive={ isFullScreenForced || isFullscreenEnabled }
+					/>
 					{ isFullScreenForced && <BackButtonContent /> }
 					{ ! isFullScreenForced && <MoreMenu /> }
 					{ currentPost.postType === 'wp_template' ? (
--- a/packages/js/email-editor/src/components/block-editor/editor.tsx
+++ b/packages/js/email-editor/src/components/block-editor/editor.tsx
@@ -72,7 +72,7 @@ export function InnerEditor( {
 				isFullscreenEnabled: settings.fullScreen,
 			};
 		},
-		[ currentPost.postType, currentPost.postId ]
+		[ currentPost.postType, currentPost.postId, settings.fullScreen ]
 	);
 
 	// @ts-expect-error Type is missing in @types/wordpress__editor
--- a/packages/js/email-editor/src/components/block-editor/editor.tsx
+++ b/packages/js/email-editor/src/components/block-editor/editor.tsx
@@ -54,7 +54,7 @@ export function InnerEditor( {
 		'post-only'
 	);
 
-	const { post, template, isFullscreenEnabled } = useSelect(
+	const { post, template, isFullScreenForced } = useSelect(
 		( select ) => {
 			const { getEntityRecord } = select( coreStore );
 			const { getEditedPostTemplate } = select( storeName );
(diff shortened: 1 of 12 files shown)


packages/js/email-editor/src/hooks/use-content-validation.ts
```diff
@@ -38,35 +73,39 @@ export const useContentValidation = (): ContentValidationData => {
...
 		removeValidationNotice,
 		hasValidationNotice,
-		rules,
 	] );
 
+	useEffect( () => {
+		const filterHandler = async ( edits: Record< string, unknown > ) => {
+			const isValid = validateContent();
+			if ( ! isValid ) {
+				throw new Error();
```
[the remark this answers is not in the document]
pavel-mailpoet: There should already be a message for the user. This error is to stop saving the email, it doesn't need an error message.

packages/js/email-editor/src/components/block-editor/editor.tsx
```diff
@@ -53,27 +54,35 @@ export function InnerEditor( {
...
+				);
+				return {
+					template:
+						currentPost.postType !== 'wp_template'
+							? getEditedPostTemplate()
+							: null,
+					post: postObject,
+					isFullscreenEnabled:
+						select( storeName ).isFeatureActive( 'fullscreenMode' ),
+					isFullScreenForced: settings.isFullScreenForced,
```
costasovo: `isFullScreenForced` doesn't need to be in `useSelect`.
It is not fetched from the store, but just passed via the `settings` prop so it can be accessed directly.
So you can do just `const { isFullScreenForced } = settings;` outside the useSelect.

packages/js/email-editor/src/components/block-editor/editor.tsx
```diff
@@ -131,14 +140,17 @@ export function InnerEditor( {
...
+					/>
+					{ isFullScreenForced && <BackButtonContent /> }
+					{ ! isFullScreenForced && <MoreMenu /> }
 					{ currentPost.postType === 'wp_template' ? (
 						<TemplateSettingsPanel />
 					) : (
 						<SettingsPanel />
 					) }
-					<PublishSave />
+					{ ! isFullScreenForced && <PublishSave /> }
```
costasovo: If I'm not mistaken this change means that when the isFullScreenForced is true we would completely hide the button.

This works for Woo, but in MailPoet we will still want to display the Send button. I think the logic must be handled differently.

Here is my perspective: 
We don't want to hide the <PublishSave /> because fullscreen mode is forced, but because we don't want and don't need to display the custom button for Woo transactional emails.
So I think this should be a different setting or perhaps a filter. That would allow to disable this button for transactional email, but we could still use it when we add support for more email types to Woo.

packages/js/email-editor/src/components/header/style.scss
```diff
@@ -10,3 +10,83 @@
 		display: none;
 	}
 }
+
+.woocommerce-email-editor__view-mode-toggle {
```
costasovo, line 14: Could you please share more details on why we need to use the custom styles and not just reuse styles for the post editor we already load?
I noticed a commit message about some problems with colors.
I ran into some problems with colors recently. The preset vars were not defined outside the editor canvas. I fixed the issue in [the PR with UI fixes](https://github.com/woocommerce/woocommerce/pull/58078) that was merged a few day ago. So perhaps if the problem was the same, we might not need the CSS 🙏

costasovo (commented):
@pavel-mailpoet Thanks for adding the functionality! I like the idea of using the editor settings.

At this point, I focused only on the code. I added some comments. I'll test the functionality when we resolve the comments.

packages/js/email-editor/src/components/block-editor/editor.tsx
pavel-mailpoet: I will update the code. But can you explain the reason for moving it outside? Is there a downside to doing it this way? Does it cause any performance issues or any other issues?

packages/js/email-editor/src/components/block-editor/editor.tsx
pavel-mailpoet: > If I'm not mistaken this change means that when the isFullScreenForced is true we would completely hide the button.

Yes, that is correct

> This works for Woo, but in MailPoet we will still want to display the Send button. I think the logic must be handled differently.

Hmm, I assumed that the forced full-screen mode is for WooCommerce, but in MailPoet, we wouldn't use it, and the Send button would be displayed. Is it not the case? I can use a different setting or filter; that's not a problem.

costasovo: > Hmm, I assumed that the forced full-screen mode is for WooCommerce, but in MailPoet, we wouldn't use it, and the Send button would be displayed. Is it not the case? 

Yes, this currently works for the two projects we manage ourselves. But we are building the editor to be consumed in other projects. The current logic is: When you force full screen, you are not able to add the send button.  This feels confusing to me. 

In the send button we have a [filter that allows adding a custom callback](https://github.com/woocommerce/woocommerce/blob/trunk/packages/js/email-editor/src/components/header/send-button.tsx#L53-L56), and if the callback is not set, we have a default callback that redirects to [`url.send` from settings](https://github.com/woocommerce/woocommerce/blob/dd824c38d82654ba176537c78ccfedb8a0143bed/packages/js/email-editor/src/components/header/send-button.tsx#L34-L36). So, we might not add any new filters, but we could check if there is an action or send url. If there are none we would not display the button. And then for Woo we unset the `url.send` because we don't need the send url in Woo. I would find this logic cleaner. I woudl implement those checks in the PublishSave component.

pavel-mailpoet: That makes sense, thank you @costasovo for the explanation. 


I prefer to control the button using the settings, but if you strongly prefer a filter, I don't mind.

packages/js/email-editor/src/components/header/style.scss
pavel-mailpoet: I am testing it now, but with the default editor classes, the CSS does not load properly: https://d.pr/i/jwViee

I'm not sure what I'm missing, but that's why I used the custom styles. 

This is the screenshot from the site editor: https://d.pr/i/pcjyfB

packages/js/email-editor/src/components/block-editor/editor.tsx
costasovo: I don't have a strong preference. Please choose what you think fits the best.

packages/js/email-editor/src/components/header/style.scss
costasovo: Aha. We load CSS for the post editor. There are still some differences. Could you please check the markup in the post editor? I see it [uses](https://github.com/WordPress/gutenberg/blob/f995b6c84d203002e4c7c4cc46e6a95204a1e0c9/packages/edit-post/src/components/back-button/fullscreen-mode-close.js#L79) `edit-post-fullscreen-mode-close` and `has-icon` classes. Maybe these could help.

pavel-mailpoet: I also tried the post editor. Unfortunately, the behaviour is different. The icon turns blue rather than swapping it for the arrow, so I couldn't reuse the markup.

costasovo: Ok. Let's stick with the custom CSS then.

packages/js/email-editor/src/components/header/back-button-content.tsx
```diff
@@ -0,0 +1,106 @@
...
+			urls: select( storeName ).getUrls(),
+		} ),
+		[]
+	);
+
+	function sendAction() {
+		if ( urls.listings ) {
+			window.location.href = urls.listings;
+		}
+	}
```
costasovo: I would suggest slightly improving the naming here. Sorry for nitpicking, but this is part of the API for editor configuration, so I think it is worth the effort.

I would suggest:
- `urls.listings` => `urls.backLink` or something similar. It might be some kind of listing in many cases, but for example, in MailPoet, when you open the editor from an automation, you go back to the automation editor. So I find the backUrl or backLink better.
- `sendAction` => `backAction`

costasovo (commented):
@pavel-mailpoet Thanks for the update!

I tested the functionality and it works as expected when the `isFullScreenForced` is enabled. 

Here is video from testing: 

https://github.com/user-attachments/assets/777c1e2a-cb53-4706-9fcb-c3e525bd93fd

The button also works in the mobile version:

Screenshot 2025-05-23 at 10 10 19

But when I disable `isFullScreenForced` and switch to the Fullscreen the button is not present. 

In the ticket we say: "By default, the icon is WordPress and it closes Fullscreen mode. It is displayed only in full-screen mode." 

So I was expecting that when fullscreen mode is active the button is there and it just closes the fullscreen. You can observe this functionality in the post editor.

Screenshot where the button is missing:
Screenshot 2025-05-23 at 10 13 55

costasovo (approved):
@pavel-mailpoet Thanks for the fix.

I can confirm that now the button is visible in fullscreen mode when the fullscreen mode is active.

PR #58162, opened 2025-05-21
Product Gallery: remove not working "crop to fit" feature and add "auto" aspect ratio to Thumbnails

kmanijak:

"Crop images to fit" option didn't work. It was available on the Product Gallery parent block. While the attribute wasn't used anywhere I decided to:
1. Remove the option from container
2. Do not add the option in Large Image - the reason is we're aiming to use Product Image inside to handle image display. 
Given this, I didn't want to introduce an attribute that would get removed soon.
3. Do not add the option in Thumbnails BUT add "Auto" aspect ratio option instead, which shows uncropped images within available space

**Before:** Option had no effect
**After:**

aspectratio

#### Crop Images to fit in Thumbnails

1. Create product with multiple images, make sure some of them are portrait and landscape
3. Go to Editor > Single Product template
5. Replace the Product Image Gallery block with the Product Gallery (beta) block.
6. Focus on Product Gallery (Beta) block
7. **Expected:** There's no "Crop images to fit" option in inspector controls
8. Focus on Thumbnails
9. Change the aspect ratio to "Auto"
10. Save and go to frontend
11. **Expected:** Images are displayed with original aspect ratio within available space
12. Go back to editor
13. Focus on Product Gallery (Beta) and change the orientation from ➡️ to ⬇️ 
14. Save and go to frontend
15. **Expected:** Images are displayed with original aspect ratio within available space
16. Repeat steps 9 to 15 for other aspect ratios (see video above for guidance)

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryThumbnails.php
```diff
@@ -68,6 +68,14 @@ protected function render( $attributes, $content, $block ) {
 			return '';
 		}
 
+		// We crop the images to square only if the aspect ratio is 1:1.
+		// Otherwise, we show the uncropped and use object-fit to crop them.
+		if ( '1' === $attributes['aspectRatio'] ) {
+			update_option( 'woocommerce_thumbnail_cropping', '1:1' );
+		} else {
+			update_option( 'woocommerce_thumbnail_cropping', 'uncropped' );
+		}
```
Aljullu: Is it intentional that we update this option on render? I feel it's a bit weird and might have performance issues. If two Product Gallery Thumbnails blocks on the same page have different `aspectRatio` attributes, that means we would be updating this option one way and the other on every render, no? :thinking: 

If I'm understanding this correctly, I wonder if using the `pre_option_woocommerce_thumbnail_cropping` filter to override the option only when rendering this block might be a good alternative. Or if we want this option to be synced across blocks and all WooCommerce, we could add a global attribute like in https://github.com/woocommerce/woocommerce/pull/54166?

Aljullu (commented):
Looking good, I left one question about the usage of `update_option()`.

I also noticed that when setting the aspect radio to `Wide - 16:9`, the images are much bigger when the thumbnails are on top than on the left:

Left thumbnails | Top thumbnails
--- | ---
imatge | imatge

Is that expected?

plugins/woocommerce/src/Blocks/BlockTypes/ProductGalleryThumbnails.php
kmanijak: Great comment @Aljullu and thank you for pointing it out! I see the performance implications now and also how it impacts other blocks (if `update_option( 'woocommerce_thumbnail_cropping', 'uncropped' );` was used last, it impacts other blocks relying on this size)! I had limited understanding of this option and now I see how I should not have used it in the first place.

I updated it to use `woocommerce_thumbnail` size for square aspect ratio and "woocommerce_single" for others as in different aspect ratios we should not start with already cropped image before cropping it to specific aspect ratio on a frontend (https://github.com/woocommerce/woocommerce/pull/58162/commits/d5d7ba2dbebf3ec0a895bb69d6c67c31c67ff622 and https://github.com/woocommerce/woocommerce/pull/58162/commits/0707e39a74d8dfb40b0b11b6c283382f2787c936)

kmanijak:
@Aljullu,

> I also noticed that when setting the aspect radio to Wide - 16:9, the images are much bigger when the thumbnails are on top than on the left:

That's something I didn't encounter. I'm wondering if that may be related to the options I used 🤔 Here's example of portrait and landscape images I use in both orientations:

Left | Top
-- | --
image | image

It's expected they're relatively bigger on top but there should not be a shrinking effect I see in here, is that what you meant? 🤔 
Screen Shot 2025-06-02 at 14 34 01 PM

Aljullu:
> It's expected they're relatively bigger on top but there should not be a shrinking effect I see in here, is that what you meant? 🤔 

Yes, apparently I can only reproduce in Firefox. It seems to be fixed by adding `width: fix-content;` to `.wc-block-product-gallery-thumbnails__thumbnail__image`. But I don't know if that might have some other consequences. :grimacing: 

Let me know if you still can't reproduce it or if you want me to create a PR.

kmanijak:
> Yes, apparently I can only reproduce in Firefox. It seems to be fixed by adding width: fix-content; to .wc-block-product-gallery-thumbnails__thumbnail__image. But I don't know if that might have some other consequences. 😬
> Let me know if you still can't reproduce it or if you want me to create a PR.

Ah, I can reproduce in Firefox, I'll take care of it, thank you!

kmanijak:
> Yes, apparently I can only reproduce in Firefox. It seems to be fixed by adding width: fix-content; to .wc-block-product-gallery-thumbnails__thumbnail__image. But I don't know if that might have some other consequences. 😬
> Let me know if you still can't reproduce it or if you want me to create a PR.

@Aljullu , I double-checked and I could reproduce the issue on 4de1c4dfcc602d386d0a27e0cecf19c7ff45da4a but couldn't on a freshest commit on this branch. I didn't apply the fix you've mentioned, removing the option update was sufficient.

Also, I made additional testing on:
- Firefox
- Safari
- Edge
- Chrome
And couldn't find any discrepancies, I hope it's now ready for final review  🙌

Aljullu (approved):
Thanks for the updates, @kmanijak! Everything seems to be looking fine on my end, now:

image

<details>

* Theme: Twenty Twenty-Five
* Store size: small
</details>

Issue #58125, opened 2025-05-20
ProductAvailabilityUtils::get_product_availability gets called multiple times in block themes


jimjasson:
### Prerequisites

- [x] I have carried out troubleshooting steps and I believe I have found a bug.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

In this PR ⬇ 
- https://github.com/woocommerce/woocommerce/pull/54156

A new function was introduced, `get_product_availability`, which performs pretty heavy operations, and it's called multiple times when a block theme is used.

The heavy operation is that for each variable product, it calls [get_available_variations](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/class-wc-product-variable.php#L308) , which loops all variations and calls `wc_get_product` , to determine if the variable product is out of stock. 

`ProductAvailabilityUtils::get_product_availability` is also called from the StoreAPI's Product Schema, in the `get_item_response` function. 

https://github.com/woocommerce/woocommerce/blob/8eeadc72263e5994b8f0bb1a3b5b0b69941dde83/plugins/woocommerce/src/Blocks/Utils/ProductAvailabilityUtils.php#L28-L38

The slowdown is noticed when using a block theme, cart and checkout blocks **and at least one product is in the cart.**

We've tested this on wordpressdotcom sites, and for a store with 25 products and 786 variations
- Product Category takes 12 seconds
- Single product takes 8 seconds
- Adding to cart takes 8 seconds
- Cart takes 8 seconds

This started happening in WooCommerce 9.6, but in 9.7 this hook ran even more times in the Cart block. 

-----------------------

On a similar note, the filter `woocommerce_product_availability` doesn't provide the product as an argument, which renders the filter pointless, as the text cannot be customized per product.

https://github.com/woocommerce/woocommerce/blob/8eeadc72263e5994b8f0bb1a3b5b0b69941dde83/plugins/woocommerce/src/Blocks/Utils/ProductAvailabilityUtils.php#L40-L46

It should have been similar to the `woocommerce_get_availability`

https://github.com/woocommerce/woocommerce/blob/8eeadc72263e5994b8f0bb1a3b5b0b69941dde83/plugins/woocommerce/includes/abstracts/abstract-wc-product.php#L2229-L2238

I can report the filter in a separate issue, but I thought it made sense to mention given they are closely coupled and introduced in the same PR.

### Expected behavior

It shouldn't take that long to load these pages. 

[...]

Issue #58026, opened 2025-05-15
Product Collection > Stock Indicator is empty


Marc-pi:
### Prerequisites

- [x] I have carried out troubleshooting steps and I believe I have found a bug.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

follow up of https://github.com/woocommerce/woocommerce/issues/42620
and https://github.com/woocommerce/woocommerce/issues/57066

With a block theme, adding the Stock Indicator Block generates an empty data
My Product Collection block is set to Featured Images
Both front & editor side are empty, despite the product is Out of Stock
I did not test with other Product Collection "Templates"

![Image](https://github.com/user-attachments/assets/d2d9bb3c-51ef-48a4-b96c-be400709296f)

![Image](https://github.com/user-attachments/assets/96e39615-f223-49ca-9a05-cb5dc450dd62)

### Expected behavior

The data should not be empty 

### Actual behavior

No data retrieved
No Dev Tools Console errors, nor on the network tab (Firefox)

### Steps to reproduce

- On home page (set with the Home Page FSE Template)
- Add the Stock Indicator within a Product Collection Featured Products
- Observe that is does not retrieve any data even on Out of Stock Products, despite the WP Query does take into consideration the due status

### WordPress Environment

- WP 6.8.1
- Woo 9.8x
- TT5 theme
- GB 20.8 but if inactivated we have the same behavior

### Isolating the problem

- [ ] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [x] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/products/storefront/).
- [ ] I can reproduce this bug consistently using the steps above.

Marc-pi:
ping @gigitux 
i'll test further on with other configurations, and try to replicate on playground later on

gigitux:
Thanks for the ping! I'm unable to replicate the issue. As mentioned in the block description, the Product Stock Indicator is visible only when the block is not in stock.

https://github.com/woocommerce/woocommerce/blob/0a9e1585635c28c3d4e0e6b380bfced88da15611/plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/stock-indicator/block.json#L5-L6


In this video, I tested the Product Stock Indicator to see if it reacts well to the Product stock changes.


https://github.com/user-attachments/assets/0957b065-661e-4af5-919e-b239c9fdf2fe.


[...]

PR #58185, opened 2025-05-21
Blocks - Improve product availability checks for variations

PanosSynetos:

**Please see this issue before proceeding** 👇 
- https://github.com/woocommerce/woocommerce/issues/58125

### First attempt 2025-05-21 

The first change I made in this PR was to change how we call `$product->get_available_variations`. Instead of calling it without parameters, I passed `objects`.

This prevents calling `get_available_variation` and saves some seconds.

https://github.com/woocommerce/woocommerce/blob/d596e0344d6eb98c860daa36909680d5a75eff35/plugins/woocommerce/includes/class-wc-product-variable.php#L330-L334

**I tried the change on a customer's staging site, where they complained about slow loading times, and pages are 40-45% faster; better but still not acceptable.**

| Page | Before | After |
| ------ | ------ | ----- |
| add to cart          |  9.2      | 4.8      |
| single product          |  8.2      | 4.5      |
| category archive          |  14      | 6.7      |
| cart          |  2.7      | 1.8      |

### Second attempt 2025-05-22 (current version of the PR)

Instead of calling `$product->get_available_variations` and _then_ checking if the variations array is empty (to determine if there is at least 1 available variation), I introduced a new function `$product->has_available_variations` (thanks @mreishus ).

This function bails out early when the first available variation is found and doesn't loop through all of them. That's a significant boost, especially for sites with products with many variations.

_Note that if all variations are not available, we would end up with the exact number of loops._

**I tried the change on a customer's staging site, where they complained about slow loading times, and pages are 80% faster.**

| Page | Before | 2025-05-21 |  2025-05-22 |
| ------ | ------ | ----- | ----- |
| add to cart          |  9.2      | 4.8      | 1.4 |
| single product          |  8.2      | 4.5      | 1.3 |
| category archive          |  14      | 6.7      | 1.6 |
| cart          |  2.7      | 1.8      | 0.9 |

This is a quick improvement, but if I were to recommend anything, it would be to revert 
- https://github.com/woocommerce/woocommerce/pull/54156
and rethink how to solve the initial issue
- https://github.com/woocommerce/woocommerce/issues/53720

Tagging @sunyatasattva @nerrad @Aljullu and @tjcafferkey  (as this function is called in the StoreAPI) to evaluate how to solve the initial issue and investigate why the function is called multiple times
- https://github.com/woocommerce/woocommerce/issues/58125

| Before | After |
| ------ | ----- |
|        |       |

#### trunk
1. Use a block theme
2. Create a couple of variable products in your store and several variations. (I can share a CSV)
3. Add a product to the cart
4. Visit a category page, single product page and cart and note down the loading times (I used Query Monitor)

#### this branch

1. Use a block theme
2. Create a couple of variable products in your store and several variations. (I can share a CSV)
3. Add a product to the cart
4. Visit a category page, single product page and cart and note down the loading times (I used Query Monitor)
5. See the difference in loading times (40-45% faster)

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/includes/class-wc-product-variable.php
+++ b/plugins/woocommerce/includes/class-wc-product-variable.php
@@ -301,6 +301,8 @@ class WC_Product_Variable extends WC_Product {
 	/**
 	 * Get an array of available variations for the current product.
 	 *
+	 * @hint If the function gets updated, make sure to update has_available_variations too, as they share similar logic.
+	 *
 	 * @param string $return Optional. The format to return the results in. Can be 'array' to return an array of variation data or 'objects' for the product objects. Default 'array'.
 	 *
 	 * @return array[]|WC_Product_Variation[]
@@ -339,6 +341,43 @@ class WC_Product_Variable extends WC_Product {
 		}
 
 		return $available_variations;
+	}
+
+	/**
+	 * Check if there are available variations for the current product.
+	 *
+	 * @hint If the function gets updated, make sure to update get_available_variations too, as they share similar logic.
+	 *
+	 * @since  9.9.0
+	 * @return boolean
+	 */
+	public function has_available_variations() {
+		$variation_ids        = $this->get_children();
+
+		if ( is_callable( '_prime_post_caches' ) ) {
+			_prime_post_caches( $variation_ids );
+		}
+
+		foreach ( $variation_ids as $variation_id ) {
+
+			$variation = wc_get_product( $variation_id );
+
+			// Hide out of stock variations if 'Hide out of stock items from the catalog' is checked.
+			if ( ! $variation || ! $variation->exists() || ( 'yes' === get_option( 'woocommerce_hide_out_of_stock_items' ) && ! $variation->is_in_stock() ) ) {
+				continue;
+			}
+
+			// Filter 'woocommerce_hide_invisible_variations' to optionally hide invisible variations (disabled variations and variations with empty price).
+			if ( apply_filters( 'woocommerce_hide_invisible_variations', true, $this->get_id(), $variation ) && ! $variation->variation_is_visible() ) {
+				continue;
+			}
+
+			// We found at least one available variation, so return true.
+			return true;
+		}
+
+		// There were either no variations, or they were hidden because of the "continues" above.
+		return false;
 	}
 
 	/**
--- a/plugins/woocommerce/includes/class-wc-product-variable.php
+++ b/plugins/woocommerce/includes/class-wc-product-variable.php
@@ -324,7 +324,13 @@ class WC_Product_Variable extends WC_Product {
 				continue;
 			}
 
-			// Filter 'woocommerce_hide_invisible_variations' to optionally hide invisible variations (disabled variations and variations with empty price).
+			/**
+			 * Filter 'woocommerce_hide_invisible_variations' to optionally hide invisible variations (disabled variations and variations with empty price).
+			 *
+			 * @param  bool                  $hide        Whether to hide invisible variations. Default true.
+			 * @param  int                   $product_id  The ID of the variation.
+			 * @param  WC_Product_Variation  $variation   The variation object.
+			 */
 			if ( apply_filters( 'woocommerce_hide_invisible_variations', true, $this->get_id(), $variation ) && ! $variation->variation_is_visible() ) {
 				continue;
 			}
@@ -352,7 +358,7 @@ class WC_Product_Variable extends WC_Product {
 	 * @return boolean
 	 */
 	public function has_available_variations() {
-		$variation_ids        = $this->get_children();
+		$variation_ids = $this->get_children();
 
 		if ( is_callable( '_prime_post_caches' ) ) {
 			_prime_post_caches( $variation_ids );
@@ -367,7 +373,13 @@ class WC_Product_Variable extends WC_Product {
 				continue;
 			}
 
-			// Filter 'woocommerce_hide_invisible_variations' to optionally hide invisible variations (disabled variations and variations with empty price).
+			/**
+			 * Filter 'woocommerce_hide_invisible_variations' to optionally hide invisible variations (disabled variations and variations with empty price).
+			 *
+			 * @param  bool                  $hide        Whether to hide invisible variations. Default true.
+			 * @param  int                   $product_id  The ID of the variation.
+			 * @param  WC_Product_Variation  $variation   The variation object.
+			 */
 			if ( apply_filters( 'woocommerce_hide_invisible_variations', true, $this->get_id(), $variation ) && ! $variation->variation_is_visible() ) {
 				continue;
 			}
--- a/plugins/woocommerce/includes/class-wc-product-variable.php
+++ b/plugins/woocommerce/includes/class-wc-product-variable.php
@@ -308,8 +308,9 @@ class WC_Product_Variable extends WC_Product {
 	 * @return array[]|WC_Product_Variation[]
 	 */
 	public function get_available_variations( $return = 'array' ) {
-		$variation_ids        = $this->get_children();
-		$available_variations = array();
+		$variation_ids           = $this->get_children();
+		$hide_out_of_stock_items = ( 'yes' === get_option( 'woocommerce_hide_out_of_stock_items' ) );
+		$available_variations    = array();
 
 		if ( is_callable( '_prime_post_caches' ) ) {
 			_prime_post_caches( $variation_ids );
@@ -320,7 +321,7 @@ class WC_Product_Variable extends WC_Product {
 			$variation = wc_get_product( $variation_id );
 
 			// Hide out of stock variations if 'Hide out of stock items from the catalog' is checked.
-			if ( ! $variation || ! $variation->exists() || ( 'yes' === get_option( 'woocommerce_hide_out_of_stock_items' ) && ! $variation->is_in_stock() ) ) {
+			if ( ! $variation || ! $variation->exists() || ( $hide_out_of_stock_items && ! $variation->is_in_stock() ) ) {
 				continue;
 			}
 
@@ -357,10 +358,11 @@ class WC_Product_Variable extends WC_Product {
 	 * @hint If the function gets updated, make sure to update get_available_variations too, as they share similar logic.
 	 *
 	 * @since  9.9.0
-	 * @return boolean
+	 * @return bool
 	 */
 	public function has_available_variations() {
-		$variation_ids = $this->get_children();
+		$variation_ids           = $this->get_children();
+		$hide_out_of_stock_items = ( 'yes' === get_option( 'woocommerce_hide_out_of_stock_items' ) );
 
 		if ( is_callable( '_prime_post_caches' ) ) {
 			_prime_post_caches( $variation_ids );
@@ -371,7 +373,7 @@ class WC_Product_Variable extends WC_Product {
 			$variation = wc_get_product( $variation_id );
 
 			// Hide out of stock variations if 'Hide out of stock items from the catalog' is checked.
-			if ( ! $variation || ! $variation->exists() || ( 'yes' === get_option( 'woocommerce_hide_out_of_stock_items' ) && ! $variation->is_in_stock() ) ) {
+			if ( ! $variation || ! $variation->exists() || ( $hide_out_of_stock_items && ! $variation->is_in_stock() ) ) {
 				continue;
 			}
 
--- a/plugins/woocommerce/includes/class-wc-product-variable.php
+++ b/plugins/woocommerce/includes/class-wc-product-variable.php
@@ -327,6 +327,8 @@ class WC_Product_Variable extends WC_Product {
 			/**
 			 * Filter 'woocommerce_hide_invisible_variations' to optionally hide invisible variations (disabled variations and variations with empty price).
 			 *
+			 * @since 2.6.8
+			 *
 			 * @param  bool                  $hide        Whether to hide invisible variations. Default true.
 			 * @param  int                   $product_id  The ID of the variation.
 			 * @param  WC_Product_Variation  $variation   The variation object.
@@ -376,6 +378,8 @@ class WC_Product_Variable extends WC_Product {
 			/**
 			 * Filter 'woocommerce_hide_invisible_variations' to optionally hide invisible variations (disabled variations and variations with empty price).
 			 *
+			 * @since 2.6.8
+			 *
 			 * @param  bool                  $hide        Whether to hide invisible variations. Default true.
 			 * @param  int                   $product_id  The ID of the variation.
 			 * @param  WC_Product_Variation  $variation   The variation object.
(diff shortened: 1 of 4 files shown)


plugins/woocommerce/includes/class-wc-product-variable.php
```diff
@@ -341,6 +351,51 @@ public function get_available_variations( $return = 'array' ) {
 		return $available_variations;
 	}
 
+	/**
+	 * Check if there are available variations for the current product.
+	 *
+	 * @hint If the function gets updated, make sure to update get_available_variations too, as they share similar logic.
+	 *
+	 * @since  9.9.0
+	 * @return boolean
```
xristos3490: NIt. Maybe use `bool` for consistency.

plugins/woocommerce/includes/class-wc-product-variable.php
```diff
@@ -341,6 +351,51 @@ public function get_available_variations( $return = 'array' ) {
...
+		if ( is_callable( '_prime_post_caches' ) ) {
+			_prime_post_caches( $variation_ids );
+		}
+
+		foreach ( $variation_ids as $variation_id ) {
+
+			$variation = wc_get_product( $variation_id );
+
+			// Hide out of stock variations if 'Hide out of stock items from the catalog' is checked.
+			if ( ! $variation || ! $variation->exists() || ( 'yes' === get_option( 'woocommerce_hide_out_of_stock_items' ) && ! $variation->is_in_stock() ) ) {
```
xristos3490: What about moving the "get_option" call outside the loop? Considering caching, this should have a minor impact, but it would still be a nice improvement.

Plus, it can be applied in the "get_available_variations()".

xristos3490 (commented):
Great work, Pano! Left a couple of comments!

plugins/woocommerce/includes/class-wc-product-variable.php
PanosSynetos: Thank you, addressed in https://github.com/woocommerce/woocommerce/pull/58185/commits/f0597dbe3fc3cba4c9ed4c3848d377685d0068fd

plugins/woocommerce/includes/class-wc-product-variable.php
PanosSynetos: Thank you, addressed in https://github.com/woocommerce/woocommerce/pull/58185/commits/f0597dbe3fc3cba4c9ed4c3848d377685d0068fd

dinhtungdu (commented):
Thanks for the PR! Looking at the code, I can see this will greatly improve the performance.

I can't reproduce a similar stat on my localhost. Even with 132 variations, the product page still loads in under half of second. So I have to use this snippet to time the execution time of the slow utility only:
```php
		$start = microtime(true);
		$availability = ProductAvailabilityUtils::get_product_availability( $product_to_render );
		$end = microtime(true);
		$time = $end - $start;
		var_dump( 'Time taken to get product availability: ' . $time );
```

On a `wp-env` instance with 2 core CPU and 2 GB Ram on my M4 Max MacBook Pr, I got this visiting the same single product page of a variable products with 132 variations:
- `trunk`: `Time taken to get product availability: 0.080392837524414`. The total load time shown in Query Monitor is 0.29s.
- This PR:  `Time taken to get product availability: 0.00078797340393066`. The total load time shown in Query Monitor is 0.21s.

This is a 10 times difference and can show the performance gain we get. 

In your test, single product page load time goes from  8.2 to 1.3s seconds, which makes me wonder if other extensions might play a role here, as the relative performance gain is much bigger compared to my localsite, which only has WooCommerce activated.

About the new method added here, I really think it’s a missing method of `WC_Product_Variable`. But I'm wary about adding a public method in such a short timeframe, under pressure of meeting PRR. We will have to maintain the public method we're adding here because it's public. Besides, we only used it inside the `ProductAvailabilityUtils::get_product_availability`, for now. I wonder if there is a way to give us more time to think about the demand and implementation of the public method, such as using the lookup table to aggregate and calculate the availability, which should be much faster than using the data store. Another thing is caching, we may want to use object cache to avoid duplicated calculations in a same request.

By the way, looking at the usage of `ProductAvailabilityUtils::get_product_availability`, I think `WC_Product_Variable` should override the `get_availability` to provide the accurate availability information, instead of having a utility to do so.

My suggestion for our safety is to put the logic inside ￼`ProductAvailabilityUtils::get_product_availability`￼ first, then move it to ￼`WC_Product_Variable`￼ later. We can solve the issue for the client, but still have more time to carefully craft the public method.

dinhtungdu:
> such as using the lookup table to aggregate and calculate the availability, which should be much faster than using the data store. Another thing is caching, we may want to use object cache to avoid duplicated calculations in a same request.

FWIW, both can be added to the public method without breaking the compatibility. So that's not a point against adding new public method.

jimjasson (approved):
Hey @PanosSynetos, 

Really good idea to bring down the time it takes to search if there are any available variations in a product by stopping the search when the first variation is found, rather that finding all the available variations and then counting them. 

I tested this on a site with a couple of products with a lot of attributes and variations and here are the results: 

| Trunk | Current Branch | Result|
|--------|--------|--------|
| Shop Page - trunk | Shop Page - branch | Shop page: 1.02s -> 0.90 (11.7% faster)
| Product Page - trunk | Product Page - branch | Product page: 1.08s -> 0.99 (8.33% faster)
| Add to Cart - trunk | Add to Cart - branch | Add to Cart: 0.86s -> 0.86s (no change) 
| Cart - trunk | Cart - branch | Cart: 1.22s-> 0.74 (39% faster).

The biggest improvement seems to be in the cart page, with significant improvements in the Catalog and single product page. 

As for the technical implementation, I think that adding a public method called `has_available_variations` inside the `WC_Product_Variable` is a useful method to have and also aligns with other `has_` methods that exist in product classes. If we decide to improve this method (e.g. via introducing caching) we could do so later on. 

I also noticed that you added an extra parameter to the `woocommerce_product_availability` hook. I agree that having access to the product object there improves the usability of this hook and adding an extra argument won't affect developers who are already consuming this filter.

PanosSynetos:
Hey @dinhtungdu  , thanks for the prompt review.

> I can't reproduce a similar stat on my localhost.

Locally, the changes are barely noticeable. Our machines are pretty powerful.

> which makes me wonder if other extensions might play a role here,

Here’s a JN site with just Woo 9.8.5 and Query Monitor enabled to see the slow loading times https://pleasant-salmon-turkey.jurassic.ninja/

- Add https://pleasant-salmon-turkey.jurassic.ninja/product/press-off-serum-blood-moon-exclusive/ to the cart
    - It takes about 6 seconds
- Go to https://pleasant-salmon-turkey.jurassic.ninja/shop/
    - Takes 14 seconds
- Go to https://pleasant-salmon-turkey.jurassic.ninja/product-category/other/
    - Takes 10 seconds

Here’s a JN site, with this branch and Query Monitor enabled

- Add https://wild-dromedary-sally.jurassic.ninja/product/press-off-serum-blood-moon-exclusive/ to the cart
    - Takes 1 second
- Go to https://wild-dromedary-sally.jurassic.ninja/shop/
    - Takes 1 second
- Go to https://wild-dromedary-sally.jurassic.ninja/product-category/other/
    - Takes 1 second

*I have wp-admin credentials if needed*

> I think `WC_Product_Variable` should override the `get_availability` to provide the accurate availability information, instead of having a utility to do so.

I agree, but this is problem to be solved on a separate PR. As I mentioned the initial PR should be carefully re-assessed on how it solves the issue

- https://github.com/woocommerce/woocommerce/pull/54156

> My suggestion for our safety is to put the logic inside ￼`ProductAvailabilityUtils::get_product_availability`￼ first, then move it to ￼`WC_Product_Variable`￼ later. We can still solve the issue for the client, but we still have more time to carefully craft the public method.

Thanks for raising your concerns and suggesting to play defensively.

The newly introduced function is a “copy” of a function that has existed since WC 2.6.x , so it’s battle-tested. As you also mentioned, the `WC_Product_Variable` class is missing a `has_available_variations` and it’s a great addition. I’m sure it can be improved in the future, and it being public will push us to improve it (caching, lookup tables, or otherwise).

I fear that if we move it to `ProductAvailabilityUtils` it will stay there forever, lost in time 🙂

**Having said that, I’d prefer to keep it public.**

xristos3490 (approved):
Looks good!  

While I agree that this may not be the ideal approach, I believe that using the lookup table or caching, as @dinhtungdu suggested, could be a better option, although not a "quick fix." There’s certainly room for improvement! Nevertheless, I like this new method as it seems to fit well in the model, so I’m happy to approve it on my end!

PanosSynetos:
Nobody mentioned about the unit tests - Boohoo :D 
I don't want to take the credit here, it was @xristos3490 suggestion to unit test the new function

dinhtungdu (approved):
Thanks for the further explanation! Great work on the PR! Please create follow up issue to improve the method with caching and/or lookup table.  🚀

plugins/woocommerce/includes/class-wc-product-variable.php
```diff
@@ -341,6 +352,52 @@ public function get_available_variations( $return = 'array' ) {
...
+		}
+
+		foreach ( $variation_ids as $variation_id ) {
+
+			$variation = wc_get_product( $variation_id );
+
+			// Hide out of stock variations if 'Hide out of stock items from the catalog' is checked.
+			if ( ! $variation || ! $variation->exists() || ( $hide_out_of_stock_items && ! $variation->is_in_stock() ) ) {
+				continue;
+			}
```
gigitux, line 378: Do you think the check for **"Hide out of stock items from the catalog"** is necessary in this case?  
I'm working on this issue: https://github.com/woocommerce/woocommerce/issues/58026#issuecomment-2903825814.

If you follow [these testing instructions](https://github.com/woocommerce/woocommerce/issues/58026#issuecomment-2897939980), the stock indicator isn’t rendered when `woocommerce_hide_out_of_stock_items` is set to `true`.

Since we're only checking available variations (and not returning them), would it make sense to remove the check?

PanosSynetos: Hey @gigitux  - the newly introduced `has_available_variations` is a copy of `get_available_variations` and should both be aligned. 

Ideally, if `get_available_variations` was performant, then `has_available_variations` would have used the `get` function and wouldn't reinvent the wheel.

If we hide out-of-stock items, then the variation technically is not available, so it makes sense to keep the check in `has_available_variations`

Aljullu: I'm a bit confused by this as well. If I follow these steps:

1. Create a variable product.
2. Set all variations to out of stock.
3. Go to the _Single Product_ template and add the _Stock Indicator_ block, it doesn't display anything.

I would expect it to show 'Out of stock...', instead.

Am I missing something? :thinking:

Aljullu: By 'available variation' I understand purchasable variations, no? So I would expect out of stock variations not to be counted no matter what the `hide_out_of_stock_items` setting is.

xristos3490: > By 'available variation' I understand purchasable variations, no?

I think this ^ is an important question here.

There's a nuance with classic themes where the return value of "get_available_variations" is directly wired into the template for display on the variation form. For classic themes, "available" variations refer to those that can be shown on the template.

Check here:
https://github.com/woocommerce/woocommerce/blob/921b01cb35975c0bd9f0ae5ba7538ae2c12381ba/plugins/woocommerce/includes/wc-template-functions.php#L1963-L1970

And here:

https://github.com/woocommerce/woocommerce/blob/921b01cb35975c0bd9f0ae5ba7538ae2c12381ba/plugins/woocommerce/templates/single-product/add-to-cart/variable.php#L24

Given this, we need to verify the "hide_out_of_stock_items" setting. If it's OFF, the available variations will be shown as "Out of Stock" in the template, and therefore included in the return value.

From my perspective, we could explore:
1. Re-purpose the term "available variations" to mean "purchasable," just like @Aljullu refers to above. Then, refactor the templates-logic to reflect the changes.
2. Approach this in a fresh way, perhaps starting with a block-specific solution, while avoiding the term "available" altogether (for instance, we could introduce some helper functions like has_purchasable_variations).

What do you all think?
cc @PanosSynetos

PanosSynetos: Thanks Albert. How does it behave with 9.8.5 ? I'll have a look again tomorrow when I'm back at my desk.

Aljullu: Oh, thanks for the explanation, @xristos3490. I was definitely confused by 'available'! :sweat_smile: 

> How does it behave with 9.8.5 ?

Good question. I just tested and apparently it didn't work properly either?

So fixing this is probably not a blocker for WC 9.9, but I'm a bit wary of introducing the `has_available_variations()` method in `WC_Product_Variable` if we know it doesn't serve our needs. To avoid introducing more last-minute changes, I think we could simply mark it as `@internal` and in WC 10.0 rename it to `has_purchasable_variations()` or something along these lines? (let me know if you would like me to handle it :slightly_smiling_face: )

gigitux: > Good question. I just tested and apparently it didn't work properly either?

#58026 is not a regression introduced by this PR. I referenced it only to confirm whether the check in the function was added unintentionally, given the function's name.

As @Aljullu mentioned, there's no need to rush. Let's make the function private for now and continue the discussion. Thanks! 🙇

PanosSynetos: Cool, I'll fire up a quick PR to mark the `has_available_variations` as `@ internal`.

Should we move parts of the discussion in this new issue I've opened
- https://github.com/woocommerce/woocommerce/issues/58284 ?

Aljullu: > https://github.com/woocommerce/woocommerce/issues/58026 is not a regression introduced by this PR. I referenced it only to confirm whether the check in the function was added unintentionally, given the function's name.

I think they are two different issues, though:

1. #58026 happens when the parent variable product is out of stock.
2. The issue [I mentioned](https://github.com/woocommerce/woocommerce/pull/58185#discussion_r2107572745) happens when the parent variable product stock isn't tracked, and instead each individual variation is out of stock.

I can open a separate issue for 2, or we can use #58284 to fix it.

gigitux: In the end, isn't it the same issue? Both issues happen because this function is strictly coupled with the `hide_out_of_stock_items` setting?

Aljullu: They way I understand it, they are different. There are currently two ways a variable product might be out of stock:

1. The variable product is set as out of stock (at the product level):

imatge

2. The variable product stock is not defined, but each variation is out of stock:

imatge

imatge

Currently, both of them appear to be broken. I think 1 is what's described in https://github.com/woocommerce/woocommerce/issues/58026 (I created a PR that would fix it: https://github.com/woocommerce/woocommerce/pull/58354), while 2 would require us removing the `$hide_out_of_stock_items` check.

PanosSynetos: This issue seems related, so sharing it just in case
- https://github.com/woocommerce/woocommerce/issues/58441

PR #58193, opened 2025-05-21
Fix: Ensure product image respects column boundaries in editor and frontend

### LINKED ISSUE #42285: `Product Image` inconsistent style between Editor and front end
### Description

The `Product Image` has different behavior in the editor and in the front end.
When inserted in a column, if the image is wider than the column, it shows under the second column content in the editor, but over it in the front-end.
It should have the same behavior in both cases.

### Steps to reproduce

1. Create a new page or post.
2. Insert a `Single Product` block.
3. Click on the `Product Image` in the left column and give it a width larger than the column.
4. Notice the image is shown **under** the right column blocks in the Editor.
5. Save and go to the front end and notice the image is shown **over** the right column blocks.

### Screenshots, screen recording, code snippet

| Editor | Front-end |
|--------|--------|
| Screenshot 2023-07-20 at 15 39 36 | Screenshot 2023-07-20 at 15 39 41|

amitraj2203:

This PR resolves an inconsistency in the layout behavior of the product image block inside a column. Specifically, it ensures that the product image does not overflow its column container in either the editor or the frontend.

The following changes have been made:

1. **Frontend Fix:**
   - Removed the `max-width: none;` style from the image rendering logic in `ProductImage.php`, allowing the image to respect column constraints.

2. **Editor Fix:**
   - Added CSS to constrain the product image inside columns using `max-width: 100%`.
   - Removed the explicit `width` and `height` styles from the editor component (`edit.tsx`) to allow natural image sizing within the layout.

Closes #42285 

#### Before:
https://github.com/user-attachments/assets/46413bad-655a-4f5f-ae34-3d264e87c7f0

#### After: 
https://github.com/user-attachments/assets/8b0cab3c-9767-4484-b178-9d37fb767a96

1. Create or edit a page/post using the block editor.
2. Insert a **Single Product** block into a **Columns** layout (e.g., two columns).
3. Set the product image width to be larger than the column width.
4. Observe that:
   - In the **editor**, the image remains within the left column boundaries.
   - In the **frontend**, the image also remains constrained and no longer overlaps the second column.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/style.scss
```diff
@@ -68,3 +68,7 @@
 .wc-block-product-image__tools-panel .components-input-control {
 	margin-bottom: 8px;
 }
+
+.wp-block-column .wc-block-components-product-image img {
+	max-width: 100%;
+}
```
kmanijak: While it's testing well in Single Product block that happens to be using Column blocks inside, I also checked other cases, for example Product Collection block.

What about making it more generic:

```suggestion
.wc-block-components-product-image img {
	max-width: 100%;
}
```

and basically apply this change everywhere? 🤔

Before | After
-- | --
image | image

---

Although, I'm currently thinking what's more important:
- setting up `max-width: 100%` or
- should we respect specific `width/height` but apply overflow.

@shivanarrthine, sorry for ping out of nowhere but would be great if you share your opinion to the above question.

kmanijak (commented):
Hey @amitraj2203 , thanks again for another great contribution! 😄 While it works fine, I suggested more generic approach. But also, that made me think if that's the right approach or maybe we should actually respect the `width/height` but manage overflow better 🤔 Let's wait for our UX designer's input before proceeding 🙏

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/style.scss
amitraj2203: Thanks @kmanijak for the review. I have addressed the feedback to make the styles more generic.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/edit.tsx
```diff
@@ -61,7 +61,7 @@ const Edit = ( {
 
 	const ref = useRef< HTMLDivElement >( null );
 
-	const blockProps = useBlockProps( { style: { width, height } } );
+	const blockProps = useBlockProps();
```
kmanijak, line 64: I see this change was added later, after my initial review. What is it responsible for? Is it redundant? 🤔

kmanijak (commented):
I double checked the PR and it works fine! Image doesn't exceed available space in

Editor | Frontend
-- | --
image | image

I just left one question about change that was added later after my initial review. Once answered and there's no concerns we can merge 🙌

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/image/edit.tsx
amitraj2203: Thanks @kmanijak for the review.

> I see this change was added later, after my initial review.

I believe this change was actually added before that, here [71292d0](https://github.com/woocommerce/woocommerce/pull/58193/commits/71292d08d10f53e0af68330de22955ecafbbd8b9)

> What is it responsible for? Is it redundant?

It’s not redundant, this change addresses the issue in the editor where, without it, inline styles are applied to the image, causing it to overflow the column width.

---

Let me know if I'm missing anything.

kmanijak: Ah, I checked this commit https://github.com/woocommerce/woocommerce/pull/58193/commits/9236500cf7de4c3a20c93976d208cb6db553ce63 and the change appeared there because inlines styles were "re-added" after bringing `trunk` changes. Sorry, misread it! OK, let's ship this PR! 🚢

kmanijak (approved):
My [question and concern](https://github.com/woocommerce/woocommerce/pull/58193#discussion_r2200121587) was addressed, all good!

PR #58194, opened 2025-05-21
Enable support for Social Link block in the Email Editor

### LINKED ISSUE #58022: Implement Core: Social Link Block
Background: [WOOPLUG-4057](https://linear.app/a8c/issue/WOOPLUG-4057/spike-on-creating-the-social-icon-block) ([GitHub](https://github.com/woocommerce/woocommerce/issues/57502))

Proposal: [WOOPLUG-4057](https://linear.app/a8c/issue/WOOPLUG-4057/spike-on-creating-the-social-icon-block#comment-4715c65f) ([GitHub](https://github.com/woocommerce/woocommerce/issues/57502#issuecomment-2873488917))

We need to implement and add support for the Social link block.

### Technical details

Social Block Proposal and Technical Implementation Details.

1. Replace SVG with PNG images.
2. Limit the number of accepted/supported social icons (top 20 for now). (Will add list later)
3. Create a PNG Image for the following (for each of the social media icons) (**Note**: The background for the PNG images will be transparent; the color here refers to the icon fill color)
   1. Black
   2. White
   3. The Default Social media color.
   4. Grey (To be done later)
4. Block Styles:
   1. Default and Pill Shape:
      1. Icon color: When either of the two block styles is selected in the wp-admin, remove or disable the Icon color block option. (Why: because we can't update the color of the PNG image, but we can set the background color)
      2. Icon background: when the background color selected is white-ish, use the Black PNG image; otherwise, set the White PNG image as the default
   2. Logos only:
      1. Icon color: When the 'Logo only' option is selected, remove the 'Icon color' block option. (Same reason as above). Use the Default Social media color.
      2. Icon background: Disable for Logos only.
   3. General colors: If possible, can we limit the Block Icon color to black and white? If not, should we register a new block option to set the image color, which will provide users with the ability to select and use either white, black, or the default social media color for the PNG image? (TODO)
5. Show Text option: The color for this will align with whatever is selected above.
6. Background: When selected, will set the TR- row background color
7. Size: We need to be able to set and update the icon sizes, but we can remove this option for this initial phase and add it later.
8. Implement a new renderer for this block

triple0t:

Closes #58022 WOOPLUG-4294

This PR implements support for the Social Links block in the WooCommerce Email Editor. The implementation includes:

- Support for rendering Social Link and Social Links blocks in the Email Editor
- Integration with core WordPress social link services
- Support for various social media platforms including Facebook, Twitter, Instagram, LinkedIn, and many more
- Customizable styling options including:
  - Icon sizes (small, normal, large, huge)
  - Brand colors for each social platform
  - Pill shape style support
  - Custom background colors
- Email client compatibility improvements
- Responsive design support

| Before | After |
| ------ | ----- |
|     New   |   Screenshot 2025-05-21 at 3 52 33 PM |
| New | Screenshot 2025-05-21 at 7 30 16 AM |
| New | Screenshot 2025-05-21 at 9 54 35 PM  |
| New | Screenshot 2025-05-21 at 11 42 59 PM |

1. Ensure the Block email editor is enabled. (`/wp-admin/admin.php?page=wc-settings&tab=advanced&section=features`)
2. Edit any email template from the email listing page. (`/wp-admin/admin.php?page=wc-settings&tab=email`)
3. Add the Social Links block
4. Add various social media links
5. Test different styling options:
   - Try different icon sizes
   - Test with and without labels
   - Test with pill shape style
   - Test with custom colors
6. Preview the email to verify rendering
7. Send a test email to verify compatibility with email clients

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


packages/php/email-editor/src/Integrations/Core/Renderer/Blocks/class-social-link.php
```diff
@@ -0,0 +1,30 @@
...
+	/**
+	 * Renders the block content.
+	 *
+	 * @param string              $block_content Block content.
+	 * @param array               $parsed_block Parsed block.
+	 * @param Settings_Controller $settings_controller Settings controller.
+	 * @return string
+	 */
+	protected function render_content( $block_content, array $parsed_block, Settings_Controller $settings_controller ): string {
+		// Not using this for now.
```
costasovo: It would be nice to add why we are not using this flow. I assume that's because the icons are rendered in the parent's block renderer, correct?

packages/js/email-editor/src/blocks/core/social-links.tsx
```diff
@@ -0,0 +1,153 @@
...
+				service: 'facebook',
+				url: 'https://www.facebook.com/WordPress/',
+			},
+		},
+		{
+			name: 'core/social-link',
+			attributes: {
+				service: 'twitter',
+				url: 'https://twitter.com/WordPress',
+			},
```
costasovo: I think we shouldn't add Twitter to the preset variation. Twitter was renamed to X so we could add X or just go with only two platforms.

packages/php/email-editor/src/Integrations/Utils/class-social-links-helper.php
```diff
@@ -104,4 +104,45 @@ public static function get_service_brand_color( $service_name ) {
...
+	 * Source: https://github.com/WordPress/gutenberg/blob/ca36a468e3ff2c564bdecbfbb5abd5ab6e603d68/packages/block-library/src/social-links/edit.js#L39-L44
+	 *
+	 * @return array The social link size options.
+	 */
+	public static function get_social_link_size_options() {
+		return array(
+			'has-small-icon-size'  => __( 'Small' ),
+			'has-normal-icon-size' => __( 'Normal' ),
+			'has-large-icon-size'  => __( 'Large' ),
+			'has-huge-icon-size'   => __( 'Huge' ),
```
costasovo: Can we use the translations without `woocommerce` as the second parameter here?

costasovo (commented):
@triple0t Great job! I think we are almost there 💪 

I left a few comments and found a couple of minor rendering issues when testing.

| Editor  | Rendered output |
| Screenshot 2025-05-22 at 15 42 41 | Screenshot 2025-05-22 at 15 42 13 |


### Issues:
1) In the editor the gaps between icons both vertical and horizontal respect Styles > Layout > Block spacing setting. VS in renderd output the horizontal gap is fixed and vertical is none. I think we should unify it and we can also decide whether we want to change the appearance in renderer or in the editor.

2) The purple icons with text on the right are Default style with text and the internal padding of the icons in the editor is different compared to the rendered. The editor version is IMO nicer.

Screenshot 2025-05-22 at 13 28 07

3) There is a weird artefact ⬆️  in the Colors section when Logos only is active. 

4) We can optimize the PNG images used for the icons. I tested a couple of random icons and I was able to reduce their size by 40% (using https://imageoptim.com/).

5) The telegram icon image is 10 times bigger than other icons. 

6) I would recommend adding an integration test for the renderer. It is packed with logic and handles many use cases. I can imagine we can easily break something with a little change.

### Testing in Email Clients
Here is the link to EmailOnAcid project I used https://app.emailonacid.com/app/editor/project/MTm4Y7h77SOwcA4QRlfWVAPY6duzMZxYc6KKvXC43wmOM

It looks good in the most email clients.  But Outlook :( 


| Google (correctly rendered)  | Outlook |
| Screenshot 2025-05-22 at 19 32 20 | Screenshot 2025-05-22 at 19 34 42 |

The rounded missing rounded corners are expected. I think we should address:
1) Icons in Outlook are not rendered on one line.
2) The text in labels should be bigger in the icons with texts.

packages/php/email-editor/src/Integrations/Core/Renderer/Blocks/class-social-link.php
triple0t: Yup. Updated

packages/js/email-editor/src/blocks/core/social-links.tsx
triple0t: Noted. Implemented in https://github.com/woocommerce/woocommerce/pull/58194/commits/2f9f53b7030f5b8934bea464c1457b2130aab2d2

packages/php/email-editor/src/Integrations/Utils/class-social-links-helper.php
triple0t: The method is unused and currently unnecessary. Removed in https://github.com/woocommerce/woocommerce/pull/58194/commits/2f9f53b7030f5b8934bea464c1457b2130aab2d2

triple0t:
Thank you for the in-depth review, @costasovo 

New commits start from https://github.com/woocommerce/woocommerce/pull/58194/commits/a1e7698484a651b75a7021f34694a4ef9b9d58dd

___

> In the editor the gaps between icons both vertical and horizontal respect Styles > Layout > Block spacing setting. VS in renderd output the horizontal gap is fixed and vertical is none. I think we should unify it and we can also decide whether we want to change the appearance in renderer or in the editor.

Screenshot 2025-05-27 at 5 28 41 PM


I'm not quite sure about this. Unlike in the [post editor,](https://github.com/user-attachments/assets/73093fe9-c4c9-4b7e-8a70-27a9ec8b1f93) we don't support justification in the email editor. The align option works well in the [rendered](https://github.com/user-attachments/assets/209fa106-d4b4-48e5-b1d6-512abf289f00) version.


> The purple icons with text on the right are Default style with text and the internal padding of the icons in the editor is different compared to the rendered. The editor version is IMO nicer.

Fixed in https://github.com/woocommerce/woocommerce/pull/58194/commits/a1e7698484a651b75a7021f34694a4ef9b9d58dd

> There is a weird artefact ⬆️ in the Colors section when Logos only is active.

Fixed in https://github.com/woocommerce/woocommerce/pull/58194/commits/a1e7698484a651b75a7021f34694a4ef9b9d58dd

> We can optimize the PNG images used for the icons. I tested a couple of random icons and I was able to reduce their size by 40% (using https://imageoptim.com/).

Implemented in https://github.com/woocommerce/woocommerce/pull/58194/commits/1d16ec4f720fc12c653425f7bf11d745aaeee24c

> The telegram icon image is 10 times bigger than other icons.

Fixed in https://github.com/woocommerce/woocommerce/pull/58194/commits/1d16ec4f720fc12c653425f7bf11d745aaeee24c

> I would recommend adding an integration test for the renderer. It is packed with logic and handles many use cases. I can imagine we can easily break something with a little change.

Implemented in https://github.com/woocommerce/woocommerce/pull/58194/commits/b1b8ad25cb6578a95aa24ae1d896d5936be791da

> Icons in Outlook are not rendered on one line.

Fixed in https://github.com/woocommerce/woocommerce/pull/58194/commits/d0ac4e528d4fb1c78bdfd6edf286ce139b373abc

Screenshot 2025-05-27 at 5 33 18 PM

Unfortunately, Outlook does not support the [border-radius](https://www.caniemail.com/features/css-border-radius/) property. The [alternative](https://learn.microsoft.com/en-us/windows/win32/vml/msdn-online-vml-roundrect-element) is deprecated. Investing time in developing a deprecated format doesn't make much sense. We can live with the box corner.

> The text in labels should be bigger in the icons with texts.

Implemented in https://github.com/woocommerce/woocommerce/pull/58194/commits/a1e7698484a651b75a7021f34694a4ef9b9d58dd.

packages/php/email-editor/src/Integrations/Utils/class-social-links-helper.php
```diff
@@ -130,4 +130,23 @@ public static function get_social_link_size_option_value( $size ) {
...
 	}
+
+	/**
+	 * Polyfill for str_starts_with() function added in PHP 8.0.
+	 *
+	 * @param string $haystack The string to search in.
+	 * @param string $needle The substring to search for.
+	 * @return bool True if $haystack starts with $needle, false otherwise.
+	 */
+	public static function str_starts_with( string $haystack, string $needle ): bool {
```
costasovo: I noticed that function in the previous review, but when checking it, I found that WordPress has a polyfill for that. See https://github.com/WordPress/WordPress/blob/5e1454dd91283a8f356c3258eba13ccc23753a85/wp-includes/compat.php#L394-L414
So this looks like a PHPStan config issue, and we shouldn't need to add the polyfill ourselves.
In case it is an issue in Unit tests, we may need to move it to integration tests, where WP is loaded.

costasovo (commented):
@triple0t Thank you for addressing the issues! Great job! Especially addressing the Outlook and adding the test coverage.

I added a comment about the polyfill that is not necessary, and I did another round of testing and found a few issues. Nothing major, though. Most of the things work as expected!

> Unfortunately, Outlook does not support the [border-radius](https://www.caniemail.com/features/css-border-radius/) property. The [alternative](https://learn.microsoft.com/en-us/windows/win32/vml/msdn-online-vml-roundrect-element) is deprecated. Investing time in developing a deprecated format doesn't make much sense. We can live with the box corner.

This is perfectly fine. We don't support border radius for Outlook.

> >In the editor the gaps between icons both vertical and horizontal respect Styles > Layout > Block spacing setting. VS in renderd output the horizontal gap is fixed and vertical is none. I think we should unify it and we can also decide whether we want to change the appearance in renderer or in the editor.

> I'm not quite sure about this. Unlike in the [post editor,](https://github.com/user-attachments/assets/73093fe9-c4c9-4b7e-8a70-27a9ec8b1f93) we don't support justification in the email editor. The align option works well in the [rendered](https://github.com/user-attachments/assets/209fa106-d4b4-48e5-b1d6-512abf289f00) version.

I think I was not clear in the description. The issue is still present. The issue is that the editor renders the gap between icons based on the styles settings. I recorded a video to illustrate the issue.


https://github.com/user-attachments/assets/59a69a24-5fd2-453b-a463-8a5b31936c50

I think we should address this. We could use the value from the style settings and apply half of it as a padding on the individual icon wrappers. Alternatively, we could alter the CSS for the editor to prevent the gap between icons, thereby achieving a WYSIWYG experience.

### Background-related issues
Screenshot 2025-05-28 at 10 48 31

In the image above, focus on the icons with the green background. On the left is the editor, and on the right is the rendered output. There are two issues:

1) The icons in the editor are not centered correctly. It happens when a background is applied to the social icons block. I found the CSS causing this is: `ul.has-background, ol.has-background { padding-left: 40px; }`

2) There is a slight gap/padding below icons in the rendered output. See the green line below the icons in the rendered output. It is not padding, but this issue is present in TD when some whitespaces are present. It can be eliminated by setting font-size: 0px; on the wrapper. I see that we set the font size for the label explicitly, so I think we can use the font-size: 0px trick as a fix here.

### Label alignment

Screenshot 2025-05-28 at 9 26 53

You can notice that the labels in the purple icons are slightly off vertically. The TD that wraps the icon has alignment set correctly to `middle`. The problem here is that font size is applied to the nested anchor element and in that case the vertical alignment doesn't work properly. It can be fixed by moving the font size from the anchor to the TD as illustrated by the arrow on the image.

packages/php/email-editor/src/Integrations/Utils/class-social-links-helper.php
triple0t: Yeah, this was due to PHPStan. Fixed in https://github.com/woocommerce/woocommerce/pull/58194/commits/7f401761ff5bc97fe7d881df12585380a4352ba8

triple0t:
Thanks for the review, @costasovo 


> I think I was not clear in the description. The issue is still present. The issue is that the editor renders the gap between icons based on the styles settings. I recorded a video to illustrate the issue.
> I think we should address this. We could use the value from the style settings and apply half of it as a padding on the individual icon wrappers. Alternatively, we could alter the CSS for the editor to prevent the gap between icons, thereby achieving a WYSIWYG experience.

Fixed in https://github.com/woocommerce/woocommerce/pull/58194/commits/21092303bbc1d9fa61a9bf75e9778c00f4cec846

> The icons in the editor are not centered correctly. It happens when a background is applied to the social icons block. I found the CSS causing this is: ul.has-background, ol.has-background { padding-left: 40px; }

Fixed in https://github.com/woocommerce/woocommerce/pull/58194/commits/21092303bbc1d9fa61a9bf75e9778c00f4cec846

> There is a slight gap/padding below icons in the rendered output. See the green line below the icons in the rendered output. It is not padding, but this issue is present in TD when some whitespaces are present. It can be eliminated by setting font-size: 0px; on the wrapper. I see that we set the font size for the label explicitly, so I think we can use the font-size: 0px trick as a fix here.

We can't really set font-size here as it would affect the display of the icons. Examples below

With Font size
Screenshot 2025-05-29 at 10 32 27 AM

Without Font size
Screenshot 2025-05-29 at 10 32 32 AM

I think we can live with this as the issue is not highly perceivable. Also, the background color is not really usable without some padding, so there's no need to change anything here.

> You can notice that the labels in the purple icons are slightly off vertically. The TD that wraps the icon has alignment set correctly to middle. The problem here is that font size is applied to the nested anchor element and in that case the vertical alignment doesn't work properly. It can be fixed by moving the font size from the anchor to the TD as illustrated by the arrow on the image.

Fixed in https://github.com/woocommerce/woocommerce/pull/58194/commits/21092303bbc1d9fa61a9bf75e9778c00f4cec846

costasovo (approved):
@triple0t Thank you for the fixes! Great job! I think it is ready for merge!

> I think we can live with this as the issue is not highly perceivable. Also, the background color is not really usable without some padding, so there's no need to change anything here.

Agree. I thought it might be an easy fix, but it is not worth complex changes.

### Testing performed

I downloaded the zip from the branch and tested on a jurrasic ninja site.

| Editor | Renderer |
| ------ | --------- |
| Screenshot 2025-05-29 at 15 13 31 |  Screenshot 2025-05-29 at 15 13 38 |

I also tested with [the Email On Acid service](https://app.emailonacid.com/app/editor/project/MTm4Y7h77SOwcA4QRlfWVAPY6duzMZxYc6KKvXC43wmOM):

Screenshot 2025-05-29 at 15 17 13
Screenshot 2025-05-29 at 15 17 33
Screenshot 2025-05-29 at 15 18 03
Screenshot 2025-05-29 at 15 18 53

PR #57994, opened 2025-05-14
Add proper error handling when WCCOM API is unavailable

bor0:
Don't crash the site.

ℹ️ Easier to review with [whitespace disabled](https://github.com/woocommerce/woocommerce/pull/57994/files?diff=unified&w=1).

### Changes proposed in this Pull Request:

Closes WCCOM-1363.

### Screenshots

| Before | After |
| ------ | ----- |
| Blank screen | Screenshot 2025-05-14 at 13 21 03 |

### How to test the changes in this Pull Request:

1. In WCCOM, disable rate limiter by adding a `return` at the top, right after the namespace declaration, in the files `plugins/wccom-plugins/api-rate-limiting/api-rate-limiting-check.php` and `plugins/wccom-plugins/api-rate-limiting/api-rate-limiting.php` - this will make your testing easier.
2. Make sure you have a connection to your local WCCOM site
    - ℹ️ Remember to add
```php
add_filter( 'woo_com_base_url', fn( $x ) => 'https://woocommerce.test' );
add_filter( 'woocommerce_helper_api_base', fn( $x ) => 'https://woocommerce.test/wp-json/helper/1.0' );
```
3. Simulate an error by adding `wp_send_json(1);` at the top of WCCOM's `WooCommerce_Helper_Endpoints::subscriptions`
4. Flush your store's transient `wp transient delete _woocommerce_helper_subscriptions`
5. Visit http://localhost:8080/wp-admin/admin.php?page=wc-admin. You should see an admin notice, not a blank screen. You should still be able to use WC core. Screenshot 2025-05-14 at 13 21 03
6. Now, assume the error has been fixed in WCCOM. Remove `wp_send_json(1);` from the top of WCCOM's `WooCommerce_Helper_Endpoints::subscriptions`.
7. Flush your store's transient `wp transient delete _woocommerce_helper_subscriptions`
8. Visit http://localhost:8080/wp-admin/admin.php?page=wc-admin. The admin notice should be gone, screen should render properly.
9. Simulate an error by adding `wp_send_json(1);` at the top of WCCOM's `WooCommerce_Helper_Endpoints::subscriptions`
10. Flush your store's transient `wp transient delete _woocommerce_helper_subscriptions`
11. Visit http://localhost:8080/wp-admin/admin.php?page=wc-admin. You should see an admin notice, not a blank screen.
10. Now, modify `WooCommerce_Helper_Endpoints::_check_permission` to `return false` (props @mcliwanow for the test case)
[...]

PR #58209, opened 2025-05-22
Perform WCCOM requests only when it's really necessary

bor0:

Refactor how and where `wccomHelper` is populated. Populate the settings that do API requests only in the necessary screens.

Fixes WCCOM-1365.

ℹ️ Makes it easier to review with [whitespace disabled](https://github.com/woocommerce/woocommerce/pull/58209/files?diff=unified&w=1).

1. First and foremost, ensure that nothing breaks! Smoke test in-app, connection/disconnection, etc.
2. One way to test this is to add `error_log( 'Requesting in ' . $_SERVER[ 'REQUEST_URI' ] );` right after [this line](https://github.com/woocommerce/woocommerce/blob/7d1e6e7feb3767aa05c5affaf8e266281f7b0bd3/plugins/woocommerce/includes/admin/helper/class-wc-helper-admin.php#L91). The log (requests) should only be triggered if you are on the Extensions screen.
3. Manually review the code (`grep`, etc) and double-check that each of these settings are only used under `Extensions`. I found out that most of them are happening in `my-subscriptions`, which shouldn't be used anywhere other than `Extensions`
    - `connected_notice`
    - `connection_url_notice`
    - `disconnected_notice`
    - `has_host_plan_orders`
    - `subscription_expired_notice`
    - `subscription_expiring_notice`
    - `subscription_missing_notice`
    - `wooUpdateCount`
    - `woocomConnectNoticeType`

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/includes/admin/helper/class-wc-helper-admin.php
```diff
@@ -83,20 +82,26 @@ function ( $product ) {
 			'wooUpdateManagerActive'     => WC_Woo_Update_Manager_Plugin::is_plugin_active(),
 			'wooUpdateManagerInstallUrl' => WC_Woo_Update_Manager_Plugin::generate_install_url(),
 			'wooUpdateManagerPluginSlug' => WC_Woo_Update_Manager_Plugin::WOO_UPDATE_MANAGER_SLUG,
-			'wooUpdateCount'             => WC_Helper_Updater::get_updates_count_based_on_site_status(),
-			'woocomConnectNoticeType'    => $woo_connect_notice_type,
 			'dismissNoticeNonce'         => wp_create_nonce( 'dismiss_notice' ),
-			'connected_notice'           => PluginsHelper::get_wccom_connected_notice( $auth_user_email ),
```
bor0, line 95: These three were eventually calling `get_subscriptions`, so moved them, too.

plugins/woocommerce/includes/admin/helper/class-wc-helper-admin.php
```diff
@@ -83,20 +82,26 @@ function ( $product ) {
...
 		);
 
-		if ( WC_Helper::is_site_connected() ) {
-			$settings['wccomHelper']['subscription_expired_notice']  = PluginsHelper::get_expired_subscription_notice( false );
-			$settings['wccomHelper']['subscription_expiring_notice'] = PluginsHelper::get_expiring_subscription_notice( false );
-			$settings['wccomHelper']['subscription_missing_notice']  = PluginsHelper::get_missing_subscription_notice();
-			$settings['wccomHelper']['connection_url_notice']        = WC_Woo_Helper_Connection::get_connection_url_notice();
-			$settings['wccomHelper']['has_host_plan_orders']         = WC_Woo_Helper_Connection::has_host_plan_orders();
```
bor0, line 103: These were doing the actual requests, so moved them all out.

plugins/woocommerce/includes/admin/helper/class-wc-helper-updater.php
```diff
@@ -409,11 +409,8 @@ public static function get_available_extensions_downloads_data() {
 		$payload = array();
 
 		// Scan subscriptions.
-		try {
-			$subscriptions = WC_Helper::get_subscriptions();
-		} catch ( Exception $e ) {
-			$subscriptions = array();
-		}
+		$subscriptions = WC_Helper::get_subscriptions();
+
```
mcliwanow, line 413: I see that `WC_Helper::get_subscriptions();`  doesn't throw, so this change makes sense

plugins/woocommerce/includes/admin/helper/class-wc-helper-updater.php
```diff
@@ -448,11 +445,7 @@ public static function get_update_data() {
 		$payload = array();
 
 		// Scan subscriptions.
-		try {
-			$subscriptions = WC_Helper::get_subscriptions();
-		} catch ( Exception $e ) {
-			$subscriptions = array();
-		}
```
mcliwanow, line 455: Same here

bor0:
@mcliwanow found one issue related to page reloads:

1. Go to Home. Check `wcSettings.admin.wccomHelper.connected_notice` in console. Expected/Actual: `undefined`
2. Navigate to `Extensions`. Check `wcSettings.admin.wccomHelper.connected_notice` in console. Expected: data. Actual: `undefined`
3. Do a page reload. Check `wcSettings.admin.wccomHelper.connected_notice` in console. Expected/Actual: data

 I found the [culprit](https://github.com/woocommerce/woocommerce/blob/2160c88/plugins/woocommerce/client/admin/client/layout/controller.js#L482-L489), introduced by [this](https://github.com/woocommerce/woocommerce/commit/aed4ea63b48234588c2f6683f4502d268bad2d6d). Will look deeper.

KokkieH (approved):
Connection/disconnection, search, discover tab all still work as expected, and I only see requests being logged on the Extensions pages, not on other WC admin pages.

This looks good to me, but I think a second review would be good here.

plugins/woocommerce/includes/admin/helper/class-wc-helper-updater.php
```diff
@@ -448,11 +445,7 @@ public static function get_update_data() {
...
+		$subscriptions = WC_Helper::get_subscriptions();
```
thilinah, line 448: Not handling the exception here can probably cause errors in many places, as `get_subscriptions` can throw an exception whenever there is a non 200 response.

bor0: After https://github.com/woocommerce/woocommerce/pull/57994, [`get_subscriptions` no longer throws](https://github.com/woocommerce/woocommerce/blob/b853322815886a13850e4a7b72ae5a17b549de84/plugins/woocommerce/includes/admin/helper/class-wc-helper.php#L1823) - I had to keep `@throws` in the docblock because phpcs was complaining.

bor0:
@mcliwanow one more round after d51974c, please.

plugins/woocommerce/client/admin/client/layout/controller.js
```diff
@@ -478,15 +478,6 @@ export function updateLinkHref( item, nextQuery, excludedScreens ) {
 
 		// Replace the href so you can see the url on hover.
 		item.href = href;
-
-		item.onclick = ( e ) => {
```
mcliwanow, line 482: I'm not a huga fan of this change. In the main WooCommerce tab it makes not a huge impact, as there there are only "Home", "Customers" and "Extensions" links that have the same `page` parameter. This change however makes the experience pretty worse on the Analytics pages, which all share the same `page`. Before this change I could navigate them without page reload, now page reloads when toggling between Analytics pages.

If we need to make changes here, maybe we should just exclude the "extensions" path from having custom onClick handler?

mcliwanow (commented):
I'm a bit unsure about the JS change introduced. Let's discuss in the comment left closer to the code. 

The rest of the changes looks good.

plugins/woocommerce/includes/admin/helper/class-wc-helper-updater.php
mcliwanow: > I had to keep @throws in the docblock because phpcs was complaining.

Can we rather then ignore the phpcs error with a comment? Keeping `@throws` declaration in the docblock is quite strong indication that there can be errors

plugins/woocommerce/client/admin/client/layout/controller.js
bor0: > now page reloads when toggling between Analytics pages

Damn, nice catch. I thought this code was specific to Extensions, and didn't think it could be used in different places.

> If we need to make changes here, maybe we should just exclude the "extensions" path from having custom onClick handler?

Yea, this makes sense. There already seems to be a mechanism for it: https://github.com/woocommerce/woocommerce/blob/9d12459/packages/js/navigation/src/index.js#L99-L105

bor0: I tried re-using that mechanism but it doesn't prevent the onclick handler. Tackled it in b771c4b12d9333c134b9016bbba93a9543413044, let's see how that one goes

plugins/woocommerce/includes/admin/helper/class-wc-helper-updater.php
bor0: Good point. 85f7641350a011b90181ced2ea076b3fc3936b62

packages/js/navigation/src/index.js
```diff
@@ -23,6 +23,8 @@ export * from './filters';
 export { useConfirmUnsavedChanges } from './hooks/use-confirm-unsaved-changes';
 
 const TIME_EXCLUDED_SCREENS_FILTER = 'woocommerce_admin_time_excluded_screens';
+const TIME_EXCLUDED_SCREENS_URL_UPDATE_FILTER =
```
mcliwanow: Old filter is used in context of:
> Get array of screens that should ignore persisted queries

So having "time" in the name makes _some_ sense. I wonder if new one shouldn't be rather something like 

```php
const NAVIGATION_UPDATE_EXCLUDED_SCREENS_FILTER = 'woocommerce_admin_nav_update_excluded_screens';
```

WDYT?

mcliwanow (approved):
It works as expected now. Thanks for the tweaks.

I left one comment but it's a nitpick, doesn't warrant another review.

packages/js/navigation/src/index.js
bor0: Good point, thank you. 06ae25e381

PR #58242, opened 2025-05-23
[dev] Clean-up ci and tests on demand workflows - remove unused inputs

adimoldovan:

Clean up some unused input choices in `ci.yml` and `tests-on-demand.yml`.
- `on-demand` trigger is not used anymore in any job config. If added back it can be used via the `custom` event, there's no need for an explicit choice.
- `wc-version` is not used.

CI should pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,10 +21,6 @@ on:
         required: true
         default: 'default'
         type: string
-      wc-version:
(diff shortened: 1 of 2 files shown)


joshuatf (approved):
✅ CI is still passing and `wc-version` doesn't appear to be used anywhere else in the workflows.

PR #58265, opened 2025-05-26
Fill out some functionality of the experimental iAPI powered mini cart

samueljseay:

As a follow up to https://github.com/woocommerce/woocommerce/pull/57951 this fills out some of the functionality of the experimental iAPI mini cart, by using conditional rendering in inner blocks to render iAPI directives.

Some of the implemented functionality includes:

* Fix drawer animations, close drawer when clicking outside
* Quantity badge server side rendered, dynamically updated and respect block settings for visibility
* Dynamically updated title text when items are in cart / show empty cart block when they're not (empty cart block incomplete)
* Footer drawer block with cart and checkout link buttons

https://github.com/user-attachments/assets/c0627d41-54af-41d9-93f5-b34626e64efb

3. Reload site on frontend, see the iAPI powered mini cart which has some functionality implemented. You can test things like adding items to cart from shop page and see that dynamic data such as quantity badge and drawer title is updated to reflect the item count.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 
[Experimental] Port more functionality from the existing mini cart to the experimental iAPI mini cart.

</details>


plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
```diff
@@ -29,6 +29,7 @@ export type Store = {
 		cart: Omit< Cart, 'items' > & {
 			items: ( OptimisticCartItem | CartItem )[];
 		};
+		totalItemsInCart: number;
```
luisherranz: Hmm... this is interesting. I wonder if we should expose this type of helpers in the public store or if we should just expose `cart` and relegate these helpers to the individual blocks.

Apart from that, this value should also be populated on the server.

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
```diff
@@ -2,28 +2,98 @@
  * External dependencies
  */
 import { store, getContext } from '@wordpress/interactivity';
+import '@woocommerce/stores/woocommerce/cart';
+import type { Store as WooCommerce } from '@woocommerce/stores/woocommerce/cart';
+
+/**
+ * Internal dependencies
+ */
+import setStyles from './utils/set-styles';
```
luisherranz, line 11: I know this is prior to this PR, but I don't like adding styles dynamically to the page like this. Feels like a hack, and it might not even work with iAPI's client-side navigation 😄

I know it's out of the scope of this PR and even the MiniCart migration initiative, but I wonder why don't we just expose the text and background colors in the Editor, as the Navigation block does.

samueljseay:
@luisherranz you mentioned to me that it appears the quantity badge is not working, but I can't seem to reproduce this in my environment. (See screen recording for what should be functioning). Can you show me the issues you're seeing? :)

plugins/woocommerce/src/Blocks/BlockTypes/MiniCart.php
```diff
@@ -474,15 +495,17 @@ protected function render_experimental_iapi_mini_cart( $attributes, $content, $b
...
+						</span>
 					<?php endif; ?>
 					<?php
 					  // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
 						echo $this->get_cart_price_markup( $attributes );
 					?>
 				</span>
 			</button>
-			<div  data-wp-bind--class="state.drawerOverlayClass" class="is-loading wc-block-components-drawer__screen-overlay wc-block-components-drawer__screen-overlay--is-hidden" aria-hidden="true">
+			<div data-wp-on--click="callbacks.overlayCloseDrawer" data-wp-bind--class="state.drawerOverlayClass" class="wc-block-components-drawer__screen-overlay wc-block-components-drawer__screen-overlay--with-slide-out wc-block-components-drawer__screen-overlay--is-hidden">
```
luisherranz, line 502: Taking over the whole class attribute feels a bit like an anti-pattern to me. First, the classes need to be added in two places (PHP and JS) and be kept in sync, which can lead to bugs if other developers add new classes only in PHP, and it also makes it impossible to modify the class attribute of that element in PHP through filters. So, although it may work for this particular case, in general, I think it would be better to avoid it.

If there are cases in which `data-wp-class--{prefix}` is not enough, we have the [idea of implementing `data-wp-class`](https://github.com/WordPress/gutenberg/pull/65803#issuecomment-2468475673) (default suffix).

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
```diff
@@ -2,28 +2,98 @@
...
+
+type MiniCartContext = {
+	isOpen: boolean;
+	productCountVisibility: 'never' | 'always' | 'greater_than_zero';
+};
+
+// Inject style tags for badge styles based on background colors of the document.
+setStyles();
 
 store( 'woocommerce/mini-cart', {
```
luisherranz: Remember to lock the stories for now.

plugins/woocommerce/src/Blocks/BlockTypes/MiniCart.php
```diff
@@ -448,10 +450,7 @@ protected function render( $attributes, $content, $block ) {
 	 */
 	protected function render_experimental_iapi_mini_cart( $attributes, $content, $block ) {
 		wp_enqueue_script_module( $this->get_full_block_name() );
-
-		$context = array(
-			'isOpen' => false,
-		);
+		$this->register_cart_interactivity( 'I acknowledge that using private APIs means my theme or plugin will inevitably break in the next version of WooCommerce' );
```
luisherranz, line 453: I wonder if it would make sense to move the [caching prevention](https://github.com/woocommerce/woocommerce/blob/947b751013c9f7cd95d4a229203062f613820a10/plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php#L212-L214) from the blocks to the `register_cart_interactivity` method. If a block is registering the cart, it will most likely use the cart info to populate the PHP (which makes the HTML non-cachable).

plugins/woocommerce/src/Blocks/BlockTypes/MiniCart.php
```diff
@@ -461,7 +460,29 @@ protected function render_experimental_iapi_mini_cart( $attributes, $content, $b
...
+
+		$cart            = $this->get_cart_instance();
+		$cart_item_count = $cart ? $cart->get_cart_contents_count() : 0;
+		$cart_item_text  = '';
+
+		if ( 'always' === $product_count_visibility ) {
+			$cart_item_text = $cart_item_count;
+		} elseif ( 'never' !== $product_count_visibility && $cart_item_count > 0 ) {
+			$cart_item_text = $cart_item_count;
+		}
```
luisherranz: Why do we need to do this? Wouldn't be a better approach to not render the badge `span` at all if its visibility is `never`?

plugins/woocommerce/src/Blocks/BlockTypes/MiniCart.php
```diff
@@ -474,15 +495,17 @@ protected function render_experimental_iapi_mini_cart( $attributes, $content, $b
 						echo $icon;
 					?>
 					<?php if ( 'never' !== $product_count_visibility ) : ?>
-						<span class="wc-block-mini-cart__badge" style="<?php echo esc_attr( $styles ); ?>"></span>
+						<span hidden data-wp-bind--hidden="!state.badgeIsVisible" data-wp-text="state.cartItemCount" class="wc-block-mini-cart__badge" style="<?php echo esc_attr( $styles ); ?>">
```
luisherranz: You shouldn't need the `hidden` attribute. The iAPI's server processing should add it depending on `state.badgeIsVisible`. Same for the inner text and `state.cartItemCount`.

plugins/woocommerce/src/Blocks/BlockTypes/EmptyMiniCartContentsBlock.php
```diff
@@ -11,4 +13,41 @@ class EmptyMiniCartContentsBlock extends AbstractInnerBlock {
...
+	 *
+	 * @param array    $attributes Block attributes.
+	 * @param string   $content    Block content.
+	 * @param WP_Block $block      Block instance.
+	 * @return string Rendered block type output.
+	 */
+	protected function render_experimental_filled_mini_cart_contents( $attributes, $content, $block ) {
+		ob_start();
+		?>
+		<div data-wp-bind--hidden="woocommerce/mini-cart::!state.cartIsEmpty" data-wp-interactive="woocommerce/empty-mini-cart-contents" class="wp-block-woocommerce-empty-mini-cart-contents-block">
```
luisherranz: I think it's fine to use `woocommerce/mini-cart` as the namespace here.

plugins/woocommerce/src/Blocks/BlockTypes/FilledMiniCartContentsBlock.php
```diff
@@ -11,4 +13,41 @@ class FilledMiniCartContentsBlock extends AbstractInnerBlock {
...
+	 *
+	 * @param array    $attributes Block attributes.
+	 * @param string   $content    Block content.
+	 * @param WP_Block $block      Block instance.
+	 * @return string Rendered block type output.
+	 */
+	protected function render_experimental_filled_mini_cart_contents( $attributes, $content, $block ) {
+		ob_start();
+		?>
+		<div data-wp-bind--hidden="woocommerce/mini-cart::state.cartIsEmpty" data-wp-interactive="woocommerce/filled-mini-cart-contents" class="wp-block-woocommerce-filled-mini-cart-contents-block">
```
luisherranz: Same here.

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartFooterBlock.php
```diff
@@ -11,4 +13,80 @@ class MiniCartFooterBlock extends AbstractInnerBlock {
...
+	 *
+	 * @param array    $attributes Block attributes.
+	 * @param string   $content    Block content.
+	 * @param WP_Block $block      Block instance.
+	 * @return string Rendered block type output.
+	 */
+	protected function render_experimental_iapi_mini_cart_footer( $attributes, $content, $block ) {
+		ob_start();
+
+		$cart                             = $this->get_cart_instance();
```
luisherranz, line 28: Hmm... it feels a bit strange to have the cart information in 2 different formats: one in the `state` from the Store API and another in this instance from `WC()->cart`.

I wonder if it would make sense to rely only on the one that is going to be serialized?

```php
$cart = wp_interactivity_state( 'woocommerce' )['cart'];
```

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartTitleItemsCounterBlock.php
```diff
@@ -11,4 +13,68 @@ class MiniCartTitleItemsCounterBlock extends AbstractInnerBlock {
...
+			'pluralItemsText'   => $plural,
+		);
+
+		ob_start();
+		?>
+		<?php // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
+		<span data-wp-text="state.itemsInCartText" data-wp-interactive="woocommerce/mini-cart-title-items-counter-block" <?php echo wp_interactivity_data_wp_context( $context ); ?> class="wp-block-woocommerce-mini-cart-title-items-counter-block">
+					<?php
+						// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+						printf( $cart_item_text, $cart_item_count );
```
luisherranz: This should be populated by the iAPI by providing `state.itemsInCartText` in the server.

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartTitleItemsCounterBlock.php
```diff
@@ -11,4 +13,68 @@ class MiniCartTitleItemsCounterBlock extends AbstractInnerBlock {
...
+
+		// translators: item is an item in cart.
+		$singular = __( '(%d item)', 'woocommerce' );
+		// translators: items is items in a cart.
+		$plural = __( '(%d items)', 'woocommerce' );
+
+		$context = array(
+			'singularItemsText' => $singular,
+			'pluralItemsText'   => $plural,
+		);
```
luisherranz: As these don't change (do they?), maybe we try using `wp_interactivity_config` instead of `context`.

luisherranz (changes requested):
It was a problem in my environment. I've already fixed it, sorry.

This looks really good! 👏👏

A few comments/suggestions 👇

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
samueljseay: @luisherranz yes, I agree.

I think once we are done the initial migration we are left with some technical debt because my approach is just to port things as they are across to iAPI and in a few cases it feels like there is a better approach or some of the existing complexity is not needed.

We could add some issues to track what technical debt is left over at the end.

plugins/woocommerce/src/Blocks/BlockTypes/MiniCart.php
samueljseay: > Taking over the whole class attribute feels a bit like an anti-pattern to me. First, the classes need to be added in two places (PHP and JS) and be kept in sync

Yep I felt like this was technical debt in my approach to porting across. Maybe a lazy port in this case. I will adjust this to add directives per class, I just didn't like the long verbose class names, but I felt that in future we could solve that by simplifying the css.

The object based approach looks useful for the future.

plugins/woocommerce/src/Blocks/BlockTypes/MiniCart.php
samueljseay: @luisherranz good idea thanks for considering that.  Will fix.

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartFooterBlock.php
samueljseay: Another good point, consistency would be best here. Thanks

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartTitleItemsCounterBlock.php
samueljseay: These should not change. Will move to config 👍🏻

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
samueljseay: yeah this could be derived state in mini cart store. 


> Apart from that, this value should also be populated on the server.

Good point 👍🏻

samueljseay: Resolved in https://github.com/woocommerce/woocommerce/pull/58265/commits/690f02887f77ffd2f408331dff8af7dbb5e71c64

plugins/woocommerce/src/Blocks/BlockTypes/MiniCart.php
samueljseay: We don't render if its never so this condition is not really needed. I believe now that I've refactored badgeIsVisible this should be solved.

plugins/woocommerce/src/Blocks/BlockTypes/MiniCart.php
samueljseay: Yeah, this was my bad, I forgot I could just populate `badgeIsVisible` in state. 

This should be solved now that I've refactored the cart item count to be populated in server state and I've removed the `hidden` attribute.

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartTitleItemsCounterBlock.php
samueljseay: Good point, I really got caught up in this pattern for some reason 😅 

Resolved in https://github.com/woocommerce/woocommerce/pull/58265/commits/b1d5a2ea53ba04144cc91b6e59d5d6a824b3a5d7

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartTitleItemsCounterBlock.php
samueljseay: Resolved in https://github.com/woocommerce/woocommerce/pull/58265/commits/5c6781f358e6711b4100b3a71101653b35cbc756

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartFooterBlock.php
samueljseay: @luisherranz I'm not sure this is worth it as part of this PR.

StoreAPI uses the exact same cart under the hood, so there is some consistency. To some degree `WC_Cart` could be thought of like a `Presenter` class in this case. StoreAPI ultimately uses the exact same method to get the cart instance.

I could duplicate some of the presentation logic from WC_Cart and run it over the data we serialized in state (I think that'd be ok because the filters would have already been run when we got the response from store API), but I think if we are to introduce new classes we should consider a separate PR.

That also makes me think, maybe when we populate the state we could consider not doing a rest request, but rather extracting the logic that serializes the cart to JSON so it could be reused to populate the iAPI state and bring more consistency that way. 

Consolidation is a good idea, but I'm just thinking I would rather not do it as part of this PR. I'd suggest we create an issue to look at this separately.

but wdyt?

plugins/woocommerce/src/Blocks/BlockTypes/MiniCart.php
samueljseay: Resolved in https://github.com/woocommerce/woocommerce/pull/58265/commits/caa92580aaa34979bec5dde695f794f1d6d81b47

samueljseay:
@luisherranz I think this is ready for some re-review 🙇

plugins/woocommerce/src/Blocks/BlockTypes/MiniCart.php
samueljseay: I forgot about actually addressing this one. If it's ok I'd like to do it in a follow up. I'd just like to get this series of PRs merged as soon as possible.

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
luisherranz: Perfect. Thanks, Sam.

plugins/woocommerce/src/Blocks/BlockTypes/MiniCart.php
luisherranz: Sure 🙂

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartFooterBlock.php
luisherranz: Yes, sorry, it wasn't a comment to do it in this PR. I think it's good to look at it later, including avoiding the need to make a REST request (which apparently has quite a bit of overhead).

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
```diff
@@ -27,73 +27,108 @@ type MiniCartContext = {
...
-		get cartItemCount() {
-			return wooStoreState.totalItemsInCart;
-		},
-	},
+// Destructure state in an empty call to the store, to ensure that state can be correctly typed.
+const { state: miniCartState } = store< MiniCart >(
+	'woocommerce/mini-cart',
+	{},
+	{ lock: true }
+);
```
luisherranz: This should not be necessary. It is a known limitation, but I think it is best resolved by typing the derived state that depends on other parts of the state. It is [documented here](https://developer.wordpress.org/block-editor/reference-guides/interactivity-api/core-concepts/using-typescript/#typing-the-derived-state).

Pushing a change to fix it.

luisherranz: Fixed here: https://github.com/woocommerce/woocommerce/pull/58265/commits/a904ee0742f8d82272b2b854bf93a9465aeff45a

luisherranz (approved):
LGTM, Sam!! Thanks for the changes 🙂👏

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
samueljseay: Thanks that was driving me crazy, I didn't realize we could workaround it that way.

PR #58268, opened 2025-05-26
Refactor price utils for use in iAPI context, display cart subtotal in iAPI mini cart footer block

samueljseay:

As a follow up to https://github.com/woocommerce/woocommerce/pull/58265 this refactors price utils so they can be used in module / iAPI context as well as the existing context they're used. The existing interfaces work as before, but have added some additional functions so that data that would come from wcSettings can be passed in instead.

**Also:** we also remove dependency on decodeEntities from the `@wordpress/html-entities` package to avoid module build errors. This is a revert basically to a change that was made to consolidate use of that library. It was nice in theory but the entity decoding we revert to is less lines of code and less complexity (we don't really need `html-entities` extra features for this one instance).

This also utilizes the refactor to display the cart subtotal in the experimental iAPI mini cart drawer

Screenshot 2025-06-02 at 1 46 06 PM

3. Reload site on frontend, see the iAPI powered mini cart which has some functionality implemented.
4. Add some items to your cart on the shop page
5. Open the cart drawer and see that the subtotal down the bottom matches the total $ value of the cart.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Refactor price utils to support passing cart data in, utilize it in the experimental iAPI mini cart drawer subtotal.

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Blocks/BlockTypes/MiniCartFooterBlock.php
```diff
@@ -32,17 +32,28 @@ protected function render_experimental_iapi_mini_cart_footer( $attributes, $cont
 		$subtotal                         = $display_cart_price_including_tax ? $cart->get_subtotal_tax() : $cart->get_subtotal();
 		$formatted_amount                 = wc_price( $subtotal );
 
+		wp_interactivity_config(
+			$this->get_full_block_name(),
+			array(
+				'displayCartPriceIncludingTax' => $display_cart_price_including_tax,
+			)
+		);
```
luisherranz, line 49: This is interesting. Why did you decide to put this setting in the block's namespace instead of `woocommerce`? (I'm not saying it's wrong, but I just want to know what reasoning you followed.)

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartFooterBlock.php
```diff
@@ -32,17 +32,28 @@ protected function render_experimental_iapi_mini_cart_footer( $attributes, $cont
...
+
+		wp_interactivity_state(
+			$this->get_full_block_name(),
+			array(
+				'formattedSubtotal' => $formatted_amount,
```
luisherranz: You never get to see it because by the time the overlay opens the JavaScript has already loaded and the value has already been computed in the client, but if you look at the HTML, what is being serialized in the state for this property is something like this: 

`"formattedSubtotal": "\u003Cspan class=\"woocommerce-Price-amount amount\"\u003E\u003Cbdi\u003E72,00&nbsp;\u003Cspan class=\"woocommerce-Price-currencySymbol\"\u003E&euro;\u003C/span\u003E\u003C/bdi\u003E\u003C/span\u003E"`

For some reason, `data-wp-text` isn't adding anything at all. I'm not [quite sure why](https://github.com/WordPress/wordpress-develop/blob/1e04fc0864544f5a224dbe4718979367ab2a208c/src/wp-includes/interactivity-api/class-wp-interactivity-api.php#L1037-L1053), but it's not a big concern.

luisherranz (approved):
This looks good to me, although I've only tested the new MiniCart block.

Should we add instructions to test that `packages/prices/utils/price.ts` is still working fine elsewhere, and ask someone else from WooCommerce who is familiar with it to take a look?

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartFooterBlock.php
samueljseay: 🤔 Thats a good question. I often think of iAPI context/state/config as being similar to React, where we want to keep data close to where it's being used. Originally I made this context on that namespace but then I moved it to config because it doesn't need to be reactive.

But in the case of config that boundary doesn't really exist in the interactivity API. ie putting it here on the block namespace specifically doesn't really change anything about how global it is. 

Maybe the only thing that's left is, I don't like how if many of our blocks put their config onto top level `woocommerce` it feels hard to reason about what the end shape of the `woocommerce` config is. I could see an argument for only putting config we want to share across a lot of blocks there and only registering it in one or 2 places so we can easily keep track of what's assigned to `woocommerce` namespace.

But I don't feel strongly about it.

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartFooterBlock.php
samueljseay: Ah dang I forgot these methods output html 😭 that's quite frustrating.

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartFooterBlock.php
luisherranz: > I could see an argument for only putting config we want to share across a lot of blocks there and only registering it in one or 2 places so we can easily keep track of what's assigned to `woocommerce` namespace.

Nice. We could follow this guideline: initially, we always put the config in the block namespaces, and only when more than one block needs to access the same config, we consider moving it to the `woocommerce` namespace.

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartFooterBlock.php
luisherranz: This would be a case where you don't need to populate this value on the server, since the HTML won't ever be visible before the JavaScript loads, although it would be more "correct" to do so. It's your call, but if you prefer to just delete it, that's fine with me too.

samueljseay: I've used html tag processor to get the raw text content of the initial value for state server side. It's an improvement at least. We can iterate on this if we need to.

https://github.com/woocommerce/woocommerce/pull/58268/commits/2b6e8d1cf079b17b5ad5de31468a52eec61372bc

samueljseay:
> This looks good to me, although I've only tested the new MiniCart block.

> Should we add instructions to test that packages/prices/utils/price.ts is still working fine elsewhere, and ask someone else from WooCommerce who is familiar with it to take a look?

I think some of this is covered by tests, but I'll reach out to someone to have a look and do some manual testing of this.

samueljseay:
Hey @opr wondering if you'd mind taking a look at this. I'm not sure what manual testing could improve confidence here so open to some suggestion.

I've refactored price utils to in theory operate the same as before but expose some of the internals for reuse in the mini cart. I want to ensure these changes haven't broken anything for existing price formatting/display and it seems like asking cart/checkout folks would make the most sense. 🙇

plugins/woocommerce/client/blocks/packages/prices/utils/currency.ts
```diff
@@ -0,0 +1,135 @@
...
+		decimalSeparator,
+		minorUnit
+	) }${ suffix }`;
+
+	// Use a textarea to decode HTML currency symbols.
+	// This used to use @wordpress/html-entities, but that was not necessary
+	// for this simple use case and avoids issues with ESM modules / @wordpress packages.
+	const txt = document.createElement( 'textarea' );
+	txt.innerHTML = formattedValue;
+	return txt.value;
```
opr, line 111: I ran this through Claude code and the thing it focused on most was this. Besides its claims that this is an XSS Vector (I don't think it is, so we can ignore it) it also claims this is unnecessary. However, I looked into it and I do think it is necessary because the currency can arrive from either API responses (where it is encoded) or from the `wc.wcSettings.allSettings.currency` object where it _isn't_ encoded.

All that to say: 👍🏼

opr (commented):
Hi @samueljseay - the changes seem OK to me. Some ideas for manual testing:

1. Change various settings related to the currency in WC -> Settings -> General and ensure the Cart/Checkout block front-end matches expectations, especially relating to the currency symbol. (this tests the wcSettings path)
2. Go to the Checkout block, then in a new tab change the currency symbol. Without reloading, go back to the original tab and select a different shipping method, or change address, ensure the currency symbol updates and there are no encoding issues. (tests the currency in API response path).

Does this sound good?

plugins/woocommerce/client/blocks/packages/prices/utils/currency.ts
samueljseay: @opr yeah thanks for looking close at this. Fwiw this is exactly what `@wordpress/html-entities` does (except it does more stuff than just that, but it still does this exact encoding via textarea)

It's a bit sad that JS doesn't already have a consistent API for doing this and in 2025 this is still the best way to do this.

samueljseay:
Thanks so much @opr I tested those pathways and everything looks great in both RTL and LTR with various settings currency settings and various symbols.

PR #58305, opened 2025-05-27
Fix flaky tests: variations in product editor

kmanijak:

This PR fixes a flaky tests. The test is about manipulating variations in the product editor. Checking the flaky test trace, it seems like "Variation" tab is clicked but the tab doesn't change (stays at initial "General" tab):

image

I was able to achieve that state on my local environment when I clicked an "Attributes" tab super-fast (similar to Playwright, I guess) but the General tab stayed open:

image

Given this, I added `expect( ... ).toPass()` making sure the tab has changed and if not, click again.

Closes https://linear.app/a8c/issue/WOOPLUG-10/flaky-test-can-individually-edit-variations
Closes https://linear.app/a8c/issue/WOOPLUG-3416/flaky-test-can-bulk-edit-variations
Closes https://linear.app/a8c/issue/WOOPLUG-3552/flaky-test-can-delete-all-variations
Closes https://linear.app/a8c/issue/WOOPLUG-3628/flaky-test-can-manage-stock-levels
Closes https://linear.app/a8c/issue/WOOPLUG-3507/flaky-test-can-set-variation-defaults
Closes https://linear.app/a8c/issue/WOOPLUG-3418/flaky-test-can-remove-a-variation

| Before | After |
| ------ | ----- |
|        |       |

1.
2.
3.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/tests/e2e-pw/tests/product/update-variations.spec.js
```diff
@@ -106,6 +106,25 @@ test.describe( 'Update variations', { tag: tags.GUTENBERG }, () => {
...
+				// so we need to wait make sure some content from the variations tab is visible.
+				const expandButton = page
+					.getByRole( 'link', { name: 'Expand' } )
+					.first();
+
+				await expect( expandButton ).toBeVisible();
+			} ).toPass( { timeout: 3_000 } );
+		} );
+	}
+
```
gigitux: NIT: I'm not sure that we should use `test.step` inside a util function. I think that utility functions should be focused on performing actions without test-specific constructs. At the same time, I'm aware that this util is used only in this file and not even exported, so I don't have a strong opinion.

plugins/woocommerce/tests/e2e-pw/tests/product/update-variations.spec.js
```diff
@@ -106,6 +106,25 @@ test.describe( 'Update variations', { tag: tags.GUTENBERG }, () => {
...
+					.click();
+
+				// Sometimes the click on link is too fast and the initial tab (General) is still visible
+				// so we need to wait make sure some content from the variations tab is visible.
+				const expandButton = page
+					.getByRole( 'link', { name: 'Expand' } )
+					.first();
+
+				await expect( expandButton ).toBeVisible();
+			} ).toPass( { timeout: 3_000 } );
```
gigitux: Is the timeout needed?

gigitux (approved):
Thanks for improving this test! I’ve left two minor comments. I don’t have a strong opinion about moving `test.step` outside the function, but I think we should avoid hardcoding the timeout if possible, otherwise, there’s a risk that others might start adding custom timeouts as well.

plugins/woocommerce/tests/e2e-pw/tests/product/update-variations.spec.js
kmanijak: I don't have strong opinion either. But to be fair, it's something I took from other core tests, like [this one](https://github.com/woocommerce/woocommerce/blob/0b7bc8225c0010451dd49925d9c52c5c290a3eca/plugins/woocommerce/tests/e2e-pw/tests/product/product-linked-products.spec.js#L41C3-L64C4). I kind of look at this more like "repeatable steps" rather than util.

gigitux: It makes sense! Let's keep it! 💪

plugins/woocommerce/tests/e2e-pw/tests/product/update-variations.spec.js
kmanijak: By default it will wait for the default assertion timeout which in our case is what, 30s? So this actually decreases the time action inside is repeated. I don't have a strong opinion, I can remove it!

gigitux: I’d prefer removing it to avoid encouraging others to add hardcoded timeouts just to make E2E tests pass

kmanijak: Removed in https://github.com/woocommerce/woocommerce/pull/58305/commits/4e0099ec8e025401335d27297f6823fadcefaa02!

> I’d prefer removing it to avoid encouraging others to add hardcoded timeouts just to make E2E tests pass 

But to be clear, this timeout is the other way around. It actually LIMITS the step time. Otherwise the step will be repeated until overall default assertions time which is I think 30s. But no limit seems cleaner anyway.

PR #58307, opened 2025-05-27
Fix tab navigation inconsistency between QuantityStepper and QuantitySelector components

### LINKED ISSUE #58011: Tab navigation inconsistency between QuantityStepper and QuantitySelector components
### Prerequisites

### Describe the bug

The `QuantityStepper` component in the `AddToCartsWithOptions` block has an incorrect tab order and accessibility issues compared to the `QuantitySelector` component used in the `Mini Cart Drawer` block.

While the `QuantityStepper` and `QuantitySelector` components are different, they appear similar and serve the same purpose, so they should provide a consistent user experience.

There are two main issues:

1. **Incorrect tab order**: When using the Tab key to navigate:
   - QuantityStepper (AddToCartsWithOptions): Tab moves to minus (-) button → input field → plus (+) button

      https://github.com/user-attachments/assets/bd6d5f7a-871c-4047-b69a-4330b1d7933e

   - QuantitySelector (Mini Cart Drawer): Tab moves to input field → minus (-) button → plus (+) button _(correct order)_
   
      https://github.com/user-attachments/assets/e756847d-33b4-499a-8ba7-a8441ade2cf0

2. **Missing disabled state for minus (-) button**: When quantity is 1:
   - QuantitySelector: The minus (-) button is properly disabled and unfocusable _(correct behavior)_  
 
      https://github.com/user-attachments/assets/d329f00a-e152-4d2f-9291-ac3fbf78a15d

   - QuantityStepper: The minus (-) button is NOT disabled and still focusable _(incorrect behavior)_

      https://github.com/user-attachments/assets/be3154d0-78b0-495c-926b-7a50af2d93d4

I'm not sure whether this is intended behavior or something else. Feel free to correct me or guide me if I'm missing anything.

### Expected behavior

The `QuantityStepper` component should follow the same tab order as the `QuantitySelector` component:
1. Tab should first focus on the input field
2. Then move to the minus (-) button
3. Finally to the plus (+) button

When quantity is 1, the minus (-) button should be:
- Visually disabled
- Not clickable
- Not focusable via tab navigation

### Actual behavior

1. Tab navigation starts with the minus (-) button instead of the input field
2. When quantity is 1, the minus (-) button is not disabled and remains focusable
3. The minus (-) button can be focused via tab and clicked, even though it performs no action

### Steps to reproduce

1. Install and activate the `WooCommerce Beta Tester plugin`
2. Go to *WooCommerce Admin Test Helper > Features* and enable `experimental-blocks` and `blockified-add-to-cart`
3. Go to *Appearance > Editor > Templates > Single Product*, select the `Add to Cart with Options` block and update it to the blockified one
4. View a product in the frontend
5. Use the Tab key to navigate through the `QuantityStepper` component
6. Observe that tab order is: minus (-) button → input field → plus (+) button
7. Set quantity to 1 (minimum value)
8. Notice the minus (-) button is not disabled (unlike in `QuantitySelector`)
9. Press Tab and observe it still focuses on the minus (-) button
10. Compare this behavior with the QuantitySelector in the `Mini Cart Drawer` block:
    - To test the `QuantitySelector` in the `Mini Cart Drawer` block, click on the mini cart icon in the header
    - Make sure to have at least one product added to the cart so the QuantitySelector shows, and use the Tab key to navigate and see that the tab order is: input field → minus (-) button → plus (+) button
11. Notice the minus (-) button is properly disabled when quantity is 1 in Mini Cart

### WordPress Environment

<details>
<summary>System Status Report</summary>

### wp-core ###

version: 6.8.1
site_language: en_US
user_language: en_US
timezone: +00:00
permalink: undefined
https_status: false
multisite: false
user_registration: 0
blog_public: 1
default_comment_status: open
environment_type: local
user_count: 1
dotorg_communication: true

### wp-paths-sizes ###

wordpress_path: /var/www/html
wordpress_size: loading...
uploads_path: /var/www/html/wp-content/uploads
uploads_size: loading...
themes_path: /var/www/html/wp-content/themes
themes_size: loading...
plugins_path: /var/www/html/wp-content/plugins
plugins_size: loading...
fonts_path: /var/www/html/wp-content/uploads/fonts
fonts_size: loading...
database_size: loading...
total_size: loading...

### wp-active-theme ###

name: Twenty Twenty-Five (twentytwentyfive)
version: 1.2
author: the WordPress team
author_website: https://wordpress.org
parent_theme: none
theme_features: core-block-patterns, post-thumbnails, responsive-embeds, editor-styles, html5, automatic-feed-links, widgets-block-editor, block-templates, post-formats, editor-style
theme_path: /var/www/html/wp-content/themes/twentytwentyfive
auto_update: Disabled

### wp-themes-inactive (3) ###

Storefront: version: 4.6.1, author: Automattic, Auto-updates disabled
Twenty Twenty-Four: version: 1.3, author: the WordPress team, Auto-updates disabled
Twenty Twenty-Three: version: 1.6, author: the WordPress team, Auto-updates disabled

### wp-plugins-active (2) ###

WooCommerce: version: 9.9.0-dev, author: Automattic (latest version: 9.9.0-beta.1), Auto-updates disabled
WooCommerce Beta Tester: version: 3.0.0, author: WooCommerce, Auto-updates disabled

### wp-plugins-inactive (2) ###

Akismet Anti-spam: Spam Protection: version: 5.3.7, author: Automattic - Anti-spam Team (latest version: 5.4), Auto-updates disabled
Hello Dolly: version: 1.7.2, author: Matt Mullenweg, Auto-updates disabled

### wp-media ###

image_editor: WP_Image_Editor_Imagick
imagick_module_version: 1691
imagemagick_version: ImageMagick 6.9.11-60 Q16 aarch64 2021-01-25 https://imagemagick.org
imagick_version: 3.6.0
file_uploads: 1
post_max_size: 1G
upload_max_filesize: 1G
max_effective_size: 1 GB
max_file_uploads: 20
imagick_limits: 
	imagick::RESOURCETYPE_AREA: 122 MB
	imagick::RESOURCETYPE_DISK: 1073741824
	imagick::RESOURCETYPE_FILE: 786432
	imagick::RESOURCETYPE_MAP: 512 MB
	imagick::RESOURCETYPE_MEMORY: 256 MB
	imagick::RESOURCETYPE_THREAD: 1
	imagick::RESOURCETYPE_TIME: 9.2233720368548E+18
imagemagick_file_formats: 3FR, 3G2, 3GP, AAI, AI, APNG, ART, ARW, AVI, AVIF, AVS, BGR, BGRA, BGRO, BIE, BMP, BMP2, BMP3, BRF, CAL, CALS, CANVAS, CAPTION, CIN, CIP, CLIP, CMYK, CMYKA, CR2, CR3, CRW, CUR, CUT, DATA, DCM, DCR, DCX, DDS, DFONT, DNG, DPX, DXT1, DXT5, EPDF, EPI, EPS, EPS2, EPS3, EPSF, EPSI, EPT, EPT2, EPT3, ERF, FAX, FILE, FITS, FRACTAL, FTP, FTS, G3, G4, GIF, GIF87, GRADIENT, GRAY, GRAYA, GROUP4, H, HALD, HDR, HEIC, HISTOGRAM, HRZ, HTM, HTML, HTTP, HTTPS, ICB, ICO, ICON, IIQ, INFO, INLINE, IPL, ISOBRL, ISOBRL6, J2C, J2K, JBG, JBIG, JNG, JNX, JP2, JPC, JPE, JPEG, JPG, JPM, JPS, JPT, JSON, K25, KDC, LABEL, M2V, M4V, MAC, MAGICK, MAP, MASK, MAT, MATTE, MEF, MIFF, MKV, MNG, MONO, MOV, MP4, MPC, MPG, MRW, MSL, MTV, MVG, NEF, NRW, NULL, ORF, OTB, OTF, PAL, PALM, PAM, PATTERN, PBM, PCD, PCDS, PCL, PCT, PCX, PDB, PDF, PDFA, PEF, PES, PFA, PFB, PFM, PGM, PGX, PICON, PICT, PIX, PJPEG, PLASMA, PNG, PNG00, PNG24, PNG32, PNG48, PNG64, PNG8, PNM, POCKETMOD, PPM, PREVIEW, PS, PS2, PS3, PSB, PSD, PTIF, PWP, RADIAL-GRADIENT, RAF, RAS, RAW, RGB, RGBA, RGBO, RGF, RLA, RLE, RMF, RW2, SCR, SCT, SFW, SGI, SHTML, SIX, SIXEL, SPARSE-COLOR, SR2, SRF, STEGANO, SUN, TEXT, TGA, THUMBNAIL, TIFF, TIFF64, TILE, TIM, TTC, TTF, TXT, UBRL, UBRL6, UIL, UYVY, VDA, VICAR, VID, VIDEO, VIFF, VIPS, VST, WBMP, WEBM, WEBP, WMV, WPG, X, X3F, XBM, XC, XCF, XPM, XPS, XV, XWD, YCbCr, YCbCrA, YUV
gd_version: bundled (2.1.0 compatible)
gd_formats: GIF, JPEG, PNG, WebP, BMP
ghostscript_version: 9.53.3

### wp-server ###

server_architecture: Linux 6.6.12-linuxkit aarch64
httpd_software: Apache/2.4.54 (Debian)
php_version: 7.4.33 64bit
php_sapi: apache2handler
max_input_variables: 1000
time_limit: 30
memory_limit: 128M
admin_memory_limit: 256M
max_input_time: -1
upload_max_filesize: 1G
php_post_max_size: 1G
curl_version: 7.74.0 OpenSSL/1.1.1n
suhosin: false
imagick_availability: true
pretty_permalinks: true
static_robotstxt_file: false
current: 2025-05-14T21:29:53+00:00
utc-time: Wednesday, 14-May-25 21:29:53 UTC
server-time: 2025-05-14T21:29:50+00:00

### wp-database ###

extension: mysqli
server_version: 11.4.5-MariaDB-ubu2404
client_version: mysqlnd 7.4.33
max_allowed_packet: 16777216
max_connections: 151

### wp-constants ###

WP_HOME: http://localhost:8888
WP_SITEURL: http://localhost:8888
WP_CONTENT_DIR: /var/www/html/wp-content
WP_PLUGIN_DIR: /var/www/html/wp-content/plugins
WP_MEMORY_LIMIT: 40M
WP_MAX_MEMORY_LIMIT: 256M
WP_DEBUG: true
WP_DEBUG_DISPLAY: true
WP_DEBUG_LOG: true
SCRIPT_DEBUG: true
WP_CACHE: false
CONCATENATE_SCRIPTS: undefined
COMPRESS_SCRIPTS: undefined
COMPRESS_CSS: undefined
WP_ENVIRONMENT_TYPE: local
WP_DEVELOPMENT_MODE: undefined
DB_CHARSET: utf8
DB_COLLATE: undefined

### wp-filesystem ###

wordpress: writable
wp-content: writable
uploads: writable
plugins: writable
themes: writable
fonts: writable

</details>

### Isolating the problem

amitraj2203:

This PR addresses accessibility and UX inconsistencies between the `QuantityStepper` component used in the `AddToCartsWithOptions` block and the `QuantitySelector` component used in the `Mini-Cart` Drawer block. While these components are structurally different, they provide a similar user interface and should behave consistently.

- Specifically, this PR includes:
  - Reordered the tab index in `QuantityStepper` so that:
  - Focus moves first to the input field, then the minus (-) button, and finally the plus (+) button - matching the `QuantitySelector` component behavior.
  - ~Applied correct disabled, unfocusable, and unclickable behavior to the minus (-) button when quantity equals the minimum allowed value.~

Part Of: #58011 

#### Blockified: 
https://github.com/user-attachments/assets/35cd31e6-a400-43f5-bcd5-6eb0e920ef03

#### Non-blockified: 
https://github.com/user-attachments/assets/4d07b5b4-4bf9-460c-899a-9097c86981d1

1. Install and activate the `WooCommerce Beta Tester plugin`
2. Go to *WooCommerce Admin Test Helper > Features* and enable `experimental-blocks` and `blockified-add-to-cart`
3. Go to *Appearance > Editor > Templates > Single Product*, select the `Add to Cart with Options` block and update it to the blockified one
4. View a product in the frontend
5. Use the Tab key to navigate through the `QuantityStepper` component
6. Ensure tab order is: **input field → plus (-) button → minus (+) button**.
7. ~Set quantity to the minimum allowed value (e.g., 1 or custom minimum via filter).~
8. ~Confirm minus (-) button is:~
   * ~Not focusable with tab~
   * ~Not clickable~

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


Aljullu (commented):
Thanks for working on this, @amitraj2203!

>   - Reordered the tab index in `QuantityStepper` so that:
>  - Focus moves first to the input field, then the minus (-) button, and finally the plus (+) button - matching the `QuantitySelector` component behavior.

This works great and code changes look good too. :100: 

>   - Applied correct disabled, unfocusable, and unclickable behavior to the minus (-) button when quantity equals the minimum allowed value.

This is also working as described, but we should try to implement this using the [Interactivity API](https://developer.wordpress.org/block-editor/reference-guides/interactivity-api/).

With it, it shouldn't be needed to query HTML elements directly, instead you could add a directive like `data-wp-bind--disabled="state.allowsDecrease"` to the minus button or something along those lines. Then `allowsDecrease` could be a state in the [stores](https://github.com/woocommerce/woocommerce/blob/8ddea4bf169c071167726833a7cc8c32653ecd84/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts#L120) that checks whether `value - step` is higher or equal than `minValue`. Similarly, we should implement the same for the plus button, as it's technically possible with custom code or extensions that products might have max quantities too.

Let me know how this sounds or if you have any doubts. :slightly_smiling_face:

amitraj2203:
Thanks for the review @Aljullu, I've now updated it accordingly. Please let me know if everything looks good now or if any further changes are needed.

Aljullu (commented):
Great job! Code looks good and it's working mostly fine, I just found one small glitch in grouped products:

If the first product's quantity is increased, then the minus button of the other products is not disabled correctly:

image

Same the other way around, if the first product can't be reduced, all minus buttons get disabled:

image

Can you reproduce as well? I guess that's because this selector is too generic and always selects the first input:

```JS
document.querySelector(
	'.wc-block-components-quantity-selector__input'
)
```

Maybe we could use the input ID or something similar to properly select the correct element? (If you go this way, I would advise rebasing or merging `trunk` first, as we modified the input IDs in https://github.com/woocommerce/woocommerce/pull/57927).

amitraj2203:
Thanks for catching this @Aljullu! You're absolutely right about the issue. 

The problem is that `const { productId } = getContext< Context >()` was giving the ID of the grouped product (e.g., 35 for "Logo Collection") not the IDs of individual products inside that grouped product (e.g., 15 for "Hoodie", 16 for "T-Shirt").

We are using a **single shared store (state.allowsIncrease and state.allowsDecrease)** for *all* quantity steppers. However, each quantity input corresponds to a different product row (e.g. ID 15, 16), but **the store logic doesn't distinguish between them**. It checks all inputs and returns false if *any* of them are at their min/max.

As a result, **both** buttons become disabled if **any one** input hits its min/max, even if others are still adjustable.

To fix this issue, I think we need to:

1. **Make the input selection context-aware**: Instead of using the generic `document.querySelector( '.wc-block-components-quantity-selector__input' )`, we should use the product ID from context to find the specific input: `document.getElementById( quantity_${ productId } )`.

2. **Ensure each product gets its own context**: Each individual product in a grouped product should have its own interactive context with the correct `productId`.
image (10)

We can also apply this approach to **all quantity inputs**, not only for grouped products. For simple/variable products, `const { productId } = getContext< Context >()` will return the correct productId of that specific product, and for grouped products, each individual product will have its own context with its own productId.

This way, each quantity stepper will only check its own input element and be independent of others.

Let me know if this approach correct, or do you have any other suggestions for handling this?

Aljullu:
Thanks for investigating this, @amitraj2203! Actually, in #58002 we might transform the `quantity` state into an object that matches product IDs with their quantities. So I think it might be a good idea to pause this PR, and update it once https://github.com/woocommerce/woocommerce/pull/58002 is merged into `trunk`. I think the work from that PR is exactly what is missing here, and I would like to avoid ending up implementing the same thing twice or differently. Do you think that's the right approach?

amitraj2203:
Sure thing @Aljullu, I'll wait for it to be merged into `trunk` and then continue with this.

Aljullu:
@amitraj2203 heads-up that https://github.com/woocommerce/woocommerce/pull/58002 has been merged. So you should be able to rebase or merge `trunk` into this branch. :slightly_smiling_face:

amitraj2203:
Hi @Aljullu,

Sorry for the delay and thank you for your patience. I was a bit tied up with other tasks recently, but I’ve now addressed the [issue](https://github.com/woocommerce/woocommerce/pull/58307#pullrequestreview-2887579946) and fixed the bug. Let me know if there’s anything else you'd like me to update or adjust.

Thanks!

https://github.com/user-attachments/assets/cb3a4a4b-95b9-4619-8328-de68e365e265

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
```diff
@@ -165,6 +166,70 @@ const addToCartWithOptionsStore = store(
...
+				) as HTMLInputElement | null;
+
+				if ( ! inputElement ) {
+					return false;
+				}
+
+				const parsedMinValue = parseInt( inputElement.min, 10 );
+				const parsedStep = parseInt( inputElement.step, 10 );
+
+				const minValue = isNaN( parsedMinValue ) ? 1 : parsedMinValue;
```
Aljullu: In grouped products children, we need to allow setting a quantity of `0`. It's being fixed in https://github.com/woocommerce/woocommerce/pull/58997, but in order to avoid conflicts between the two PRs, I would set the default `minValue` to `0` when `childProductId` is defined.

Aljullu (commented):
Thanks for updating this PR again, @amitraj2203! :raised_hands: 

This is mostly working fine, I noticed however the same issue as before in the existing _Add to Cart with Options_ block (the non-blockified one): when I increase or decrease the quantity of a grouped product child, all the quantity selector buttons update according to the last modified one:

https://github.com/user-attachments/assets/eb7868ca-c018-4f06-b10f-06018814d5ca

:point_up_2: notice that after setting _T-Shirt_ to `0`, the button to decrease the quantity of _Hoodie with Logo_ is disabled.

Can you reproduce as well? Btw, I'm fine if we fix that in this PR or if we split this improvement in two separate PRs, one for each block, as you prefer. :slightly_smiling_face:

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
amitraj2203: Thanks I've done it that way now.

amitraj2203: @Aljullu I've included the fix for the _Add to Cart with Options block (the non-blockified version)_ in this PR only. Additionally, there was a bug where changing the quantity through the input field didn't activate the decrease stepper button — I've included a fix for that too.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
```diff
@@ -169,6 +170,74 @@ const addToCartWithOptionsStore = store(
...
+					productType === 'grouped' && childProductId
+						? quantity[ childProductId ] || 0
+						: quantity[ context.productId ] || 0;
+
+				const selector =
+					productType === 'grouped' && childProductId
+						? `.wc-block-components-quantity-selector__input[name="quantity[${ childProductId }]"]`
+						: `.wc-block-components-quantity-selector__input[name="quantity"]`;
+
+				const inputElement = document.querySelector(
```
Aljullu: Using `document.querySelector()` here might be problematic because there might be the same product displayed as a top-level product and as a grouped product child. Ie:

```
Single Product: Logo collection
--> Beanie
--> T-Shirt
Single Product: Beanie
```

The first `Beanie` will have a min value of `0`, while the second one will have a min value of `1`. So `document.querySelector( ... ).min` will return 1 or 0 depending on the order of the blocks.

This will require a significant refactor, so we could handle that in a separate PR, but I think we should move away from getting the `min`, `step` and `max` values from the HTML input and instead pass it as the Interactivity API context from the backend. What do you think?

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
```diff
@@ -169,6 +170,74 @@ const addToCartWithOptionsStore = store(
...
+				const parsedStep = parseInt( inputElement.step, 10 );
+
+				const maxValue = isNaN( parsedMaxValue )
+					? undefined
+					: parsedMaxValue;
+				const step = isNaN( parsedStep ) ? 1 : parsedStep;
+
+				return (
+					maxValue === undefined || currentQuantity + step <= maxValue
+				);
```
Aljullu: Small nit, but I think using an early return when `parsedMaxValue` is NaN would make it easier to read:

```suggestion
				const parsedMaxValue = parseInt( inputElement.max, 10 );

				if ( isNaN( parsedMaxValue ) ) {
					return true;
				}
				
				const parsedStep = parseInt( inputElement.step, 10 );
				const step = isNaN( parsedStep ) ? 1 : parsedStep;

				return currentQuantity + step <= maxValue;
```

This is quite subjective, so I'm fine leaving it as-is if you prefer. :slightly_smiling_face: In case you end up changing this, it can be changed in the `add-to-cart-form/frontend.ts` file as well.

plugins/woocommerce/client/blocks/assets/js/blocks/product-elements/add-to-cart-form/frontend.ts
```diff
@@ -32,34 +39,118 @@ const getInputData = ( event: HTMLElementEvent< HTMLButtonElement > ) => {
...
-			const { currentValue, maxValue, step, inputElement } = inputData;
+			const context = getContext< Context >();
+			const {
+				currentValue,
+				maxValue,
+				step,
+				inputElement,
+				childProductId,
+			} = inputData;
 			const newValue = currentValue + step;
```
Aljullu: I wonder if we should save a variable like this at the beginning of the functions:

```TS
			const productId = childProductId || context.productId;
			if ( ! productId ) {
				return false;
			} 
```

If you do so, you will be able to avoid the `if { ... } else { ... }` and won't have to add `as number` every time you use `context.productId`.

Aljullu (commented):
Thanks for all the updates, @amitraj2203! This looks promising, but I left a few small comments inline. I'm a bit concerned about the usage of `document.querySelector()` to get the input values, as it might return any `input` on the page: from other blocks, classic templates, etc. And I feel that won't be completely reliable.

I know this PR is growing quite a lot, so I'm fine if you prefer splitting it. I think at least these two items:

> * Reordered the tab index in QuantityStepper so that:
> * Focus moves first to the input field, then the minus (-) button, and finally the plus (+) button - matching the QuantitySelector component behavior.

Are stable now, so they could be merged separately from the other one, which seems to be the most complex one:

> * Applied correct disabled, unfocusable, and unclickable behavior to the minus (-) button when quantity equals the minimum allowed value.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
amitraj2203: @Aljullu I've reverted the change that _disables the minus (-) button when the quantity equals the minimum allowed value_.

This PR now includes only the following:

> - Reordered the tab index in `QuantityStepper` so that:
>  - Focus moves first to the input field, then to the minus (-) button, and finally to the plus (+) button — matching the behavior of the `QuantitySelector` component.

---

I’ll raise a new PR to address the issue of **disabling the minus (-) button when the quantity equals the minimum allowed value**, and I’ll incorporate the three comments you provided in this PR as part of that work.

Aljullu (approved):
Thanks for splitting this in two separate PRs, this way we can merge these improvements already. :slightly_smiling_face: 

I tested, and the focus order is as described now.

https://github.com/user-attachments/assets/358cb570-ca6b-4854-84cb-0f2b856e4294

<details>

* Theme: Woo Starter Theme
* Store creation: June 2025
* Store size: small
</details>

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartForm.php
```diff
@@ -75,14 +75,14 @@ protected function enqueue_data( array $attributes = [] ) {
 	private function add_steppers( $product_html, $product_name ) {
 		// Regex pattern to match the <input> element with id starting with 'quantity_'.
 		$pattern = '/(<input[^>]*id="quantity_[^"]*"[^>]*\/>)/';
-		// Replacement string to add button BEFORE the matched <input> element.
+		// Replacement string to add button AFTER the matched <input> element.
 		/* translators: %s refers to the item name in the cart. */
-		$minus_button = '<button aria-label="' . esc_attr( sprintf( __( 'Reduce quantity of %s', 'woocommerce' ), $product_name ) ) . '"type="button" data-wp-on--click="actions.removeQuantity" class="wc-block-components-quantity-selector__button wc-block-components-quantity-selector__button--minus">-</button>$1';
+		$minus_button = '$1<button aria-label="' . esc_attr( sprintf( __( 'Reduce quantity of %s', 'woocommerce' ), $product_name ) ) . '"type="button" data-wp-on--click="actions.removeQuantity" class="wc-block-components-quantity-selector__button wc-block-components-quantity-selector__button--minus">-</button>';
```
Aljullu: Sorry, I missed that. There is a space missing here:

```suggestion
		$minus_button = '$1<button aria-label="' . esc_attr( sprintf( __( 'Reduce quantity of %s', 'woocommerce' ), $product_name ) ) . '" type="button" data-wp-on--click="actions.removeQuantity" class="wc-block-components-quantity-selector__button wc-block-components-quantity-selector__button--minus">-</button>';
```

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/Utils.php
```diff
@@ -21,14 +21,14 @@ class Utils {
 	public static function add_quantity_steppers( $quantity_html, $product_name ) {
 		// Regex pattern to match the <input> element with id starting with 'quantity_'.
 		$pattern = '/(<input[^>]*id="quantity_[^"]*"[^>]*\/>)/';
-		// Replacement string to add button BEFORE the matched <input> element.
+		// Replacement string to add button AFTER the matched <input> element.
 		/* translators: %s refers to the item name in the cart. */
-		$minus_button = '<button aria-label="' . esc_attr( sprintf( __( 'Reduce quantity of %s', 'woocommerce' ), $product_name ) ) . '" type="button" data-wp-on--click="actions.decreaseQuantity" class="wc-block-components-quantity-selector__button wc-block-components-quantity-selector__button--minus">-</button>$1';
+		$minus_button = '$1<button aria-label="' . esc_attr( sprintf( __( 'Reduce quantity of %s', 'woocommerce' ), $product_name ) ) . '"type="button" data-wp-on--click="actions.decreaseQuantity" class="wc-block-components-quantity-selector__button wc-block-components-quantity-selector__button--minus">-</button>';
```
Aljullu: And here:

```suggestion
		$minus_button = '$1<button aria-label="' . esc_attr( sprintf( __( 'Reduce quantity of %s', 'woocommerce' ), $product_name ) ) . '" type="button" data-wp-on--click="actions.decreaseQuantity" class="wc-block-components-quantity-selector__button wc-block-components-quantity-selector__button--minus">-</button>';
```

Aljullu (changes requested):

plugins/woocommerce/src/Blocks/BlockTypes/AddToCartForm.php
amitraj2203: Fixed the missing space. Also I just changed the description of this PR to `Part Of`. Hoping it shouldn't closes the issue 😅

Aljullu (approved):
Thanks for the updates and the contribution, @amitraj2203! LGTM! :raised_hands:

PR #58309, opened 2025-05-27
Try: Make Product Details hooked block testable

dinhtungdu:


dinhtungdu:
@ockham  I have to add a new property to make the filters testable. What do you think about this approach?

ockham:
> @ockham I have to add a new property to make the filters testable. What do you think about this approach?

Hmm, I think it'd be nice if we could avoid introducing this kind of statefulness (`$current_block`) just for the sake of testing 🤔 Plus the division into `register_hooked_block_type` and `register_hooked_block_data` is also a bit clunky 😬 (although I see where it's coming from).

A few thoughts on breaking down this problem:
- The public interface we'd need to cover with tests is the `woocommerce_product_details_hooked_blocks` filter, right?
- I don't think we necessarily need to test whether that injects the hooked block(s) into a given piece of block markup when running `apply_block_hooks_to_content`. Instead, it should be sufficient to check whether the `woocommerce_product_details_hooked_blocks` results in `hooked_block_types` and `hooked_block_...` filter invocations that add the expected hooked blocks correctly. (IMO, that's also the more accurate way of adding a _unit_ test for `woocommerce_product_details_hooked_blocks`.)

It should be possible to add a mock filter to `hooked_block_types` and `hooked_block_....`, respectively, to assert that the hooked block is correctly passed to those filters. There are some examples in Core on how to use `MockAction` to do that, e.g. [here](https://github.com/WordPress/wordpress-develop/blob/666ebd8731818af64f135a28aaffb90881b7ed8c/tests/phpunit/tests/blocks/updateIgnoredHookedBlocksPostMeta.php#L201-L221).

There's still a problem with method visibility we need to solve (since `initialize` is protected). We can probably either derive a dummy subclass or use [introspection](https://www.php.net/manual/en/class.reflectionfunction.php) to call it, or maybe introduce an intermediary public method that we can call from tests.

LMK if all this makes sense!

dinhtungdu:
> Instead, it should be sufficient to check whether the woocommerce_product_details_hooked_blocks results in hooked_block_types and hooked_block_... filter invocations that add the expected hooked blocks correctly. (IMO, that's also the more accurate way of adding a unit test for woocommerce_product_details_hooked_blocks.)

This is where I hit the wall tbh. I can't find a way to add the filter from the test before we load WooCommerce. At the time we add the filter, the apply filter is already executed.

> There's still a problem with method visibility we need to solve (since initialize is protected). We can probably either derive a dummy subclass or use [introspection](https://www.php.net/manual/en/class.reflectionfunction.php) to call it, or maybe introduce an intermediary public method that we can call from tests.

This is the first thing I tried, I added a public method that calls the `initialize()` methods. Then I used `BlockTemplateUtils` to prepare the template. It works, but the problem is that I can't remove the two filters added by `register_hooked_block()` method.

> just for the sake of testing

Yeah, that's my feeling too. I will work on merging the original PR, this one is not a blocker for it, but let's keep this open and continue the convo.

ockham:
> > Instead, it should be sufficient to check whether the woocommerce_product_details_hooked_blocks results in hooked_block_types and hooked_block_... filter invocations that add the expected hooked blocks correctly. (IMO, that's also the more accurate way of adding a unit test for woocommerce_product_details_hooked_blocks.)
> 
> This is where I hit the wall tbh. I can't find a way to add the filter from the test before we load WooCommerce. At the time we add the filter, the apply filter is already executed.

Gotcha 🤔 I guess initialization of all those blocks happens very early on.

> > There's still a problem with method visibility we need to solve (since initialize is protected). We can probably either derive a dummy subclass or use [introspection](https://www.php.net/manual/en/class.reflectionfunction.php) to call it, or maybe introduce an intermediary public method that we can call from tests.
> 
> This is the first thing I tried, I added a public method that calls the `initialize()` methods. Then I used `BlockTemplateUtils` to prepare the template. It works, but the problem is that I can't remove the two filters added by `register_hooked_block()` method.

Right. I worked bit on a tentative unit test today and came across WP's [`remove_all_filters` method](https://developer.wordpress.org/reference/functions/remove_all_filters/) (which was also new to me).

I've filed a PR: https://github.com/woocommerce/woocommerce/pull/58363
(Since I wasn't able to get unit tests to run locally due to an issue with wp-env, I'll have to wait for CI to tell me if it's any good 😅 )

dinhtungdu:
@ockham I updated this PR with my original direction that doesn't require updating the block class. I see you also created one, so pushing it to use as reference.

ockham:
> @ockham I updated this PR with my original direction that doesn't require updating the block class.

Nice, I like that! It has a much smaller footprint now 👍 

FWIW, I also tweaked [mine](https://github.com/woocommerce/woocommerce/pull/58363) a bit. My philosophy there is to avoid running `apply_block_hooks_to_content` at all, as IMO it's not part of the "unit" we're writing a test for -- which is the `woocommerce_product_details_hooked_blocks` hook. (I've added a paragraph about this to the PR description). But ultimately, I'm fine with either approach -- I'll leave the decision with you! 😊

dinhtungdu:
Closing in favor of https://github.com/woocommerce/woocommerce/pull/58363

PR #58310, opened 2025-05-27
[NOX in LYS] Generate a different Jetpack connection link for LYS

### LINKED ISSUE #58281: [NOX in LYS] Logic: Backend for NOX should generate a different Jetpack connection link for LYS
In the "Connect with WP.com" step, the current link redirects to the Settings page after completion.

When the `GET /onboarding` request is completed from LYS, we should return a Jetpack link that redirects back to LYS, allowing us to continue the process.

This can be achieved by adding a query string to the request or controller, but other options may also be considered.

elazzabi:

When using NOX, we assumed that the only way to access the flow was through the `Settings > Payments` screen. The return URL we had is for that screen.

For LYS, we want to redirect to the LYS flow to continue the onboarding.

Closes WOOPLUG-4407
Closes #58281 

1. Create a new Jurassic Ninja site with WooCommerce Beta on this branch
2. Once ready, skip the onboarding and choose a country with WooPayments (like UK)
3. Go to `Settings > Payments` and click Install on WooPayments to get the latest version. You can skip this step if you have already selected it in JN.
4. **Make sure to override the payment task status**. Go to `Tools > Plugin File Editor`, select WooCommerce, and open this file `src/Admin/Features/OnboardingTasks/Tasks/Payments.php`. Return `false` from `is_complete()` and return `''` from `get_action_url()`.
5. Go to `WooCommerce > Home` and click `Launch your store`
6. Click `Get paid`
7. Click `Continue` on the first step, and you'll reach the WP.com connection step
8. Click on the connect button and go to Jetpack
9. After finishing the login, make sure you're redirected back to the LYS flow instead of the Settings page

Here's a video of the process:

https://github.com/user-attachments/assets/b4bed45e-2f13-4f44-bea0-527d877abf6c

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

This PR will not be merged into trunk. No changelog required

</details>

### DIFF
--- a/plugins/woocommerce/src/Internal/Admin/Settings/PaymentProviders/WooPayments/WooPaymentsService.php
+++ b/plugins/woocommerce/src/Internal/Admin/Settings/PaymentProviders/WooPayments/WooPaymentsService.php
@@ -1443,15 +1443,27 @@ class WooPaymentsService {
 		// If the WPCOM connection is already set up, we don't need to add anything more.
 		if ( self::ONBOARDING_STEP_STATUS_COMPLETED !== $wpcom_step['status'] ) {
 			// Craft the return URL.
-			// By default, we return the user to the onboarding modal.
-			$return_url = $this->proxy->call_static(
-				Utils::class,
-				'wc_payments_settings_url',
-				self::ONBOARDING_PATH_BASE,
-				array(
-					'wpcom_connection_return' => '1', // URL query flag so we can properly identify when the user returns.
-				)
-			);
+			switch($source) {
+				case 'launch-your-store':
+					// If the source is 'launch-your-store', we return the user to the Launch Your Store flow.
+					$return_url = $this->proxy->call_function(
+						'admin_url',
+						'admin.php?page=wc-admin&path=/launch-your-store' . self::ONBOARDING_PATH_BASE . '&sidebar=hub&content=payments&wpcom_connection_return=1'
+					);
+					break;
+				default:
+					// By default, we return the user to the onboarding modal in the Settings > Payments page.
+					$return_url = $this->proxy->call_static(
+						Utils::class,
+						'wc_payments_settings_url',
+						self::ONBOARDING_PATH_BASE,
+						array(
+							'wpcom_connection_return' => '1', // URL query flag so we can properly identify when the user returns.
+						)
+					);
+					break;
+			}
+
 			// Try to generate the authorization URL.
 			$wpcom_connection = $this->get_wpcom_connection_authorization( $return_url );
 			if ( ! $wpcom_connection['success'] ) {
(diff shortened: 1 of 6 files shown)


plugins/woocommerce/src/Internal/Admin/Settings/PaymentProviders/WooPayments/WooPaymentsService.php
```diff
@@ -1441,15 +1443,27 @@ private function get_onboarding_steps( string $location, string $rest_path ): ar
...
-				'wc_payments_settings_url',
-				self::ONBOARDING_PATH_BASE,
-				array(
-					'wpcom_connection_return' => '1', // URL query flag so we can properly identify when the user returns.
-				)
-			);
+			switch ( $source ) {
+				case 'launch-your-store':
+					// If the source is 'launch-your-store', we return the user to the Launch Your Store flow.
+					$return_url = $this->proxy->call_function(
```
elazzabi, line 1449: @vladolaru, out of curiosity, what's the difference between defining URLs like this versus what we already had? Is it safe to define it this way?

vladolaru: The `LegacyProxy::call_function()` method just proxies the user function call. It is the same as calling:

```
admin_url( 'admin.php?page=wc-admin&path=/launch-your-store' . self::ONBOARDING_PATH_BASE . '&sidebar=hub&content=payments&wpcom_connection_return=1' );
```

The benefit is that we can properly unit test it since we can easily mock any such calls like this:

```
$this->mockable_proxy = wc_get_container()->get( LegacyProxy::class );
$this->mockable_proxy->register_function_mocks(
			array(
				// Mock the $this->current_time.
				'time'         => function () {
					return $this->current_time;
				},
				// Everything is callable by default.
				'is_callable'  => function () {
					return true;
				},
				'class_exists' => function ( $class_to_check ) {
					// By default, the WooPayments extension is mocked as active.
					if ( '\WC_Payments' === $class_to_check ) {
						return true;
					}

					return false;
				},
                'admin_url' => function ( $relative_path ) {
                    return 'https://example.com/wp-admin/something';
                },
			),
		);
```

Not every user function call needs to be wrapped. Probably `admin_url` doesn't need to be because it's a quite pure function (barring the filtering).

mordeth (approved):
Thank you for working on this, Ahmed!

✅ Code changes are looking good and make sense.
✅ The testing steps pass successfully, and the user is properly redirected back to LYS after connecting Jetpack.

PR #58320, opened 2025-05-27
Default FlexSlider jQuery easing now correctly set to "swing"

apenchev:

The default FlexSlider easing when not using CSS3 3D Transforms `(useCSS == false || video == true)` is now correctly set to jQuery "swing".

Also a small refactor to make it more obvious that:

- we use jQuery easings only when not using CSS3 3D Transforms `(useCSS == false || video == true)`
- we use the CSS "ease" easing when using CSS3 3D Transforms `(useCSS == true && video == false)`

Updates #36689.

1. Using the Twenty Twenty-Five, navigate to a single product page with multiple images on desktop.
2. Switch between different product images either using the arrow keys or by clicking on different image thumbnails.
3. Confirm that there is an easing transition when switching between product images - it should behave exactly the same as before this PR.

Tested locally using the Twenty Twenty-Five theme on a single product page with Chrome and Firefox on Windows.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
The default FlexSlider easing when not using CSS3 3D Transforms is now correctly set to jQuery "swing".
</details>


plugins/woocommerce/client/legacy/js/flexslider/jquery.flexslider.js
```diff
@@ -106,7 +106,7 @@
         slider.started = false;
         slider.startTimeout = null;
         // TOUCH/USECSS:
-        slider.transforms = slider.transitions = !slider.vars.video && !fade && slider.vars.useCSS;
+        slider.transforms = slider.transitions = !slider.vars.video && !fade && slider.vars.useCSS; // TODO: Delete "slider.transitions", this variable name is only kept for backwards compatibility
```
prettyboymp: This TODO comment appears unrelated to the easing refactor. Could you remove it from this PR and address it separately?

prettyboymp (changes requested):
@apenchev,

Thank you for this cleanup! The changes on lines 72 and 817 work together to preserve the original functionality in the common case.

However, when an invalid/unrecognized easing name is passed, the fallback behavior changes from CSS `"ease"` to jQuery `"swing"`. Since FlexSlider is a jQuery plugin and defaults to `"swing"`, this is arguably more consistent — but it's technically a behavior change rather than a pure refactor.  Because of that, we should make sure that the changelog file reflects that change.

apenchev:
@prettyboymp You are correct, thanks for checking! I've updated the changelog and PR description to clarify that.

prettyboymp (changes requested):
Took another look — my earlier review missed something on the line 817 restructure. Walking it back:

I said the changes "preserve the original functionality in the common case." That's not right. On line 109, `slider.transforms = !slider.vars.video && !fade && slider.vars.useCSS`. For the default product gallery (`animation: "slide"`, `useCSS: true`) on any modern browser, `slider.transforms` is `true`. Before this PR, the JS never wrote `transition-timing-function` on that path — themes controlled it via CSS. After the PR, `transition-timing-function: ease` is written inline on every slide change, and inline styles win over stylesheet rules without `!important`. So any theme that customizes this property with normal CSS silently stops working.

The PR description scopes the change to "when not using CSS3 3D Transforms," but the line 817 restructure also affects the transforms path. The line 72 change alone matches the described scope — the `easing` variable only feeds into non-transforms paths.

Two options:

**Close the PR.** Once line 817 reverts, what's left (line 72) only fires for callers passing an easing name that isn't in the `easings` map. Core doesn't hit that path (`single-product.js` doesn't set the option, default `"swing"` resolves directly), so the reach is third-party callers with typos or unsupported names. Probably not worth the overhead.

**Pare back to match the PR description.** Revert line 817 so the `else` block stays as it was, keep the line 72 fallback fix. If you want the readability the restructure was aiming for, a comment does it without the regression:

```js
if (slider.transforms) {
  // transition-timing-function intentionally left to CSS so themes can customize it.
  target = (vertical) ? "translate3d(0," + target + ",0)" : "translate3d(" + (parseInt(target)+'px') + ",0,0)";
} else {
  // jQuery-style easing names only apply here (no CSS3 transforms in use).
  slider.container.css("transition-timing-function", easing);
}
```

If you go that route, the changelog and description need a tweak too. "Correctly set to jQuery swing" reads like a bug fix, but `easing` is interpolated into CSS strings — never passed to jQuery's animation engine — and CSS `"ease"` was a valid easing function. Something like "Aligns the FlexSlider easing fallback for unrecognized names with the documented `swing` default" matches what's actually changing.

PR #58356, opened 2025-05-28
MIC: Add Customer Stock Notifications menus and Notifications List Tables

jimjasson:

This PR:
- adds a new submenu item under WooCommerce, called 'Notifications' which;
- contains a page called Stock Notifications, that prints a list table with all the stock notifications registered in the database. 

Markup on 2025-05-30 at 15:24:43

## Technical specification

File structure:
```
-- Automattic\WooCommerce\Internal\StockNotifications\Admin\AdminManager.php
  |
  -- Automattic\WooCommerce\Internal\StockNotifications\Admin\MenusController.php
     |
     -- Automattic\WooCommerce\Internal\StockNotifications\Admin\NotificationsPage.php
        |
        -- Automattic\WooCommerce\Internal\StockNotifications\Admin\NotificationsListTable.php
```

**AdminManager.php**

This is a Manager class that gets enqueued only in admin context and is responsible for:
- initializing the `SettingsController` and the `MenusController` and;
- enqueuing the admin scripts.

**MenusController.php**

This is a Controller responsible for:
- adding the Notifications submenu under WooCommerce,
- creating a Stock Notifications page, 
- adding screen options support for the Stock Notifications page so that users can change how many notifications show up per page, 
- registering the `woocommerce_page_customer_stock_notifications` screen ID for the Stock Notifications page, 
- triggering the display of the Notifications list table and; 
- linking the new page to WC Admin. 

This file also contains some scaffolding for the create new and edit actions for notifications, that will be implemented in a later PR.

**NotificationsPage.php**

This is a class that is responsible for:
- displaying the Notifications list table and printing success/error messages related to the list table actions and; 
- handling the deletion of notifications. 

**NotificationsListTable.php**

This is the class that constructs the list table with the following columns:
- ID, 
- Status (Pending, Cancelled, Sent, or Active),
- User, 
- Product, 
- SKU, 
- Signed Up date and;
- Waiting since. 

It also adds 5 views, one for All notifications, one for Active, one for Pending, one for Cancelled and one for Sent. Each view displays a count for how many notifications have that status, based on the active filters. 

Speaking of filters, this list table supports filtering by date (only dates with associated notifications are used), user and product. It is also possible to search notifications using a user email -- for performance reasons, an exact user e-mail need to be provided for a match. We can re-iterate on that after the release. 

Then, there are also 3 Bulk actions, one for activating a notification, one for cancelling it and one for deleting it. 

Finally, each notification has a edit (not yet functional) and delete button that appear on hover. If merchants click to delete a notification, either via this button or via bulk actions, a pop-up modal shows up where they need to confirm their action.

**html-admin-notifications.php**

This is where all the HTML code related to the list table lives. 

**StockNotificationsDataStore.php**

There has been some updates in the `query` method to support getting notifications that are created between specific dates and also ordering notifications.

Moreover, a new method was introduced that gets all the months for which there was a notification created.

Closes [WOOPRD-539: Create Notifications List Table](https://linear.app/a8c/issue/WOOPRD-539/create-notifications-list-table)
Closes [WOOPRD-538: Create a Stock Notifications WooCommerce submenu](https://linear.app/a8c/issue/WOOPRD-538/create-a-stock-notifications-woocommerce-submenu)

Start with a clean database, without any notifications
1. Go to WooCommerce > Notifications.
2. In the Stock Notifications page, confirm that you see a notice that says that you don't have any notifications yet and a link to the documentation.
3. Ensure that the blank state page uses similar styles to the Orders blank state page.

Prepare the database with notifications that have different statuses and different creation dates. 

1. Go to WooCommerce > Notifications.

Sanity check on the table:
- Ensure that all columns show up and are populated correctly. 
- Ensure that notifications are paginated and that pagination works.
- Ensure that the Signed Up column is sortable. 

Test screen options:
- Ensure that it is possible to change how many notifications show up per page. 

Test views:
- Ensure that the All, Active, Sent, Cancelled and Pending views show up with the correct count. 
- Ensure that you can click on each view and get only notifications associated with that view. 

Test filters:
- Ensure that you can filter notifications by user, product and date.
- Ensure that the views count is updated when filtering. 

Test search:
- Ensure that you can search notifications by providing the full user e-mail. 

Test bulk actions:
- Ensure that you can activate, cancel and delete notifications in bulk. 
- When deleting ensure that you see a pop-up modal that asks you to confirm your action. 
- After every bulk action, ensure that you see a success message. 

Test deleting individual notifications:
- Hover over individual notifications and click on Delete. 
- Ensure that you see a pop-up modal that asks you to confirm your action. 
- In another notification, right click on the Delete button and copy the permalink. 
- Change the notification ID with an ID that doesn't exist. 
- Visit that page and confirm you get an error notice. 

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 
This is part of a bigger feature, a single changelog entry will be added at the end.
</details>


plugins/woocommerce/src/Internal/StockNotifications/Admin/MenusController.php
```diff
@@ -0,0 +1,196 @@
...
+	 * Constructor.
+	 */
+	public function __construct() {
+
+		// Add Stock Notifications menu item.
+		add_action( 'admin_menu', array( $this, 'add_menu' ), 10 );
+
+		// Integrate WooCommerce breadcrumb bar.
+		add_action( 'admin_menu', array( $this, 'wc_admin_connect_customer_stock_notifications_pages' ) );
+		add_filter( 'woocommerce_navigation_pages_with_tabs', array( $this, 'wc_admin_navigation_pages_with_tabs' ) );
```
jimjasson: @xristos3490 I don't think that this hook is needed, but if I remove it, the white WC Admin header is removed from the page.

plugins/woocommerce/client/legacy/css/admin.scss
```diff
@@ -8276,6 +8277,7 @@ table.bar_chart {
 	}
 }
 
+.woocommerce_page_customer_stock_notifications .tablenav,
```
xristos3490: Nit. Maybe we can harmonize _ and - in the names?

plugins/woocommerce/client/legacy/js/admin/wc-customer-stock-notifications.js
```diff
@@ -0,0 +1,35 @@
+/* global wc_admin_customer_stock_notifications_params, woocommerce_admin_meta_boxes */
+;( function( $, window, document ) {
```
xristos3490: The `document` lookup isn't needed?

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -455,12 +459,35 @@ public function query( array $args ) {
...
+		$order_by         = '';
+		$order_by_clauses = array();
+
+		if ( $args['order_by'] && is_array( $args['order_by'] ) ) {
+			foreach ( $args['order_by'] as $what => $how ) {
+				$order_by_clauses[] = $table . '.' . esc_sql( strval( $what ) ) . ' ' . esc_sql( strval( $how ) );
+			}
+		}
+
+		$order_by_clauses = empty( $order_by_clauses ) ? array( $table . '.id, ASC' ) : $order_by_clauses;
```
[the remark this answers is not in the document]
xristos3490: This needs fixing as described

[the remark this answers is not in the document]
xristos3490: Probably a bug in the extension.

plugins/woocommerce/src/Internal/StockNotifications/Admin/AdminManager.php
```diff
@@ -0,0 +1,60 @@
...
+	 * @return void
+	 */
+	public static function admin_resources() {
+
+		$screen    = get_current_screen();
+		$screen_id = $screen ? $screen->id : '';
+		$suffix    = Constants::is_true( 'SCRIPT_DEBUG' ) ? '' : '.min';
+		$version   = Constants::get_constant( 'WC_VERSION' );
+
+		wp_register_script( 'wc-admin-customer-stock-notifications', WC()->plugin_url() . '/assets/js/admin/wc-customer-stock-notifications' . $suffix . '.js', array( 'jquery', 'jquery-ui-datepicker', 'wp-util', 'sw-admin-select-init', 'wc-backbone-modal' ), $version, true );
```
xristos3490: I believe we don't need all these deps, no?

plugins/woocommerce/src/Internal/StockNotifications/Admin/AdminManager.php
```diff
@@ -0,0 +1,60 @@
...
+
+		$params = array(
+			'i18n_wc_delete_notification_warning'       => __( 'Delete this notification permanently?', 'woocommerce' ),
+			'i18n_wc_bulk_delete_notifications_warning' => __( 'Delete the selected notifications permanently?', 'woocommerce' ),
+		);
+
+		/*
+		 * Enqueue specific styles & scripts.
+		 */
+		if ( 'woocommerce_page_customer_stock_notifications' === $screen_id ) {
```
xristos3490: Nit. Let's keep it clean by bailing with `return;` instead of wrapping code in `if`

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -455,12 +459,35 @@ public function query( array $args ) {
 			$where_values[] = esc_sql( $args['user_email'] );
 		}
 
+		if ( $args['start_date'] ) {
+			$where[]        = 'date_created_gmt >= %s';
+			$where_values[] = $args['start_date'];
```
xristos3490: Run `esc_sql` for best practices here.

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -455,12 +459,35 @@ public function query( array $args ) {
...
+		}
+
+		if ( $args['end_date'] ) {
+			$where[]        = 'date_created_gmt < %s';
+			$where_values[] = $args['end_date'];
```
xristos3490: Same

plugins/woocommerce/src/Internal/StockNotifications/Admin/AdminManager.php
```diff
@@ -0,0 +1,60 @@
...
+class AdminManager {
+
+	/**
+	 * Initialize admin components.
+	 *
+	 * @internal
+	 *
+	 * @return void
+	 */
+	final public function init() {
```
xristos3490: Let's make that a `__construct` there isn't really a reason to go with the `init` if there is no DI involved.

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php
```diff
@@ -0,0 +1,799 @@
...
+
+	/**
+	 * Handles the product column output.
+	 *
+	 * @param Notification $notification The notification object.
+	 * @return void
+	 */
+	public function column_product( $notification ) {
+		$product = $notification->get_product();
+		if ( is_a( $product, 'WC_Product' ) ) {
```
xristos3490: Same for bailing early, let's avoid long `ifs`

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php
```diff
@@ -0,0 +1,799 @@
...
+			if ( $formatted_variation_list ) {
+				/* translators: product name, identifier */
+				$name .= '<span class="description">' . $formatted_variation_list . '</span>';
+			}
+
+			echo wp_kses_post(
+				sprintf(
+					'<a target="_blank" href="' . admin_url( 'post.php?post=%d&action=edit' ) . '">%s</a>',
+					$product->get_parent_id() ? absint( $product->get_parent_id() ) : absint( $product->get_id() ),
+					wp_kses_post( $name )
```
xristos3490: Is there a reason to run the `wp_kses_post()` twice here? Maybe simplify?

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php
```diff
@@ -0,0 +1,799 @@
...
+	 * Handles the notification date column output.
+	 *
+	 * @param Notification $notification The notification object.
+	 * @return void
+	 */
+	public function column_date_created_gmt( $notification ) {
+		$date_created = $notification->get_date_created()->getTimestamp();
+
+		if ( ! $date_created ) {
+			$t_time = __( 'Unpublished', 'woocommerce' );
```
xristos3490: Let's just print an `mdash` here, considering this is a rare case and the "unpublished" doesn't fit here.

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php
```diff
@@ -0,0 +1,799 @@
...
+	/**
+	 * Display dates dropdown filter.
+	 *
+	 * @return void
+	 */
+	protected function display_months_dropdown() {
+		global $wp_locale;
+
+		$months      = $this->data_store->get_distinct_dates();
+		$month_count = count( $months );
```
xristos3490, line 798: It's not guaranteed that the `wpdb->get_results()` will return an array here.

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsPage.php
```diff
@@ -0,0 +1,81 @@
...
+	 */
+	public static function delete() {
+
+		check_admin_referer( 'delete_customer_stock_notification' );
+
+		$notification_id = isset( $_GET['notification'] ) ? absint( $_GET['notification'] ) : 0;
+
+		try {
+			if ( ! $notification_id ) {
+				throw new \Exception( 'Missing notification ID.' );
```
xristos3490: Nit. -> "Notification not found."

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsPage.php
```diff
@@ -0,0 +1,81 @@
...
+			}
+
+			$notification = new Notification( $notification_id ); // <- this can throw
+			\WC_Data_Store::load( 'stock_notification' )->delete( $notification );
+
+			wp_safe_redirect( add_query_arg( 'notice', 'deleted', admin_url( self::PAGE_URL ) ) );
+			exit;
+
+		} catch ( \Exception $e ) {
+			wp_safe_redirect( add_query_arg( 'notice', 'not_found', admin_url( self::PAGE_URL ) ) );
```
xristos3490: I'm not entirely sure I dig this approach. Why not use a session-based flash notice via `WC_Admin_Notices` and get rid of the handling in `output`?

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php
```diff
@@ -0,0 +1,799 @@
...
+
+		if ( ! $month_count || ( 1 === $month_count && 0 === (int) $months[0]->month ) ) {
```
xristos3490: Similarly, the `count` would never return anything other than an int. We don't need this checking here, maybe just `count > 1`?

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php
```diff
@@ -0,0 +1,799 @@
...
+	 * Handles the status column output.
+	 *
+	 * @param Notification $notification The notification object.
+	 * @return void
+	 */
+	public function column_status( $notification ) {
+
+		if ( $notification->get_status() === NotificationStatus::PENDING ) {
+			$status = 'cancelled';
+			$label  = __( 'Pending', 'woocommerce' );
```
xristos3490: I feel this would need to be centralized, for now and for this PR, let's add a `_x()` function since these can be tricky to translate without a context.

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php
```diff
@@ -0,0 +1,799 @@
...
+	public $data_store;
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		global $status, $page;
+
+		$this->data_store              = \WC_Data_Store::load( 'stock_notification' );
+		$this->total_items             = $this->data_store->query( array( 'return' => 'count' ) );
```
xristos3490: These queries should happen in the prepare_items, no need to store them in props.

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php
```diff
@@ -0,0 +1,799 @@
...
+			)
+		);
+
+		// Count sent notifications.
+		$this->total_sent_items = $this->data_store->query(
+			array(
+				'return' => 'count',
+				'status' => NotificationStatus::SENT,
+			)
+		);
```
xristos3490: Similarly, there is no need to have these run on the constructor or have them under class props. Check how the get_views() method works on the WP_List_Table base class.

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php
```diff
@@ -0,0 +1,799 @@
...
+
+		parent::__construct(
```
xristos3490, line 80: This should be the only thing living in the constructor.

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php
```diff
@@ -0,0 +1,799 @@
...
+ * Notifications list table for Customer Stock Notifications.
+ */
+class NotificationsListTable extends \WP_List_Table {
+
+	/**
+	 * Page URL.
+	 *
+	 * @const PAGE_URL
+	 */
+	const PAGE_URL = 'admin.php?page=customer_stock_notifications';
```
xristos3490: 👌

plugins/woocommerce/src/Internal/StockNotifications/Admin/MenusController.php
```diff
@@ -0,0 +1,196 @@
...
+			return (int) $value;
+		}
+		return $status;
+	}
+
+	/**
+	 * Displays the Notifications list table.
+	 */
+	public function notifications_page() {
+		wc_get_container()->get( NotificationsPage::class );
```
xristos3490: This seems to be a separation of concerns issue. Looks like the actions (+ request checks and parsing) can and should happen within the NotificationPage itself, no?

xristos3490 (commented):
Great job rewriting this entire admin flow, @jimjasson! I left some comments as a first round of review.

plugins/woocommerce/client/legacy/css/admin.scss
jimjasson: Done: https://github.com/woocommerce/woocommerce/pull/58356/commits/4aff14edda58f81950f4c459d6121d2090ea6a63

plugins/woocommerce/client/legacy/js/admin/wc-customer-stock-notifications.js
jimjasson: Done: https://github.com/woocommerce/woocommerce/pull/58356/commits/acaad5dea59274db1b5af0f865cb3f58745aa9c5

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
jimjasson: Fixed: https://github.com/woocommerce/woocommerce/pull/58356/commits/9d75653274d0db1ce76979c0ced4bd506e6acf62

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
jimjasson: Fixed: https://github.com/woocommerce/woocommerce/pull/58356/commits/9d75653274d0db1ce76979c0ced4bd506e6acf62

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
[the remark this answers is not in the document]
jimjasson: Fixed: https://github.com/woocommerce/woocommerce/pull/58356/commits/9d75653274d0db1ce76979c0ced4bd506e6acf62

plugins/woocommerce/src/Internal/StockNotifications/Admin/AdminManager.php
jimjasson: You are right, fixed: https://github.com/woocommerce/woocommerce/pull/58356/commits/b34998a1232a898e0b5931d083b002c650dffa17

plugins/woocommerce/src/Internal/StockNotifications/Admin/AdminManager.php
jimjasson: Done: https://github.com/woocommerce/woocommerce/pull/58356/commits/b34998a1232a898e0b5931d083b002c650dffa17

plugins/woocommerce/src/Internal/StockNotifications/Admin/AdminManager.php
jimjasson: Done: https://github.com/woocommerce/woocommerce/pull/58356/commits/b34998a1232a898e0b5931d083b002c650dffa17

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php
jimjasson: Done: https://github.com/woocommerce/woocommerce/pull/58356/commits/d0c0bdf3752da2daf293e9c85ab5976d6a35b35f

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php
jimjasson: Looks like a carry over from the extension -- fixed: https://github.com/woocommerce/woocommerce/pull/58356/commits/d0c0bdf3752da2daf293e9c85ab5976d6a35b35f

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php
jimjasson: Done: https://github.com/woocommerce/woocommerce/pull/58356/commits/d0c0bdf3752da2daf293e9c85ab5976d6a35b35f

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsPage.php
jimjasson: @xristos3490 I explored the option of using `WC_Admin_Notices` and I noticed that if I use `add_notice`, I would need to register the notice in the `WC_Admin_Notices` class. The only other option is to use `add_custom_notice`:  https://github.com/woocommerce/woocommerce/blob/01466a2eda027a7458ceddb0acd32eed208b1fd0/plugins/woocommerce/includes/admin/class-wc-admin-notices.php#L344-L349. Is using `add_custom_notice` what you had in mind?

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php
jimjasson: Done: https://github.com/woocommerce/woocommerce/pull/58356/commits/d9a9de3977c8e90e374305bbe010ad05117c049d

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsPage.php
jimjasson: Done: https://github.com/woocommerce/woocommerce/pull/58356/commits/d9a9de3977c8e90e374305bbe010ad05117c049d

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php
jimjasson: Done: https://github.com/woocommerce/woocommerce/pull/58356/commits/d9a9de3977c8e90e374305bbe010ad05117c049d

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php
jimjasson: Done via: https://github.com/woocommerce/woocommerce/pull/58356/commits/37755585754300903253944cca948ea5724a4949

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php
jimjasson: @xristos3490 The `get_views()` method on `WP_List_Table` base class doesn't have any content: https://github.com/WordPress/wordpress-develop/blob/6.8.1/src/wp-admin/includes/class-wp-list-table.php#L500-L502

I had a look at WordPress core classes that are extending this, like `WP_Posts_List_Table` and this class also calculates some queries in the constructor class and stores them in properties that are then latest used in views, like for example `sticky_posts_count` or `user_posts_count`.

plugins/woocommerce/src/Internal/StockNotifications/Admin/MenusController.php
jimjasson: Done via: https://github.com/woocommerce/woocommerce/pull/58356/commits/8ae350d4fcb60bbaac647e8eb7e9ef19ad133b07

jimjasson:
Thank you @xristos3490 for the thorough review! 

Many of the things you raised were carry-overs from the extension that I missed when moving the code. Good catches! 

The only two things that I left for further discussion are:
- handling admin notices and;
- queries in the Notification List Table. 

See my comments above. 

I also tested with a notification with a NULL creation date and I found some issues in `StockNotificationsDataStore`, where we were calling `getTimestamp` without first validating if the date exists. I fixed those as well, but heads-up if you need them in any of the PRs you are working on.

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php
xristos3490: > I had a look at WordPress core classes that are extending this, like WP_Posts_List_Table and this class also calculates some queries in the constructor class and stores them in properties that are then latest used in views, like for example sticky_posts_count or user_posts_count.

Don't have a strong opinion on where to count, honestly. As long as we are using the `get_views()` to render, it's OK! Thanks for mentioning the Posts list table. I drew inspiration from the HPOS orders list table, but this one works just as well.

plugins/woocommerce/src/Internal/StockNotifications/Admin/AdminManager.php
```diff
@@ -0,0 +1,62 @@
...
+		);
+
+		/*
+		 * Enqueue specific styles & scripts.
+		 */
+		if ( 'woocommerce_page_wc-customer-stock-notifications' !== $screen_id ) {
+			return;
+		}
+
+		wp_enqueue_script( 'wc-admin-customer-stock-notifications' );
```
xristos3490, line 59: This is loaded across the admin interface. Let's restrict it to the list table screen only.

xristos3490: I'd also suggest looking for opportunities to incorporate this in the `WCAdminAssets` class.

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsPage.php
xristos3490: I see what you mean! Thanks for explaining! Looking good now.

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsPage.php
```diff
@@ -0,0 +1,107 @@
...
+	const PAGE_URL = 'admin.php?page=wc-customer-stock-notifications';
+
+	/**
+	 * Init.
+	 *
+	 * @internal
+	 *
+	 * @return void
+	 */
+	final public function init() {
```
xristos3490: No need for `init()`, let's keep it in the constructor. Considering this change, the `add_submenu...` callback needs to run a method from this class specifically.

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsPage.php
```diff
@@ -0,0 +1,107 @@
...
+		// Select section.
+		$section = '';
+
+		// Nonce is checked in NotificationsPage::delete and NotificationsPage::output just displays the page.
+		if ( isset( $_GET['section'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
xristos3490: Is there a way to make this a bit more semantic? Why is a "section" a delete action? maybe use a "action" param?

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsPage.php
```diff
@@ -0,0 +1,107 @@
...
+				break;
+		}
+	}
+
+	/**
+	 * Render page.
+	 */
+	public static function output() {
+
+		if ( isset( $_GET['notice'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
xristos3490: Does this mean that we will be left with a notice GET in the URL? This suggests that every refresh will produce the same notice regardless of the action, as long as the GET is in the URL. Can we explore a way to avoid leaving footprints in the URL after actions?

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -164,8 +164,8 @@ public function create( &$notification ) {
 				'user_id'               => $notification->get_user_id( 'edit' ),
 				'user_email'            => $notification->get_user_email( 'edit' ),
 				'status'                => $notification->get_status( 'edit' ),
-				'date_created_gmt'      => gmdate( 'Y-m-d H:i:s', $notification->get_date_created( 'edit' )->getTimestamp() ),
-				'date_modified_gmt'     => gmdate( 'Y-m-d H:i:s', $notification->get_date_modified( 'edit' )->getTimestamp() ),
+				'date_created_gmt'      => $notification->get_date_created( 'edit' ) ? gmdate( 'Y-m-d H:i:s', $notification->get_date_created( 'edit' )->getTimestamp() ) : null,
```
xristos3490: Why is these needed? Considering that the model will fill-in missing dates on lines 153:158 just above?

jimjasson: @xristos3490 You are right that this fill-in mechanism exists for the `create` method, but not for the `update`, which resulted in this error when updating a notification without a creation date: 

```
[05-Jun-2025 07:16:01 UTC] PHP Fatal error:  Uncaught Error: Call to a member function getTimestamp() on null in /Users/jasonkytros/Local Sites/wp/app/public/wp-content/plugins/woocommerce-mono/plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php:280
Stack trace:
#0 /Users/jasonkytros/Local Sites/wp/app/public/wp-content/plugins/woocommerce-mono/plugins/woocommerce/includes/class-wc-data-store.php(196): Automattic\WooCommerce\Internal\DataStores\StockNotifications\StockNotificationsDataStore->update(Object(Automattic\WooCommerce\Internal\StockNotifications\Notification))
#1 /Users/jasonkytros/Local Sites/wp/app/public/wp-content/plugins/woocommerce-mono/plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php(502): WC_Data_Store->update(Object(Automattic\WooCommerce\Internal\StockNotifications\Notification))
#2 /Users/jasonkytros/Local Sites/wp/app/public/wp-content/plugins/woocommerce-mono/plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsListTable.php(382): Automattic\WooCommerce\Internal\StockNotifications\Admin\NotificationsListTable->process_bulk_action()
#3 /Users/jasonkytros/Local Sites/wp/app/public/wp-content/plugins/woocommerce-mono/plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsPage.php(50): Automattic\WooCommerce\Internal\StockNotifications\Admin\NotificationsListTable->prepare_items()
#4 /Users/jasonkytros/Local Sites/wp/app/public/wp-content/plugins/woocommerce-mono/plugins/woocommerce/src/Internal/StockNotifications/Admin/MenusController.php(109): Automattic\WooCommerce\Internal\StockNotifications\Admin\NotificationsPage::output()
#5 /Users/jasonkytros/Local Sites/wp/app/public/wp-includes/class-wp-hook.php(324): Automattic\WooCommerce\Internal\StockNotifications\Admin\MenusController->notifications_page('')
#6 /Users/jasonkytros/Local Sites/wp/app/public/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters('', Array)
#7 /Users/jasonkytros/Local Sites/wp/app/public/wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#8 /Users/jasonkytros/Local Sites/wp/app/public/wp-admin/admin.php(260): do_action('woocommerce_pag...')
#9 {main}
  thrown in /Users/jasonkytros/Local Sites/wp/app/public/wp-content/plugins/woocommerce-mono/plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php on line 280
```

I see that in the `update` function there is a fill-in mechanism just for the modified date, perhaps we should add something similar for the creation date too?

plugins/woocommerce/src/Internal/StockNotifications/Admin/AdminManager.php
jimjasson: @xristos3490 This is already enqueued only in the list table screen only: https://github.com/woocommerce/woocommerce/pull/58356/files/245a041fd771e923815576263c23a2839ce3dc9c#diff-38cab0041e3b51210695897b522fa450f6a248c54cc2f4a91e2bb3c1f966f504R55-R56

But will look into using `WCAdminAssets` methods to enqueue this script.

jimjasson: Looking at `WCAdminAssets`, there doesn't seem to be a way to use this class for scripts that don't have an `asset.php` file, which I don't think is needed here.

See: https://github.com/woocommerce/woocommerce/blob/4e08bee588c97aec594f0043b9f2c8e94e85a459/plugins/woocommerce/src/Internal/Admin/WCAdminAssets.php#L129-L145

xristos3490: >This is already enqueued only in the list table screen only: https://github.com/woocommerce/woocommerce/pull/58356/files/245a041fd771e923815576263c23a2839ce3dc9c#diff-38cab0041e3b51210695897b522fa450f6a248c54cc2f4a91e2bb3c1f966f504R55-R56

I missed that, thanks! This looks OK!

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
xristos3490: Oh, I see! Sure, let's make sure it doesn't break if an update happens with a null create date :) 

Curious about how you ended up in this situation?

jimjasson: @xristos3490 I deleted the creation date for a notification from the database and then using the list table Bulk actions, I tried to Activate/Cancel it. 

Fixed via: https://github.com/woocommerce/woocommerce/pull/58356/commits/22b89d30805e71a21cdaa97d1c3999273c680f1b

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsPage.php
jimjasson: This is a carry-over from the extension -- you are right that it can be improved. Done: https://github.com/woocommerce/woocommerce/pull/58356/commits/3a7a4d6b9916d5681fa1a00c4ed0406469b9a898

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsPage.php
jimjasson: I replaced `init` with a constructor, but not sure I understand which function to call in `add_submenu...`. I am calling the `NotificationsPage` constructor there: https://github.com/woocommerce/woocommerce/pull/58356/commits/3a7a4d6b9916d5681fa1a00c4ed0406469b9a898

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsPage.php
jimjasson: @xristos3490 The only other option I can think of is to use core's `\WC_Admin_Notices::add_custom_notice` but these notices are not automatically cleared on refresh. The user must manually dismiss the notice. 

Do you have any other suggestions perhaps?

jimjasson: As discussed on our call, decided to use a custom option to temporarily store and display notices: https://github.com/woocommerce/woocommerce/pull/58356/commits/5457a8eb0a7c60df961196abc21d18014572ca83

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
```diff
@@ -455,12 +459,35 @@ public function query( array $args ) {
...
+		$order_by         = '';
+		$order_by_clauses = array();
+
+		if ( $args['order_by'] && is_array( $args['order_by'] ) ) {
+			foreach ( $args['order_by'] as $what => $how ) {
+				$order_by_clauses[] = $table . '.' . esc_sql( strval( $what ) ) . ' ' . esc_sql( strval( $how ) );
+			}
+		}
+
+		$order_by_clauses = empty( $order_by_clauses ) ? array( $table . '.id ASC' ) : $order_by_clauses;
```
xristos3490: Let's avoid having a default ORDER here. It's better for the MySQL optimizer.

plugins/woocommerce/src/Internal/StockNotifications/Admin/MenusController.php
xristos3490: Alright! Let's remove it!

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsPage.php
```diff
@@ -0,0 +1,91 @@
...
+		$table->prepare_items();
+		include __DIR__ . '/Views/html-admin-notifications.php';
+	}
+
+	/**
+	 * Delete notification.
+	 *
+	 * @throws \Exception If notification ID is missing.
+	 */
+	public static function delete() {
```
xristos3490: Nit. Separation of concerns here would help. Ideally, the delete method should accept a notification ID. The parsing + nonce checking should be done in the constructor.

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsPage.php
```diff
@@ -0,0 +1,91 @@
...
+				break;
+		}
+	}
+
+	/**
+	 * Render page.
+	 */
+	public static function output() {
+
+		$search = isset( $_REQUEST['s'] ) ? sanitize_text_field( wp_unslash( $_REQUEST['s'] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
xristos3490: How is this used here? 🤔

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsPage.php
```diff
@@ -0,0 +1,91 @@
...
+		$table  = new NotificationsListTable();
+		$table->prepare_items();
+		include __DIR__ . '/Views/html-admin-notifications.php';
```
xristos3490: No need to capitalize "Views"; it can just be "views". Typically, we capitalize folders that are part of the namespace itself.

xristos3490 (commented):
Left a few more comments before diving into manual review!

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsPage.php
jimjasson: It is used in the `html-admin-notifications.php` view.

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsPage.php
jimjasson: Done!

plugins/woocommerce/src/Internal/StockNotifications/Admin/NotificationsPage.php
jimjasson: Done!

plugins/woocommerce/src/Internal/DataStores/StockNotifications/StockNotificationsDataStore.php
jimjasson: Done!

xristos3490 (approved):
Looking good, @jimjasson! 

Made a few minor adjustments for better flow, mostly by reorganizing and simplifying a bit.

Approved!

PR #58370, opened 2025-05-29
Render cart items in Mini Cart drawer with iAPI, support increase / decrease quantity

samueljseay:

This PR introduces rendering of the mini cart items in the drawer via the interactivity API (behind experimental flag like in previous PRs).

Not all functionality is yet supported but increasing/decreasing quantities is supported via the UI steppers. 

Also note that there is a flicker of items when changing quantities. I plan to address improvements to the store and this issue in a follow up PR. I will also support removing items entirely in a separate PR

https://github.com/user-attachments/assets/c16105df-9f1e-4df3-b34d-2d6574de28c5

3. Reload site on frontend, see the iAPI powered mini cart which has some functionality implemented.
4. Add some items to your cart on the shop page
5. Open drawer, see the items rendered
6. Use the `+` and `-` steppers to increase/decrease quantities of the items (noting that removing/going to 0 is not implemented yet)

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 
[Experimental] Render mini cart items using interactivity API when feature flag is enabled.

</details>


plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
```diff
@@ -164,3 +171,151 @@ store(
...
+
+		*decrementQuantity(): Generator< unknown, void > {
+			const { cartItem } = getContext< CartItemContext >();
+			const { actions } = store< WooCommerce >(
+				'woocommerce',
+				{},
+				{ lock: universalLock }
+			);
+
+			yield actions.addCartItem( {
```
samueljseay, line 315: I realize this API as is, is not well suited to this purpose. I'll do some additions to the store itself in a separate follow up PR.

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartItemsBlock.php
```diff
@@ -11,4 +13,126 @@ class MiniCartItemsBlock extends AbstractInnerBlock {
...
+		$reduce_quantity_label = __( 'Reduce quantity of %s', 'woocommerce' );
+
+		// translators: %s is the name of the product in cart.
+		$increase_quantity_label = __( 'Increase quantity of %s', 'woocommerce' );
+
+		// translators: %s is the name of the product in cart.
+		$quantity_description_label = __( 'Quantity of %s in your cart', 'woocommerce' );
+
+		// translators: %s is the name of the product in cart.
+		$remove_from_cart_label = __( 'Remove %s from cart', 'woocommerce' );
```
luisherranz, line 55: In the Product Button block, [Luigi replaced `%s` with `###`](https://github.com/woocommerce/woocommerce/blob/bfe52f1f5f11b1b60eba0b71217254b206de94fe/plugins/woocommerce/src/Blocks/BlockTypes/ProductButton.php#L114-L119). I guess that's not necessary, right? We should probably change it there too, for consistency.

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartItemsBlock.php
```diff
@@ -11,4 +13,126 @@ class MiniCartItemsBlock extends AbstractInnerBlock {
...
+				'increaseQuantityLabel'    => $increase_quantity_label,
+				'quantityDescriptionLabel' => $quantity_description_label,
+				'removeFromCartLabel'      => $remove_from_cart_label,
+			)
+		);
+
+		wp_interactivity_state(
+			$this->get_full_block_name(),
+			array(
+				'cartItems' => wp_interactivity_state( 'woocommerce' )['cart']['items'] ?? [],
```
luisherranz, line 70: Hmmm... this means we're sending the cart items information to the frontend twice. I don't like that. It's a devex optimization that has a performance cost. I didn't realize it when we talked about it.

To be honest, right now I don't know if it would be better to use the context of the `woocommerce` namespace, or if we should allow changing the namespace in the Interactivity API. Probably the former.

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
```diff
@@ -164,3 +170,156 @@ store(
...
+store( 'woocommerce/mini-cart-items-block', {
+	state: {
+		// Intended to be used in context of a cart item in wp-each
+		get reduceQuantityLabel() {
+			const { reduceQuantityLabel } = getConfig(
+				'woocommerce/mini-cart-items-block'
+			);
+			const { cartItem } = getContext< CartItemContext >();
+
+			return reduceQuantityLabel.replace( '%s', cartItem.name );
```
luisherranz: Maybe? (also in the other getters)

```suggestion
			return reduceQuantityLabel.replace( '%s', state.cartItemName );
```

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
```diff
@@ -164,3 +170,156 @@ store(
...
+
+			return removeFromCartLabel.replace( '%s', cartItem.name );
+		},
+
+		get cartItemName() {
+			const { cartItem } = getContext< CartItemContext >();
+			const txt = document.createElement( 'textarea' );
+			txt.innerHTML = cartItem.name;
+			return txt.value;
+		},
```
luisherranz: I wonder about this. I understand that you are trying to espace the values, but should we allow HTML to be injected instead? How does it work as of today? And if we shouldn't allow HTML, wouldn't it be better/cleaner to directly remove the tags (`innerText`)?

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartItemsBlock.php
```diff
@@ -11,4 +13,126 @@ class MiniCartItemsBlock extends AbstractInnerBlock {
...
+										<a data-wp-text="state.cartItemName" data-wp-bind--href="context.cartItem.permalink" class="wc-block-components-product-name"></a>
+										<div class="wc-block-cart-item__prices">
+											<span class="price wc-block-components-product-price">
+												<span data-wp-text="state.itemPrice" class="wc-block-formatted-money-amount wc-block-components-formatted-money-amount wc-block-components-product-price__value">
+												</span>
+											</span>
+										</div>
+										<div class="wc-block-components-product-metadata">
+											<div data-wp-watch="state.itemShortDescription" >
+												<div class="wc-block-components-product-metadata__description" data-wp-ignore></div>
```
luisherranz: Oh, this is very interesting. I wonder why you need `data-wp-ignore` here. I was thinking about deprecating and the removing it, as we haven't found any use case for it (and it's a very complex and confusing directive).

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
```diff
@@ -164,3 +170,156 @@ store(
...
+
+		get cartItems() {
+			return wooStoreState.cart.items;
+		},
+
+		// Intended to be used in context of a cart item in wp-each
+		itemShortDescription() {
```
luisherranz: My take is that this should be moved to `callbacks`. I understand that it is due to the limitation of not yet having a directive for the HTML, but I'd move it anyway to keep only state and derived state inside `state`.

luisherranz (approved):
LGTM, Sam. This is looking great so far 🙂👏

Pre-approving with some comments.

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartItemsBlock.php
luisherranz: I tried switching to `woocommerce` here https://github.com/woocommerce/woocommerce/pull/58370/commits/80387de367ae0abb1357e2314155903fa16d5ce1. It doesn't seem too terrible to me. At least for now.

luisherranz: Another alternative would be to derive only the cart item:

```js
get cartItem() {
  const { cartItem } = getContext< CartItemContext >( 'woocommerce' );
  return cartItem;
}
```

luisherranz: I did so in https://github.com/woocommerce/woocommerce/pull/58370/commits/1058bc4ef4d2c02762a8ea3b93e96249369b9bf3. 

This is probably a bit cleaner. What do you think?

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
```diff
@@ -164,3 +170,145 @@ store(
...
+	state: {
+		get cartItem() {
+			return getContext< CartItemContext >( 'woocommerce' ).cartItem;
+		},
+
+		// Intended to be used in context of a cart item in wp-each
+		get reduceQuantityLabel(): string {
+			const { reduceQuantityLabel } = getConfig(
+				'woocommerce/mini-cart-items-block'
+			);
```
luisherranz: I was thinking. Maybe it's not necessary to execute the `getConfig` every time a getter is executed. Since these values don't change, they could be moved out of the store. 

For reference: `getConfig` inherits the namespace, so `getConfig()` will get the store namespace automatically, which is nice, but it doesn't have anything to do with the scope, so it's safe to execute outside of a store as long as you provide the namespace.

luisherranz: I've tried that and it seems to work well: https://github.com/woocommerce/woocommerce/pull/58370/commits/ed722b0dca25765d16519b7881a7935f9371107a

Let me know what you think.

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
samueljseay: Escaping? I'm decoding here.  

The sample data for products has product descriptions coming through with HTML encoded entities like `Shirt &#8211; Cream` Thats the kind of thing we're getting from the Cart response, so I'm just trying to decode it here. But I don't believe there is actual HTML (I better check that assumption).

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
samueljseay: Yeah that makes sense 👍🏻

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
samueljseay: Good point. For some reason I assumed scope was necessary, just lack of knowledge of the API. its a good change 👍🏻

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
luisherranz: Oh, yeah, sorry. I misunderstood it. All good then, thanks 🙂

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartItemsBlock.php
samueljseay: I don't see why it would be necessary. I can ask him about it and do a follow up for consistency.

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartItemsBlock.php
samueljseay: Thats a nice clean up thanks!

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartItemsBlock.php
samueljseay: Ah I forgot to revisit this 🤦🏻‍♂️ . I assumed by the name of it that it was a way to tell interactivity API that I don't want it to control rendering this region, now looking at it I see it is sorta like that, but would likely only make sense to do it, if it has some initial inner content to cache? I think I can remove this.

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
samueljseay: This is basically solved by your commit with derived state for the cart item. 🎉

plugins/woocommerce/src/Blocks/BlockTypes/MiniCartItemsBlock.php
luisherranz: I think it's okay to delete it and see if everything works as it should. If it doesn't work, we'll research why.

samueljseay:
@luisherranz Could you help me debug tomorrow, there seem to be some issues with reflecting state updates, I can't work out where it's going wrong.
I've incorporated the wp-key change needed for the each to work properly, but cart item quantities don't update using the steppers anymore with the new derived state for the item context, but I can't see why they shouldn't. 

I'm wondering if server side directive handling is part of the issue?

samueljseay:
@luisherranz just adding some more thoughts:

A revert to the way we were using context before works: https://github.com/woocommerce/woocommerce/pull/58370/commits/3ff87857ed5d017a507ae5213aee53c77c03b0e1

We should still figure out whats going on, but one thing I don't like about the changes you made is its quite hard to reason about. I come back to finding the "context" being set to `woocommerce` quite confusing (and how that becomes the context for items in the loop), but also that we add another layer more indirection for a derived getter for the `woocommerce` context's cart item. It feels like coming back to maintain it that finding the source of the context feels like detective work.

It results in less code, but overall a lot more confusing in my opinion.

Still would be great to pair and try figure out why the current pattern is not working properly for me.

EDIT: I kept going with a new branch for the follow on quantity limits work. This branch: https://github.com/woocommerce/woocommerce/compare/dev/redo-cart-set-qty?expand=1 removes the duplicated `cartItems` state, but still relies on getContext from within the derived getters and everything functions as expected. I just can't get your original code of getting state directly from woocommerce namespace in the directive to work for me.

luisherranz (changes requested):
After investigating with @samueljseay  and @DAreRodz, we discovered that the problem is a bug in `data-wp-each`, and @DAreRodz will prepare a pull request in the Interactivity API to solve it.

There is still an issue in the `lineItemTotal` update, for which we have to find a workaround.

samueljseay:
@luisherranz how do you feel about merging this so we can continue? I've continued to iterate in https://github.com/woocommerce/woocommerce/pull/58589 and also in https://github.com/woocommerce/woocommerce/compare/wooplug-4502-interactivity-api-powered-mini-cart-display-salediscount?expand=1

luisherranz (approved):
Yes, let's do that. Sorry for the request changes.

PR #58379, opened 2025-05-29
[NOX in LYS] Introduce installation step for WooPayments if the plugin is not active

mordeth:

This PR introduces a new step in the onboarding flow to handle the installation of the WooPayments plugin when it is not currently activated/installed.

If the plugin is missing or inactive, users will now see a dedicated step prompting them to install and activate WooPayments before continuing with the rest of the setup process.

Closes WOOPLUG-4402.

| Before | After |
| ------ | ----- |
| Screenshot 2025-05-30 at 11 34 34 | Screenshot 2025-05-30 at 11 30 52 |

- Create a new JN store with this PR's branch build (here's a [direct create link](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce-beta-tester&woocommerce&branches.woocommerce=update/WOOPLUG-4402-install-woopayments-before-lys-onboarding)).
- Go to the new store's dashboard, click on WooCommerce > Home to trigger the core profiler, skip it, and set your store's location to Austria.
- Go to WooCommerce > Home and click on Launch your store task.
Screenshot 2025-05-30 at 12 14 13

- Click on the Set up payments link.
Screenshot 2025-05-30 at 12 15 33

- Verify that a page to install WooPayments is displayed. Click on Install.
Screenshot 2025-05-30 at 12 16 24

- Once the NOX onboarding starts, verify that the 'Install WooPayments' step is not removed from the list but is completed.
Screenshot 2025-05-30 at 12 38 28

- Click on the Set up WooPayments back button.
Screenshot 2025-05-30 at 12 38 28

- Click again on the Set up payments.
Screenshot 2025-05-30 at 12 15 33

- Verify that the 'Install WooPayments' step is not removed from the list but is completed.
Screenshot 2025-05-30 at 12 38 28

- Refresh the page and verify the Install WooPayments step is removed from the list.
Screenshot 2025-05-30 at 12 38 42

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/admin/client/launch-your-store/data/installation-context.tsx
+++ b/plugins/woocommerce/client/admin/client/launch-your-store/data/installation-context.tsx
@@ -33,6 +33,7 @@ export const InstallationProvider: React.FC< {
 	children: React.ReactNode;
 	closeModal: () => void;
 } > = ( { children, closeModal } ) => {
+	// Check if WooPayments is active by looking for the plugin in the active plugins list
 	const isWooPaymentsActive = useSelect(
 		( select ) =>
 			select( pluginsStore )
--- a/plugins/woocommerce/client/admin/client/launch-your-store/data/installation-context.tsx
+++ b/plugins/woocommerce/client/admin/client/launch-your-store/data/installation-context.tsx
@@ -1,0 +1,92 @@
+/**
+ * External dependencies
+ */
+import { createContext, useContext, useState } from '@wordpress/element';
+import { useSelect } from '@wordpress/data';
+import { pluginsStore } from '@woocommerce/data';
+import { getNewPath } from '@woocommerce/navigation';
+
+/**
+ * Internal dependencies
+ */
+import { LYSPaymentsSteps } from '~/settings-payments/onboarding/providers/woopayments/steps';
+import { OnboardingProvider } from '~/settings-payments/onboarding/providers/woopayments/data/onboarding-context';
+
+interface InstallationContextType {
+	isWooPaymentsActive: boolean;
+	wooPaymentsRecentlyActivated: boolean;
+	setWooPaymentsRecentlyActivated: ( value: boolean ) => void;
+}
+
+/**
+ * Context to manage onboarding steps
+ */
+const InstallationContext = createContext< InstallationContextType >( {
+	isWooPaymentsActive: false,
+	wooPaymentsRecentlyActivated: false,
+	setWooPaymentsRecentlyActivated: () => undefined,
+} );
+
+export const useInstallationContext = () => useContext( InstallationContext );
+
+export const InstallationProvider: React.FC< {
+	children: React.ReactNode;
+	closeModal: () => void;
+} > = ( { children, closeModal } ) => {
+	const isWooPaymentsActive = useSelect(
+		( select ) =>
+			select( pluginsStore )
+				.getActivePlugins()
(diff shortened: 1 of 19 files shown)


elazzabi (changes requested):
I just looked at your PR. Thank you for giving it a spin 🙌 I knew it wouldn't be easy! 

While it works, I'm worried about the changes made to the context.

We've made some early progress and removed most of the hardcoded values, including the list of steps and the navigation strategy. However, it appears we are reverting to hardcoding with the `install_woopayments` step id.

Additionally, the changes made are useful in LYS only, which will make the context less and less general.

Can we instead try this to see if it's a better solution:
1. On the frontend, if WooPayments isn't installed, let's render a fake list of steps (which can be one step, such as `Install WooPayments`, or if we need all steps, let's use the list of LYS steps directly) and a fake NOX by showing the Install screen you worked on here.
2. Once the installation is done, we then switch to the NOX.

This will have less (or possibly no) impact on the context, while keeping things clean and easier to understand. Also, we can keep the thinking model that NOX is only available if WooPayments is installed, and isn't responsible for installing it.

WDYT?

mordeth:
Thank you for the review, Ahmed!

I agree that moving the logic out of the context provider is a better approach. I’ve now excluded the “Install WooPayments” step from the step list, as modifying it outside the provider would be difficult. Would you mind taking another look?

mordeth:
Thank you for the feedback, @elazzabi!

I agree that WooPayments installation logic shouldn’t be mixed with the `OnboardingContext`. That’s why I introduced a separate `InstallationContext` to handle that responsibility and conditionally wrap the `OnboardingContext` only when needed. This change also resolves the issue of unnecessary requests to the WooPayments onboarding endpoint when the plugin isn’t activated, which previously led to errors.

Additionally, we’re no longer relying on onboarding API errors to determine whether the plugin is activated.

Would you mind giving it another review, please?

plugins/woocommerce/client/admin/client/launch-your-store/data/installation-context.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+	return (
+		<InstallationContext.Provider
+			value={ {
+				isWooPaymentsActive,
+				wooPaymentsRecentlyActivated,
+				setWooPaymentsRecentlyActivated,
+			} }
+		>
+			{ isWooPaymentsActive && (
```
elazzabi: This would be better if rendered outside the installation context. This way, it will be clear in the LYS hub that there is an installation + onboarding context. Now it looks like it's the installation only, but there are both.

Wdyt?

elazzabi (approved):
The new changes look perfect! I left a small comment but this is looking much better! Thank you 🙇

plugins/woocommerce/client/admin/client/launch-your-store/data/installation-context.tsx
mordeth: That makes total sense. I’ve updated the provider name to the more general SetUpPaymentsProvider, which will handle both the installation and onboarding of payment plugins.

elazzabi (approved):
This looks great! Thank you 🚢 

CleanShot 2025-06-02 at 15 10 09@2x

Issue #56870, opened 2025-03-31
Blockified Add to Cart with Options block: display variation description, price and stock status when attributes are selected in a variable product


Aljullu:
In the classic templates, when selecting the attributes for a specific variation, the variation description, price, and stock status (when out of stock) is displayed:

[video.webm](https://uploads.linear.app/f629fcde-6f9b-48e8-8af3-60eed8823bff/59b98fc3-f4f7-4ed7-839f-5edff37318c8/357f3a78-9e64-4999-b0ff-79b9e434b009?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2Y2MjlmY2RlLTZmOWItNDhlOC04YWYzLTYwZWVkODgyM2JmZi81OWI5OGZjMy1mNGY3LTRlZDctODM5Zi01ZWRmZjM3MzE4YzgvMzU3ZjNhNzgtOWU2NC00OTk5LWIwZmYtNzliOWU0MzRiMDA5IiwiaWF0IjoxNzQzNDM0MjgyLCJleHAiOjMzMzEzOTk0MjgyfQ.ET_qfZiQq6BYN2rD8iOiQLBSsMP_iftW92KwtE2Sfp4)

In the blockified *Add to Cart with Options* block we would like to change this a bit, and it should work like this:

* The Product Price that is displaying the price range (ie: 42.00 € – 45.00 €) would update to display the selected variation price (ie: 45.00 €).
  * https://github.com/woocommerce/woocommerce/pull/58384.
* The Product Description block would be updated to display the variation description (ie: This is a variable product would become Lorem ipsum… when that variation is selected).
  * https://github.com/woocommerce/woocommerce/issues/59439
* Similarly, the SKU block should be updated.
  * https://github.com/woocommerce/woocommerce/issues/59440
* The Stock Indicator block is not displayed by default. So in this case, it would appear when necessary. I guess after the Variation Selector and before the Quantity Selector and Add to Cart Button is the best place?
  * https://github.com/woocommerce/woocommerce/issues/59441
* _Additional Information_ in Product Details should update with the data for the selected variation (weight and dimensions).
  * https://github.com/woocommerce/woocommerce/issues/59442
* The _Product Quantity_ block might need to hid/show depending on whether the variation is sold individually. Min, max and step values should be updated as well.
  * https://github.com/woocommerce/woocommerce/issues/59443
[...]

PR #58384, opened 2025-05-29
implement context management for displayed product

gigitux:

This PR adds **context and state management** for the displayed product on the **Single Product** page and within the **Single Product block**.

All atomic blocks related to the `product` entity should now access product data via:

- **State**: in the Single Product Template  
- **Context**: within the Single Product block

This ensures allows atomic blocks to dynamically respond to changes in the product data.

Part of https://github.com/woocommerce/woocommerce/issues/56870.

#### Current Scope

At this stage, only the **Product Price** block reads data from the state/context.  
If this approach is approved, the goal is to migrate the other atomic blocks (if needed) to follow the same pattern.

#### Technical Details

When the Single Product Template is rendered, a global state is created containing all relevant data for the displayed product.  

https://github.com/user-attachments/assets/624cdc03-79c6-4f98-8861-37115d8023aa

1. Install and activate the [WooCommerce Beta Tester](https://woocommerce.com/products/woocommerce-beta-tester/) extension.
2. Go to *WooCommerce Admin Test Helper* > *Features* and enable `experimental-blocks` and `blockified-add-to-cart`.
3. Go to *Appearance* > *Editor* > *Templates* > *Single Product*.
4. Click on the *Add to Cart with Options* block and update to the blockified version.
5. Add the Single Product block and select a product variable type.
6. Save.
7. Visit a Variable Product.
8. Select a different configuration, ensuring that the Product Price is updated.
9. Do the same for the Product Variable inside the Single Product block.

## Summary by CodeRabbit

- **New Features**
  - Product price now updates dynamically in response to attribute selections within the Add to Cart + Options block for variable products.
  - Enhanced interactivity and real-time updates for product price display in relevant product blocks.
  - Added interactive attributes and scripts to product price and single product blocks for variable products.
- **Bug Fixes**
  - Ensured accurate price display when selecting different product variations.
- **Tests**
  - Added tests to verify that product price updates correctly when variation options are changed.
- **Documentation**
  - Updated changelog to reflect new dynamic pricing functionality.
- **Style**
  - Improved configuration formatting for consistency and readability.


plugins/woocommerce/src/Blocks/Utils/BlocksSharedState.php
```diff
@@ -101,6 +101,21 @@ public function register_cart_interactivity( $consent_statement ) {
...
+	 * Register interactivity state for single product template that is needed by multiple blocks.
+	 *
+	 * @param \WC_Product $product - The product to register.
+	 * @return void
+	 */
+	public function register_single_product_interactivity( $product ) {
+		wp_interactivity_state(
+			'woocommerce',
+			array(
+				'displayedProduct' => ProductDataUtils::get_product_data( $product ),
```
gigitux: I feel that "displayedProduct" is too much generic. What do you think about:

```
array(
    'singleProductTemplate' => array(
        'displayedProduct' => ...
    )
)
```

plugins/woocommerce/src/Blocks/Utils/ProductDataUtils.php
```diff
@@ -0,0 +1,34 @@
...
+	 *
+	 * @param \WC_Product $product Product object.
+	 * @return string The product price HTML.
+	 */
+	public static function get_product_price_html( $product ) {
+		return array(
+			'price_html' => $product->get_price_html(),
+		);
+	}
+}
```
gigitux: The idea is that this class is used by the Single Product Block and the Single Product template to register product data that atomic blocks can consume.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/price/frontend.ts
```diff
@@ -0,0 +1,49 @@
...
+
+				const context = getContext< {
+					displayedProduct: DisplayedProduct;
+				} >( 'woocommerce/single-product' );
+
+				if ( context ) {
+					element.ref.innerHTML = context.displayedProduct.price_html;
+				} else if ( wooState.displayedProduct ) {
+					element.ref.innerHTML =
+						wooState.displayedProduct.price_html;
```
gigitux: We need to sanitize the HTML before setting it. I can take care of it if you agree with the approach. Unfortunately, we have to use the “price HTML” directly, as this part of WooCommerce contains a lot of logic and is highly extensible.

gigitux: Otherwise, we should re-think the Product Price block and move it to a full block-paradigm.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/price/frontend.ts
```diff
@@ -0,0 +1,57 @@
...
+					displayedProduct: DisplayedProduct;
+				} >( 'woocommerce/single-product' );
+
+				if ( singleProductContext ) {
+					element.ref.innerHTML =
+						singleProductContext.displayedProduct.price_html;
+				} else if ( productCollectionContext ) {
+					element.ref.innerHTML =
+						productCollectionContext.displayedProduct.price_html;
+				} else if ( wooState.displayedProduct ) {
```
gigitux: We could make a generic util that can be used across atomic blocks

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/price/frontend.ts
Aljullu: > We need to sanitize the HTML before setting it. I can take care of it if you agree with the approach.

@gigitux did you have an idea on the best way to sanitize this HTML? We have a [`sanitizeHTML()` util](https://github.com/woocommerce/woocommerce/blob/2e1679402d2526fd8cc719781b4a84e326e6a6cd/plugins/woocommerce/client/blocks/assets/js/utils/sanitize-html.ts#L9), but wondering if you had something else in mind. :thinking:

plugins/woocommerce/src/Blocks/Utils/BlocksSharedState.php
Aljullu: In 153eb4337f162c93bfaef6bab5d8bb17702ab11b I did a couple of changes:

1. Renamed `displayedProduct` to `productData`. This way it's in consistent with the `ProductDataUtils` class.
2. Moved it under `singleProductTemplate` in the global state.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/price/frontend.ts
gigitux: Not really, but I'm fine with using `sanitizeHTML` util. At the same time, I would check the impact on the bundle size. According [to bundlephobia](https://bundlephobia.com/package/dompurify@3.2.6), dompurify is pretty heavy, but it is tree-shakeable. I don't know why, but we are importing the entire module:

https://github.com/woocommerce/woocommerce/blob/181e81e3e2c6d5be8b9dd9f0d3034ac50fcced5c/plugins/woocommerce/client/blocks/assets/js/utils/sanitize-html.ts#L9-L21

We could import only `sanitize` to reduce the size. 


Also, I think this is a good interim solution, but we should avoid to set HTML even from the extensibility point of view: third-party developers could abuse it.

plugins/woocommerce/src/Blocks/Utils/BlocksSharedState.php
gigitux: It makes sense! Thanks! 🙇

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/price/frontend.ts
Aljullu: Done in https://github.com/woocommerce/woocommerce/pull/58384/commits/a43523724dc4f2abb10303d84e25e7d1f0b46bb8.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/price/frontend.ts
```diff
@@ -0,0 +1,84 @@
...
+				const element = getElement();
+
+				if ( ! element.ref ) {
+					return;
+				}
+
+				const newPrice = getProductData( 'price_html' );
+
+				if ( newPrice ) {
+					element.ref.innerHTML = newPrice;
```
dinhtungdu: I can see that we're doing this because we use `get_price_html` method, which is filterable. But updating the price this way is not the "Interactivity API" way.

As the PR description states:

> For example, a coupon plugin could update the displayed price in real time by modifying the product state or context.

If we're doing this, extenders want to update the price will have to update the price HTML, not just the value of regular and sale prices. And how could they do it? Can we try updating the price as an extender to see how it looks?

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
```diff
@@ -228,6 +233,46 @@ const addToCartWithOptionsStore = store(
...
 			},
 		},
+		callbacks: {
+			setProductData: () => {
+				const { availableVariations, selectedAttributes } =
+					getContext< Context >();
+				const matchedVariation = getMatchedVariation(
+					availableVariations,
+					selectedAttributes
+				);
```
dinhtungdu, line 350: If the price HTML is stored in the `context.availableVariations` already, can we derive it instead of syncing like we're doing now? Syncing state makes it easier for synced data to be outdated, I learned it the hard way while working on Product Filters, which leads to store rewrite at the end.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
```diff
@@ -110,6 +113,8 @@ const dispatchChangeEvent = ( inputElement: HTMLInputElement ) => {
 	inputElement.dispatchEvent( event );
 };
 
+let originalPriceHtml = '';
```
dinhtungdu: This causes the data leak. For example, I added the Single Product block to the Single Product Template and the price of main product is leaked into the block:
image

dinhtungdu (changes requested):
I tested this PR and it's nearly working as expected. But I'm wary about syncing state as we're doing here. Instead, I'd recommend to try using derived state first, like Luiz recommended [here](https://github.com/woocommerce/woocommerce/pull/58097#pullrequestreview-2850048798).

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/price/frontend.ts
Aljullu: I've removed that part from the PR description, I would like to handle extensibility all at once as part of #55700, so we can figure out the best APIs with a wide view instead of in small parts. We might or might not want to let extensions modify `productData` (or `productData.price_html`) directly and I don't think it's a decision that we need to do now. I agree with your point that this is not the "Interactivity API way".

We can decide to keep `price_html` internal and have specific APIs for extensions to update the price using the iAPI, for example only allowing extensions to define certain values that we require to be strings/numbers (price prefix, old price, new price, price range, price suffix, etc.) but not HTML. `price_html` could eventually be a derived state of all those strings/numbers so extensions update them but don't update `price_html` directly.

However, for backwards compatibility, we need to support setting the price as an HTML internally. Even if that's not something we plan to expose for iAPI-powered extensions, it's still used by legacy extensions. Do you think that makes sense?

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
Aljullu: How do you envision this? The `price_html` and other `productData` need to be stored either in the _Single Product_ block context or the global state. We can't know where from the _Add to Cart + Options_ store, that's why we are syncing it here. I'm not sure how it could be implemented as derived state. :thinking:

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
Aljullu: Ups, good catch! I refactored that part in the last commit. Now instead of saving a global, we use two values in the context/state: `originalProductData` and `productData`. `productData` is only used to override the values to display and we set it to `null` when we want to return to displaying `originalProductData`. This way, the values of different products don't get mixed up.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
dinhtungdu: There are two cases here, right? Add to Cart + Options block is used inside Single Product Template and inside Single Product block. In Add to Cart + Options store, I think we can do something like this:

```js
const addToCartWithOptionsStore = store(
	'woocommerce/add-to-cart-with-options',
	{
		state: {
			get productData() {
				const singleProductContext = getContext('woocommerce/single-product');
				if(singleProductContext?.productData){
					return singleProductContext.productData;
				}
				const { state: wooState } = store(
					'woocommerce',
					{},
					{ lock: universalLock }
				);
				if(wooState?.productData){
					return wooState.productData;
				}
				return null;
			},
```
Then in other state/action, we can use `state.productData` without worrying about where it comes from.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/price/frontend.ts
dinhtungdu: I agree that we can work on extensibility later.

About the backward compatibility, I just want to make sure we're on the same page. To me, backward compat in the case of product price is supporting PHP filters in `wc_price()` and `get_price_html()` methods and I don't think we can support more than that. For example, when you choose an addon added by Product Add-Ons extension, it shows the subtotal price its own way, not updating the product price as we want. Extensions need to update to support iAPI (which is the extensibility btw).

What I'm wary about is that by updating whole HTML instead of subscribing to state/context, we're somewhat losing reactivity. Ideally, IMO, the variation data will have regular and sale price in numeric value, not the `price_html` string. And the Product Price block has `regularPrice` and `salePrice` as derived state and uses `data-wp-text` directive to subscribe to their value. But in reality, the variable product price can be a range, making the HTML markup structure different. I don't know how to solve it, just explaining my chain of thought.

I'm not totally against updating price by replacing `innerHTML` as we have a reason for it. But I don't think it's a good pattern to follow.

Aljullu: I have implemented an alternative approach in my last commit: 97134145089abc185f84e951e243f8e32368798f. There are some things to flesh out (like the deep import of a _Product Filters_ util), but it works, and I think it's more aligned with what you had in mind?

https://github.com/user-attachments/assets/355ed4ab-b3c0-44f2-9a59-1e18305a3390

It relies on `product_html` to display the product price to avoid breaking the existing behavior. But when variations are selected, it updates the price via the iAPI, based on the `display_price` and `display_regular_price` values.

I don't think we can use `data-wp-text` because we need to change the markup based on the selected variation (ie, adding a `<del>` element for products on sale), but I might be missing something. :thinking: 

Of course, this approach has the _con_ that it doesn't support extensions that modify the variations' `price_html`. I'm fine with it if we consider this a fair trade-off, specially considering the _Add to Cart + Options_ block is still in beta.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
Aljullu: Thanks for all the feedback, @dinhtungdu, I really appreciate it! :raised_hands: 

> Then in other state/action, we can use state.productData without worrying about where it comes from.

I don't fully understand which other state/actions you mean. The only block that "reads" `productData` is the _Product Price_ block, and it reads it from the _Single Product_ store or the global state, not from the _Add to Cart + Options_ state.

That's because _Product Price_ can be outside the _Add to Cart + Options_ block. In fact, the _Add to Cart + Options_ block might not be on the page at all. And while currently _Add to Cart + Options_ is the only block that can modify `productData`, in the future there might be others.

So what's the benefit of having `productData` as state in the _Add to Cart + Options_ store if blocks are not going to read `productData` from it? :thinking:

dinhtungdu: It's my bad to mention the Add to Cart + Options block. I meant that if a block need to access product data from parent region, it can define the derived `productData` state like above, and then use that state in other place within their store. Product Price can defined the `productData` state, then in the `setNewPrice` callback for example, we can do something like this:

```
				const { display_price, display_regular_price} = state.productData;
```

Aljullu:
@dinhtungdu I updated this PR with what we discussed in Slack. Now, there is a `woocommerce/single-product` store that has `productData` and `originalProductData` as derived state and `setProductData()` as an action. This way, no matter if the consumers interact with the global state or a local store, the API is the same. Let me know how this approach looks. :slightly_smiling_face: 

(Tests are failing because of this unrelated issue: https://github.com/woocommerce/woocommerce/pull/58783)

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/price/frontend.ts
```diff
@@ -0,0 +1,130 @@
...
+					) }</span> <ins aria-hidden="true"><span class="woocommerce-Price-amount amount"><bdi>${ formatPrice(
+						newPrice
+					) }</bdi></span></ins><span class="screen-reader-text">${ sprintf(
+						currentPriceText,
+						formatPrice( newPrice )
+					) }</span></span>`;
+				} else if ( newPrice ) {
+					return `<span class="price"><span class="woocommerce-Price-amount amount"><bdi>${ formatPrice(
+						newPrice
+					) }</bdi></span></span>`;
```
dinhtungdu: Unfortunately, if we want to support legacy PHP filter used in `get_price_html()` method, we can't do this. Because the HTML structure of the price is filterable and can be changed, which may tied to styling.

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/single-product.ts
dinhtungdu, line 1: Is this one a global store or belong to the Single Product block?

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
```diff
@@ -337,6 +341,36 @@ const addToCartWithOptionsStore = store(
 				}
 			},
 		},
+		callbacks: {
+			setProductData: () => {
+				const { availableVariations, selectedAttributes } =
+					getContext< Context >();
```
dinhtungdu, line 346: We can revisit this context later but sharing my thoughts here to not lose it.

I think we can even move availableVariations and selectedAttribute to Single Product Template and Single Product block under the `productData`, by doing so, we don't have to sync the state anymore.

plugins/woocommerce/client/blocks/bin/webpack-helpers.js
```diff
@@ -97,6 +97,10 @@ const getAlias = ( options = {} ) => {
 			__dirname,
 			`../assets/js/templates/`
 		),
+		'@woocommerce/base-stores': path.resolve(
+			__dirname,
+			`../assets/js/base/stores/`
+		),
```
dinhtungdu: Why do we need this given that we already have some store in this directory?

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/price/frontend.ts
Aljullu: Yes, it was mostly to try the approach discussed in this thread https://github.com/woocommerce/woocommerce/pull/58384#discussion_r2136894275. But if we all agree, I also prefer using `price_html`, as it offers backwards compatibility and keeps things simpler. I switched back to using `price_html` in c90b6dd7b679a9f0d65a51049b4848f38b6dc3fa.

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/single-product.ts
Aljullu: It's a global store intended to be used when there is the Single Product template or block on the page. You mean we should try to use a different slug for it?

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
Aljullu: Does it sound good to revisit this in a follow-up? I'm a bit wary of adding too many things to the global store because once extensions start hooking into it, they might start depending on these data being always in the global store. Instead, if we keep it in the child blocks/stores, it's easier for us to introduce changes because other blocks can't rely on its existence.

plugins/woocommerce/client/blocks/bin/webpack-helpers.js
Aljullu: I think that was a leftover from a previous version of this PR where we weren't importing the store directly. I removed it in 986b79d4a63a1cf5fb0f1e0cc16857697b527339.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/frontend.ts
dinhtungdu: Yeah, I comment here to not forget about it. It's better to visit this topic later. This PR is too big already.

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/price/frontend.ts
```diff
@@ -0,0 +1,83 @@
...
+		callbacks: {
+			updatePrice: () => {
+				const element = getElement();
+
+				if ( ! element.ref ) {
+					return;
+				}
+
+				const { priceHTML, originalPriceHtml } =
+					productPriceStore.state;
```
dinhtungdu: if we don't use this state in any directive, we can use it directly from the singleProductState without creating immediate derived state.

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/single-product.ts
dinhtungdu: Yeah, we're matching the block name with store namespace, so I think we need to use different namespace for global store. Btw, can we reuse `woocommerce` store?

dinhtungdu (approved):
LGTM! I left a minor comment but this is looking good already.

#### 🖼️ Screenshots/ screen recordings

https://github.com/user-attachments/assets/65db04cb-b670-41b6-9f9e-c7db4a45659a

#### 🌐 Site details

```
### WordPress Environment ###

WC Version: 10.0.0-dev
Log Directory Writable: ✔
WP Version: 6.8.1
WP Multisite: –
WP Memory Limit: 2 GB
WP Debug Mode: ✔
WP Cron: ✔
Language: en_US
External object cache: –

### Server Environment ###

Server Info: nginx/1.25.4
Server Architecture: Darwin 24.5.0 arm64
PHP Version: 7.4.33
PHP Post Max Size: 8 MB
PHP Time Limit: 30
PHP Max Input Vars: 1000
cURL Version: 8.5.0
OpenSSL/3.1.4

SUHOSIN Installed: –
MySQL Version: 8.0.33
Max Upload Size: 2 MB
Default Timezone is UTC: ✔
fsockopen/cURL: ✔
SoapClient: ✔
DOMDocument: ✔
GZip: ✔
Multibyte String: ✔
Remote Post: ✔
Remote Get: ✔

### Database ###

[REDACTED]

### Post Type Counts ###

attachment: 25
global_product_addon: 1
page: 18
post: 3
product: 19
product_tab: 1
product_variation: 7
revision: 224
shop_order_placehold: 1
wp_font_face: 36
wp_font_family: 12
wp_global_styles: 3
wp_navigation: 1
wp_template: 7
wp_template_part: 7

### Security ###

Secure connection (HTTPS): ✔
Hide errors from visitors: ❌Error messages should not be shown to visitors.

### Active Plugins (2) ###

Laravel DD for Wordpress: by Peter Hegman – 1.0.1
WooCommerce: by Automattic – 10.0.0-dev

### Inactive Plugins (15) ###

Akismet Anti-spam: Spam Protection: by Automattic - Anti-spam Team – 5.3.6
bbPress: by The bbPress Contributors – 2.6.13
Block Hooks Experiments for WooCommerce: by Automattic –
Custom Product Tabs Manager: by Addify – 1.2.2
Gutenberg: by Gutenberg Team – 20.3.0
Hello Dolly: by Matt Mullenweg – 1.7.2
Product Filters for WooCommerce: by WooCommerce – 1.4.38
What Can Wp Scripts Do: by The WordPress Contributors – 0.1.0
WooCommerce: by Automattic – 9.7.1 (update to version 9.9.3 is available)
WooCommerce PayPal Payments: by PayPal – 3.0.0 (update to version 3.0.6 is available)
WooCommerce Product Add-Ons: by Woo – 7.8.1 (update to version 7.9.0 is available)
WooPayments: by WooCommerce – 9.1.0 (update to version 9.5.0 is available)
WP BASIC Auth: by wokamoto – 1.1.4

### Must Use Plugins (1) ###

WooCommerce Product Details Extension Demo: by  – 1.0.0

### Settings ###

Legacy API Enabled: –
Force SSL: –
Currency: USD ($)
Thousand Separator: .
Decimal Separator: ,
Number of Decimals: 1
grouped (grouped)
simple (simple)
variable (variable)

featured (featured)
outofstock (outofstock)
rated-1 (rated-1)
rated-2 (rated-2)
rated-3 (rated-3)
rated-4 (rated-4)
rated-5 (rated-5)

HPOS feature enabled: ✔
HPOS data sync enabled: –
Enabled Features: analytics
marketplace
order_attribution
site_visibility_badge
remote_logging
email_improvements
point_of_sale
custom_order_tables


### Logging ###

Enabled: ✔
Retention period: 30 days
Level threshold: –
Log directory size: 4 MB

### WC Pages ###

Shop base: #7 - /shop/

### Theme ###

Name: Twenty Twenty-Five
Version: 1.2
Theme type: Block theme

### Templates ###

Overrides: /Users/tung/workspace/woocommerce/plugins/woocommerce/templates/block-notices/error.php
/Users/tung/workspace/woocommerce/plugins/woocommerce/templates/block-notices/notice.php
/Users/tung/workspace/woocommerce/plugins/woocommerce/templates/block-notices/success.php


### Admin ###

analytics
product-block-editor
coupons
core-profiler
customize-store
customer-effort-score-tracks
import-products-task
shipping-smart-defaults
shipping-setting-tour
homescreen
marketing
mobile-app-banner
onboarding
onboarding-tasks
product-custom-fields
remote-inbox-notifications
remote-free-extensions
payment-gateway-suggestions
printful
shipping-label-banner
subscriptions
store-alerts
transient-notices
woo-mobile-welcome
wc-pay-promotion
wc-pay-welcome-page
launch-your-store

experimental-blocks
experimental-iapi-mini-cart
experimental-iapi-runtime
minified-js
product-pre-publish-modal
settings
use-wp-horizon

Daily Cron: ✔ Next scheduled: 2025-06-14 10:10:56 +00:00
Options: ✔
Notes: 76
Onboarding: completed

### Action Scheduler ###

Complete: 107
Oldest: 2025-05-14 08:35:39 +0000
Newest: 2025-06-13 03:23:51 +0000

Failed: 2
Oldest: 2025-03-05 03:00:25 +0000
Newest: 2025-06-05 11:13:52 +0000

Pending: 2
Oldest: 2025-06-13 13:19:26 +0000
Newest: 2025-06-14 03:23:51 +0000


Generated at: 2025-06-13 10:58:50 +00:00
```
</details>

plugins/woocommerce/client/blocks/assets/js/atomic/blocks/product-elements/price/frontend.ts
Aljullu: Good call! Changed in https://github.com/woocommerce/woocommerce/pull/58384/commits/ff49f1c6ae7a0b152e81349f28cc5ba462f07001.

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/utils/price-currency.ts
```diff
@@ -165,7 +173,7 @@ export const formatPrice = (
 	) }${ applyDecimal(
 		afterDecimal,
 		decimalSeparator,
-		minorUnit
+		precision
 	) }${ suffix }`;
```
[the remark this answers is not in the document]
Aljullu: Changes in this file were removed.

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/utils/price-currency.ts
```diff
@@ -148,10 +148,18 @@ export const formatPrice = (
...
-	const { minorUnit, prefix, suffix, decimalSeparator, thousandSeparator } =
-		currency;
+	const {
+		minorUnit = 0,
+		precision,
+		prefix,
+		suffix,
+		decimalSeparator,
+		thousandSeparator,
+	} = currency;
```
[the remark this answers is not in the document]
Aljullu: Changes in this file were removed.

plugins/woocommerce/src/Blocks/BlockTypes/ProductPrice.php
```diff
@@ -70,10 +75,26 @@ protected function render( $attributes, $content, $block ) {
...
+
+			$wrapper_attributes = array(
+				'class' => 'wp-block-woocommerce-product-price',
+			);
+
+			if ( $is_interactive ) {
+				wp_enqueue_script_module( 'woocommerce/product-price' );
+				$wrapper_attributes['data-wp-interactive'] = 'woocommerce/product-price';
+				$wrapper_attributes['data-wp-watch']       = 'callbacks.updatePrice';
+				$this->initialize_shared_config( 'I acknowledge that using private APIs means my theme or plugin will inevitably break in the next version of WooCommerce' );
```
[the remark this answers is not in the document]
Aljullu: This could be removed: e797fde38ce29253f803fffd3eb3db25599587e5. :slightly_smiling_face:

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/product-data.ts
```diff
@@ -0,0 +1,72 @@
...
+
+				if ( context?.productData ) {
+					context.productData[ key ] = value;
+				} else if (
+					productDataStore?.state?.templateState?.productData
+				) {
+					productDataStore.state.templateState.productData[ key ] =
+						value;
+				}
+			},
```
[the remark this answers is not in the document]
Aljullu: I'm not sure about this one, I would leave things as they are until we have a clear understanding of how extensibility will be implemented.

plugins/woocommerce/src/Blocks/BlockTypes/SingleProduct.php
```diff
@@ -156,6 +161,43 @@ protected function replace_post_for_single_product_inner_block( $block, &$contex
...
+
+		if ( $html->next_tag( array( 'tag_name' => 'div' ) ) ) {
+			$html->set_attribute( 'data-wp-interactive', 'woocommerce/product-data' );
+			$html->set_attribute( 'data-wp-context', wp_json_encode( $interactivity_context, JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_AMP ) );
+		}
+
+		$content = $html->get_updated_html();
+
+		return $content;
+	}
```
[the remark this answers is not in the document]
Aljullu: Made some improvements to the render function in: 0cbbd7a6acf339ab41bb467b0be68de3992b1c81.

Aljullu:
Thanks for all the reviews, @dinhtungdu! :bow: 

> Yeah, we're matching the block name with store namespace, so I think we need to use different namespace for global store. Btw, can we reuse woocommerce store?

I renamed the store to `woocommerce/product-data` instead of `woocommerce/single-product`, this way it doesn't conflict with the block name. I'm not sure about reusing the `woocommerce` store, I feel that if we know we only need this store in the _Single Product_ block/template, it makes more sense to keep it separate from the `woocommerce` store that might end up being loaded everywhere. But not a strong opinion. In any case, it's something that can be changed in a follow-up.

dinhtungdu:
> I'm not sure about reusing the woocommerce store, I feel that if we know we only need this store in the Single Product block/template, it makes more sense to keep it separate from the woocommerce store that might end up being loaded everywhere. 

Luiz once shared with me on one of my P2 post (pdnLyh-7Ok-p2):

> The Interactivity API is in fact designed so a system like this can share a single store. The store function can be called multiple times from different files, and the objects get merged. This means that each block loaded on the frontend can add the parts it needs but can also use the parts loaded by other blocks.

You only inject the single product template data to the single product template only. On the other page, it's simple doesn't exist and returns undefined when retrieving.

> But not a strong opinion. In any case, it's something that can be changed in a follow-up.

Yes, let's not care about it for now.

plugins/woocommerce/src/Blocks/BlockTypes/SingleProduct.php
```diff
@@ -156,6 +161,43 @@ protected function replace_post_for_single_product_inner_block( $block, &$contex
...
+
+		$interactivity_context = array(
+			'originalProductData' => ProductDataUtils::get_product_data( $product ),
+			'productData'         => array(),
+		);
+
+		$html = new \WP_HTML_Tag_Processor( $content );
+
+		if ( $html->next_tag( array( 'tag_name' => 'div' ) ) ) {
+			$html->set_attribute( 'data-wp-interactive', 'woocommerce/product-data' );
```
dinhtungdu: I don't think we should set the namespace here to `woocommerc/product-data`, it's should be `woocommerce/single-product`.

dinhtungdu: The `woocommerce/product-data` is a global store that doesn't attach to any element. The `woocommerce/single-product` is a store stick to the Single Product block.

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/product-data.ts
```diff
@@ -0,0 +1,72 @@
...
+			key: keyof ProductData,
+			value: string | null
+		) => void;
+	};
+} >(
+	'woocommerce/product-data',
+	{
+		state: {
+			get productData(): ProductData {
+				const context = getContext< Context >();
```
dinhtungdu:
```suggestion
				const context = getContext< Context >( 'woocommerce/single-product' );
```

plugins/woocommerce/src/Blocks/BlockTypes/SingleProduct.php
dinhtungdu: It's not a hard rule, but I think we should distinguish between global store like `woocommerce` or `woocommerce/product-data` that only deal with data and local store attached to specific block that can have actions/callbacks. Imagine if later on, we want to add some actions to the Single Product block, what should we do?

Aljullu:
> Luiz once shared with me on one of my P2 post (pdnLyh-7Ok-p2):
> 
> > The Interactivity API is in fact designed so a system like this can share a single store. The store function can be called multiple times from different files, and the objects get merged. This means that each block loaded on the frontend can add the parts it needs but can also use the parts loaded by other blocks.
> 
> You only inject the single product template data to the single product template only. On the other page, it's simple doesn't exist and returns undefined when retrieving.

Hmmm, I would like to hear more about the benefits of this approach over having different stores, each one tailored for a specific use-case. The P2 suggests creating a single store for the _Product Filters_ blocks, but I think that's different from having everything in the same `woocommerce` global store.

I worry about two things:

* Working with one single global store with state and actions that can potentially not exist might be more difficult than working with smaller separate stores with clear and reliable states and actions.
* When unlocking the stores so extensions can start using them, it's going to be trickier if everything is in the same store than if we have more intentional individual stores, because it will force us to unlock everything at once.

I don't mind changing it in this PR, and we can always refactor it later. But I would delay this decision until we work on the extensibility APIs and can decide with a better understanding of the extension use cases. What do you think?

plugins/woocommerce/src/Blocks/BlockTypes/SingleProduct.php
Aljullu: Yes, it makes sense! I changed it in c6883be5293c620ce72fc50f810f258c700e35b2.

plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/product-data.ts
Aljullu: Same: c6883be5293c620ce72fc50f810f258c700e35b2. :slightly_smiling_face:

dinhtungdu:
> I don't mind changing it in this PR, and we can always refactor it later. But I would delay this decision until we work on the extensibility APIs and can decide with a better understanding of the extension use cases. What do you think?

As I mentioned above, I agree to do it later, the PR is already big and complex. And that's something we can change without affecting existing stores.

dinhtungdu (approved):
I can confirm it still works as expected after addressing feedback.

helgatheviking:
@gigitux this is :fire:. What would I need to listen for  to have my own custom block update on the variation change? Is the new [`ProductPriceStore` ](https://github.com/woocommerce/woocommerce/pull/58384/files#diff-75a5656079a63979ca5241d104fd81fef01ebb603651109a28aa4f7a3ecdde25R42)watching the `ProductDataStore`?

PR #58391, opened 2025-05-29
Remove unused woocommerce_product_availability filter

Aljullu:

As @jimjasson mentioned in https://github.com/woocommerce/woocommerce/issues/58125, we have a filter in the codebase which is useless: `woocommerce_product_availability`. While it has been improved in #58185 (shipping in WC 9.9), my suggestion is to remove the filter until we see there is demand and valid use cases for it. Otherwise, we will need to support it forever.

This PR will require a CFE for WooCommerce 9.9.

1. Add the _Product Stock Indicator_ somewhere and verify it still correctly shows the availability text for products which are out of stock or in backorder.

### DIFF
--- a/plugins/woocommerce/src/Blocks/Utils/ProductAvailabilityUtils.php
+++ b/plugins/woocommerce/src/Blocks/Utils/ProductAvailabilityUtils.php
@@ -36,13 +36,6 @@ class ProductAvailabilityUtils {
 			$product_availability = $product->get_availability();
 		}
 
-		/**
-		 * Filters the product availability information.
-		 *
-		 * @since 9.7.0
-		 * @param array $product_availability The product availability information.
-		 * @param \WC_Product $product Product object.
-		 */
-		return apply_filters( 'woocommerce_product_availability', $product_availability, $product );
+		return $product_availability;
 	}
 }
--- a/plugins/woocommerce/changelog/fix-remove-woocommerce_product_availability-filter
+++ b/plugins/woocommerce/changelog/fix-remove-woocommerce_product_availability-filter
@@ -1,0 +1,4 @@
+Significance: minor
+Type: update
+
+Remove unused woocommerce_product_availability filter


dinhtungdu (approved):
As 9.9 is not release yet, so this change is backward compatible, so I'm approving. :shipit:

PanosSynetos (commented):
I don't think that anyone would have used this filter, as it didn't have a `$product` argument, before my fix.

I'm fine with removing it in 9.9 - but we would need to "hurry" as the release is on Monday

PanosSynetos (approved):

Aljullu:
Rebased this branch to point to `release/9.9`.

Aljullu:
CFE created: https://github.com/woocommerce/woocommerce/issues/58408.

plugins/woocommerce/src/Blocks/Utils/ProductAvailabilityUtils.php
```diff
@@ -36,13 +36,6 @@ public static function get_product_availability( $product ) {
...
 		}
 
-		/**
-		 * Filters the product availability information.
-		 *
-		 * @since 9.7.0
-		 * @param array $product_availability The product availability information.
-		 * @param \WC_Product $product Product object.
-		 */
-		return apply_filters( 'woocommerce_product_availability', $product_availability, $product );
```
kalessil, line 46: It appears that we didn't have any deprecation notices before this removal, and just dropped the filter. Out of curiosity, are there any criteria in core development when we skip the deprecation step?

Aljullu: The rationale is that between WC 9.7 (when it was introduced) and WC 9.8 (last released version) the filter looked like this:

```PHP
return apply_filters( 'woocommerce_product_availability', $product_availability );
```

Given that the product context wasn't passed to the filter, the filter didn't solve any real world use case and it was, in fact, useless and introduced by mistake. It was also never documented and never promoted publicly. That's why I think it should be safe to remove.

kalessil: Gotcha, thanks for explaining!

PR #58402, opened 2025-05-29
Adds tests to verify error notices when adding out-of-stock products

### LINKED ISSUE #57318: Add automated tests that verify Product Collection errors notices are displayed correctly
In [https://github.com/woocommerce/woocommerce/issues/57193](https://github.com/woocommerce/woocommerce/issues/57193) we introduced a regression that caused Product Collection error notices not to be attached to the page, so they weren't visible to shoppers. This issue was luckily found before it was released, but we should have automated tests that catch it.

This task is about creating automated tests that verify error notices get displayed when adding to cart from the Product Collection block causes an error.

BurakParsAydin:

Closes #57318  .

1. Run the following command:
`pnpm --filter=@woocommerce/block-library test:e2e tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts --debug`

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 
Add E2E test to verify error notice displays correctly when adding out-of-stock products from Product Collection
</details>

## Summary by CodeRabbit

- **Tests**
  - Added an end-to-end test to ensure users see an error notice when trying to add out-of-stock products from a Product Collection in WooCommerce.


BurakParsAydin:
I'm getting the following error when running the command. Could I get some help with this?: 

Screenshot from 2025-05-30 22-02-11

Aljullu (commented):
Thanks for working on this, @BurakParsAydin! Some things that might help:

1. The test file name should end in `.block_theme.spec.ts` (notice the `block_theme` part), this tells the e2e test tooling to run the test in a block theme. It's also possible to use `classic_theme` or `block_theme_with_templates` when needed, but in generic cases like this one, `block_theme` is the correct one to use.
2. I would avoid creating any new category during the tests, as this can slow down the test run significatively.
3. Similarly, I would avoid creating a new product and instead would try to use one of the existing ones. The state is reset between tests, so you can do any edits you need to them without worrying to affect follow-up tests.
4. As a tip, you can add `--debug` to the command you use to run tests and that will open a browser window. That's very helpful because you can see the test live and you can also open new tabs and, with that, navigate the site, see the categories, products, etc.

Hope that helps! :slightly_smiling_face:

BurakParsAydin:
@Aljullu Thanks for your suggestion above 🙂. It really helped me simplify the test flow.  I've adjusted the test. Apparently, there was no need to add so much unnecessary code.
For the test, I only got stuck on how to capture the error when adding the product. Since it’s an out-of-stock product, there’s no "Add to Cart" button, so I can’t add it to the cart. Could you help me with this part?

Aljullu (commented):
Thanks for updating this, @BurakParsAydin!

> For the test, I only got stuck on how to capture the error when adding the product. Since it’s an out-of-stock product, there’s no "Add to Cart" button, so I can’t add it to the cart. Could you help me with this part?

You're right. I initially thought testing with an _Out of stock_ product was the easiest approach. That changes the _Add to Cart_ button with a _Read more_ link as you say. So instead of using an _Out of stock_ product, it might be better to test with a product sold individually. After pressing the _Add to Cart_ button twice, an error should be displayed.

Aljullu:
Btw, to set a product as sold individually, it might be good to do that from the[ `products.sh` script](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/tests/e2e/bin/scripts/products.sh) that is used to create the products for tests. It will be more performant and it will impact all tests. This way, if we ever have another test where we need a product sold individually, we could benefit from the changes in this PR.

It would be similar to how it's done for the out-of-stock product:

https://github.com/woocommerce/woocommerce/blob/501974d86bbe813f8afc00c965e4cb02f7a92144/plugins/woocommerce/client/blocks/tests/e2e/bin/scripts/products.sh#L34-L36

How does it sound?

BurakParsAydin:
@Aljullu Thank you for your help. It really helped me implement the test.
Good news: I've mostly completed the test and pushed the commit.

The only issue I’m facing is with the final step, which refers to [this](https://github.com/woocommerce/woocommerce/blob/3a7ef3359026e270aebbfbedaeb4c09d3ba97f89/plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts#L80-L83).

The script can't find `.wc-block-components-notice-banner.is-error` because the banner only appears inside the sidebar that opens when clicking the cart (bag) icon. If you get a chance to run the test, you’ll see what I mean.

Do you have any suggestions on how I could approach this part?

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+test.describe( 'Product Page: error notices when adding out-of-stock products', () => {
+	test( 'displays error notice when attempting to add product beyond stock limit', async ( {
+		admin,
+		editor,
+		blockUtils,
+		page,
+	} ) => {
+		await blockUtils.createManagedStockProduct();
+		await admin.createNewPost();
+		await editor.insertBlock( { name: 'woocommerce/single-product' } );
```
Aljullu: We should be testing the _Product Collection_ block (`woocommerce/product-collection`) instead of the _Single Product_ block, here.

Aljullu (commented):
I think the issue is that you were adding the wrong block to the page. You should test the _Product Collection_ block, not the _Single Product_ block. :slightly_smiling_face:  The latter doesn't have error notices itself, so it's expected they won't show up.

BurakParsAydin:
@Aljullu Thanks for your review 🙂 I've added the Store Notice block - please refer to my [latest commit](https://github.com/woocommerce/woocommerce/pull/58402/commits/ad0bc378cb0a65efba499cb0dfb861e3c4def9b4), which addresses the issue. 
The test is now running well on my end 🚀 Looking forward to your feedback

Aljullu:
Thanks for the updates, @BurakParsAydin! Not sure if you saw my [last comment](https://github.com/woocommerce/woocommerce/pull/58402#pullrequestreview-2909423078), but the issue is that we need to test the _Product Collection_ block (`woocommerce/product-collection`), not the _Single Product_ block (`woocommerce/single-product`). Your test [is adding the _Single Product_ block to the page](https://github.com/woocommerce/woocommerce/pull/58402/files#diff-205ca2b430ba5da692d90de898dd88731849a890f322ddd936dd3f85e615cafdR67) and testing it, instead of testing the _Product Collection_ block.

https://github.com/woocommerce/woocommerce/issues/57318 is about adding the test for the _Product Collection_ block, so this PR needs to be updated so the block that is added in the test is the _Product Collection_. Hope that makes sense!

BurakParsAydin:
@Aljullu That was a helpful insight. I appreciate it. FYI, I adjusted the code. 
Waiting for your feedback.

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+		await editor.publishAndVisitPost();
+
+		const productCart = page
+			.locator( '.wc-block-product' )
+			.filter( { hasText: productName } );
+
+		await expect( productCart ).toBeVisible();
+
+		// Add to cart once — succeeds.
+		await productCart.locator( '.add_to_cart_button' ).click();
```
Aljullu: It's usually a good practice to use [`.getByText()`, `.getByRole()`, etc. locators](https://playwright.dev/docs/locators) instead of using class names. In this case, could we use `.getByText( 'Add to cart' )`?

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+
+		// Add to cart again — triggers out-of-stock error.
+		await productCart.locator( '.add_to_cart_button' ).click();
+
+		// Verify error notice is displayed.
+		await expect(
+			page.locator( '.wc-block-components-notice-banner.is-error' )
```
Aljullu: Similarly, here it would be better to check for the error text message (or part of it) instead of the element class name.

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+
+		await singleProduct
+			.locator(
+				'.wc-blocks-product-collection__collections-create button'
+			)
+			.click();
+
+		const productName = 'A Managed Stock';
+
+		await editor.publishAndVisitPost();
```
Aljullu: I wonder if we could replace all of this with the `pageObject` utils:

```JS
			await pageObject.createNewPostAndInsertBlock();
			await pageObject.publishAndGoToFrontend();
```

Like it's done in [this test](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection.block_theme.spec.ts#L30), for example.

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+	constructor( { editor, page }: { editor: Editor; page: Page } ) {
+		this.editor = editor;
+		this.page = page;
+	}
+
+	async createManagedStockProduct() {
+		await wpCLI(
+			'wc product create --name="A Managed Stock" --regular_price=10 --manage_stock=true --stock_quantity=1 --user=admin'
+		);
+	}
```
Aljullu: Given that this method is only used in this test, I wonder if we should simply have it inline in the test. That would help simplify the code.

Aljullu (commented):
Thanks for the updates, @BurakParsAydin! The test is passing now and it's correctly testing that the error notices are visible in the _Product Collection_ block. :raised_hands: 

I left a few inline comments with some suggestions which I think would help make the test simpler and more reliable.

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts
BurakParsAydin: @Aljullu I'm thinking to target this via `getByRole()`. What paremeter should I use? `dialog` doesn't work

Aljullu: The role is `alert`, if I'm not wrong. :slightly_smiling_face: 

In this case, it might also be good to check that it contains an error message, not only that it's visible. Something like this, maybe?

```JS
await expect( page.getByRole( 'alert' ) ).toBeVisible();
await expect( page.getByRole( 'alert' ) ).toHaveText(
	/maximum quantity/i
);
```

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+		await editor.publishAndVisitPost();
+
+		const productCart = page
+			.locator( '.wc-block-product' )
+			.filter( { hasText: productName } );
+
+		await expect( productCart ).toBeVisible();
+
+		// Add to cart once — succeeds.
+		await productCart.getByRole( 'button' ).click();
```
Aljullu: I think it will be more reliable if we make sure the button has the `Add to Cart` text:

```suggestion
		await productCart.getByRole( 'button', { name: 'Add to cart' } ).click();
```

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts
BurakParsAydin: @Aljullu FYI, I've adjusted the test as per your request. There's just **this part** left I keep getting errors here. If possible, could you make a commit suggestion for this part?

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+test.describe( 'Product Page: error notices when adding out-of-stock products', () => {
+	test( 'displays error notice when attempting to add product beyond stock limit', async ( {
+		admin,
+		editor,
+		page,
+	} ) => {
+		const productName = 'A Managed Stock';
+
+		wpCLI(
```
Aljullu: Heads-up that `wpCLI()` needs to be awaited:

```suggestion
		await wpCLI(
```

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+			'woocommerce/product-collection'
+		);
+
+		await expect(
+			singleProduct.getByText( 'create your own' )
+		).toBeVisible();
+
+		await singleProduct.getByText( 'create your own' ).click();
+
+		await editor.publishAndVisitPost();
```
Aljullu: You can replace this with:

```suggestion
		await pageObject.createNewPostAndInsertBlock();
		await pageObject.publishAndGoToFrontend();
```

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+import { Page } from '@playwright/test';
+import { test, expect, Editor, wpCLI } from '@woocommerce/e2e-utils';
+
+test.describe( 'Product Page: error notices when adding out-of-stock products', () => {
+	test( 'displays error notice when attempting to add product beyond stock limit', async ( {
+		admin,
+		editor,
+		page,
+	} ) => {
```
Aljullu: Then you must add the `pageObject` here:

```suggestion
const test = base.extend< { pageObject: ProductCollectionPage } >( {
	pageObject: async ( { page, admin, editor }, use ) => {
		const pageObject = new ProductCollectionPage( {
			page,
			admin,
			editor,
		} );
		await use( pageObject );
	},
} );

test.describe( 'Product Page: error notices when adding out-of-stock products', () => {
	test( 'displays error notice when attempting to add product beyond stock limit', async ( {
		page,
		pageObject,
	} ) => {
```

You will also need to add the import at the top of the file:

```TS

/**
 * Internal dependencies
 */
import ProductCollectionPage from './product-collection.page';
```

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+		await expect( productCart ).toBeVisible();
+
+		// Add to cart once — succeeds.
+		await productCart
+			.getByRole( 'button', { name: 'Add to cart' } )
+			.click();
+
+		// Add to cart again — triggers out-of-stock error.
+		await productCart.getByRole( 'button' ).click();
```
Aljullu: I would update this line that the one above, making sure the `button` has the `Add to cart` text.

Aljullu (commented):
Thanks for the updates, @BurakParsAydin, I left a few more comments, but it's shaping up good. :100:

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts
Aljullu: I left the comments inline. Hope that helps. :slightly_smiling_face:

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts
BurakParsAydin: @Aljullu Since the button text changes after the first click and says "1 in cart," I thought it might be better to leave it as is. Do you have any recommendations at this point?

BurakParsAydin:
@Aljullu FYI, I’ve adjusted the test as you suggested. The only part left is the one related to the add-to-cart event you mentioned above. Looking forward to your feedback 🙂

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts
Aljullu: Ah, my bad, you're right. In that case, could we add `{ name: '1 in cart' }`?

What I don't like about simply using `getByRole( 'button' )` is that if any other button is added to the Product Collection block in the future, the test might fail because it could try to click on the wrong button. For example, imagine in the future an amplifier button is added to the product image, this test might fail because it won't know which button it needs to click. So, in general, it's better to be as specific as possible with this kind of selectors. Do you agree?

BurakParsAydin: @Aljullu Thanks for the recommendation. I’ve updated that line, but I ended up targeting the button a bit differently. Hope that’s okay with you.

It seems` { name: '1 in cart' } `doesn’t work, since the button’s accessible name is actually "Add to cart."

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+		const productCart = page
+			.locator( '.wc-block-product' )
+			.filter( { hasText: productName } );
+
+		await expect( productCart ).toBeVisible();
+
+		// Add to cart once — succeeds.
+		await productCart
+			.getByRole( 'button', { name: 'Add to cart' } )
+			.click();
```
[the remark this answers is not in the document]
Aljullu: @BurakParsAydin I think this is a good suggestion. Do you think we can add it to the test?

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+		const pageObject = new ProductCollectionPage( {
+			page,
+			admin,
+			editor,
+		} );
+		await use( pageObject );
+	},
+} );
+
+test.describe( 'Product Page: error notices when adding out-of-stock products', () => {
```
Aljullu: I would simplify this `describe`, in the test itself we already describe what the test is doing:

```suggestion
test.describe( 'Product Page: error notices', () => {
```

Aljullu (commented):
Thanks for the updates, @BurakParsAydin! I left a couple of last comments, but overall everything is looking good.

BurakParsAydin:
@Aljullu Thanks! FYI, I've adjusted the test as per your request. Waiting for your feedback.

plugins/woocommerce/client/blocks/tests/e2e/tests/product-collection/product-collection-errors.block_theme.spec.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * External dependencies
+ */
+import { Page } from '@playwright/test';
+import { test as base, expect, Editor, wpCLI } from '@woocommerce/e2e-utils';
```
Aljullu: `Page` and `Editor` aren't used in this test, so they shouldn't be imported. :slightly_smiling_face:

Aljullu (commented):
Thanks, @BurakParsAydin! Everything looks good to me. Apparently, there is only one issue remaining related to linting. I commented it inline. :point_down:

BurakParsAydin:
@Aljullu Thanks for the review  🙂 I've adjusted it as you addressed. 
I also renamed the constant name for clarity, removed await - it doesn’t need to be awaited if `use` itself is synchronous internally I think.

If there’s a related task you think I can help with, feel free to let me know - I'd be happy to jump in!

Apart from that, just wanted to say a big thank you for all the guidance and support you gave me on this task 🙏🏼
Your detailed feedback and clear direction really helped me understand things better and move forward with confidence. I truly appreciate the time and effort you put into reviewing and mentoring - means a lot! ❤️

Aljullu (commented):
> removed await - it doesn’t need to be awaited if `use` itself is synchronous internally I think.

Apparently lint is complaining about that, so I think we need to add that `await` again. :smile: 

> Apart from that, just wanted to say a big thank you for all the guidance and support you gave me on this task 🙏🏼
> Your detailed feedback and clear direction really helped me understand things better and move forward with confidence. I truly appreciate the time and effort you put into reviewing and mentoring - means a lot! ❤️

Not at all, thanks a lot for your help and contribution to WooCommerce! :raised_hands:

BurakParsAydin:
@Aljullu FYI, I've added `await`

Aljullu (approved):
Awesome, thanks for all your work, @BurakParsAydin! This is looking good now, merging it into `trunk`! :partying_face:

Issue #35022, opened 2022-10-10
[COT/HPOS] Investigate/Add adding `post_author` field to core DB schema.


vedanshujain:
Some extension store data in the post_author field; we currently don't migrate it. This issue scopes investigating whether that's needed and adding the field if it is.

coreymckrill:
@vedanshujain do you know a specific example of an extension that uses the `post_author` field for orders?

coreymckrill:
I searched the (unzipped) all plugins repo using the following:

```
grep -nR -C 2 post_author ./ | grep -E -e "\.php[-+:]+[0-9]+[-+:]+.*order\b" | sort
```

There were a few results, but all were false positives. So it seems unclear that there's actually a use case for the `post_author` field. Perhaps we can deprioritize this unless an extension developer comes forward with a documented use case? @vedanshujain

vedanshujain:
Yep, makes sense, thanks for doing the analysis, going to mark it as low prio.

rrennick:
Here is a back reference to an [issue encountered with post_author](https://github.com/woocommerce/woocommerce/issues/21656). I thought of this because some of the larger shops looking to migrate to HPOS could be 5+ years old.

cc @vedanshujain @coreymckrill

coreymckrill:
Thanks @rrennick! Do you think this means that we should, in fact, prioritize adding an `author` field to the wc_orders table? My read of that issue is that we aren't using `post_author` anymore, and presumably any current shops with orders older than that would have gone through upgrade routines by now. Plus my grepping efforts (above) didn't turn up anything in WC Subscriptions. But I could be misinterpreting what that issue is really saying.

rrennick:
> Do you think this means that we should, in fact, prioritize adding an author field to the wc_orders table?

@coreymckrill I didn't go back and do a full refresh on the issue. I was mainly bringing it to attention here on the chance that an issue crops up over the next few months.

ZILZAL:
This would be useful when giving specific permissions to shop managers to manage and see only thier own created orders.
i came across this thread when i was searching for this use.

jorgeatorres:
Going to close this one as this hasn't come up again or at any point. We do have an open issue related to managing orders based on role (#39289), which possibly captures the above request.

PR #58423, opened 2025-05-30
HPOS: allow users with `edit_shop_orders` permission to access to the orders screen

### LINKED ISSUE #52389: [Enhancement]: Read only access to orders after HPOS

Hi there,

I have a shop manager who should see orders but can't edit them via admin panel.
It was possible before with this plugin.
But after HPOS move, when I give Read access to a user, he can't see orders page at all.

Below tutorials doesn't work now.
https://youtu.be/fFoST7We0wY?si=JHnN3fwrBAFDQc9k&t=205
https://publishpress.com/blog/capabilities/allow-users-to-view-and-not-edit-woocommerce-orders/

I tried 3 different plugin and lot's of custom codes.
All have same results.

when I give Read access to a user, he can't see orders page at all.

What we and these plugin developers need is to have READ only access to orders.

jorgeatorres:

This PR addresses a permission issue where users with `edit_shop_orders` capability were unable to access the Orders screen when HPOS is enabled.

While this might be a bit of an uncommon scenario it was supported in the legacy implementation (posts) and this change brings HPOS in line.

Closes #52389.

1. Create a new role (named "order viewer") with `edit_shop_orders` capability using WP-CLI:
   ```
   wp role create orderviewer "order viewer"
   wp cap add orderviewer edit_shop_orders read view_admin_dashboard
   ```
1. Create a test user (username "orderviewer") on your site with the "order viewer" role:
   ```
   wp user create orderviewer orderviewer@woo.test --user_pass=password
   wp user add-role orderviewer orderviewer
   ```
1. Log in as `orderviewer` on the backend (`/wp-admin`) in an incognito window and keep an admin session in another browser window.
1. As admin, go to WC > Settings > Advanced > Features and ensure that `WordPress posts storage (legacy)` is set as datastore. Save changes if needed.
   Sync orders with `wp wc hpos sync` if needed to enable the option.
1. **Confirm legacy behavior with HPOS disabled:**
   1. Switch to your `orderviewer` session.
   1. Confirm that you don't see the **WooCommerce** menu but do see an **Orders** menu.
      Screenshot 2025-09-12 at 15 55 51
   1. Clicking the **Orders** menu should give you a list of all orders.
   1. You should not be able to edit or make changes to any order.
1. As admin, go to WC > Settings > Advanced > Features and ensure that ` High-performance order storage` is set as datastore. Save changes if needed.
1. **Confirm HPOS supports the legacy behavior:** Repeat the steps from step 5 (_Confirm legacy behavior with HPOS disabled_) above. Results should be identical.
1. Check out `trunk` and confirm that the above step doesn't work. That is, you can't access the _Orders_ menu as `orderviewer` when HPOS is enabled.
1. Remove the `edit_shop_orders` capability from the role (`wp cap remove orderviewer edit_shop_orders`) and confirm that you no longer see the **Orders** menu when logged in as `orderviewer`.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/src/Internal/Admin/Orders/PageController.php
```diff
@@ -261,7 +262,7 @@ public function register_menu(): void {
 			$post_type = get_post_type_object( $order_type );
 
 			add_submenu_page(
-				'woocommerce',
+				\WC_Admin_Menus::can_view_woocommerce_menu_item() ? 'woocommerce' : null,
```
joshuatf: I think we need to avoid registration of this item instead of passing `null` as this causes a number of deprecation warnings:
Screenshot 2025-09-15 at 10 10 08 AM

plugins/woocommerce/src/Internal/Admin/Orders/PostsRedirectionController.php
```diff
@@ -184,5 +192,34 @@ private function maybe_redirect_to_edit_order_page(): void {
 		exit;
 	}
 
+	/**
+	 * Updates legacy menu items to point to the new Orders page.
```
joshuatf: Nitpick: Could we be a bit more descriptive with this text and the function name?

It seems like this only applies when the main WC menu item is not present and this could reflect that.

plugins/woocommerce/includes/admin/list-tables/class-wc-admin-list-table-orders.php
```diff
@@ -376,6 +376,11 @@ public static function get_order_preview_actions_html( $order ) {
 		$actions        = array();
 		$status_actions = array();
 
+		$wp_post_type = get_post_type_object( $order->get_type() ) ?? get_post_type_object( 'shop_order' );
+		if ( ! current_user_can( $wp_post_type->cap->edit_post, $order->get_id() ) ) {
+			return;
```
joshuatf: This should probably return an empty string to keep consistent with the return type of this method.

joshuatf (commented):
Testing pretty well!  Just a few minor nitpicks and deprecation issues to address.

Also, just checking if this is expected (same behavior on `trunk`), but with this newly created user we can't access individual orders:
Screenshot 2025-09-15 at 10 57 12 AM

plugins/woocommerce/includes/admin/list-tables/class-wc-admin-list-table-orders.php
jorgeatorres: Addressed in d575f486f448bad58255ee98a10d7e17ead8117c.

plugins/woocommerce/src/Internal/Admin/Orders/PageController.php
jorgeatorres: Fixed with 937ff611553b1929f8ba56674c78265da9fded70.

plugins/woocommerce/src/Internal/Admin/Orders/PostsRedirectionController.php
jorgeatorres: Rewrote the docblock a bit in 3b419999832a5582232eed6cc0049dd6c1b025be, but let me know if you think further clarification is needed.

jorgeatorres:
Hey @joshuatf!

Thanks for the review. I've addressed your feedback. Please take another look when you have the chance!

As for your question:

> Also, just checking if this is expected (same behavior on trunk), but with this newly created user we can't access individual orders:

Yes, that's expected and in line with the CPT behavior.

plugins/woocommerce/src/Internal/Admin/Orders/PageController.php
```diff
@@ -261,7 +262,7 @@ public function register_menu(): void {
 			$post_type = get_post_type_object( $order_type );
 
 			add_submenu_page(
-				'woocommerce',
+				\WC_Admin_Menus::can_view_woocommerce_menu_item() ? 'woocommerce' : '',
```
joshuatf: I think the `''` is causing this warning:

```
Deprecated: strip_tags(): Passing null to parameter #1 ($string) of type string is deprecated in /Users/joshua/Sites/merchant/app/public/wp-admin/admin-header.php on line 41
```

I think instead we can conditionally use `add_submenu_page` or `add_menu_page` when `can_view_woocommerce_menu_item()` is false.

joshuatf (commented):
Found one more warning to take care of during review.

Also, optional and a little edge case but nice to have for bookmarks would be to redirect the post type screen to the `wc-orders` screen when we change storage methods.

`/wp-admin/edit.php?post_type=shop_order` -> `/wp-admin/admin.php?page=wc-orders`

To reproduce:

1. Change to legacy storage via admin role
2. Visit the orders page using `orderviewer` role in an incognito window
3. Change the storage to HPOS via admin role
4. Refresh the orders page
5. Note the error and page URL

Screenshot 2025-09-24 at 9 19 44 AM

plugins/woocommerce/src/Internal/Admin/Orders/PageController.php
jorgeatorres: Yes, for some reason I wanted to avoid this as I didn't want the new code to be "in charge" of a full top level menu. I'm now using `admin.php` (see bd794e54f4e42faf20f532851ae61be9c988bc91) which should be valid (according to the codex) and produces no warning.

jorgeatorres:
@joshuatf: Thanks for the latest round of review. Hopefully third's a charm. I've changed the code to use `admin.php` as parent page (which also matches the `admin_page_wc-orders` slug used elsewhere). The CPT > HPOS redirection should also be working now (77b902e5f8ee6fadbf798296fe4c3bfc52b67165).

joshuatf (approved):
Pre-approving these changes as the following are testing well:

* ✅ Redirection works as expected for both order types when viewing the opposite orders route that is enabled
* ✅ The order viewer user can access the orders screen, regardless of what storage type is configured
* 🛑  I ran into one more issue (sorry I didn't see this before!).  When HPOS is enabled we're missing the "Add Order" option for the order viewer role, which shows up when we're using legacy order storage.  Pre-approving this PR, but we can either update here or ship this and fix in a follow-up.

Screenshot 2025-09-25 at 2 20 06 PM

jorgeatorres:
> * I ran into one more issue (sorry I didn't see this before!).  When HPOS is enabled we're missing the "Add Order" option for the order viewer role, which shows up when we're using legacy order storage.  Pre-approving this PR, but we can either update here or ship this and fix in a follow-up.

Under HPOS, I believe it is best to hide this menu item as it wouldn't work anyways. This is because in legacy mode, where orders are also posts, certain users are allowed to create orders and then manage those orders they created (this regardless of the customer associated with the order). Currently, HPOS doesn't have support for the concept of author, so we wouldn't be able to offer this functionality, which is kind of an edge case.

We considered this in the past (see https://github.com/woocommerce/woocommerce/issues/35022) but decided against adding it as it doesn't seem a very common/popular use case.

PR #58440, opened 2025-05-31
Add `Products by Category` and `Products by Tag` Collections to Product Collection Block

### LINKED ISSUE #54631: Product Collection: Add "Products by Category" Collection
## Background
With the transition to the Product Collection block in WooCommerce 9.5, users have requested a simpler way to display products by category, similar to the previous "Products by Category" block functionality. While this can currently be achieved through custom/"create your own" collection, having a dedicated collection would improve the user experience.

More context: https://github.com/woocommerce/woocommerce/issues/54394

## Proposed Solution
Add a new collection called "Products by Category" to the Product Collection block similar to "Hand-Picked Products" collection. This new collection will have control at top of sidebar to select categories, similar to "Hand-Picked Products" collection.

## Benefits
- Improved user experience for this common use case
- Reduces steps needed to display category-based product listings

amitraj2203:

This PR introduces two new collections to the `Product Collection` block: `Products by Category` and `Products by Tag`.

#### Key Changes:

- Added new `by-category.tsx` and `by-tag.tsx` collection files in `client/blocks/assets/js/blocks/product-collection/collections/`
Modified `taxonomy-controls/index.tsx` to lift the controls for both collections to the top and hide the other collection settings in the `Filters Panel`
- Removed `by category` and `by tag` keyword from existing Product Collection block.json

Closes #54631 

<details>
  <summary><strong>Click here to view</strong></summary>

  #### Picker
Picker

 #### Initial (empty) state - Category
Initial (empty) state - Category

 #### Category Selected
Category Selected

 #### Initial (empty) state - Tag
Initial (empty) state - Tag  

 #### Tag Selected
Tag Selected

</details>

1. Create a new post.
2. Add the `Product Collection` block and select the `Products by Category` or `Products by Tag` collection.
3. Ensure the initial/empty state of the block displays `No products to display. Try adjusting the filters in the block settings panel.`
4. Use the top-most control in the Inspector Controls to choose categories/tags to display products.
5. Save the post without selecting any category or tag and ensure the empty block is not rendered on the front end.
6. Add some categories or tags in random order via the control.
7. Ensure the products associated with the categories or tags are preserved in both the editor and front end.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/collections/by-category.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const collection = {
+	name: CoreCollectionNames.BY_CATEGORY,
+	title: __( 'Products by Category', 'woocommerce' ),
+	icon: <Icon icon={ category } />,
+	description: __(
+		'Display products from specific categories.',
+		'woocommerce'
+	),
+	keywords: [ 'category', 'products' ],
```
kmanijak: Minor but these keywords can be omitted as they are literally part of the collection title which is taken into account when typing in the inserter.

```suggestion
```

Applies to Products by Tag as well.

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/collections/by-category.tsx
```diff
@@ -0,0 +1,71 @@
...
+			justifyContent: 'center',
+		},
+	},
+];
+
+const innerBlocks: InnerBlockTemplate[] = [
+	heading,
+	INNER_BLOCKS_PRODUCT_TEMPLATE,
+	pagination,
+];
```
kmanijak: This could be simplified a bit by importing `INNER_BLOCKS_PAGINATION_TEMPLATE` from `'../constants'` and then:

```suggestion

const innerBlocks: InnerBlockTemplate[] = [
	heading,
	INNER_BLOCKS_PRODUCT_TEMPLATE,
	INNER_BLOCKS_PAGINATION_TEMPLATE,
];
```

Applies to Products by Tag as well.

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/edit/inspector-controls/index.tsx
```diff
@@ -318,14 +323,18 @@ const CollectionSpecificControls = (
...
 	return (
 		<InspectorControls>
 			{
 				/**
-				 * Hand-Picked collection-specific controls.
+				 * "Hand-Picked" collection-specific controls.
 				 */
 				props.attributes.collection ===
-					'woocommerce/product-collection/hand-picked' && (
+					CoreCollectionNames.HAND_PICKED && (
```
kmanijak: Thank you for attention to details and fixing minor stuff as you go through the code! 👏

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/edit/inspector-controls/index.tsx
```diff
@@ -337,11 +346,23 @@ const CollectionSpecificControls = (
...
 			}
+			{
+				/**
+				 * "Category and Tag" collection-specific controls.
+				 */
+				isByCategoryOrTag && (
+					<PanelBody>
+						<TaxonomyControlsField
+							{ ...queryControlProps }
+							collection={ props.attributes.collection }
```
kmanijak: Nitpick but what do you think about destructuring `collection` in upper scope and reuse it inside `CollectionSpecificControls`? It seems like `props.attributes.collection` is repeated many times while it could be:

```
const { collection } = props.attributes;
```

and used in the conditions above or in here. Super nitpick, feel free to ignore.

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/edit/inspector-controls/taxonomy-controls/index.tsx
```diff
@@ -36,35 +40,61 @@ export const useTaxonomies = (): Taxonomy[] => {
...
+		// For filters panel, show the complementary taxonomy.
+		if ( collection === CoreCollectionNames.BY_CATEGORY ) {
+			return taxonomies.filter(
+				( taxonomy ) => taxonomy.slug === 'product_tag'
+			);
+		}
+		if ( collection === CoreCollectionNames.BY_TAG ) {
+			return taxonomies.filter(
+				( taxonomy ) => taxonomy.slug === 'product_cat'
+			);
```
kmanijak: This logic has to be reverted from filtering to excluding.

Tags and Categories are taxonomies added by WooCommerce Core but there may be more taxonomies like brands etc. and they're all handled generically in here. That means, this would ignore other potential taxonomies and just left one. Should be something like: 

```suggestion
		// For filters panel, show the complementary taxonomy.
		if ( collection === CoreCollectionNames.BY_CATEGORY ) {
			return taxonomies.filter(
				( taxonomy ) => taxonomy.slug !== 'product_cat'
			);
		}
		if ( collection === CoreCollectionNames.BY_TAG ) {
			return taxonomies.filter(
				( taxonomy ) => taxonomy.slug !== 'product_tag'
			);
```

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/edit/inspector-controls/taxonomy-controls/index.tsx
```diff
@@ -107,4 +137,70 @@ function TaxonomyControls( {
 	);
 }
 
+export function TaxonomyControlsField( {
```
kmanijak: What's the difference between `TaxonomyControls` and `TaxonomyControlsField`? Feels identical and redundant 😅 but I may be missing something obvious.

kmanijak (changes requested):
Hey @amitraj2203! As usual, great contribution!

I tested the solution and it works fine by default, appreciate it! I left some comments though, especially around Products by Cat/Tag working correctly with other taxonomies. I think we'll be able to merge it once it's addressed.

Also, do you mind adding some tests to these new collections? 🙌

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/edit/inspector-controls/taxonomy-controls/index.tsx
amitraj2203: The `TaxonomyControls` is what we see on the Filters panel, while the `TaxonomyControlsField` represents the top-most setting we see in the InspectorControls when selecting a collection.

That's why the `TaxonomyControls` is wrapped inside `ToolsPanelItem`, whereas `TaxonomyControlsField` isn't. The logic is similar to the hand-picked collections.

As for the similarity, you're right, they do appear redundant, and a cleanup would be beneficial, which I addressed  in https://github.com/woocommerce/woocommerce/pull/58440/commits/7194f2b77b019b38196f65e3e5676c6be7091b32.

Let me know if I'm missing anything 🙂

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/edit/inspector-controls/taxonomy-controls/index.tsx
amitraj2203: Thanks, I have addressed it.

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/edit/inspector-controls/index.tsx
amitraj2203: Done

amitraj2203:
> Also, do you mind adding some tests to these new collections? 🙌

Oops I didn't saw this, I'll work on it and will let you know once done.

amitraj2203:
Hey @kmanijak, I've added tests for both the collections now. Please review it when you get a chance.

Thanks!

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/edit/inspector-controls/taxonomy-controls/index.tsx
```diff
@@ -36,66 +40,125 @@ export const useTaxonomies = (): Taxonomy[] => {
...
+		} );
+		trackInteraction( `${ CoreFilterNames.TAXONOMY }__${ slug }` );
+	};
+
+	const createDeselectCallback = ( slug: string ) => () => {
+		createHandleChange( slug )( [] );
 	};
 
+	const shouldShowTaxonomyControl =
+		taxonomies && taxonomies.length > 0 && filteredTaxonomies.length > 0;
```
kmanijak: Feels like this can be simplified to:

```suggestion
	const shouldShowTaxonomyControl = filteredTaxonomies.length > 0;
```

This is the only condition that eventually matters 🤔 WDYT?

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/edit/inspector-controls/taxonomy-controls/index.tsx
```diff
@@ -36,66 +40,125 @@ export const useTaxonomies = (): Taxonomy[] => {
...
+			taxQuery: {
+				...taxQuery,
+				[ slug ]: newTermIds,
+			},
+		} );
+		trackInteraction( `${ CoreFilterNames.TAXONOMY }__${ slug }` );
+	};
+
+	const createDeselectCallback = ( slug: string ) => () => {
+		createHandleChange( slug )( [] );
```
kmanijak: Since this is used only within `TaxonomyControls` I'd remove it from this hook and init it in the `TaxonomyControls`. I guess something like:

```
const handleChange = createHandleChange( slug );
const deselectCallback = () => handleChange( [] );
```

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/edit/inspector-controls/taxonomy-controls/index.tsx
```diff
@@ -36,66 +40,125 @@ export const useTaxonomies = (): Taxonomy[] => {
...
 
-		return name
-			.split( ' ' )
-			.map( ( word ) => word.charAt( 0 ).toUpperCase() + word.slice( 1 ) )
-			.join( ' ' );
+		if ( collection === CoreCollectionNames.BY_CATEGORY ) {
+			return taxonomies.filter( ( taxonomy ) =>
+				isFiltersPanel
+					? taxonomy.slug !== 'product_cat'
+					: taxonomy.slug === 'product_cat'
```
kmanijak: Let's please add a comment explaining this logic for others who may look into here in future. E.g.

```suggestion
                                 // If it's in filter panel, we want to show everything BUT the category control.
                                 // Otherwise, it's a collection specific filter and we want to show ONLY the category control.
				isFiltersPanel
					? taxonomy.slug !== 'product_cat'
					: taxonomy.slug === 'product_cat'
```

kmanijak (commented):
Hey @amitraj2203, thanks for addressing my comments, especially adding the tests! I left couple minor comments, please let me know what you think of them! Once we agreed and resolve them, I'm happy to approve! 🙌

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/edit/inspector-controls/taxonomy-controls/index.tsx
kmanijak: Thanks for explanation and effort to share as much logic between them as possible, appreciate it!

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/edit/inspector-controls/taxonomy-controls/index.tsx
amitraj2203: Yup, updated it now!

kmanijak (changes requested):
Hey @amitraj2203 , I was doing another round of testing and spotted that using `Reset All` from Filters section in inspector controls, resets also the lifted control at the top of inspector controls. Apologies if that was a result of my suggestion to simplify `deselectCallback` 🙇 Here's a video showcasing what I mean. I think it's the only remaining issue at the moment 🙌 

https://github.com/user-attachments/assets/3f8304f5-5b7a-4d6b-83a1-74e423529d62

amitraj2203:
Hi @kmanijak, thanks again for the review, really appreciate it!

> I was doing another round of testing and spotted that using `Reset All` from Filters section in inspector controls, resets also the lifted control at the top of inspector controls. Apologies if that was a result of my suggestion to simplify `deselectCallback` 🙇 Here's a video showcasing what I mean. I think it's the only remaining issue at the moment 🙌

Regarding this bug, I don't think it was caused by simplifying `deselectCallback`. Interestingly, I noticed a strange behavior: when the block is inserted for the first time, clicking the `Reset All` button in the Filters panel also resets the lifted control at the top of the inspector controls. However, if we save the post/page once after inserting the block, refresh the editor and then test it again using the same steps, it works as expected.

This issue also occurs with the `Hand-Picked Products` collection and possibly with other collections as well _(though I haven't tested all of them yet)_. I'm attaching a quick demo showing the behavior for both the `Products by Category` and `Hand-Picked Products` collections to illustrate what I mean.

Let me know if I’m missing anything. Once again, thanks so much!

### Products by Category
https://github.com/user-attachments/assets/2efd98f8-de18-4bbb-9940-638a29bfa5ec

### Hand-Picked Products
https://github.com/user-attachments/assets/476fd93f-96c7-41ab-89ad-7038f911d78c

kmanijak:
Hi @gigitux , sorry for ping out of nowhere, do you mind making a final review of this PR? I didn't get a chance before AFK. Thank you 🙇

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/block.json
```diff
@@ -9,8 +9,6 @@
 		"WooCommerce",
 		"Products (Beta)",
 		"all products",
-		"by category",
-		"by tag",
```
gigitux: I don't think that we should remove these keywords. In this case, if the user searches by `by category` or `by tag` keyword, the Product Collection will not appear.

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/edit/inspector-controls/taxonomy-controls/index.tsx
```diff
@@ -36,66 +40,122 @@ export const useTaxonomies = (): Taxonomy[] => {
...
+	};
+
+	const shouldShowTaxonomyControl = filteredTaxonomies.length > 0;
+
+	return {
+		filteredTaxonomies,
+		taxQuery,
+		createHandleChange,
+		shouldShowTaxonomyControl,
 	};
```
gigitux: Let's move this file to a dedicated file in the `taxonomy-controls` folder. `use-taxonomy-controls` looks like a good  name for the file.

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/edit/inspector-controls/taxonomy-controls/index.tsx
gigitux: If I understood correctly, the only difference is that one component is wrapped in `ToolsPanelItem` and the other is not.  
If that's the case, let's update the `TaxonomyControls` implementation to remove the `ToolsPanelItem` wrapper.  
It will be up to the consumer to decide whether to wrap the block in `ToolsPanelItem` or not.

plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/Controller.php
```diff
@@ -307,6 +309,7 @@ public function update_rest_query_in_editor( $query, $request ): array {
 				'featured'            => $featured,
 				'timeFrame'           => $time_frame,
 				'priceRange'          => $price_range,
+				'taxonomies_query'    => $this->query_builder->get_filter_by_taxonomies_query( $raw_tax_query_from_rest_params ),
```
gigitux: What do you think if only `$raw_tax_query_from_rest_params` is passed?

`get_filter_by_taxonomies_query` can be invoked in `update_rest_query_in_editor` function. With this approach, `get_filter_by_taxonomies_query` can be a private function.

plugins/woocommerce/tests/php/src/Blocks/BlockTypes/ProductCollection.php
```diff
@@ -1581,4 +1581,268 @@ public function test_collection_hand_picked_order() {
 		$this->assertEquals( $product_ids, $result_frontend['post__in'] );
 		$this->assertEquals( $product_ids, $result_editor['post__in'] );
 	}
+
+	/**
+	 * Tests that the by-category collection handler works as expected.
+	 */
+	public function test_collection_by_category() {
```
gigitux, line 1588: Just a comment: this file is starting to be huge. We need to split tests. 

Did you have a chance to explore [this](https://github.com/woocommerce/woocommerce/pull/57892#issuecomment-2886666468)?

gigitux (changes requested):
Thanks for your help! I left a few comments!

> This issue also occurs with the Hand-Picked Products collection and possibly with other collections as well (though I haven't tested all of them yet). I'm attaching a quick demo showing the behavior for both the Products by Category and Hand-Picked Products collections to illustrate what I mean.

Thanks! Can you open an issue, so we can keep track of it?

plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/Controller.php
amitraj2203: > `get_filter_by_taxonomies_query` can be invoked in `update_rest_query_in_editor` function. With this approach, `get_filter_by_taxonomies_query` can be a private function.

Thank you for the suggestion! I’m a bit unclear on this part — could you please explain it in a bit more detail when you have a moment? 🙇

Just to clarify, the code for `get_filter_by_taxonomies_query` is currently added inside the `update_rest_query_in_editor` function. Let me know if I might be missing something or if there's a better way to structure it. Thanks again!

plugins/woocommerce/tests/php/src/Blocks/BlockTypes/ProductCollection.php
amitraj2203: Oh yes, apologies for missing this, I've created an issue for it [here](https://github.com/woocommerce/woocommerce/issues/59239). Could you please take a look and let me know if I’ve missed anything? Thanks!

amitraj2203:
Hi @gigitux @kmanijak,

Apologies for the delay, and thank you both for your patience. I was caught up with some other tasks recently, but I’ve now addressed the feedback.

I’ve also opened the following issues:

* **Product Collection: Reset All button resets inspector controls on initial insert of block**
  [https://github.com/woocommerce/woocommerce/issues/59238](https://github.com/woocommerce/woocommerce/issues/59238)

* **Product Collection: Separate tests into unit/integration and add editor-side integration tests**
  [https://github.com/woocommerce/woocommerce/issues/59239](https://github.com/woocommerce/woocommerce/issues/59239)

Could you please take a look and let me know if I’ve missed anything? Thanks again!

plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/Controller.php
gigitux: Sorry if I wasn't clear, but I mean something like this:

```php
''taxonomies_query' => $raw_tax_query_from_rest_params
```

And we update `get_final_query_args` like this:

image

Does it make sense?

amitraj2203: Understood, I've done it this way now.

plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/edit/inspector-controls/taxonomy-controls/index.tsx
gigitux: @amitraj2203 Can you check my previous comment?

amitraj2203: Hey @gigitux, actually the `TaxonomyControls` component is what we see in the Filters panel, and it’s always wrapped in a `ToolsPanelItem`.
For example, in [this usage](https://github.com/woocommerce/woocommerce/blob/b8d1046c8637c2532e3e1ca95bbc57413ae96ff4/plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/edit/inspector-controls/index.tsx#L217-L219), every other component is also wrapped in a `ToolsPanelItem` internally.
So, if we modify `TaxonomyControls` to allow the consumer to decide whether to wrap it in a `ToolsPanelItem` or not, it could break consistency in the code.

Do you think we should go ahead with that change?

https://github.com/woocommerce/woocommerce/blob/b8d1046c8637c2532e3e1ca95bbc57413ae96ff4/plugins/woocommerce/client/blocks/assets/js/blocks/product-collection/edit/inspector-controls/index.tsx#L193-L229

gigitux: Thanks for giving me more context. What about passing another prop like `renderMode` that accepts `panel` and `standalone`? When the `renderMode` is `panel`, the component is wrapped into the `ToolsPanelItem` component. What do you think?

amitraj2203: Thanks! I have implemented this approach now in this commit: https://github.com/woocommerce/woocommerce/pull/58440/commits/617ac2347d87b1c678e8c030352dbda2eeb925e6

plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/QueryBuilder.php
```diff
@@ -260,6 +260,7 @@ public function get_final_query_args(
 		$visibility_query = is_array( $query['stock_status'] ) ? $this->get_product_visibility_query( $stock_query, $query['stock_status'] ) : array();
 		$featured_query   = $this->get_featured_query( $query['featured'] ?? false );
 		$attributes_query = $this->get_product_attributes_query( $query['product_attributes'] );
+		$taxonomies_query = $this->get_filter_by_taxonomies_query( $query['taxonomies_query'] ) ?? array();
 		$taxonomies_query = $query['taxonomies_query'] ?? array();
```
gigitux: `$taxonomies_query` is assigned twice.

gigitux (dismissed):
I reviewed only the code and everything looks good, but I noticed that `$taxonomies_query` is assigned twice. Are you sure that `get_filter_by_taxonomies_query` is necessary? 🤔

plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/QueryBuilder.php
amitraj2203: I did some testing, and it seems it's not required so I've removed it now.

Also, I noticed that due to [this](https://github.com/woocommerce/woocommerce/pull/58440/files#diff-1b7c3c6fca3c97e9e29604197a61c256f6eaccc68aa4ce05ca6424d8d8358696R37-R39), when we add Products by category/tag, the order of products gets randomized on each reload. While, in the Inspector Controls, the `Order by` setting is shown as `A → Z`, which can be confusing.

Should we change `post__in` to `title` so that products are initially sorted alphabetically (ascending) by title?

kmanijak: > Should we change post__in to title so that products are initially sorted alphabetically (ascending) by title?

Yes, definitely, good catch!

kmanijak: Actually seems like you can remove the `order` field completely and it'll fallback to alphabetical order which I think would be the best.

kmanijak (changes requested):
Thanks for addressing @gigitux's feedback! The PR looks pretty good. The only thing missing is the incorrect order as you pointed out in your comment. I think we should remove the `query.orderBy` definition from these two collections: 

```
	query: {
		orderBy: 'post__in',
	},
```

and we'll be good to go! 💪

plugins/woocommerce/src/Blocks/BlockTypes/ProductCollection/QueryBuilder.php
amitraj2203: Thanks for the review, I've done it now in [536eaa0](https://github.com/woocommerce/woocommerce/pull/58440/commits/536eaa01127b44cb1389b67e90667a02b0952635)

kmanijak (changes requested):
Hey @amitraj2203 , thanks for addressing the comment. I was doing final testing and realised there's something wrong when Product by Category is mixed with some other taxonomy filter.

Here's an example

Case | Editor | Frontend
-- | -- | --
Only Category filter applied | image | image
Category filter + tags filter | image | image

When I apply additional filter it gets applied in Editor (the number of products is limited) but nothing is rendered on the frontend. When I create default Product Collection ("create your own") and apply the same filters, it works:

Editor | Frontend
-- | --
image | image

amitraj2203:
Hi @kmanijak,
I just tested the `Product by Category` collection, and it seems to be working fine in all the following cases:
- When no filter is applied
- When the `Product Tags` filter is applied
- When both `Product Tags` and `Product Price` filters are applied

Please see the attached screen recording for more details, and let me know if I'm missing anything.

---

### Testing `Product by Category` collection with no filters and `Product Tags` filter
https://github.com/user-attachments/assets/87644e78-ec8d-4e78-8f9a-365b4436c036

### Testing `Product by Category` collection with `Product Tags` and `Product Price` filters
https://github.com/user-attachments/assets/61ad0f61-059c-4187-9120-68ade3e0bf10

kmanijak (approved):
> I just tested the Product by Category collection, and it seems to be working fine in all the following cases:

Thanks a lot @amitraj2203 for retesting and providing videos. I gave it another go and it works as expected, I can no longer reproduce the issue so apologies, most likely I've had a repo in some weird state and didn't notice. 🙇 

It's all good from my perspective it seems. I'm happy to merge once CI pass, thanks a lot for your contirbution! 🎉

PR #58444, opened 2025-06-01
Remove support for obsolete Google Web Light

westonruter:
### Changes proposed in this Pull Request:

This removes support for Google Web Light which was introduced in #26858 and https://github.com/woocommerce/woocommerce/commit/9f6104598ba70425177a058f024cb98724c34e82. The Google Web Light service was shut down in 2022 [per Wikipedia](https://en.wikipedia.org/wiki/Google_Web_Light):

> On December 19, 2022, Google announced that it discontinued the Web Light service by removing the documentation and retiring the Web Light user agent. Increased affordability of more powerful [smartphones](https://en.wikipedia.org/wiki/Smartphones) was cited as the reason for this decision.[[5]](https://en.wikipedia.org/wiki/Google_Web_Light#cite_note-5)

See also [the update from Google Search Central](https://developers.google.com/search/updates#2022):

> December 19: Removed the [Web Light documentation](https://developers.google.com/search/docs/crawling-indexing/mobile/web-light) and retired the [Web Light user agent](https://developers.google.com/search/docs/crawling-indexing/overview-google-crawlers#retired). We introduced Web Light to enable us to serve faster, lighter pages to people searching on entry-level devices. While this feature has worked as intended and enabled broader access to the richness of the web, increased affordability of more powerful smartphones has diminished the need for such functionality. We remain committed to evolving and refining the Search experience to meet the changing needs of our users.

So this is removing technical debt.

### Changelog entry

-   [x] Automatically create a changelog entry from the details below.

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Details</summary>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

-   [ ] Fix - Fixes an existing bug
-   [ ] Add - Adds functionality
-   [ ] Update - Update existing functionality
-   [ ] Dev - Development related task
-   [x] Tweak - A minor adjustment to the codebase
-   [ ] Performance - Address performance issues
-   [ ] Enhancement - Improvement to existing functionality

#### Message 

</details>

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

PR #58445, opened 2025-06-01
Remove nocache headers from Cart page to enable bfcache (instant back/forward navigations)

westonruter:

I found that nocache headers were being sent on the Cart page seemingly unnecessarily. This appears to have been done for the sake of the browser's back/forward cache (bfcache) holding onto a stale cart page which wouldn't reflect a change made from another page after hitting the back button. However, as can be seen in the current version of Chrome, this no longer seems to be the case. (See screen recording below.)

In particular, this is no longer necessary because there is a `pageshow` event listener running on the Cart page to update the cart whenever landing on the cart after a bfcache navigation. For example:

https://github.com/woocommerce/woocommerce/blob/4697b875ed8bf161691deb312ce8e459a07cc319/plugins/woocommerce/client/legacy/js/frontend/cart-fragments.js#L109-L115

This PR therefore improves performance of navigating to and from the cart page by allowing the page to be served from bfcache.

Finally, this PR ensures the original `Cache-Control` directives from `nocache_headers()` are preserved when additional nocache headers are set. Namely, in WP core changeset [r55968](https://core.trac.wordpress.org/changeset/55968) the use of `private` was introduced.

----

This includes a commit (33c6961f54d4e9a4e55150089fc7617cc4eae7fb) from https://github.com/woocommerce/woocommerce/pull/58444 to remove the obsolete Google Web Light support. Once that PR is merged, that change will be removed from this PR.

https://github.com/user-attachments/assets/0fc35e38-19fc-463e-a4be-3f6d50be3857

0. Log out of WordPress (if logged-in, since `Cache-Control: no-store` is currently sent for authenticated users).
1. Add products to the cart.
2. Open DevTools to see the Back/Forward Cache test panel.
3. Click on one of the products in the cart.
4. From the mini-cart in the header, remove the first item in the cart.
5. Click the back button.
6. See that the Cart page was loaded instantly, that the page was loaded from bfcache, and see that the Cart is updated with the product removed.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Remove nocache headers from Cart page to enable bfcache

</details>

<details>

#### Comment 

</details>


westonruter:
The pages for Checkout and My Account might also be suitable for this change, but I didn't test that they have `pageshow` event handlers in the same way that the Cart page does.

senadir:
Thank you for this PR @westonruter, our perops team was looking into those headers and actually looking to add them to more pages, I will bring them to the discussion.

westonruter:
@senadir Adding `no-store` to more pages will degrade the performance of browsing a site. See https://web.dev/articles/bfcache#minimize-no-store

senadir:
not more pages, just the whole cart/checkout pages, as we aggressively cache other pages and we don't want to do that for cart/checkout.

westonruter:
Oh good. So in other words, in addition to removing `cart` (as done in my PR here), you also intend to remove `checkout` from this line as well:

https://github.com/woocommerce/woocommerce/blob/74a558f33c61f91570ca2b4fb31a773185722899/plugins/woocommerce/includes/class-wc-cache-helper.php#L153

thingalon:
Removing the Chrome-specific stuff seems ok - and I particularly support the idea of preserving core's cache-control directives, especially "private".

But this will remove the "no-cache" directive for all visitors on the cart page. There are likely thousands of WooCommerce sites behind edge caches which rely on those headers to prevent caching shopping carts across users.

I think it would be better to add "private" to cache-control on cart and checkout pages instead of no-cache.

westonruter:
@thingalon By `no-cache` do you mean `no-store`? 

This PR removes `no-store` for all visitors to the cart page, yes. However, because `nocache_headers()` is being called, core is sending the `private` directive which prevents edge caches from caching those responses.

Core added the `private` directive in WordPress 6.3, which is far before the 6.7 minimum supported version. So it appears that `private` will already be getting sent.

thingalon:
@westonruter - this diff affects both the `no-store` and `no-cache` headers.

On a fresh temp WordPress / WooCommerce install:
```
$ curl -sI 'https://test-host.local/temp-site/cart/' | grep Cache-Control
Cache-Control: no-cache, must-revalidate, max-age=0, no-store, private
```

With the patch installed:
```
$ curl -sI 'https://test-host.local/temp-site-patched/cart/' | grep Cache-Control
(nothing)
```

The diff prevents `nocache_headers()` from being called on the cart page. Since shopping carts get served embedded in pages for unauthenticated users, it could cause full shopping carts to get caught up in reverse-proxy caches. :)

I think it'd be more appropriate to set `Cache-Control: private` on cart and checkout pages for unauthenticated users. i.e.: I think it'd be ok to drop the `no-cache` and `no-store` directives from cart pages, but I think we should retain `private` for the sake of reverse-proxy caches. 

Thinking on it further: this will also bypass calling `set_nocache_constants()` on the cart page. The `DONOTCACHEPAGE` const defined there affects how common page-caching plugins like WP Rocket and WP Super Cache handle pages too.

westonruter:
@thingalon Oh yes, you're right. I confused myself because I also have in mind removing `no-store` from the default set of directives sent when calling `nocache_headers()`, leaving only `private`.

**Update:** See https://core.trac.wordpress.org/ticket/63636

So yes, `Cache-Control: private` is what needs to be added instead of calling `nocache_headers()` on the cart and checkout pages.

westonruter:
@thingalon OK, how about 3d0d954?

State | Before | After
--|--|--
Unauthenticated | `no-transform, no-cache, no-store, must-revalidate` | `private, no-cache, must-revalidate, max-age=0`
Authenticated | `no-transform, no-cache, no-store, must-revalidate` | `no-cache, must-revalidate, max-age=0, no-store, private`

In the authenticated state, the `no-store` directive is coming from `wp_get_nocache_headers()` sent from WordPress while logged-in. Otherwise, the other same directives from `wp_get_nocache_headers()` remain since they don't impact bfcache while also helping to ensure freshness. As noted above, I'm in the process of putting together a proposal to remove `no-store` from that. (**Update:** See [Core-63636](https://core.trac.wordpress.org/ticket/63636).)

The `set_nocache_constants()` call remains in place.

thingalon:
@westonruter - that looks good to me, and seems like a clear improvement around caching headers. Thanks for working on this! 

@senadir - can you please review it for merging into WooCommerce? :)

westonruter:
Here's a demo of the impact. While not being logged-in to WordPress and emulating a Slow 4G connection, compare entering a coupon on the cart page, then navigating to the shop, and then going back to the cart page. Notice how much faster the experience is when bfcache is enabled, and notice that the coupon dropdown retains its open state with the unsubmitted coupon still populating the field:

Before | After
--|--
<video src="https://github.com/user-attachments/assets/a03c0e61-70ff-4378-904e-a81a1ea0ce2e"> | <video src="https://github.com/user-attachments/assets/fdd169da-2f71-4710-8155-7f2cb81745e0">

Granted, the speed effect is exaggerated here because I had DevTools open with cache disabled, besides emulating Slow 4G.

plugins/woocommerce/includes/class-wc-cache-helper.php
```diff
@@ -26,56 +26,52 @@ class WC_Cache_Helper {
...
-	 * Set additional nocache headers.
+	 * Prevent caching on certain pages.
 	 *
-	 * @param array $headers Header names and field values.
 	 * @since 3.6.0
+	 *
+	 * @param array<string, string> $headers Header names and field values.
+	 * @return array<string, string> Filtered headers.
 	 */
-	public static function additional_nocache_headers( $headers ) {
```
senadir, line 46: Thank you for the changes, I'm a bit hesitant about deleting this public function outright but I can't see any usage [for it publicly](https://wpdirectory.net/search/01JZJ1KBY6DPWFX4M9Q93KS8CC) so I think we're good to go.

plugins/woocommerce/includes/class-wc-cache-helper.php
```diff
@@ -26,56 +26,52 @@ class WC_Cache_Helper {
...
+		// These directives are all included in `wp_get_nocache_headers()`, with the exclusion of `no-store` for the sake of bfcache.
+		$new_directives = array(
+			// Prevent caching the response in reverse proxies.
+			'private',
+
+			// Ensure freshness of the response but without `no-store` so that bfcache won't be disabled.
+			'no-cache',
+			'must-revalidate',
+			'max-age=0',
+		);
```
senadir: So why not use `wp_get_nocache_headers` and remove `no-store` from it instead of maintaining our own list here? that function also benefit from having a filter built-in.

plugins/woocommerce/includes/class-wc-cache-helper.php
```diff
@@ -26,56 +26,52 @@ class WC_Cache_Helper {
...
+		$old_directives = array();
+		if ( isset( $headers['Cache-Control'] ) ) {
+			$old_directives = preg_split( '/\s*,\s*/', $headers['Cache-Control'] );
 		}
+		$headers['Cache-Control'] = implode( ', ', array_unique( array_merge( $old_directives, $new_directives ) ) );
 
-		if ( $set_cache ) {
-			$headers['Cache-Control'] = 'no-transform, no-cache, no-store, must-revalidate';
-		}
 		return $headers;
```
senadir, line 113: The PR discussion mentioned adding back `no-store` for authenticated users, at which point are we doing that because I can't seem to see it? unless you're assuming being on cart/checkout/my-account equal being authenticated? which is technically true but doesn't cover all cases of being authenticated.

senadir (commented):
This is looking good to me, I left a couple of comments, one of them is slightly blocking unless you have which is about not using `wp_get_nocache_headers` if we're not longer hooking into `nocache_headers` filter.

plugins/woocommerce/includes/class-wc-cache-helper.php
westonruter: Adding `no-store` or _removing_ the directive? 

Are you referring to [this comment](https://github.com/woocommerce/woocommerce/pull/58445#issuecomment-2935741965)? 

In https://core.trac.wordpress.org/ticket/63636 I'm working on removing `no-store` for authenticated users in core.

plugins/woocommerce/includes/class-wc-cache-helper.php
westonruter: Yeah, I think that makes sense. 

How about this: we merge the headers from `wp_get_nocache_headers()` into `$headers`, and then _if the user is not logged-in_, we also remove `no-store` from `Cache-Control`. Then when https://core.trac.wordpress.org/ticket/63636 lands in core (hopefully) the `no-store` directive will also be removed for logged-in users automatically.

plugins/woocommerce/includes/class-wc-cache-helper.php
senadir: I was about this comment that includes `no-store` for authed users

https://github.com/woocommerce/woocommerce/pull/58445#issuecomment-2993821380

plugins/woocommerce/includes/class-wc-cache-helper.php
senadir: Happy to go with that!

senadir:
@westonruter I'm going on sabbatical by the end of this week; I'd love if we could get this merged ASAP so I don't need to hand it over. Let me know if you need help applying feedback.
Merging it this week will also land it in the WooCommerce 10.1 release coming out early August.

westonruter:
I will try to amend the PR tomorrow morning to do so, but I'll be AFK for the rest of the week.

plugins/woocommerce/includes/class-wc-cache-helper.php
westonruter: See 47f9b9c36a8c46f203cc73efa2e2ec3de22945a2

senadir (approved):
Looking good, thank you @westonruter

westonruter:
I've opened a similar pull request for Jetpack: https://github.com/Automattic/jetpack/pull/44322

PR #46429, opened 2024-04-10
Fix wrong Shop title shown in classic themes after deleting the page

Aljullu:
### Changes proposed in this Pull Request:

This PR partially reverts some changes introduced in https://github.com/woocommerce/woocommerce/pull/44537.

Rationale: we had some code that made sure the Shop page title was _Shop_. While we initially wrote that code to be applied only to block themes, given that `BlockTemplatesController` was run also in classic themes, it caused the change to apply to classic themes as well. When we moved that logic to the `ProductCatalogTemplate` class in #44537, it stopped being run in classic themes, causing the regression.

This PR moves the code back to `BlockTemplatesController` to make sure it runs in classic themes as well.

This is a temporary solution. If this code is expected to run in classic and block themes, ideally it shouldn't be in `BlockTemplatesController`.

Kudos to @jorgeatorres for finding the root of the issue!

### How to test the changes in this Pull Request:

1. Activate a classic theme (ie: Storefront).
2. Go to Pages and send the _Shop_ page to the _Trash_.
3. Visit the shop in the frontend (ie: `/shop`).
4. Verify the page title is _Shop_ instead of _Products_ or something similar. Important: check the title of the page that is in the `<title>` tags, that means the title that appears in the tab bar.
5. Restore the _Shop_ page so it's not in the _Trash_ anymore.
6. Switch to a block theme (ie: Twenty Twenty Four).
7. Repeat steps 2-4.

Before | After
--- | ---
imatge | imatge

### Changelog entry

-   [ ] Automatically create a changelog entry from the details below.

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

-   [ ] Fix - Fixes an existing bug
-   [ ] Add - Adds functionality
-   [ ] Update - Update existing functionality
-   [ ] Dev - Development related task
-   [ ] Tweak - A minor adjustment to the codebase
-   [ ] Performance - Address performance issues
-   [ ] Enhancement - Improvement to existing functionality

#### Comment 

</details>

PR #58449, opened 2025-06-01
Add unit test for Shop page behavior when trashed

### LINKED ISSUE #56975: Migrate "could-be-lower-level-test" tagged tests in shop folder
This is a sub issue of [https://github.com/woocommerce/woocommerce/issues/56964](https://github.com/woocommerce/woocommerce/issues/56964)

Check `plugins/woocommerce/tests/e2e-pw/tests/shop/`

Find tests tagged with `@could-be-lower-level-test` Playwright tag

For each test:

* Review the tagging. Does it makes sense?
* Ensure the functionality is covered in a lower layer of testing (unit, integration, etc.)
* Once coverage is confirmed, remove the test
* Review the specs. Any other candidates for lower level coverage? Anything else to improve?

BurakParsAydin:

Closes woocommerce/woocommerce#56975  .

1. Run the following command:<br>`pnpm --filter=@woocommerce/plugin-woocommerce test:unit:env -- --filter=WC_Tests_Shop_Page_Trashed`

### Changelog entry

- [X] Automatically create a changelog entry from the details below.

#### Significance

- [X] Patch
- [ ] Minor
- [ ] Major

#### Type

- [X] Dev - Development related task

#### Message

</details>

#### Comment

Add unit test to cover shop page behavior when trashed

</details>


BurakParsAydin:
Hi @adimoldovan 
I've prepped the unit test for the shop page scenario. This refers to the `shop-title-after-deletion.spec.js` file.
From my side, it's ready for review. Please feel free to take a look and let me know if any adjustments are needed!

For `cart-redirection.spec.js`:
The cart logic (adding products, validating cart contents, totals, etc.) appears to already be covered under `WC_Tests_Cart`.
The remaining part of this E2E test focuses on browser redirection after adding to cart, which obviously can't really be covered in unit tests. Because of that, I believe there's no need for additional unit tests in this case - but happy to hear your thoughts.

Overall, these are my observations for the tests under `plugins/woocommerce/tests/e2e-pw/tests/shop`.
If I'm missing anything, I'm more than happy to adjust based on your feedback.

adimoldovan (changes requested):
@BurakParsAydin Thanks for your contribution!

I have a few comments:

1. On the unit test - see #46429 which describes the issue:
  - the main assert should not be on the page id (which is supposed to be deleted) but on the page title, which should fallback to 'Shop'
  - I'm not entirely sure, but the fact that `wc_get_page_id( 'shop' );` returns the trashed page id could be a bug - pinging @Aljullu as the author of the fix to confirm this
  - it would be great if the test would run on 2 different themes, a blocks (e.g. 2024) and a classic one (e.g. Storefront)

2. With this flow covered on unit level the e2e spec `shop-title-after-deletion.spec.js` can be removed
3. There are 2 different changelog file, only one is needed so it's safe to remove one of them.

Aljullu:
>   - I'm not entirely sure, but the fact that `wc_get_page_id( 'shop' );` returns the trashed page id could be a bug - pinging @Aljullu as the author of the fix to confirm this

I'm not sure what's the expected behavior, to be honest. I just checked and `woocommerce_shop_page_id` isn't reset when trashing the _Shop_ page. I guess there are two ways to solve this:

1. We could hook into the action of deleting pages, check if the deleted page is the _Shop_ page, and, in case it is, delete the `woocommerce_shop_page_id` option. Then we would need to restore it if the page is restored.
2. Update `wc_get_page_id()` so it checks if the page is not in the trash before returning the value.

Having said that, I think it would be good to know if there are any real-life issues caused by the current behavior. If not, I would lean towards leaving it as-is to reduce the risk of breaking existing stores. What do you think @adimoldovan?

BurakParsAydin:
@adimoldovan Thanks for the review! Applied the requested changes.

adimoldovan:
> > * I'm not entirely sure, but the fact that `wc_get_page_id( 'shop' );` returns the trashed page id could be a bug - pinging @Aljullu as the author of the fix to confirm this
> 
> I'm not sure what's the expected behavior, to be honest. I just checked and `woocommerce_shop_page_id` isn't reset when trashing the _Shop_ page. I guess there are two ways to solve this:
> 
> 1. We could hook into the action of deleting pages, check if the deleted page is the _Shop_ page, and, in case it is, delete the `woocommerce_shop_page_id` option. Then we would need to restore it if the page is restored.
> 2. Update `wc_get_page_id()` so it checks if the page is not in the trash before returning the value.
> 
> Having said that, I think it would be good to know if there are any real-life issues caused by the current behavior. If not, I would lean towards leaving it as-is to reduce the risk of breaking existing stores. What do you think @adimoldovan?

Thank you for taking a look at this @Aljullu. My testing confirms what you're saying. Even more that that, `woocommerce_shop_page_id` isn't reset even when the page is permanently deleted. 
The fact that we're displaying the trashed page when navigating to `/shop` is a bug to me. And it looks like the e2e test checking this flow produces a false negative, because the title says 'Shop' because of the trashed page title, not because the `update_product_archive_title` function is executed. The fact that the trashed page has the same title as the fall back doesn't make a great test.

I will raise a bug to track this and the owner team can decide on the resolution.

adimoldovan:
> @adimoldovan Thanks for the review! Applied the requested changes.

@BurakParsAydin Thanks for the updates! Given the above comments, looks like this test should actually be failing. Let's put a pin on this and wait for a resolution on this issue so we'll know what to actually test for.

adimoldovan:
Opened https://github.com/woocommerce/woocommerce/issues/58504

nerrad:
Thanks for the contribution but this looks stale and no longer relevant given there's an issue created addressing the bug (and any PR related to that issue should add regression tests).

PR #58453, opened 2025-06-02
Product Details: running description through the_content filter

dinhtungdu:

This PR fix the issue that product description isn't wrapped in paragraph tags event though it's rich text. The issue is fixed by applying `the_content` filter like the `core/content` block.

Closes WOOPLUG-4475

1. Ensure the Single Product template is using Blockified Product Details (use WooCommerce Beta Tester to enable experimental blocks).
2. Visit a product on the frontend.
3. See the product description is wrapped by `<p>` tags.
4. Check with product that has multiple paragraph and image, see the HTML semantic is properly output.

### DIFF
--- a/plugins/woocommerce/src/Blocks/BlockTypes/ProductDescription.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/ProductDescription.php
@@ -57,6 +57,8 @@ class ProductDescription extends AbstractBlock {
 
 		// Get the description content.
 		$description = $product->get_description();
+		// phpcs:ignore WooCommerce.Commenting.CommentHooks.MissingHookComment
+		$description = apply_filters( 'the_content', str_replace( ']]>', ']]&gt;', $description ) );
 		if ( empty( $description ) ) {
 			unset( self::$seen_ids[ $product_id ] );
 			return '';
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/src/Blocks/BlockTypes/ProductDescription.php
```diff
@@ -57,6 +57,8 @@ protected function render( $attributes, $content, $block ) {
 
 		// Get the description content.
 		$description = $product->get_description();
+		// phpcs:ignore WooCommerce.Commenting.CommentHooks.MissingHookComment
+		$description = apply_filters( 'the_content', str_replace( ']]>', ']]&gt;', $description ) );
```
Aljullu, line 65: Why do we need this `str_replace()`? Should we add a comment so it's clear in the future? :thinking:

Aljullu (approved):
It works as described: paragraphs are wrapped in `<p>` elements and any other HTML like images is displayed correctly as well.

image

I left one question about the `str_replace()`, which I think would be good to document, but besides that, LGTM.

<details>

* Theme: Twenty Twenty-Five
* Store size: small
</details>

plugins/woocommerce/src/Blocks/BlockTypes/ProductDescription.php
dinhtungdu: I added a comment explaining the reason:

```
		/**
		 * This filter is documented in wp-includes/post-template.php.
		 * We follow core/content block to replace ]]> with ]&gt;
		 */
```

dinhtungdu: It was added to handle CDATA I guess, I'm not totally sure tbh. But as `core/content` does that, I think we should follow.

PR #58455, opened 2025-06-02
Add private data store to detect site editor template editing mode

amitraj2203:

This PR introduces a private WooCommerce data store to reliably detect whether the site editor is currently in template editing mode. This store addresses the issue where `getEditedPostId()` doesn't reset when navigating from template editing view back to the templates overview page.

#### Background:
For context, please refer to the related [issue](https://github.com/woocommerce/woocommerce/issues/56198) and the ongoing [PR discussion](https://github.com/woocommerce/woocommerce/pull/57901).

When implementing template-restricted block registration (as discussed [here](https://github.com/woocommerce/woocommerce/issues/56198)), we encountered a [challenge](https://github.com/woocommerce/woocommerce/issues/56198#issuecomment-2871043399) where blocks would appear broken in template preview cards. This occurs [because](https://github.com/woocommerce/woocommerce/pull/57901#issuecomment-2909916869) `getEditedPostId()` retains its value even when transitioning from template editing mode to the templates overview, leading to incorrect block registration decisions.

#### Solution:
The new data store monitors the URL for the presence of `canvas=edit` parameter to determine if we're actively editing a template, similar to the approach used in [WordPress Calypso](https://github.com/Automattic/wp-calypso/blob/532c4d6e3baeb3bc59a413cca0fbba11959ec46f/apps/help-center/hooks/use-canvas-mode.js).

Related to #56198 

Since this store hasn’t been implemented anywhere yet, we’ll need to temporarily use it somewhere to test. I’ve tested it using the `registerTemplateRestrictedBlockType` API we plan to introduce [here](https://github.com/woocommerce/woocommerce/pull/57901)
1. Checkout to this branch and edit the [Catalog Sorting](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/assets/js/blocks/catalog-sorting/edit.tsx) block to use this store and add a few logs.
2. Import the store and add the following code:
   ```tsx
   import { useCanvasMode } from '../../shared/redux-stores/canvas-mode';

   const { isEditMode } = useCanvasMode();

   isEditMode
   	? console.log('We are in edit mode.')
   	: console.log('We are not in edit mode.');

   ```
7. Build the changes and go to _Appearance > Editor > Template List_ page
8. Open Chrome DevTools and check the console. You should see: `We are not in edit mode.`
9. Edit any template that includes the catalog-sorting block like `Product Catalog`, `Product Search Results`, `Product by Brand` and you should then see the log: `We are in edit mode.`
10. Furthermore, you can use this store to select and dispatch data:
      ```js
      wp.data.select( 'woocommerce/canvas-mode' ).isEditMode();
      wp.data.dispatch( 'woocommerce/canvas-mode' ).setCanvasMode( true );
      ```

---

#### To run the tests for this store:
```shell
npm run test:js assets/js/shared/redux-stores/canvas-mode/test
```

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


amitraj2203:
Hi @tjcafferkey,
As discussed [here](https://github.com/woocommerce/woocommerce/pull/57901#issuecomment-2921656460), I've raised this PR to add a store for detecting template edit mode.

Since this is my first time adding a store in WooCommerce, I might have made some mistakes, particularly regarding where to place the store or in following general guidelines.

Please review the PR and feel free to let me know if the approach looks good or if any changes are needed. Thanks!

tjcafferkey:
Thank you for this Amit! Looping in @gigitux as I am on parental leave currently 😊

plugins/woocommerce/client/blocks/assets/js/shared/stores/canvas-mode/index.ts
```diff
@@ -0,0 +1,88 @@
...
+			type: ACTION_SET_CANVAS_MODE,
+			isEditMode,
+		};
+	},
+};
+
+const selectors = {
+	isEditMode() {
+		// Always check the current URL for the most up-to-date state.
+		const url = new URL( window.location.href );
```
gigitux: `new URL` could return an error. I think that we should wrap the URL parsing in a try/catch - what do you think?

plugins/woocommerce/client/blocks/assets/js/shared/stores/canvas-mode/README.md
gigitux, line 1: Thanks for writing a README 🙇

gigitux (changes requested):
Thanks for this PR! I left a comment. Also, I think that we should move this store in a different folder. Currently, in `stores` there are interactivity API stores. What do you think if we create a `redux-stores` folder?



Additionally, I would add some unit tests. Can you take care of it?

plugins/woocommerce/client/blocks/assets/js/shared/stores/canvas-mode/index.ts
amitraj2203: Done, also I have:
- Added the tests for it
- Moved this store into `client/blocks/assets/js/shared/redux-stores` folder
- Updated the PR description with the command for running the tests

plugins/woocommerce/client/blocks/assets/js/shared/redux-stores/canvas-mode/index.ts
```diff
@@ -0,0 +1,103 @@
...
+			const url = new URL( window.location.href );
+			const canvas = url.searchParams.get( 'canvas' );
+			const isEditMode = canvas === 'edit';
+
+			// Update store state to trigger any subscribed components.
+			const storeDispatch = dispatch( STORE_NAME ) as {
+				setCanvasMode: ( isEditMode: boolean ) => void;
+			};
+
+			await Promise.resolve();
```
gigitux: Is there a reason why this function is a Promise?

plugins/woocommerce/client/blocks/assets/js/shared/redux-stores/canvas-mode/index.ts
```diff
@@ -0,0 +1,103 @@
...
+	const handleUrlChange = async () => {
+		try {
+			const url = new URL( window.location.href );
+			const canvas = url.searchParams.get( 'canvas' );
+			const isEditMode = canvas === 'edit';
+
+			// Update store state to trigger any subscribed components.
+			const storeDispatch = dispatch( STORE_NAME ) as {
+				setCanvasMode: ( isEditMode: boolean ) => void;
+			};
```
gigitux: We can avoid defining the type manually using:

```suggestion
			dispatch(store).setCanvasMode(isEditMode)
```

plugins/woocommerce/client/blocks/assets/js/shared/redux-stores/canvas-mode/index.ts
```diff
@@ -0,0 +1,103 @@
...
+
+			await Promise.resolve();
+			storeDispatch.setCanvasMode( false );
+		}
+	};
+
+	window.addEventListener( 'popstate', () => handleUrlChange() );
+	window.addEventListener( 'pushstate', () => handleUrlChange() );
+
+	handleUrlChange();
```
gigitux: Should we invoke the function here?

plugins/woocommerce/client/blocks/assets/js/shared/redux-stores/canvas-mode/index.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export const store = createReduxStore( STORE_NAME, {
+	reducer,
+	actions,
+	selectors,
+} );
+
+if ( ! select( STORE_NAME ) ) {
+	register( store );
+}
```
gigitux: I don't think that this check is necessary. We can just register the store.

plugins/woocommerce/client/blocks/assets/js/shared/redux-stores/canvas-mode/test/index.ts
```diff
@@ -0,0 +1,123 @@
...
+	beforeEach( () => {
+		// Reset the store before each test.
+		if ( ! select( 'woocommerce/canvas-mode' ) ) {
+			register( store );
+		}
+	} );
+
+	describe( 'Initial State', () => {
+		it( 'should initialize with isEditMode as false', () => {
+			const state = select( 'woocommerce/canvas-mode' ).isEditMode();
```
gigitux: To get types, we can import the `store` object and use it.

```suggestion
			const state = select( store ).isEditMode();
```

plugins/woocommerce/client/blocks/assets/js/shared/redux-stores/canvas-mode/use-canvas-mode.ts
gigitux, line 1: I would move this file in the `hooks` folder

gigitux (changes requested):
Thanks for adding unit tests! 🙇 

I left a few comments that allow us to rely more on typescript types (without specifying them)

plugins/woocommerce/client/blocks/assets/js/shared/redux-stores/canvas-mode/index.ts
amitraj2203: Nope, I've removed it now.

plugins/woocommerce/client/blocks/assets/js/shared/redux-stores/canvas-mode/index.ts
amitraj2203: Without the Promise, the store update happens too fast, before WP has finished processing the URL change. This creates a timing issue where components get the old state instead of the new one, making everything appear _one step behind_.

I tested it in [PR](https://github.com/woocommerce/woocommerce/pull/57901) without the promise and it wasn't working properly.

plugins/woocommerce/client/blocks/assets/js/shared/redux-stores/canvas-mode/index.ts
amitraj2203: I've removed the check although this is being done [here](https://github.com/woocommerce/woocommerce/blob/5ea87a7fc5829f82efab612f899f59b9ea7c414d/plugins/woocommerce/client/blocks/assets/js/shared/stores/product-type-template-state/index.ts#L137-L139) too

plugins/woocommerce/client/blocks/assets/js/shared/redux-stores/canvas-mode/use-canvas-mode.ts
amitraj2203: Done

plugins/woocommerce/client/blocks/assets/js/shared/redux-stores/canvas-mode/index.ts
gigitux: Could you create a custom branch with all the changes? I want to double-check it! I don't think that this function should be a promise 🤔

amitraj2203: Sure, here’s the [PR](https://github.com/woocommerce/woocommerce/pull/58664) with all the changes. The rest of the details are mentioned in the [PR description](https://github.com/woocommerce/woocommerce/pull/58664#issue-3131252138).

plugins/woocommerce/client/blocks/assets/js/shared/redux-stores/canvas-mode/index.ts
```diff
@@ -0,0 +1,103 @@
...
+		try {
+			const url = new URL( window.location.href );
+			const canvas = url.searchParams.get( 'canvas' );
+			return canvas === 'edit';
+		} catch ( error ) {
+			// If URL parsing fails, return default state.
+			return false;
+		}
+	},
+};
```
gigitux, line 44: You should get the value from the state:

```suggestion
const selectors = {
	isEditMode( state: StoreState ) {
		return state.isEditMode;
	},
};
```

plugins/woocommerce/client/blocks/assets/js/shared/redux-stores/canvas-mode/index.ts
```diff
@@ -0,0 +1,103 @@
...
+				setCanvasMode: ( isEditMode: boolean ) => void;
+			};
+
+			await Promise.resolve();
+			storeDispatch.setCanvasMode( false );
+		}
+	};
+
+	window.addEventListener( 'popstate', () => handleUrlChange() );
+	window.addEventListener( 'pushstate', () => handleUrlChange() );
```
gigitux: I did some tests and I noticed that the real issue is here.

In some cases, the URL isn't the updated one:


https://github.com/user-attachments/assets/50179f3f-c4d3-4738-9ea8-b57ba6fc6973

gigitux: Also I noticed that `dispatch(store).setCanvasMode` returns a Promise. So we should add await:

```ts
await dispatch(store).setCanvasMode
```

gigitux (changes requested):
> Sure, here’s the PR with all the changes. The rest of the details are mentioned in the PR description.

Thanks for sharing the branch. I did some tests and I noticed that the root of the issue is that the `url` in the `pushstate` event isn't always the updated one. 


> This store addresses the issue where getEditedPostId() doesn't reset when navigating from template editing view back to the templates overview page.

What do you think if we try to fix this on the Gutenberg side? Does it make sense for you?

amitraj2203:
> What do you think if we try to fix this on the Gutenberg side? Does it make sense for you?

Hi @gigitux, yes, that sounds like a good idea. Actually, we briefly discussed it [here](https://github.com/woocommerce/woocommerce/pull/57901#:~:text=In%20the%20meantime,remove%20ours%20safely.) before raising this PR.

Would you suggest we pause work on this PR and proceed with addressing it on the Gutenberg side instead?

gigitux:
> > What do you think if we try to fix this on the Gutenberg side? Does it make sense for you?
> 
> Hi @gigitux, yes, that sounds like a good idea. Actually, we briefly discussed it [here](https://github.com/woocommerce/woocommerce/pull/57901#:~:text=In%20the%20meantime,remove%20ours%20safely.) before raising this PR.
> 
> Would you suggest we pause work on this PR and proceed with addressing it on the Gutenberg side instead?

Yes! I think that we can try to address the real issue:

> So the reason you're seeing this issue you describe is because when we load the site editor we land on an ineligible template (e.g. home) and when you transition out of the edit template view and into the all templates view it does not reset the getEditedPostId() value to undefined even though you're not editing a template (so it still has the value home). Because of this, the catalog block is not registered.

gigitux:
@amitraj2203 Given the outcome, can I close this PR?

amitraj2203:
@gigitux Sure, I’ll raise an issue in Gutenberg in a while regarding this.

PR #52253, opened 2024-10-23
Checkout: Collapsible Order Summary in mobile view

samueljseay:
### Changes proposed in this Pull Request:

We decided not to implement sticky behaviour of #45285 but this solves:

Implement an order summary (on mobile) that is:

✅ collapsed by default
✅ add a final Order summary before Place order button

See https://github.com/woocommerce/woocommerce/issues/45285#issuecomment-2428080020 if you'd like to see the rationale.

When in mobile size screen you'll see a collapsed by default order summary. The desktop experience should behave as before. Additionally at mobile size a duplicate order summary is rendered near bottom of checkout so that the user has convenient access to see the order summary at top or bottom of screen.

There is also some adjustment and shuffling of layout in the mobile mode. E.g. the duplicate summary at bottom of screen  has a header outside of the summary which is a little different from the one rendered in desktop view that has a header inside the summary border. 

I have not updated font sizes of headers to match the design, maybe this could be considered as a follow up to make it look more consistent with the design.

I also fixed a small visual bug in mobile where the `+ Add apartment, suite, etc.` toggle was not spaced correctly.

Screenshots:

Screenshot 2024-10-24 at 11 27 45 AM
Screenshot 2024-10-24 at 11 27 56 AM

### How to test the changes in this Pull Request:

1. View your checkout page with some sample products added to cart.
2. See that the desktop page still has the summary on the right side and a duplicate order summary should *not* be rendered
3. Now view at a mobile size
4. See that there is a collapsed by default order summary header which includes a summary of the total price.
5. Tap it to open the summary and see the rendered summary of order items.
6. Scroll to bottom of order and see that another non-collapsible order summary is rendered. 

### Changelog entry

-   [x] Automatically create a changelog entry from the details below.

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Details</summary>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

-   [ ] Fix - Fixes an existing bug
-   [ ] Add - Adds functionality
-   [ ] Update - Update existing functionality
-   [ ] Dev - Development related task
-   [ ] Tweak - A minor adjustment to the codebase
-   [ ] Performance - Address performance issues
-   [x] Enhancement - Improvement to existing functionality

#### Message 
Checkout: Add a collapsible order summary for smaller screens.

</details>

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

PR #58476, opened 2025-06-02
Order summary: Switch to css container width detection for the cart and checkout blocks

### LINKED ISSUE #59695: [Cart & Checkout block] Switch to CSS container width detection for Order summary
For the Cart and Checkout blocks, we need to step away from the JS added `is-large`, `is-small` , etc classes so we can prevent content jump when the blocks switch to a progressive rendering.

The blocks will render on the page with loading skeleton states where data is not available. The current behaviour renders an empty div before all the inner blocks are rendered, and then proceeds with a full render. At this point the JS added classes are available.

Without the change the user would see the content jumping in the blocks as the JS classes get addend after the initial render

ralucaStan:
Using JS to detect container width impacts the initial rendering of the blocks adding unnecessary layout shifts. These changes only account for the order summary area. 

The `is-large`, `is-small` etc classes are kept for backwards compatibility. The classes are public and they could be used by 3PD as they are covered in our docs. 
The PR introduces CSS mixins that can be used instead of the `is-large`, `is-small`, etc classes, which are added to the DOM after the initial render.

Closes https://github.com/woocommerce/woocommerce/issues/59695
Closes https://linear.app/a8c/issue/WOOPLUG-5033/cart-and-checkout-block-switch-to-css-container-width-detection-for

Things not handled by this PR:
- updating other areas of the blocks to use the container use mixing - will be handled in upcoming PRS

Note:
_While working on WOOPLUG-3274 and improving the checkout block loading I noticed that the order summary was loaded twice when rendering the block directly after the placeholder. These changes prevent the order summary to be rendered twice in the checkout block in the initial render. However this behaviour is not noticeable on trunk, as we initially render an empty div. I've created this PR on trunk for ease of testing, but it's a step towards improving the loading of the checkout block._

Part of WOOPLUG-3274.

The testing steps apply to the Order summary, in both the Checkout and the Cart areas:
1. Go to the block in fronted, and notice that the order summary looks and works as expected
2. Got to the block in the editor and notice the same thing. (use the tablet, mobile and desktop view options as well)
3. Resize the viewport and notice the Order summary adapting

I've tested this:
-  with and without Gutenberg active
- in the editor and frontend
- with block and non block themes

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/base/components/select/style.scss
```diff
@@ -1,5 +1,7 @@
 .wc-blocks-components-select {
 	width: 100%;
+	// override the max-width set in the editor
+	max-width: none;
```
ralucaStan, line 4: unrelated to the changes; a quick fix to the shorter country field we see in the editor. I can move it to a separate PR if necessary.

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-order-summary-block/frontend.tsx
```diff
@@ -100,27 +95,25 @@ const FrontendBlock = ( {
...
-						<div className="checkout-order-summary-block-fill">
-							{ children }
-							<div className="wc-block-components-totals-wrapper">
-								<TotalsFooterItem
-									currency={ totalsCurrency }
-									values={ cartTotals }
-								/>
-							</div>
-							<OrderMetaSlotFill />
+			<CheckoutOrderSummaryFill>
```
ralucaStan: this is handled in the CSS, we hide the fill for container width large

plugins/woocommerce/client/blocks/assets/css/abstracts/_mixins.scss
```diff
@@ -290,3 +290,42 @@ $fontSizes: (
...
+// Small: > 400px
+// Mobile: <= 400px
+
+@mixin cart-checkout-large-container {
+	@container (min-width: 700px) {
+		@content;
+	}
+}
+
+@mixin cart-checkout-not-large-container {
```
opr: Kind of a nit pick, but I believe this being used to describe all breakpoints that result in the layout becoming stacked right (e.g. as if on mobile)? What if we renamed it to something like `cart-checkout-narrow` or if that's too specific, perhaps `cart-checkout-below-large` is better since it gives more info about it.

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/order-summary/style.scss
```diff
@@ -1,9 +1,10 @@
-.wc-block-components-order-summary {
-	&.is-large {
-		// Compensate for removing Panel.
+@include cart-checkout-large-container {
```
opr: Could this be nested inside the selector so it directly replaces the `&.is-large {` ? Is there a reason you changed the order like this and duplicated `.wc-block-components-order-summary` ?

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-order-summary-block/frontend.tsx
opr: Previously, the fill would not render if not large, what are the implications of rendering this, but not showing it? For example, if components rendered here have side effects, could this break things?

opr (changes requested):
Thanks for working on this Raluca, I also tested in a few different themes, block and non-block, as well as on Firefox and Safari. 

Left a couple of comments, one is potentially a blocker (about the slot fill) so not approving for now, but besides that outstanding question it looks great

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/order-summary/style.scss
ralucaStan: Some context, [this PR](https://github.com/woocommerce/woocommerce-blocks/pull/2370/) introduced the `is-large`class on the `.wc-block-components-sidebar-layout` div. This class was later used on other containers like `.wc-block-components-order-summary`. Imo, reusing the class was a wrong approach because `is-large` is about the width of the block container, not the order summary container.

So on my PR I am referencing the width of the container which is available on first render, and ditching the dependency on the `is-large` class. I kept the classes though, because those are public API now.

Screenshot 2025-06-05 at 14 40 30

So that's why I changed the order of the selectors.
I removed the duplication in this commit https://github.com/woocommerce/woocommerce/pull/58476/commits/77513620c1dbcc9ed08b22bc963e52ce6fd7ff68

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-order-summary-block/frontend.tsx
ralucaStan: Great question, and it's something I should have considered.
After looking into it, the danger is that we are rendering an two areas with the same Slot twice. We have duplicated handlers, and store subscriptions, while also having some a11y issues. And we don't know what extensions are doing in the slot fill. 

But to be honest we already have this issue since  the duplicate order summary was introduced for mobile in this PR https://github.com/woocommerce/woocommerce/pull/52253. Because we are having the duplicated components already for the mobile view.

Testing on trunk with subscriptions one can see that the implementation is problematic and buggy.
- on mobile, the top order summary does not show information about the subscription; it's shown in the bottom order summary
- when enlarging the window size to a desktop, the subscriptions content from the slot disappears 

I didn't dive into it, but we should definitely reconsider the approach and implementation from https://github.com/woocommerce/woocommerce/pull/52253. I'll open another ticket for this. 

> Left a couple of comments, one is potentially a blocker (about the slot fill) so not approving for now, but besides that outstanding question it looks great

The `useContainerQueries()` is still available, so I will revert this part, keeping the JS handling of the Slot Fill Order summary. 

I don't think this will be a blocker for this PR, because we are still benefiting from the improvements at the CSS level for the order summary elements, making sure there is no content jump before those width classes are added. 

But a refactoring is still in place for the duplicated Order summary we have now.

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-order-summary-block/frontend.tsx
```diff
@@ -100,7 +96,7 @@ const FrontendBlock = ( {
 				</div>
 			</div>
 
-			{ ! isLarge && (
+			{ ( isMedium || isSmall || isMobile ) && (
```
ralucaStan, line 100: Reverted to JS handling of the CheckoutOrderSummaryFill
Using the reverted condition, to make sure this doesn't get rendered before we have values width, which was the case for `!isLarge`

ralucaStan: @opr  Screenshot 2025-06-06 at 19 18 45
As discussed, I considered rendering the <OrderMetaSlotFill /> from line 99 (see image) only on desktop, so we don't have the same Slot twice in the page. 

But I decided against it because:
- it means the component still unmounts and remounts when switching between large and small screens
- it can lead to potential memory leaks if cleanup isn't handled properly
- it might impact the functionality of some plugins, which work in the current setup

ralucaStan: @opr this is ready for another review

opr (approved):
Thanks Raluca, the changes look good. I tested with a plugin that adds a fill to the `ExperimentalOrderMeta` slot and it renders fine. I also checked Cart and it renders there too.

The fill doesn't render with padding, which I think is OK, I think we should leave that up to the component. It happens on trunk too, I don't think it's a regression.

image

PR #58487, opened 2025-06-02
Fix fatal error when WC_Email child class doesn't set template_html

### LINKED ISSUE #58486: Fatal error when WC_Email child class doesn't set template_html
In WC Tracker, we track usage of core WC emails. But some extensions, which use the same `WC_Email_` class prefix don't set `template_html` property, which caused fatal error.

```
PHP Fatal error: Uncaught TypeError: basename(): Argument #1 ($path) must be of type string, null given in /wordpress/plugins/woocommerce/9.9.1/src/Internal/Admin/EmailImprovements/EmailImprovements.php:183
```

NeosinneR:

==== **Don't merge this PR unless https://github.com/woocommerce/woocommerce/issues/58489 is approved.** ====

In WC Tracker, we track usage of core WC emails, which are detected by checking for `WC_Email_` prefix in the registered emails class name. In all core emails, the `template_html` is properly set, but some 3rd-party extension use the same prefix and don't set `template_html`, which ends up in fatal error, because the `template_html` is later used and is expected to be a string (but is `null`). 

This PR adds one more check for `template_html` property to pass `is_string()`.

Closes WOOPLUG-4488, closes #58486.

(For Bug Fixes) Bug introduced in PR #58286.

1. This isn't easily replicable in the UI, so some code needs to be changed to test this.
2. Before checking out this branch, add `var_dump( WC_Tracker::get_tracking_data() );` anywhere, which will indicate if the error is present.
3. Update any of the `includes/emails/class-wc-email-*` classes and remove `$this->template_html =` file from construct.
4. Reload the page to observe the fatal error.
5. Check out this branch. 
6. Reload the page to see there is no error.

The above.

### DIFF
--- a/plugins/woocommerce/changelog/wooplug-4488-fatal-error-when-wc_email-child-class-doesnt-set
+++ b/plugins/woocommerce/changelog/wooplug-4488-fatal-error-when-wc_email-child-class-doesnt-set
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix fatal error when WC_Email child class doesn't set template_html
--- a/plugins/woocommerce/src/Internal/Admin/EmailImprovements/EmailImprovements.php
+++ b/plugins/woocommerce/src/Internal/Admin/EmailImprovements/EmailImprovements.php
@@ -165,7 +165,7 @@ class EmailImprovements {
 		return array_filter(
 			WC()->mailer()->get_emails(),
 			function ( $email ) {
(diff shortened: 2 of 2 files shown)


pavel-mailpoet (approved):
Screenshot 2025-06-03 at 12 56 30

All the changes look good, including only the screenshot with the error, but I also tested with this branch, but there is nothing to show.

PR #57275, opened 2025-04-15
Fix: Remove Product Collection's Store Notices wrapper

luisherranz:
### Submission Review Guidelines:

-   I have followed the [WooCommerce Contributing Guidelines](https://github.com/woocommerce/woocommerce/blob/trunk/.github/CONTRIBUTING.md) and the [WordPress Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/).
-   I have checked to ensure there aren't other open [Pull Requests](https://github.com/woocommerce/woocommerce/pulls) for the same update/change.
-   I have reviewed my code for [security best practices](https://developer.wordpress.org/apis/security/).
-   Following the above guidelines will result in quick merges and clear and detailed feedback when appropriate.

### Changes proposed in this Pull Request:

Closes #57162.
Closes WOOPLUG-3883.

#### What 

This PR removes a wrapper introduced in https://github.com/woocommerce/woocommerce/pull/56117.

#### Why

The wrapper is causing problems with the CSS.

- https://github.com/woocommerce/woocommerce/issues/57162#issuecomment-2802084312

#### How

For now, I have moved the context of the notices to the namespace of the Product Collection block.

In the future, once [more than one context directive can be added to the same element](https://github.com/WordPress/gutenberg/discussions/62720), we can reintroduce the store-notices namespace.

---

Bug introduced in PR #56117.

### How to test the changes in this Pull Request:

1. Open a post, page or template with the _Product Collection_ block.
2. In a separate tab, edit one of the products in the grid and set it as out of stock.
3. Try adding the product to your cart.
4. Verify an error notice is displayed.

imatge

5. With your browser devtools, focus on the Navigation links at the bottom of the product grid. Verify they match the `:root :where(.is-layout-flow) > *` selector and (if using TT5, they have `margin-block-start: 1.2rem;`):

Captura de pantalla de 2025-04-15 10-01-33

### Changelog entry

-   [ ] Automatically create a changelog entry from the details below.

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Details</summary>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

-   [ ] Fix - Fixes an existing bug
-   [ ] Add - Adds functionality
-   [ ] Update - Update existing functionality
-   [ ] Dev - Development related task
-   [ ] Tweak - A minor adjustment to the codebase
-   [ ] Performance - Address performance issues
[...]

PR #58491, opened 2025-06-03
[Experimental] Display errors in Add to Cart + Options block

### LINKED ISSUE #57725: Blockified Add to Cart with Options block: show error notice when adding to cart fails

Aljullu:

Closes #57725.
Closes WOOPLUG-4152.

This PR adds support for displaying errors (and other notices) in the _Add to Cart + Options_ block. Currently, the Store Notices store is tightly coupled to the Product Collection block. Even though there were efforts to update it, we had to revert them (see #57275). In the _Add to Cart + Options_ implementation, I went for what I understand is the approach we want: using the Store Notices store context instead of the specific block context to handle the notices.

#### Next steps

As a next step, we should make it so notices are cleaned up when the user interacts with the form again (changes an attribute, submits again, etc.). However, I would do that in a follow-up issue, as I don't think it's a beta release blocker and this way we don't increase this PR complexity.

* Update: issue opened #58584 (WOOPLUG-4523).

#### Preparation

#### Testing the PR

1. Set a product as sold individually.
2. Go to the frontend and try to add the product twice.
3. Verify an error appears.

image

### DIFF
--- a/plugins/woocommerce/changelog/fix-57725-add-to-cart-options-errors
+++ b/plugins/woocommerce/changelog/fix-57725-add-to-cart-options-errors
@@ -1,0 +1,4 @@
+Significance: patch
+Type: add
+
+Display errors in Add to Cart + Options block
--- a/plugins/woocommerce/client/blocks/assets/js/base/stores/store-notices.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/base/stores/store-notices.ts
@@ -18,6 +18,7 @@ type NoticeWithId = Notice & {
 };
 
 const getStoreNoticeContext = getContextFn< {
+	notices: NoticeWithId[];
 	notice: NoticeWithId;
 } >;
 
@@ -66,7 +67,7 @@ const generateNoticeId = () => {
 };
 
 // Todo: export this store once the store is public.
-store< Store >(
+const { state } = store< Store >(
 	'woocommerce/store-notices',
 	{
 		state: {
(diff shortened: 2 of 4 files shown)


Aljullu:
@luisherranz I tagged you for review as you worked on notices in the past and might have some thoughts on this PR. Hope that's ok!

gigitux:
Thanks for this PR! 🙇  Everything works as expected, but I think that we should avoid rendering the same notices multiple time. What do you think?


https://github.com/user-attachments/assets/69f56262-62fc-4283-b861-0ff2825474ee

gigitux (approved):
LGTM! I tested it in the Single Product block too and everything works as expected:

https://github.com/user-attachments/assets/69f56262-62fc-4283-b861-0ff2825474ee


> As a next step, we should make it so notices are cleaned up when the user interacts with the form again (changes an attribute, submits again, etc.). However, I would do that in a follow-up issue, as I don't think it's a beta release blocker and this way we don't increase this PR complexity.

It makes sense, and I'm fine to do in a follow-up issue! 🚀

luisherranz:
> As a next step, we should make it so notices are cleaned up when the user interacts with the form again (changes an attribute, submits again, etc.). However, I would do that in a follow-up issue, as I don't think it's a beta release blocker and this way we don't increase this PR complexity.

Before making improvements to the current way of showing notices, it would be good to continue this discussion to decide if the way to show store notices is definitely how we are doing it now (putting a store notice in each block), or if it would be a better user experience to do it with snackbars/toasts or some other option.

- https://github.com/woocommerce/woocommerce/discussions/55731

PR #58492, opened 2025-06-03
Experimental iAPI Mini Cart: Remove item, set item quantity, support min and maximum

samueljseay:

This PR fills out more functionality of the iAPI powered mini cart:

1. When using the stepper buttons, they are disabled when min and max are reached
2. Editing the text of the input updates the quantity (and also respects min and max with same behaviour as the original mini cart)
3. Remove item button works to remove an item entirely
4. Fixes a bug where updating quantities makes the mini cart items flicker

To support removing items I've added another action to the store that uses the `remove-item` endpoint of the store API.

To fix the flickering bug, I've implemented a first attempt at updating the store state without blowing away all the signals. Its a rudimentary implementation and could be improved iteratively in future. It's still (imo) better than blowing away all the signals like we do now with `state.cart = json;`

https://github.com/user-attachments/assets/463b39f8-0c31-47fd-842e-7e92e9f33b74

TBD

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
@@ -15,6 +15,7 @@ import type { Store as StoreNotices } from '@woocommerce/stores/store-notices';
  * Internal dependencies
  */
 import { triggerAddedToCartEvent } from './legacy-events';
+import { updateCartItemsByKey } from './helpers';
 
 export type OptimisticCartItem = {
 	key?: string;
@@ -33,6 +34,7 @@ export type Store = {
 		};
 	};
 	actions: {
+		removeCartItem: ( key: string ) => void;
 		addCartItem: ( args: OptimisticCartItem ) => void;
 		// Todo: Check why if I switch to an async function here the types of the store stop working.
 		refreshCartItems: () => void;
@@ -82,6 +84,55 @@ const { state, actions } = store< Store >(
 	'woocommerce',
 	{
 		actions: {
+			*removeCartItem( key: string ) {
+				// optimistically update the cart
+				state.cart.items = state.cart.items.filter(
+					( item ) => item.key !== key
+				);
+
+				const previousCart = JSON.stringify( state.cart );
+
+				try {
+					const res: Response = yield fetch(
+						`${ state.restUrl }wc/store/v1/cart/remove-item`,
+						{
+							method: 'POST',
+							headers: {
+								Nonce: state.nonce,
+								'Content-Type': 'application/json',
+							},
+							body: JSON.stringify( { key } ),
+						}
+					);
+
+					const { items, ...rest }: Cart = yield res.json();
+
+					// @ts-ignore
+					updateCartItemsByKey( state.cart.items, items );
+
+					state.cart = {
+						items: state.cart.items,
+						...rest,
+					};
+				} catch ( error ) {
+					const { items: previousItems, ...rest } =
+						JSON.parse( previousCart );
+
+					updateCartItemsByKey(
+						// @ts-ignore
+						state.cart.items,
+						previousItems
+					);
+
+					state.cart = {
+						items: state.cart.items,
+						...rest,
+					};
+
+					// Shows the error notice.
+					actions.showNoticeError( error as Error );
+				}
+			},
 			*addCartItem( { id, quantity, variation }: OptimisticCartItem ) {
 				let item = state.cart.items.find(
 					( { id: productId } ) => id === productId
@@ -125,8 +176,14 @@ const { state, actions } = store< Store >(
 						actions.showNoticeError( error );
 					} );
 
-					// Updates the local cart.
-					state.cart = json;
+					const { items, ...rest } = json;
+					// @ts-ignore
+					updateCartItemsByKey( state.cart.items, items );
+
+					state.cart = {
+						items: state.cart.items,
+						...rest,
+					};
 
 					// Dispatches a legacy event.
 					triggerAddedToCartEvent( {
@@ -138,7 +195,14 @@ const { state, actions } = store< Store >(
 				} catch ( error ) {
 					// Reverts the optimistic update.
 					// Todo: Prevent racing conditions with multiple addToCart calls for the same item.
-					state.cart = JSON.parse( previousCart );
+					const { items, ...rest } = JSON.parse( previousCart );
+					// @ts-ignore
+					updateCartItemsByKey( state.cart.items, items );
+
+					state.cart = {
+						items: state.cart.items,
+						...rest,
+					};
 
 					// Shows the error notice.
 					actions.showNoticeError( error as Error );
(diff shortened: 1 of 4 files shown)


plugins/woocommerce/client/blocks/assets/js/base/stores/woocommerce/cart.ts
```diff
@@ -82,6 +84,48 @@ const { state, actions } = store< Store >(
...
+							body: JSON.stringify( { key } ),
+						}
+					);
+
+					const { items, ...rest }: Cart = yield res.json();
+					updateCartItemsByKey( state.cart.items, items );
+
+					state.cart = {
+						items: state.cart.items,
+						...rest,
```
samueljseay, line 113: later we could more granularly update all of the cart, for now I just needed to avoid regenerating references for all the cart items.

I view this as something we can iterate on and improve more.

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
```diff
@@ -177,6 +177,30 @@ type CartItemContext = {
 
 store( 'woocommerce/mini-cart-items-block', {
 	state: {
+		// Intended to be used in context of a cart item in wp-each
+		get cartItemMinimum() {
```
samueljseay, line 181: For functionality we don't really need to bind these to the html attributes, but from accessibility perspective (ie screen readers) it makes sense to.

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
```diff
@@ -294,6 +318,49 @@ store( 'woocommerce/mini-cart-items-block', {
...
+				{},
+				{ lock: universalLock }
+			);
+			const { minimum, maximum } = cartItem.quantity_limits;
+			const quantity = parseInt( qty, 10 );
+
+			let finalQuantity = quantity;
+
+			if ( quantity < minimum ) {
+				input.value = `${ minimum }`;
```
luisherranz, line 337: Hmmm... why do you have to change the input value manually? Shouldn't you change its value in the context, since the `value` attribute is bound with a directive?

`data-wp-bind--value="context.cartItem.quantity"`

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
```diff
@@ -294,6 +318,49 @@ store( 'woocommerce/mini-cart-items-block', {
...
+			const input = e.target as HTMLInputElement;
+			const qty = input.value;
+
+			const { cartItem } = getContext< CartItemContext >();
+			const { actions } = store< WooCommerce >(
+				'woocommerce',
+				{},
+				{ lock: universalLock }
+			);
+			const { minimum, maximum } = cartItem.quantity_limits;
```
luisherranz, line 331: I wonder if it would be okay to abstract this logic of the minimum and maximum within the `actions.addCartItem` action instead. I guess it's something that needs to happen for every block, right? What do you think?

luisherranz (changes requested):
In general it looks good to me, although I've added a couple of questions below.

What I don't like so much is the need to do the deep equal when updating the cart. That might be some kind of optimization we can do in the future, but now I would like to investigate why the flickering is happening, since in principle, despite replacing the signals, there should be no rendering without data. That is, the flickering shouldn't exist.

Another note, not important right now, but I was wondering if it's really necessary to import the `@woocommerce/stores/woocommerce/cart` store when the page loads. We can do some tests there to see how easy it would be to load it dynamically instead.

luisherranz:
It seems the flickering was due to us not telling Preact what the key of the array was, so Preact was deleting the `tr` element and recreating it. Adding `data-wp-each-key` solves the issue.

`<template data-wp-each--cart-item="state.cartItems" data-wp-each-key="context.cartItem.key">`

Still, the flickering shouldn't exist if the HTML matches (the browser simply deletes the markup and adds it again before repainting the screen), but the flickering was noticeable because the item description is being modified using a `data-wp-watch` that doesn't modify the HTML synchronously because it's behind a `useEffect`.

The same flickering will be seen if a product is added to the cart while the MiniCart is open, but I don't know if that's possible. So, for now we can leave it like that.

EDIT: I would revert all the changes when overwriting the cart and leave it as it was. It is an optimization that we can do later and with more care.

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
samueljseay: this is to match behaviour in the existing mini cart (which i dont particularly like) which is, that after you type, if the value doesn't match the constraint we overwrite it in place. Its quite bad UX because in some cases like where you're typing with 0's there are certain values you cannot enter.

I don't see how we can get that behaviour by simply binding the value via directive? this has to happen after the user typed a value?

We could suggest improving the UX of this as a follow up?

samueljseay:
@luisherranz thanks for investigating that. I assumed that blowing away all signals would still force a lot of reactivity changes, I guess thats still true but we shouldn't get the flicker at least 👍🏻 

> The same flickering will be seen if a product is added to the cart while the MiniCart is open, but I don't know if that's possible. So, for now we can leave it like that.

There is only the case with block setting "open drawer when adding to cart" where maybe it flickers as the cart is opening. It might not be visible. I agree we can revert all those changes and explore later if optimization is needed.

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
samueljseay: I think I need to check one thing about this. I think in the original mini cart this overwriting of the input doesn't trigger the value change until leaving the input, whereas I've left this as on-input right now. Maybe in either situation context will work fine here. 😄

luisherranz: Switching to `onChange` makes sense. Still, we should try to change the value of the context instead of the value of the input in the DOM. If that doesn't work, we can investigate to understand what's happening.

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
samueljseay: I'm not sure about this. we're using this here as UX to limit the input quantity, but if you did this work in the store, what would you do, throw an error that should be caught? 

For the current implementation (including things like product button) the only reason we need this is that the user can set an unreasonable value in this field, but for trying to increment by 1, the stepper becomes disabled, and for product button, it returns an error displayed after the API returns, rather than us having to do anything calculation of min or max in the button itself.

I think we could consider this for sure, but I would want to do it as a follow up, or perhaps wait to see other use cases before implementing it generally there.

samueljseay:
Closing this in favor of a new branch, will create a new PR

plugins/woocommerce/client/blocks/assets/js/blocks/mini-cart/iapi-frontend.ts
luisherranz: Yeah, no problem. We can investigate this when we analyze in detail what shape the `woocommerce` store is taking as we progress (before making it public).

PR #58496, opened 2025-06-03
[NOX in LYS] Redirect to the LYS hub after successful account onboarding

mordeth:

Closes WOOPLUG-4491

Video: https://www.loom.com/share/64a4e8061fda454f8ae3c386b86ee6d8

1. Create a new JN store with this PR's branch build (here's a [direct create link](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce-beta-tester&woocommerce&branches.woocommerce=update/WOOPLUG-4491-redirect-after-onboarding)).
2. Go to the new store's dashboard, click on WooCommerce > Home to trigger the core profiler, skip it, and set your store's location to Austria.
3. Go to Plugins > Add plugin and install&activate the [WCPay Dev Tools plugin](https://github.com/Automattic/woocommerce-payments-dev-tools/tree/trunk).
4. Go to "WooCommerce > Home" and click on Launch your store task.
5. Click on the "Set up payments".
6. Proceed with the account onboarding.
7. Once the account is onboarded, verify you are redirected back to LYS hub.

Check above.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/data/onboarding-context.tsx
+++ b/plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/data/onboarding-context.tsx
@@ -214,8 +214,7 @@ export const OnboardingProvider: React.FC< {
 					areStepDependenciesCompleted( step, allSteps )
 			);
 
-			// If the next step is the same as the current step, it means we are at the last step
-			// and we should call onFinish if provided.
+			// If the current step is the last one, then we should call onFinish if provided.
 			if ( currentStepIndex === allSteps.length - 1 ) {
 				onFinish?.();
 			}
--- a/plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/data/onboarding-context.tsx
+++ b/plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/data/onboarding-context.tsx
@@ -62,7 +62,6 @@ const OnboardingContext = createContext< OnboardingContextType >( {
 	closeModal: () => undefined,
 	justCompletedStepId: null,
 	setJustCompletedStepId: () => undefined,
-	source: null,
 } );
 
 export const useOnboardingContext = () => useContext( OnboardingContext );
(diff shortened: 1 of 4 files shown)


elazzabi (changes requested):
Thank you for working on this!

I'm not a fan of hardcoding this logic inside the steps. Ideally, the step shouldn't care about any redirection, and it shouldn't care if it's the last one or not. In all cases, the step should call `navigateToNextStep` when finished.

This is where having an `onFinish` function in the context would be helpful. Currently, the logic we have checks for the next step and then navigate:

```js
if ( nextStep ) {
	navigateToStep( nextStep.id );
}
```

We can add a simple `onFinish` there:

```js
if ( nextStep ) {
	navigateToStep( nextStep.id );
} else {
	// All steps are completed, call onFinish if provided
	onFinish?.();
}
```

WDYT?

mordeth:
@elazzabi, thank you for the feedback!

Implementing the `onFinish` method definitely makes sense. However, we couldn’t use the suggested approach because, for some reason, the `nextStep` always returns the current step when it’s the last one. Instead, I’ve tied the `onFinish` logic to the step index to handle this case.

Would you mind taking another look?

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/data/onboarding-context.tsx
```diff
@@ -206,6 +214,12 @@ export const OnboardingProvider: React.FC< {
 					areStepDependenciesCompleted( step, allSteps )
 			);
 
+			// If the next step is the same as the current step, it means we are at the last step
+			// and we should call onFinish if provided.
```
elazzabi:
```suggestion
			// If the current step is the last one, then we should call onFinish if provided.
```

elazzabi: And I'll also move it to the top so that it can be executed before `const nextStep` (given we'll not use it anyway). Maybe even `return` to stop the execution?

elazzabi (commented):
It looks cleaner now, and I left a last request for change. Should be okay to merge once fixed

plugins/woocommerce/client/admin/client/settings-payments/onboarding/providers/woopayments/data/onboarding-context.tsx
mordeth: Updated. I had intentionally left it without a return if `onFinish` failed, but on second thought, I agree it makes sense to include it as we don't need to navigate to the same step.

elazzabi (approved):
Looks good after the recent changes 🚢

PR #58497, opened 2025-06-03
Round individual taxes in fee lines items in orders REST endpoint

### LINKED ISSUE #58351: Tax line items are not rounded in order endpoint response
## **Describe the bug**

In the order endpoint response, individual tax line items are not rounded despite the total being rounded.

This leads to a discrepancy in the total vs summed value of the line items and makes it challenging to fully refund specific tax line items without knowing how they will be rounded on the backend.

```
"fee_lines": [
      {
        "id": 381,
        "name": "$14.70 fee",
        "tax_class": "",
        "tax_status": "taxable",
        "amount": "14.7",
        "total": "14.70",
        "total_tax": "1.33",
        "taxes": [
          {
            "id": 3,
            "total": "0.588",
            "subtotal": ""
          },
          {
            "id": 2,
            "total": "0.735",
            "subtotal": ""
          }
        ],
        "meta_data": []
      }
    ]
```

## **To Reproduce**

1. Add two taxes to your store (e.g., 5% and 4%)
2. Create a test order that includes both of these tax rates
3. Add a fee of $14.70 to the order
4. Make a request to the orders endpoint
5. Note the sum of the tax line items does not match the total

## **Expected behavior**

Expected the sum of the tax line items to match the total so that we can fully and accurately refund a set of line items.

Note that the alternative would be to allow summing on the backend and provide the same non-summed values back to the refund endpoint, but this leads to us using inconsistent numbers and potentially unexpected behavior.

## **Screenshots**

```
"fee_lines": [
      {
        "id": 381,
        "name": "$14.70 fee",
        "tax_class": "",
        "tax_status": "taxable",
        "amount": "14.7",
        "total": "14.70",
        "total_tax": "1.33",
        "taxes": [
          {
            "id": 3,
            "total": "0.588",
            "subtotal": ""
          },
          {
            "id": 2,
            "total": "0.735",
            "subtotal": ""
          }
        ],
        "meta_data": []
      }
    ]
```

## **Additional context**

Note that rounding is dependent on [this setting](https://github.com/woocommerce/woocommerce/blob/d4bbb38d0a8c4d6e329a597a692bdc26324e6c45/plugins/woocommerce/includes/class-wc-order-item-fee.php/#L202-L206) and needs to be taken into consideration in the response.

joshuatf:

This PR rounds the individual taxes in the order fee line items in the orders REST endpoint.  Providing unrounded numbers means the client must know the number of decimal places and formatting used for currencies in the client and manually calculate them for each line item, which is not ideal.

This also matches the expected values in other areas (various order screens and WC Store API endpoint).

Closes #58351 .

| Before | After |
| ------ | ----- |
|    Screenshot 2025-06-03 at 6 43 34 AM   |    Screenshot 2025-06-03 at 6 43 17 AM  |

1. Create REST API credentials at `/wp-admin/admin.php?page=wc-settings&tab=advanced&section=keys`
2. Create two sets of taxes, one at 5% and one at 4%.  Assign them different priorities so that both get used in your order
3. Create an order with a fee of 14.70 that the taxes apply to
4. Make a request with your favorite API app to `/wp-json/wc/v3/orders/{ORDER_ID}` replacing with the order ID just created
5. Check that under `fee_lines`, the individual taxes are rounded to your stores decimal places (e.g., `0.74` and `0.59` if your store uses 2 decimal places).
6. Provide the `dp` argument with a custom value (e.g., `3`).
7. Make sure that the response for individual taxes under fee lines is rounded to the provided number of decimal points

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-orders-v2-controller.php
+++ b/plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-orders-v2-controller.php
@@ -458,7 +458,6 @@ class WC_REST_Orders_V2_Controller extends WC_REST_CRUD_Controller {
 		}
 
 		// Format decimal values.
-		// Need this for fee_lines['taxes'] to be formatted.
 		foreach ( $format_decimal as $key ) {
 			$data[ $key ] = wc_format_decimal( $data[ $key ], $this->request['dp'] );
 		}
--- a/plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-orders-v2-controller.php
+++ b/plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-orders-v2-controller.php
@@ -252,8 +252,8 @@ class WC_REST_Orders_V2_Controller extends WC_REST_CRUD_Controller {
 			foreach ( $data['taxes']['total'] as $tax_rate_id => $tax ) {
 				$taxes[] = array(
 					'id'       => $tax_rate_id,
-					'total'    => $tax,
-					'subtotal' => isset( $data['taxes']['subtotal'][ $tax_rate_id ] ) ? $data['taxes']['subtotal'][ $tax_rate_id ] : '',
+					'total'    => wc_format_decimal( $tax, $this->request['dp'] ),
+					'subtotal' => isset( $data['taxes']['subtotal'][ $tax_rate_id ] ) ? wc_format_decimal( $data['taxes']['subtotal'][ $tax_rate_id ], $this->request['dp'] ) : '',
 				);
 			}
 			$data['taxes'] = $taxes;
@@ -458,6 +458,7 @@ class WC_REST_Orders_V2_Controller extends WC_REST_CRUD_Controller {
 		}
 
 		// Format decimal values.
+		// Need this for fee_lines['taxes'] to be formatted.
 		foreach ( $format_decimal as $key ) {
 			$data[ $key ] = wc_format_decimal( $data[ $key ], $this->request['dp'] );
 		}
--- a/plugins/woocommerce/changelog/fix-58351
+++ b/plugins/woocommerce/changelog/fix-58351
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Round individual taxes in fee lines items in orders REST endpoint


plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-orders-v2-controller.php
```diff
@@ -252,8 +252,8 @@ protected function get_order_item_data( $item ) {
 			foreach ( $data['taxes']['total'] as $tax_rate_id => $tax ) {
 				$taxes[] = array(
 					'id'       => $tax_rate_id,
-					'total'    => $tax,
-					'subtotal' => isset( $data['taxes']['subtotal'][ $tax_rate_id ] ) ? $data['taxes']['subtotal'][ $tax_rate_id ] : '',
+					'total'    => wc_format_decimal( $tax, $this->request['dp'] ),
```
joshuatf, line 255: I went back and forth on whether or not we wanted to change this based on the option value of `woocommerce_tax_round_at_subtotal`, but I decided against it for 2 reasons:

1. There are no obvious use cases for using unrounded numbers.  If we wanted to calculate the total taxes, there is already a calculated total taxes item provided that can be used on the frontend.
2. If more precision is needed for some reason, the `dp` param can be used

kalessil:
:eyes:

kalessil (approved):
LGTM and tests well! I'm also sharing the rationale from the comment.

plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-orders-v2-controller.php
hichamboushaba: @joshuatf I think there could an issue with this choice for the clients, as the goal of this change was to allow us to correctly calculate the refund amount based on the various tax rates applied to the order, returning the items rounded when `woocommerce_tax_round_at_subtotal` is enabled could lead to some inconsistencies, for example in the above example, this change will mean that we'll get: `0.74` and `0.59` taxes, which gives `1.33`, while the the actual `total_tax` would be `1.32` as the rounding is done on the subtotal level, which means that we might refund more than what was charged.

If the response was based on `woocommerce_tax_round_at_subtotal`, the API response should return `0.588` and `0.735` and would match the `1.32` total tax.

Please correct me if I'm misunderstanding something here 🙏.

joshuatf: Hey @hichamboushaba 👋  as noted above, I don't see a real use case for having those numbers not rounded because it means the client becomes responsible for:

* Knowing the rounding and decimal configuration 
* Knowing if `woocommerce_tax_round_at_subtotal` is true or not
* Calculating totals based on the above value

In general, it's not wise that the client handles any calculations, which is why the `total_tax` is included in the response.  If the total is needed, that's the value that should be used.  The use case for individual tax lines would be for display purposes and not calculation and because of that, they should be formatted with the correct number of decimals to match all other currencies.  This is also the same expectation from the StoreApi.

If you have a specific use case for wanting to calculate totals manually, you can use the `dp` parameter to get a higher level of precision.  But if you have some other use case where `total_tax` doesn't work, please share.

PR #58499, opened 2025-06-03
Update non-marketplace recommendations placements

KokkieH:

This PR further enhances the new endpoint we've switched to in #58187 to allow the number of recommendations shown to be set remotely. We're also removing some recommendations that are no longer needed.

Related changes to the API in https://github.com/Automattic/woocommerce.com/pull/23982

- Remove recommendations on the empty products list page - as of 9.8 that page is redirected to the onboarding checklist, so it's no longer reachable.
- Get number of recommendations to display from the API instead of using a hard-coded value

Closes [WCCOM-1177](https://linear.app/a8c/issue/WCCOM-1177/update-non-marketplace-placements-to-use-api#comment-348a9b5e)

1. Check out this branch
2. Navigate to `/plugins/woocommerce` and run `pnpm --filter='@woocommerce/plugin-woocommerce' build`.
3. Make sure your local WCCOM is running on the `update/WCCOM-1177-recommendations-non-marketplace-placements` branch, and connect your Core env to your local WCCOM using `docker network connect woocommerce.test <container id>` (container-id for the `wordpress` container in the Core wp-env).
4. Change the endpoint URL here to woocommerce.test:
https://github.com/woocommerce/woocommerce/blob/23182138e0ada04d34f8b2aa9406e06b6a1f079f/plugins/woocommerce/includes/admin/marketplace-suggestions/class-wc-marketplace-updater.php#L53
5. Update this condition to always be true:
https://github.com/woocommerce/woocommerce/blob/23182138e0ada04d34f8b2aa9406e06b6a1f079f/plugins/woocommerce/includes/admin/marketplace-suggestions/class-wc-marketplace-suggestions.php#L189
6. Visit http://localhost:8888/wp-admin/admin.php?page=wc-orders, then go to http://localhost:8888/wp-admin/tools.php?page=action-scheduler&s=woocommerce_update_marketplace_suggestions&bulk_action=Apply&paged=1 and run the pending action
7. Refresh the Orders page and confirm you now see suggestions returned from Algolia instead of the hard-coded ones from the JSON file on WCCOM.
Orders_‹_woocommerce_—_WordPress

8. Create a new product, and view the Get more options tab of Product Data section. It should also show Algolia recommendations.
Add_new_product_‹_woocommerce_—_WordPress

9. On WCCOM, increase the `RECOMMENDATIONS_COUNT` value (see testing instructions for https://github.com/Automattic/woocommerce.com/pull/23982) and clear the cache. Then refresh the Orders page and run the scheduled action again to fetch new recommendations.
10. On both the empty Order screen and the product editor you should now see more than five recommendations returned, depending on the number you set in step 9, as well as other conditions included in the data (some of the hard-coded recommendations for the product editor only show if particular plugins are already active on the site)
11. Change the endpoint you modified in step 4 to the 1.0 endpoint. Repeat the steps to trigger the action and re-fetch recommendations
12. Confirm you now see the five hard-coded recommendations in each location.
Orders_‹_WP_Test_Site_—_WordPress

13. Please test around this issue

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/legacy/js/admin/marketplace-suggestions.js
+++ b/plugins/woocommerce/client/legacy/js/admin/marketplace-suggestions.js
@@ -349,6 +349,12 @@
 		function displaySuggestions( marketplaceSuggestionsApiData ) {
 			var usedSuggestionsContexts = [];
 
+			// Extract recommendations count from the first element if it exists
+			var recommendationsCount = 5; // default fallback
+			if ( marketplace_suggestions.suggestions_data && marketplace_suggestions.suggestions_data[0]['recommendations-count'] ) {
+				recommendationsCount = marketplace_suggestions.suggestions_data[0]['recommendations-count'];
+			}
+
 			// iterate over all suggestions containers, rendering promos
 			$( '.marketplace-suggestions-container' ).each( function() {
 				// determine the context / placement we're populating
@@ -357,8 +363,8 @@
 				// find promotions that target this context
 				var promos = getRelevantPromotions( marketplaceSuggestionsApiData, context );
 
-				// shuffle/randomly select five suggestions to display
-				var suggestionsToDisplay = _.sample( promos, 5 );
+				// shuffle/randomly select suggestions to display based on API count
+				var suggestionsToDisplay = _.sample( promos, recommendationsCount );
 
 				// render the promo content
 				for ( var i in suggestionsToDisplay ) {
--- a/plugins/woocommerce/client/legacy/js/admin/marketplace-suggestions.js
+++ b/plugins/woocommerce/client/legacy/js/admin/marketplace-suggestions.js
@@ -352,7 +352,11 @@
 			// Extract recommendations count from the first element if it exists
 			var recommendationsCount = 5; // default fallback
 			if ( marketplace_suggestions.suggestions_data && marketplace_suggestions.suggestions_data[0]['recommendations-count'] ) {
-				recommendationsCount = marketplace_suggestions.suggestions_data[0]['recommendations-count'];
+				var apiCount = marketplace_suggestions.suggestions_data[0]['recommendations-count'];
+				// Validate that it's a positive number and within reasonable bounds
+				if ( typeof apiCount === 'number' && apiCount > 0 && apiCount <= 50 ) {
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/client/legacy/js/admin/marketplace-suggestions.js
```diff
@@ -349,6 +349,16 @@
 		function displaySuggestions( marketplaceSuggestionsApiData ) {
 			var usedSuggestionsContexts = [];
 
+			// Extract recommendations count from the first element if it exists
+			var recommendationsCount = 5; // default fallback
+			if ( marketplace_suggestions.suggestions_data && marketplace_suggestions.suggestions_data[0]['recommendations-count'] ) {
```
bor0: I suppose it'd be better if we made `suggestions_data` a JSON object instead of an array, but as we discussed, this might affect BWC, so I'm not sure if there's anything we can do about it. Creating a new JSON is acceptable, but adds more stuff to maintain, so 🤷

mcliwanow: +1 for maintaining BWC. This will allow us to update 1.0 endpoint with better recommendations too in the future, if we don't break the data structure.

mcliwanow: Not sure how likely this is, but we might be even more defensive:
```
if ( marketplace_suggestions.suggestions_data && marketplace_suggestions.suggestions_data[0] && marketplace_suggestions.suggestions_data[0]['recommendations-count'] ) {
```

mcliwanow (approved):
It took me a while to test this one, but everything seems to work as expected 👍  Recommendations count is based on API response and I can see dynamic recommendations coming from 2.0/suggestions endpoint

plugins/woocommerce/client/legacy/js/admin/marketplace-suggestions.js
KokkieH: > we might be even more defensive

Definitely no harm in being cautious. Added in a310175

PR #58517, opened 2025-06-04
Product Filter Rating: replace shared Rating component with local RatingStars component

dinhtungdu:

This PR refactors the star component of the Rating filter block. I decided to use the local RatingStars component instead of reusing the shared component. Doing so has several benefits:
- Reduce CSS. In fact, it removed the entire CSS file for the Rating filter block.
- The shared rating component does too much. Using the local component simplifies the implementation without the risk of adding more unnecessary things to the shared component. 

Closes WOOPLUG-4439

| Before | After |
| ------ | ----- |
|    image    |   image    |
| image | image |

1. Add Product Filters to the Product Catalog template.
2. See the Rating filter block use the new component.
3. The count doesn't show in the editor, but it will be fixed by https://github.com/woocommerce/woocommerce/pull/58507.
4. See the updated rating stars on the frontend.


plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterCheckboxList.php
```diff
@@ -99,7 +99,7 @@ class="wc-block-product-filter-checkbox-list__input"
 							</span>
 							<span class="wc-block-product-filter-checkbox-list__text-wrapper">
 								<span class="wc-block-product-filter-checkbox-list__text">
-									<?php echo wp_kses_post( $item['label'] ); ?>
```
gigitux, line 102: Is there a particular reason for this change? I think that we should escape this value. If I'm not wrong, context can easily be updated via `render_block_context` hook.

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterCheckboxList.php
```diff
@@ -57,7 +57,7 @@ function ( $item ) {
 			'data-wp-key'         => wp_unique_prefixed_id( $this->get_full_block_name() ),
 			'data-wp-context'     => '{}',
 			'class'               => esc_attr( $classes ),
-			'style'               => esc_attr( $style ),
+			'style'               => esc_attr( $style ) . ';',
```
gigitux: Could you clarify the reason for this change? Currently, when style is empty, the generated HTML looks like this:

```
<div data-block-name="woocommerce/product-filter-checkbox-list" style=";" class="wp-block-woocommerce-product-filter-checkbox-list wc-block-product-filter-checkbox-list wp-block-woocommerce-product-filter-checkbox-list" data-wp-interactive="woocommerce/product-filters" data-wp-key="woocommerce/product-filter-checkbox-list4" data-wp-context="{}">
```

gigitux (commented):
Thanks for working on this. The PR works as expected:

| Editor | Frontend |
|--------|--------|
|Image| Image | 


I have some doubts regarding some changes - can you take a look?

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterCheckboxList.php
dinhtungdu: Ah, I should check if the style is empty or not, will update it.

The reason for that is, the generated style by Supports API doesn't have the `;` at the end. When passing it to `get_block_wrapper_attributes()`, we need to append the `;`, otherwise, the style will break.

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterCheckboxList.php
dinhtungdu: I removed it because `wp_kses_post` strips out `svg`.

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterCheckboxList.php
dinhtungdu: Updated in [12843ff](https://github.com/woocommerce/woocommerce/pull/58517/commits/12843ff01568d8c59626917e41360befa777114a)

gigitux: Ah it makes sense! Thanks! 🙇

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterCheckboxList.php
dinhtungdu: > context can easily be updated via render_block_context hook.

Yeah, but that's server side code. Can you explain in more details what we should worry about?

gigitux: I might be overthinking this, but a poorly designed extension could modify the context, causing the block to output dangerous HTML. That said, the same risk exists with any filter like `render_block`. All clear now—thanks for the clarification, @dinhtungdu!

gigitux (approved):
LGTM! Thanks for updating the PR and for the clarification! :shipit:

plugins/woocommerce/src/Blocks/BlockTypes/ProductFilterCheckboxList.php
dinhtungdu: Yeah, like you said the same can happen with `render_block`. In this case, we pull the filter data from lookup tables, process, and send it to display blocks through context. We don't deal with user input, so I think it's safe to echo the HTML.

PR #58525, opened 2025-06-04
Product Filters: improve Checkbox List color setting, now allowing setting label color only

dinhtungdu:

This PR improves the Checkbox List color setting:
- Removes the Text and Background settings.
- Adds new color setting for Label, which affect the label only, not the count or checkbox element.

The motivation of this change:
- The label of the list is more important than the count and checkbox element, so its color should be changable.
- To support changing color of the rating stars.
- The default color can be set at the filter block level (Attribute/Status/Rating) or is ancestor, and the Checkbox List block will inherit that color.

image

1. Add Product Filters to the Product Catalog template.
2. Enable product count for Attribute/Status/Rating filters block.
3. Select a List block.
4. Open the inspector settings.
5. See the Label color setting.
6. Change the color, see editor reacts.
7. Save, see the color applies on the front end.

### DIFF
--- a/plugins/woocommerce/changelog/feat-list-label-color
+++ b/plugins/woocommerce/changelog/feat-list-label-color
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+Product Filters: improve Checkbox List color setting, now allowing setting label color only
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/inner-blocks/checkbox-list/block.json
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/inner-blocks/checkbox-list/block.json
@@ -13,9 +13,6 @@
 		"woocommerce/product-filter-rating"
 	],
 	"supports": {
-		"color": {
-			"enableContrastChecker": false
-		},
 		"interactivity": true
 	},
 	"usesContext": [ "filterData" ],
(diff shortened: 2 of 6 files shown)


dinhtungdu:
@shivanarrthine Can you take a look at this PR please as it changes the UX of the block? 🙏🏽

kmanijak (approved):
I tested the solution and it works fine! I can style the labels separately and it doesn't include count:

Editor | Frontend
-- | --
image | image

I'm approving, however could you explain why we need both `labelElement` and `customLabelElement` while one seems to be a fallback for another. Wouldn't it be sufficient to have just `labelElement`? What am I missing? 😅

dinhtungdu:
I followed this example to implement the color setting [https://developer.wordpress.org/news/2023/11/how-to-add-custom-color-options-to-blocks/](https://developer.wordpress.org/news/2023/11/how-to-add-custom-color-options-to-blocks/). About your question, here is the reason for using two attributes:

> Technically, you can use a single attribute for this and add some custom logic behind the scenes for handling both presets and custom colors. But using two separate attributes makes for a simpler codebase. You’ll probably find that there are many different ways of wrangling this once you start customizing this for your own blocks.

dinhtungdu:
Probably with 4 colors and 8 attributes, this is becoming harder to maintain. Let me take a look.

shivanarrthine:
@dinhtungdu I tested it, and it works great! :)

kmanijak:
> I followed this example to implement the color setting https://developer.wordpress.org/news/2023/11/how-to-add-custom-color-options-to-blocks/. About your question, here is the reason for using two attributes:
> ...
> Probably with 4 colors and 8 attributes, this is becoming harder to maintain. Let me take a look.

@dinhtungdu, I wasn't aware of this examples and guidelines. And with this explanation I'm totally fine with this approach, no need to reinventing the wheel! 🙌

Issue #58482, opened 2025-06-02
Investigate `fetch_patterns` scheduled on every admin page load


kaushikasomaiya:
## **Describe the bug**

We've got two reports so far

* [https://wordpress.org/support/topic/scheduled-action-fetch_patterns-runs-every-few-seconds/page/2/](https://wordpress.org/support/topic/scheduled-action-fetch_patterns-runs-every-few-seconds/page/2/)
* #9842349-zen

Suspect I think is:

$last_request = get_transient( 'last_fetch_patterns_request' );

In the second case, there was no presence of this transient in the database making me wonder it is likely some sort of persistent caching storing false value of the transient thus the action is scheduled repeatedly.

However, `set_transient( 'last_fetch_patterns_request', time(), HOUR_IN_SECONDS )` should then correctly set it unless there is an issue with host 🤔  I will add anything else as I discover.

Just opening this for awareness and investigation.

gigitux:
I'm closing this issue given that it will be handled as part of WOOPLUG-4456

PR #58528, opened 2025-06-04
Refactor PTKPatternsStore to use options instead of transients for PTK patterns storage

gigitux:

This PR refactor the `PTKPatternsStore` class to use options instead of the transients. For this use case, transients aren't good because there isn't the guarantee that they are always available. This causes some issues, like https://github.com/woocommerce/woocommerce/issues/58482. Furthermore, transients are always autoloaded instead, with options, autoload can be disabled, increasing the performance.

Closes # .

| Before | After |
| ------ | ----- |
|        |       |

### New site

1. Start from a fresh site. Install this zip
[woocommerce.zip](https://github.com/user-attachments/files/20592108/woocommerce.zip)
2. Visit `/wp-admin/admin.php?page=wc-settings&tab=advanced&section=woocommerce_com`
3. Enable the tracking.
4. Visit the `/wp-admin/admin.php?page=wc-status&tab=action-scheduler`.
5. Ensure that the `fetch_patterns` action is scheduled.
6. Install [Options View](https://wordpress.org/plugins/options-view/).
7. Visit `/wp-admin/tools.php?page=optionsview`
8. Check that the `ptk_patterns` option has a json with patterns.
9. Check that the `last_fetch_patterns_request` option has unix timestamp.
10. Remove the `ptk_patterns` option. 
11. Visit `/wp-admin/admin.php?page=wc-settings&tab=advanced&section=woocommerce_com`.
12. Disable and enable tracking.
13. Visit the `/wp-admin/admin.php?page=wc-status&tab=action-scheduler`.
14. Ensure that the `fetch_patterns` action **is not scheduled**.
15. Visit `/wp-admin/tools.php?page=optionsview`
16. Ensure that the `ptk_patterns` option doesn't exist.
17. Update the `last_fetch_patterns_request` option with a unix timestamp older than one day ( you can use `1717498972`)
18. Visit `/wp-admin/admin.php?page=wc-settings&tab=advanced&section=woocommerce_com`.
19. Disable and enable tracking.
20.  Visit the `/wp-admin/admin.php?page=wc-status&tab=action-scheduler`.
21. Ensure that the `fetch_patterns` action is scheduled.

### Site with ptk_patterns transient

1. Install WooCommerce 9.9.
2. Visit `/wp-admin/admin.php?page=wc-settings&tab=advanced&section=woocommerce_com`
3. Enable the tracking.
4. Install [Transients Manager](https://wordpress.org/plugins/transients-manager/)
5. Visit `/wp-admin/tools.php?page=transients-manager`.
6. Search for the `ptk_patterns`.
7. Ensure that `ptk_patterns` is present.
8. Update WooCommerce: [woocommerce.zip](https://github.com/user-attachments/files/20592108/woocommerce.zip)
9. Visit `/wp-admin/tools.php?page=transients-manager`
10. Search for the `ptk_patterns`.
11. Ensure that `ptk_patterns` is not present.
12. Install [Options View](https://wordpress.org/plugins/options-view/)
13. Visit `/wp-admin/tools.php?page=optionsview`
14. Check that the `ptk_patterns` option has a json with patterns.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 
Refactor PTKPatternsStore to use options instead of transients for PTK patterns storage.

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/includes/class-wc-install.php
```diff
@@ -197,99 +197,102 @@ class WC_Install {
...
 			'wc_update_985_enable_new_payments_settings_page_feature',
 		),
-		'9.9.0' => array(
+		'9.9.0'  => array(
 			'wc_update_990_remove_wc_count_comments_transient',
 			'wc_update_990_remove_email_notes',
 		),
+		'10.0.0' => array(
+			'wc_update_1000_remove_patterns_toolkit_transient',
+		),
```
gigitux, line 295: I added only this line

plugins/woocommerce/includes/wc-update-functions.php
```diff
@@ -3031,3 +3031,14 @@ function wc_update_990_remove_email_notes() {
...
+/**
+ * Remove the transient ptk_patterns.
+ * This was used to store the Patterns Toolkit patterns in the database.
+ * The patterns are now stored in the option ptk_patterns.
+ *
+ * @return void
+ */
+function wc_update_1000_remove_patterns_toolkit_transient() {
+	delete_transient( 'ptk_patterns' );
+}
```
gigitux, line 3044: It doesn't make sense to keep the transient in the database, given that these will be read from options. 

Also, once the update, PTK patterns will be refetched:

https://github.com/woocommerce/woocommerce/blob/5e2572cecdaea9742c394020e0ed027c2678be98/plugins/woocommerce/src/Blocks/Patterns/PTKPatternsStore.php#L32-L53

josephscott:
I tested this in a different way, in an effort to make sure it was only this change that was making a difference.

Here are the top level steps:

- A fresh WordPress 6.8.1 install on my laptop
- Installed and activated WooCommerce 9.8.5 ( selected the location and such, just enough to get it active )
- Confirmed the PTK Patterns transient in the wp_options table, with autoload on
- Ran `oha -z 60s https://localhost:7770/` from my laptop against the WP+WooC test install, repeated a few times

That gave me a baseline for what things were like with the PTK Patterns transient autoloading on every request.  Then I setup the change in this PR to see how it compared

- Deleted the PTK Patterns transient from the wp_options table
- Deleted last_fetch_patterns_request from the wp_options table
- Applied changes to PTKPatternsStore.php from this PR
- Loaded wp-admin and front of site a few times
- Confirmed the `ptk_patterns` entry in the wp_options table existed, with autoload off
- Ran `oha -z 60s https://localhost:7770/` again, a few times

No object cache, no other plugins active.

I am seeing about a 1.5% improvement in requests per second and p75 response time without autoloading the PTK Patterns data on every request.  There was some variability between runs of oha though, so this could also be as wash in other environments.

Looking just at the `wp_load_alloptions()` query of:

```sql
SELECT option_name, option_value
FROM wp_options
WHERE autoload IN ( 'yes', 'on', 'auto-on', 'auto' )
```

Not loading that one PTK Patterns entry reduced the query time by nearly 50%.


Since this is data that is needed only in specific areas, I do think that not autoloading it is still the right thing to do.

dinhtungdu:
@gigitux thanks for the PR. I always love performance improvement, and Joseph tests above proves the gain we got. But I'm not fully convinced that switching switching from transient to option is the way to go.

> transients are always autoloaded

I don't think this is always true. Transients with expiration are not autoload. See https://github.com/WordPress/wordpress-develop/blob/085bdca192498cc1ac492931c4ad3e05407a33a7/src/wp-includes/option.php#L1312-L1315

> there isn't the guarantee that they are always available

Can you explain why? Transient is just an option after all, what makes option more reliable than transient?

gigitux:
> I don't think this is always true. Transients with expiration are not autoload. See https://github.com/WordPress/wordpress-develop/blob/085bdca192498cc1ac492931c4ad3e05407a33a7/src/wp-includes/option.php#L1312-L1315
> 
> >

TIL

> Can you explain why? Transient is just an option after all, what makes option more reliable than transient?



Based on host configuration, transients could not be stored in the database. You can read pcShBQ-2ES-p2#comment-4211 and https://pressidium.com/blog/wordpress-object-caching-redis-memcached-and-native-apis/

dinhtungdu:
> Based on host configuration, transients could not be stored in the database.

To me, not storing in the database is not a valid reason to not use transient. It can be stored in memory or anywhere and it's fine. It's not fine only when it's not stored anywhere.

About the Darren comment you linked above, it's the reason for validating the transient data, not really for not using it. We already add validation in response to the incident. So it's again not a strong point for not using transient to me.

Taking https://github.com/woocommerce/woocommerce/issues/58482 for consideration, `set_transient( 'last_fetch_patterns_request', time(), HOUR_IN_SECONDS )` seem to fail to set the transient and we haven't known why, yet. We know that transients can disappear at any time and you can’t predict when that will occur, but it's not the same as transient will disappear right away after you set it. It seems to hosting configuration to me than transient limitation.

I still think transient is the right tool for this job, but we're using it wrong by not setting the expiration for `ptk_patterns`. But let's be real that using option might be a good alternative too as long as we handle the cleanup correctly. Knowing why the transient is failed to set in #58482 may not worth the effort finding it.

With all that said, I'm still on the fence as I'm not fully convinced. Let's me loop in other for more perspective.

gigitux:
> Taking #58482 for consideration, `set_transient( 'last_fetch_patterns_request', time(), HOUR_IN_SECONDS )` seem to fail to set the transient and we haven't known why, yet. We know that transients can disappear at any time and you can’t predict when that will occur, but it's not the same as transient will disappear right away after you set it. It seems to hosting configuration to me than transient limitation.

I don’t think this is a transient-related limitation. However, in some cases, if the transient expires too early (even caused by hosting configuration), it can lead to an infinite loop of fetching PKT patterns similar to what happens in #58482. I think that we should avoid this. 

> With all that said, I'm still on the fence as I'm not fully convinced. Let's me loop in other for more perspective.

Thanks for your feedback and put in the loop other folks 🙇 To give you more context, we discussed this along with PTK incident and there is some alignment on [this migration](https://linear.app/a8c/issue/WOOPLUG-4486/investigate-fetch-patterns-scheduled-on-every-admin-page-load#comment-77d6e471).

Aljullu:
I missed some discussions following up the incident, so I don't have the full context of what was decided. Similarly to @dinhtungdu, I feel transients are better suited for this use case than options. But ultimately, they are both the same, so I understand we can use one or the other without major concerns as long as we keep track of invalidation, cleaning them up, etc. So not a strong opinion from my side in one direction or the other.

---

This is probably out of the scope for this PR, and I'm not sure if it was discussed, but I think WC core patterns should be stored in a file, not in an option nor transient. That file should be shipped with WooCommerce, and it should be specific to each WooCommerce version. Currently, we are shipping the same patterns to all WooCommerce stores, independently of their version.

plugins/woocommerce/src/Blocks/Patterns/PTKPatternsStore.php
```diff
@@ -176,7 +187,7 @@ public function fetch_patterns_on_plugin_update( $upgrader_object, $options ) {
 	 * @return void
 	 */
 	public function flush_cached_patterns() {
-		delete_transient( self::TRANSIENT_NAME );
+		delete_option( self::OPTION_NAME );
```
dinhtungdu, line 191: I think we should also delete the `last_fetch_patterns_request` option. So if the tracking is disabled accidentally then activated again right after, the patterns can be fetched right away without waiting for another day.

dinhtungdu (approved):
Thanks for the PR and explanation! I followed the testing instruction and can verify the PTK patterns are fetched as described. There is a minor issue I pointed out in the review comment, but otherwise, this is LGTM!

plugins/woocommerce/src/Blocks/Patterns/PTKPatternsStore.php
gigitux: Good point! Fixed with a200530

dinhtungdu: Thanks for the update! LGTM! The failed metric workflow seems flaky.

PR #58575, opened 2025-06-05
Improve checks for missing context in Variation Selector blocks

Aljullu:

The way blocks are parsed is from the most inner one to the outer ones. In our case, we are passing context from the parent blocks to the children, so there is the chance that we might try to render blocks when context hasn't been set.

We already had some checks that would return early if context wasn't set:

https://github.com/woocommerce/woocommerce/blob/8603154be0d36135d19382c1e915015cd005e3f7/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/VariationSelectorAttributeName.php#L34-L36

However, there is the case that the Add to Cart + Options block can be added inside a post. In which case all blocks get a `postId` context. In those cases, `empty( $block->context )` is no longer true and we weren't returning early. That caused some warnings to appear in certain cases (see testing steps).

_tl;dr_ this PR strengthens the context checks, so instead of simply checking if `$block->context` is not empty, we make sure the data which is required by blocks actually exists.

#### Preparation

#### Testing the PR

1. Create a post, add the _Single Product_ block, select a variable product and update the add to cart form to the blockified _Add to Cart + Options_ block.
2. Go to your site's front page or blog page (any page that lists the posts).
3. Verify there are no warnings on the screen.

Before | After
--- | ---
imatge | imatge

### DIFF
--- a/plugins/woocommerce/changelog/fix-add-to-cart-options-better-context-checks
+++ b/plugins/woocommerce/changelog/fix-add-to-cart-options-better-context-checks
@@ -1,0 +1,5 @@
+Significance: patch
+Type: fix
+Comment: Improve checks for missing context in Variation Selector blocks
+
+
--- a/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/VariationSelectorAttributeName.php
+++ b/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/VariationSelectorAttributeName.php
@@ -35,8 +35,8 @@ class VariationSelectorAttributeName extends AbstractBlock {
 			return '';
 		}
 
-		$attribute_id   = $block->context['woocommerce/attributeId'];
-		$attribute_name = $block->context['woocommerce/attributeName'];
+		$attribute_id   = isset( $block->context['woocommerce/attributeId'] ) ? $block->context['woocommerce/attributeId'] : null;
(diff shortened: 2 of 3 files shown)


kmanijak (approved):
Great fix! In my case, when trying on `trunk` I even encountered a fatal, not only warning.

```
Fatal error: Uncaught Error: array_merge(): Argument #2 must be of type array, null given
in /Users/karolm/automattic/repositories/woocommerce/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/VariationSelectorAttributeOptions.php on line 210

Call stack:

array_merge()
/Users/karolm/automattic/repositories/woocommerce/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/VariationSelectorAttributeOptions.php:210
Automattic\W\B\B\A\VariationSelectorAttributeOptions::render_dropdown()
/Users/karolm/automattic/repositories/woocommerce/plugins/woocommerce/src/Blocks/BlockTypes/AddToCartWithOptions/VariationSelectorAttributeOptions.php:72
Automattic\W\B\B\A\VariationSelectorAttributeOptions::render()
/Users/karolm/automattic/repositories/woocommerce/plugins/woocommerce/src/Blocks/BlockTypes/AbstractBlock.php:98
```

But I confirm changes in this PR fix it and the Home Page displaying the post is displayed as expected with no errors or warnings 👍

PR #57373, opened 2025-04-17
Fix mobile unresponsiveness of product gallery after ‘Reviews’ tab click

gigitux:
### Submission Review Guidelines:

-   I have followed the [WooCommerce Contributing Guidelines](https://github.com/woocommerce/woocommerce/blob/trunk/.github/CONTRIBUTING.md) and the [WordPress Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/).
-   I have checked to ensure there aren't other open [Pull Requests](https://github.com/woocommerce/woocommerce/pulls) for the same update/change.
-   I have reviewed my code for [security best practices](https://developer.wordpress.org/apis/security/).
-   Following the above guidelines will result in quick merges and clear and detailed feedback when appropriate.

### Changes proposed in this Pull Request:

This PR addresses a regression introduced in [#53158](https://github.com/woocommerce/woocommerce/pull/53158), where, in mobile view, the touch event on the description and reviews tabs sets the animation to `true` without resetting it. This prevents the animation logic in [jquery.flexslider.js](https://github.com/woocommerce/woocommerce/blob/f3eac49e1d8bd39fbfde10bc9a037e5131a27775/plugins/woocommerce/client/legacy/js/flexslider/jquery.flexslider.js#L619-L724) from executing, causing the animation to fail.

Additionally, this PR resolves the issue noted in [#53018](https://github.com/woocommerce/woocommerce/pull/53018#issuecomment-2498921373). After investigation, the following fixes were implemented:

- Use the `keyup` event, consistent with [FlexSlider's implementation](https://github.com/woocommerce/woocommerce/blob/297c33addb4d0e7faa9e4e4b1fd2400c2c84b2b8/plugins/woocommerce/client/legacy/js/flexslider/jquery.flexslider.js#L140-L154), to handle keyboard interactions.
- Now, that the same event is used, prevent event bubbling to the gallery via `stopPropagation`, ensuring that keyboard updates to the stars do not trigger unintended gallery behavior.

Given that there is a back and forth with these changes, I created E2E tests to covere these scenario.

These changes fix the regression and ensure proper behavior for both touch and keyboard interactions.

I'm going to see how we can fix it, but sharing in case you have a better idea to avoid setting animating to true.

Closes #57292 (WOOPLUG-3943) .

(For Bug Fixes) Bug introduced in PR #53158 .

### How to test the changes in this Pull Request:

[...]

PR #58598, opened 2025-06-06
Add to Cart + Options block: Fix gallery keyboard navigation

### LINKED ISSUE #57682: Add to Cart + Options block: Navigating between variations with the keyboard causes the gallery images to change
When using the arrow keys to navigate between variations in the *Add to Cart with Options Variation Selector*, the *Product Image Gallery* block triggers an image swap.

### Steps to reproduce

1. Install and activate the [WooCommerce Beta Tester](https://woocommerce.com/products/woocommerce-beta-tester/) plugin.
2. Go to *Tools* > *WCA Test Helper* > *Features* and enable `experimental-blocks` and `blockified-add-to-cart`.
3. Go to *Appearance* > *Editor* > *Templates* > *Single Product* and update the *Add to Cart with Options* block to the blockified version.
4. Make sure in the same template there is the *Product Image Gallery* block.
5. Visit a variable product in the frontend.
6. With the keyboard arrow keys, navigate through the variation options.
7. Notice the images from the gallery switch when you press the left or right keys even though the variation pills have focus.

https://github.com/user-attachments/assets/4e9ed26c-f6fd-4e57-92fe-638414b8ba38

Expected behavior: when the pills have focus, the _Product Image Gallery_ shouldn't switch images.

mikachan:

Closes https://github.com/woocommerce/woocommerce/issues/57682.

This attempts to fix the issue described in https://github.com/woocommerce/woocommerce/issues/57682, where the product gallery slider always navigates when the keyboard left and right arrows are pressed, regardless of whether the slider is in focus or not.

The primary fix is updating the FlexSlider implementation in `single-product.js` to disable the default keyboard navigation and use a custom keyboard navigation which only listens to the left and right arrow keys if the gallery is in focus.

I've also updated the variation pills to include `aria-checked` and `tabindex` attributes, which I believe should improve the accessibility of these pills following the guidelines from [this w3 example](https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio/). If it's easier, I can create a separate PR for this instead.

| Before | After |
| ------ | ----- |
| <video src="https://github.com/user-attachments/assets/7fe88611-9734-42f3-8a8e-2b59fd54f92b" /> | <video src="https://github.com/user-attachments/assets/2159fd99-b9a6-428c-9a3a-168c4c0b398b" /> |

1. Install and activate the [WooCommerce Beta Tester](https://woocommerce.com/products/woocommerce-beta-tester/) plugin.
2. Go to *Tools* > *WCA Test Helper* > *Features* and enable `experimental-blocks` and `blockified-add-to-cart`.
3. Go to *Appearance* > *Editor* > *Templates* > *Single Product* and update the *Add to Cart with Options* block to the blockified version.
4. Make sure in the same template there is the *Product Image Gallery* block.
5. Visit a variable product in the frontend.
6. With the keyboard arrow keys, navigate through the variation options.
7. Ensure that the images from the gallery do not switch at the same time as navigating through the variation options.
8. Use the keyboard to focus the product gallery.
9. Use the left or right arrow keys to navigate the gallery and ensure the gallery images switch correctly.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [x] Minor
-   [ ] Major

#### Type

#### Message 

Blockified Add to Cart with Options block: only allow the product gallery to navigate via the keyboard left and right arrows when the gallery is in focus.

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/legacy/js/frontend/single-product.js
+++ b/plugins/woocommerce/client/legacy/js/frontend/single-product.js
@@ -239,7 +239,9 @@ jQuery( function( $ ) {
 			},
 			after: function( slider ) {
 				gallery.initZoomForTarget( gallery.$images.eq( slider.currentSlide ) );
-			}
+			},
+			// Disable the built-in keyboard navigation and use custom keyboard navigation instead.
+			keyboard: false,
 		}, args );
 
 		$target.flexslider( options );
@@ -264,7 +266,7 @@ jQuery( function( $ ) {
 			}
 		} );
 
-		// Add custom keyboard navigation that applies only to the slider container.
+		// Custom keyboard navigation that applies only when the slider container is focused.
 		$target
 			.on( 'keydown', function ( e ) {
 				if ( e.which === 39 ) {
@@ -279,7 +281,6 @@ jQuery( function( $ ) {
 					e.stopPropagation();
 				}
 			} )
-			// Make the slider focusable.
 			.attr( 'tabindex', '0' );
 	};
 
--- a/plugins/woocommerce/client/legacy/js/frontend/single-product.js
+++ b/plugins/woocommerce/client/legacy/js/frontend/single-product.js
@@ -263,6 +263,24 @@ jQuery( function( $ ) {
 				$( this ).trigger( 'load' );
 			}
 		} );
+
+		// Add custom keyboard navigation that applies only to the slider container.
+		$target
+			.on( 'keydown', function ( e ) {
+				if ( e.which === 39 ) {
+					// Right arrow
+					$( this ).flexslider( 'next' );
+					e.preventDefault();
+					e.stopPropagation();
+				} else if ( e.which === 37 ) {
+					// Left arrow
+					$( this ).flexslider( 'prev' );
+					e.preventDefault();
+					e.stopPropagation();
+				}
+			} )
+			// Make the slider focusable.
+			.attr( 'tabindex', '0' );
 	};
 
 	/**
--- a/plugins/woocommerce/changelog/58598-fix-product-gallery-keyboard-nav
+++ b/plugins/woocommerce/changelog/58598-fix-product-gallery-keyboard-nav
@@ -1,0 +1,4 @@
+Significance: minor
+Type: fix
+
+Blockified Add to Cart with Options block: only allow the product gallery to navigate via the keyboard left and right arrows when the gallery is in focus.
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts
@@ -51,15 +51,6 @@ function setAttribute( name: string, value: string | null ) {
 
 function setDefaultSelectedAttribute() {
 	const context = getContext< Context >();
-
-	// Select the first attribute option if no value is selected.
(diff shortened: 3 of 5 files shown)


plugins/woocommerce/client/legacy/js/frontend/single-product.js
```diff
@@ -263,6 +265,23 @@ jQuery( function( $ ) {
 				$( this ).trigger( 'load' );
 			}
 		} );
+
+		// Custom keyboard navigation that applies only when the slider container is focused.
```
gigitux, line 269: I’m happy to merge this, as it appears to follow the correct accessibility behavior.
However, do you think there’s a risk of breaking something?
Some extensions might rely on the current implementation. For example, here’s what happens when I run this code:


```js
const event = jQuery.Event("keyup", {
  key: "ArrowRight",
  code: "ArrowRight",
  keyCode: 39,
  which: 39,
});



jQuery("body").trigger(event);
```


| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/1b0a1a0a-dadb-467f-bf68-9dbe96148301" /> |<video src="https://github.com/user-attachments/assets/36d8c9b1-b8b1-4533-b6d1-b6d84b1a06c0" /> |

mikachan: The main difference is that the gallery now needs to be in focus for the arrow keys to work as gallery navigation. This could break extensions that rely on the arrow keys to navigate the gallery no matter what is in focus.

How are cases like this usually handled? I'm wondering if there's some outreach we could do to extenders, or maybe this functionality could be put behind a setting?

gigitux: > How are cases like this usually handled? I'm wondering if there's some outreach we could do to extenders, or maybe this functionality could be put behind a setting?

Usually, we evaluate the impact of these changes via https://wpdirectory.net/ and manual testing. 

Given that this is old code, and we are working on a new Product Gallery block, I'm wondering if we could keep the behavior as it is. Instead, we could fix the bug only for the Add to Cart With Options block. For instance, we could try to disable the event when an element of the Add To Cart With Options blocks is focused. I did a similar approach here: https://github.com/woocommerce/woocommerce/pull/57373#issuecomment-2815241231 


What do you think?

mikachan: Yeah I think it's OK to leave the old behaviour as it is, as the gallery is being replaced with the new block. I can't reproduce the bug on the new Product Gallery block as it's caused by FlexSlider, which the new gallery isn't using. 

Perhaps we could still add the `tabindex` changes to the Add to Cart With Options block, and remove the FlexSlider changes from this PR?

gigitux: > Perhaps we could still add the tabindex changes to the Add to Cart With Options block, and remove the FlexSlider changes from this PR?

I'm fine with this! If we get some reports, we can investigate on other approaches or merge this PR. What are your thoughts on opening a dedicated PR for `tabindex` patch? In this way, we can keep this code easily accessible if needed in the future.

mikachan: Yep that's fine, I'll open another PR just for the `tabindex` changes. Thank you!

gigitux: Feel free to add me as a reviewer!

mikachan:
Closing as the issue this was addressing, https://github.com/woocommerce/woocommerce/issues/57682, has been closed.

PR #58600, opened 2025-06-06
Add $order argument to woocommerce_email_downloads_columns filter

### LINKED ISSUE #37757: [Enhancement]: Add order object to woocommerce_email_downloads_columns filter

In the file `templates/emails/plain/email-downloads.php`, the filter `woocommerce_email_downloads_columns` is called like this:

```
$columns   = apply_filters(
			'woocommerce_email_downloads_columns',
			array(
				'download-product' => __( 'Product', 'woocommerce' ),
				'download-expires' => __( 'Expires', 'woocommerce' ),
				'download-file'    => __( 'Download', 'woocommerce' ),
			)
		);
```

Note that whatever hooks the function isn't passed any parameter that relates to the order (and in particular, not `$order`, the WC_Order object).

This makes the filter into something that is "always or never", rather than something that can be varied depending upon any details of the order (which is what I want to achieve).

Proposal: add `$order` as a parameter.

Other possibilities: You can over-write the whole of the `templates/emails/email-downloads.php` and `templates/emails/plain/email-downloads.php` templates, which is a lot of work for a small change.

### Additional context

_No response_

NeosinneR:

Adds a new `$order` argument to  `woocommerce_email_downloads_columns` to allow conditionally add a new column based on the value from order properties. 

Closes WOOPLUG-1200, closes #37757.

1. Start testing on the `trunk` to check the previous behaviour. 
2. Set up a product with downloadable content.
3. Make an order with this product and mark it as completed. 
4. Install **Code Snippets**.
5. Add this snippet
```php
add_filter( 'woocommerce_email_downloads_columns', 'add_recent_order_downloads_column', 10, 2 );
function add_recent_order_downloads_column( $columns, $order ) {
    $columns['new-column'] = 'Order #' . $order->id;
    return $columns;
}
```
6. Send an order detail to the customer - notice that this throws an error (because the filter expects 2 parameters, but only 1 is passed).
7. Check out this branch.
8. Send an order detail to the customer - notice that the downloads section includes a new column which includes order ID.

### DIFF
--- a/plugins/woocommerce/changelog/wooplug-1200-enhancement-add-order-object-to
+++ b/plugins/woocommerce/changelog/wooplug-1200-enhancement-add-order-object-to
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+Add $order argument to woocommerce_email_downloads_columns filter
--- a/plugins/woocommerce/includes/class-wc-emails.php
+++ b/plugins/woocommerce/includes/class-wc-emails.php
@@ -525,7 +525,9 @@ class WC_Emails {
 		 * Filter the columns of the order downloads table.
 		 *
 		 * @since 3.2.0
-		 * @param array $columns Array of columns.
+		 * @since 10.0.0 Added $order parameter.
+		 * @param array    $columns Array of columns.
+		 * @param WC_Order $order  Order object.
 		 */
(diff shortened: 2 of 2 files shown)


yuliyan (approved):
LGTM!

PR #58602, opened 2025-06-06
Remove Add to Cart + Options feature flags

### LINKED ISSUE #58548: Add to Cart + Options block: remove feature gates
We are planning to release the blockified *Add to Cart + Options* block in WooCommerce 10.0. Based on this, we should start removing the feature flags that make it unavailable in public builds.

So this issue is about:

1. Completely removing the `blockified-add-to-cart` feature flag.
2. Making it so *Add to Cart + Options* (and all its inner blocks) and template parts are registered even when `experimental-blocks` is false.

Aljullu:

Closes #58548.
Closes WOOPLUG-4510.

This PR removes the feature flags around the _Add to Cart + Options_ block, so it will available by default in all block themes starting from WooCommerce 10.0. :partying_face: 

1. Make sure you have WooCommerce Beta Tester plugin disabled.
2. Go to Appearance > Editor > Patterns > Add to Cart + Options and verify there are four template parts visible (one for each product type).
3. Go to Appearance > Editor > Templates > Single Product click on the _Add to Cart with Options_ and update it to the blockified _Add to Cart + Options_ block.


plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/index.tsx
```diff
@@ -3,16 +3,19 @@
...
  */
 import metadata from './block.json';
 import AddToCartWithOptionsGroupedProductSelectorEdit from './edit';
 import AddToCartWithOptionsGroupedProductSelectorSave from './save';
-import { shouldBlockifiedAddToCartWithOptionsBeRegistered } from '../utils';
 
-if ( shouldBlockifiedAddToCartWithOptionsBeRegistered ) {
+const isBlockTheme = getSettingWithCoercion( 'isBlockTheme', false, isBoolean );
+
+if ( isBlockTheme ) {
```
dinhtungdu: We can remove this check on the client side. Controlling the registration server side is enough.

dinhtungdu (changes requested):
I'm excited for us to reach this state! Great work! Apart from the review comment, this is looking good to me.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/grouped-product-selector/index.tsx
Aljullu: Good call! I changed that in the last commit.

dinhtungdu (approved):
LGTM! I can confirm that the Add to Cart + Options block is registered and worked as expected for the block theme. The block isn't registered if the classic theme is in-used.

PR #58613, opened 2025-06-06
Adjusted data visibility rules in multisite network

naman03malhotra:
Testing instructions and other details: pcShBQ-3hR-p2
Context: pcShBQ-3hR-p2

Fixes: WOO6-14


plugins/woocommerce/includes/class-wc-install.php
```diff
@@ -290,6 +290,9 @@ class WC_Install {
 			'wc_update_990_remove_wc_count_comments_transient',
 			'wc_update_990_remove_email_notes',
 		),
+		'9.9.1' => array(
```
jorgeatorres: Can we change that to 10.0.0 which is the version this will target?

jorgeatorres (changes requested):

plugins/woocommerce/includes/class-wc-install.php
naman03malhotra: Good catch, https://github.com/woocommerce/woocommerce/pull/58613/commits/1c9d54d60b964b4f515464a0c19a379a654b4b54

jorgeatorres (changes requested):
@naman03malhotra: The change introduced some linting issues (due to misaligned `=>`). Once that's addressed we should be good to go.

jorgeatorres (approved):

PR #58616, opened 2025-06-06
Improved Block Chips Style for Add to Cart + Options and Product Filter Block

### LINKED ISSUE #58550: Add to Cart + Options and Filters blocks: unify chips styling
We currently display 'chips' in two different blocks: *Variation Selector* (inside the *Add to Cart + Options* block) and *Chips* (inside the *Filters* block). However, they look slightly different:

Image

This issue is about unifying their style. I think we should:

1. Keep the text color from the _Add to Cart + Options_ block.
2. Keep the border from the _Filters_ block when not focused or hovered. That should probably use `currentColor` with `opacity` (like using a pseudo-element with `::before` or `::after`).
3. Use the border color from the _Add to Cart + Options_ block when focused or hovered.
4. Make sure the other styling rules are the same: padding, border radius, etc.

Kallyan01:

Unify the visual styling of 'chips' used in two different blocks:

- Variation Selector (inside the Add to Cart + Options block)

- Chips (inside the Filters block)

**Changes :**

- Added text color from the Add to Cart + Options block.
- Added border from the Filters block ( not focused & hovered ).
- Used the border color from the Add to Cart + Options block ( on focused or hovered ).

Closes #58550

| Before | After |
| ------ | ----- |
| chips-style-before | update-chips-style

1. Navigate to Appearance → Editor → Templates → Single Product. In the Document Overview, select the Add to Cart with Options block. Then, open the right sidebar and click Upgrade to the Add to Cart + Options block.

2. Open any product page that supports variations (e.g., color, size). You can now test the chip display using the Add to Cart + Options block (Beta).

3. Create a new Page or Post and insert a Product Collection block (you can choose any collection type).

4. Within the selected collection block, insert a Product Filters block. Inside the filter options, add a Status block and switch its Display Option to Chips. Publish the page and view it to see the chips in action.

image

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


Kallyan01:
@Aljullu Please do review and share you thoughts on the changes . Thanks !

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/inner-blocks/chips/style.scss
```diff
@@ -13,7 +13,20 @@
 	font-size: 0.875em;
 	font-family: inherit;
 	cursor: pointer;
-	color: var(--wc-product-filter-chips-text, currentColor);
+	color: inherit;
+
+	&:focus {
+		outline: 1px solid var(--wp--preset--color--accent-4);
+		outline-offset: 1px;
+	}
```
Aljullu: One thing I would like to try is not to have an `outline` on focus/hover and instead set the border to: `1px solid currentColor`. Do you think that would work?

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/inner-blocks/chips/style.scss
```diff
@@ -27,7 +40,16 @@
 }
 
 .wc-block-product-filter-chips__item[aria-checked="true"] {
-	background: currentColor;
+	background: var(--wp--preset--color--contrast);
+	color: var(--wp--preset--color--base);
```
Aljullu: To invert the background and text color we should ideally do the same we are doing in the _Attribute Options_ block.

* Frontend ([1](https://github.com/woocommerce/woocommerce/blob/b3e36591003e0dda88a19aa572426ebbfc5c1979/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/frontend.ts#L31-L32), [2](https://github.com/woocommerce/woocommerce/blob/b3e36591003e0dda88a19aa572426ebbfc5c1979/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/set-styles.ts))
* [Editor](https://github.com/woocommerce/woocommerce/blob/b3e36591003e0dda88a19aa572426ebbfc5c1979/plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/edit.tsx#L70-L80)

It's a bit verbose from the JS-side, but it guarantees it works in all themes, as `var(--wp--preset--color--contrast)` and `var(--wp--preset--color--base)` might not exist in certain themes.

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/style.scss
```diff
@@ -8,9 +8,8 @@
 }
 
 .wc-block-add-to-cart-with-options-variation-selector-attribute-options__pill {
-	border-width: 1px;
-	border-style: solid;
 	border-radius: 2px;
+	border: 1px solid color-mix(in srgb, currentColor 20%, transparent);
```
Aljullu: I know this is how it was in the Filter chips, but wondering if we can raise that 20% to something higher like 40% or 50% in both cases. What do you think? Otherwise pills are difficult to see in some color combinations.

imatge

Aljullu (commented):
Thanks for working on this, @Kallyan01! It's looking great, and nice to see the styles of the two pills/chips elements being more consistent! :raised_hands: 

I just left a few suggestions that I think would help improve the styles even more.

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/inner-blocks/chips/style.scss
MaggieCabrera: var(--wp--preset--color--accent-4) will have no value on the woostarter theme for example. We can't rely on color names, which is why I suggested to rely on the elements api

Aljullu: > var(--wp--preset--color--accent-4) will have no value on the woostarter theme for example

Yes, we should definitely get rid of this kind of variables. In this specific case, removing the outline would solve it.

> which is why I suggested to rely on the elements api

https://github.com/woocommerce/woocommerce/pull/57260 has been in my todo list to take another look, but I don't think I will have time before WooCommerce 10.0. @mikachan wondering if you have capacity? Given your background, #57260 might be a good one for you.

Having said that, I would keep going ahead with this PR (#58616), as it improves what we have in `trunk` and #57260 might take longer to land. What do you think?

MaggieCabrera: To me none of these are blockers for the theme release, we don't need to nail it on the first go, specially if the user is not going to be able to customize any of these via the UI, we can iterate later. Also, now we can actually make core PRs that will let us get closer to what we want without doing so much work

Kallyan01: image
```css
&:focus {
		outline: 1px solid currentColor;
		outline-offset: 1px;
	}
```

image
```css
&:focus {
		border: 1px solid currentColor;
	}
```


@Aljullu i think outline looks good in this case than border . what's your view on this ?

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/style.scss
Kallyan01: Yes ! that make sense here is the preview after adding 40%
image

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/inner-blocks/chips/style.scss
Aljullu: Yes, good call. I also realized that using border doesn't allow seeing the focus on active pills. Let's leave it with `outline`, but using `currentColor` instead of the CSS variable. Sounds good?

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/style.scss
Aljullu: Much better. :100:

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/inner-blocks/chips/style.scss
```diff
@@ -17,7 +17,20 @@
 	font-size: 0.875em;
```
Aljullu: I would update these values too. Setting the `font-size` to `0.8em` and the `padding` to `0.25em 0.75em` to match the _Add to Cart + Options_ block.

Aljullu (commented):
Thanks for the updates, @Kallyan01! I left a couple of extra comments inline.

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/inner-blocks/chips/style.scss
Kallyan01: Yes ! i've used `currentColor ` and tested , it looks good to me .

Kallyan01:
@Aljullu Implemented the suggested changes , please do take a look and let me know if anything needs improvement . Thanks !

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/inner-blocks/chips/set-styles.ts
```diff
@@ -0,0 +1,88 @@
...
+ * Sets the appropriate styles for variation selector pills to ensure
+ * visibility in both light and dark themes.
+ *
+ * This function swaps the text and background colors for selected pills
+ * to create better contrast in all theme environments.
+ */
+function setStyles(): void {
+	/**
+	 * Get the background color of the body then set it as the background color
+	 * of the Variation Selector selected pills.
```
Aljullu: This should probably say "Product Filter selected pills", no? :smile: 

```suggestion
	 * of the Product Filter selected pills.
```

plugins/woocommerce/client/blocks/assets/js/blocks/add-to-cart-with-options/variation-selector/attribute-options/style.scss
```diff
@@ -25,22 +29,11 @@
...
-				in srgb,
-				var(--wp--preset--color--contrast) 85%,
-				transparent
-			);
-		}
-	}
-
 	&:has(.wc-block-add-to-cart-with-options-variation-selector-attribute-options__pill-input:disabled) {
-		border-color: var(--wc-subtext);
 		color: var(--wc-subtext);
```
Aljullu: I know it wasn't mentioned in the original issue, but given that we are improving this, it might be a good opportunity to do it. :slightly_smiling_face: 

`var(--wc-subtext)` doesn't work very well in some color combinations, and it might be unavailable in some themes. I wonder if we should replace it with the border color, how does it sound?

```suggestion
		color: color-mix(in srgb,currentColor 40%,transparent);
```

Before | After
--- | ---
imatge | imatge

Aljullu (commented):
Good job, @Kallyan01! This is working great, I left two last minor suggestions, but besides that, LGTM! Thanks for all your work on this. :raised_hands: 

<details>

* Theme: Twenty Twenty-Five
* Store creation: June 13th 2025
* Store size: small
</details>

plugins/woocommerce/client/blocks/assets/js/blocks/product-filters/inner-blocks/chips/set-styles.ts
Kallyan01: Yes ! mybad , will fix it .

Aljullu (approved):
Everything looks good to me now. Thanks for your contribution @Kallyan01, this will land in WooCommerce 10.0. :partying_face:

Issue #58364, opened 2025-05-28
"Function wp_is_block_theme was called incorrectly." notice when installed as a network plugin


BrandonXLF:
### Prerequisites

- [x] I have carried out troubleshooting steps and I believe I have found a bug.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

When WooCommerce is installed as a network plugin, the notice:
> Notice: Function wp_is_block_theme was called incorrectly. This function should not be called before the theme directory is registered. Please see [Debugging in WordPress](https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/) for more information. (This message was added in version 6.8.0.) in C:\www\wordpress\wp-includes\functions.php on line 6121

is outputted on every page of the site. This bugs occurs in both 9.8.5 and 9.9.0-rc.1.

### Expected behavior

The notice should not be triggered.

### Actual behavior

The notice is shown on every page of the site.

### Steps to reproduce

1. Install WooCommerce as a network plugin

### WordPress Environment

<details>

<summary>System Status Report  (9.8.5)</summary>

```
### WordPress Environment ###

WordPress address (URL): [Redacted]
Site address (URL): [Redacted]
WC Version: 9.8.5
Legacy REST API Package Version: The Legacy REST API plugin is not installed on this site.
Action Scheduler Version: ✔ 3.9.2
Log Directory Writable: ✔
WP Version: 6.8.1
WP Multisite: ✔
WP Memory Limit: 4 GB
WP Debug Mode: ✔
WP Cron: ✔
Language: en_CA
External object cache: –

### Server Environment ###

Server Info: Apache/2.4.58 (Win64) OpenSSL/3.1.3 PHP/8.1.26
PHP Version: 8.1.26
PHP Post Max Size: 8 MB
PHP Time Limit: 60
PHP Max Input Vars: 1000
cURL Version: 8.4.0
OpenSSL/1.1.1w

SUHOSIN Installed: –
MySQL Version: 11.5.2-MariaDB
Max Upload Size: 1 MB
Default Timezone is UTC: ✔
fsockopen/cURL: ✔
SoapClient: ❌ Your server does not have the SoapClient class enabled - some gateway plugins which use SOAP may not work as expected.
DOMDocument: ✔
GZip: ✔
Multibyte String: ✔
Remote Post: ✔
Remote Get: ✔

### Database ###

[REDACTED]

### Post Type Counts ###

attachment: 147
[...]

PR #58640, opened 2025-06-09
Fix:ensure wp_is_block_theme() is called after theme directories registration 

dinhtungdu:

Closes wooplug-4444.

This PR ensures that we don't call `wp_is_block_theme()` too early which throw the warning as pointed out in https://github.com/woocommerce/woocommerce/issues/58364. 

Smoke test the Customizer in the follow case:
- Directly access the customizer when using block theme.
- Using Customizer with classic theme.

### DIFF
--- a/plugins/woocommerce/includes/class-woocommerce.php
+++ b/plugins/woocommerce/includes/class-woocommerce.php
@@ -285,6 +285,7 @@ final class WooCommerce {
 		register_shutdown_function( array( $this, 'log_errors' ) );
 
 		add_action( 'plugins_loaded', array( $this, 'on_plugins_loaded' ), -1 );
+		add_action( 'plugins_loaded', array( $this, 'init_customizer' ) );
 		add_action( 'plugins_loaded', array( $this, 'init_jetpack_connection_config' ), 1 );
 		add_action( 'admin_notices', array( $this, 'build_dependencies_notice' ) );
 		add_action( 'after_setup_theme', array( $this, 'setup_environment' ) );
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/includes/class-woocommerce.php
```diff
@@ -1381,4 +1382,23 @@ public function get_tracking_history( $old_value, $value ) {
...
+
+	/**
+	 * Initialize the customizer on the plugins_loaded action.
+	 * If WooCommerce is network activated, wp_is_block_theme() will be called too early,
+	 * which cause the warning in #58364. By initializing the customizer on plugins_loaded,
+	 * we ensure that wp_is_block_theme() is called after theme directories registration.
+	 *
+	 * See https://github.com/woocommerce/woocommerce/issues/58364
+	 */
+	public function init_customizer() {
```
Aljullu, line 1395: I would mark the method as `@internal`, as I don't think there is any use case for extensions to use it.

Aljullu (approved):
I wasn't able to reproduce the original issue in a multisite installation, nor in a simple installation. However, the code changes in this PR make sense, and I couldn't find any regressions. I left one minor comment about marking the method as internal, but pre-approving.

<details>

* Theme: Twenty Twenty-Five
* Store size: small
</details>

Aljullu (approved):
Thanks for adding `@internal` to the method, @dinhtungdu! LGTM

> I wasn't able to reproduce the original issue in a multisite installation, nor in a simple installation.

I could finally reproduce it, I think I didn't have WooCommerce enabled at the network level in my first try. :sweat: 

Apparently the warning was visible in any screen, ie _Plugins_:

Before | After
--- | ---
imatge | imatge

PR #58653, opened 2025-06-09
Update the font sizes of the Add to Cart + Options blocks

### LINKED ISSUE #58465: Add to Cart + Options block: revisit font sizes
With a designer, we should revisit the font sizes used in the *Add to Cart + Options* block, as they look a bit out of place in some cases.

**Grouped products**

* Item Selector buttons are larger than Add to cart button (I think we could even consider converting the style of these buttons into links or something like that to make them less noticeable).
* Product titles, prices and Add to cart button are smaller than product description.

**Variable products**

* Attribute name, pills and Add to cart button are smaller than product description.

Aljullu:

Closes #58465.
Closes WOOPLUG-4480.

This PR removes some `fontStyle` attributes we had in Add to Cart + Options template parts and `font-size` properties we had in CSS files, with the goal to better inherit the styles of the page. In general, it makes most blocks to look a bit bigger, but I think that's the way to go, to keep it consistent with the other elements of the page.

#### Preparation

#### Testing the PR

1. Visit an external, a simple, a variable and a grouped product in the frontend.
2. Verify _Add to Cart + Options_ inner blocks inherit the font sizes from the context.
3. Test with different themes to verify there isn't any where things look very out of place.

Product type | Before | After
--- | --- | ---
External product | imatge | imatge
Simple product | imatge | imatge
Variable | imatge | imatge
Grouped product | imatge | imatge


gigitux (approved):
>  In general, it makes most blocks to look a bit bigger, but I think that's the way to go, to keep it consistent with the other elements of the page.

LGTM! I tested it with TT5 theme that has big fonts as default setting:

| Product | Before | After |
|--------|--------|--------|
| External Product | image | image |
| Simple Product | image | image |
| Grouped Product | image | image | 
| Variable Product | image | image |

Issue #58637, opened 2025-06-09
[dev] Replace actions sending Slack messages with existing monorepo-util


adimoldovan:
`archive/github-actions-slack` is used across many workflows, while there is a \`slack\`  command in the monorepo-utils: [https://github.com/woocommerce/woocommerce/tree/c51c6676145465758e4bbd8f4af64f609884d222/tools/monorepo-utils/src/slack](https://github.com/woocommerce/woocommerce/tree/c51c6676145465758e4bbd8f4af64f609884d222/tools/monorepo-utils/src/slack)
We should replace the usage of these actions with the util, for consistency and better security.
More context: pdV5qK-Te-p2#comment-1388

PR #58655, opened 2025-06-09
[dev] Refactor slack util for simplicity and testability

adimoldovan:

Prerequisite for https://github.com/woocommerce/woocommerce/issues/58637

This PR refactors the Slack utility code in the monorepo utils for improved simplicity and testability. 
The initial scope was to simply use environment variables for Slack token and channels instead of cli arguments, but while working on it I found it appropriate to refactor a bit deeper and add some tests.

Changes:
- Use environment variables for token and channels instead of cli arguments
- Consolidated CLI commands, arguments and options so that logic is not duplicated
- Split code logic into testable functions
- Added unit tests
- Documented the tool

1. Unit tests should obviously pass
2. To test end-to-end you'll need a Slack token. Refer to the readme file for instructions on basic usage.

- text only: p1749484894333559-slack-C02180M7JFM
- file upload: p1749485294673329-slack-C02180M7JFM


prettyboymp (approved):
Changes look good.

Issue #57898, opened 2025-05-12
Move release workflow to GitHub workflow


joshuatf:
Currently our release process is handled internally by a private project that handles releases.  Not only does this reduce transparency around the release, but it also makes it difficult for new release leads who have yet to set this up locally.

This issue will involve moving both RC and final release generation process to GitHub to be run as a workflow.

## Acceptance Criteria

- [ ] The workflow should only be run when approved by those with either admin or maintainer status
- [ ] Any sensitive keys or information should be removed from the workflow before publishing
- [ ] Automatically verify that all CFEs/PRRs have been closed
- [ ] The workflow should produce a GH release and tag, as well as zip of the core plugin with the correct version number

PR #58657, opened 2025-06-09
Allow GH releases to be created from workflow

jorgeatorres:

This PR introduces a few enhancements in the ["build ZIP" workflow](https://github.com/woocommerce/woocommerce/actions/workflows/release-build-zip-file.yml). Specifically:

- Simplifies the language for the various options.
- Fixes a workflow failure when the verification step was skipped.
- Adds support for running the workflow from `trunk` while building a specific release branch, so backporting fixes to this workflow to the release branch will no longer be required before building.
- Allows a draft release to be created from while building the ZIP.

Screenshot 2025-06-12 at 11 06 11

Related to #57898.

1. Create a fork of the WooCommerce repository and ensure that `trunk` is set to this specific branch and that it has at least the `release/9.9` branch (useful for testing the workflow).
   You can achieve the above using the CLI:

   ```
   git remote add my_fork git@github.com:<GITHUB USERNAME>/woocommerce.git
   git push my_fork enhancement/57898:trunk -f
   git push my_fork release/9.9
   ```

   ⚠️ Be careful **not** to push to `origin` but always to `my_fork`.

1. Go to **Actions > Release: Build ZIP file** in your forked repo and run the workflow with the following configurations.

   Ensure the outcome is as expected and in all cases confirm that the ZIP attached to both the workflow and the release (if any) has the correct version.

   |**Source branch**|**Create a draft GitHub release**|**Expected Result**|
   |-|-|-|
   |`release/9.9`|(checked)|Draft 9.9.4 release in **Releases** with attached file `woocommerce.zip`.|
   |`trunk`|(checked)|Should fail because `trunk` is not a release branch.|
   |`my-super-branch`|(checked)|Build fails due to incorrect source branch name.|
   |(empty)|(unchecked)|ZIP attached to workflow as asset should be 10.0.0-dev.<br>No draft release created.|

   I suggest running with _Skip the PR verification step_ checked off to speed up testing. Do at least one run with it unchecked to make sure the checks are correct. In this case you might need to temporarily bump the version to 9.9.4 in `release/9.9` since the checks prevent releasing a tag that already exists on wporg.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/.github/workflows/release-build-zip-file.yml
+++ b/.github/workflows/release-build-zip-file.yml
@@ -212,29 +212,21 @@ jobs:
 
       - name: Create draft release
         id: create_release
-        uses: actions/create-release@v1
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        with:
-          tag_name: ${{ steps.get_version.outputs.version }}
-          release_name: ${{ steps.get_version.outputs.version }}
-          draft: true
-          prerelease: ${{ contains(steps.get_version.outputs.version, '-rc') }}
-          commitish: ${{ inputs.branch || github.ref }}
-          body: ''
-
-      - name: Upload ZIP to release
-        uses: actions/upload-release-asset@v1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        with:
-          upload_url: ${{ steps.create_release.outputs.upload_url }}
-          asset_path: ./woocommerce.zip
-          asset_name: woocommerce.zip
-          asset_content_type: application/zip
-
-      - name: Show link to release.
-        env:
-          RELEASE_URL: ${{ steps.create_release.outputs.html_url }}
+          GH_TOKEN: ${{ github.token }}
+          GH_REPO: ${{ github.repository }}
+          IS_PRERELEASE: ${{ contains(steps.get_version.outputs.version, '-rc') }}
         run: |
-          echo "::notice::Draft release available at ${RELEASE_URL}."
+          FLAGS=""
+          if [ "$IS_PRERELEASE" = "true" ]; then
+            FLAGS="--prerelease=true"
+          fi
+
+          gh release create ${{ steps.get_version.outputs.version }} \
+            './woocommerce.zip' \
+            --title '${{ steps.get_version.outputs.version }}' \
+            --notes '' \
+            --target '${{ inputs.branch || github.ref }}' \
+            --latest=false \
+            --draft \
+            $FLAGS
--- a/.github/workflows/release-build-zip-file.yml
+++ b/.github/workflows/release-build-zip-file.yml
@@ -7,7 +7,7 @@ on:
         required: false
         default: ''
       skip_verify:
-        description: 'Skip the PR verification step.'
+        description: 'Skip verification steps.'
         type: boolean
         required: true
         default: false
--- a/.github/workflows/release-build-zip-file.yml
+++ b/.github/workflows/release-build-zip-file.yml
@@ -35,7 +35,7 @@ jobs:
                   let runBuildZipJob = true;
                   const event = context.payload;
 
-                  if (event.inputs.skip_verify !== false) {
+                  if (event.inputs.skip_verify === 'true') {
                     core.setOutput('runBuildZipJob', runBuildZipJob);
                     console.log('Skipping verification step');
                     return;
--- a/.github/workflows/release-build-zip-file.yml
+++ b/.github/workflows/release-build-zip-file.yml
@@ -84,7 +84,7 @@ jobs:
                   core.setOutput('runBuildZipJob', runBuildZipJob);
 
         - name: Notify Slack on failure
-          if: ${{ steps.verify-prs.outputs.failureMessage != '' }}
+          if: ${{ steps.verify-prs.outputs.failureMessage != '' && inputs.create_github_release }}
           uses: archive/github-actions-slack@f530f3aa696b2eef0e5aba82450e387bd7723903 #v2.0.0
           with:
               slack-bot-user-oauth-access-token: ${{ secrets.CODE_FREEZE_BOT_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release-build-zip-file.yml
```diff
@@ -94,65 +100,67 @@ jobs:
...
+        if: ${{ inputs.create_github_release && ! inputs.skip_verify }}
+        env:
+          BRANCH: ${{ inputs.branch || github.ref }}
         run: |
-            branch_name="${GITHUB_REF#refs/heads/}"
+            # Branch name must be 'release/*'.
+            branch_name="${BRANCH#refs/heads/}"
             if  [[ $branch_name != release/* ]] ; then
-                echo "[ERR] Pre-build verification: the branch name is NOT matching 'release/*' pattern (identified as $branch_name)"
+                echo "::error::Pre-build verification: branch name '$branch_name' is not matching 'release/*' pattern."
```
kalessil, line 122: `::error::` is nice update in the checks :+1:

.github/workflows/release-build-zip-file.yml
```diff
@@ -94,65 +100,67 @@ jobs:
...
-                echo "[OK] Pre-build verification: release version in branch is matching '$version_prefix.*' pattern ($branch_plugin_version)"
             fi
-            if [[ $branch_plugin_version == $svn_plugin_version ]] ; then
-                echo "[ERR] Pre-build verification: release version in branch is matching the one in SVN trunk (identified as $branch_plugin_version)"
-                exit 1
-            else
-                echo "[OK] Pre-build verification: release version in branch is NOT matching the one in SVN trunk ($branch_plugin_version vs $svn_plugin_version)"
+
+            # Release should not already exist on wporg.
+            tag_exists=$(curl -s 'https://api.wordpress.org/plugins/info/1.2/?action=plugin_information&request\[slug\]=woocommerce' | jq "(.versions | has(\"$branch_plugin_version\"))")
```
kalessil, line 135: Good change!

.github/workflows/release-build-zip-file.yml
```diff
@@ -94,65 +100,67 @@ jobs:
   verify-release-versions:
     name: 'Verify release and stable tag versions'
     runs-on: ubuntu-latest
-    if: ${{ inputs.skip_verify == 'false' }}
     steps:
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.branch || github.ref }}
+          path: checkout-branch
```
kalessil, line 107: Can be a bit faster with e.g. (looks like we need the single file from source code):

```
sparse-checkout: |
  plugins/woocommerce/woocommerce.php
```

.github/workflows/release-build-zip-file.yml
```diff
@@ -173,4 +181,56 @@ jobs:
...
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: ./woocommerce.zip
+          asset_name: woocommerce.zip
+          asset_content_type: application/zip
+
+      - name: Show link to release.
+        env:
+          RELEASE_URL: ${{ steps.create_release.outputs.html_url }}
+        run: |
+          echo "::notice::Draft release available at ${RELEASE_URL}."
```
kalessil: Do we need to perform any additional steps with the created pre-release draft?

.github/workflows/release-build-zip-file.yml
```diff
@@ -173,4 +181,56 @@ jobs:
...
+        id: get_version
+        run: |
+          version=$(cat ./woocommerce/woocommerce.php | grep -oP '(?<=Version: )(.+)' | head -n1)
+          echo "version=$version" >> $GITHUB_OUTPUT
+
+      - name: Generate release ZIP file
+        run: |
+          zip -q -9 -r ./woocommerce.zip ./woocommerce/
+
+      - name: Create draft release
```
kalessil, line 213: Is it possible to use GitHub cli for creating the pre-release draft (occasionally GitHub have issues with downloading actions, while the cli is always there)?

kalessil (commented):
The changes look good overall, though I did a code review only yet.

.github/workflows/release-build-zip-file.yml
jorgeatorres: Done.

.github/workflows/release-build-zip-file.yml
jorgeatorres: Updated to use GH CLI and removed the actions for creating the release and uploading the asset.

.github/workflows/release-build-zip-file.yml
```diff
@@ -173,4 +181,56 @@ jobs:
...
+      - name: Get version from plugin file
+        id: get_version
+        run: |
+          version=$(cat ./woocommerce/woocommerce.php | grep -oP '(?<=Version: )(.+)' | head -n1)
+          echo "version=$version" >> $GITHUB_OUTPUT
+
+      - name: Generate release ZIP file
+        run: |
+          zip -q -9 -r ./woocommerce.zip ./woocommerce/
+
```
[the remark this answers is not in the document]
jorgeatorres: This is not correct. `zipfile` is the name of a directory in one of the actions that upload the asset, but GH by default compresses directories into just a ZIP, which is what is used later. The name `zipfile` wouldn't be present anymore.

jorgeatorres:
@kalessil: This is ready for another look whenever you have the chance. Thanks!

.github/workflows/release-build-zip-file.yml
```diff
@@ -94,65 +100,71 @@ jobs:
...
+          ref: ${{ inputs.branch || github.ref }}
+          path: checkout-branch
+          sparse-checkout: |
+            /plugins/woocommerce/woocommerce.php
+            /plugins/woocommerce/readme.txt
+          sparse-checkout-cone-mode: false
+        if: ${{ inputs.create_github_release && ! inputs.skip_verify }}
 
       - name: 'Pre-build verification'
+        if: ${{ inputs.create_github_release && ! inputs.skip_verify }}
```
kalessil, line 115: Looks like both steps have identical if-conditions. How about moving it into the job itself: as per my understanding if the job skipped, it still satisfied `needs: [ verify-prs, verify-release-versions ]` of the next job.

kalessil:
> This is ready for another look whenever you have the chance. Thanks!

Thank you @jorgeatorres. Leaving a single comment and meanwhile going through the testing steps.

.github/workflows/release-build-zip-file.yml
```diff
@@ -2,14 +2,20 @@ name: 'Release: Build ZIP file'
...
-        description: 'By default the zip file is generated from the branch the workflow runs from, but you can specify an explicit reference to use instead here (e.g. refs/tags/tag_name or refs/heads/release/x.x). The resulting file will be available as an artifact on the workflow run.'
+      branch:
+        description: 'Source branch to use. Defaults to the branch the workflow is run from.'
         required: false
         default: ''
       skip_verify:
-        description: 'Skip the PR verification step (default: false) pass true to skip'
-        required: false
-        default: 'false'
+        description: 'Skip the PR verification step.'
```
kalessil: The description is outdated. How about something like `Skip verification steps.`?

kalessil: Also, I think we need to update the workflow name: currently it's saying `Release: Build ZIP file` and not reflecting the GitHub release creation steps.

kalessil (approved):
:heavy_check_mark: manual testing results: works as intended.

Leaving some cleanup comments and pre-approving!

.github/workflows/release-build-zip-file.yml
jorgeatorres: > [...] as per my understanding if the job skipped, it still satisfied `needs: [ verify-prs, verify-release-versions ]` of the next job.

Sadly, that's not the case:

> If a run contains a series of jobs that need each other, a failure or skip applies to all jobs in the dependency chain from the point of failure or skip onwards.

([Source](https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/using-jobs-in-a-workflow#defining-prerequisite-jobs.))

I agree that some reorg can be useful, so if it's ok with you, I can leave this as is and will take a stab at reorganizing in a future PR. Does that sound good to you?

.github/workflows/release-build-zip-file.yml
jorgeatorres: Addressed in a0c7d97964b6b02f960dc46b09e53fbd9eb62628.

.github/workflows/release-build-zip-file.yml
kalessil: Yep, that works for me. This one is pretty much polishing anyway.

.github/workflows/release-build-zip-file.yml
kalessil: Thanks!

PR #58658, opened 2025-06-09
Prevent adding items via StoreAPI with 0 quantity 

opr:

- If quantity isn't supplied during add to cart request, it sets quantity to the minimum allowed
- If quantity is supplied, but is 0, no item is added and the API returns an error

Closes WOOPLUG-4542

1. Using Store API add an item to your cart with no quantity (POST to `/wp-json/wc/store/v1/cart/add-item` with the following body (change id to be a product ID in your store)

 ```json
{
	"id": 15
}
```

2. Repeat this, and expect the item in the returned cart to have a quantity of 2 now (1 item added for each request)
3. Expect the response to show the item to have been added with quantity of 1.
4. POST again, but set quantity to 0

```json
{
	"id": 15,
	"quantity": 0
}
```

5. Expect the API to return an error.
6. Try again with `quantity: "0"` (as a string) ensure the same error is returned.
7. Try with regular, positive integer quantities, expect the items to be added to the cart with the expected quantity.
8. Try with negative integer quantities, arrays, non-numeric strings (e.g. "abc"), etc. expect none of these to work.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/src/StoreApi/Utilities/CartController.php
+++ b/plugins/woocommerce/src/StoreApi/Utilities/CartController.php
@@ -266,8 +266,8 @@ class CartController {
 				'woocommerce_rest_product_invalid_quantity',
 				sprintf(
 					/* translators: %s: product name */
-					__( 'You cannot add &quot;%s&quot; with a quantity of 0 to the cart.', 'woocommerce' ),
-					$product->get_name()
+					esc_html__( 'You cannot add &quot;%s&quot; with a quantity of 0 to the cart.', 'woocommerce' ),
+					esc_html( $product->get_name() )
 				),
 				400
 			);
--- a/plugins/woocommerce/src/StoreApi/Utilities/CartController.php
+++ b/plugins/woocommerce/src/StoreApi/Utilities/CartController.php
@@ -261,12 +261,12 @@ class CartController {
 			$this->throw_default_product_exception( $product );
 		}
 
-		if ( 0 === (int) $request['quantity'] ) {
+		if ( floatval( $request['quantity'] ) <= 0 ) {
 			throw new RouteException(
 				'woocommerce_rest_product_invalid_quantity',
 				sprintf(
 					/* translators: %s: product name */
-					esc_html__( 'You cannot add &quot;%s&quot; with a quantity of 0 to the cart.', 'woocommerce' ),
+					esc_html__( 'You cannot add &quot;%s&quot; with a quantity less than or equal to 0 to the cart.', 'woocommerce' ),
 					esc_html( $product->get_name() )
 				),
 				400
--- a/plugins/woocommerce/src/StoreApi/Utilities/CartController.php
+++ b/plugins/woocommerce/src/StoreApi/Utilities/CartController.php
@@ -261,6 +261,18 @@ class CartController {
 			$this->throw_default_product_exception( $product );
 		}
 
+		if ( 0 === (int) $request['quantity'] ) {
+			throw new RouteException(
+				'woocommerce_rest_product_invalid_quantity',
+				sprintf(
+					/* translators: %s: product name */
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/src/StoreApi/Utilities/CartController.php
```diff
@@ -112,9 +112,14 @@ public function add_to_cart( $request ) {
 			$request['cart_item_data']
 		);
 
-		$this->validate_add_to_cart( $product, $request );
+		$quantity_limits = new QuantityLimits();
+
+		// If quantity was not passed, it should default to the minimum allowed quantity.
```
opr, line 117: The `wp_parse_args` above would not add `quantity => 1` _unless_ quantity had been removed by the `woocommerce_store_api_add_to_cart_data` filter

mikachan (approved):
This is working well for me:

- If I don't set any quantity, 1 item is added to the cart for each request ✅ 
- If I set quantity to 0, I see the new `woocommerce_rest_product_invalid_quantity` error response ✅ 
- If I try again with 0, I see the same new error response again ✅ 
- Positive integer quantities work as expected; the quantities are added to the cart ✅ 
- Negative integer quantities, arrays, and non-numeric strings don't work - I see `rest_invalid_param` ✅ 

Thanks for looking into this so quickly!

plugins/woocommerce/src/StoreApi/Utilities/CartController.php
ralucaStan: It's good that we are setting a quantity, but because we don't consider the total quantity of the same product that might already exist in the cart we could be creating an invalid cart. 

I have two use cases in mind:
- product is sold individually (`$product->is_sold_individually()`)
- product has a maximum that we might be exceeding it if we add another one to the cart

For example, in my testing, I could add a one-time product multiple times to the cart. 

Maybe we could include these checks in the `validate_add_to_cart` function?

ralucaStan (changes requested):
I've tested this following the testing steps and worked as expected.
Additionally I tested with a one time product (see my comment); Also with variations and with batch requests which worked as expected

plugins/woocommerce/src/StoreApi/Utilities/CartController.php
opr: > For example, in my testing, I could add a one-time product multiple times to the cart.

How did you test this, can you share your request? I just tested this out and I see:

```json
{
	"code": "readonly_quantity",
	"message": "The quantity of &quot;Hoodie&quot; cannot be changed",
	"data": {
		"status": 400
	}
}
```

Also the item quantity cannot be modified on the Cart if it is sold individually.

image

> product has a maximum that we might be exceeding it if we add another one to the cart

How did you do this too? I have this snippet:

```php
add_filter( 'woocommerce_store_api_product_quantity_maximum', function( $value, $product, $cart_item ) {
	return 3;
}, 10, 3 );
```

and adding an item more than its maximum quantity causes:

```json
{
	"code": "invalid_quantity",
	"message": "The maximum quantity of &quot;Hoodie&quot; allowed in the cart is 3",
	"data": {
		"status": 400
	}
}
```

ralucaStan: @opr My apologies, I was using the incorrect ID for the individually sold product. Tested again and worked as expected. 

Screenshot 2025-06-11 at 13 29 52


I assumed those checks were missing. But tested again with the quantity filters, also with `woocommerce_store_api_product_quantity_limit`, and those work as expected. 

Thank you for double-checking. I'm approving the PR.

ralucaStan (approved):
Approving the PR, all my testing passed successfully. Good job on the testing coverage.

PR #58676, opened 2025-06-10
Add `Release: Update stable tag` workflow

### LINKED ISSUE #58645: Create a workflow that updates the stable tag
We need to replicate what the `woorelease update-stable` command used to do (example run `php woorelease.php update-stable --product_path=plugins/woocommerce `[`https://github.com/woocommerce/woocommerce/tree/trunk`](https://href.li/?https://github.com/woocommerce/woocommerce/tree/trunk)` 8.2.0)`

See: [https://github.com/woocommerce/woorelease/blob/trunk/includes/command/class-update-stable.php](https://github.com/woocommerce/woorelease/blob/trunk/includes/command/class-update-stable.php)

albarin:

This PR introduces the `Release: Update stable tag` GitHub workflow, which, given a version number, updates the stable tag in SVN and creates two PRs to update the tag in the GitHub repository as well (in trunk and the release branch).

This workflow will replace the current `php woorelease.php update-stable` command.

Closes https://github.com/woocommerce/woocommerce/issues/58645

⚠️ If you want to test on your own Slack workspace, follow the instructions from the Steps to set up a Slack app from https://github.com/woocommerce/woocommerce/pull/58417, otherwise follow these 👇

1. Fork the `woocommerce/woocommerce` repo with all branches into your GH account.
2. In your forked repo, go to `Settings > Secrets and variables > Actions` and create new repository secrets:
- `SVN_REPO_URL`, `SVN_USERNAME`, and `SVN_PASSWORD` (you can get them from the secret store, search for `Flux SVN test repo`).
3. In the forked repo, create a PR with the branch `58645/update-stable-workflow` and merge it (to get this new workflow into `trunk`).
4. Now go to `Actions`, click on the `Release: Update stable tag`, and run the flow with different input combinations, make sure that it fails if the following happens:
- When running with an invalid version format.
- When running without checking the `I confirm...` checkbox.
- When running without the branch `release/x.y` existing.
- When the tag does not exist on SVN.
- When the current stable tag and the input version are the same.
- When the input version is lower than the current stable, and the `Revert` checkbox is unchecked.

5. Run it again and make sure it completes successfully in the following cases 👇  Two PRs should be created (one for trunk and another for the release branch), and the stable tag is updated to the input version in both the PRs and the SVN repository:
- When the input version format is correct, the `I confirm...` checkbox is checked, the branch exists on the repo, the tag exists on SVN, and the input version is bigger than the current.
- When the input version format is correct, the `I confirm...` checkbox is checked, the branch exists on the repo, the tag exists on SVN, and the input version is lower than the current AND the `Revert` checkbox is checked.
 

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

## Summary by CodeRabbit

- **New Features**
  - Introduced a manual workflow to automate updating the stable tag version in plugin readme files on WordPress.org and GitHub.
  - Validates version input and ensures corresponding release branches exist before updating.
  - Automatically commits changes and opens pull requests for stable tag updates on GitHub.

### DIFF
--- a/.github/workflows/release-update-stable-tag.yml
+++ b/.github/workflows/release-update-stable-tag.yml
@@ -222,10 +222,6 @@ jobs:
 
           README_PATH="plugins/woocommerce/readme.txt"
 
-          echo "📋 Current stable tag line on $BRANCH_NAME:"
-          grep "^Stable tag:" "$README_PATH" || echo "::error::Stable tag line not found"
-          echo "=========================="
-
           # Update the stable tag
           sed -i "s/^Stable tag: .*/Stable tag: $TARGET_VERSION/" "$README_PATH"
 
--- a/.github/workflows/release-update-stable-tag.yml
+++ b/.github/workflows/release-update-stable-tag.yml
@@ -17,8 +17,11 @@ permissions:
   pull-requests: write
 
 jobs:
-  validate-confirmation:
+  validate-release:
     runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.extract-version.outputs.version }}
+      branch: ${{ steps.extract-version.outputs.branch }}
     steps:
       - name: Check confirmation
         run: |
@@ -29,13 +32,6 @@ jobs:
             echo "✅ Confirmation received - proceeding with stable tag update"
           fi
 
-  validate-release-version:
-    runs-on: ubuntu-latest
-    needs: validate-confirmation
-    outputs:
-      version: ${{ steps.extract-version.outputs.version }}
-      branch: ${{ steps.extract-version.outputs.branch }}
-    steps:
       - name: Validate release version
         id: extract-version
         uses: actions/github-script@v7
@@ -60,10 +56,6 @@ jobs:
               process.exit(1);
             }
 
-  verify-release-branch:
-    runs-on: ubuntu-latest
-    needs: validate-release-version
-    steps:
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
@@ -71,7 +63,7 @@ jobs:
 
       - name: Verify release branch exists
         run: |
-          RELEASE_BRANCH="${{ needs.validate-release-version.outputs.branch }}"
+          RELEASE_BRANCH="${{ steps.extract-version.outputs.branch }}"
           if ! git ls-remote --heads origin "$RELEASE_BRANCH" | grep -q "refs/heads/$RELEASE_BRANCH"; then
             echo "::error::Release branch '$RELEASE_BRANCH' does not exist"
             exit 1
@@ -81,16 +73,16 @@ jobs:
 
   update-wporg-stable-tag:
     runs-on: ubuntu-latest
-    needs: [validate-release-version, verify-release-branch]
+    needs: validate-release
     steps:
       - name: Install SVN
         run: sudo apt-get install -y subversion
 
       - name: Update SVN stable tag
         env:
-          TARGET_VERSION: ${{ needs.validate-release-version.outputs.version }}
+          TARGET_VERSION: ${{ needs.validate-release.outputs.version }}
         run: |
-          echo "🔄 Checking out SVN repository..."
+          echo "🔄 Checking out SVN trunk..."
           svn checkout "${{ secrets.WPORG_SVN_URL }}/trunk" ./svn-trunk --depth files --username "${{ secrets.WPORG_SVN_USERNAME }}" --password "${{ secrets.WPORG_SVN_PASSWORD }}" --non-interactive
           cd ./svn-trunk
 
@@ -123,12 +115,12 @@ jobs:
 
   update-gh-stable-tag:
     runs-on: ubuntu-latest
-    needs: [validate-release-version, verify-release-branch, update-wporg-stable-tag]
+    needs: [validate-release, update-wporg-stable-tag]
     strategy:
       matrix:
         branch:
           - trunk
-          - ${{ needs.validate-release-version.outputs.branch }}
+          - ${{ needs.validate-release.outputs.branch }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -145,7 +137,7 @@ jobs:
 
       - name: Update stable tag on ${{ matrix.branch }}
         env:
-          TARGET_VERSION: ${{ needs.validate-release-version.outputs.version }}
+          TARGET_VERSION: ${{ needs.validate-release.outputs.version }}
           BRANCH_NAME: ${{ matrix.branch }}
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
--- a/.github/workflows/release-update-stable-tag.yml
+++ b/.github/workflows/release-update-stable-tag.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Check confirmation
         run: |
           if [ "${{ github.event.inputs.confirm-update }}" != "true" ]; then
-            echo "❌ ERROR: You must check the confirmation checkbox to proceed with the stable tag update"
+            echo "::error::You must check the confirmation checkbox to proceed with the stable tag update"
             exit 1
           else
             echo "✅ Confirmation received - proceeding with stable tag update"
@@ -76,7 +76,7 @@ jobs:
         run: |
           RELEASE_BRANCH="${{ needs.validate-release-version.outputs.branch }}"
           if ! git ls-remote --heads origin "$RELEASE_BRANCH" | grep -q "refs/heads/$RELEASE_BRANCH"; then
-            echo "❌ ERROR: Release branch '$RELEASE_BRANCH' does not exist"
+            echo "::error::Release branch '$RELEASE_BRANCH' does not exist"
             exit 1
           else
             echo "✅ Release branch '$RELEASE_BRANCH' exists"
@@ -108,17 +108,20 @@ jobs:
           grep "^Stable tag:" "$README_PATH" || echo "::error::Stable tag line not found"
           echo "=========================="
 
-          if sed -i "s/^Stable tag: .*/Stable tag: $TARGET_VERSION/" "$README_PATH"; then
-            echo "📝 Updated stable tag line:"
-            grep "^Stable tag:" "$README_PATH" || echo "::error::Stable tag line not found"
-            echo "=========================="
+          # Update the stable tag
+          sed -i "s/^Stable tag: .*/Stable tag: $TARGET_VERSION/" "$README_PATH"
 
-            echo "📤 Committing changes to SVN..."
+          echo "📝 Updated stable tag line:"
+          grep "^Stable tag:" "$README_PATH" || echo "::error::Stable tag line not found"
+          echo "=========================="
+
+          # Check if there are any changes
+          if svn diff --quiet; then
+            echo "ℹ️ No changes detected - stable tag is already $TARGET_VERSION"
+          else
+            echo "📤 Changes detected. Committing changes to SVN..."
             svn commit -m "Setting stable tag to $TARGET_VERSION" --username "${{ secrets.WPORG_SVN_USERNAME }}" --password "${{ secrets.WPORG_SVN_PASSWORD }}" --non-interactive
             echo "✅ Successfully committed stable tag update to $TARGET_VERSION"
-          else
-            echo "❌ ERROR: Unable to update stable tag in SVN"
-            exit 1
           fi
 
   update-gh-stable-tag:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-update-stable-tag.yml
```diff
@@ -0,0 +1,149 @@
...
+              git config user.name github-actions
+              git config user.email github-actions@github.com
+
+              update_branch="update/stable-tag-$TARGET_VERSION"
+              git checkout -b "$update_branch"
+              git add "$README_PATH"
+              git commit -m "Update stable tag to $TARGET_VERSION"
+              git push origin "$update_branch"
+
+              gh pr create \
```
jorgeatorres: Not reviewing or anything, but I think we should update the stable tag not only on trunk but also the branch for the version we've just released. This makes things tidy and allows us to release point versions more easily.

albarin: Should we do that by pushing the commit directly to the branch, or with a PR?

jorgeatorres: I don't have a strong preference, but I'm not sure if branch protections will allow a direct push to an existing branch. We could revisit the permissions or just go with a PR for now.

.github/workflows/release-update-stable-tag.yml
```diff
@@ -0,0 +1,192 @@
...
+          else
+            echo "✅ Release branch '$RELEASE_BRANCH' exists"
+          fi
+
+  update-wporg-stable-tag:
+    runs-on: ubuntu-latest
+    needs: [validate-release-version, verify-release-branch]
+    steps:
+      - name: Install SVN
+        run: sudo apt-get update && sudo apt-get install -y subversion
```
jorgeatorres: We possibly don't need to update all packages in the system before running the workflow, but it doesn't hurt to keep it. In any case, sometimes `apt-get update` can also ask interactive questions (particularly when config files are updated), so we should also add the `-y` flag to that command.

.github/workflows/release-update-stable-tag.yml
```diff
@@ -0,0 +1,192 @@
...
+          TARGET_VERSION: ${{ needs.validate-release-version.outputs.version }}
+        run: |
+          echo "🔄 Checking out SVN repository..."
+          svn checkout "${{ secrets.SVN_REPO_URL }}" ./svn-woocommerce-repo --username "${{ secrets.SVN_USERNAME }}" --password "${{ secrets.SVN_PASSWORD }}" --non-interactive
+          cd ./svn-woocommerce-repo
+
+          README_PATH="./trunk/readme.txt"
+
+          if [ ! -f "$README_PATH" ]; then
+            echo "❌ ERROR: readme.txt not found at $README_PATH"
```
jorgeatorres: For real errors I think it'd be best to use proper workflow commands (see https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-error-message), so that any errors are also quickly available on the workflow run on GH.

.github/workflows/release-update-stable-tag.yml
```diff
@@ -0,0 +1,192 @@
...
+            exit 1
+          fi
+
+          echo "📋 Current stable tag line:"
+          grep "^Stable tag:" "$README_PATH" || echo "Stable tag line not found"
```
jorgeatorres: I don't think this will ever happen, but if no stable tag line is found, we probably should exit with an error as well.

.github/workflows/release-update-stable-tag.yml
```diff
@@ -0,0 +1,192 @@
...
+          echo "=========================="
+
+          if sed -i "s/^Stable tag: .*/Stable tag: $TARGET_VERSION/" "$README_PATH"; then
```
jorgeatorres: `sed` will exit with success even if it didn't do anything, so maybe it's not necessary to check the exit status and instead we should use `svn diff` (as coderabbit suggested below) to actually check that there's something to commit.

.github/workflows/release-update-stable-tag.yml
```diff
@@ -0,0 +1,192 @@
...
+              console.log(`Normalized version: ${version}`);
+              console.log(`Release branch: ${branch}`);
+              core.setOutput('version', version);
+              core.setOutput('branch', branch);
+            } else {
+              core.setFailed(`Version ${versionInput} does not match the expected format (X.Y.Z or X.Y)`);
+              process.exit(1);
+            }
+
+  verify-release-branch:
```
jorgeatorres: Given this job has the previous one as a prerequisite and it's only a couple of steps, maybe we can just merge both jobs and perform all validation in one?

I also think validation should include checking that:
- The version we're trying to set exists as a tag in SVN (`svn list <SVN_URL>/tags/<TAG>` can be used for this).
- The stable tag in `trunk`'s readme.txt is different from the version we're trying to set as stable (`svn cat` can be useful to fetch the readme from `trunk`).
  - Under normal circumstances, the new version should be _higher_ than the current one except if we're doing a revert, in which case I think we should have a checkbox to explicitly allow reverts, so that we don't mistakenly set the new stable tag to an old release.

.github/workflows/release-update-stable-tag.yml
```diff
@@ -0,0 +1,192 @@
...
+        with:
+          script: |
+            const versionInput = '${{ github.event.inputs.version }}';
+
+            // Match X.Y.Z or X.Y formats
+            const versionMatch = versionInput.match(/^(\d+\.\d+)(?:\.(\d+))?$/);
+
+            if (versionMatch) {
+              const major = versionMatch[1]; // X.Y
+              const patch = versionMatch[2] || '0'; // Z (default to 0 if not provided)
```
jorgeatorres: Given this is sort of a risky workflow, I think it'd be best to require the full version number and not make any assumptions.

jorgeatorres (changes requested):
Hey @albarin!

Thanks for working on this one. I tested it and it works great 💯, so my feedback can mostly be considered nitpicks.

I left some inline comments, but also would like us to discuss a couple more things:

- In https://github.com/woocommerce/woocommerce/pull/58666, the secrets I'm using are `WPORG_SVN_URL`, `WPORG_SVN_USERNAME` and `WPORG_SVN_PASSWORD`. Happy to adjust my PR if necessary, but we should probably use the same names.
- If we have proper validation in place (which we have) I'm not entirely sure what the value of the dry run is, though I might be missing something. I'd prefer if we later implement [environments](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/managing-environments-for-deployment) for risky workflows instead but for now, maybe a checkbox (unchecked by default) that serves as confirmation that you _indeed_ intend to run the stable tag update should be enough. If not checked, nothing runs at all.
   I believe no harm is done even if something fails while running the workflow, so maybe the dry run is not really necessary.

What do you think?

.github/workflows/release-update-stable-tag.yml
```diff
@@ -0,0 +1,192 @@
...
+    steps:
+      - name: Install SVN
+        run: sudo apt-get update && sudo apt-get install -y subversion
+
+      - name: Update SVN stable tag
+        env:
+          TARGET_VERSION: ${{ needs.validate-release-version.outputs.version }}
+        run: |
+          echo "🔄 Checking out SVN repository..."
+          svn checkout "${{ secrets.SVN_REPO_URL }}" ./svn-woocommerce-repo --username "${{ secrets.SVN_USERNAME }}" --password "${{ secrets.SVN_PASSWORD }}" --non-interactive
```
jorgeatorres: We can use `svn checkout --depth` (see https://svnbook.red-bean.com/en/1.6/svn.advanced.sparsedirs.html) to only retrieve `trunk` or even just the readme.txt. It's not necessary to checkout the full repo.

albarin:
>  * In [Add commit to wporg workflow #58666](https://github.com/woocommerce/woocommerce/pull/58666), the secrets I'm using are `WPORG_SVN_URL`, `WPORG_SVN_USERNAME` and `WPORG_SVN_PASSWORD`. Happy to adjust my PR if necessary, but we should probably use the same names.

I'll update them to have the `WPORG_` prefix 👌 


>  * If we have proper validation in place (which we have) I'm not entirely sure what the value of the dry run is, though I might be missing something. I'd prefer if we later implement [environments](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/managing-environments-for-deployment) for risky workflows instead but for now, maybe a checkbox (unchecked by default) that serves as confirmation that you _indeed_ intend to run the stable tag update should be enough. If not checked, nothing runs at all.
>       I believe no harm is done even if something fails while running the workflow, so maybe the dry run is not really necessary.

Yes, I agree. Since I started replicating the functionality of the `php woorelease.php update-stable` command, I added the same dry-run option it has. But yes, it's not really helpful for us. I will remove it and just add the confirmation checkbox, good suggestion!

.github/workflows/release-update-stable-tag.yml
albarin: Good idea, I've added those two checks as well.

albarin:
@jorgeatorres I think I have addressed all your comments and updated the testing steps. Could you give it another go? TY!

.github/workflows/release-update-stable-tag.yml
```diff
@@ -0,0 +1,268 @@
...
+        env:
+          TARGET_VERSION: ${{ needs.validate-release.outputs.version }}
+        run: |
+          echo "🔄 Checking out SVN trunk..."
+          svn checkout "${{ secrets.WPORG_SVN_URL }}/trunk" ./svn-trunk --depth files --username "${{ secrets.WPORG_SVN_USERNAME }}" --password "${{ secrets.WPORG_SVN_PASSWORD }}" --non-interactive
+          cd ./svn-trunk
+
+          README_PATH="./readme.txt"
+
+          if [ ! -f "$README_PATH" ]; then
```
jorgeatorres, line 154: I guess this check doesn't hurt but my understanding is that for the default shell, execution stops if a command fails, so, in theory, `readme.txt` will exist or otherwise the `svn checkout` command would've failed. I might be wrong though.

.github/workflows/release-update-stable-tag.yml
```diff
@@ -0,0 +1,268 @@
...
+            echo "::error::readme.txt not found at $README_PATH"
+            exit 1
+          fi
+
+          echo "📋 Current stable tag line:"
+          grep "^Stable tag:" "$README_PATH" || echo "::error::Stable tag line not found"
+          echo "=========================="
```
jorgeatorres: Do we really need to print the current stable tag at this point? If we wanted to keep the check then I don't think it should just print an error but exit.

.github/workflows/release-update-stable-tag.yml
```diff
@@ -0,0 +1,268 @@
...
+
+          # Update the stable tag
+          sed -i "s/^Stable tag: .*/Stable tag: $TARGET_VERSION/" "$README_PATH"
+
+          echo "📝 Updated stable tag line:"
+          grep "^Stable tag:" "$README_PATH" || echo "::error::Stable tag line not found"
+          echo "=========================="
```
jorgeatorres: Again, is this output necessary?

.github/workflows/release-update-stable-tag.yml
```diff
@@ -0,0 +1,268 @@
...
+          BRANCH_NAME: ${{ matrix.branch }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          echo "🔄 Processing branch: $BRANCH_NAME..."
+
+          git checkout "$BRANCH_NAME"
+
+          README_PATH="plugins/woocommerce/readme.txt"
+
+          if [ ! -f "$README_PATH" ]; then
```
jorgeatorres: This shouldn't ever happen if the checkout action above succeeded. Maybe this check can be removed?

.github/workflows/release-update-stable-tag.yml
```diff
@@ -0,0 +1,268 @@
...
+            echo "::error::readme.txt not found at $README_PATH"
+            exit 1
+          fi
+
+          echo "📋 Current stable tag line on $BRANCH_NAME:"
```
jorgeatorres: Similar to other cases above: not sure if this check is really necessary. The `readme.txt` should have the correct format. If not, we should probably catch it below when we detect that `git diff` doesn't report any changes. Or, if we consider this bad enough, then beyond just printing the error we should probably stop execution.

.github/workflows/release-update-stable-tag.yml
```diff
@@ -0,0 +1,268 @@
...
+          grep "^Stable tag:" "$README_PATH" || echo "::error::Stable tag line not found"
+          echo "=========================="
+
+          # Check if file was actually modified
+          if git diff --quiet "$README_PATH"; then
+            echo "ℹ️ No changes needed on $BRANCH_NAME - stable tag is already $TARGET_VERSION"
+          else
+            echo "📤 Changes detected. Creating PR for $BRANCH_NAME..."
+
+            UPDATE_BRANCH="update/stable-tag-$TARGET_VERSION-$BRANCH_NAME"
```
jorgeatorres: I'd suggest we add the timestamp to the branch name so it doesn't conflict with an existing branch that maybe wasn't already deleted.

```suggestion
            UPDATE_BRANCH="update/stable-tag-$TARGET_VERSION-$BRANCH_NAME-$(date +%s)"
```

.github/workflows/release-update-stable-tag.yml
```diff
@@ -0,0 +1,268 @@
...
+            git checkout -b "$UPDATE_BRANCH"
+            git add "$README_PATH"
+            git commit -m "Update stable tag to $TARGET_VERSION on $BRANCH_NAME"
+            git push origin "$UPDATE_BRANCH"
+
+            gh pr create \
+              --title "Update stable tag to $TARGET_VERSION ($BRANCH_NAME)" \
+              --body "This PR updates the stable tag to $TARGET_VERSION." \
+              --base "$BRANCH_NAME" \
+              --head "$UPDATE_BRANCH"
```
jorgeatorres: Optional: add `--reviewer ${{ github.actor }}` so that whoever triggers the workflow has to review the resulting PRs 😄.

.github/workflows/release-update-stable-tag.yml
```diff
@@ -0,0 +1,268 @@
...
+
+              for (let i = 0; i < maxLength; i++) {
+                const aPart = aParts[i] || 0;
+                const bPart = bParts[i] || 0;
+
+                if (aPart > bPart) return 1;
+                if (aPart < bPart) return -1;
+              }
+              return 0;
+            };
```
jorgeatorres: Ideally we wouldn't have to write our own comparison. If we performed it at the previous step (in the shell) we could just use PHP's `version_compare()`. If there's a JS package we could use I wouldn't oppose it, as it seems a bit frail to write our own version comparison. This is definitely not a blocker, but something to maybe explore.

If we definitely want to keep this on the JS side and implement it on our own, I think there's room to simplify, given we're sure versions always have the same number of parts (3).

I understand this is a "generic" comparison function, but both versions should have exactly 3 parts so maybe we can simplify a bit.

Maybe rewrite the check below as

```js
if (aPart[i] !== bPart[i] ) {
    return aPart[i] > bPart[i] ? 1 : -1;
}
```

Or even the whole function can be replaced with something like

```js
a.split('.').map( (n, i) => Math.sign(n - b.split('.')[i] ) ).find( n => n ) || 0;
```

What do you think?

.github/workflows/release-update-stable-tag.yml
```diff
@@ -0,0 +1,268 @@
...
+          # Update the stable tag
+          sed -i "s/^Stable tag: .*/Stable tag: $TARGET_VERSION/" "$README_PATH"
+
+          echo "📝 Updated stable tag line:"
+          grep "^Stable tag:" "$README_PATH" || echo "::error::Stable tag line not found"
+          echo "=========================="
+
+          # Check if there are any changes
+          if svn diff --quiet; then
+            echo "ℹ️ No changes detected - stable tag is already $TARGET_VERSION"
```
jorgeatorres: Shouldn't this trigger an error? We've validated the version before so if this happens now something weird happened while the workflow was running.

jorgeatorres (commented):
Hey @albarin!

Tested again and this works great. I left some comments about simplifying things a bit and some light improvements, but this is already working great.

Let me know what you think!

.github/workflows/release-update-stable-tag.yml
albarin: But the `svn checkout` is checking out `trunk`, so it wouldn't fail if there's no `readme.txt`, right?

.github/workflows/release-update-stable-tag.yml
albarin: Yep, agree, I saw you were using `version_compare` in the other workflow while reviewing your PR. Will update to use it here as well. Thanks for the suggestion.

jorgeatorres (approved):
Thank you @albarin! 🫶 
Loved playing with this and updating the stable tag automagically running a single workflow. This automation is 💯.

PR #58684, opened 2025-06-10
Fix setting pickup location causing too many state updates

straku:

Bug introduced in PR https://github.com/woocommerce/woocommerce/pull/57675
Original issue https://github.com/woocommerce/woocommerce/issues/57542

`useEffect` that was supposed to react only to data store changes was fired for local state changes too, caused too many updates to fire that resulted in bad visual flicker, especially on slower connections.

Both recorder with network throttling set to Slow 4G

#### Before

https://github.com/user-attachments/assets/050e5365-2a80-4009-bf49-1ac898f4c635

#### After

https://github.com/user-attachments/assets/db3f84e8-94c8-4ae6-97bf-a058b0448e7c

1. Go to WC → Settings → Shipping → Local pickup.
2. Enable Local Pickup and set up at least two locations.
3. Add an item to your cart and go to the Checkout block.
4. Go to Local Pickup and switch between the locations and verify that the location is properly picked (and that the experience is smooth).

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment

This is a follow-up fix to a fix that is already in `trunk` and will target the same release. The other fix has a changelog entry.

</details>

### DIFF
--- a/plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-pickup-options-block/block.tsx
+++ b/plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-pickup-options-block/block.tsx
@@ -74,7 +74,7 @@ const renderPickupLocation = (
 ): RadioControlOptionType => {
 	const priceWithTaxes = getSetting( 'displayCartPricesIncludingTax', false )
 		? parseInt( option.price, 10 ) + parseInt( option.taxes, 10 )
-		: option.price;
+		: parseInt( option.price, 10 );
 	const location = getPickupLocation( option );
 	const address = getPickupAddress( option );
 	const details = getPickupDetails( option );
@@ -85,7 +85,7 @@ const renderPickupLocation = (
 	let secondaryLabel = <em>{ __( 'free', 'woocommerce' ) }</em>;
 
 	// If there is a cost for local pickup, show the cost per package.
-	if ( parseInt( priceWithTaxes, 10 ) > 0 ) {
+	if ( priceWithTaxes > 0 ) {
 		// If only one package, show the price and not the package count.
 		if ( packageCount === 1 ) {
 			secondaryLabel = (
@@ -144,7 +144,7 @@ const Block = () => {
 	const { shippingRates, selectShippingRate } = useShippingData();
 
 	// Memoize pickup locations to prevent re-rendering when the shipping rates change.
-	const pickupLocations = useMemo( () => {
+	const pickupLocations: CartShippingPackageShippingRate[] = useMemo( () => {
 		return ( shippingRates[ 0 ]?.shipping_rates || [] ).filter(
 			isPackageRateCollectable
 		);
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-pickup-options-block/block.tsx
```diff
@@ -74,7 +74,7 @@ const renderPickupLocation = (
 ): RadioControlOptionType => {
 	const priceWithTaxes = getSetting( 'displayCartPricesIncludingTax', false )
 		? parseInt( option.price, 10 ) + parseInt( option.taxes, 10 )
-		: option.price;
+		: parseInt( option.price, 10 );
```
straku, line 77: Decided to fix some TypeScript errors while being here

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-pickup-options-block/block.tsx
```diff
@@ -85,7 +85,7 @@ const renderPickupLocation = (
 	let secondaryLabel = <em>{ __( 'free', 'woocommerce' ) }</em>;
 
 	// If there is a cost for local pickup, show the cost per package.
-	if ( parseInt( priceWithTaxes, 10 ) > 0 ) {
+	if ( priceWithTaxes > 0 ) {
```
straku, line 88: TypeScript again

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-pickup-options-block/block.tsx
```diff
@@ -144,7 +144,7 @@ const Block = () => {
 	const { shippingRates, selectShippingRate } = useShippingData();
 
 	// Memoize pickup locations to prevent re-rendering when the shipping rates change.
-	const pickupLocations = useMemo( () => {
+	const pickupLocations: CartShippingPackageShippingRate[] = useMemo( () => {
```
straku, line 147: TS party

opr (dismissed):
Hey @straku thanks for working on this! The code changes look fine. Probably need to add a message or comment to the changelog.

Would it be possible to hide/show the descriptions immediately on select? It would be nice to prevent this:

https://www.loom.com/share/411cc51aa1c043d288433f669328ef3d#Edit

I also tested with extensions changing the shipping method in code, if you switch between pickup methods, it's fine, but if the shopper selects pickup, then the extension changes it to a shipping method, then it doesn't work.

This works on trunk.

Here is an example plugin that illustrates it, you'll need to update line 41 of `switch-shipping-methods.php` to use a method ID you have on your site. You can look at the cart response to see what method IDs are available. They are in `shipping_rates[0].shipping_rates`
[switch-shipping-methods.zip](https://github.com/user-attachments/files/20680031/switch-shipping-methods.zip)

straku:
> Probably need to add a message or comment to the changelog.

There was already this as the comment. Should I amend it?

> This is a follow-up fix to a fix that is already in trunk and will target the same release. The other fix has a changelog entry.

opr:
> > Probably need to add a message or comment to the changelog.
> 
> There was already this as the comment. Should I amend it?
> 
> > This is a follow-up fix to a fix that is already in trunk and will target the same release. The other fix has a changelog entry.

Hmm, yeah then that's fine if the changelog job accepts it. I thought the comment would be added in the changelog under `Comment:` but it's OK to leave as is. Thanks!

opr (approved):
Thanks @straku I can actually replicate the bug I found on trunk, so not a regression introduced here. Your changes look good, thanks!

PR #58688, opened 2025-06-10
Don't disable shipping selection while change is processing

straku:

I've noticed that we're showing a disabled state for the selector that is only visual (you can still click on the other option and it will activate) while shipping change is still in-flight. I am proposing to remove it. Also did some more cleanup:
- remove support for passing `className` as we've never used it
- fix some TypeScript issues
- remove unused class names

Recorder with network throttling set to Slow 4G

#### Before

https://github.com/user-attachments/assets/ebbf2620-0155-4dbf-a6c0-eccb4db51bf7

#### After

https://github.com/user-attachments/assets/389d0b22-753e-4d72-a9b1-24951838daac

Prerequisites:
- store with at least two shipping rates configured (you need to switch between them)
- ideally the store should have local pickup enabled with at least two locations to choose from

1. Add some item to the cart and proceed to checkout
2. Input your address to be able to see available shipping rates
3. Go to your browser devtools and enable network throttling (I suggest the equivalent of Slow 4G in Chrome)
4. Switch between the available shipping rates and assess the UX
5. Make sure that the total value is always properly updated (it can lag a bit behind the click but the final value should always reflect the selected rate)
6. Optionally: go to local pickup and verify that the experience is the same as for shipping

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/shipping-rates-control-package/index.tsx
```diff
@@ -137,22 +120,7 @@ export const ShippingRatesControlPackage = ( {
...
-				className,
-				{
-					'wc-block-components-shipping-rates-control__package--disabled':
-						isSelectingRate,
-					'wc-block-components-shipping-rates-control__package--first-selected':
-						! isSelectingRate && selectedOptionNumber === 0,
-					'wc-block-components-shipping-rates-control__package--last-selected':
-						! isSelectingRate &&
-						selectedOptionNumber ===
-							packageData?.shipping_rates?.length - 1,
```
straku, line 152: I am not sure whether we should be worried about it being a potentially breaking change. If so, we can steer folks to use classes in the underlying `RadioControl` component that serve a similar purpose https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/packages/components/radio-control/index.tsx#L41-L45

mikejolley: There are some styles linked to these but I didn't spot a visual regression so I think ok to remove.

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/shipping-rates-control-package/index.tsx
```diff
@@ -19,16 +18,14 @@ import './style.scss';
 
 export const ShippingRatesControlPackage = ( {
 	packageId,
-	className = '',
```
mikejolley: Since this component is shared externally I would keep this for BW compat. Its not being used in subs (https://github.com/woocommerce/woocommerce-subscriptions/blob/14e1cff09c1dc25cd388a6a5e3a72325fa065343/assets/src/js/recurring-packages/index.js#L54) but it stands to reason it could be by other integrations for customisations.

Does it hurt to keep this supported?

mikejolley (changes requested):
It tested as described! I made some notes inline; I don't think it hurts to keep `className` support in place.

One thing i noticed in testing which is not related to this PR but does happen, if you click around but time your clicks, you can select something and then the UI switches between the new selection and a previous selection. I'm not sure if thats fixable but its a rough experience for user. Did you encounter this?

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/shipping-rates-control-package/index.tsx
straku: I wasn't aware that this is used externally, thanks for spotting. I will bring it back.

straku: It's back in https://github.com/woocommerce/woocommerce/pull/58688/commits/b1a8fc2e7e25b6ef0b47cd70511e590e859ccbb2

straku:
> One thing i noticed in testing which is not related to this PR but does happen, if you click around but time your clicks, you can select something and then the UI switches between the new selection and a previous selection. I'm not sure if thats fixable but its a rough experience for user. Did you encounter this?

@mikejolley We fought with something that looks similar to what you describe in https://github.com/woocommerce/woocommerce/issues/57542 (you can see multiple PRs attached). If you'll find a reproduction steps for what you're experiencing, please share, we might need to reopen it once again 😅

mikejolley:
@straku just keep clicking between different methods :D Try to time your input about a second apart. When it happens you'll see the inputs ignore your clicks and toggle randomly. Give it a go and let me know?

straku:
> Give it a go and let me know?

Yeah, I can reproduce 🤦‍♂️ It happens on `trunk` too, I hope we can merge this one regardless of these problems. I will reopen the ticket (again).

straku:
@mikejolley Fixed in https://github.com/woocommerce/woocommerce/pull/59448 😌

mikejolley (approved):
Approving since the other issue is being handled in another PR

PR #58713, opened 2025-06-11
Add fulfillments to customer order details page

### LINKED ISSUE #57424: [Order fulfillments] Update customer orders page
Issue: On the customer's My Account > Orders  page, we need to include fulfillment data on the page that shows the order information. 

Proposed solution: Modify the order details page to display fulfillment data as described in [https://wooorderstatusp2.wordpress.com/2025/01/18/adding-fulfillment-to-my-account/#comment-161](https://wooorderstatusp2.wordpress.com/2025/01/18/adding-fulfillment-to-my-account/#comment-161)

tpaksu:

This PR decorates the customer order details page, and the order tracking page contents with the fulfillment information available. Only the fulfilled fulfillments are displayed on the order information page, draft fulfillments are not shown. This PR also includes a new `FulfillmentUtils` utility class that contain static methods to help extracting the information required from the fulfillments and the order.

Closes #57424 WOOPLUG-4582 WOOPLUG-4019

#### Order details page: 

image

#### Order tracking page: 

image

1. Checkout this branch
2. Run `composer dumpautoload` just in case the `FulfillmentUtils` file is not loaded via the autoloader.
3. Create some fulfillments for an order with refunds, multiple quantities, different fulfillment statuses, different items.
5. Go to the store and open My account > Orders page
6. Open the order details and see the new structure
7. Open an order without fulfillments, see that the old structure is displayed
8. Create a new page, and add `[woocommerce_order_tracking]` shortcode into it
9. Save the page, and navigate to it
10. Use the order ID and the billing email address to open the order tracking information
11. Check the contents and see if it contains fulfillment information
12. Track the order with no fulfillments, see that the old design is applied.
13. Check if the design matches with 2xydCrrB4yb2TAqpXt607y-fi-3274_51696

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 
Modify customer order details page, and the order tracking page to display fulfillment information.
</details>


plugins/woocommerce/src/Internal/Fulfillments/FulfillmentsRenderer.php
```diff
@@ -177,41 +177,6 @@ function ( $provider ) {
 		}
 	}
 
-	/**
```
tpaksu, line 180: Moved to `FulfillmentUtils` file.

plugins/woocommerce/templates/order/order-details-fulfillments.php
```diff
@@ -0,0 +1,257 @@
+<?php
+/**
+ * Order details
+ *
+ * This template can be overridden by copying it to yourtheme/woocommerce/order/order-details.php.
```
ismaeldcom: The path needs to be updated, and in the other template too 🙂

ismaeldcom (commented):
The testing instructions work as expected, but there are some minor design issues and a couple of questions about the implementation.

**Implementation questions**
I wonder if we should use `woocommerce_order_details_before_order_table` to inject the fulfillment details or embed the `orders-details` template in the `order-details-fullfilments` one. Otherwise, we will need to ensure both templates are in sync after every update.

This is more about the project itself, are we implementing fulfillments as an optional feature, or will we auto-fulfill some cases by default? 🤔 I'm thinking about a store with digital products that could end up having `It has no fulfillments yet` on every order. 

**Design issues**
- The issue with Storefront is how the second column is not aligned with the order details one, as the design states, although I don't see this that important.
- With other designs I noticed that the fulfillment status is `bolder` instead of `bold` and doesn't use the `mark` tag. And the table doesn't have a 100% width.

| Storefront | Twenty Twenty-Five |
|--------|--------|
|Screenshot 2025-06-17 at 11 31 42|Screenshot 2025-06-17 at 11 31 54|

tpaksu:
Thanks for the review @ismaeldcom!

> I wonder if we should use woocommerce_order_details_before_order_table to inject the fulfillment details or embed the orders-details template in the order-details-fullfilments one. Otherwise, we will need to ensure both templates are in sync after every update.

That makes sense. I needed the order-details-fulfillments because of the loop inside, we need to show different quantities of some items which exist in a fulfillment, than showing the full order items, so I needed a different loop. For the header part that shows the shipping date, we can use the hook. 

> This is more about the project itself, are we implementing fulfillments as an optional feature, or will we auto-fulfill some cases by default? 🤔 I'm thinking about a store with digital products that could end up having It has no fulfillments yet on every order.

We will be wrapping the project with a feature flag, but I was thinking of doing that after finishing the project work. There are many things to discuss after the implementation finished, such as: 

- UI Design discrepancies
- Missing/wrong/broken parts of project features
- Compatibility with other themes

I'm just now focusing on finishing what's written on the TSD (the created issues), putting the whole work on a test site, share final builds with HE's, designers, stakeholders, external integrators, etc., and planning to do the refinements later.

> The issue with Storefront is how the second column is not aligned with the order details one, as the design states, although I don't see this that important.

> With other designs I noticed that the fulfillment status is bolder instead of bold and doesn't use the mark tag. And the table doesn't have a 100% width.

Thanks for bringing that up, I wasn't sure about how to design this, it is using the design of the orders table (so I don't need to add another set of styles to the plugin), but it's hardcoded as bold. What I did was setting the table header inline style to have normal weight fonts, and then used `<strong>` to make the bold parts on the Figma bold. Maybe the `<strong>` is bolder than expected. I'll check why that table doesn't have 100% width like the orders table.

plugins/woocommerce/templates/order/order-details-fulfillments.php
tpaksu: Fixed in [bc9d95b](https://github.com/woocommerce/woocommerce/pull/58713/commits/bc9d95bec39d60718ea6f54582b4f3a366ca8c23)

ismaeldcom (approved):
Thanks for addressing the comments and providing additional details. I was just curious about the approach, but I agree on focusing on the feature implementation for now 🙂 

It looks good to me and works as expected now.

I still think we should avoid duplicating the order details template. Instead, we could do something like the `$show_downloads` one, rendering it conditionally in the original template. However, this case is much more complex, so I'll leave it up to you to decide whether to iterate on it again at a later stage of the project.

PR #58727, opened 2025-06-11
Add to Cart + Options block: Add e2e test for grouped products

mikachan:

As a follow-up to https://github.com/woocommerce/woocommerce/pull/58002, this adds an e2e test for adding a grouped product to the cart using the Add to Cart With Options blocks.

1. Ensure new test passes in CI.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

Add e2e test for adding a grouped product to the cart using the Add to Cart With Options block

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/client/blocks/tests/e2e/tests/add-to-cart-with-options/add-to-cart-with-options.block_theme.spec.ts
+++ b/plugins/woocommerce/client/blocks/tests/e2e/tests/add-to-cart-with-options/add-to-cart-with-options.block_theme.spec.ts
@@ -186,7 +186,7 @@ test.describe( 'Add to Cart + Options Block', () => {
 
 		await addToCartButton.click();
 
-		await expect( page.getByText( 'Added to cart' ).first() ).toBeVisible();
+		await expect( page.getByText( 'Added to cart' ) ).toBeVisible();
 
 		await expect( page.getByLabel( '4 items in cart' ) ).toBeVisible();
 	} );
--- a/plugins/woocommerce/client/blocks/tests/e2e/tests/add-to-cart-with-options/add-to-cart-with-options.block_theme.spec.ts
+++ b/plugins/woocommerce/client/blocks/tests/e2e/tests/add-to-cart-with-options/add-to-cart-with-options.block_theme.spec.ts
@@ -162,6 +162,41 @@ test.describe( 'Add to Cart + Options Block', () => {
 		} );
 	} );
 
+	test( 'allows adding grouped products to cart', async ( {
+		page,
+		pageObject,
+		editor,
+	} ) => {
+		await pageObject.setFeatureFlags();
+
+		await pageObject.updateSingleProductTemplate();
+
+		await editor.saveSiteEditorEntities( {
+			isOnlyCurrentEntityDirty: true,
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/client/blocks/tests/e2e/tests/add-to-cart-with-options/add-to-cart-with-options.block_theme.spec.ts
```diff
@@ -162,6 +162,41 @@ test.describe( 'Add to Cart + Options Block', () => {
...
+
+		const addToCartButton = page.getByText( 'Add to cart' ).first();
+
+		await addToCartButton.click();
+
+		await page.waitForResponse(
+			( response ) =>
+				response.url().includes( 'wp-json/wc/store/v1/cart' ) &&
+				response.status() === 200
+		);
```
Aljullu: Do we really need to wait for the network request? I think the next line will already handle that for us, as it will wait for the `Added to cart` text to appear on the screen.

plugins/woocommerce/client/blocks/tests/e2e/tests/add-to-cart-with-options/add-to-cart-with-options.block_theme.spec.ts
```diff
@@ -162,6 +162,41 @@ test.describe( 'Add to Cart + Options Block', () => {
...
+
+		await expect( page.getByText( 'Added to cart' ).first() ).toBeVisible();
```
Aljullu: Is `.first()` needed here? I think there is only one `Added to cart` on the page, no? :thinking:

Aljullu (approved):
Thanks for following up and adding the test, @mikachan! It looks good to me, so pre-approving. I left a couple of suggestions which, I think, would help simplify the test while keeping it reliable.

plugins/woocommerce/client/blocks/tests/e2e/tests/add-to-cart-with-options/add-to-cart-with-options.block_theme.spec.ts
mikachan: I was seeing a timeout when checking for `Added to cart` yesterday, but now when I remove this `waitForResponse` check, it's passing fine 🤔 I think it's better without so I've removed it in https://github.com/woocommerce/woocommerce/pull/58727/commits/f19945d58d3bd423b525cbb2f2e1f775e5f733de, and if we see any issues we can always add this back in.

plugins/woocommerce/client/blocks/tests/e2e/tests/add-to-cart-with-options/add-to-cart-with-options.block_theme.spec.ts
mikachan: Yes that's right, thank you! Removed in https://github.com/woocommerce/woocommerce/pull/58727/commits/801c22ea20e9763218307bd1cb0c8bfefd8bb0f9.

PR #58731, opened 2025-06-11
Add initial shipping providers support

### LINKED ISSUE #57404: [Order fulfillments] Add shipping providers support
**Issue:** We need a way to manually and programmatically add shipping providers through a taxonomy form and a filter.

**Proposed solution:** 

* Add a "Shipping providers" taxonomy and initial shipping providers as terms of the taxonomy. You can use WP_Install class to add this taxonomy and terms.
* Make sure the terms have a key, a label, and a logo URL.
* Create an UI on the Shipping settings page that can edit items on the taxonomy.
* Add a `wc_fulfillments_shipping_providers` filter that returns a list of shipping providers in this format: 

```
add_filter('wc_fulfillment_shipping_providers', function($providers){
    $providers['provider_key'] = [
        'key' => 'provider_key',
        'label' => 'Provider Label',
        'logo' => 'https://provider_site.com/favicon.png'|'local image'|null,
    ];
    return $providers;
}
```

tpaksu:

Closes WOOPLUG-3999 #57404

image

1. Checkout this branch
2. Run `composer dumpautoload` just in case the autoloader doesn't catch the new files
3. Rebuild the client assets with `pnpm run watch:build` in `plugins/woocommerce` folder

#### Testing the shipping providers

1. Create a new order if you don't have one
2. On the fulfillments screen, go to manual shipment provider entry form, and click the providers dropdown. It should now show the new providers outlined in https://docs.google.com/spreadsheets/d/1EGdA9QCEh-tLtH_6SYc0dm4wKiYcZ2VGufJ9mBmugl4/edit?usp=sharing
3. Save the tracking information, and see if it displays correctly on the saved fulfillment box

#### Testing the fulfillment status strings

1. Save a draft fulfillment, the badge should now show `Unfulfilled` instead of `unfulfilled` because it now gets the translated strings via `wcFulfillmentSettings.statuses` JS variable export.
2. Fulfill the fulfillment, it should now show "Fulfilled" instead of "fulfilled".

#### Testing the currency symbols

1. Change the store currency to another currency (possibly having a distinctable currency symbol, like  that your payment provider supports, on WooCommerce > Settings > General tab
2. Create an order with that currency
4. See that the currency shows correctly on the fulfillment screen with the correct symbol

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 
Move the shipping providers list from frontend to backend, add additional providers
</details>


plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/components/fulfillments/fulfillment-line-item.tsx
```diff
@@ -53,8 +53,8 @@ export default function FulfillmentLineItem( {
 			return currencyContext.formatAmount( total );
 		}
 
-		// TODO: Find a way to get the currency symbols from the store.
-		const symbol = false;
+		const symbol =
```
samnajian, line 56: Wouldn't the following work?
```
const currency = new wc.currency.CurrencyFactory();
consty currency = c.getCurrencyConfig().symbol;
```
?
Besides wouldn't the following basically do what the current `getFormattedItemTotal` does?

```
currency.formatAmount( total );
```


Do you need to add support for multiple-currency scenario and for this reason the currency is passed down as prop from order? 

Apart from the above, the usage is like the following:
```
	<FulfillmentLineItem
						item={ item.item }
						quantity={ item.selection.length }
						editMode={ editMode }
						currency={ order?.currency ?? '' }
						toggleItem={ handleToggleItem }
						isChecked={ isChecked }
						isIndeterminate={ isIndeterminate }
					/>
```
And I don't see `currency = ''` defaulting the selected currency to shop's currency.

plugins/woocommerce/src/Internal/Fulfillments/FulfillmentsManager.php
```diff
@@ -48,4 +49,28 @@ public function translate_fulfillment_meta_key( $meta_key ) {
 		);
 		return isset( $meta_key_translations[ $meta_key ] ) ? $meta_key_translations[ $meta_key ] : $meta_key;
 	}
+
+	/**
+	 * Get initial shipping providers.
+	 *
+	 * This method returns an array of initial shipping providers that will be available
+	 * in the fulfillment process. It allows for filtering to add or modify the list of
```
samnajian: I guess adding the name of those filter here to this doc block can be useful.

plugins/woocommerce/src/Internal/Fulfillments/FulfillmentsRenderer.php
```diff
@@ -227,14 +230,56 @@ public function render_fulfillment_drawer_slot() {
...
 			return;
 		}
 		WCAdminAssets::register_style( 'fulfillments', 'style', array( 'wp-components' ) );
 		WCAdminAssets::register_script( 'wp-admin-scripts', 'fulfillments', true );
 	}
 
+	/**
+	 * Prints the fulfillments object in the admin header.
+	 */
+	public function print_fulfillments_object() {
```
samnajian, line 365: How come [wp_localize_script](https://developer.wordpress.org/reference/functions/wp_localize_script/) is not used instead?

samnajian (commented):
Functionality is as expected, left some comments.

plugins/woocommerce/src/Internal/Fulfillments/FulfillmentsRenderer.php
tpaksu: It needs a registered script which I don't have :) This is also a valid way of rendering JS objects.

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/components/fulfillments/fulfillment-line-item.tsx
tpaksu: IIRC that had a deprecated notice, and not all currencies were returned. But I'll check again.

tpaksu: > Do you need to add support for multiple-currency scenario and for this reason the currency is passed down as prop from order?

Yes, I realized that on my store, I had orders from WooPayments' multicurrency, and only fulfillments of the orders paid with the store currency rendered correctly. I tried to create the same orders without WooPayments, and changed my store currency to Euro, Pound, Polish Zloti, Japanese Yen, Kazakhstan Tenge etc. And some of them only printed the three-letter ISO code besides the amount. By adding the symbols from the backend, it rendered all correctly.

> And I don't see currency = '' defaulting the selected currency to shop's currency.

Yep, that was something I missed, on the `getFormattedItemTotal` function, I'll add this: 

```
if ( ! orderCurrencyCode ) {
    orderCurrencyCode = storeCurrency?.code || 'USD';
}
```

The code you provided will only work for ~10 currencies, which is hardcoded in the currency package.

plugins/woocommerce/src/Internal/Fulfillments/FulfillmentsRenderer.php
tpaksu: Okay, I managed to convert it to `wp_localize_script`. I just needed to prepend `wc-admin-` to the handle I provided. The replacement is in [71ea793](https://github.com/woocommerce/woocommerce/pull/58731/commits/71ea793ae5fcd44e1217f5753a3c1a3f90c61387)[71ea793](https://github.com/woocommerce/woocommerce/pull/58731/commits/71ea793ae5fcd44e1217f5753a3c1a3f90c61387).

tpaksu: Okay, needed to revert it because it needs a registered script, and we don't register the fulfillment drawer script on the customer area, but we render the settings object. And I can't register a script without a source file, so it's better to keep it like that.

samnajian (approved):
LGTM, thanks for the updates.

plugins/woocommerce/src/Internal/Fulfillments/FulfillmentsRenderer.php
samnajian: I'd suggest creating a follow up issue to address this, as this implementation is not the cleanest I've seen.

tpaksu: Created follow up issue: [WOOPLUG-4693](https://linear.app/a8c/issue/WOOPLUG-4693/order-fulfillments-clean-up-script-loading)

PR #58737, opened 2025-06-11
Sanitize address data only once in CartUpdateCustomer

### LINKED ISSUE #58679: cart/update-customer endpoint escapes HTML twice and leaves WC_Customer in wrong state
### Prerequisites

### Describe the bug

Address data submitted via the checkout block, e.g. billing/shipping address may contain special chars, e.g. `&`. When, for example, submitting a company name like `XY & Z`, the `cart/update-customer` calls `CartUpdateCustomer::get_route_post_response()` which triggers sanitization twice, first on line 135/143 and again in 158/159. This leads to special chars being escaped twice.

`XY & Z` will be sanitized to `XY &amp; Z` after line 143. After running through the sanitization callbacks again in line 158/159 the company name is set to `XY &amp;amp; Z` and that's what is persisted within `WC_Customer`. Any extension/core calling `WC_Customer::get_shipping_company()` (or any other field) afterwards retrieves the wrong value.

### Expected behavior

Special chars should only be escaped/sanitized once and should not break.

### Actual behavior

Special chars are escaped twice and lead to a wrongful state of `WC_Customer`.

### Steps to reproduce

1. Add a snippet to your functions.php to output the value persisted within the WC_Customer object, e.g.: `add_action( 'woocommerce_store_api_cart_update_customer_from_request', function( $customer ) { var_dump( $customer->get_shipping_company() ); exit(); } );`.
2. Open the block-based checkout and dev tools. Edit the shipping address. Use a special char, e.g. `&` within the company field and check the XHR response.

### WordPress Environment

Latest WP + Woo.

### Isolating the problem

mikejolley:

#58679 reported that address data was being sanitized twice which is true and seems to be a mistake. While I was unable to replicate the reported double encoding bug, the sanitization should only run once for other reasons such as performance.

Closes #58679

N/A

Given this is a minor change this won't require specific testing steps and should be covered by CI tests which post and validate customer address data via Store API and Checkout.

If you do want to test this issue specifically,

1. Go to checkout
2. Enter some address data containing `&`.
3. After placing the order, confirm you still see a single `&` and its not visibly escaped.

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>

### DIFF
--- a/plugins/woocommerce/changelog/wooplug-4557-cartupdate-customer-endpoint-escapes-html-twice-and-leaves
+++ b/plugins/woocommerce/changelog/wooplug-4557-cartupdate-customer-endpoint-escapes-html-twice-and-leaves
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Sanitize address data only once in CartUpdateCustomer
--- a/plugins/woocommerce/src/StoreApi/Routes/V1/CartUpdateCustomer.php
+++ b/plugins/woocommerce/src/StoreApi/Routes/V1/CartUpdateCustomer.php
@@ -131,23 +131,9 @@ class CartUpdateCustomer extends AbstractCartRoute {
 		$cart     = $this->cart_controller->get_cart_instance();
 		$customer = wc()->customer;
 
-		// Get data from request object and merge with customer object, then sanitize.
-		$billing  = $this->schema->billing_address_schema->sanitize_callback(
-			wp_parse_args(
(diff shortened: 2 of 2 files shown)


ralucaStan (approved):
I did an analyze of the code and it spotted it as an improvement with no side effects. Tested with an order and worked as expected. 

Approving.

Issue #58622, opened 2025-06-06
[Checkout block] The shipping address form is collapsed on page refresh hiding validation errors in some cases


ralucaStan:
## **Describe the bug**

The shipping address form is collapsed on page refresh hiding validation errors in some cases. Applicable to the billing form as well

## **To Reproduce**

1. As a guest checkout with a product. Store shouldn't default to any country.
2. Enter the first name, address line 1 and last name
3. Refresh page
4. Notice the address form is replaced by the address card
5. Try to place the order
6. Toggle the card, notice the validation errors

## **Expected behavior**

The form should appear if it has required fields that are not filled in. 

## Context: 

The default state for the Address form (collapsed or not) is based on the `address_1`, `first_name` and `last_name` fields. [see code](https://github.com/woocommerce/woocommerce/blob/cabf4583af4b337e1259abec020b595d82adee6d/plugins/woocommerce/client/blocks/assets/js/data/checkout/default-state.ts#L37)

If these fields have values, and we refresh the page, the Address card will appear, but it will be replaced by the form because the form is invalid (not all fields are filled). [This logic](https://github.com/woocommerce/woocommerce/blob/5c0170d5b97ceb4a8559f1cccd84e07b8118fc69/plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-shipping-address-block/customer-address.tsx#L58) does not seem to work on trunk.

We need to improve the detection of the default state. We used to have an isAddressComplete utility ([removed in this PR](https://github.com/woocommerce/woocommerce/pull/58030) because it wasn't used); we can bring it back the part that checked for all the fields and reuse it in the initial state detection

## **Screenshots**

[...]

Issue #58285, opened 2025-05-26
Audit usage of parse_blocks in the checkout area


ralucaStan:
`parse_blocks()` is a heavy tool and we should be using `scan_delimiters()` when trying to inspect the blocks from a template.

Found 4 instances which should be easy to replace once scan_delimiters is available in the monorepo. See [https://github.com/Automattic/jetpack/pull/43648/](https://github.com/Automattic/jetpack/pull/43648/)

[https://github.com/woocommerce/woocommerce/blob/c98e25245462c519b7be0de179d596fd3f87fb73/plugins/woocommerce/src/Blocks/BlockTypes/Checkout.php#L356](https://github.com/woocommerce/woocommerce/blob/c98e25245462c519b7be0de179d596fd3f87fb73/plugins/woocommerce/src/Blocks/BlockTypes/Checkout.php#L356)

[https://github.com/woocommerce/woocommerce/blob/c98e25245462c519b7be0de179d596fd3f87fb73/plugins/woocommerce/src/Blocks/Utils/CartCheckoutUtils.php#L117](https://github.com/woocommerce/woocommerce/blob/c98e25245462c519b7be0de179d596fd3f87fb73/plugins/woocommerce/src/Blocks/Utils/CartCheckoutUtils.php#L117)

[https://github.com/woocommerce/woocommerce/blob/c98e25245462c519b7be0de179d596fd3f87fb73/plugins/woocommerce/src/Blocks/BlockTypes/CheckoutExpressPaymentBlock.php#L88](https://github.com/woocommerce/woocommerce/blob/c98e25245462c519b7be0de179d596fd3f87fb73/plugins/woocommerce/src/Blocks/BlockTypes/CheckoutExpressPaymentBlock.php#L88)

[https://github.com/woocommerce/woocommerce/blob/c98e25245462c519b7be0de179d596fd3f87fb73/plugins/woocommerce/tests/php/src/Blocks/BlockTypes/MiniCart.php#L53](https://github.com/woocommerce/woocommerce/blob/c98e25245462c519b7be0de179d596fd3f87fb73/plugins/woocommerce/tests/php/src/Blocks/BlockTypes/MiniCart.php#L53)

ralucaStan:
p7H4VZ-5l1-p2

pierorocca:
@joshuatf in case your team is starting checkout perf work before we get to this.

joshuatf:
[...]

PR #58782, opened 2025-06-12
Update Checkout block initial render loading

### LINKED ISSUE #57826: [Checkout block] Enable progressive rendering
As part of this issue we are going to:

~~* move away from hydrating the cart data before the initial rendering~~
* remove the Suspense component and switch to rendering a partially loaded checkout
* use the new Skeleton component for sections still fetching data (cart items, shipping options, express payment methods, payment options, order summary)

Things to look out for:

* initial rendering for guests & returning customers

This issue won't handle Partial updates loading states that happen after the initial rendering

### Additional context

* P2: pfHfG4-OB-p2
* Figma: fsQtn884prBGas6UkruLd2-fi

ralucaStan:

This PR refactors the Checkout block loading architecture from component-driven loading to data-driven loading: 

- it eliminates Suspense boundaries and lazy loading of inner components
- it adds immediate component rendering with skeleton states based on data presence (aka cart loading states).

The PR also improves the initial loading of the content in the block by fixing some content jumps and switching to container width mixins instead of styles based on a JS added class after the initial render.

Before the PR we rendered two Suspense components, which impacted the loading of the checkout block:
- One at the Checkout block parent level, [see code](https://github.com/woocommerce/woocommerce/blob/7acddbdf26db0e38263b469250120fc6f28aef4b/plugins/woocommerce/client/blocks/assets/js/base/utils/render-frontend.tsx#L74), which was hidden and resulted in an empty page before the block rendered
- The inner blocks were wrapped in Suspenses without a fallback, [see code](https://github.com/woocommerce/woocommerce/blob/7acddbdf26db0e38263b469250120fc6f28aef4b/plugins/woocommerce/client/blocks/assets/js/atomic/utils/render-parent-block.tsx#L220), which didn't add any value to the loading 

Issues with keeping the Suspense at those levels:
- At the block parent level, the current fallback is a `Loading...` div which replaces the Block HTML structure, which is required for iAPI to function. If we were to fix the Skeletong fallback, we'd have to use a Full Checkout Skeleton fallback component, which is hard to maintain as the checkout block evolves
- At the inner block level: I tried using Suspense fallbacks defined for each child, but this triggers a cascade loading for the checkout block as the components get loaded one by one via lazy loading, which is not a good checkout loading experience. 

Advantages of the approach: 
- Checkout structure appears instantly on page load
- Instead of replacing it with an empty div, we keep the initial DOM structure in the page throughout the block loading process; this enables iAPI usage
- We avoid the need to maintain Skeletons for the whole block, as well for the inner components. We favour skeletons used only for the data-dependent elements from the block
- We' improve the Speed index of the checkout page. The measurements below are made locally, using Lighthouse in the terminal. 

Disadvantages: (generated with Claude 4 Sonnet)
- This change increases the checkout frontend bundle size by 6.22 kB (+81.54%) to 13.8 kB total. While the percentage increase appears significant, the absolute impact is minimal in modern web contexts, equivalent to a single small image.  
For checkout flows, immediate responsiveness is more critical than bundle size optimization. The 13.8 kB total remains well below industry standards (most e-commerce bundles are 300-800 kB), and the performance gains from instant component availability directly improve conversion rates—where every millisecond of delay can impact revenue. Major e-commerce platforms like Amazon and Shopify follow similar patterns, prioritizing checkout responsiveness over bundle size minimization.
- The transition from lazy to direct imports exposes several testing challenges due to immediate component initialization. Previously isolated tests now load checkout dependencies through the import chain. Jest mocks require earlier application since components no longer load lazily, leading to timing issues where real modules initialize before mocks are applied. 

What this PR doesn't include:
- does not handle Express Checkout, will be added separately
- does not refactor the Skeleton components, those will be cleaned in a later PR
- does not fully improve the page placeholder (before the component loads), only adds a style update to match the new placeholders used in the block; there is another ticket to improve the page placeholder

### Performance measurements

I tested the page performance using this command in the terminal:
```
lighthouse 'http://woo-2.local/checkout-link?products=18:1' \
    --throttling-method=provided \
  --screenEmulation.disabled \
  --no-emulatedUserAgent \
  --chrome-flags="--no-sandbox --disable-extensions --disable-http-cache"
```

[Lighthouse report on branch.html.zip](https://github.com/user-attachments/files/20776289/Lighthouse.report.on.branch.html.zip)

[Lighthouse report on trunk.html.zip](https://github.com/user-attachments/files/20776247/Lighthouse.report.on.trunk.html.zip)

| Before | After |
| ------ | ----- |
|   Screenshot 2025-06-17 at 15 24 40 |   Screenshot 2025-06-17 at 15 21 46|

Closes https://github.com/woocommerce/woocommerce/issues/57826
Closes https://linear.app/a8c/issue/WOOPLUG-4214/enable-progressive-rendering-for-the-checkout-block-load

Showcasing the loading of the block using the Performance tab of the browser

| Before | After |
| ------ | ----- |
|   Screenshot 2025-06-17 at 15 46 11 |   Screenshot 2025-06-17 at 15 49 12    |

In your testing to better observe the loading:
- zoom out the window so that the whole block is visible
- simulate slower network connections
- additionally consider commenting hydration to observe the loading of data

In all of the cases make sure there are no regressions. Observe checkout section making sure they load and work as expected.

1. Test the checkout block in the Editor and in the Post Editor
2. Test the checkout block on mobile, desktop and tablet wieport size
3. Test the checkout with guest shopper, returning guest and logged in shopper

### Changelog entry

<details>

#### Significance

-   [ ] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 

</details>


plugins/woocommerce/client/blocks/assets/js/atomic/utils/render-parent-block.tsx
```diff
@@ -216,54 +216,58 @@ const renderInnerBlocks = ( {
 			? blockWrapper
 			: Fragment;
 
-		return (
-			<Suspense
-				key={ `${ block }_${ depth }_${ index }_suspense` }
-				fallback={ <div className="wc-block-placeholder" /> }
+		const isCheckoutBlock = block === 'woocommerce/checkout';
```
ralucaStan: Cart will also be updated. The only block using the suspense after this might be the Mini Carts contents

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/order-summary/index.tsx
```diff
@@ -1,10 +1,13 @@
...
  */
-import { useContainerWidthContext } from '@woocommerce/base-context';
+import {
+	useContainerWidthContext,
+	useStoreCart,
+} from '@woocommerce/base-context';
 import type { CartItem } from '@woocommerce/types';
 import clsx from 'clsx';
-
+import { CartLineItemsSkeleton } from '@woocommerce/base-components/skeleton/patterns/cart-line-items';
```
ralucaStan, line 10: the skeleton components will be cleaned up in another PR

plugins/woocommerce/client/blocks/assets/js/base/components/skeleton/layouts/checkout/index.tsx
```diff
@@ -5,7 +5,7 @@ import { CheckoutOrderSummaryMobileSkeleton } from '../../patterns/checkout-orde
 import { CheckoutContactSkeleton } from '../../patterns/checkout-contact';
 import { CheckoutDeliverySkeleton } from '../../patterns/checkout-delivery';
```
ralucaStan, line 6: I might end up removing most of these components, in a later PR

plugins/woocommerce/client/blocks/assets/js/base/components/state-input/state-input.tsx
```diff
@@ -9,7 +9,6 @@ import { clsx } from 'clsx';
 /**
  * Internal dependencies
  */
-import './style.scss';
```
ralucaStan, line 12: this was referencing the .wc-blocks-components-select__container, so I just removed it and updated the style. there was some content jump happening because of this CSS

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/address-wrapper/index.tsx
```diff
@@ -16,15 +16,18 @@ export const AddressWrapper = ( {
 	isEditing = false,
 	addressCard,
 	addressForm,
+	shouldAnimate = false,
```
ralucaStan, line 19: Temporary fix. The address form gets collapsed if only address_1 and names are filled in. On first render it opens with an animation. I am disabling the animation until this is fixed in https://github.com/woocommerce/woocommerce/issues/58622

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-billing-address-block/block.tsx
```diff
@@ -53,18 +51,11 @@ const Block = (): JSX.Element => {
...
-			cartDataLoaded: store.hasFinishedResolution( 'getCartData' ),
-		};
-	} );
-
 	return (
 		<>
 			<StoreNoticesContainer context={ noticeContext } />
 			<WrapperComponent>
-				{ cartDataLoaded ? <CustomerAddress /> : null }
+				<CustomerAddress />
```
ralucaStan, line 58: we show the empty form before the customer data comes.

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/styles/style.scss
```diff
@@ -1,3 +1,37 @@
+@mixin skeleton-animation {
```
ralucaStan, line 1: This is a temporary improvement of the Skeleton we show until the block loads. I'll reiterate and look for better ways to improve it in a different PR

plugins/woocommerce/client/blocks/assets/js/data/cart/index.ts
```diff
@@ -60,7 +60,7 @@ register( store );
 window.addEventListener( 'load', () => {
 	if (
 		( ! hasCartSession() || persistenceLayer.get() ) &&
-		! isAddingToCart
+		! isAddingToCart()
```
ralucaStan, line 63: A bug I found while debugging. Can be moved to separate PR

plugins/woocommerce/client/blocks/packages/components/totals/item/index.tsx
```diff
@@ -49,12 +51,19 @@ const TotalsItem = ( {
...
 }: TotalsItemProps ): ReactElement => {
+	const { cartIsLoading } = useStoreCart();
+
 	return (
 		<div className={ clsx( 'wc-block-components-totals-item', className ) }>
 			<span className="wc-block-components-totals-item__label">
 				{ label }
 			</span>
-			<TotalsItemValue value={ value } currency={ currency } />
+			{ cartIsLoading ? (
```
ralucaStan: Showing skeletons for prices before cart data comes;

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-fields-block/style.scss
```diff
@@ -1,17 +1,24 @@
 .wc-block-checkout__form {
 	margin: 0;
 	max-width: 100%;
+
+	// Fix content jump for address forms when using select as first input
+	.wc-blocks-components-select {
+		margin-top: $gap;
+	}
```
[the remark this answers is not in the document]
ralucaStan: css class is valid

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/inner-blocks/checkout-shipping-address-block/customer-address.tsx
```diff
@@ -72,17 +73,21 @@ const CustomerAddress = () => {
...
 			addressCard={
 				<AddressCard
 					address={ shippingAddress }
 					target="shipping"
-					onEdit={ () => {
-						setEditing( true );
-					} }
-					isExpanded={ editing }
+					onEdit={ handleEditClick }
+					isExpanded={ true }
```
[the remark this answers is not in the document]
ralucaStan: this doesn't need to reset, the flag needs to be false on initial render

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/product-details/style.scss
```diff
@@ -18,8 +18,10 @@
...
 
-.is-large:not(.wc-block-checkout) {
-	.wc-block-components-product-details__name {
-		font-weight: bold;
+@include cart-checkout-large-container {
+	.wc-block-cart__main {
+		.wc-block-components-product-details__name {
+			font-weight: bold;
+		}
 	}
```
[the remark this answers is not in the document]
ralucaStan: that style was intendet only for the cart block, in the checkout block we highlight prices

plugins/woocommerce/client/blocks/assets/js/atomic/utils/render-parent-block.tsx
```diff
@@ -216,54 +216,58 @@ const renderInnerBlocks = ( {
...
+				</InnerBlockComponentWrapper>
+			</BlockErrorBoundary>
+		);
+
+		// Temporary return until the Cart block is also updated
+		return isCheckoutBlock ? (
+			blockContent
+		) : (
+			<Suspense fallback={ null }>{ blockContent }</Suspense>
 		);
```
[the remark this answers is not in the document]
ralucaStan: fixed

opr:
Hi @ralucaStan noting that this PR caused a notification about a flaky test

https://github.com/woocommerce/woocommerce/issues/55993#event-18155543760

I think it's worth looking into this test and checking if the changes here could have caused the test to become flaky, perhaps some tweaks are needed to the test. I assigned that flaky test issue to you, but let me know what you think. Thanks!

plugins/woocommerce/client/blocks/assets/js/base/utils/render-frontend.tsx
```diff
@@ -69,6 +69,22 @@ export const renderBlock = <
...
+		if ( isCheckoutBlock ) {
+			return (
+				<BlockErrorBoundary { ...errorBoundaryProps }>
+					{ Block && (
+						<Block { ...props } attributes={ attributes } />
+					) }
+				</BlockErrorBoundary>
+			);
+		}
+
```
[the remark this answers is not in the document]
ralucaStan: Keeping lazy loading and suspense at this level would mean having sequential "popping" effect where components appear one by one as they load, which is not a good loading experience for checkout. Suspense is required for lazy loaded components. So the alternative was looking at removing the lazy loaded components.

After some analysis, the impact on the bundle size is minimal. With this approach, we're actually improving the speed score for the page, from 1.7 to 1.1s.
Moving forward, we could consider a hybrid approach (lazy loaded non-essentioal components + essential lazy loaded components) if we ever move to a stepped checkout

ralucaStan:
@coderabbitai update the ai generated summary based on the latest commits.

ralucaStan:
> Hi @ralucaStan noting that this PR caused a notification about a flaky test
> 
> [#55993 (comment)](https://github.com/woocommerce/woocommerce/issues/55993#event-18155543760)
> 
> I think it's worth looking into this test and checking if the changes here could have caused the test to become flaky, perhaps some tweaks are needed to the test. I assigned that flaky test issue to you, but let me know what you think. Thanks!

Thank you for checking @opr . I don't think it's related to my branch. Tested locally and manually and there is no issue. I will try to fix the flakiness this week; it might be coming [from here](https://github.com/woocommerce/woocommerce/blob/2b0202f4395cd29896d881a08cfe9242b419f033/plugins/woocommerce/client/blocks/tests/e2e/utils/frontend/frontend-utils.page.ts#L53); from the look of it the 2nd product isn't added to cart by the time the test reaches the checkout page

plugins/woocommerce/client/blocks/assets/js/data/cart/persistence-layer.ts
```diff
@@ -29,7 +29,7 @@ export const hasCartSession = () => {
 };
 
 export const isAddingToCart = () => {
-	return !! window.location.search.match( /add-to-cart/ );
+	return !! window.location?.search?.match( /add-to-cart/ );
 };
```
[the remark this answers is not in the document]
ralucaStan: fixed in 2a25600cd83260db27e6b9b5c40fb7edfc872fc9

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/test/index.tsx
```diff
@@ -1,22 +1,141 @@
...
  * Internal dependencies
  */
 import TotalsFooterItem from '../index';
 
+jest.mock( '@wordpress/data', () => ( {
+	__esModule: true,
+	...jest.requireActual( '@wordpress/data' ),
+	useSelect: jest.fn(),
+} ) );
+
```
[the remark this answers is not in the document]
ralucaStan: fixed in 69f32cf3327f3c1962a7495bc7c86b3ad7ca9d5f

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/discount/index.tsx
```diff
@@ -30,6 +31,7 @@ export interface TotalsDiscountProps {
 		CartTotalsItem,
 		'total_discount' | 'total_discount_tax'
 	>;
+	isLoading: boolean;
 }
```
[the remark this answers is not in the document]
ralucaStan: fixed in 3880a03b8e4b14273baba285713e67ea08b2630d

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/test/index.tsx
```diff
@@ -1,22 +1,141 @@
...
+// Mock use select so we can override it when wc/store/checkout is accessed, but return the original select function if any other store is accessed.
+wpData.useSelect.mockImplementation(
+	jest.fn().mockImplementation( ( passedMapSelect ) => {
```
ralucaStan, line 24: Needed because TotalsFooterItem component calls hasSelectedShippingRate

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/test/index.tsx
```diff
@@ -1,22 +1,141 @@
...
  * External dependencies
  */
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import { allSettings } from '@woocommerce/settings';
-import { Currency, CartResponseTotals } from '@woocommerce/types';
+import { CurrencyCode } from '@woocommerce/types';
+import * as baseContextHooks from '@woocommerce/base-context/hooks';
+import * as wpData from '@wordpress/data';
+import { previewCart as mockPreviewCart } from '@woocommerce/resource-previews';
```
ralucaStan, line 10: switched this component from snapshot testing to unit tests with assertions

ralucaStan:
@mikejolley the PR is ready for review. All the AI comments have been addressed. Happy to jump on a pair review session and to hear your thoughts on the approach.

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/discount/index.tsx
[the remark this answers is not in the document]
ralucaStan: fixed in 3880a03b8e4b14273baba285713e67ea08b2630d

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/discount/index.tsx
```diff
@@ -119,6 +122,8 @@ const TotalsDiscount = ( {
 					: __( 'Coupons', 'woocommerce' )
 			}
 			value={ discountTotalValue ? discountTotalValue * -1 : '-' }
+			showSkeleton={ isLoading }
+			skeleton={ <Skeleton width="45px" height="1em" /> }
```
mikejolley: I'm curious why you decided to pass the skeleton and showSkeleton props into the component, rather than handling at component level? I'm also wondering if this could be derived instead of being explicitly defined. e.g. if a `TotalsDiscount` component has no value, should it render a placeholder/skeleton, instead of being fed `showSkeleton`?

plugins/woocommerce/client/blocks/assets/js/atomic/utils/render-parent-block.tsx
```diff
@@ -216,53 +216,64 @@ const renderInnerBlocks = ( {
...
+								node.childNodes,
+								blockWrapper
+							)
+						}
+					</InnerBlockComponent>
+				</InnerBlockComponentWrapper>
+			</BlockErrorBoundary>
+		);
+
+		// Temporary return until the Cart block is also updated
```
mikejolley, line 269: When cart is done we'll drop `Suspense`?

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/order-summary/index.tsx
mikejolley: What did you have planned in terms of cleanup?

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/payment-method-label/style.scss
```diff
@@ -8,12 +8,19 @@
 	}
 }
 
-.is-mobile,
-.is-small {
+@mixin small-medium-styles {
```
mikejolley: What is "medium"? Is there a more easy to understand grouping we could use?

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
```diff
@@ -158,6 +159,8 @@ const TotalsFooterItem = ( {
 					) }
 				</>
 			}
+			showSkeleton={ cart.cartIsLoading }
+			skeleton={ <Skeleton width="45px" height="1em" /> }
```
mikejolley: R/e my other comment, not saying this is better, but we could make the Totals components handle skeletons by making them render skeleton when they are provided no data, or marked as loading. Then we won't need to pass through the `<Skeleton>` as a prop. Unless you had a reason to do this?

plugins/woocommerce/client/blocks/assets/js/base/components/skeleton/patterns/cart-line-items/index.tsx
```diff
@@ -8,35 +14,43 @@ import './style.scss';
...
-									<Skeleton maxWidth="173px" />
-									<Skeleton width="78px" />
-								</div>
-							</td>
-							<td className="wc-block-cart-item__total">
-								<Skeleton />
-							</td>
+		<>
+			<VisuallyHidden aria-live="polite">
+				{ __( 'Loading cart line items…', 'woocommerce' ) }
```
mikejolley: If this is user facing maybe "loading your cart"?

plugins/woocommerce/client/blocks/assets/js/base/utils/render-frontend.tsx
```diff
@@ -69,6 +69,20 @@ export const renderBlock = <
 			}
 		}, [] );
 
+		const isCheckoutBlock = container.classList.contains(
+			'wp-block-woocommerce-checkout'
+		);
+
+		// Temporary return until the Cart block is also updated
```
mikejolley, line 76: 👍🏻

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/address-wrapper/style.scss
```diff
@@ -3,10 +3,16 @@
 
 	.wc-block-components-address-card-wrapper,
 	.wc-block-components-address-form-wrapper {
-		transition: all 300ms ease-in-out;
 		width: 100%;
 	}
 
+	&.should-animate {
```
mikejolley: Is should-animate a good classname? I understand it makes sense as a prop, but seems an odd choice for a classname. `is-animated`?

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/discount/index.tsx
ralucaStan: `TotalsItem` is part of the components package. And we use no hooks in that package to get cart data in any of the other components. That's why I decided to handle this at upper levels.

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/styles/style.scss
mikejolley: Does this need adding in a comment?

mikejolley (commented):
Thanks @ralucaStan I've looked file by file and asked a few questions. I've tested logged in/out, desktop, mobile width, editor. Spotted no regressions but difficult to compare. 

Did you have a testing snippet for hydration disabling?

I could definitely see the skeleton in the payments area. That one was the easiest to spot visually, and I like it. I was expecting shipping methods to switch to skeleton on change but it looks like they just fade. Was this intentional to prevent layout shifting? I assume yes.

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/order-summary/index.tsx
ralucaStan: There are a lot of components defined that we don't need. I also had to redo some of them. As I work through the Cart block, I will see what is needed and remove the rest of the skeleton patterns. A good overview of the current Skeletons is on Stylebook.

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/payment-method-label/style.scss
ralucaStan: Fixed in f0626c74f44f71efc468a21da578f65656244477 & ba783e43cdd763d509e51ef146662938b45cdaf7. I should have used mobile instead of medium; mixins are defined here https://github.com/woocommerce/woocommerce/blob/6bf26048344e72fd9eb0c3628048961bf5613711/plugins/woocommerce/client/blocks/assets/css/abstracts/_mixins.scss#L321-L331

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
ralucaStan: Fixed in fddaf88625a0d7660b57f7c214d3a1f4e5c969c0.

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/address-wrapper/style.scss
ralucaStan: Fixed in ff8aea832a12cb25730e7b667649194257dd11a8

plugins/woocommerce/client/blocks/assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
ralucaStan: For some context, I didn't want to add a dependence on another component in TotalsItem. But I saw that we do that for other components from that package.

plugins/woocommerce/client/blocks/assets/js/blocks/checkout/styles/style.scss
ralucaStan: I have a ticket for it in my project https://linear.app/a8c/issue/WOOPLUG-4212/update-the-look-and-feel-of-the-primary-loading-states-for-the; but let me know if you think a comment is necessary at this stage

ralucaStan:
> I was expecting shipping methods to switch to skeleton on change but it looks like they just fade. Was this intentional to prevent layout shifting? I assume yes.

I haven't worked on secondary loading states yet. Those are to be handled in another ticket; this is just for the initial rendering, but there is some handling at the shipping options level.

Did you mean something else?

https://github.com/user-attachments/assets/abc70b9e-dcd5-4cf1-be27-0c98563d9c6f

plugins/woocommerce/client/blocks/assets/js/atomic/utils/render-parent-block.tsx
ralucaStan: Actually, some other components outside of Cart and Checkout blocks are lazy loaded:

- Mini Cart Contents: it makes sense to keep this one lazy loaded, because it's not appearing on load. But I saw @luisherranz created a ticket with a similar goal. https://linear.app/a8c/issue/WOOPLUG-4672/interactivity-api-powered-mini-cart-code-split-the-store;  @luisherranz I expect we're moving away from the current block rendering for the Mini Cart? Asking because we're removing lazy loading Suspense wrappers for Cart and checkout blocks.
- Some Product blocks based on AbstractProductGrid soft-deprecated blocks: https://github.com/woocommerce/woocommerce/pull/52209; we might discuss if it's worth keeping it just for these. In the Cart block we use the Cross-sell block which we also want to deprecate once we remove Suspense from the Cart block.

ralucaStan:
> Did you have a testing snippet for hydration disabling?

No, I just disabled it in [here].(https://github.com/woocommerce/woocommerce/blob/39308d309a563ca735758a2979f71d7da2b3689a/plugins/woocommerce/src/Blocks/BlockTypes/Checkout.php#L538) 

For some context, disabling hydration doesn't bring any speed improvement; the speed index is at 1.2 -1.4 in my testing

plugins/woocommerce/client/blocks/assets/js/base/components/skeleton/patterns/cart-line-items/index.tsx
ralucaStan: Good point! fixed in 39308d309a563ca735758a2979f71d7da2b3689a

ralucaStan:
@mikejolley I'd say this is ready for another review. Thank you so much for your help!

pierorocca:
Kudos @ralucaStan for the Lighthouse performance checks. Awesome improvements in the key metrics and in the UI flows.

Can you help me figure out the discrepancy I'm seeing in the Before screenshots where FCP is 1.9s in the first screenshot but I'm seeing the skeleton at around .7s and checkout rendered at around 1.1s in the second screenshot?

If the After flow, the skeleton doesn't disappear before rendering checkout which is a nice improvement. It's still .9 seconds before it renders. Is there room there for further improvement? On the before the skeleton comes in at .6s.

ralucaStan:
> Can you help me figure out the discrepancy I'm seeing in the Before screenshots where FCP is 1.9s in the first screenshot but I'm seeing the skeleton at around .7s and checkout rendered at around 1.1s in the second screenshot?

@pierorocca I used Ghat GPT by providing the 2 HTML file reports, here are some answers that I think will help:

The Lighthouse-reported FCP of 1.9s in the “Before” report seems higher than the actual perceived paint, because Lighthouse only counts text or image content that the browser rendering pipeline considers significant. In this case, the skeleton loads at ~0.7s and the checkout UI appears by ~1.1s, but those aren’t counted as “first contentful” by Chrome’s paint timing API. This discrepancy is common when skeleton screens or dynamic content are used. Based on the filmstrip, the perceived performance is indeed faster than the reported metric suggests.

Even though the skeleton appears slightly later in the “After” flow (0.9s vs. 0.6s), it’s recognized by Lighthouse as real content because it smoothly transitions into the final UI without removal. This leads to better FCP, LCP, and Speed Index scores, and a better-perceived experience.
The earlier skeleton in the “Before” doesn’t help scores if it’s removed before rendering real content — it’s not counted toward performance metrics and may introduce a perceived layout shift.

> If the After flow, the skeleton doesn't disappear before rendering checkout which is a nice improvement. It's still .9 seconds before it renders. On the before the skeleton comes in at .6s.

The time the skeleton appears can vary slightly between reports (e.g., 0.6s vs. 0.9s) due to natural fluctuations in rendering and network conditions during each Lighthouse run. These differences are normal and expected in synthetic performance testing, which is why we focus more on overall trends and meaningful content rendering (FCP/LCP) rather than exact milliseconds of skeleton appearance.

>  Is there room there for further improvement? 

I didn't touch the backend logic and I am sure we could find some ways to improve the time it takes for the content to be served. One thing that comes to mind is the usage of `parse_blocks()`, which is a costly operation. More details here pcShBQ-2sN-p2. I've also created an issue for this https://github.com/woocommerce/woocommerce/issues/58285

mikejolley:
> I haven't worked on secondary loading states yet. Those are to be handled in another ticket; this is just for the initial rendering, but there is some handling at the shipping options level.

👍🏻 Those were what I meant

Ill test without hydration and approve if all is well.

plugins/woocommerce/client/blocks/assets/js/atomic/utils/render-parent-block.tsx
luisherranz: > I expect we're moving away from the current block rendering for the Mini Cart?

We are, yes. So don't spend any time on the MiniCart at this moment, as it will be replaced with an iAPI-powered version (likely in WooCommerce 10.1).

mikejolley (approved):
I retested this earlier and tried removing hydration which resulted in a more obvious cart line item skeleton display. I can see tests are now passing and everything in CI is green, so I am approving.

pierorocca:
Thanks so much @ralucaStan. That was super helpful in understanding in how Lighthouse is actually measuring FCP.

> The Lighthouse-reported FCP of 1.9s in the “Before” report seems higher than the actual perceived paint, because Lighthouse only counts text or image content that the browser rendering pipeline considers significant. In this case, the skeleton loads at ~0.7s and the checkout UI appears by ~1.1s, but those aren’t counted as “first contentful” by Chrome’s paint timing API.

pierorocca:
> I didn't touch the backend logic and I am sure we could find some ways to improve the time it takes for the content to be served. One thing that comes to mind is the usage of parse_blocks(), which is a costly operation. More details here pcShBQ-2sN-p2. I've also created an issue for this https://github.com/woocommerce/woocommerce/issues/58285I didn't touch the backend logic and I am sure we could find some ways to improve the time it takes for the content to be served. One thing that comes to mind is the usage of parse_blocks(), which is a costly operation. More details here pcShBQ-2sN-p2. I've also created an issue for this https://github.com/woocommerce/woocommerce/issues/58285

Seems like a really good opportunity and low hanging fruit to improve performance @ralucaStan given you've found that it's used 3 times in checkout and once on the mini-cart.

yjailin:
Hello @ralucaStan! 

The initial skeleton design is very different from what was [intended](https://www.figma.com/design/fsQtn884prBGas6UkruLd2/Checkout-Block-Explorations?node-id=413-16157&t=nUxiiiwm4kj5SOxh-1). Is it something meant to be refined later? 

Capture d’écran 2025-06-27 à 12 45 15

Also, it's loading so fast we don't have the time to see it. Is there anything we can do to have a smoother loading that anyone can experience, even with a good connection? [Figma prototype](https://www.figma.com/proto/fsQtn884prBGas6UkruLd2/Checkout-Block-Explorations?page-id=2%3A79135&node-id=176-17178&viewport=-3806%2C-248%2C0.25&t=p4VaB341FXfKaVul-1&scaling=scale-down&content-scaling=fixed&starting-point-node-id=176%3A17178)

My reference here is the Checkout from Shopify (i.e.: allbirds, daytona73)
https://github.com/user-attachments/assets/cfbbb49d-990e-4bbc-9f10-53d0edafaed9

ralucaStan:
> The initial skeleton design is very different from what was fsQtn884prBGas6UkruLd2-fi-413_16157. Is it something meant to be refined later?

I mentioned in the PR description that the PR does not fully improve the page placeholder (before the component loads), only adds a style update to match the new placeholders used in the block; there is another ticket to improve the page placeholder. 

My goals will be to try and be as close as possible to the design, but there are some limits to what we can do there, we can discuss when the time comes.

> Also, it's loading so fast we don't have the time to see it. Is there anything we can do to have a smoother loading that anyone can experience, even with a good connection?
Not that I know of. We can sync next week and I can change my local env to reach a state you need.

PR #58812, opened 2025-06-13
Add filter and action for fulfillment delete

### LINKED ISSUE #57418: [Order fulfillments] Add `wc_fulfillments_before_delete` hook
**Issue:** We need a hook to validate the action before a fulfillment gets deleted and return an error if validation fails. This also can be used to run preliminary actions before a fulfillment is deleted.

**Proposed solution:** We will add a `wc_fulfillments_before_delete` filter that returns an error when something is not right about the deletion. Or will return the `$error` input as is, when everything is correct. See [WOOPLUG-4007](https://linear.app/a8c/issue/WOOPLUG-4007/%5Border-fulfillments%5D-add-wc-fulfillment-before-create-hook) for an example.

### LINKED ISSUE #57419: [Order fulfillments] Add `wc_fulfillments_after_delete` hook
Issue: We need an action that will run after a fulfillment is deleted, to let plugins know about the update.

Proposed solution: We will add a `wc_fulfillments_after_delete` action that runs on the point where a fulfillment deletion is completed.

tpaksu:

This PR adds before and after fulfillment delete hooks to `FulfillmentsDataStore`:

- new filter `wc_fulfillment_before_delete`: Allows modification of the fulfillment before deleting, or interrupting the fulfillment deletion process with an exception. 
- new action `wc_fulfillment_after_delete`: Notifies when a fulfillment is deleted. 

Closes WOOPLUG-4013 WOOPLUG-4014.

Changes should make sense, tests should pass.

#### Preparation

1. Checkout this branch
2. Rebuild assets with `npm run watch:build`
3. Run `composer dumpautoload` just in case the autoloader can't load a file.

#### Testing fulfillment deletion 

1. Add a filter that modifies the fulfillment before delete to your theme's _functions.php_ file (or any other place that will run the code): 

```php
add_filter( 'wc_fulfillment_before_delete', function ( $fulfillment ) {
    $fulfillment->add_meta_data( 'dummy meta', 'test_delete' );
    return $fulfillment;
} );
```

4. Try to delete a fulfillment from the UI
5. It should show the new meta data on the DB meta table after deletion.

#### Testing fulfillment deletion blocking

1. Remove the previous filter.
1. Add a filter that interrupts the deletion of the fulfillment to your theme's _functions.php_ file (or any other place that will run the code): 

```php
add_filter( 'wc_fulfillment_before_delete', function ( $fulfillment ) { 
    throw new FulfillmentException( 'This will show on the UI.' );
} );
```

2. Try to delete a fulfillment from the UI
3. It should show the error on the fulfillment UI.

#### Testing fulfillment after delete hook

1. Remove all previous filters.
2. Add this action to see if the hook works:

```php
add_action( 'wc_fulfillment_after_delete', function ( $fulfillment ) {
   error_log( 'Ran after delete hook', 3, WP_CONTENT_DIR . '/uploads/wc-logs/debug.log' );
} );
```

3. Delete a fulfillment, and see if the log is visible on the debug log, which you can reach from WooCommerce > Status > Logs.

### Changelog entry

<details>

#### Significance

-   [x] Patch
-   [ ] Minor
-   [ ] Major

#### Type

#### Message 

</details>

<details>

#### Comment 
Add before delete filter and after delete action for fulfillments.
</details>


plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/components/action-buttons/remove-button.tsx
```diff
@@ -23,29 +23,31 @@ export default function RemoveButton( {
...
 	const [ isOpen, setOpen ] = useState( false );
 	const openModal = () => setOpen( true );
 	const closeModal = () => setOpen( false );
 
-	const handleDeleteFulfillment = () => {
+	const handleDeleteFulfillment = async () => {
 		setError( null );
 		setIsExecuting( true );
 		if ( ! fulfillment || ! fulfillment.id || ! order || ! order.id ) {
 			setIsExecuting( false );
```
samnajian: I `setIsExecuting( true );` is invoked, and the immediately, if the if state is truthy, it's set to false, can you please either adjust this to remove unnecessary invocation of  `setIsExecuting` or create a follow up issue to address this since it's not added in this PR but still exists in this feature branch?

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/components/action-buttons/remove-button.tsx
```diff
@@ -23,29 +23,31 @@ export default function RemoveButton( {
...
-				setIsEditing( false );
-			} )
-			.catch( ( error ) => {
-				setError( error );
-			} )
-			.finally( () => {
-				setIsExecuting( false );
-			} );
+		await deleteFulfillment( order.id, fulfillment.id, notifyCustomer );
+		const error = getError( order.id );
```
samnajian: [suggestion] Doing `const error = wpSelect( FulfillmentStore ).getError( order.id );` directly here would work as well, na? where `wpSelect` is assigned as `import { select as wpSelect } from '@wordpress/data';`

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/components/action-buttons/remove-button.tsx
```diff
@@ -23,29 +23,31 @@ export default function RemoveButton( {
 	const { order, fulfillment, notifyCustomer } = useFulfillmentContext();
 	const [ isExecuting, setIsExecuting ] = useState< boolean >( false );
 	const { deleteFulfillment } = useDispatch( FulfillmentStore );
+	const { getError } = useSelect(
+		( select ) => ( { getError: select( FulfillmentStore ).getError } ),
+		[]
+	);
```
samnajian: [suggestion] You could directly do the following as well, it's only more concise though:
```
	const { getError } = useSelect(
		( select ) => select( FulfillmentStore ),
		[]
	);
```

samnajian (commented):
While this feature works I observed the followings:

```
add_filter( 'wc_fulfillment_before_delete', function ( $fulfillment ) {
    $fulfillment->add_meta_data( 'dummy meta', 'test_delete' );
    return $fulfillment;
} );

``` 
- Returns error that I can't edit a deleted fulfillment.
- When deletion fails the error text is logged far away from the delete button, so I had to scroll up to find it. This should be delegated to another issue though.

I left some comments, please ping me when I should take another look.

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/components/action-buttons/remove-button.tsx
tpaksu: Well, I'm new to wpData, and I'm using how they described in the docblocks of the methods. What you suggested looks definitely cleaner.

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/components/action-buttons/remove-button.tsx
tpaksu: I can do it in this branch but I'm asking this because I want to understand, what's the downside of it in this code? Will it trigger an extra render?

tpaksu:
Thanks @samnajian for the initial round of review.

> Returns error that I can't edit a deleted fulfillment.

Yea, I need a better way to test, than trying to add a metadata to a deleted fulfillment. Maybe I can use `error_log` or something that doesn't do anything with the fulfillment.

**Edit:** You shouldn't get that in `wc_fulfillment_before_delete` as the fulfillment isn't deleted yet. Are you sure it isn't `wc_fulfillment_after_delete` ?

> When deletion fails the error text is logged far away from the delete button, so I had to scroll up to find it. This should be delegated to another issue though.

I'll ask design about it and see if I can add another error line just before the buttons. @JanaMW27 fyi.

tpaksu:
@samnajian can you do a re-review? I've fixed the issue. It was trying to delete an already deleted fulfillment, which was trying to create the record again 🙂

plugins/woocommerce/client/admin/client/wp-admin-scripts/fulfillments/components/action-buttons/remove-button.tsx
samnajian: It potentially won't trigger an extra render now as the shuffling happens fast enough, but what if later on some other code starts listening to the value of `isExecuting` to perform something?

samnajian (approved):
LGTM, thanks for the updates.

